PR #3613, opened 2023-08-18
[docs] Update the docs on subgraph url overrides

smyrick:
Update the docs to make clear you can set this at "start up time" with variable expansion and link to how you change them at runtime with Rhai or Coprocessors

PR #3652, opened 2023-08-22
docs: add persisted query configuration options info

Meschreiber:
This PR adds information about persisted queries router configuration options. The "security levels" or "modes" were already documented, but this adds reference information about the individual options. I've purposefully put them after the modes, since those are likely to be more useful to readers.

I'm targeting `main` since this is only a docs change. LMK if that's not right.


Geal (approved):

bnjjj (approved):

Meschreiber:
Thanks for reviewing @Geal and @bnjjj . It looks like this and [one other docs PR](https://github.com/apollographql/router/pull/3613) (cc @smyrick ) are failing CI checks and therefore can't be merged. Not sure why `.mdx` only changes would cause that. Would you be able to take a look whenever you get a chance? TIA!

Geal:
@Meschreiber sorry, I reconfirmed out of band: it's fine to push a documentation update directly to main, we'll handle it during the release process

PR #3659, opened 2023-08-24
feat(apollo-router): add `QueryPlan.subgraph_fetches`

### LINKED ISSUE #3658: feat: Expose the number of subgraph fetches for `QueryPlan`s

When writing a plugin with a `execution_service` method, [`execution::Request`](https://docs.rs/apollo-router/1.21.0-alpha.0/apollo_router/services/execution/struct.Request.html) exposes the `QueryPlan`. However, the [`QueryPlan`](https://docs.rs/apollo-router/1.21.0-alpha.0/apollo_router/services/execution/struct.QueryPlan.html) itself is an opaque object that only exposes if this is a mutation or not.

In my case, I would need to access the number of expected fetches for a query plan before the query executes in a custom Rust plugin.

Add a new method on `QueryPlan` that would traverse the query plan and count the number of `Fetch` nodes:

```rust
impl QueryPlan {
    pub fn subgraph_fetches(&self) -> usize {
        // algorithm that traverse the plan here
        todo!()
    }
}
```

`QueryPlan` implements `Serialize`, which can be used to extract the query plan tree and manually parse it. However, this is a brittle implementation that would require manual verification for each Router version release.

**Additional context**

I'm happy to create a PR for this.

nmoutschen:
*Description here*

Fixes #3658

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

Right now, I've set it to take the max possible value for `PlanNode::Condition`. An alternative would be to return the min-max possible values for the number of subgraph fetches instead. Let me know what you folks prefer.

### DIFF
--- a/apollo-router/src/query_planner/plan.rs
+++ b/apollo-router/src/query_planner/plan.rs
@@ -199,8 +199,9 @@ impl PlanNode {
                         .map(|n| n.node.as_ref().map_or(0, |n| n.subgraph_fetches()))
                         .sum::<usize>()
             }
+            // A `SubscriptionNode` makes a request to a subgraph, so counting it as 1
             PlanNode::Subscription { rest, .. } => {
-                rest.as_ref().map_or(0, |n| n.subgraph_fetches())
+                rest.as_ref().map_or(0, |n| n.subgraph_fetches()) + 1
             }
             // Compute the highest possible value for condition nodes
             PlanNode::Condition {
--- a/apollo-router/src/query_planner/plan.rs
+++ b/apollo-router/src/query_planner/plan.rs
@@ -186,6 +186,40 @@ impl PlanNode {
         }
     }
 
+    pub(crate) fn subgraph_fetches(&self) -> usize {
+        match self {
+            PlanNode::Sequence { nodes } => nodes.iter().map(|n| n.subgraph_fetches()).sum(),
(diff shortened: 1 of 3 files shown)


apollo-router/src/query_planner/plan.rs
```diff
@@ -186,6 +186,40 @@ impl PlanNode {
...
+            PlanNode::Flatten(node) => node.node.subgraph_fetches(),
+            PlanNode::Defer { primary, deferred } => {
+                primary.node.as_ref().map_or(0, |n| n.subgraph_fetches())
+                    + deferred
+                        .iter()
+                        .map(|n| n.node.as_ref().map_or(0, |n| n.subgraph_fetches()))
+                        .sum::<usize>()
+            }
+            PlanNode::Subscription { rest, .. } => {
+                rest.as_ref().map_or(0, |n| n.subgraph_fetches())
```
bnjjj, line 203: The primary should be counted as a subgraph fetch I think. It's doing a request to a subgraph

nmoutschen: I've added a `+ 1` and small comment explaining this

o0Ignition0o (approved):

bnjjj (approved):
I don't have the right answer what we should do with the condition node but at least we should document the behavior on the method. What do you think @nmoutschen ?

PR #3669, opened 2023-08-24
move the SpanMetricsExporter to a tracing Layer

Geal:
SpanMetricsExporter is an OpenTelemetry Exporter, which means that if it is loaded, then the entire OpenTelemetry handling infrastructure is loaded, and especially the part that allocates data for the entire list of spans.
Unfortunately, the SpanMetricsExporter was always loaded, even if we do not export the metrics outside of the router, which means that there's a constant overhead of telemetry even when it is not used. This moves the SpanMetricsExporter to a lightweight tracing-subscriber Layer which performs the same busy/idle accounting as OpenTelemetryLayer, then generates the same events as before when the span closes

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/plugins/telemetry/mod.rs
```diff
@@ -624,13 +624,8 @@ impl Telemetry {
...
         // For metrics
-        builder = builder.with_span_processor(
-            BatchSpanProcessor::builder(
-                metrics::span_metrics_exporter::Exporter::default(),
-                opentelemetry::runtime::Tokio,
-            )
-            .build(),
-        );
+
+        //builder = builder.with_simple_exporter(metrics::span_metrics_exporter::Exporter::default());
```
bnjjj, line 628: to delete

bnjjj (approved):

apollo-router/src/plugins/telemetry/metrics/span_metrics_exporter.rs
```diff
@@ -39,55 +37,104 @@ impl Default for Exporter {
...
-                .end_time
-                .duration_since(span.start_time)
-                .unwrap_or_default()
-                .as_secs_f64();
+impl<S> Layer<S> for SpanMetricsLayer
+where
+    S: Subscriber + for<'span> LookupSpan<'span>,
+{
+    fn on_new_span(&self, attrs: &span::Attributes<'_>, id: &span::Id, ctx: Context<'_, S>) {
+        let span = ctx.span(id).expect("Span not found, this is a bug");
```
o0Ignition0o, line 45: Not comfortable with an expect here, are we 100% sure this wont ever ever happen ever for anyone ?

Geal: for some context, this is directly taken from `OpenTelemetryLayer` so thereis no additional risk we are taking right now, since that layer is always active https://github.com/tokio-rs/tracing-opentelemetry/blob/4fa752748116f37af82b59e9b74cb042afa035c1/src/layer.rs#L672

BrynCooke (approved):

PR #3674, opened 2023-08-25
Add logs and extend metrics for `graphql_validation_mode: both`

goto-bus-stop:
This adds logging for query validation errors with either Rust or JS when there is a mismatch, i.e. one of them validates but the other does not. In other cases we are not really interested in the specific error (it will just go back to the user), so we don't need to log there.

To log the Rust validation error well, I now store the ApolloDiagnostics that were produced on `Query{}`. `Query` is serializable for caching, but ApolloDiagnostic is not. Here I just skipped serializing `ApolloDiagnostic` so if `Query` is loaded from cache, it does not have the validation error stored. I'm not sure this is the right thing to do. The ApolloDiagnostics are later used after query planning (which may produce a JS validation error). So it's correct if we can ~safely assume that we only have valid Query instances cached. Otherwise we might get spurious error logs from this.
- [ ] So is that a safe assumption? Reading the CachingQueryPlanner implementation I think it does only store errors (then it's not a `Query` instance) and fully successful planning (then it has run both Rust and JS validation already). So it looks fine, but it could be a bit brittle to rely on this.

I also simplified the validation error printing which
- [x] depends on https://github.com/apollographql/apollo-rs/pull/630.
- [x] and on https://github.com/apollographql/router/pull/3675

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


Geal (approved):

apollo-router/src/error.rs
```diff
@@ -549,26 +557,15 @@ impl std::fmt::Display for ValidationErrors {
...
-            // Best effort to display errors
-            self.errors.iter().for_each(|diag| {
-                println!("{}", diag.data);
-            });
-        };
+        }
+
+        self.errors.iter().for_each(|err| {
+            // Outputs a pretty colourised report on TTYs
+            eprintln!("{err}");
```
BrynCooke, line 567: We should really gate this entire function to test scope. We should never call println or variants in prod code. Otherwise if this is something that should  be used then it should use tracing.

BrynCooke (changes requested):

apollo-router/src/error.rs
goto-bus-stop: Makes sense! I've removed this function as it is only used in one place (supergraph schema validation error) and that already bubbles up the error. The bubbling up lost the actual underlying error message as described in #3681, so I've changed that in 525f376e4d4f288a1750742b427e736ccb88554b.

BrynCooke (approved):

Issue #3686, opened 2023-08-27
Memory leak on router reload


xuorig:
We've recently bumped the router from `1.23.0` to `1.27.0`. After the deployment, our instances started to get killed due to OOM errors within a day. 

<img width="731" alt="Screenshot 2023-08-27 at 6 15 11 PM" src="https://github.com/apollographql/router/assets/1919498/c2b100cf-5f24-4c8a-8ce6-d3e19eb0e4b0">

*(Ignore the temporary drops as these were due to profiling)*

**Here is a heap profile done by enabling jemalloc profile and comparing to base period, about an hour apart: https://github.com/apollographql/router/assets/1919498/26753a47-fa8e-4379-8612-13e370417da6**

Meanwhile my colleague Stephen @fotoetienne bisected to what seems this PR:  https://github.com/apollographql/router/pull/3341

There seems to be something retaining memory between restarts, most likely schema objects / the salsa DB behind apollo-compiler.

I did notice other things:

 - [`Notify` tasks](https://github.com/apollographql/router/blob/dev/apollo-router/src/notification.rs#L559) are being added and never dropped. Running `tokio-console` revealed more than 20 of those tasks on a router running locally for an hour.
 - The only way I see these tasks not [exiting their loop](https://github.com/apollographql/router/blob/dev/apollo-router/src/notification.rs#L638) is if the sender is never dropped.
 - If the sender is never dropped, that probably means `Notify` is not getting dropped either, which means `Configuration` are probably laying around as well.
 - Then I noticed this piece of code: https://github.com/apollographql/router/pull/3341/files#diff-67b9590c5fe973fea43d5bc5a29f37da7b961823665694909f45a00ec50e873dR174-R176
   - The PR added a circular reference. Configuration has a notify object, and notify possibly has a reference to the `Configuration` in its `router_broadcasts`.

Early testing seems to indicate that removing this circular reference at least stops leaking Notify tasks. But I haven't confirmed 100% this is the source of the leak / increased memory usage just yet. Putting my notes out here for now, but will continue investigating tomorrow morning.

BrynCooke:
Thank you for the detailed investigation. We'll get a fix in ASAP.

xuorig:
Thanks @BrynCooke!

Here's what happens when a [`Weak` reference is used instead](https://github.com/apollographql/router/commit/2842b935c8f2205c73dfb093ff41be9c8ae3f918)

[...]

PR #3692, opened 2023-08-28
Fix memory leak caused by Arc circular reference in `Notify`

xuorig:
The issue is described in detail here: https://github.com/apollographql/router/issues/3686

This PR uses `Weak` references to `Configuration` inside `Notify`'s broadcast channel. This was causing a memory leak on router reload.

If the `upgrade` in the `subscription_task` is not successful, we cannot update, so it is simply skipped this time, waiting for the next valid configuration broadcast.

I'm not sure what the best way to test this is within the router test suite. Happy to hear suggestions. I've validated it remediates the memory leak issue on our end.

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/.changesets/fix_arc_config_mem_leak.md
+++ b/.changesets/fix_arc_config_mem_leak.md
@@ -1,4 +1,4 @@
-### Fix memory leak caused by Arc circular reference in `Notify` ([PR #3692](https://github.com/apollographql/router/pull/3692))
+### Fix memory leak caused by Arc circular reference in `Notify` ([Issue #3686](https://github.com/apollographql/router/issues/3686))
 
 A [memory leak](https://github.com/apollographql/router/issues/3686) caused by a [change](https://github.com/apollographql/router/pull/3341) to subscription handling was fixed.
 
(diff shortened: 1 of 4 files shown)


o0Ignition0o (approved):

BrynCooke (approved):

.changesets/fix_arc_config_mem_leak.md
```diff
@@ -0,0 +1,5 @@
+### Fix memory leak caused by Arc circular reference in `Notify` ([PR #3692](https://github.com/apollographql/router/pull/3692))
```
abernix, line 1:
```suggestion
### Fix memory leak caused by Arc circular reference in `Notify` ([Issue #3686](https://github.com/apollographql/router/issues/3686))
```

abernix: Mostly changing in this PR just to trigger CI as me, since it was failing for unclear CircleCI auth/config reasons.  Thanks for the changelog! 😄 😍

PR #3718, opened 2023-09-01
Deal with interfaces on fragment spreads when no __typename is queried

### LINKED ISSUE #2587: Use of `__typename` affects presence of fragment spread data on interfaces
**Describe the bug**

We are seeing a difference in behavior between Apollo Gateway and Router that looks like a bug in Router.
Not sure what the spec really should be, but we have these in the wild.

Given schema:

```
schema {
  query: Query
}

type Query {
  animal: Animal!
  dog: Dog!
}

interface Animal {
  id: String!
}

type Dog implements Animal {
  id: String!
  name: String!
}
```

Issuing the query

```
query dog {
  dog {
    ...on Animal {
      id
      # __typename
      ...on Dog {
        name
      }
    }
  }
}
```

Gives the result

```
{
  "data": {
    "dog": {
      "id": "123"
    }
  }
}
```

Uncomment `__typename` (or use Gateway as the server), and you get

```
{
  "data": {
    "dog": {
      "id": "123",
      "name": "Rover"
    }
  }
}
```

Also, changing the query to not nest will work:

```
query test {
  dog {
    ... on Animal {
      id
    }
    ... on Dog {
      name
    }
  }
}
```

**To Reproduce**
Implement the above schema in a Router service and run the given query(ies).

**Expected behavior**
We expect that the results from Router would be the same as from Gateway.

**Output**
See above.

 - OS: Linux
 - Version: v1.10.3

**Additional context**

o0Ignition0o:
Fix #2587

Operations would over rely on the presence of __typename to resolve selection sets on interface implementers. This changeset checks for the parent type in an InlineFragment, so we don't drop relevant selection set when applicable.


o0Ignition0o:
@Geal @BrynCooke not 100% sure of the change I just made, can we walk through the code together ?

bnjjj (approved):

Issue #3687, opened 2023-08-28
config metric for subscriptions not detecting callback


BrynCooke:
Path should have been preview_callback rather than callback

PR #3724, opened 2023-09-01
Fix metrics attribute types

BrynCooke:
Metrics attributes were being coerced to strings. This is now fixed.
In addition, the logic around types accepted as metrics attributes has been simplified. It will log and ignore values of the wrong type.

Fixes: #3687

@o0Ignition0o I've gone through the code again and think I have caught all metrics, unfortunately it's impossible to test this stuff without setting the global subscriber. As everything has to change in the otel 0.20 upgrade anyway think it's better to just thoroughly review and I'll try and add tests on the new code for the updgrade.

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/plugins/telemetry/metrics/layer.rs
+++ b/apollo-router/src/plugins/telemetry/metrics/layer.rs
@@ -24,8 +24,6 @@ use super::METRIC_PREFIX_HISTOGRAM;
 use super::METRIC_PREFIX_MONOTONIC_COUNTER;
 use super::METRIC_PREFIX_VALUE;
 
-const I64_MAX: u64 = i64::MAX as u64;
-
 #[derive(Default)]
 pub(crate) struct Instruments {
     u64_counter: MetricsMap<Counter<u64>>,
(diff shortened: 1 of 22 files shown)


bnjjj (approved):

apollo-router/src/plugins/telemetry/metrics/layer.rs
```diff
@@ -159,69 +157,247 @@ pub(crate) struct MetricVisitor<'a> {
...
-        self.custom_attributes.push(KeyValue::new(
-            Key::from_static_str(field.name()),
-            Value::from(value.to_string()),
-        ));
+    fn record_i64(&mut self, field: &Field, value: i64) {
+        if let Some(metric_name) = field.name().strip_prefix(METRIC_PREFIX_MONOTONIC_COUNTER) {
+            tracing::error!(
+                metric_name,
+                "monotonic counter must be u64 or f64. This metric will be ignored"
+            );
```
o0Ignition0o, line 203: let's make all of these failfast_debug! instead of tracing::error.

o0Ignition0o (changes requested):
https://github.com/apollographql/router/blob/8e86ef41a3b5eef7f528fc827eeb227f33c27445/apollo-router/src/lib.rs#L23 failfast_debug or failfast_error would allow us to spot attribute type issues.

BrynCooke:
@o0Ignition0o I ended up adding another macro in [cc9cb64](https://github.com/apollographql/router/pull/3724/commits/cc9cb649ec2de45823e986116e9bb72d8f2d29b0)

The behavior is that we will log at debug level and panic on debug builds.

apollo-router/src/plugins/telemetry/metrics/layer.rs
```diff
@@ -24,6 +24,14 @@ use super::METRIC_PREFIX_HISTOGRAM;
 use super::METRIC_PREFIX_MONOTONIC_COUNTER;
 use super::METRIC_PREFIX_VALUE;
 
+macro_rules! log_and_panic_in_debug_build {
+    ($($tokens:tt)+) => {{
+        tracing::debug!($($tokens)+);
+        #[cfg(debug_assertions)]
+        panic!("metric type error, see DEBUG log for details. Release builds will not panic but will still emit an debug log message");
```
o0Ignition0o, line 31:
```suggestion
        panic!("metric type error, see DEBUG log for details. Release builds will not panic but will still emit a debug log message");
```

o0Ignition0o (approved):
Looking great! 1 small nit otherwise LGTM

PR #3735, opened 2023-09-04
Subgraph authentication: Make sure Request signing happens after Compression and APQ

### LINKED ISSUE #3608: Subgraph level APQ should execute before AWS Sigv4 signing
Right now SigV4 signing happens as a subgraph service plugin, so before going through rhai, coprocessor and traffic shaping (dedup, timeout, retry, rate limit), and then into the subgraph service. Inside the subgraph service, there is the APQ implementation.
APQ removes the query from the body and replaces it with a hash, and if APQ is disabled by the subgraph, or not yet available in the subgraph's APQ cache, then a new request with the query is sent. Both queries will modify the body that was signed by SigV4, so signature verification will fail.
And after APQ, the body is compressed.

Here is the order that would make sense:
- deduplication
- timeout: it must cover an entire subgraph call from the point of view of the execution service, even if that means multiple subgraph calls due to retries or APQ). It could be used before deduplication, but if deduplicated queries have the same timeout, then only the first one could reach the limit
- APQ
- compression (happens before retry: no need to do that work multiple times)
- retry
- rate limiting
- AWS SigV4 (the signature holds the current time, so it must change on each retry)

Here we're reaching a design issue: we are missing a service between the subgraph service and the actual HTTP call. By separating in 2 services, we would be able to do:

- subgraph service plugin:
  - deduplication
  - timeout
  - rhai
  - coprocessor
- in subgraph service:
  - APQ
- subgraph HTTP service plugin:
  - compression
  - retry
  - rate limiting
  - AWS SigV4
- in subgraph HTTP service:
  - TLS
  - HTTP call
  
 This would require splitting some parts of the traffic shaping plugin, but at this point I feel they would live better directly in the relevant service instead of all in a layer

o0Ignition0o:
Fix #3608

The router now adds SigningParams to the private context, which the subgraph service can use to sign http calls before the HTTP fetch is made (for websocket connection and regular http calls)


o0Ignition0o:
I'm pretty satisfied with the code, but I need to figure out a proper way to test things.

Given we're using a specific hyper client at the subgraph level, i ll test signingparams in isolation.

apollo-router/src/plugins/authentication/subgraph.rs
```diff
@@ -509,21 +546,22 @@ mod test {
         mock.expect_call()
             .times(1)
             .withf(|request| {
-                let authorization_regex = Regex::new(r"AWS4-HMAC-SHA256 Credential=id/\d{8}/us-east-1/s3/aws4_request, SignedHeaders=content-length;content-type;host;x-amz-content-sha256;x-amz-date, Signature=[a-f0-9]{64}").unwrap();
-                let authorization_header_str = request.subgraph_request.headers().get("authorization").unwrap().to_str().unwrap();
+                let http_request = get_signed_request(request);
+                let authorization_regex = Regex::new(r"AWS4-HMAC-SHA256 Credential=id/\d{8}/us-east-1/s3/aws4_request, SignedHeaders=host;x-amz-content-sha256;x-amz-date, Signature=[a-f0-9]{64}").unwrap();
```
o0Ignition0o, line 550: cc @nmoutschen @BlenderDude I had to change the signedheaders value because we decided to only use the ones that are mandatory (sigv4 is not happy with the connection header, and maybe others).

I'd like to be 100% sure it's not gonna be a problem eventually. What do you both think?

grnikhil:
> Fix #3608
> 
> The router now adds SigningParams to the private context, which the subgraph service can use to sign http calls before the HTTP fetch is made (for websocket connection and regular http calls)

I deep dived the changes and see that now credential provider object is stored in the request context and at the time of signing it fetches the temp aws credentials every time the subgraph service is invoked. Since fetching credentials is an async operation and credentials remain valid for atleast 15 mins, I was wondering if it would optimize the process is we create a lazy static in memory timed cache to store the credentials directly instead of context and only call the credential provider when the value in cache is not found. It might help shave off few ms. I tried using the below crate while playing around the code and was able to init the cache on plugin invocation directly access the credentials in the subgraph_service

https://docs.rs/cached/0.45.1/cached/stores/struct.TimedCache.html

tried something like this to define the cache

lazy_static!{
  static ref CREDENTIAL_CACHE: Mutex<TimedCache<String,SigningConfig>> = Mutex::new(TimedCache::with_lifespan(120));
}

nmoutschen (approved):

o0Ignition0o:
@grnikhil This sounds like a great idea!

Maybe we can do it as a followup? I'd like to land this before we iterate

o0Ignition0o:
@nmoutschen @Meschreiber I just updated the docs to highlight services we have tested SIGV4 against, lmk if you like it! :)

grnikhil:
> @grnikhil This sounds like a great idea!
> 
> Maybe we can do it as a followup? I'd like to land this before we iterate

@o0Ignition0o Yes, I agree. Optimization could go as a follow up. Also wanted to confirm that I tested you changes in our dev environment and was able to successfully get data from our backend

apollo-router/src/services/subgraph_service.rs
```diff
@@ -488,11 +489,28 @@ async fn call_websocket(
...
     let request = get_websocket_request(service_name.clone(), parts, subgraph_cfg)?;
+
     let display_headers = context.contains_key(LOGGING_DISPLAY_HEADERS);
     let display_body = context.contains_key(LOGGING_DISPLAY_BODY);
+
+    let signing_params = {
+        let ctx = context.private_entries.lock();
+        let sp = ctx.get::<SigningParamsConfig>();
+        sp.cloned()
+    };
```
bnjjj, line 500: NIT
```suggestion
    let signing_params = context.private_entries.lock().get::<SigningParamsConfig>().cloned();
```

apollo-router/src/services/subgraph_service.rs
```diff
@@ -674,6 +702,18 @@ async fn call_http(
     let display_headers = context.contains_key(LOGGING_DISPLAY_HEADERS);
     let display_body = context.contains_key(LOGGING_DISPLAY_BODY);
 
+    let signing_params = {
+        let ctx = context.private_entries.lock();
+        let sp = ctx.get::<SigningParamsConfig>();
+        sp.cloned()
+    };
```
bnjjj, line 709: NIT
```suggestion
  let signing_params = context.private_entries.lock().get::<SigningParamsConfig>().cloned();
```

bnjjj (approved):

apollo-router/src/services/subgraph_service.rs
o0Ignition0o: good catch! fixed in [aca6ce3](https://github.com/apollographql/router/pull/3735/commits/aca6ce3cb4ce2c7abc777fd4e1dca4a09e9c64eb)

apollo-router/tests/logging_test.rs
```diff
@@ -179,6 +179,7 @@ async fn simple_query_should_display_logs_for_subgraph_and_supergraph() {
         logging_count.count(event.fields());
     }
 
+    dbg!(&logging_count);
```
BrynCooke, line 182: nit tests should not log by default

BrynCooke (approved):

apollo-router/tests/logging_test.rs
o0Ignition0o: oh ok!

o0Ignition0o: fixed in 698b6554

Geal (approved):
that looks way better than the previous code: one less checkpoint_async, and the signing code is extracted in a function :+1:

PR #3736, opened 2023-09-04
Ensure release build tag matches Cargo.toml version

### LINKED ISSUE #3664: Release process should check project files to see that versions match the git tag
This way we prevent releases where the wrong branch was tagged.
We can probably get away with just checking one location.

BrynCooke:
When release builds are started that are not nightly, the tag name must match that of the Cargo.toml

Fixes #3664

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


.circleci/config.yml
```diff
@@ -543,6 +543,21 @@ jobs:
...
+                command: >
+                  # Validate that version in the Cargo.toml matches the tag version
+                  # Fail the release if the versions don't match
+                  # Note that the tag has the prefix `v`, so this needs to be used to prefix the VERSION variable.
+                  VERSION=v$(cargo metadata --format-version=1 --no-deps | jq --raw-output '.packages[0].version')
+                  if [[ "${VERSION}" != "${CIRCLE_TAG}" ]]; then
+                    echo "Version in Cargo.toml (${VERSION}) does not match tag (${CIRCLE_TAG})"
+                    exit 1
+                  fi
+
```
abernix, line 560: This will run X separate runs for each build of the same release.  Is there any easy way we could make this run in a pre-flight job in front of all the release jobs?  We have several things we need to check, and it would make sense to club them together.

.circleci/config.yml
```diff
@@ -543,6 +543,21 @@ jobs:
...
+      - when:
+          condition:
+            not: << pipeline.parameters.nightly >>
+          steps:
+            - run:
+                command: >
+                  # Validate that version in the Cargo.toml matches the tag version
+                  # Fail the release if the versions don't match
+                  # Note that the tag has the prefix `v`, so this needs to be used to prefix the VERSION variable.
+                  VERSION=v$(cargo metadata --format-version=1 --no-deps | jq --raw-output '.packages[0].version')
```
abernix, line 555: This should be quoted, per best practices, since it could end up with unknown input.

```suggestion
                  VERSION="v$(cargo metadata --format-version=1 --no-deps | jq --raw-output '.packages[0].version')"
```

abernix: Are we sure that `jq` exists here?

abernix, line 555: Just as a thought: Is there a reason this can't be an `xtask` command?  I feel like a series of commands that are written (**and tested**) in Rust, including being able to run them locally, is more manageable than just adding more and more to our CircleCI file.

.circleci/config.yml
abernix: If the answer is "yes, but let's move it out later", then that's fine.  I just started introducing the "Existing image" check last week, and it would be best positioned _before_ starting 4 (somewhat expensive) jobs.

.circleci/config.yml
BrynCooke: Let's talk about this.
I actually don't think xtask is a good build tool in general, it's very high cost to add simple features, adds a ton to our build times and doesn't have many of the basic features that a build tool should have.

I understand the concern about being stuck with circle-ci, but in that case we could consider: https://github.com/sagiegurari/cargo-make

BrynCooke:
Closing so that it can be rewrittten as an xtask

PR #3738, opened 2023-09-04
Subscriptions: Correct v1.28.x regression allowing panic via un-named subscription operation

o0Ignition0o:
This change allows the contexts OPERATION_NAME to be set to None.

### DIFF
--- a/apollo-router/src/services/subgraph_service.rs
+++ b/apollo-router/src/services/subgraph_service.rs
@@ -431,6 +431,13 @@ async fn call_websocket(
     subgraph_cfg: &WebSocketConfiguration,
     subscription_hash: String,
 ) -> Result<SubgraphResponse, BoxError> {
+    let operation_name = request
+        .subgraph_request
+        .body()
+        .operation_name
+        .clone()
+        .unwrap_or_default();
+
     let SubgraphRequest {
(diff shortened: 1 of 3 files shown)


BrynCooke (approved):

apollo-router/src/services/subgraph_service.rs
```diff
@@ -431,6 +431,13 @@ async fn call_websocket(
     subgraph_cfg: &WebSocketConfiguration,
     subscription_hash: String,
 ) -> Result<SubgraphResponse, BoxError> {
+    let operation_name = request
+        .subgraph_request
+        .body()
+        .operation_name
```
Geal, line 437: is that really the expected behaviour here? The operation name that was set in the subgraph request is defined by query planner, so it will be different from the operation name extracted in query analysis

Geal: oh I see, that was introduced in https://github.com/apollographql/router/commit/ed6a9fb94b024f8144942cecf4ccc3c4ab66f902#diff-6ef5a208ca8622f30eef88f75c18566e0304d59856b66293dcd6811555e6382e

Geal (approved):

PR #3748, opened 2023-09-05
fix: handle correctly multipart stream if the original stream is empty

### LINKED ISSUE #3293: Test: add test with empty stream for multipart protocol

bnjjj:
Fixes #3293


Geal (approved):

apollo-router/src/protocols/multipart.rs
```diff
@@ -205,4 +234,35 @@ mod tests {
...
+            } else {
+                match curr_index {
+                    0 => {
+                        assert_eq!(
+                            res,
+                            "\r\n--graphql\r\ncontent-type: application/json\r\n\r\n{}\r\n--graphql--\r\n"
+                        );
+                    }
+                    _ => {
+                        panic!("should not happened, test failed");
```
o0Ignition0o, line 261:
```suggestion
                        panic!("should not happen, test failed");
```

o0Ignition0o (approved):
Let's make sure it works end2end with a client, otherwise lgtm!

PR #3752, opened 2023-09-05
test(subscription): add integration test for subscription

### LINKED ISSUE #3740: Anonymous subscriptions: write a regression test
Followup to #3738, we need to add a regression test that ensures an unnamed subscription makes it through the request pipeline.

bnjjj:
Add a regression test that ensures an unnamed subscription makes it through the request pipeline.

Fixes #3740

### DIFF
--- a/apollo-router/tests/subscription_load_test.rs
+++ b/apollo-router/tests/subscription_load_test.rs
@@ -10,8 +10,24 @@ use crate::common::Telemetry;
 mod common;
 
 const SUBSCRIPTION_CONFIG: &str = include_str!("fixtures/subscription.router.yaml");
-const SUB_QUERY: &str = r#"subscription {  userWasCreated {    name reviews { body } }}"#;
+const SUB_QUERY: &str =
+    r#"subscription {  userWasCreated(intervalMs: 5, nbEvents: 10) {    name reviews { body } }}"#;
 const UNFEDERATED_SUB_QUERY: &str = r#"subscription {  userWasCreated { name username }}"#;
+
+#[tokio::test(flavor = "multi_thread")]
+async fn test_subscription() -> Result<(), BoxError> {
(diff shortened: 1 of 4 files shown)


o0Ignition0o:
Looking great!

apollo-router/tests/subscription_load_test.rs
```diff
@@ -10,9 +10,25 @@ use crate::common::Telemetry;
...
+async fn test_subscription() -> Result<(), BoxError> {
+    let mut router = create_router(SUBSCRIPTION_CONFIG).await?;
+    router.start().await;
+    router.assert_started().await;
+
+    let (_, response) = router.run_subscription(SUB_QUERY).await;
+    assert!(response.status().is_success());
+
+    let mut stream = response.bytes_stream();
+    while let Some(_chunk) = stream.next().await {}
```
Geal, line 27: maybe add a snapshot to guarantee that the response stays of the right shape?

apollo-router/tests/subscription_load_test.rs
```diff
@@ -10,9 +10,25 @@ use crate::common::Telemetry;
 mod common;
 
 const SUBSCRIPTION_CONFIG: &str = include_str!("fixtures/subscription.router.yaml");
-const SUB_QUERY: &str = r#"subscription {  userWasCreated {    name reviews { body } }}"#;
+const SUB_QUERY: &str =
+    r#"subscription {  userWasCreated(intervalMs: 5, nbEvents: 10) {    name reviews { body } }}"#;
 const UNFEDERATED_SUB_QUERY: &str = r#"subscription {  userWasCreated { name username }}"#;
 
+#[tokio::test(flavor = "multi_thread")]
+async fn test_subscription() -> Result<(), BoxError> {
```
abernix, line 18: Look for `TEST_GRAPH_REF` and friends in the existing code!

PR #3764, opened 2023-09-06
Add apollo.router.id to otlp metrics metadata

BrynCooke:
Currently apollo metrics may have the same metadata between multiple routers. This means that we can't disambiguate when multiple routers are sending data.

Add a random UUID to the metadata.

This has no user facing impact.

Note: there is no unit test or integration test for this. We need to take a broader look at how we do telemetry testing, and the solution should probably involve otel collector. However that is beyond the scope of this ticket.

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/plugins/telemetry/metrics/apollo.rs
+++ b/apollo-router/src/plugins/telemetry/metrics/apollo.rs
@@ -1,6 +1,7 @@
 //! Apollo metrics
 use std::sync::atomic::AtomicBool;
 use std::sync::atomic::Ordering;
+use std::sync::OnceLock;
 use std::time::Duration;
 
 use opentelemetry::sdk::export::metrics::aggregation;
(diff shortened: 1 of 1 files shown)


apollo-router/src/plugins/telemetry/metrics/apollo.rs
```diff
@@ -31,6 +33,9 @@ fn default_buckets() -> Vec<f64> {
     ]
 }
 
+// Random unique UUID fot the Router. This doesn't actually identify the router, it just allows disambiguation between multiple routers with the same metadata.
```
bkinzle, line 36:
```suggestion
// Random unique UUID for the Router. This doesn't actually identify the router, it just allows disambiguation between multiple routers with the same metadata.
```

PR #3773, opened 2023-09-06
fix(subscription): force the deduplication to be enabled by default as it's documented

bnjjj:
`subscription.enable_deduplication` was documented to be `true` by default but it wasn't really the case.

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/.changesets/fix_bnjjj_subscription_default_configuration.md
+++ b/.changesets/fix_bnjjj_subscription_default_configuration.md
@@ -1,0 +1,5 @@
+### fix(subscription): force the deduplication to be enabled by default as it's documented ([PR #3773](https://github.com/apollographql/router/pull/3773))
+
+`subscription.enable_deduplication` was documented to be `true` by default but it wasn't really the case.
+
+By [@bnjjj](https://github.com/bnjjj) in https://github.com/apollographql/router/pull/3773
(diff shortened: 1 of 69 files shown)


Geal (approved):

.changesets/fix_bnjjj_subscription_default_configuration.md
```diff
@@ -0,0 +1,5 @@
+### fix(subscription): force the deduplication to be enabled by default as it's documented ([PR #3773](https://github.com/apollographql/router/pull/3773))
+
+`subscription.enable_deduplication` was documented to be `true` by default but it wasn't really the case.
```
o0Ignition0o, line 3:
```suggestion
A bug was introduced in router v1.25.0 which caused `enable_deduplication` to be disabled by default.
As documented, the router will enable deduplication by default, providing you with subscriptions that scale.

Should you decide to disable it, you can still explicitly set `enable_deduplication` to `false`.
```

apollo-router/src/plugins/subscription.rs
```diff
@@ -1268,6 +1269,22 @@ mod tests {
...
+                    "public_url": "http://localhost:4000",
+                    "path": "/subscription/callback",
+                    "subgraphs": ["test"]
+                }
+            }
+        }))
+        .unwrap();
+
+        assert!(sub_config.enabled);
+        assert!(sub_config.enable_deduplication);
```
o0Ignition0o, line 1285: awesome!

Issue #3226, opened 2023-06-06
Improve / formalize configuration of telemetry / logging / observability / metrics / tracing


BrynCooke:
Currently we have an `experimental_logging` section that we should bring out of experimental and in addition support json formatting options. There have also been several other users asking for much more control over what is output in both spans and logs, so it's maybe time to take a more holistic look.

The things that users have asked us for are:

* Secrets redaction
* The ability to exclude spans from logging output
* The ability to disable attributes on spans
* The ability to only have attributes on spans if headers are present on the root request.
* Optional log events for when there are http or graphql errors
* Optionally set the supergraph span to error if there were graphql errors in the response
* `supergraph` span is wrapped in an `operation` span that follows [semantic conventions](https://opentelemetry.io/docs/specs/otel/trace/semantic_conventions/instrumentation/graphql/). Existing attributes on supergraph that are from semantic conventions are moved onto the new span.
~~* Log levels are made easier to use. (see below yaml)~~

## Config

Suggest the following format:
```yaml
telemetry:

  tracing:
    common: # Renamed from trace_config
      max_attributes_per_event: 128
      max_attributes_per_span: 128
      max_attributes_per_link: 128
      max_events_per_span: 128
      max_links_per_span: 128
      parent_based_sampler: true
      sampler: always_on
      service_name: router
      service_namespace: "default"
      resource:
        d: e

      # Resources are otel config not represented in the yaml config


    propagation:
      baggage: false
      jaeger: false
      datadog: false
      request:
        header_name: "X-REQUEST-ID"
      trace_context: false
      zipkin: false

    otlp:
      enabled: true
      endpoint: "http://localhost:4317/v1/traces"




  metrics:
    common:
      service_namespace: "default"
      service_name: router
      buckets:
        - 0.1

      resource:
        test: foo
    prometheus:
      enabled: true
      path: /metrics
    otlp:
      enabled: true


  instruments:
    default_attribute_requirement_level: required
    router:
      http.server.active_requests: true
      my_instrument:
        value: unit
        type: counter
        unit: kb
        description: "my description"
        event: on_error
        attributes:

          http.response.status_code: false
          "my_attribute":
            response_header: "X-MY-HEADER"
            default: "unknown"
            redact: "foo"

    supergraph:
      my_instrument:
        value: unit
        event: on_error
        type: counter
        unit: kb
        description: "my description"
    subgraph:
      my_instrument:
        value: unit
        event: on_error
        type: counter
        unit: kb
[...]

PR #3787, opened 2023-09-07
fix the session count metric

### LINKED ISSUE #3485: apollo_router_session_count metric is suspect
As described in: https://github.com/apollographql/router/issues/3392#issuecomment-1645524436 the metric seems to provide odd values at times. A negative session count is often reported and clearly that's not right.

Geal:
Fix #3485 

using a counter for the session count meant that sometimes, the counter is reset to 0 on the metric collector side while the router has active sessions, which means that the next time a session is closed, the counter will send -1 and the metric will be negative. By storing the session count in an atomic, we can make sure to always keep the right( value in the router and have it survive across configuration reloads

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


BrynCooke:
Sorry, let's put this off until after the otel update merges. There's going to be merge conflicts if we do this now. 

I agree that the PR will fix things.

bnjjj (approved):

Geal:
@BrynCooke no rush on this, just a status update: shall I update this to current dev and make it mergeable now, or wait for the work on #3226 to end?

bnjjj:
feel free to merge @Geal it won't affect our telemetry PR

garypen (commented):
Very nice!

I think the docs will need to be updated to document the change to gauge.

Geal:
there's an issue with the active session count that I cannot solve. In every test it fails with `panicked at 'attempt to subtract with overflow', apollo-router/src/axum_factory/axum_http_server_factory.rs:593:33`, while there's clearly been a write to the atomic before, with the same kind of ordering

garypen:
> there's an issue with the active session count that I cannot solve. In every test it fails with `panicked at 'attempt to subtract with overflow', apollo-router/src/axum_factory/axum_http_server_factory.rs:593:33`, while there's clearly been a write to the atomic before, with the same kind of ordering

https://stackoverflow.com/questions/50631189/why-are-const-atomic-variables-not-updated-but-static-atomic-variables-are

If you make the Atomic `static` the tests start passing.

Geal:
yup, that worked!

Geal:
the last remaining issue here is that `apollo_router_session_count_total` is recorded, while `apollo_router_session_active` is not, and I cannot figure out whyn they're both gauges of the same type.

They both go through

https://github.com/apollographql/router/blob/dev/apollo-router/src/metrics/layer.rs#L248

and

https://github.com/apollographql/router/blob/dev/apollo-router/src/metrics/layer.rs#L152-L159

but only the total metric ends up in prometheus

BrynCooke:
This PR has been around for a while now. I'm OK with merging it if we think it's working.

Longer term we'll be deprecating this metric in favor of https://opentelemetry.io/docs/specs/semconv/http/http-metrics/#metric-httpserveractive_requests. 

At this point we'll be allowing configuration of metrics in the telemetry plugin, and we'll need to do the work to move it before any other layers.

Geal:
@BrynCooke there's still the issue from my previous comment, where `apollo_router_session_active` does not appear in the tests, while the other metric does. I still don't understand why, they're written in the same way

Geal:
I tried to fix the test again but couldn't. I validated manually again that the `apollo_router_session_active` metric is generated and available through prometheus, it just does not appear in the metrics reloading test. Since it will soon be deprecated, I will just remove the failing test([c973be2](https://github.com/apollographql/router/pull/3787/commits/c973be22c71127bd6616f049862abfc888d42139)) and move on, there's no need to waste more time on this

PR #3790, opened 2023-09-08
docs: Update federation-version-support.mdx

abernix:
Update the [Federation version matrix](https://www.apollographql.com/docs/router/federation-version-support/) in our docs to have ⚠️ indicators as per our convention.

### DIFF
--- a/docs/source/federation-version-support.mdx
+++ b/docs/source/federation-version-support.mdx
@@ -27,7 +27,7 @@ The table below shows which version of federation each router release is compile
     <tbody>
     <tr>
         <td>
-            v1.29.0 and later (<a href="https://github.com/apollographql/router/releases">see latest releases</a>)
+            v1.29.1 and later (<a href="https://github.com/apollographql/router/releases">see latest releases</a>)
         </td>
         <td>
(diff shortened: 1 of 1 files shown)


docs/source/federation-version-support.mdx
```diff
@@ -27,15 +27,23 @@ The table below shows which version of federation each router release is compile
...
             2.5.3
         </td>
     </tr>
     <tr>
         <td>
-            v1.28.0 - v1.28.1
+            ⚠️ v1.29.0
+        </td>
+        <td>
+            2.5.3
```
BrynCooke, line 41: 2.5.3 is listed in the row above is it intentional to have two rows with the same value?.

abernix: Yes, the version with the ⚠️ is specific.

abernix: v1.29.1 with 2.5.3 is great.  1.29.0 with 2.5.3 is not.

BrynCooke (approved):

shorgi (approved):

PR #3794, opened 2023-09-08
TLS client authentication

### LINKED ISSUE #3414: Support for Mutual TLS (mTLS) Between Router and Subgraphs

For many organizations, security is a paramount concern. As they scale, they need to ensure that traffic coming from one service is actually coming from that service. [Mutual TLS (mTLS)](https://www.cloudflare.com/learning/access-management/what-is-mutual-tls/) is a common tactic to enable this, as it ensures both the client and server must exchange certificates to validate they are who they say they are. 

The router should support mTLS between itself and the subgraphs via configuration. 

There are three alternatives currently available, neither fully address the concerns: 
- Use of normal TLS, which while a step in the right direction, also means that it's only validation that the server knows the client's certificate. The traffic to the client can still be sent from anywhere, so it's not proof of server/source. 
- Using a coprocessor to sign subgraph requests with bespoke tokens. This is another layer to the above, and can help significantly, but requires that the server validate the tokens as well. 
- Configuring network isolation. This helps ensure traffic is coming from dedicated sources, but for many organizations, they'll have configured bastion hosts that will make this issue moot. 

While mTLS doesn't address all of these concerns, it does help with points of all and many of them would be layered on top.

Geal:
Fix #3414

The router now supports TLS client authentication when connecting to subgraphs. It can be configured as follows:

```yaml
tls:
  subgraph:
    all:
      client_authentication:
        certificate_chain: ${file./path/to/certificate_chain.pem}
        private_key: ${file./path/to/key.pem}
    # if configuring for a specific subgraph:
    subgraphs:
      # subgraph name
      products:
        client_authentication:
          certificate_chain: ${file./path/to/certificate_chain.pem}
          private_key: ${file./path/to/key.pem}
```

This also adds tests for subgraph TLS connections with self signed server certificates, with root certificate authority override, and with client authentication.

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


bnjjj (approved):
I'm wondering if we should write the `mtls` keyword somewhere in our documentation if our users are looking for mtls in our documentation.

BrynCooke (approved):
Approved, but please create a followup ticket to create metrics for the `tls` feature as a whole. Currently we don't even have config metrics so we have no idea on the usage.

PR #3801, opened 2023-09-11
Activate query planner cache warm up by default

### LINKED ISSUE #3766: Activate query planner warm up by default
Since it was experimental, it was deactivated by default, but we feel now that it should be active by default.

What's the proper default value? In general, queries follow a power law: the few most frequent queries take up the large majority of the traffic. So we do not need to cover the entire cache to get the most benefit out of it.
Warming up queries for the entire cache would be fine though, it's just that it makes the schema update longer.
I propose that we use as default value a ratio, like 30% of the queries in cache get warmed up, that should be enough

Geal:
Fix #3766

Cache warm up has been in testing for long enough, and has proven highly ezffective in smoothing out latencies when updating a schema. So this will activate it by default. It can still be deactivated by configuring it to 0:

```yaml
supergraph:
  query_planning:
    warmed_up_queries: 0
```

By default, it will now warm up the cache with the 30% most used queries

*Description here*

Fixes #issue_number

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/configuration/mod.rs
+++ b/apollo-router/src/configuration/mod.rs
@@ -827,11 +827,12 @@ impl Default for Apq {
 pub(crate) struct QueryPlanning {
     /// Cache configuration
     pub(crate) experimental_cache: Cache,
-    /// Warm up the cache on reloads by running the query plan over
-    /// a list of the most used queries
-    /// Defaults to 0 (do not warm up the cache)
+    /// Warms up the cache on reloads by running the query plan over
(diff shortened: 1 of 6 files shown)


bnjjj (approved):

bnjjj:
Missing a changelog

aaronArinder (approved):

PR #3806, opened 2023-09-12
Coprocessors: Allow JSON and String body shapes for ControlFlow::Break

o0Ignition0o:
This PR fixes a regression that would prevent coprocessor writers to ControlFlow::Break with a body `String`.


apollo-router/src/plugins/coprocessor.rs
```diff
@@ -47,6 +47,9 @@ use crate::tracer::TraceId;
 
 pub(crate) const EXTERNAL_SPAN_NAME: &str = "external_plugin";
 const POOL_IDLE_TIMEOUT_DURATION: Option<Duration> = Some(Duration::from_secs(5));
+// TODO: are there better extension codes we could use here?
```
garypen, line 50: I think it's important to clearly identify errors. I agree that it might not be ideal to leak this to a client, but I don't think it's a big deal. Knowing that a problem occurred in an EXTERNAL component quickly narrows down problem resolution.

garypen (approved):
I think this looks fine. I'm approving, but remember to remove the TODO at line 50 if you merge.

apollo-router/src/plugins/coprocessor.rs
o0Ignition0o: works for me! thanks!

Geal (approved):

bnjjj (approved):

PR #3815, opened 2023-09-13
document the query plan cache warm up

### LINKED ISSUE #3704: Query planning cache warm up improvements
the [warm_up_queries option](https://www.apollographql.com/docs/router/configuration/in-memory-caching#caching-query-plans) takes the most used queries from the in memory cache, and creates query plans for them when a new schema comes in, before switching the router to this new schema. This makes schema updates much smoother, because it won't delay most of the queries coming, since they already have a query plan available.

It could benefit from a few improvements though:

* [x] #3767 
* [x] #3766 
* [x] #3621
* [x] #3769
* [x] #3770
* [x] #3815

Follow up:
* [ ] #3768
* [ ] #3771

Geal:
Fix #3704

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/docs/source/configuration/in-memory-caching.mdx
+++ b/docs/source/configuration/in-memory-caching.mdx
@@ -42,6 +42,45 @@ supergraph:
         limit: 512
 ```
 
+### Cache warm-up
+
+When loading a new schema, a query plan might change for some queries, so cached query plans cannot be reused. 
+
+To prevent increased latency upon query plan cache invalidation, the Router precomputes query plans for:
+* The most used queries from the cache.
+* The entire list of persisted queries.
+
+Precomputed plans will be cached before the Router switches traffic over to the new schema.
+
+By default, the Router warms up the cache with 30% of the queries already in cache, but it can be configured as follows:
+
+```yaml title="router.yaml"
+supergraph:
+  query_planning:
+    # Pre-plan the 100 most used operations when the supergraph changes
+    warmed_up_queries: 100
+```
+
+To get more information on the planning and warm-up process use the following metrics:
+
+(`<storage>` can be `redis` for distributed cache or `memory`)
+* counters:
+  * `apollo_router_cache_size{kind="query planner", storage="<storage>}`: current size of the cache (only for in-memory cache)
+  * `apollo_router_cache_hit_count{kind="query planner", storage="<storage>}`
+  * `apollo_router_cache_miss_count{kind="query planner", storage="<storage>}`
+
+* histograms:
+  * `apollo_router_query_planning_time`: time spent planning queries
+  * `apollo_router_schema_loading_time`: time spent loading a schema
+  * `apollo_router_cache_hit_time{kind="query planner", storage="<storage>}`: time to get a value from the cache
+  * `apollo_router_cache_miss_time{kind="query planner", storage="<storage>}`
+
+Typically, we would look at `apollo_router_cache_size` and the cache hit rate to define the right size of the in memory cache,
+then look at `apollo_router_schema_loading_time` and `apollo_router_query_planning_time` to decide how much time we want to spend warming up queries.
+
+#### Cache warm-up with distributed caching
+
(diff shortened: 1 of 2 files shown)


docs/source/configuration/in-memory-caching.mdx
```diff
@@ -42,6 +42,38 @@ supergraph:
...
+By default, the router warms up the cache with 30% of the queries already in cache, but it can be configured as follows:
+
+```yaml title="router.yaml"
+supergraph:
+  query_planning:
+    # Pre-plan the 100 most used operations when the supergraph changes
+    warmed_up_queries: 100
+```
+
+To configure it finely, we can get more information on the planning and warm up process by looking at the following metrics:
```
BrynCooke, line 57:
```suggestion
we can get more information on the planning and warm up process by looking at the following metrics:
```

docs/source/configuration/in-memory-caching.mdx
```diff
@@ -42,6 +42,38 @@ supergraph:
         limit: 512
 ```
 
+### Cache warm up
+
+When loading a new schema, the query plan might change for some queries, so we cannot reuse that cache. If we invalidated the entire cache, this would incur latency on queries after the reload, because we would have to wait for them to be planned. To avoid that, the router takes the most used queries from the cache, and the entire list of persisted queries, generates query plans for them in advance and caches them before switching traffic over to the new schema.
+By default, the router warms up the cache with 30% of the queries already in cache, but it can be configured as follows:
```
BrynCooke, line 48:
```suggestion
When loading a new schema, a query plan might change for some queries, so cached query plans cannot be reused. 

To prevent increased latency upon query plan cache invalidation, the Router warms will pre-compute query plans for:
* the most used queries from the cache
* the entire list of persisted queries.

Pre-computed plans will be cached before the Router switches traffic over to the new schema.

By default, the Router warms up the cache with 30% of the queries already in cache, but it can be configured as follows:
```

docs/source/configuration/in-memory-caching.mdx
```diff
@@ -42,6 +42,38 @@ supergraph:
...
+  * `apollo_router_schema_loading_time`: time spent loading a schema
+  * `apollo_router_cache_hit_time{kind="query planner", storage="<storage>}`: time to get a value from the cache
+  * `apollo_router_cache_miss_time{kind="query planner", storage="<storage>}`
+
+Typically, we would look at `apollo_router_cache_size` and the cache hit rate to define the right size of the in memory cache,
+then look at `apollo_router_schema_loading_time` and `apollo_router_query_planning_time` to decide how much time we want to spend warming up queries.
+
+#### Cache warm up with distributed caching
+
+If the router is using distributed caching for query plans, the warm up phase will also store the new query plans in Redis. Since all router instances might have the same distributions of queries in their in memory cache, the list of queries is shuffled before warm up, so each router instance can plan queries in a different order and share their results through the cache.
```
BrynCooke, line 76:
```suggestion
If the Router is using distributed caching for query plans, the warm up phase will also store the new query plans in Redis. Since all Router instances might have the same distributions of queries in their in memory cache, the list of queries is shuffled before warm up, so each Router instance can plan queries in a different order and share their results through the cache.
```

docs/source/configuration/in-memory-caching.mdx
```diff
@@ -42,6 +42,45 @@ supergraph:
...
+  query_planning:
+    # Pre-plan the 100 most used operations when the supergraph changes
+    warmed_up_queries: 100
+```
+
+To get more information on the planning and warm up process use the following metrics:
+
+(`<storage>` can be `redis` for distributed cache or `memory`)
+* counters:
+  * `apollo_router_cache_size{kind="query planner", storage="<storage>}`: current size of the cache (only for in memory cache)
```
shorgi, line 68:
```suggestion
  * `apollo_router_cache_size{kind="query planner", storage="<storage>}`: current size of the cache (only for in-memory cache)
```

shorgi (commented):
Copyedit comments

o0Ignition0o (approved):

PR #3818, opened 2023-09-13
chore(xtask): fix changeset create with the new pr template

bnjjj:
Fix a regression introduced in #3814


abernix (approved):
heh.  I assumed that `<!-- start metadata -->\n` alone would have been the delineation.  I am now not sure why this includes checklist.  Anyhow, merge away.  Probably won't change this template again for a while.

abernix:
Ah, I know why it's like that right now — it was for backwards compatibility at the time! The design was definitely to make it more durable in the future by only using the metadata tag, as the comment suggests.

https://github.com/apollographql/router/blob/11bfde1e03c2b31869a9844ebe0662664fb81607/xtask/src/commands/changeset/mod.rs#L396-L398

Still, I'd just merge for now.

PR #3819, opened 2023-09-13
Introduce OneShotAsyncCheckpoint

garypen:
The existing AsynCheckpoint requires `Clone` and thus introduces a need for Service Buffering that reduces the performance and resiliance of the router.

This new service set of Services, Layers and utility functions removes the requirement for `Clone` and thus the requirement for service buffering.

Existing uses of AsyncCheckpoint within the router are replaced with OneShotAsyncCheckpoint along with the requirement to `buffer()` such services.

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/layers/async_checkpoint.rs
+++ b/apollo-router/src/layers/async_checkpoint.rs
@@ -67,6 +67,82 @@ where
     }
 }
 
+/// [`Service`] for OneShot (single use) Asynchronous Checkpoints. See [`ServiceBuilderExt::oneshot_checkpoint_async()`](crate::layers::ServiceBuilderExt::oneshot_checkpoint_async()).
+#[allow(clippy::type_complexity)]
+pub struct OneShotAsyncCheckpointService<S, Fut, Request>
+where
+    Request: Send + 'static,
+    S: Service<Request, Error = BoxError> + Send + 'static,
+    <S as Service<Request>>::Response: Send + 'static,
+    <S as Service<Request>>::Future: Send + 'static,
+    Fut: Future<Output = Result<ControlFlow<<S as Service<Request>>::Response, Request>, BoxError>>,
+{
+    inner: Option<S>,
+    checkpoint_fn: Arc<Pin<Box<dyn Fn(Request) -> Fut + Send + Sync + 'static>>>,
+}
+
+impl<S, Fut, Request> OneShotAsyncCheckpointService<S, Fut, Request>
+where
+    Request: Send + 'static,
+    S: Service<Request, Error = BoxError> + Send + 'static,
+    <S as Service<Request>>::Response: Send + 'static,
+    <S as Service<Request>>::Future: Send + 'static,
+    Fut: Future<Output = Result<ControlFlow<<S as Service<Request>>::Response, Request>, BoxError>>,
+{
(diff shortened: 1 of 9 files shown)


BrynCooke (approved):
Can we add a test for when next_response is called twice?

apollo-router/src/layers/async_checkpoint.rs
```diff
@@ -67,6 +67,82 @@ where
...
+
+    type Future = BoxFuture<'static, Result<Self::Response, Self::Error>>;
+
+    fn poll_ready(
+        &mut self,
+        cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<Result<(), Self::Error>> {
+        self.inner
+            .as_mut()
+            .expect("One shot must only be called once")
```
Geal, line 126: could this return an error instead of panicking? Panicking means that the entire router is brought down, while an error would only affect the current request. We can still have a `debug_assert!(self.inner.is_some())` though

garypen: Good suggestion. Bryn also made the same point. I'm going to re-work this.

garypen:
> Can we add a test for when next_response is called twice?

The thing to test isn't calling `next_response()` twice, but trying to use the service stack multiple times. I've added a new test that does this with a `buffered()` service stack and which reliably panics (as it should) rather than hanging.

apollo-router/src/layers/async_checkpoint.rs
garypen: I've re-worked the poll_ready() to improve the resilience: b72a52862d9d80ca82258612ade4f3dad27f8d8e

Geal (approved):

bnjjj (approved):

apollo-router/src/layers/async_checkpoint.rs
```diff
@@ -67,6 +68,83 @@ where
...
+            None => Poll::Ready(Err("One shot must only be called once".into())),
+        }
+    }
+
+    fn call(&mut self, req: Request) -> Self::Future {
+        let checkpoint_fn = Arc::clone(&self.checkpoint_fn);
+        let inner = self
+            .inner
+            .take()
+            .expect("One shot must only be called once");
```
o0Ignition0o, line 137: same as with poll_ready, could we return a boxerror instead ?

o0Ignition0o: https://github.com/apollographql/router/pull/3878

garypen: We could. This merged yesterday, but I could do a follow-up to change:
```
     fn call(&mut self, req: Request) -> Self::Future {
         let checkpoint_fn = Arc::clone(&self.checkpoint_fn);
-        let inner = self
-            .inner
-            .take()
-            .expect("One shot must only be called once");
+        let inner = match self.inner.take() {
+            Some(inner) => inner,
+            None => {
+                return Box::pin(async move { Err("No inner service found".into()) });
+            }
+        };
         Box::pin(async move {
             match (checkpoint_fn)(req).await {
                 Ok(ControlFlow::Break(response)) => Ok(response),
```

PR #3821, opened 2023-09-13
special case for __typename in authorization

Geal:
When evaluating authorization directives on fields returning interfaces,
we require the usage of fragments with type conditions if the interface
implementors have different security requirements. For the __typename
field though, we must make an exception, because it should be available
for all implementors

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


.changesets/fix_geal_authorization_typename.md
```diff
@@ -0,0 +1,12 @@
+### special case for __typename in authorization ([PR #3821](https://github.com/apollographql/router/pull/3821))
+
+When evaluating authorization directives on fields returning interfaces,
+we require the usage of fragments with type conditions if the interface
+implementors have different security requirements. For the __typename
+field though, we must make an exception, because it should be available
+for all implementors
+
+<!-- start metadata -->
+---
```
bnjjj, line 10:
```suggestion
```

this bug should be fixed now

bnjjj (approved):

.changesets/fix_geal_authorization_typename.md
```diff
@@ -0,0 +1,10 @@
+### special case for __typename in authorization ([PR #3821](https://github.com/apollographql/router/pull/3821))
+
+When evaluating authorization directives on fields returning interfaces,
```
BrynCooke, line 3: I don't fully understand this fix. Is there a ticket that goes into more detail?

Could `__typename` represent a leakage of data?

BrynCooke: If a concrete type is not accessible then a caller can still check to see if something exists by asking for the typename.

In the examples below a caller can see that the blog concrete type is a `PrivateBlog` even if they don't have access.

Geal: Yes, it is expected. The authorization feature assumes that clients know about the schema, and the schema indicates the condition to access types or fields. If the data is not there and the subgraph did not return a "not found" explicitely, it's easy to infer that the type was `PrivateBlog`. If we actually want to return "not found" on unauthorized data, it is better handled at the subgraph level

BrynCooke (approved):

PR #3837, opened 2023-09-15
experimental support for query batching

### LINKED ISSUE #126: query batching
**Describe the bug**
In our project we use BatchHttpLink. When attempting to run against `router`, an error is thrown

**To Reproduce**
1. Clone https://github.com/sjungling/apollo-router-batch
2. Checkout the `broken` branch
3. Run `yarn install`
4. Run `yarn start:router`
5. Run `yarn start`
6. Navigate to http://localhost:3000 and observe an error in the terminal for `router` (see **Outputs** below)
7. (Optional) Checkout `main` branch and refresh the browser where you can observe the query succeeded.

**Expected behavior**
I would assume a successful query response.

**Output**
```shell
Nov 10 21:05:43.343 DEBUG hyper::proto::h1::conn: incoming body is content-length (104 bytes)
Nov 10 21:05:43.343 DEBUG hyper::proto::h1::conn: incoming body completed
Nov 10 21:05:43.343 DEBUG warp::filters::body: request json body error: invalid type: map, expected a string at line 1 column 1
Nov 10 21:05:43.343 DEBUG warp::filter::service: rejected: Rejection([BodyDeserializeError { cause: Error("invalid type: map, expected a string", line: 1, column: 1) }, MethodNotAllowed])
```

 - OS: MacOS
 - Version 12.0.1
 - Apollo Client 3.4.17
 - Router v0.1.0-alpha.0

**Additional context**
[Delta between two branches in sample repo](https://github.com/sjungling/apollo-router-batch/compare/broken?w=1)

garypen:
Things needing attention

 - [x] Configuration to match the design
 - [x] Decide if cloning `parts` is acceptable (in review) It isn't, so not cloned but re-built.
 - [x] Decide if cloning context for each `SupergraphRequest` is acceptable (in review)
 - [x] Should we do special handling if `@defer` or `subscription` detected in batch?
 - [x] Metrics
 - [x] Find someone happy to test from an appropriately configured Apollo Client and verify it works
 - [x] Modify Apollo telemetry to create separate root traces for each batch entry
 
Fixes #126

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [x] Integration Tests
    - [x] Manual Tests

**Exceptions**

Manual testing was performed with `curl` and [apollo client](https://www.apollographql.com/docs/react/api/link/apollo-link-batch-http/)

**Notes**


apollo-router/src/services/router_service.rs
```diff
@@ -228,119 +229,125 @@ impl RouterService {
...
-        let mut request_res = self
-            .persisted_query_layer
-            .supergraph_request(supergraph_request);
+        let mut results = vec![];
 
-        if let Ok(supergraph_request) = request_res {
-            request_res = self.apq_layer.supergraph_request(supergraph_request).await;
-        }
+        for supergraph_request in supergraph_requests {
+            let mut request_res = self
```
Geal, line 235: split the loop body in a separate function maybe?

garypen: Yeah, the formatter has given up on this completely. I'll do some tidying up at some point. Just making the basic code work at the moment.

apollo-router/src/configuration/snapshots/apollo_router__configuration__tests__schema_generation.snap
```diff
@@ -1009,6 +1009,27 @@ expression: "&schema"
...
+      "type": "object",
+      "properties": {
+        "enabled": {
+          "description": "Activates Batching (disabled by default)",
+          "default": false,
+          "type": "boolean"
+        },
+        "mode": {
+          "description": "Batching mode (batch_http_link is supported)",
+          "default": "",
```
smyrick, line 1027: Should we default to `batch_http_link` mode too?

garypen: The [specification document](https://docs.google.com/document/d/1CJH17-X9f8htfOOhvGP8t8RwpVu9HvBIFOhagnOmS8Y/edit) specified default to empty.

BrynCooke:
There are a couple of things that need to be verified:
1. What to traces look like in studio?
2. Are we OK with the metrics being generated and sent to Apollo e.g. op counts?

apollo-router/src/configuration/snapshots/apollo_router__configuration__tests__schema_generation.snap
BrynCooke: I think the specification doc is wrong. 
The idea is that if we improve batching then users won't suddenly find that their queries are working in a different way.

garypen: So default to `batch_http_link` mode?

docs/source/executing-operations/query-batching.mdx
```diff
@@ -0,0 +1,170 @@
...
+
+counter for the number of batched requests received
+
+</td>
+</tr>
+
+<tr>
+<td style="min-width: 150px;">
+
+##### `apollo.router.config.batching`
```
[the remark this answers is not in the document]
garypen: done: 3ba599f6c5ed0d9d62b01d45cb20e1695f239fc9

apollo-router/src/configuration/snapshots/apollo_router__configuration__tests__schema_generation.snap
smyrick: I don't have access to that doc, but just from an user perspective there definitely should be an opt-in model with `enable: true` but if there is only one mode we support why not default to that for now?

garypen: I don't have any feelings about default to None or default to batch_http_link.

@BrynCooke was the original author and (by trawling the spec doc's history) he originally did have it set to default to `batch_http_link` mode. Then, on September 7th, he removed the default and the docs team set it to be `empty`.

Since @BrynCooke was the original author and he now thinks `empty` is wrong, I'm happy to update the default to `batch_http_link`.

BrynCooke: I've updated the doc. The reason that we want to make mode explicit is that we may want to change the default mode in future in a breaking way.

.changesets/exp_garypen_126_query_batching.md
```diff
@@ -0,0 +1,19 @@
...
+
+The Apollo Router must be configured to receive batch requests, otherwise it rejects them. When processing a batch request, the router deserializes and processes each operation of a batch independently, and it responds to the client only after all operations of the batch have been completed.
+
+```yaml
+experimental_batching:
+  enabled: true
+  mode: batch_http_link
+```
+
+Do not attempt to use subscriptions or `@delay` queries within a batch as they are not supported.
```
bnjjj, line 17: Is `@defer` supported ?

.changesets/exp_garypen_126_query_batching.md
```diff
@@ -0,0 +1,19 @@
...
+If you’re using Apollo Client, you can leverage the in-built support for batching to reduce the number of individual requests sent to the Apollo Router.
+
+Once [configured](https://www.apollographql.com/docs/react/api/link/apollo-link-batch-http/), Apollo Client will automatically combine multiple operations into a single HTTP request. The number of operations within a batch is client configurable, including the maximum number of operations in a batch and the maximum duration to wait for operations to accumulate before sending the batch request. 
+
+The Apollo Router must be configured to receive batch requests, otherwise it rejects them. When processing a batch request, the router deserializes and processes each operation of a batch independently, and it responds to the client only after all operations of the batch have been completed.
+
+```yaml
+experimental_batching:
+  enabled: true
+  mode: batch_http_link
```
bnjjj, line 14: Do we have another mode ?

apollo-router/src/configuration/mod.rs
```diff
@@ -1255,3 +1265,16 @@ fn default_graphql_path() -> String {
...
+#[derive(Debug, Default, Clone, Deserialize, Serialize, JsonSchema)]
+#[serde(deny_unknown_fields)]
+#[serde(default)]
+pub(crate) struct Batching {
+    /// Activates Batching (disabled by default)
+    pub(crate) enabled: bool,
+
+    /// Batching mode (batch_http_link is supported)
+    #[serde(default)]
+    pub(crate) mode: String,
```
bnjjj, line 1279: Why not using an enum here ?

apollo-router/src/configuration/mod.rs
```diff
@@ -1255,3 +1265,16 @@ fn default_graphql_path() -> String {
...
+/// Configuration for Batching
+#[derive(Debug, Default, Clone, Deserialize, Serialize, JsonSchema)]
+#[serde(deny_unknown_fields)]
+#[serde(default)]
+pub(crate) struct Batching {
+    /// Activates Batching (disabled by default)
+    pub(crate) enabled: bool,
+
+    /// Batching mode (batch_http_link is supported)
+    #[serde(default)]
```
bnjjj, line 1278: This is useless as the struct is already tagged with `serde(default)`

apollo-router/src/configuration/mod.rs
```diff
@@ -1255,3 +1265,16 @@ fn default_graphql_path() -> String {
 fn default_graphql_introspection() -> bool {
     false
 }
+
+/// Configuration for Batching
+#[derive(Debug, Default, Clone, Deserialize, Serialize, JsonSchema)]
+#[serde(deny_unknown_fields)]
+#[serde(default)]
```
bnjjj, line 1272:
```suggestion
#[serde(deny_unknown_fields, default)]
```

apollo-router/src/request.rs
```diff
@@ -165,27 +165,84 @@ impl Request {
...
+    /// cannot be turned into a valid GraphQL `Request`.
+    pub(crate) fn batch_from_bytes(bytes: &[u8]) -> Result<Vec<Request>, serde_json::Error> {
+        let value: serde_json::Value =
+            serde_json::from_slice(bytes).map_err(serde_json::Error::custom)?;
+
+        Request::process_batch_values(&value)
+    }
+
+    fn process_batch_values(value: &serde_json::Value) -> Result<Vec<Request>, serde_json::Error> {
+        let mut result = vec![];
```
bnjjj, line 189: I think there is a way to predict the capacity needed and initialize it using `with_capacity()`

apollo-router/src/request.rs
```diff
@@ -165,27 +165,84 @@ impl Request {
...
-        };
+            let bytes = serde_json::to_vec(value)?;
+            result.push(Request::deserialize_from_bytes(&bytes.into())?);
+        }
+        Ok(result)
+    }
 
-        let query = if let Some(serde_json::Value::String(query)) = urldecoded.get("query") {
+    fn process_query_values(value: &serde_json::Value) -> Result<Vec<Request>, serde_json::Error> {
+        let mut result = vec![];
```
bnjjj, line 211: Same for capacity

apollo-router/src/request.rs
```diff
@@ -165,27 +165,84 @@ impl Request {
...
+
+        if value.is_array() {
+            tracing::info!(
+                monotonic_counter.apollo_router.operations.batching = 1u64,
+                mode = "batch_http_link" // Only supported mode right now
+            );
```
bnjjj, line 217: Is it interesting to also have an histogram on how many queries we have in a batched query ?

apollo-router/src/services/router_service.rs
```diff
@@ -339,61 +324,208 @@ impl RouterService {
...
+                    .build();
+            }
+        };
+
+        let mut results = vec![];
+
+        for supergraph_request in supergraph_requests {
+            self.process_supergraph_request(&mut results, supergraph_request)
+                .await?;
+        }
```
bnjjj, line 384: Can we predict a minimal capacity using the `supergraph_requests.len()` ?

apollo-router/src/services/router_service.rs
```diff
@@ -339,61 +324,208 @@ impl RouterService {
...
+            for result in results {
+                array_result += ", ";
+                array_result += &String::from_utf8(
+                    hyper::body::to_bytes(result.response.into_body())
+                        .await?
+                        .to_vec(),
+                )
+                .unwrap();
+            }
+            array_result += "]";
```
bnjjj, line 405: I'm wondering if it would not be more optimized to only use bytes and then convert it to string if needed afterwards. For example we used [this](https://github.com/apollographql/router/blob/dev/apollo-router/src/protocols/multipart.rs#L149) and [this](https://github.com/apollographql/router/blob/dev/apollo-router/src/protocols/multipart.rs#L140) in the multipart protocol

apollo-router/src/services/router_service.rs
```diff
@@ -339,61 +324,208 @@ impl RouterService {
...
+
+            match graphql::Request::from_urlencoded_query(q.to_string()) {
+                Ok(request) => {
+                    result.push(request);
+                }
+                Err(err) => {
+                    // It may be a batch of requests, so try that (if config allows) before
+                    // erroring out
+                    if self.experimental_batching.enabled
+                        && self.experimental_batching.mode == "batch_http_link"
```
bnjjj, line 428: Would be less error prone if it was an enum

apollo-router/src/services/router_service.rs
```diff
@@ -339,61 +324,208 @@ impl RouterService {
...
+    ) -> Result<Vec<graphql::Request>, TranslateError> {
+        let mut result = vec![];
+
+        match graphql::Request::deserialize_from_bytes(bytes) {
+            Ok(request) => {
+                result.push(request);
+            }
+            Err(err) => {
+                if self.experimental_batching.enabled
+                    && self.experimental_batching.mode == "batch_http_link"
```
bnjjj, line 481: enum

apollo-router/src/services/router_service.rs
```diff
@@ -406,52 +538,74 @@ impl RouterService {
...
                     })
             }
         };
 
-        Ok(SupergraphRequest {
-            supergraph_request: http::Request::from_parts(parts, graphql_request?),
-            context,
-        })
+        let mut results = vec![];
+        let mut ok_results = graphql_requests?;
```
bnjjj, line 576:
```suggestion
        let mut ok_results = graphql_requests?;
        let mut results = Vec::with_capacity(ok_results.len() + 1);
        
```

docs/source/executing-operations/query-batching.mdx
```diff
@@ -0,0 +1,191 @@
...
+  },
+  {"data":{"me":{"name":"Ada Lovelace"}}}
+]
+```
+
+## Known limitations
+
+### Unsupported query modes
+ 
+When batching is enabled, any mode of querying that results in a stream of responses is unsupported, including:
```
bnjjj, line 189: You mentioned `@delay` in the changelog

bnjjj (commented):
Should we also have a configuration to limit the maximum number of queries accepted in a batched query ?

apollo-router/src/configuration/mod.rs
garypen: We don't know what future versions might be added, or removed, so I'm just using a String to keep it open.

.changesets/exp_garypen_126_query_batching.md
garypen: Not at this time.

garypen:
> There are a couple of things that need to be verified:
> 
> 1. What to traces look like in studio?
> 2. Are we OK with the metrics being generated and sent to Apollo e.g. op counts?

1. Hard to say right now. The traces are created and sent from the router and appear to be correct, but studio wasn't displaying them. Also: when we ran from dev with non-batch traces, studio didn't display those traces either. So, hard to verify right now until that problem is understood.
2. I believe we are.

apollo-router/src/configuration/mod.rs
garypen: Yup. I changed this to match the spec change, so neither the struct or the member are default now.

apollo-router/src/services/router_service.rs
bnjjj: If you keep it as a `String` could you at least put this in a constant and use it everywhere @garypen ?

apollo-router/src/services/router_service.rs
bnjjj: or const

docs/source/executing-operations/query-batching.mdx
garypen: I updated this to `@defer` in [33f5934](https://github.com/apollographql/router/pull/3837/commits/33f593473b6bd3989d3bdb5151c9a7f16a98c4e4)

apollo-router/src/configuration/mod.rs
garypen: This has changed with the change to the spec. There is no default on the struct now.

apollo-router/src/services/router_service.rs
garypen: done: a8359234e9910bb848692ab13d96a24f5ef07c0f

apollo-router/src/request.rs
garypen: done: a8359234e9910bb848692ab13d96a24f5ef07c0f

apollo-router/src/request.rs
garypen: done: a8359234e9910bb848692ab13d96a24f5ef07c0f

apollo-router/src/services/router_service.rs
garypen: done: 6088b38a8b78078c2a8dba66e1d11e64b57783fd

garypen:
> Should we also have a configuration to limit the maximum number of queries accepted in a batched query ?

There's nothing in the spec. We could perhaps add that in future if we decided it was a good idea. What do you think @BrynCooke ?

apollo-router/src/request.rs
garypen: That could be interesting. What do you think @BrynCooke ? Again, we could add that in the future if we liked.

BrynCooke: Yes, I think that's be very interesting. It'd help inform us on any future improved batching support.

apollo-router/src/services/router_service.rs
```diff
@@ -339,61 +326,219 @@ impl RouterService {
...
+                    .status_code(err.status)
+                    .header(CONTENT_TYPE, APPLICATION_JSON.essence_str())
+                    .context(context)
+                    .build();
+            }
+        };
+
+        let mut results = Vec::with_capacity(supergraph_requests.len());
+
+        for supergraph_request in supergraph_requests {
```
BrynCooke, line 383: Does this mean that each request is currently processed one at a time rather than concurrently?

garypen: Yes, to preserve ordering. Otherwise, I'd be concerned about mutation side-effects and the difficulties of re-assembling the batch response out of order.

apollo-router/src/services/router_service.rs
garypen: done as enum: 3cc67f8da4bc72db559801292a1573b338c02484

apollo-router/src/services/router_service.rs
garypen: done as enum: 3cc67f8da4bc72db559801292a1573b338c02484

apollo-router/src/services/router_service.rs
BrynCooke: Did the original user submitted PR do parallel execution? (Also worth checking the gateway)

garypen: The original user submitted PR did concurrent execution. I don't think it should have because of the "mutation side-effects" point from above.

It might be possible to check if batch contains mutations and execute concurrently if it doesn't. That would be a viable followup issue for performance.

apollo-router/tests/snapshots/apollo_reports__condition_if.snap
garypen, line 1: I don't know why these changes have results in an update to this snapshot. My best guess is that the somewhat fragile nature of `apollo_reports` and the hackery that I did to get the new batch tests to pass is at least part of the reason.
I think it would be wisest to either:
 - review my apollo_reports changes with someone who has a better understanding of how those tests should work.
 - or remove the new batch tests until we can figure out a better way to perform them.

BrynCooke: I'd be happy to take a look with you tomorrow.

apollo-router/src/services/router_service.rs
```diff
@@ -594,13 +596,50 @@ impl RouterService {
...
+        // but the concern is that there may be other private_entries that wish to propagate into
+        // each request or we may add them in future and not know about it here...
+        //
+        // (Technically we could clone private entries, since it is held under an `Arc`, but that
+        // would mean all the requests in a batch shared the same set of private entries and review
+        // comments expressed the sentiment that this may be a bad thing...)
+        //
         for graphql_request in ok_results {
-            // XXX Lose extensions, is that ok?
+            // XXX Lose http extensions, is that ok?
```
bnjjj, line 624: I think so

apollo-router/tests/snapshots/apollo_reports__condition_if.snap
bnjjj: I think that's because we're relying on the root trace which is `router` span or something from the router_service and then we find it but there is several sub traces and it doesn't follow the logic we wrote in the apollo exporter

apollo-router/src/services/router_service.rs
garypen: resolution: batches should be executed concurrently.
done: [88c8326](https://github.com/apollographql/router/pull/3837/commits/88c83261b315afb6d309fdc087606c7ce292bf61)

docs/source/executing-operations/query-batching.mdx
```diff
@@ -0,0 +1,208 @@
...
+
+</td>
+<td>
+
+mode
+
+</td>
+<td>
+
+Counter for the number received batch requests.
```
Meschreiber, line 77:
```suggestion
Counter for the number of received batch requests.
```

docs/source/executing-operations/query-batching.mdx
```diff
@@ -0,0 +1,208 @@
...
+</td>
+</tr>
+</tbody>
+</table>
+
+
+## Query batch formats
+
+### Request format
+
```
Meschreiber, line 106:
```suggestion

A batched request is an array of operations.

```

docs/source/executing-operations/query-batching.mdx
```diff
@@ -0,0 +1,208 @@
...
+  me {
+    name
+  }
+}
+]
+```
+
+### Response format
+
+Responses are provided in JSON array, with ordering of responses matching the supplied request.
```
Meschreiber, line 124:
```suggestion
Responses are provided in JSON array, with the order of responses matching the order of operations in the batched request.
```

docs/source/executing-operations/query-batching.mdx
```diff
@@ -0,0 +1,208 @@
...
+  {"data":{"me":{"id":"1"}}},
+  {"data":{"me":{"name":"Ada Lovelace"}}}
+]
+```
+
+## Error handling for query batching
+
+### Batch error
+
+If a batch of queries cannot be processed, the entire batch will fail.
```
Meschreiber, line 137:
```suggestion
If a batch of queries cannot be processed, the entire batch fails.
```

docs/source/executing-operations/query-batching.mdx
```diff
@@ -0,0 +1,208 @@
...
+]
+```
+
+## Error handling for query batching
+
+### Batch error
+
+If a batch of queries cannot be processed, the entire batch will fail. 
+
+For example, this query is invalid because it has two commas to separate the batch of queries:
```
Meschreiber, line 139:
```suggestion
For example, this batch request is invalid because it has two commas to separate the constituent queries:
```

docs/source/executing-operations/query-batching.mdx
```diff
@@ -0,0 +1,208 @@
...
+},,
+query MySecondQuery {
+  me {
+    name
+  }
+}
+]
+```
+
+As a result, an error is returned for the invalid batch:
```
Meschreiber, line 156:
```suggestion
As a result, the router returns an invalid batch error:
```

docs/source/executing-operations/query-batching.mdx
```diff
@@ -0,0 +1,208 @@
...
+{"errors":
+  [
+    {"message":"Invalid GraphQL request","extensions":{"details":"failed to deserialize the request body into JSON: expected value at line 1 column 54","code":"INVALID_GRAPHQL_REQUEST"}}
+  ]
+}
+```
+
+### Individual query error
+
+If a single query in a batch cannot be processed, this will result in an individual error.
```
Meschreiber, line 168:
```suggestion
If a single query in a batch cannot be processed, this results in an individual error. 
```

docs/source/executing-operations/query-batching.mdx
```diff
@@ -0,0 +1,208 @@
...
+  },
+  {"data":{"me":{"name":"Ada Lovelace"}}}
+]
+```
+
+## Known limitations
+
+### Unsupported query modes
+ 
+When batching is enabled, any mode of querying that results in a stream of responses is unsupported, including:
```
Meschreiber, line 206: Does this mean that the router cannot process these types of operations _at all_ or just in batches? I assume the latter, but let's make it clear, either way.

docs/source/executing-operations/query-batching.mdx
```diff
@@ -0,0 +1,208 @@
...
+- `@defer`
```
Meschreiber, line 207:
```suggestion
- [`@defer`](/graphos/operations/defer/)
```

docs/source/executing-operations/query-batching.mdx
```diff
@@ -0,0 +1,208 @@
...
+- subscriptions
```
Meschreiber, line 208:
```suggestion
- [subscriptions](/graphos/operations/subscriptions/)
```

docs/source/executing-operations/query-batching.mdx
```diff
@@ -0,0 +1,208 @@
...
+},
+query MySecondQuery {
+  me {
+    name
+  }
+}
+]
+```
+
+As a result, an error is returned for the individual query:
```
Meschreiber, line 187:
```suggestion
As a result, an error is returned for the individual invalid query and the other (valid) query returns a response.
```

Meschreiber (commented):
Docs are super nicely structured! Besides inline nits/suggestions, I'm wondering about the naming/terminology. (I might be hyper-sensitive to this because of persisted queries 😅).

We seem to use "query," "request," and "operation," interchangeably throughout the doc, e.g. "query batching" and "request batching". We should try to unify our terminology within this doc and across Apollo. (E.g. see client docs that refer to this as “operation batching": https://www.apollographql.com/docs/kotlin/advanced/query-batching/, https://www.apollographql.com/docs/react/api/link/apollo-link-batch-http/ vs this [blog post](https://www.apollographql.com/blog/apollo-client/performance/batching-client-graphql-queries/))

I propose keeping "query batching" as the feature name and then referring to "operations" (not queries or requests) within a batch.

apollo-router/tests/snapshots/apollo_reports__condition_if.snap
garypen: This is resolved now. I've re-written the `apollo_router.rs` tests and this anomaly has gone away.

docs/source/executing-operations/query-batching.mdx
garypen: You assume correctly. How about this?:
```
When batching is enabled, any batch operation that results in a stream of responses is unsupported, including:
```
I've made that edit. If you don't like it, feel free to suggest improvements.

Meschreiber: That looks good, thank you! 🙇‍♀️

docs/source/executing-operations/query-batching.mdx
```diff
@@ -0,0 +1,208 @@
...
+
+To enable batching in a client, configure `BatchHttpLink`. For details on implementing `BatchHttpLink`, see [batching operations](https://www.apollographql.com/docs/react/api/link/apollo-link-batch-http/).
+
+### Configuration compatibility
+
+If the router receives a query batch from a client, and batching is *not* enabled, the router sends a `BATCHING_NOT_ENABLED` error to the client, which clearly identifies why the query batch failed.
+
+## Metrics for query batching
+
+Metrics in the Apollo Router for query batching:
```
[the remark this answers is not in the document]
garypen: It depends how your router is configured. For example, if you enable prometheus then you can see the metrics in prometheus. If you configure DataDog, then the metrics are exported to DataDog. You may also configure any other metrics ingrestion mechanism that understand OpenTelemetry.

I think that anyone who is interested in consuming metrics from the router will know this. We could add links to:
https://www.apollographql.com/docs/router/configuration/metrics/#using-prometheus but that's not telling the full story.

docs/source/executing-operations/query-batching.mdx
```diff
@@ -0,0 +1,208 @@
...
+| `enabled` | Flag to enable reception of client batches | boolean | `false` |
+| `mode` | Supported client batching mode | `batch_http_link`:  the client uses Apollo Link and its [`BatchHttpLink`](/react/api/link/apollo-link-batch-http) link. | No Default |
+
+### Configure client
+
+To enable batching in a client, configure `BatchHttpLink`. For details on implementing `BatchHttpLink`, see [batching operations](https://www.apollographql.com/docs/react/api/link/apollo-link-batch-http/).
+
+### Configuration compatibility
+
+If the router receives a query batch from a client, and batching is *not* enabled, the router sends a `BATCHING_NOT_ENABLED` error to the client, which clearly identifies why the query batch failed.
```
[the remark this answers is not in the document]
garypen: The error does contain additional information. It would look something like this:
```
{
  "errors": [
    {
      "message": "Invalid GraphQL request",
      "extensions": {
        "details": "batching not enabled",
        "code": "BATCHING_NOT_ENABLED"
      }
    }
  ]
}
```
However, I agree that the dependent clause is superfluous.

docs/source/executing-operations/query-batching.mdx
```diff
@@ -0,0 +1,208 @@
...
+```
+
+| Attribute | Description | Valid Values | Default Value |
+| :-- | :-- | :-- | :-- |
+| `enabled` | Flag to enable reception of client batches | boolean | `false` |
+| `mode` | Supported client batching mode | `batch_http_link`:  the client uses Apollo Link and its [`BatchHttpLink`](/react/api/link/apollo-link-batch-http) link. | No Default |
+
+### Configure client
+
+To enable batching in a client, configure `BatchHttpLink`. For details on implementing `BatchHttpLink`, see [batching operations](https://www.apollographql.com/docs/react/api/link/apollo-link-batch-http/).
```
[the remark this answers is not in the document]
garypen: Well, that's a specific Apollo React client feature. It's not a requirement to make use of batching, but we call it out here because we know that we have customers/users using this client and who want to make use of batching. Perhaps:

To enable batching in an Apollo React client, configure `BatchHttpLink`. For details on implementing `BatchHttpLink`, see [batching operations](/react/api/link/apollo-link-batch-http/).

?

docs/source/executing-operations/query-batching.mdx
```diff
@@ -0,0 +1,208 @@
...
+> **Experimental Feature**  
+> The Apollo Router provides *experimental* support for client query batching.
+
+Learn about query batching and how to configure the Apollo Router to receive query batches.
+
+## About query batching
+
+Modern applications use several operations to render a single page. This is usually the result of a component-based architecture where individual micro-frontends (MFE) make requests separately to fetch data relevant to them. Not only does this cause a performance overhead (i.e. different components may be requesting the same data), it can also cause a consistency issue. To combat this, MFE-based UIs batch multiple client queries and operations, issued close together, into one HTTP request. This is supported in Apollo Client and Apollo Server.
+
+The Apollo Router now supports client request batching. If you’re using Apollo Client, you can leverage the built-in support for batching to reduce the number of individual requests sent to the router.
```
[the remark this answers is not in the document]
garypen: No we don't. I've removed "now"

docs/source/executing-operations/query-batching.mdx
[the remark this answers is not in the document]
Meschreiber: Got it. We can leave as is. :)

BrynCooke (approved):

apollo-router/src/request.rs
```diff
@@ -165,27 +165,101 @@ impl Request {
...
+        }
+    }
+
+    fn process_batch_values(value: &serde_json::Value) -> Result<Vec<Request>, serde_json::Error> {
+        let mut result = Request::allocate_result_array(value);
+
+        if value.is_array() {
+            tracing::info!(
+                histogram.apollo_router.operations.batching.size = result.len() as f64,
+                mode = "batch_http_link" // Only supported mode right now
```
bnjjj, line 201: NIT: for these metrics you should implement Display on the enum Mode and just use `%BatchingMode::HttpLink` to avoid typo

apollo-router/src/services/supergraph_service.rs
```diff
@@ -229,6 +230,29 @@ async fn service_call(
             let is_deferred = plan.is_deferred(operation_name.as_deref(), &variables);
             let is_subscription = plan.is_subscription(operation_name.as_deref());
 
+            if let Some(batching) = context.private_entries.clone().lock().get::<Batching>() {
```
bnjjj, line 233: Is clone mandatory here ?

docs/source/executing-operations/query-batching.mdx
```diff
@@ -0,0 +1,210 @@
...
+
+### Configure client
+
+To enable batching in an Apollo client, configure `BatchHttpLink`. For details on implementing `BatchHttpLink`, see [batching operations](/react/api/link/apollo-link-batch-http/).
+
+### Configuration compatibility
+
+If the router receives a query batch from a client, and batching is *not* enabled, the router sends a `BATCHING_NOT_ENABLED` error to the client.
+
+## Metrics for query batching
```
bnjjj, line 50: I think we should at least mention them in the metrics docs where we list all metrics

apollo-router/src/services/supergraph_service.rs
garypen: Yes, or the borrow checker gets angry that I'm moving out of `context` at line 249. It's a lightweight `Arc` clone.

apollo-router/src/request.rs
garypen: Yep, that's a hangover from when I first implemented it as a `String`. `Display` is implemented already, but I overlooked updating it.

done: 4fd50dd142795bee0c2ef053bc54050fc9d19fee

apollo-router/src/services/supergraph_service.rs
bnjjj: Oh yes indeed

bnjjj (approved):

PR #3839, opened 2023-09-18
detect lists properly in error paths for authorization

Geal:
The lists were not correctly recognized when building the error path in the authorization plugin. This has no impact on actual authorization, types that require authorization are still properly recognized even in a list or non null

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/plugins/authorization/authenticated.rs
+++ b/apollo-router/src/plugins/authorization/authenticated.rs
@@ -995,6 +995,7 @@ mod tests {
        name: String
        phone: String @authenticated
        activeOrganization: Organization
+       addresses: [Address!]
    }
    type Organization
    @join__owner(graph: ORGA)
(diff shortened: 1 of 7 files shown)


Geal:
putting this one back in draft, because there's an issue I had not predicted: when we remove fields from types inside a list, we don't know how many elements will be in that list. IF we have the following schema:

```graphql
type Query {
  organizations: [Organization]
}

type Organization {
  name: String
  members: [User]
}

type User @authenticated {
  name: String
}
```

For the query `query { organizations { name members { name } } }`, if we are not authenticated, we will get the filtered query `query organizations { name }` and the error path `/organizations/@/members/@`, with `@` the "flatten" path element we use in query planning, that indicates it applies to all elements of the array.
So if we had the following data:
```json
{
  organizations: [{
    name: "first",
    members: [{
        name: "A"
      },
      {
        name: "B"
      }]
    },
    {
      name: "second",
      members: [{
        name: "C"
      }]
    }]
}
```

We need to generate the following paths:

* `/organizations/0/members/0`
* `/organizations/0/members/1`
* `/organizations/1/members/0`

That error dispatching should happen in the execution service, after response formatting

garypen (approved):

garypen:
I approved this, because the code changes looked good, and then changed my mind after I read [your comment](https://github.com/apollographql/router/pull/3839#issuecomment-1723092542) . Is it something we still need?

PR #3852, opened 2023-09-19
h2c support

### LINKED ISSUE #3535: Provide support for subgraph connections to only be made with h2c (http2 cleartext/prior knowledge)
In [subgraph_service.rs](https://github.com/apollographql/router/blob/19ea4d9b4cc675ba26326c9098ae068686edac21/apollo-router/src/services/subgraph_service.rs#L161-L189) provide explicit support allowing configuration of router to result in all outbound subgraph connections to be made using http2 cleartext (h2c).  Due to using h2c, there is no need to initialize the tls_config setup for these connections at all.

**Additional context**
Our environment's service to service communication uses envoy.  Each service's outbound connections connect via h2c to its own envoy sidecar.  That envoy sidecar provides load balancing and cross datacenter failover capabilities and establishes http2/tls connections to all destination service instances envoy sidecars which then communicate with the destination service's ingress h2c endpoint listener.  Thus for router to exist and communicate using the approved/golden path used by services, we need to be able to tell router to only use h2c for its outbound connections to subgraphs.

https://docs.rs/hyper/latest/hyper/client/struct.Builder.html#method.http2_only
we have used [tonic](https://github.com/hyperium/tonic) to make grpc calls over service mesh, and that requires h2c - and tonic is built on hyper 

In Apollo Gateway which we are working to migration off of, we have been using h2c outbound and accomplished this via the fetcher interface on RemoteGraphQLDataSource and used [@adobe/fetch](https://github.com/adobe/fetch).

Geal:
Fix #3535

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


bnjjj (approved):

docs/source/configuration/traffic-shaping.mdx
```diff
@@ -151,6 +152,10 @@ traffic_shaping:
     deduplicate_query: true # Enable query deduplication for all subgraphs.
 ```
 
+### HTTP/2
+
+The router supports subgraph connections over HTTP/2 with TLS. It is enabled by default, but can be disabled through the `experimental_http2` option. To use the HTTP/2 Cleartext protocol (h2c), which uses HTTP/2 over plaintext connections, the subgraph URL must have the `http` scheme and the `experimental_http2` option must be set to `http2only`.
+
```
shorgi, line 158:
```suggestion
### HTTP/2

The router supports subgraph connections over
- HTTP/2 with TLS. This is the default configuration.
- HTTP/2 Cleartext protocol (h2c). This uses HTTP/2 over plaintext connections.

To use h2c, the subgraph URL must have the `http` scheme, and the `experimental_http2` option must be set to `http2only`.

```

shorgi (commented):
Copyedit suggestions

apollo-router/src/services/subgraph_service.rs
```diff
@@ -2985,4 +2991,73 @@ mod tests {
...
+                    .into(),
+                )
+                .unwrap())
+        }
+
+        let make_svc = make_service_fn(|_conn| async { Ok::<_, Infallible>(service_fn(handle)) });
+        let server = Server::from_tcp(listener)
+            .unwrap()
+            .http2_only(true)
+            .serve(make_svc); //Server::from_tcp(listener).unwrap().serve(make_svc);
```
o0Ignition0o, line 3017:
```suggestion
            .serve(make_svc);
```

o0Ignition0o (approved):

PR #3859, opened 2023-09-20
preverify release version

### LINKED ISSUE #3664: Release process should check project files to see that versions match the git tag

BrynCooke:
Adds a new `xtask release pre-verify`

Adds a check that the cargo toml matches the local tagged tree.

Fixes #3664

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -869,6 +869,11 @@ workflows:
           matrix:
             parameters:
               platform: [ amd_linux_build ]
+          filters:
+            branches:
+              ignore: /.*/
+            tags:
+              only: /v.*/
       - build_release:
           matrix:
             parameters:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -380,18 +380,7 @@ commands:
                   - target
   xtask_release_preverify:
     steps:
-      - restore_cache:
-          keys:
-            - "<< pipeline.parameters.merge_version >>-preverify"
       - run: xtask release pre-verify
-      - when:
-          condition:
-            equal: [ "dev", "<< pipeline.git.branch >>" ]
-          steps:
-            - save_cache:
-                key: "<< pipeline.parameters.merge_version >>-preverify"
-                paths:
-                  - target
 
   xtask_check_helm:
     steps:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -378,6 +378,20 @@ commands:
...
                   - target
+  xtask_release_preverify:
+    steps:
+      - restore_cache:
+          keys:
+            - "<< pipeline.parameters.merge_version >>-preverify"
+      - run: xtask release pre-verify
+      - when:
+          condition:
+            equal: [ "dev", "<< pipeline.git.branch >>" ]
```
abernix, line 389: This isn't intuitive to me.  Is this the right conditional branch we want to check?  "When the current pipeline's branch is `dev` do this?"  How about releases that happen off a tag on a release branch?

xtask/src/commands/release.rs
```diff
@@ -410,3 +414,54 @@ impl Prepare {
...
+            .exec()?;
+        let version = format!(
+            "v{}",
+            metadata
+                .root_package()
+                .expect("expected root")
+                .version
+                .to_string()
+        );
+
```
abernix, line 434: Could this be replaced by using `PKG_VERSION`?  As we do in other places?

https://github.com/apollographql/router/blob/2f2462f8b06fda28d21369f7c8e1912efbdb8804/xtask/src/lib.rs#L25-L37

.circleci/config.yml
BrynCooke: Ahh, we can just remove the stuff around caching.

xtask/src/commands/release.rs
BrynCooke: [a85f195](https://github.com/apollographql/router/pull/3859/commits/a85f195794e5bd29b50f4d4e46d80a50da770a94)

.circleci/config.yml
BrynCooke: [a85f195](https://github.com/apollographql/router/pull/3859/commits/a85f195794e5bd29b50f4d4e46d80a50da770a94)

abernix (approved):

PR #3868, opened 2023-09-21
fix(telemetry): support more types for metric counters

### LINKED ISSUE #3865: Panic in DEBUG when using the wrong type with metrics macros and MetricsLayer
```
tracing::info!(
            monotonic_counter
                .apollo
                .router
                .operations
                .authentication
                .jwt = 1,
            authentication.jwt.failed = true
        )
```

it breaks because you must specify the type lie `= 1u64`

bnjjj:
Add more supported types for metric counters in `MetricsLayer`.

Now it's not mandatory and won't panic in debug mode if you don't specify `1u64` in this example:

```rust
tracing::info!(
    monotonic_counter
        .apollo
        .router
        .operations
        .authentication
        .jwt = 1,
    authentication.jwt.failed = true
)
```

Fixes #3865 

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/plugins/telemetry/metrics/layer.rs
+++ b/apollo-router/src/plugins/telemetry/metrics/layer.rs
@@ -26,7 +26,7 @@ use super::METRIC_PREFIX_VALUE;
 
 macro_rules! log_and_panic_in_debug_build {
     ($($tokens:tt)+) => {{
-        tracing::debug!($($tokens)+);
+        tracing::error!($($tokens)+);
         #[cfg(debug_assertions)]
         panic!("metric type error, see DEBUG log for details. Release builds will not panic but will still emit a debug log message");
     }};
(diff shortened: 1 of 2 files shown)


Geal (approved):

BrynCooke (approved):

apollo-router/src/plugins/telemetry/metrics/layer.rs
```diff
@@ -26,7 +26,7 @@ use super::METRIC_PREFIX_VALUE;
 
 macro_rules! log_and_panic_in_debug_build {
     ($($tokens:tt)+) => {{
-        tracing::debug!($($tokens)+);
+        tracing::error!($($tokens)+);
```
BrynCooke, line 29: Note that this does mean there is the potential to spam logs if they have their attributes incorrect.

bnjjj: Yes, but for now it's completely invisible. I'm tempted to put this either as a warn or error but it should be louder than today because it's discarded. What do you think ?

PR #3876, opened 2023-09-21
docs: minimum versions of Federation packages and Server for subscriptions

### LINKED ISSUE #3870: Specify minimal version required to use federated subscription with @apollo/subgraph
We should add a note somewhere in our documentation for people using apollo server that they have to upgrade their version because it's not available in previous versions. (I don't know which one exactly introduced the federated subscription support)

shorgi:
Fixes #3870 

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


docs/source/executing-operations/subscription-support.mdx
```diff
@@ -98,11 +98,18 @@ flowchart LR;
...
+1. **If you compose your router's supergraph schema with GraphOS** (instead of with the Rover CLI), [update your build pipeline](/graphos/graphs/updating#2-update-your-build-pipeline) to use Apollo Federation 2.4 or later.
     - Previous versions of Apollo Federation don't support subscription operations.
-4. Modify your subgraph schemas to use Apollo Federation 2.4 or later:
+1. Each package that makes up Apollo Federation needs to be updated to version 2.4 or later:
+    - `@apollo/gateway`
+    - `@apollo/query-planner`
+    - `@apollo/subgraph`
+    - `@apollo/federation-internals`
+    - `@apollo/query-graphs`
+    - `@apollo/composition`
```
bnjjj, line 111: I think we should mention it's only related to npm packages if they use apollo server

docs/source/executing-operations/subscription-support.mdx
```diff
@@ -115,6 +116,16 @@ flowchart LR;
 
     - You can _skip_ modifying subgraph schemas that don't define any `Subscription` fields.
 
+1. [Update your Apollo Server instances to version 4 or later](/apollo-server/migration).
```
bnjjj, line 119: I think it's better to use conditional here, because it's not mandatory, it's required only if your subgraphs are using Apollo Server. So I would suggest to start with something like `If you're using Apollo Server....` . I just want to not have a confusing documentation to let our users think they MUST use apollo server to make federated subscriptions work. What do you think ?

shorgi: Good suggestion

docs/source/executing-operations/subscription-support.mdx
```diff
@@ -115,6 +116,16 @@ flowchart LR;
...
 
+1. If you're using Apollo Server, [update your Apollo Server instances to version 4 or later](/apollo-server/migration).
+    - Follow the Apollo Server 4 [guide for enabling subscriptions](/apollo-server/data/subscriptions#enabling-subscriptions). 
+    - Update each `npm` package that makes up Apollo Federation to version 2.4 or later:
+        - `@apollo/gateway`
+        - `@apollo/query-planner`
+        - `@apollo/subgraph`
+        - `@apollo/federation-internals`
+        - `@apollo/query-graphs`
+        - `@apollo/composition`
```
trevor-scheer, line 127: Most of these aren't relevant to running an Apollo Server _subgraph_, and we shouldn't be recommending the installation of most of these packages either.

For Apollo Server as a subgraph, users should install `@apollo/server` and `@apollo/subgraph`.

For Apollo Server as a _gateway_, users should install `@apollo/server` and `@apollo/gateway`. However, this is a guide for router so there shouldn't be a mention of gateway here at all.

The remaining packages are all considered dependencies of `@apollo/gateway` and `@apollo/subgraph` and should not (in the majority of cases) be installed into a user's project.

PR #3878, opened 2023-09-22
Don't panic in OneShotAsyncCheckpoint.

o0Ignition0o:
Followup to #3819, this removes a panic which would trigger if a user misuses OneShotAsyncCheckpoint.


garypen (approved):
Approve. I think it's fine to merge this without a changeset.

PR #3879, opened 2023-09-22
Fix Studio reporting when it is not configured

### LINKED ISSUE #3871: v1.30.0: Traces not appearing in Apollo GraphOS/Studio
Summary, Apollo Studio reporting was broken in v1.30.0 due to a bug introduced in https://github.com/apollographql/router/pull/2999.

Updating to v1.30.1 is the best solution, but **as a work-around for v1.30.0**, it's possible to just specify the following in your Router's YAML:

```yaml
telemetry:
  apollo:
 ```

Geal:
Fix #3871

The sampling filter introduced in
b58a67c92d65b00c6c11c246e220dce9e2755de1 was making sure that the Apollo exporter was configured when the key is present, but not in the right place: that configuration is used in various places to set up information in the spans and the context, and that is used to create the Studio report and injecting the traces there.
In previous versions, the code was relying on the inject_schema_id function in router_factory.rs to always insert a configuration for the Apollo exporter.
This commit adds back the code that creates that configuration, but only if the Apollo key and graph reference are present

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


Geal:
closed in favor of #3881 that directly targets 1.31

PR #3908, opened 2023-09-26
Make release a step that is performed after tests have passed.

BrynCooke:
A new `release_approval` step is added which requires that tests have been run. Once approved either a nightly or regular release will take place depending on if the branch is tagged or not.


BrynCooke:
Let's close this for now.

PR #3919, opened 2023-09-27
Move `apollo.router.operations` to the outer layer of the router pipeline.

### LINKED ISSUE #3915: telemetry should happen first
As the router pipeline has evolved various things have crept in before the telemetry plugin. This means that we may be inaccurate operation counts and spans will be missing some processing time.

`apollo.router.operations` metric should be moved to the fist thing in the pipeline in factory.

In addition, the make span layer should also move.

BrynCooke:
Our metrics are currently mutated at the supergraph service. 
This is a throwback from when the router service did not exist.

Move operation count metrics to the first thing that happens in the pipeline. Note that this won't catch disconnects because they are reliant on the status code of the response, which will stop processing before the metric is incremented.

This also move the creation of the root span to the just after so that cors and extensions are traced.

Fixes #3915

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/axum_factory/axum_http_server_factory.rs
+++ b/apollo-router/src/axum_factory/axum_http_server_factory.rs
@@ -404,6 +404,8 @@ where
         ))
         .layer(Extension(service_factory))
         .layer(cors)
+        // Telemetry layers MUST be last. This means that they will be hit first during execution of the pipeline
+        // Adding layers after telemetry will cause us to lose metrics and spans.
         .layer(TraceLayer::new_for_http().make_span_with(PropagatingMakeSpan { license }))
         .layer(middleware::from_fn(metrics_handler));
 
--- a/apollo-router/src/axum_factory/axum_http_server_factory.rs
+++ b/apollo-router/src/axum_factory/axum_http_server_factory.rs
@@ -402,9 +402,10 @@ where
             (license, Instant::now(), Arc::new(AtomicU64::new(0))),
             license_handler,
         ))
-        .layer(TraceLayer::new_for_http().make_span_with(PropagatingMakeSpan { license }))
         .layer(Extension(service_factory))
-        .layer(cors);
+        .layer(cors)
+        .layer(TraceLayer::new_for_http().make_span_with(PropagatingMakeSpan { license }))
+        .layer(middleware::from_fn(metrics_handler));
 
     let route = endpoints_on_main_listener
         .into_iter()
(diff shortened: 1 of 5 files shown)


apollo-router/src/axum_factory/axum_http_server_factory.rs
```diff
@@ -414,6 +415,17 @@ where
     Ok(ListenAddrAndRouter(listener, route))
 }
 
+async fn metrics_handler<B>(request: Request<B>, next: Next<B>) -> Response {
+    let resp = next.run(request).await;
+    u64_counter!(
+        "apollo.router.operations",
+        "The number of graphql operations performed by the Router",
+        1,
+        "http.response.status_code" = resp.status().as_u16() as i64
```
garypen, line 424: Should that be `as u64`? It's a `u64_counter`
(I just spotted it's a copy and paste of the original, but now I'm wondering about it...)

garypen (approved):
One NIT style question about i64 vs u64.

apollo-router/src/axum_factory/axum_http_server_factory.rs
BrynCooke: Although monotonic counters are `u64` attributes on metrics are `i64`

apollo-router/src/axum_factory/axum_http_server_factory.rs
```diff
@@ -402,9 +402,10 @@ where
             (license, Instant::now(), Arc::new(AtomicU64::new(0))),
             license_handler,
         ))
-        .layer(TraceLayer::new_for_http().make_span_with(PropagatingMakeSpan { license }))
         .layer(Extension(service_factory))
-        .layer(cors);
+        .layer(cors)
+        .layer(TraceLayer::new_for_http().make_span_with(PropagatingMakeSpan { license }))
+        .layer(middleware::from_fn(metrics_handler));
```
garypen, line 408: Maybe add a comment for this layer about the importance of its position in this chain of layers.

BrynCooke: [4d67075](https://github.com/apollographql/router/pull/3919/commits/4d67075301b56ba15ed323014917efbfb4f3ccd4)

Issue #3853, opened 2023-09-19
Improve error handling for JWKS with missing alg


andyvajda:
**Describe the bug**
When a JWKS url is used in the router authentication config that includes keys without an "alg" field, none of the keys will be imported.  Instead one, or both of the following errors will be thrown:
```
ERROR: could not get url
ERROR: could not create JWKS from url content
```

**To Reproduce**
Configure router with JWKS

**Expected behavior**
Error response indicates why JWKS was unable to load.  
Possibly skip loading the offending keys while loading the others.

@flyboarder

garypen:
I don't think the problem is because an `alg` field is missing from the JWKS entries, but cause an `alg` field is specified which isn't supported by our underlying `jsonwebtoken` crate. For example: `ES512`

I've gone with an approach, #3922, which leverages our existing support for specifying a list of algorithms to use via configuration for each JWKS url. I chose this approach because it naturally aligns with the existing functionality and avoids any requirement to create extra configuration.

Here's an example
```
authentication:
  router:
    jwt:
      jwks: # This key is required.
        - url: https://my-jwks-url
          algorithms: [ "HS256", "HS384", "HS512", "ES256", "ES384", "RS256", "RS384", "RS512", "PS256", "PS384", "PS512", "EdDSA"]
```

In this case, I'm telling the router to support all the algorithms supported by our underlying `jsonwebtoken` crate. This list is now used to specify a retention list from the JWKS retrieved from the specified url. This means that if the source JWKS contains an algorithm not in the `algorithms` list, it will be ignored at download time as well as during execution time.

PR #3922, opened 2023-09-27
Ignore JWKS keys which are not supported by the router

garypen:
If you have a JWKS which contains a key which has an algorithm (alg) which the router doesn't recognise, then the entire JWKS is disregarded. This is unsatisfactory, since there are likely to be many other keys in the JWKS which the router could use.

We have changed the JWKS processing logic so that we remove entries with an unrecognised algorithm from the list of available keys. We print a warning with the name of the algorithm for each removed entry.

fixes: #3853

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [x] Manual Tests

**Exceptions**

I manually tested with a set of keys which are known to contain alg ES512 entries. I'll add some unit tests before I merge the PR if this approach is agreed to be acceptable.

**Notes**


BrynCooke (approved):

Geal:
Thisbsolution is a bit bruteforcing the problem. Could you try this approach: parde to a json Value, then for each element in the array, try to parse to a Jwk, accumulating the errors. That way, when the lib starts supporting ES512, it will be active right away, and when new algorithms appear, the JWKS parsing will already handle it

garypen:
> This solution is a bit bruteforcing the problem. Could you try this approach: parse to a json Value, then for each element in the array, try to parse to a Jwk, accumulating the errors. That way, when the lib starts supporting ES512, it will be active right away, and when new algorithms appear, the JWKS parsing will already handle it

Neat suggestion. That's what I've done: 7bca1709eb4df746b840254f6f8d220ffc73b7d5

Geal (approved):

PR #3927, opened 2023-09-28
[docs] Update Rhai docs around response headers

smyrick:
This is called out in our docs so it should probably live in the examples too

> Note: Be particularly careful when interacting with headers in a response context. For router_service(), supergraph_service() and execution_service(), response headers only exist for the first response in a deferred response stream. You can handle this by making use of the is_primary() function which will return true if a response is the first (or primary) response. If you do try to access the headers in a non-primary response, then you'll raise an exception which can be handled like any other rhai exception, but is not so convenient as using the is_primary() method.

https://www.apollographql.com/docs/router/customizations/rhai-api/#response-interface


abernix:
🙇  Don't worry about `@github-actions
NEXT_CHANGELOG.md reminder` in case you were.

bnjjj (approved):

Geal (approved):

PR #3935, opened 2023-09-29
add invariant fuzz target for the router

bnjjj:
Add an invariant fuzz target for the router which sends arbitrary bytes to the router both as a raw body and as a query


lrlna (approved):
this is greattt, let's do it!

fuzz/fuzz_targets/invariant_router.rs
```diff
@@ -0,0 +1,77 @@
+//! Fuzz target to generate random invalid body and query to the router and check it doesn't panic
+#![no_main]
```
Geal, line 2:
```suggestion
#![allow(unused_attributes)]
#![no_main]
```

PR #3950, opened 2023-10-02
Move processing of endpoints to UriEndpoint and SocketEndpoint

BrynCooke:
We have multiple places in the code that deal with endpoint parsing for otel exporters. 
This PR removes duplicated logic across exporters and adds some tests.

The new types implement default and will allow us to use serde default in a follow up PR.

Part of #3226

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/configuration/snapshots/apollo_router__configuration__tests__schema_generation.snap
+++ b/apollo-router/src/configuration/snapshots/apollo_router__configuration__tests__schema_generation.snap
@@ -4771,7 +4771,6 @@ expression: "&schema"
                 },
                 "endpoint": {
                   "description": "The endpoint to send to",
-                  "default": "default",
                   "type": "string"
                 }
               },
@@ -4813,7 +4812,6 @@ expression: "&schema"
                       "properties": {
                         "endpoint": {
                           "description": "The endpoint to send to",
-                          "default": "default",
                           "type": "string"
                         }
                       },
@@ -4949,8 +4947,7 @@ expression: "&schema"
                       "properties": {
                         "endpoint": {
                           "description": "The endpoint to send reports to",
-                          "type": "string",
-                          "format": "uri"
+                          "type": "string"
                         },
                         "password": {
                           "description": "The optional password",
@@ -5336,6 +5333,9 @@ expression: "&schema"
             "zipkin": {
               "description": "Zipkin exporter configuration",
               "type": "object",
+              "required": [
+                "endpoint"
+              ],
               "properties": {
                 "batch_processor": {
                   "description": "Batch processor configuration",
@@ -5395,7 +5395,6 @@ expression: "&schema"
                 },
                 "endpoint": {
                   "description": "The endpoint to send to",
-                  "default": "default",
                   "type": "string"
                 }
               },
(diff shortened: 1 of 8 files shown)


apollo-router/src/configuration/snapshots/apollo_router__configuration__tests__schema_generation.snap
```diff
@@ -5336,6 +5333,9 @@ expression: "&schema"
             "zipkin": {
               "description": "Zipkin exporter configuration",
               "type": "object",
+              "required": [
```
BrynCooke, line 5336: This should have been required all along.

apollo-router/src/configuration/snapshots/apollo_router__configuration__tests__schema_generation.snap
```diff
@@ -4949,8 +4947,7 @@ expression: "&schema"
                       "properties": {
                         "endpoint": {
                           "description": "The endpoint to send reports to",
-                          "type": "string",
-                          "format": "uri"
```
BrynCooke, line 4953: I may follow up to reintroduce `format` into the schema.

BrynCooke: Actually, this is a mistake. We accept default, which is not a URI, so this must just be a string.

apollo-router/src/plugins/telemetry/tracing/zipkin.rs
```diff
@@ -1,70 +1,42 @@
...
-where
-    D: Deserializer<'de>,
-{
-    let mut s = String::deserialize(deserializer)?;
-    if s == "default" {
-        return Ok(AgentEndpoint::Default(AgentDefault::Default));
-    }
-    let mut url = Url::parse(&s).map_err(serde::de::Error::custom)?;
-
-    // support the case of 'collector:4317' where url parses 'collector'
```
BrynCooke, line 46: Note this is tested in UriEndpoint:
```
        assert_eq!(
            UriEndpoint::from(Uri::from_static("example.com:2000"))
                .to_uri(&Uri::from_static("http://localhost:9411/path2"))
                .unwrap(),
            Uri::from_static("http://example.com:2000/path2")
        );
        ```

File: apollo-router/src/plugins/telemetry/endpoint.rs [hunk 4]
```diff
@@ -0,0 +1,298 @@
...
+use schemars::schema::Schema;
+use schemars::JsonSchema;
+use serde::de::Error;
+use serde::de::Visitor;
+use serde::Deserialize;
+use serde::Deserializer;
+
+#[derive(Debug, Clone, Default, Eq, PartialEq)]
+pub(crate) struct UriEndpoint {
+    uri: Option<Uri>,
```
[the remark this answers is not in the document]
BrynCooke: [14ae52b](https://github.com/apollographql/router/pull/3950/commits/14ae52bebc100f02f961f317167984e02c1f469a)

bnjjj (approved):

PR #3961, opened 2023-10-03
set a minimal size in the compression output buffer

Geal:
it should be at least large enough to hold the 10 byte gzip header. Otherwise the compressor goes into an infinite loop

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/axum_factory/compression/mod.rs
+++ b/apollo-router/src/axum_factory/compression/mod.rs
@@ -79,7 +79,7 @@ where {
                         }
                     }
                     Ok(data) => {
-                        let mut buf = BytesMut::zeroed(std::cmp::max(10, data.len()));
+                        let mut buf = BytesMut::zeroed(10 + data.len());
 
                         let mut partial_input = PartialBuffer::new(&*data);
                         let mut partial_output = PartialBuffer::new(&mut buf);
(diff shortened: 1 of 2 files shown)


BrynCooke (changes requested):
Can we have a test for compression?

Geal:
there's a test now

BrynCooke (approved):

apollo-router/src/axum_factory/compression/mod.rs
```diff
@@ -79,7 +79,9 @@ where {
                         }
                     }
                     Ok(data) => {
-                        let mut buf = BytesMut::zeroed(data.len());
+                        // the buffer needs at least 10 bytes for a gzip header if we use gzip, then more
+                        // room to store the data itself
+                        let mut buf = BytesMut::zeroed(10 + data.len());
```
garypen, line 84: 10 should be a const GZIP_HEADER_LEN.

Geal: done in [374a868](https://github.com/apollographql/router/pull/3961/commits/374a8688cb6b34b927af839388e3dafd8335e72d)

PR #3968, opened 2023-10-04
Remove the use of Option telemetry code

BrynCooke:
Removed the use of Option in config and related consuming code. This significantly simplifies the codebase.

Part of #3226

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/plugins/telemetry/apollo.rs
```diff
@@ -35,59 +34,72 @@ pub(crate) const ENDPOINT_DEFAULT: &str =
...
     pub(crate) apollo_key: Option<String>,
 
     /// The Apollo Studio graph reference.
     #[schemars(skip)]
-    #[serde(skip)]
+    #[serde(default = "apollo_graph_reference")]
     pub(crate) apollo_graph_ref: Option<String>,
 
     /// The name of the header to extract from requests when populating 'client nane' for traces and metrics in Apollo Studio.
     #[schemars(with = "Option<String>", default = "client_name_header_default_str")]
```
garypen, line 61: Is it still an Option in schemars? I find that odd anyway with a default, but that was already there I guess...

apollo-router/src/plugins/telemetry/apollo.rs
```diff
@@ -35,59 +34,72 @@ pub(crate) const ENDPOINT_DEFAULT: &str =
...
-    #[serde(deserialize_with = "deserialize_header_name")]
+    #[serde(
+        deserialize_with = "deserialize_header_name",
+        default = "client_name_header_default"
+    )]
     pub(crate) client_name_header: HeaderName,
 
     /// The name of the header to extract from requests when populating 'client version' for traces and metrics in Apollo Studio.
     #[schemars(with = "Option<String>", default = "client_version_header_default_str")]
```
garypen, line 69: Same question.

apollo-router/src/plugins/telemetry/apollo.rs
```diff
@@ -104,15 +116,17 @@ pub(crate) struct SubgraphErrorConfig {
...
 
 #[derive(Debug, Clone, Deserialize, JsonSchema)]
-#[serde(deny_unknown_fields, default)]
+#[serde(deny_unknown_fields)]
 pub(crate) struct ErrorConfiguration {
     /// Send subgraph errors to Apollo Studio
+    #[serde(default = "default_send_errors")]
     pub(crate) send: bool,
     /// Redact subgraph errors to Apollo Studio
+    #[serde(default = "default_redact_errors")]
```
garypen, line 129: I think this is going to change the serialization behaviour of this struct. Previously, the default for the type would have been used, [which is](https://doc.rust-lang.org/std/primitive.bool.html#impl-Default-for-bool) `false` for a `bool`.

The two functions will return `true`.

apollo-router/src/plugins/telemetry/config.rs
```diff
@@ -140,25 +139,25 @@ impl Default for MetricsCommon {
 
 /// Tracing configuration
 #[derive(Clone, Default, Debug, Deserialize, JsonSchema)]
-#[serde(deny_unknown_fields, rename_all = "snake_case")]
+#[serde(deny_unknown_fields, rename_all = "snake_case", default)]
```
garypen, line 142: NIT: Could we get rid of the `rename_all` here as well?

apollo-router/src/plugins/telemetry/metrics/prometheus.rs
```diff
@@ -50,7 +49,7 @@ fn prometheus_default_path() -> String {
 impl Default for Config {
     fn default() -> Self {
         Self {
-            enabled: true,
+            enabled: false,
```
garypen, line 52: It concerns me that this wasn't always `false`...

garypen (commented):
Nice cleanup. I have a few questions.

apollo-router/src/plugins/telemetry/metrics/prometheus.rs
BrynCooke: Agree, the root of our issues is this: https://github.com/serde-rs/serde/issues/2622. Depending on where the defaulting happens the value may have been true or false.

apollo-router/src/plugins/telemetry/apollo.rs
BrynCooke: The serde defaulting did not agree with the Default impl: https://github.com/apollographql/router/pull/3968/files/61b41cbe171c2b31c954da63ce13245409410079#diff-7737729cbb75d74b8921ae972883e90ba02b94cce0217b662822ad5183e966e7L119

The root of our issues is this: https://github.com/serde-rs/serde/issues/2622.

apollo-router/src/plugins/telemetry/config.rs
BrynCooke: [99476b7](https://github.com/apollographql/router/pull/3968/commits/99476b7ac34f96453ce2ac75384a867574afb0b6)

apollo-router/src/plugins/telemetry/apollo.rs
garypen: Please raise that issue.

garypen (approved):

apollo-router/src/plugins/telemetry/apollo.rs
BrynCooke: https://github.com/apollographql/router/issues/3982

PR #3969, opened 2023-10-04
test graceful shutdown with idle connections

### LINKED ISSUE #3941: Better handling of idle initial connections
If a connection to the router is opened without sending anything, it will not be closed right away by hyper's graceful shutdown feature.
In browsers like Chrome, opening a page like the health check or prometheus results in one connection created for the initial request, and another eagerly created but not used (probably in case multiple requests must be done to download assets).
If we try to close the router at this point, graceful shutdown closes the first connection, but not the second one, because it has not sent anything. The connection will close after a while, once the tab is closed in the browser or the keepalive timeout triggers.

Proposal to solve this:

In this part of the code:
https://github.com/apollographql/router/blob/beca2acb8b1abb9e7a0f4c86762f43bc9d773099/apollo-router/src/axum_factory/listeners.rs#L232-L255

`app` is an axum router. Hyper expects a tower service in the second argument of the `serve_connection` function.
We can make a wrapper service for app, that contains an `AtomicBool` shared with the task:
- once the service's `call()` method receives the request, the bool is set to true
- in the graceful shutdown code, we check for that bool. If it is false, that means we never receive the initial request, so we don't wait for the connection to shutdown, we close the task right away
- if we have received at least one header, then hyper's read timeout code kicks in and can handle idle connections after the first initial requests
- as an added bonus, we can add a timeout to cover the span between initial connection and receiving the entire request, using that bool

Geal:
Fix #3941

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/tests/lifecycle_tests.rs
+++ b/apollo-router/tests/lifecycle_tests.rs
@@ -171,6 +171,22 @@ async fn test_reload_via_sighup() -> Result<(), BoxError> {
     Ok(())
 }
 
+#[tokio::test(flavor = "multi_thread")]
+async fn test_shutdown_with_idle_connection() -> Result<(), BoxError> {
+    let mut router = IntegrationTest::builder()
+        .config(HAPPY_CONFIG)
+        .build()
+        .await;
+    router.start().await;
+    router.assert_started().await;
+    std::net::TcpStream::connect("0.0.0.0:4000").unwrap();
+    router.execute_default_query().await;
+    tokio::time::timeout(Duration::from_millis(100), router.graceful_shutdown())
+        .await
+        .unwrap();
+    Ok(())
+}
+
 async fn command_output(command: &mut Command) -> String {
     let output = command.output().await.unwrap();
     let success = output.status.success();


.changesets/fix_geal_idle_connections.md
```diff
@@ -0,0 +1,9 @@
+### Fix router hang when opening the explorer, prometheus or health check page ([Issue #3941](https://github.com/apollographql/router/issues/3941))
+
+The router's graceful shutdown was not handling idle connections well, and waiting indefinitely for them to close instead of shutting them down directly, which resulted in the router hanging on CTRL+C.
+
+This tends to happen in particular when a browser opens the explorer or prometheus page provided by the router, because browsers eagerly open new connectionsi in anticipation of future traffic, while those pages only need a single request to the router.
```
BrynCooke, line 5:
```suggestion
The Router did not gracefully shutdown when an idle connections are made by a client, and would instead hang. In particular, web browsers make such connection in anticipation of future traffic.

This is now fixed, and the Router will now gracefully shut down in a timely fashion.
```

BrynCooke (approved):
Approved with changelog suggestion

garypen (changes requested):
We have the same problem with UDS. I just manually tested and confirmed that if I have a connection open to the router which never sends any data, then shutdown will hang.

We need to add the same checker for UDS.

garypen:
I merged in additional code to address Unix Domain Sockets.

garypen (approved):

Issue #3856, opened 2023-09-20
improve security context of our container images


garypen:
Many kubernetes environments impose security restrictions on containers. For example:
 - Don't run as root
 - Can't bind to ports < 1024

Currently our container runs on port 80 as root.

It would be straightforward to address both of these points and would improve the security profile of our containers and make it easier to deploy into such restricted environments.

PR #3971, opened 2023-10-04
Improve the secure deployability of our helm chart and docker image

garypen:
Many kubernetes environments impose security restrictions on containers. For example:
 - Don't run as root
 - Can't bind to ports < 1024

Right now our container runs the router as `root` on port 80 and this makes it harder to deploy in such secure environments.

This change moves our helm chart's default port from 80 to 4000 and introduces a new user `router` in our docker container which we will use to run our router process.

fixes: #3856 

---
**Impact**

1. Changing the default port in the helm chart from 80 to 4000 is an innocuous change. I can't see how this would impact anyone.
2. Changing the default user from `root` to `router` will have an impact. You will no longer be able to `exec` to the executing container (kubernetes or docker) and install packages, etc... The container is now "locked down", by default. Good for security, but less convenient for support/debugging.

If you wish to restore the behaviour of the router to execute as `root` and listen on port 80, that can be achieved with this configuration:

```
router:
  configuration:
    supergraph:
      listen: 0.0.0.0:80
securityContext:
  runAsUser: 0
```
I don't recommend that you do this for debugging/support (there are more secure ways to achieve your goals in kubernetes), but am documenting for completeness.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [x] Manual Tests

**Exceptions**

Manual testing of the impact of this change was performed on a local PC (using docker) and in a Kubernetes cluster. The changes are working as described. It's not viable to automate this testing.

**Notes**


Geal (approved):

abernix:
Curiously: Is it still possible to `sudo` when `exec`'ing into the container?

garypen:
> Curiously: Is it still possible to `sudo` when `exec`'ing into the container?

No. We'd have to specifically install it and set up passwords etc...

garypen:
> > Curiously: Is it still possible to `sudo` when `exec`'ing into the container?
> 
> No. We'd have to specifically install it and set up passwords etc...

Additional info: A good way to debug secure stuff in kubernetes is to use something like: https://kubernetes.io/docs/tasks/debug/debug-application/debug-running-pod/#ephemeral-container

.changesets/maint_garypen_3856_docker_security.md
```diff
@@ -0,0 +1,28 @@
...
+```
+router:
+  configuration:
+    supergraph:
+      listen: 0.0.0.0:80
+securityContext:
+  runAsUser: 0
+```
+We don't recommend that you do this for debugging/support (there are more secure ways to achieve your goals in kubernetes), but we are documenting for completeness.
+
```
abernix, line 27: I'd prefer to see this CL written differently.  My suggestions are:

1.  The tense is hard to understand in isolation — e.g., when reading this "Right now" is not the case with the change.  I recommend leading with what has changed and is the _new_ state as of this version ("_Improved the security by switching the Helm port from 80 to 4000 and executing as a `router` user rather than `root`._"). **Then**, explain motivation (Kubernetes sec. requirements and best practices).  And finally, talk about restoring the previous behavior as necessary.
2. Clearly convey the update in base operating system from Bullseye to Bookworm (maybe even make it a separate CL within the same PR?) which seems material on its own but not captured right now.
3. I'd just massage the "Impact" heading into the words rather than having it as a heading.

Wdyt?

abernix (approved):
Sounds like a good idea to me.  My requested changes are on the changelog messaging.

.changesets/maint_garypen_3856_docker_security.md
garypen: done: 4eae6afea362be388b6a6e532c619a61b78f0a93

PR #3978, opened 2023-10-05
test fragment in requires directive

### LINKED ISSUE #3983: Value selection for representations should be able to infer the type
If we had a field defined like this:
```isEnabled: Boolean @join__field(graph: SUB2, requires: "thing { ... on Thing { text } }")```

where previous subgraph requests did not add a `__typename` field under `thing`, because the fragment selection only looks at the `__typename` field: https://github.com/apollographql/router/blob/3eb70b79c09c2a650ba53d6fe8075af980b9303f/apollo-router/src/query_planner/selection.rs#L105

In a lot of cases, at this point the router should be able to recognize the type using only the schema

### LINKED ISSUE #4052: Values selection will not produce a field if the field was null
may be related to #3972

In a `@key ` or `@requires` that selects an array field, like `@key(fields: "id members { id }"), if we have the following data: `{ "id": "1", "members": null }`, then the representation produced by the router would be `{ "__typename": "Ty", "id": "1" }` while it should contain a `"member": null` field

### PR DESCRIPTION
**Pull request by Geal**
related to https://github.com/apollographql/router/issues/3972
Fix #3983 (not yet, this only contains tests for now)
Fix #4052

The representation selection code in https://github.com/apollographql/router/blob/3eb70b79c09c2a650ba53d6fe8075af980b9303f/apollo-router/src/query_planner/selection.rs#L105

should be able to recognize the current type even when `__typename` is not present if it's a concrete type. This could be achieved by passing the query compiler along the selection code, or using the `Schema` representation.

**The plan right now:**
* I am trying to simplify the selection code a bit, but it will have to be refactored to follow more closely the behaviour for `@requires`: https://www.apollographql.com/docs/federation/entities-advanced/#contributing-computed-entity-fields
* I'll pass a `ParsedDocument` to `Variables::new` (getting it from context)
* we need a function that takes an `ExecutableDocument` and a `Path` and returns a `Field` with its type and selection set
* we pass that selection set to `select_object` and friends so we know exactly which type is required where

could we work with the schema instead of the query document? (since the schema is already transmitted there).
we may need to work from the query because the path depends on the query(example: there was an alias somewhere).
we would need to work with the schema because the `@requires` selection are not present in the query.

can we fix the requires selection by post-processing the query plan to add type information then caching the result?

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### CODE REVIEW

File: apollo-router/src/query_planner/selection.rs [hunk 1]
```diff
@@ -48,97 +48,171 @@ pub(crate) struct InlineFragment {
     pub(crate) selections: Vec<Selection>,
 }
 
-pub(crate) fn select_object(
-    content: &Object,
+pub(crate) fn execute_selection_set<'a>(
```
Geal, line 51: this replaces our initial implementation with a function that closely follows what is happening in the gateway's `executeSelectionSet` function

apollo-router/src/query_planner/selection.rs
```diff
@@ -48,97 +48,171 @@ pub(crate) struct InlineFragment {
...
+                        // Unfortunately, we don't validate subgraph responses yet
+                        if field_type
+                            .as_ref()
+                            .map(|ty| !ty.is_non_null())
+                            .unwrap_or(false)
+                        {
+                            output.insert(ByteString::from(selection_name.to_owned()), Value::Null);
+                        } else {
+                            return Value::Null;
+                        }
```
Geal, line 115: I am not entirely confident about this part because it differs from the gateway's implementation. The rationale from the gateway's side is that a missing field in the data we select from should not have happened, and should have been caught by response validation, but I don't see where this validation is happening

apollo-router/src/spec/query.rs
```diff
@@ -593,6 +593,31 @@ impl Query {
...
+                    if name.as_str() == TYPENAME {
+                        let input_value = input
+                            .get(field_name.as_str())
+                            .cloned()
+                            .filter(|v| v.is_string())
+                            .unwrap_or_else(|| {
+                                Value::String(ByteString::from(
+                                    parent_type.inner_named_type().as_str().to_owned(),
+                                ))
+                            });
```
Geal, line 605: this is specifically a check to replace a null `__typename` with the type we currently know. Maybe in general, we should make sure to replace the typename with the one we know from the query and the API schema

apollo-router/src/query_planner/tests.rs
```diff
@@ -985,3 +985,489 @@ async fn missing_fields_in_requires() {
...
+        .build()
+        .unwrap();
+
+    let mut stream = service.clone().oneshot(request).await.unwrap();
+    let response = stream.next_response().await.unwrap();
+    insta::assert_json_snapshot!(serde_json::to_value(&response).unwrap());
+}
+
+#[tokio::test]
+async fn typename_propagation2() {
```
Geal, line 1278: this test is still failing. I do not know yet if we should fix it (it requires some work to find out the right type) or accept that the implementation does what it can, and won't cover all cases (missing `__typename` can generally be fixed by adding it to the `@requires` selection)

apollo-router/src/query_planner/selection.rs
```diff
@@ -48,97 +48,171 @@ pub(crate) struct InlineFragment {
...
+    let content = match input_content.as_object() {
+        Some(o) => o,
+        None => return Value::Null,
+    };
+
+    current_type = current_type.or_else(|| content.get("__typename").and_then(|v| v.as_str()));
+
     let mut output = Object::with_capacity(selections.len());
     for selection in selections {
+        println!(
```
o0Ignition0o, line 66:
```suggestion
        tracing::trace!(
```

apollo-router/src/query_planner/selection.rs
o0Ignition0o: @trevor-scheer @clenfest is something you could help us figure out? 🙏

apollo-router/src/query_planner/selection.rs
```diff
@@ -48,97 +48,171 @@ pub(crate) struct InlineFragment {
...
+                type_condition,
+                selections,
+            }) => match type_condition {
+                None => continue,
+                Some(condition) => {
+                    let b = is_object_of_type(current_type, condition, schema);
+                    println!(
+                        "is_object_of_type({condition})={b} for {}",
+                        serde_json::to_string(&content).unwrap(),
+                    );
```
o0Ignition0o, line 160:
```suggestion
                    tracing::trace!(
                        "is_object_of_type({condition})={b} for {}",
                        serde_json::to_string(&content).unwrap(),
                    );
```

apollo-router/src/query_planner/selection.rs
```diff
@@ -48,97 +48,171 @@ pub(crate) struct InlineFragment {
...
-        };
-    }
-    if output.is_empty() {
-        return Ok(None);
+            },
+        }
+        println!(
+            "execute_selection_set: output is now: {}",
+            serde_json::to_string(&output).unwrap(),
+        );
```
o0Ignition0o, line 184:
```suggestion
        tracing::trace!(
            "execute_selection_set: output is now: {}",
            serde_json::to_string(&output).unwrap(),
        );
```

apollo-router/src/query_planner/selection.rs
```diff
@@ -48,97 +48,171 @@ pub(crate) struct InlineFragment {
...
+        None => return false,
+        Some(t) => t,
+    };
+
+    if conditional_type.is_interface() || conditional_type.is_union() {
+        return (object_type.is_object() || object_type.is_interface())
+            && schema.is_subtype(condition, typename);
     }
+
+    false
```
o0Ignition0o, line 215:
```suggestion
   (conditional_type.is_interface() || conditional_type.is_union()).then(|| (object_type.is_object() || object_type.is_interface())
            && schema.is_subtype(condition, typename)).unwrap_or(false)
```

not tested, i just wanted to be nitpicky 😂

apollo-router/src/query_planner/selection.rs
```diff
@@ -166,19 +241,23 @@ mod tests {
...
+                    /*match (value, selections) {
                     (Value::Object(content), requires) => {
-                        select_object(content, requires, schema).transpose()
+                        select_object(content, requires, schema) //.transpose()
                     }
-                    (_, _) => Some(Err(FetchError::ExecutionInvalidContent {
+                    (_, _) => Err(FetchError::ExecutionInvalidContent {
                         reason: "not an object".to_string(),
-                    })),
+                    }),*/
```
o0Ignition0o, line 255: uh did we remove a test? :D

o0Ignition0o (approved):
Looking good overall, i just got a bit nitpicky trying to understand things.

The failing test i d love us to investigate a bit (looks like a breaking change?)

apollo-router/src/query_planner/selection.rs
Geal: no, the `select_object` was just replaced with `execute_selection_set`

apollo-router/src/query_planner/selection.rs
BrynCooke: From memory (a long time ago) I was under the impression that we never validated responses from subgraphs as this was an expensive operation.

Geal:
> What shall we do with this? Raise a followup ticket?
This is a case that this PR can't fix. We will need to revisit this when we get deeper integration with the compiler and the new planner

PR #3994, opened 2023-10-09
Primarily use Rust validation errors when `graphql_validation_mode: both`

goto-bus-stop:
This PR upgrades to a new apollo-compiler version that supports serializing errors to JSON. `ApolloDiagnostic`s are turned into the JSON representation ASAP (as diagnostics are not serializable themselves) and threads the JSON errors through the different error types so they end up in responses unchanged. With the better quality errors, it would now be feasible to switch entirely to `graphql_validation_mode: new`.

This PR also changes the `both` validation mode so it primarily uses the Rust errors. Previously, it would run Rust validation and JS validation, compare, and then throw away the Rust errors. Now it's the other way around. I would suggest that we keep running `both` in our infra so we continue to check against the JS implementation, as we don't yet have good results from fuzzing validation yet.

This PR has one minor user-facing issue that I definitely couldn't address in time and that maybe we should solve separately, but it was also true for the JS validation. Rust validation runs **after** some query transformations, eg. defer labeling, which means that the location information in rust validation errors may be wrong. Ideally, we would run Rust validation as early as possible on the raw user input, but I got a bit stuck trying to move things around. It will be easier with compiler 1.0 and if we don't need to also compare the results to JS.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


lrlna (approved):
This looks good to me!

>This PR has one minor user-facing issue that I definitely couldn't address in time and that maybe we should solve separately, but it was also true for the JS validation. 

If this already exists with JS validation, it seems like it's something we can work on as a follow up. It might even be easier with `apollo-compiler@1.0.0`.

PR #3995, opened 2023-10-09
add support for unix domain sockets idle shutdown

garypen:
Implement the same logic as for TLS and TCP.

I manually confirmed (with: `socat - UNIX-CONNECT:/tmp/router.sock`) that the same problem existed without this fix for unix domain sockets. The problem goes away with the fix.


BrynCooke (approved):
Approved, but please open a ticket for adding a test and also improving the testing harness to make testing viable.

garypen:
> Approved, but please open a ticket for adding a test and also improving the testing harness to make testing viable.

I've filed: https://github.com/apollographql/router/issues/3996

Issue #2710, opened 2023-03-03
Use apollo-compiler to resolve type of selections in query parsing


SimonSapin:
Part of https://github.com/apollographql/router/issues/2483

[`hir::Field::ty`](https://docs.rs/apollo-compiler/0.6.0/apollo_compiler/database/hir/struct.Field.html#method.ty) could replace this logic:

https://github.com/apollographql/router/blob/4947da278c31eae49c89474f600066fbbc2a5d5c/apollo-router/src/spec/selection.rs#L75-L103

I have a local branch doing so, but some tests still fail when either introspection or type extensions are involved. This needs changes in apollo-compiler.
```[tasklist]
- [ ] https://github.com/apollographql/apollo-rs/issues/468
- [ ] https://github.com/apollographql/apollo-rs/issues/439
```

PR #4038, opened 2023-10-16
Port to apollo-compiler 1.0 beta

SimonSapin:
Fixes https://github.com/apollographql/router/issues/2710

# apollo-compiler background

The public API of apollo-compiler 1.0 is almost completely different from 0.x. Notably it doesn’t have compiler objects anymore. Instead, users mainly manipulate:

* `apollo_compiler::ast::Document` is the result of parsing one input file/string. May contain parse errors. Finding anything requires a linear scan of a `Vec` of top-level definitions.
* `apollo_compiler::Schema` has stuff indexed by name, with GraphQL extensions "applied". Conversion from AST may record "build errors" (e.g. for a name collision). Stuff with errors (e.g. that second definition with the same name) may be missing in the main data structure.
* `apollo_compiler::ExecutableDocument` similarly has operations and fragments indexed by name, has build errors, and may be missing stuff related to errors. Creating one requires `&Schema`, where it’ll find a field definition to associate with every field selection, and resolve the type of every selection set.

`Schema` and `ExecutableDocument` have `validate` methods to run full GraphQL spec validation, but it needs to be called explicitly. A future beta will likely change the API to make it less easy to forget dealing with build errors or validation errors.

# This PR

Instead of creating and passing around a `Arc<Mutex<ApolloCompiler>>` for every GraphQL request, the Router now creates a `Arc<ParsedDocument>` which contains *both* an `ast::Document` and an `ExecutableDocument`. In a later PR we’ll want to run Rust validation early and have the rest of a request lifecycle only deal with `ExecutableDocument`, but as long as we rely on TypeScript validation some things still need to work at the AST level.

~~This PR starts as a draft because of remaining failing tests:~~

* `cargo test --lib`
  - [x] plugins::telemetry::metrics::apollo::test::apollo_metrics_validation_failure
  - [x] router::tests::schema_update_test
  - [x] spec::query::tests::reformat_response_data_best_effort
* `cargo test --test integration_tests`
  - [x] api_schema_hides_field
  - [x] defer_path_with_disabled_config
  - [x] validation_errors_from_rust

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/plugins/authorization/mod.rs
```diff
@@ -233,12 +227,12 @@ impl AuthorizationPlugin {
...
     ) -> Result<Option<FilteredQuery>, QueryPlannerError> {
-        // we create a compiler to filter the query. The filtered query will then be used
+        // The filtered query will then be used
         // to generate selections for response formatting, to execute introspection and
         // generating a query plan
-        let mut compiler = ApolloCompiler::new();
-        compiler.set_type_system_hir(schema.type_system.clone());
-        let _id = compiler.add_executable(&key.filtered_query, "query");
+
+        // TODO: do we need to (re)parse here?
```
SimonSapin, line 234: I’ll open an issue to track this

SimonSapin: https://github.com/apollographql/router/issues/4040

apollo-router/src/spec/query/tests.rs
```diff
@@ -522,9 +522,9 @@ fn reformat_response_data_best_effort() {
                         "baz": "2",
                     },
                     "array": [
-                        {},
+                        {"bar": null, "baz": "3"},
                         null,
-                        {}
+                        {"bar": "5", "baz": null}
                     ],
```
SimonSapin, line 528: I don’t know:
* How did response formatting end up emitting empty maps here, before this PR?
* What in this PR changed the output for this case?

… but the new output seems to me to be more correct. @Geal what do you think?

apollo-router/src/plugins/authorization/authenticated.rs
```diff
@@ -353,20 +326,19 @@ impl<'a> transform::Visitor for AuthenticatedVisitor<'a> {
...
-            .get(node.name())
-            .ok_or("MissingFragmentDefinition")?
-            .type_condition();
+        node: &ast::FragmentSpread,
+    ) -> Result<Option<ast::FragmentSpread>, BoxError> {
+        let condition = &self
+            .fragments
+            .get(&node.fragment_name)
+            .ok_or("MissingFragment")?
+            .type_condition;
```
Geal, line 335: unrelated to this PR again, but what would happen if all the selections under a fragment were removed by the authz filter? Shouldn't we remove the spread then?

apollo-router/src/query_planner/bridge_query_planner.rs
```diff
@@ -434,45 +427,39 @@ impl Service<QueryPlannerRequest> for BridgeQueryPlanner {
...
-                    compiler_guard.db.set_token_limit(None);
-                    compiler_guard.update_executable(file_id, &modified_query);
+                    doc = Arc::new(ParsedDocumentInner {
+                        executable: modified_query.to_executable(schema),
+                        ast: modified_query,
+                    });
+                    context
+                        .private_entries
+                        .lock()
+                        .insert::<ParsedDocument>(doc.clone());
```
Geal, line 450: why is this added here?

apollo-router/src/plugins/authorization/snapshots/apollo_router__plugins__authorization__authenticated__tests__array.snap
```diff
@@ -17,7 +17,7 @@ query:
         }
         
 filtered:
-query {
+{
```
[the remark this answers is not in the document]
SimonSapin: The short syntax is allowed when a query operation has no name, variable, or directive. I made the new serializer use it when possible without realizing apollo-encoder doesn’t.

apollo-router/src/plugins/authorization/mod.rs
```diff
@@ -247,97 +241,76 @@ impl AuthorizationPlugin {
...
-        let compiler = match filter_res {
-            None => compiler,
-            Some((query, paths)) => {
+        let doc = match filter_res {
+            None => doc,
+            Some((filtered_doc, paths)) => {
                 unauthorized_paths.extend(paths);
 
-                if query.is_empty() {
+                // FIXME: consider only `filtered_doc.get_operation(key.operation_name)`?
```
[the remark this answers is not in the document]
SimonSapin: Added to https://github.com/apollographql/router/issues/4040

apollo-router/src/spec/query/traverse.rs
```diff
@@ -1,198 +1,188 @@
...
+                    FieldLookupError::NoSuchField(_, _) => {
+                        format!("no field `{}` in type `{parent_type}`", &def.name)
+                    }
+                })?
+                .clone();
+            visitor.field(parent_type, field_def, def)
+        }
+        ast::Selection::FragmentSpread(def) => visitor.fragment_spread(def),
+        ast::Selection::InlineFragment(def) => {
+            let fragment_type = def.type_condition.as_deref().unwrap_or(parent_type);
```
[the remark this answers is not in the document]
SimonSapin: That issue is the same before and after this PR, right? The previous code had `let parent_type = hir.type_condition().unwrap_or(parent_type);` for inline fragments. Should this be filed as a new issue?

apollo-router/src/services/layers/allow_only_http_post_mutations.rs
```diff
@@ -264,20 +254,20 @@ mod forbid_http_get_mutations_tests {
 
     fn create_request(method: Method, operation_kind: OperationKind) -> SupergraphRequest {
         let query = match operation_kind {
-            OperationKind::Query => "query { a }",
-            OperationKind::Mutation => "mutation { a }",
+            OperationKind::Query => "query { a }\n type Query { a: Int }",
```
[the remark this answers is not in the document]
SimonSapin: an `ExecutableDocument` is already created for a given schema. Without a `Query` definition there would be an error about undefined root operation type. `to_mixed` below creates both a `Schema` and `ExecutableDocument` from the same input document.

apollo-router/src/services/layers/query_analysis.rs
```diff
@@ -152,4 +149,9 @@ impl QueryAnalysisLayer {
     }
 }
 
-pub(crate) struct Compiler(pub(crate) Arc<Mutex<ApolloCompiler>>);
+pub(crate) type ParsedDocument = Arc<ParsedDocumentInner>;
```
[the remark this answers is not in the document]
SimonSapin: Yes! Everything is `Send + Sync`. All mutation happens through `Arc::make_mut`.

If we want shared ownership _and_ mutation seen by all owners then a mutex is still needed. Here `Context::private_entries` provides that mutex.

apollo-router/src/spec/query.rs
```diff
@@ -472,51 +456,53 @@ impl Query {
...
+                if parameters
+                    .schema
+                    .definitions
+                    .get_scalar(type_name)
+                    .is_some()
+                {
                     *output = input.clone();
                     return Ok(());
-                } else if let Some(enum_type) = definitions.enums.get(type_name) {
+                } else if let Some(enum_type) = parameters.schema.definitions.get_enum(type_name) {
```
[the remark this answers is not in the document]
SimonSapin: Done!

apollo-router/src/spec/query.rs
```diff
@@ -536,13 +522,20 @@ impl Query {
...
+                            if parameters
+                                .schema
+                                .definitions
+                                .get_object(input_type)
+                                .is_none()
+                                && parameters
+                                    .schema
+                                    .definitions
+                                    .get_interface(input_type)
+                                    .is_none()
```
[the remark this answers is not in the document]
SimonSapin: What does the API schema look like in this example? Does it have an interface without any implementer? That feels wrong exactly for this reason: what is `__typename` supposed to return?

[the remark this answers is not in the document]
SimonSapin: :thinking: 

https://specs.apollo.dev/inaccessible/v0.2/#note-833f6
> Applying this process may result in an invalid schema. For example, fields which return `@inaccessible` types which are not themselves marked `@inaccessible` will now return an invalid type which is not present in the schema. This is intentional. `@inaccessible` does NOT cascade. If applying `@inaccessible` results in an invalid schema, the serving process [SHOULD](https://tools.ietf.org/html/rfc2119#section-3) apply standard polices to determine whether or how to serve it. Generally, invalid schemas [SHOULD NOT](https://tools.ietf.org/html/rfc2119#section-4) be served, though some server configurations—particularly those used for development—may [OPTIONAL](https://tools.ietf.org/html/rfc2119#section-5)LY elect to serve such schemas in a degraded mode. The semantics of such a mode are not within the scope of this spec.

apollo-router/src/spec/selection.rs
```diff
@@ -59,99 +59,61 @@ impl Selection {
...
-                                        .inner_type_name()
-                                        .map(ToString::to_string)
-                                        .unwrap_or_else(|| current_type.to_string()),
-                                )
-                            })?
-                            .ty()
-                            .into()
-                    }
-                };
+                let field_type = FieldType::from(field.ty());
```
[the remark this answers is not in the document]
SimonSapin: Yes. Meta-fields are not part of the explicit `ObjectType::fields` map but they are returned by [`Schema::type_field(type_name, field_name)`](https://docs.rs/apollo-compiler/1.0.0-beta.4/apollo_compiler/schema/struct.Schema.html#method.type_field) which is called to associate field selections with their definitions

apollo-router/src/services/supergraph_service.rs
```diff
@@ -1451,7 +1451,6 @@ mod tests {
             .unwrap();
         let mut stream = service.oneshot(request).await.unwrap();
         let res = stream.next_response().await.unwrap();
-        assert_eq!(&res.data, &Some(serde_json_bytes::Value::Null));
```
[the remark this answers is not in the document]
SimonSapin: The following line is `insta::assert_json_snapshot!(res);` which will also fail if that `assert_eq` would fail, but give more useful info (namely, include `res.errors`)

apollo-router/src/plugins/authorization/authenticated.rs
```diff
@@ -127,79 +115,91 @@ impl<'a> traverse::Visitor for AuthenticatedCheckVisitor<'a> {
...
         }
         false
     }
 
-    fn implementors_with_different_type_requirements(&self, t: &TypeDefinition) -> bool {
-        if t.is_interface_type_definition() {
+    fn implementors_with_different_type_requirements(
+        &self,
+        type_name: &str,
+        t: &schema::ExtendedType,
```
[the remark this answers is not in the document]
SimonSapin: Yes it’s a bit weird. I’m not sure it should stay that way but right now `ExtendedType` does not contain the type’s name. This is because `Schema` contains an `IndexMap<Name, ExtendedType>`, so the name is already in the map key. With everything being mutable, omitting it in the map value eliminates the possibility of inconsistency.

However this pattern is not applied everywhere. For example object fields and type definitions are in a name-indexed map but include a name in the map value. We should pick one pattern or the other and stick to it.

apollo-router/src/plugins/authorization/authenticated.rs
SimonSapin: Yes we probably should :sweat_smile:

apollo-router/src/query_planner/bridge_query_planner.rs
SimonSapin: This replaces `compiler_guard.update_executable()`. Now that apollo-compiler does not provide interior mutability anymore we rely on the one from `context.private_entries` instead.

apollo-router/src/spec/query/traverse.rs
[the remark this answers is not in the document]
Geal: yes, a new issue sounds good

apollo-router/src/services/layers/query_analysis.rs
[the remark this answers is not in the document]
Geal: I think we'll basically never need shared mutation in the router

apollo-router/src/spec/query.rs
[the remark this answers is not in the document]
Geal: this does not say anything about __typename though

apollo-router/src/plugins/authorization/authenticated.rs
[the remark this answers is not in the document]
Geal: ok, I see

apollo-router/src/query_planner/bridge_query_planner.rs
Geal: ok, I see

o0Ignition0o (commented):
halfway through

apollo-router/src/plugins/authorization/authenticated.rs
```diff
@@ -1,84 +1,74 @@
...
+    ) -> Result<(), BoxError> {
+        let field_name = &node.name;
         if self
-            .compiler
-            .db
-            .types_definitions_by_name()
-            .get(parent_type)
-            .and_then(|def| def.field(&self.compiler.db, field_name))
+            .schema
+            .type_field(field_name, field_name)
```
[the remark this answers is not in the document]
SimonSapin: that definitely looks like a bug, good catch!

bnjjj (approved):

apollo-router/src/services/layers/persisted_queries/manifest_poller.rs
```diff
@@ -679,7 +661,7 @@ mod tests {
         ]));
 
         let is_allowed =
-            |body: &str| -> bool { safelist.is_allowed(body, Parser::new(body).parse()) };
+            |body: &str| -> bool { safelist.is_allowed(body, &ast::Document::parse(body, "")) };
```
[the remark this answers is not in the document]
bnjjj: More an apollo-rs question, but why not using an `Option` then ?

Geal (approved):

apollo-router/src/services/layers/persisted_queries/mod.rs
```diff
@@ -175,33 +171,20 @@ impl PersistedQueryLayer {
...
                         ErrorCacheStrategy::DontCache,
                     ));
                 }
-                Some(c) => c.0.clone(),
-            }
-        };
-
-        let compiler_guard = compiler.lock().await;
-        let db = &compiler_guard.db;
-        let file_id = match db.source_file(QUERY_EXECUTABLE.into()) {
```
o0Ignition0o, line 197: Is the operation now always present? (lot of context i m missing so just asking ^^)

apollo-router/src/services/layers/persisted_queries/manifest_poller.rs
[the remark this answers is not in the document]
SimonSapin: To avoid encouraging everyone to be lazy like i was here :p

apollo-router/src/spec/query/tests.rs
o0Ignition0o: @Geal we went through it while working on https://github.com/apollographql/router/pull/3718

IIRC I had a change that would output this, and we decided to not fix it that way, but I don't remember exactly why.
Can we walk through it again together tomorrow? gotta be 100% sure it's ok

apollo-router/src/spec/schema.rs
```diff
@@ -63,53 +61,39 @@ impl Schema {
...
-        let mut compiler = ApolloCompiler::new();
-        let id = compiler.add_type_system(sdl, "schema.graphql");
-
-        let ast = compiler.db.ast(id);
+        let mut parser = apollo_compiler::Parser::new();
+        let ast = parser.parse_ast(sdl, "schema.graphql");
 
         // Trace log recursion limit data
-        let recursion_limit = ast.recursion_limit();
+        let recursion_limit = parser.recursion_reached();
```
o0Ignition0o, line 68: recursion_limit_reached?

Oh nvm it's the compiler's public api?

apollo-router/tests/integration_tests.rs
```diff
@@ -110,20 +110,21 @@ async fn api_schema_hides_field() {
...
         actual.errors[0].extensions["code"].as_str(),
-        Some("GRAPHQL_VALIDATION_FAILED"),
+        Some("PARSING_ERROR"),
     );
 }
 
 #[tokio::test(flavor = "multi_thread")]
 async fn validation_errors_from_rust() {
     let request = supergraph::Request::fake_builder()
-        .query(r#"{ topProducts { name inStock } notAField }"#)
```
o0Ignition0o, line 126: why wouldn't this fail ?

apollo-router/tests/snapshots/integration_tests__validation_errors_from_rust.snap
```diff
@@ -4,23 +4,17 @@ expression: response.errors
...
   },
   {
-    "message": "cannot query field `notAField` on type `Query`",
+    "message": "compiler error: the argument `notAnArg` is not supported",
     "locations": [
       {
-        "line": 6,
-        "column": 3
+        "line": 1,
+        "column": 22
```
o0Ignition0o, line 17: I thought the locations refered to the graphql operation, if so why do the lines and columns change?

o0Ignition0o (commented):
Looking great so far! 8 files remaining and a couple of comments / questions

apollo-router/src/services/layers/query_analysis.rs
```diff
@@ -51,8 +50,8 @@ impl QueryAnalysisLayer {
         }
     }
 
-    pub(crate) fn make_compiler(&self, query: &str) -> (ApolloCompiler, FileId) {
-        Query::make_compiler(query, self.schema.api_schema(), &self.configuration)
+    pub(crate) fn parse_document(&self, query: &str) -> ParsedDocument {
+        Query::parse_document(query, self.schema.api_schema(), &self.configuration)
```
lrlna, line 54: I would love to see a future where we also don't reparse operations. Given how incredibly large some operations can be, a reparse would be significant. Perhaps a problem for future us.

apollo-router/src/spec/field_type.rs
```diff
@@ -31,14 +31,18 @@ impl Serialize for FieldType {
             {
                 #[derive(Serialize)]
                 enum NestedBorrowed<'a> {
-                    NonNull(BorrowedFieldType<'a>),
-                    List(BorrowedFieldType<'a>),
                     Named(&'a str),
+                    NonNullNamed(&'a str),
+                    List(BorrowedFieldType<'a>),
+                    NonNullList(BorrowedFieldType<'a>),
```
lrlna, line 37: Hmm these types aren't quite inline with the spec. Is it just to make it easier with the borrows?

apollo-router/src/spec/field_type.rs
```diff
@@ -80,114 +82,83 @@ impl std::fmt::Display for FieldType {
 }
 
 fn validate_input_value(
```
lrlna, line 84: Validation should definitely live in `apollo-rs`. I take it this is execution-specific validation? (i.e. we can't quite strip this out of the router)

apollo-router/src/spec/schema.rs
lrlna: it's the compiler api

lrlna (approved):
HUGE!

apollo-router/src/plugins/telemetry/metrics/apollo.rs
```diff
@@ -310,7 +310,7 @@ mod test {
 
     #[tokio::test(flavor = "multi_thread")]
     async fn apollo_metrics_validation_failure() -> Result<(), BoxError> {
-        let query = "query {topProducts{unknown}}";
+        let query = "query {topProducts(minStarRating: 4.7){name}}";
```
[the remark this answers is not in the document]
SimonSapin: This test needs TypeScript validation to run, but an undefined field now causes an error to be triggered before that happens. IIRC, when the query visitor tries to find the field definition. Switching to an undefined argument lets the visitor run successfully but have TypeScript validation still fail.

apollo-router/src/services/layers/persisted_queries/mod.rs
SimonSapin: Before this PR:

* Router context private entries contain `Arc<Mutex<ApolloCompiler>>`
* `ApolloCompiler` contains any number of files, identified by `FileId`, each with an associated "file name"
* This particular line looks for a file named `"query.graphql"` and finds its ID. We expect it to always exist so `unwrap` would not be wrong, but in case of a bug causing it to be missing we return an error instead of panicking.
* With that ID, we can later query the equivalent of `Arc<ExecutableDocument>` from the compiler

After this PR:

* Router context private entries contain `Arc<ParsedDocument>`
* `ParsedDocument` (a new Router struct) contains `ExecuableDocument`

File IDs still exist in apollo-compiler 1.0 but users almost never need to care about them.

apollo-router/src/query_planner/fetch.rs
```diff
@@ -61,6 +61,26 @@ impl OperationKind {
     }
 }
 
+impl From<OperationKind> for apollo_compiler::ast::OperationType {
```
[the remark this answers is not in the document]
Geal: `OperationKind` is the internal type on the router side, that comes from the query plan IIRC

apollo-router/src/spec/schema.rs
SimonSapin: Yes [`recursion_reached`](https://docs.rs/apollo-compiler/1.0.0-beta.4/apollo_compiler/struct.Parser.html#method.recursion_reached) is from apollo-compiler. It says how much recursion was actually used/reached during the last parse, even if the limit was not exceeded. This can be useful to decide how high/low to configure the limit.

apollo-router/tests/integration_tests.rs
SimonSapin: Same story as https://github.com/apollographql/router/pull/4038#discussion_r1366699959

apollo-router/tests/snapshots/integration_tests__validation_errors_from_rust.snap
SimonSapin: The input changed to have a different error: https://github.com/apollographql/router/pull/4038#discussion_r1365851704
The new error has a different message and location

apollo-router/src/router/mod.rs
```diff
@@ -487,11 +487,11 @@ mod tests {
         let response = router_handle.request(request).await.unwrap();
 
         assert_eq!(
-            "cannot query field 'name' on type 'User'",
-            response.errors[0].message
+            "parsing error: no field `name` in type `User`", response.errors[0].message,
```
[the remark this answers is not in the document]
SimonSapin: After we fully switch to Rust validation, we can run it earlier in the pipeline and work on error messages there.

I believe `parse error:` here comes from the enum variant used when the query visitor returns an error. With early validation that error path should become unreachable.

apollo-router/src/services/layers/query_analysis.rs
SimonSapin: This PR already eliminates the serialize + reparse step at the end of every query transform visitor. There’s one maybe-unnecessary maybe-reparse in `AuthorizationPlugin::filter_query` tracked in https://github.com/apollographql/router/issues/4040, but other than that this function is used for the initial parse.

apollo-router/src/query_planner/fetch.rs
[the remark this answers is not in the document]
Geal: it comes from the query planner's `FetchNode` structure: https://github.com/apollographql/router/blob/dev/apollo-router/src/query_planner/fetch.rs#L67-L86
For now we are tightly integrated with it, but if apollo-compiler's type can deserialize in the same way then we can use it instead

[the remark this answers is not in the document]
Geal: there's duplication here mainly because that `OperationKind` predates the apollo-compiler by a very long time, and we have not updated everything in the process

apollo-router/src/spec/field_type.rs
SimonSapin: Having an ad-hoc rust enum here at all is to deal with borrows. The change in variants is for a corresponding one [in the compiler 1.0 AST](https://docs.rs/apollo-compiler/1.0.0-beta.4/apollo_compiler/ast/enum.Type.html):

```rust
pub enum Type {
    Named(NamedType),
    NonNullNamed(NamedType),
    List(Box<Type>),
    NonNullList(Box<Type>),
}
```

This representation is more in line with the spec than compiler 0.11 which allowed non-null of non-null. The spec excludes non-null of non-null [in its grammar](https://spec.graphql.org/October2021/#Type):

* Type:
  -  NamedType
  - ListType
  - NonNullType
* NamedType:
  - Name
* ListType:
  - `[` Type `]`
* NonNullType
  - NamedType `!`
  - ListType `!`

Which simplifies to:

* Type:
  - Name
  - `[` Type `]`
  - Name `!`
  - `[` Type `]` `!`

apollo-router/src/spec/field_type.rs
SimonSapin: This is what the spec calls [CoerceVariableValues](https://spec.graphql.org/October2021/#CoerceVariableValues()) except instead of returning `Result<CoercedValues, RequestError>` this version only returns a boolean.

So far apollo-compiler doesn’t deal with variable values at all, but i agree it makes sense to add this to the project scope. Introspection needs it, and https://github.com/apollographql/apollo-rs/pull/645 includes it. I commented in https://github.com/apollographql/apollo-rs/pull/645#issuecomment-1772432060.

apollo-router/src/services/layers/query_analysis.rs
lrlna: gotcha!

apollo-router/src/query_planner/fetch.rs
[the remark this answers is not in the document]
lrlna: ahh cool cool one more thing to look forward to simplifying with this replatforming

apollo-router/src/spec/field_type.rs
```diff
@@ -80,114 +82,83 @@ impl std::fmt::Display for FieldType {
...
+        .get(type_name)
+        .ok_or(InvalidValue)?;
+    match (type_def, value) {
+        // Custom scalar: accept any JSON value
+        (schema::ExtendedType::Scalar(_), _) => Ok(()),
+
+        // TODO: check enum value?
+        // (schema::ExtendedType::Enum(def), Value::String(s)) => {
+        //     from_bool(def.values.contains_key(s))
+        // },
```
o0Ignition0o, line 136: do we have followup tickets for todos in this file ?

apollo-router/src/plugins/telemetry/metrics/apollo.rs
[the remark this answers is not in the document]
o0Ignition0o: I see, thanks!

apollo-router/src/spec/schema.rs
o0Ignition0o: thanks!

apollo-router/tests/snapshots/integration_tests__validation_errors_from_rust.snap
o0Ignition0o: thanks!

o0Ignition0o (approved):
LGTM Overall.

3 concerns:
- I wonder why https://github.com/apollographql/router/pull/4038#discussion_r1365453718 wasn't caught by a test :/
- Still unsure about https://github.com/apollographql/router/pull/4038#discussion_r1365843363
- This PR was fairly big, and I can't be 100% positive i understood everything and caught everything. I'm not sure what our testing story looks like, and how we'll catch potential bugs.

apollo-router/src/spec/field_type.rs
SimonSapin: https://github.com/apollographql/router/issues/4040

apollo-router/src/services/layers/persisted_queries/mod.rs
```diff
@@ -677,7 +660,7 @@ mod tests {
     async fn pq_layer_freeform_graphql_with_safelist() {
         let manifest = HashMap::from([(
             "valid-syntax".to_string(),
-            "fragment A on T { a }    query SomeOp { ...A ...B }    fragment,,, B on U{b c  } # yeah"
+            "fragment A on Query { me { id } }    query SomeOp { ...A ...B }    fragment,,, B on Query{me{name,username}  } # yeah"
```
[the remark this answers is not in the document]
SimonSapin: At some point while working on this PR I tried to make this code rely on `ExecutableDocument` instead of AST, which required queries to be valid for some schema. This turned out to cause other tests to fail because some checks happened earlier in the request pipelines, so I reverted back to AST. Quite possibly these test changes turned out unnecessary then.

After we fully migrate away from graphql-js validation and intentionally run apollo-rs validation early in the pipeline, I still plan to reduce AST usage and migrate things to `ExecutbleDocument` as much as possible.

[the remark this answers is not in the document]
glasser: Makes sense. For the purpose of safelisting I think we should try to keep the safelist check before the validity check, though.

Issue #3400, opened 2023-07-10
Simplify service request types


Geal:
Right now we carry a lot of data from service to service:

Router service:
```rust
pub struct Request {
    /// Original request to the Router.
    pub router_request: http::Request<Body>,

    /// Context for extension
    pub context: Context,
}
```

Supergraph service:
```rust
pub struct Request {
    /// Original request to the Router.
    pub supergraph_request: http::Request<graphql::Request>,

    /// Context for extension
    pub context: Context,
}
```

Execution service:
```rust
pub struct Request {
    /// Original request to the Router.
    pub supergraph_request: http::Request<graphql::Request>,

    pub query_plan: Arc<QueryPlan>,

    pub context: Context,
}
```

Subgraph service:

```rust
pub struct Request {
    /// Original request to the Router.
    pub supergraph_request: Arc<http::Request<graphql::Request>>,

    pub subgraph_request: http::Request<graphql::Request>,

    pub operation_kind: OperationKind,

    pub context: Context,

    /// Channel to send the subscription stream to listen on events coming from subgraph in a task
    pub(crate) subscription_stream: Option<mpsc::Sender<HandleStream<String, graphql::Response>>>,
    /// Channel triggered when the client connection has been dropped
    pub(crate) connection_closed_signal: Option<broadcast::Receiver<()>>,
}
```

These types make service manipulation complex (especially the fact the `http::Request` is not easily clonable).

## Proposal

*this is a breaking change*

we do not need to carry HTTP requests along the entire pipeline. If a supergraph, execution or subgraph plugin needs access to the request headers or the request line, they could set a hook point at the router service level, extract the header, store it in their context, and then use it where needed. This is what we already do in the content negotiation layer.
With this approach, the request types would only carry what is needed, which means less cloning, less memory usage per session, easier request type manipulation.

The request types could look like this:

Router service:
```rust
pub struct Request {
    /// Original request to the Router.
    pub router_request: http::Request<Body>,

    /// Context for extension
    pub context: Context,
}
```

Supergraph service:
```rust
pub struct Request {
    pub graphql_request: graphql::Request,

    /// Context for extension
    pub context: Context,
}
```

Execution service:
```rust
pub struct Request {
    pub query_plan: Arc<QueryPlan>,

    pub context: Context,
}
```

Subgraph service:

```rust
pub struct Request {
    pub subgraph_request: http::Request<graphql::Request>,

    pub context: Context,

[...]

PR #4057, opened 2023-10-18
Propagate headers for source stream events with subscription

### LINKED ISSUE #3731: Subscriptions: Headers for source stream events aren't propagated to subgraphs
STRs:

Configure the router to forward Authorization header to all subgraphs.
Open a subscription with an authorization header.

Expected:
Subsequent subgraph calls propagated by the source stream events contain the Authorization header initially provided by the client.

Actual:
No header is forwarded.

Caveat:
Let's be 100% sure dedup still works as expected.

bnjjj:
Fixes #3731

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/services/supergraph_service.rs
+++ b/apollo-router/src/services/supergraph_service.rs
@@ -293,21 +293,8 @@ async fn service_call(
                     let (subs_tx, subs_rx) = mpsc::channel(1);
                     let query_plan = plan.clone();
                     let execution_service_factory_cloned = execution_service_factory.clone();
-                    let mut cloned_supergraph_req = SupergraphRequest::builder()
-                        .extensions(req.supergraph_request.body().extensions.clone())
-                        .and_query(req.supergraph_request.body().query.clone())
-                        .context(context.clone())
-                        .method(req.supergraph_request.method().clone())
-                        .and_operation_name(req.supergraph_request.body().operation_name.clone())
-                        .uri(req.supergraph_request.uri().clone())
-                        .variables(req.supergraph_request.body().variables.clone());
-
-                    for (header_name, header_value) in req.supergraph_request.headers().clone() {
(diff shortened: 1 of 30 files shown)


Geal (approved):
Approved, with the caveat that the current request types encourage wrong decisions, like having the headers plugin extract headers from the supergraph request at the subgraph service plugin, instead of extracting at the router service level, putting the data in the context and then using in the subgraph service.
All of that will have to change when we implement  https://github.com/apollographql/router/issues/3400

apollo-router/src/services/supergraph_service.rs
```diff
@@ -293,6 +293,21 @@ async fn service_call(
...
+                        .and_operation_name(req.supergraph_request.body().operation_name.clone())
+                        .uri(req.supergraph_request.uri().clone())
+                        .variables(req.supergraph_request.body().variables.clone());
+
+                    for (header_name, header_value) in req.supergraph_request.headers().clone() {
+                        if let Some(header_name) = header_name {
+                            cloned_supergraph_req =
+                                cloned_supergraph_req.header(header_name, header_value);
+                        }
+                    }
```
garypen, line 310: Maybe make this into a function and call the function at line 533?

bnjjj: I wanted to make a function here but, it requires to clone the context everytime because we're using it before, so I didn't want to create a clone for every operations instead of only for subscription

bnjjj: Fixed in https://github.com/apollographql/router/pull/4057/commits/c15c60286fab0528970ccadf1b6579f12447b5dc

garypen (approved):

PR #4061, opened 2023-10-19
Telemetry roadmap config

BrynCooke:
This PR creates a config structure that will set us up a better telemetry future.

Documentation is current in progress.

The existing configuration has grown organically, and now that inconsistencies have been removed we can now look to adding new features.

This PR contains a feature gate: `telemetry_next` that must be enabled to get the new config.

Run the following to generate the schema for exploration:
```
cargo run --features telemetry_next -- --schema > schema.json
```

The goals of this config are:

* Follow the Otel specs. There's plenty of standard attributes and instruments as part of the spec.
* Custom events and instruments.
* Standardize access to data at the relevant service.
* Consistency across metrics/tracing/logging.

Example config:

```yaml
telemetry:

  logging:
    common:
      service_name: router
    stdout:
      enabled: false
      format: bunyan
    file:
      enabled: false
      format: bunyan

  tracing:
    common: # Renamed from trace_config
      max_attributes_per_event: 128
      max_attributes_per_span: 128
      max_attributes_per_link: 128
      max_events_per_span: 128
      max_links_per_span: 128
      parent_based_sampler: true
      sampler: always_on
      service_name: router
      service_namespace: "default"
      resource:
        d: e

      # Resources are otel config not represented in the yaml config


    propagation:
      baggage: false
      jaeger: false
      datadog: false
      request:
        header_name: "X-REQUEST-ID"
      trace_context: false
      zipkin: false

    otlp:
      enabled: true
      endpoint: "http://localhost:4317/v1/traces"




  metrics:
    common:
      service_namespace: "default"
      service_name: router
      buckets:
        - 0.1

      resource:
        test: foo
    prometheus:
      enabled: true
      path: /metrics
    otlp:
      enabled: true


  instruments:
    default_attribute_requirement_level: required
    router:
      http.server.active_requests: true
      my_instrument:
        value: unit
        type: counter
        unit: kb
        description: "my description"
        event: on_error
        attributes:
          http.response.status_code: false
          "my_attribute":
            response_header: "X-MY-HEADER"
            default: "unknown"
            redact: "foo"

    supergraph:
      my_instrument:
        value: unit
        event: on_error
        type: counter
        unit: kb
        description: "my description"
    subgraph:
      my_instrument:
        value: unit
        event: on_error
        type: counter
        unit: kb
        description: "my description"


  events:
    router:
      request: true
      response: false
      error: false
      test:
        message: "foo"
        level: info
        attributes:
          http.response.body.size: false


  spans:
    default_attribute_requirement_level: required
    legacy_request_span: true
    # The request span will be disappearing
    # router is the new root span
    router:
      attributes:
        dd.trace_id: false
        http.request.body.size: false
        http.response.body.size: false
        http.request.method: false
        http.request.method.original: false
        http.response.status_code: false
        network.protocol.name: false
        network.protocol.version: false
        network.transport: false
        error.type: false
        network.type: false
        trace_id: false
        user_agent.original: false
        client.address: false
        client.port: false
        http.route: false
        network.local.address: false
        network.local.port: false
        network.peer.address: false
        network.peer.port: false
        server.address: false
        server.port: false
        url.path: false
        url.query: false
        url.scheme: false
        "x-custom1":
          trace_id: datadog
        "x-custom2":

          response_header: "X-CUSTOM2"

          default: "unknown"
        "x-custom3":
          request_header: "X-CUSTOM3"
        "x-custom5":
          response_context: "X-CUSTOM3"
        "x-custom8":
          env: "ENV_VAR"

      #etc...
    supergraph:
      attributes:
        graphql.document: false
        graphql.operation.name: true
        graphql.operation.type: true

        "x-custom":
          query_variable: "arg1"
          default: "unknown"
          redact: ""
        "x-custom2":
          response_body: "arg2"
        "x-custom4":
          request_context: "X-CUSTOM3"
        "x-custom5":
          response_context: "X-CUSTOM3"
        "x-custom6":
          operation_name: string
        "x-custom7":
          operation_name: hash
        "x-custom8":
          env: "ENV_VAR"
      #etc...
    subgraph:

      attributes:

        graphql.federation.subgraph.name: false
        graphql.operation.name: false
        graphql.operation.type: true
        "x-custom":
          subgraph_operation_name: string
          default: "unknown"
        "x-custom2":
          subgraph_response_body: "arg2"
        "x-custom4":
          request_context: "X-CUSTOM3"
        "x-custom5":
          response_context: "X-CUSTOM3"
```

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


bnjjj:
Is it a draft PR ? If not I would suggest to check if we can just hide it from the json schema

BrynCooke:
@bnjjj Yes, sorry it's still draft

apollo-router/src/plugins/telemetry/config.rs
```diff
@@ -56,12 +56,26 @@ pub(crate) struct Conf {
...
     pub(crate) tracing: Tracing,
     /// Apollo reporting configuration
     pub(crate) apollo: apollo::Config,
+
+    #[cfg(feature = "telemetry_next")]
+    /// Event configuration
+    pub(crate) events: config_new::events::Events,
+    #[cfg(feature = "telemetry_next")]
+    /// Span configuration
+    pub(crate) spans: config_new::spans::Spans,
```
bnjjj, line 75: as spans are only related to tracing why don't we put that in `tracing` section ?

apollo-router/src/plugins/telemetry/config.rs
```diff
@@ -56,12 +56,26 @@ pub(crate) struct Conf {
...
     /// Metrics configuration
     pub(crate) metrics: Metrics,
     /// Tracing configuration
     pub(crate) tracing: Tracing,
     /// Apollo reporting configuration
     pub(crate) apollo: apollo::Config,
+
+    #[cfg(feature = "telemetry_next")]
+    /// Event configuration
+    pub(crate) events: config_new::events::Events,
```
bnjjj, line 72: Is an event a log ?

apollo-router/src/plugins/telemetry/config_new/attributes.rs
```diff
@@ -0,0 +1,702 @@
...
+        /// Optional redaction pattern.
+        redact: Option<String>,
+        /// Optional default value.
+        default: Option<String>,
+    },
+    SubgraphOperationKind {
+        /// The kind of the subgraph operation (query|mutation|subscription).
+        subgraph_operation_kind: OperationKind,
+        /// Optional default value.
+        default: Option<String>,
```
bnjjj, line 279: Do we need a default value for that ? a subgraph operation should always have an operation kind

apollo-router/src/plugins/telemetry/config_new/attributes.rs
```diff
@@ -0,0 +1,702 @@
...
+        subgraph_query_variable: String,
+        /// Optional redaction pattern.
+        redact: Option<String>,
+        /// Optional default value.
+        default: Option<String>,
+    },
+    SubgraphResponseBody {
+        /// The subgraph response body json path.
+        subgraph_response_body: String,
+        /// Optional default value.
```
bnjjj, line 292: Should we also have a way to `redact` it ?

apollo-router/src/plugins/telemetry/config_new/attributes.rs
```diff
@@ -0,0 +1,702 @@
...
+        default: Option<String>,
+    },
+    SupergraphOperationKind {
+        /// The supergraph query operation kind (query|mutation|subscription).
+        supergraph_operation_kind: OperationKind,
+        /// Optional redaction pattern.
+        redact: Option<String>,
+        /// Optional default value.
+        default: Option<String>,
```
bnjjj, line 326: same than previously, I think default here is useless

apollo-router/src/plugins/telemetry/config_new/attributes.rs
bnjjj: We have a `redact` for the supergraph version

apollo-router/src/plugins/telemetry/config_new/attributes.rs
```diff
@@ -0,0 +1,702 @@
...
+    },
+    SupergraphResponseHeader {
+        /// The supergraph response header name.
+        supergraph_response_header: String,
+        /// Optional redaction pattern.
+        redact: Option<String>,
+        /// Optional default value.
+        default: Option<String>,
+    },
+    RequestContext {
```
bnjjj, line 360: at which stage ? supergraph? subgraph ?

apollo-router/src/plugins/telemetry/config_new/attributes.rs
```diff
@@ -0,0 +1,702 @@
...
+        /// The request context key.
+        request_context: String,
+        /// Optional redaction pattern.
+        redact: Option<String>,
+        /// Optional default value.
+        default: Option<AttributeValue>,
+    },
+    ResponseContext {
```
bnjjj, line 368: at which stage ? supergraph? subgraph ?

apollo-router/src/plugins/telemetry/config_new/events.rs
```diff
@@ -0,0 +1,66 @@
...
+
+#[allow(dead_code)]
+#[derive(Clone, Deserialize, JsonSchema, Debug, Default)]
+#[serde(deny_unknown_fields, default)]
+struct SupergraphEvents {}
+
+#[allow(dead_code)]
+#[derive(Clone, Deserialize, JsonSchema, Debug, Default)]
+#[serde(deny_unknown_fields, default)]
+struct SubgraphEvents {}
```
bnjjj, line 38: Do you have example of fields we might have here ?

apollo-router/src/plugins/telemetry/config_new/instruments.rs
```diff
@@ -0,0 +1,110 @@
...
+
+#[allow(dead_code)]
+#[derive(Clone, Deserialize, JsonSchema, Debug)]
+#[serde(deny_unknown_fields, rename_all = "snake_case")]
+pub(crate) enum InstrumentType {
+    /// A monotonic counter
+    Counter,
+
+    /// A duration histogram
+    Duration,
```
bnjjj, line 109: Why don't we keep `Histogram` instead of `Duration` ? Why not also adding `Gauge` ?

apollo-router/src/plugins/telemetry/config_new/logging.rs
```diff
@@ -0,0 +1,68 @@
...
+#[serde(deny_unknown_fields, rename_all = "snake_case")]
+enum Rollover {
+    /// Roll over every minute.
+    Minutely,
+    /// Roll over every hour.
+    Hourly,
+    /// Roll over every day.
+    Daily,
+    /// Never roll over.
+    #[default]
```
bnjjj, line 66: Should we really put `Never` as default ?

apollo-router/src/plugins/telemetry/config_new/logging.rs
```diff
@@ -0,0 +1,68 @@
...
+
+/// The format for logging.
+#[allow(dead_code)]
+#[derive(Deserialize, JsonSchema, Clone, Default, Debug)]
+#[serde(deny_unknown_fields, rename_all = "snake_case")]
+enum Format {
+    Aws,
+    Bunyan,
+    Gelf,
+    Google,
```
bnjjj, line 48: Is it really called the `google` format ?

apollo-router/src/plugins/telemetry/config.rs
```diff
@@ -56,12 +56,26 @@ pub(crate) struct Conf {
...
+    #[cfg(feature = "telemetry_next")]
+    /// Span configuration
+    pub(crate) spans: config_new::spans::Spans,
+    #[cfg(feature = "telemetry_next")]
+    /// Instrument configuration
+    pub(crate) instruments: config_new::instruments::Instruments,
```
bnjjj, line 78: It looks like it's only related to metrics, so same question why not putting that in `metrics` ?

bnjjj:
Do we agree that for custom attributes, if it's set to false, users will still see them on spans but they will be unset. Because the attributes list must be static, it's a constraint coming from the tracing crate

BrynCooke:
> Do we agree that for custom attributes, if it's set to false, users will still see them on spans but they will be unset. Because the attributes list must be static, it's a constraint coming from the tracing crate

We've got a workaround for that.
When the tracing crate is called it uses a fieldset, the issue is that it requires `&'static str` as keys. The way we're going to tackle this is to introduce some sort of intern functionality This will leak a small amount of memory memory, but will allow us to create dynamic tracing attributes.

apollo-router/src/plugins/telemetry/config_new/logging.rs
BrynCooke: The documentation is here: https://cloud.google.com/logging/docs/structured-logging
I'm not sure if it's officially called google. Let's at least add a rustdoc that details it.

apollo-router/src/plugins/telemetry/config.rs
BrynCooke: [9ac775a](https://github.com/apollographql/router/pull/4061/commits/9ac775a02e509a68dec99653ce010d109eab098e)

apollo-router/src/plugins/telemetry/config_new/logging.rs
BrynCooke: [9ac775a](https://github.com/apollographql/router/pull/4061/commits/9ac775a02e509a68dec99653ce010d109eab098e)

apollo-router/src/plugins/telemetry/config.rs
BrynCooke: Yes, although they only happen as part of a trace.

https://opentelemetry.io/docs/specs/otel/logs/event-api/

The way to think about it is that logs are the output, and events are the things that make it into traces and logs.

apollo-router/src/plugins/telemetry/config.rs
BrynCooke: The reason that they are separate is that span data does end up in logs. Events also make it into traces, so that's the reason why both are separate.

apollo-router/src/plugins/telemetry/config_new/attributes.rs
BrynCooke: [210cac9](https://github.com/apollographql/router/pull/4061/commits/210cac92202f7130b601728e934bab11c15fc74d)

apollo-router/src/plugins/telemetry/config_new/attributes.rs
BrynCooke: [a7d501c](https://github.com/apollographql/router/pull/4061/commits/a7d501cdb0ffad517b7420e328d01b8dcfd8f0b7)

apollo-router/src/plugins/telemetry/config_new/attributes.rs
BrynCooke: [a7d501c](https://github.com/apollographql/router/pull/4061/commits/a7d501cdb0ffad517b7420e328d01b8dcfd8f0b7)

apollo-router/src/plugins/telemetry/config.rs
BrynCooke: For consistency with spans and events this is at the same level. But let's discuss.

apollo-router/src/plugins/telemetry/config_new/logging.rs
BrynCooke: Open to something else, let's discuss.

apollo-router/src/plugins/telemetry/config_new/instruments.rs
BrynCooke: [d2c0b05](https://github.com/apollographql/router/pull/4061/commits/d2c0b05c1d34d8e3268e3c1d8efaa3cc82e02cb3)
Agree instrument type now is Counter and Histogram, I didn't add gauge though as I'm not sure how it'd work.
There is a separate field for Value, with contains Duration, Unit, but maybe we also add Active which would align with gague.

apollo-router/src/plugins/telemetry/config_new/attributes.rs
BrynCooke: Mirrors Subgraph structures

apollo-router/src/plugins/telemetry/config_new/attributes.rs
BrynCooke: Mirrors Subgraph structures

apollo-router/src/plugins/telemetry/config_new/instruments.rs
```diff
@@ -105,6 +103,25 @@ pub(crate) enum InstrumentType {
...
 
-    /// A duration histogram
+    /// A counter
+    UpDownCounter,
+
+    /// A histogram
+    Histogram,
+
+    /// A counter
+    Gague,
```
bnjjj, line 113: typo and bad doc

apollo-router/src/plugins/telemetry/config_new/events.rs
BrynCooke: [7c3414c](https://github.com/apollographql/router/pull/4061/commits/7c3414c45c8c4e49a6e26c607865e19ab1a227fe)

apollo-router/src/plugins/telemetry/config_new/instruments.rs
BrynCooke: [a120e6a](https://github.com/apollographql/router/pull/4061/commits/a120e6ae3b0e323f72eeab9db85092f55a15fd90)

apollo-router/src/plugins/telemetry/config_new/logging.rs
BrynCooke: [a95c1e4](https://github.com/apollographql/router/pull/4061/commits/a95c1e4261d31749bf58de7bda49890bc3b8a458) Set to daily

apollo-router/src/plugins/telemetry/config_new/events.rs
```diff
@@ -17,35 +17,56 @@ pub(crate) struct Events {
...
 #[serde(deny_unknown_fields, default)]
-struct SupergraphEvents {}
+struct SupergraphEvents {
+    /// Log the supergraph request
+    request: EventLevel,
+    /// Log the supergraph request
+    response: EventLevel,
+    /// Log the supergraph error
+    error: EventLevel,
+}
```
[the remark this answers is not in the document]
BrynCooke: I think if users want to to that then they can create a custom event.
The default events are just the basics.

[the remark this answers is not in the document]
BrynCooke: There's a typo in the comment doc.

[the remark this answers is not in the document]
BrynCooke: [3fc278f](https://github.com/apollographql/router/pull/4061/commits/3fc278f6c03a35bbb98b09afdd054db0fa9bd539)

lennyburdette (approved):
approving after discussing in person. the changes are worthwhile but will require a lot of documentation and probably a migration guide.

PR #4068, opened 2023-10-20
Update `reports.proto` to latest version

abernix:
Using:

    curl -f https://usage-reporting.api.apollographql.com/proto/reports.proto >
      apollo-router/src/plugins/telemetry/proto/reports.proto

### DIFF
--- a/apollo-router/tests/snapshots/apollo_reports__batch_send_header.snap
+++ b/apollo-router/tests/snapshots/apollo_reports__batch_send_header.snap
@@ -1088,6 +1088,7 @@ traces_per_query:
     stats_with_context: []
     referenced_fields_by_type: {}
     internal_traces_contributing_to_stats: []
+    query_metadata: ~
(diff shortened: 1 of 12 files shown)


apollo-router/tests/snapshots/apollo_reports__batch_send_header.snap
```diff
@@ -1088,6 +1088,7 @@ traces_per_query:
     stats_with_context: []
     referenced_fields_by_type: {}
     internal_traces_contributing_to_stats: []
+    query_metadata: ~
```
abernix, line 1091: This maps expectedly to the updates brought in by the `QueryMetadata` type in the `reports.proto` update itself. ✔️

BrynCooke (approved):

bnjjj (approved):

o0Ignition0o (approved):

PR #4076, opened 2023-10-23
add a flag to disable authorization error logs

Geal:
Authorization errors can be seen as common usage of the service when filtering fields from queries depending on the client's rights, so they might not warrant error logs to be analyzed by the router operators

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/plugins/authorization/mod.rs
+++ b/apollo-router/src/plugins/authorization/mod.rs
@@ -70,14 +70,26 @@ pub(crate) struct Conf {
     /// `@authenticated` and `@requiresScopes` directives
     #[serde(default)]
     preview_directives: Directives,
+    /// Log authorization errors
+    #[serde(default = "enable_log_errors")]
+    log_errors: bool,
 }
 
+fn enable_log_errors() -> bool {
+    true
+}
 #[derive(Clone, Debug, Default, Deserialize, JsonSchema)]
 #[allow(dead_code)]
 pub(crate) struct Directives {
     /// enables the `@authenticated` and `@requiresScopes` directives
     #[serde(default)]
     enabled: bool,
+}
+
+#[derive(Clone, Debug, Default, PartialEq, Eq, Serialize, Deserialize)]
+pub(crate) struct UnauthorizedPaths {
+    pub(crate) paths: Vec<Path>,
+    pub(crate) log_errors: bool,
 }
 
 pub(crate) struct AuthorizationPlugin {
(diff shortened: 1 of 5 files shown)


apollo-router/src/plugins/authorization/mod.rs
```diff
@@ -107,6 +120,17 @@ impl AuthorizationPlugin {
...
+    pub(crate) fn log_errors(configuration: &Configuration) -> bool {
+        let has_config = configuration
+            .apollo_plugins
+            .plugins
+            .iter()
+            .find(|(s, _)| s.as_str() == "authorization")
+            .and_then(|(_, v)| v.get("preview_directives").and_then(|v| v.as_object()))
+            .and_then(|v| v.get("log_errors").and_then(|v| v.as_bool()));
+        has_config.unwrap_or(true)
+    }
```
bnjjj, line 132: I know that looks useless but could you write a small test to check that function ? Basically it's to prevent breaking change when we will remove `preview_` prefix and if we decide to change the configuration shape later

.changesets/feat_geal_authz_log_errors.md
```diff
@@ -0,0 +1,13 @@
...
+
+Authorization errors can be seen as common usage of the service when filtering fields from queries depending on the client's rights, so they might not warrant error logs to be analyzed by the router operators
+
+THos logs can be disabled in the configuration:
+
+```yaml
+authorization:
+  preview_directives:
+    log_errors: true
+```
```
garypen, line 11: I don't have strong feelings about this, but ould this be better as:

```yaml
authorization:
  preview_directives:
    logging:
      enabled: true
```
?
We seem to be settling on this `enabled` `true|false` style and I wonder if it's just about enabling `logging` for directives or specifically `errors`.

also: We should specify if enabled or not by default here and in the docs.

o0Ignition0o: cc @BrynCooke as you might have a preference in light of the telemetry recent work

.changesets/feat_geal_authz_log_errors.md
```diff
@@ -0,0 +1,13 @@
+### add a flag to disable authorization error logs ([PR #4076](https://github.com/apollographql/router/pull/4076))
+
+Authorization errors can be seen as common usage of the service when filtering fields from queries depending on the client's rights, so they might not warrant error logs to be analyzed by the router operators
+
+THos logs can be disabled in the configuration:
```
BrynCooke, line 5:
```suggestion
Those logs can be disabled in the configuration:
```

bnjjj (approved):

.changesets/feat_geal_authz_log_errors.md
BrynCooke: It depends if we see further logging configuration in the future.
If we think there will be more config:

```yaml
authorization:
  preview_directives:
    log:
      errors: true
```

We can always tackle this with a migration though, so happy with for now:
```yaml
authorization:
  preview_directives:
    log_errors: true
```

Issue #3843, opened 2023-09-18
Dry run option for authorization directives


Geal:
This option would allow the router to run the authorization code without modifying the query: it would only generate errors at the path that would be removed. This is meant to test authorization without modifying traffic

PR #4079, opened 2023-10-23
Authorization: add a dry run option

Geal:
Fix https://github.com/apollographql/router/issues/3843

This adds an option to execute authorization without modifying the
query: it will generate the list of unauthorized paths and send it in
the response, allowing the user to understand which parts of the query
would be affected

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/plugins/authorization/scopes.rs
```diff
@@ -451,7 +464,7 @@ impl<'a> transform::Visitor for ScopeFilteringVisitor<'a> {
         // If we modified the transform visitor implementation to modify the fragment definitions before the
         // operations, we would be able to store the list of unauthorized paths per fragment, and at the point
         // of application, generate unauthorized paths starting at the operation root
-        if !fragment_is_authorized {
+        if !fragment_is_authorized && !self.dry_run {
```
BrynCooke, line 467: Nit: the nots here make my head spin.

bnjjj:
I'm wondering if instead of having a global dry_run we could not set in the configuration a header to specifiy we want to run that request in dry_run mode. You would be able to still run the same instance in production and run a simple request in dry_run mode. What do you think ?

Geal:
the header could be done as a follow-up, but for now I don't want to add an option where the client can decide to disable authorization. We'll see later if there's demand for that header

apollo-router/src/plugins/authorization/scopes.rs
Geal: I'll switch the branches

Geal: does it look more reasonable now?

BrynCooke (approved):

apollo-router/src/plugins/authorization/scopes.rs
BrynCooke: Much easier to read, thanks!

.changesets/feat_geal_authz_dry_run.md
```diff
@@ -0,0 +1,5 @@
+### Authorization: dry run option ([Issue #3843](https://github.com/apollographql/router/issues/3843))
+
+It is now possible to execute authorization directives without modifying the query, but still returning the list of affected paths in the response. This allows testing authorization without breaking existing traffic
```
lennyburdette, line 3:
```suggestion
It is now possible to execute authorization directives without modifying the query, but still return the list of affected paths as top-level errors in the response. This allows testing authorization without breaking existing traffic.
```

docs/source/configuration/authorization.mdx
```diff
@@ -644,7 +644,19 @@ This option will entirely reject queries that would have some parts filtered by
 ```yaml title="router.yaml"
 authorization:
   preview_directives:
+    enabled: true
     log_errors: true
 ```
 
-By default, when part of a query is filtered by authorization, the list of filtered paths is added to the response and logged by the router. Filtering parts of a query according to the client's rights may be seen as normal operation, and not warrant investigation by platform operators, so those logs can be disabled.
\ No newline at end of file
+By default, when part of a query is filtered by authorization, the list of filtered paths is added to the response and logged by the router. Filtering parts of a query according to the client's rights may be seen as normal operation, and not warrant investigation by platform operators, so those logs can be disabled.
```
lennyburdette, line 651:
```suggestion
By default, when part of a query is filtered by authorization, the list of filtered paths is added to the response as top-level errors and logged by the router. Filtering parts of a query according to the client's rights may be seen as normal operation, and not warrant investigation by platform operators, so those logs can be disabled.
```

My first question when reading this was "are the filtered paths added as errors or extensions"?

docs/source/configuration/authorization.mdx
```diff
@@ -644,7 +644,19 @@ This option will entirely reject queries that would have some parts filtered by
...
+### dry_run
+
+```yaml title="router.yaml"
+authorization:
+  preview_directives:
+    enabled: true
+  dry_run: true # default: false
+```
+
+This option executes the authorization directives without modifying the query: it only generates the list of unauthorized paths and returns it as part of the response. This is a way to evaluate the impact of authorization policies without breaking existing traffic.
```
lennyburdette, line 662: Have you validated that returning top-level errors is what customers will expect with a `dry_run` option? Apollo Client defaults to raising exceptions when it sees top-level errors, so most client apps will fail.

When I see "dry_run" I assume that there will be no runtime effects on my applications. Maybe it should attach the filtered paths as extensions instead?

Geal: the opinion from the client devs was that authorization errors could be easily ignored so it wouldn't be an issue. Even in normal operation, you can have a valid query for which there are authorization errors, that does not warrant raising the error up to the user

Geal: will follow up in https://github.com/apollographql/router/issues/4116

Issue #4030, opened 2023-10-12
Traffic routing issue in a federated GraphQL setup with multiple subgraphs.


purushottam-anand:
**Describe the bug**
*Traffic routing issue in a federated GraphQL setup with multiple subgraphs.*

**To Reproduce**
*Steps to reproduce the behavior:*
1. Set up a federated GraphQL system with router and subgraphs.
2. Implement blue-green deployment for the subgraphs.
3. Shift traffic from blue stack to green stack (or vice-versa) by adjusting Route53 weights.
4. Observe the routing behavior over time.

**Expected behavior**
*We expect that after a minute, all traffic from the router should be routed to the new stack (e.g., green) due to the 60-second TTL on Route53.*

**Output**
*However, we have observed that the router continues to send requests to the old stack (e.g., blue) for a prolonged period, roughly 40-50 minutes.*

**Additional context**
*This issue raises questions about whether the router caches the subgraph's load balancer DNS or IP. If it does cache, we need to understand how to flush that cache within a minute to ensure timely traffic redirection. Is there a way to configure this behavior to align with the expected routing changes?*

BrynCooke:
Suspect that this is connections to subgraphs are remaining open, we may need to make sure that we respect the DNS TTL.

garypen:
I setup the following:

- a private network containing 4 systems
  - a DNS server (TTL 60)
  - router federating over 2 subgraps
  - subgraph 1
  - subgraph 2
 
I then queried the router and found that my queries resolved as expected.
I then modified my DNS configuration so that the entry for subgraph 2 pointed to a non-existing host
I then re-ran my query and found that it immediately failed because subgraph2 could not be accessed
I then restored my valid DNS configuration and re-ran the query and found that it worked fine.

I realise that this isn't the "exact" issue that is being reported by the user, but I think it's a reasonable enough approximation to show that the router is using DNS and is respecting TTL as reported by the DNS server.

@purushottam-anand It seems more likely, at this point, that there is a networking issue in your testing infrastructure. If you take the router out of the equation can you validate/confirm that your DNS server is behaving as you expect?

I'm going to mark this as blocked for now until we get concrete evidence that this is a router issue.

PR #4088, opened 2023-10-23
Use trust dns for hyper client resolver

garypen:
Investigating memory revealed that the default hyper client DNS resolver had a negative impact on the memory footprint of the router.

It may also not be respecting TTL correctly. Let's replace the default with Trust DNS.

fixes: #4030

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/Cargo.toml
```diff
@@ -97,6 +97,7 @@ humantime = "2.1.0"
 humantime-serde = "1.1.1"
 hyper = { version = "0.14.27", features = ["server", "client"] }
 hyper-rustls = { version = "0.24.1", features = ["http1", "http2"] }
+hyper-trust-dns-connector = "0.4.0"
```
Geal, line 100: that crate's last release is over a year old now and it brings in old versions of trust-dns. Considering the amount of code this represent, we could have an in house implementation: https://github.com/paullgdc/hyper-trust-dns-connector/blob/master/src/lib.rs

garypen: Yep. I did have those thoughts. Ok. I'll do an in-house one since we think it's useful to have.

garypen: done: ef2dfcdfa9c31b9b9d07785e330be7c391ed61d0

apollo-router/src/services/trust_dns_connector.rs
```diff
@@ -0,0 +1,64 @@
...
+    type Error = io::Error;
+
+    fn poll_ready(&mut self, _cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        Poll::Ready(Ok(()))
+    }
+
+    fn call(&mut self, name: Name) -> Self::Future {
+        let resolver = self.0.clone();
+
+        let closure = || async move {
```
Geal, line 44: why a closure here?

garypen: Just porting the code that was already there.

I cleaned it up: fc57d3230bba7ca9f804a80a88307e9c698847d6

PR #4094, opened 2023-10-24
Remove the most significant memory leaks caused by HashMap::into_iter()

garypen:
Cloning the full metrics entries results in leaks.

The root cause is a memory leak in the rust standard HashMap (see changelog for more details).

This PR mitigates the impact of the leak on the router by replacing our use of `into_iter()` with `iter()` in the area of the router greatest impacted by this problem.

When rust 1.74.0 is available this problem will be fixed since the issue was fixed 2 days ago.

**Checklist**

**Exceptions**

*Note any exceptions here*

**Notes**


garypen:
Closing this. Further investigation reveals that this may be a small part of the problem. I'll open a new PR soon which addresses the issues more completely.

PR #4097, opened 2023-10-25
Add DR to extend plugin init

BrynCooke:
Adds a DR for extending plugin init:
Target ratification date: 2023-30-10


Geal (commented):
could we make the entire configuration available to the plugin?

We would still deserialize the plugin specific config and give it in PluginInit, but I think there were some use cases for some plugins accessing the rest of the configuration too

bnjjj (approved):

BrynCooke:
> could we make the entire configuration available to the plugin? We would still deserialize the plugin specific config and give it in PluginInit, but I think there were some use cases for some plugins accessing the rest of the configuration too

Yes let's do this

garypen (approved):

PR #4098, opened 2023-10-25
Add DR for unstable plugin API

BrynCooke:
Adds a decision record for adding an unstable plugin API.
Aim to ratify by 2023-10-30

### DIFF
--- a/dev-docs/decision-records/unstable-plugins.md
+++ b/dev-docs/decision-records/unstable-plugins.md
@@ -12,7 +12,7 @@ We will promote hook points from private to unstable to public as we gain confid
 
 ## Hook points that could be introduced
 
-1. Query planning - Allows plugins tom implement caching, or other optimizations.
+1. Query planning - Allows plugins to implement caching, or other optimizations.
 2. Validation - Allows plugins to place additional constraints on the query before hitting the query planner.
 3. Parse - Allows plugins to modify the query before it is parsed. e.g. APQ or persisted queries.
 4. Http subgraph request - Allows signing of subgraph requests to take place.
--- a/dev-docs/decision-records/unstable-plugins.md
+++ b/dev-docs/decision-records/unstable-plugins.md
@@ -1,16 +1,17 @@
-# Unstable Plugin API
+# Title [ADR-4]
+
+Provide a mechanism which supports stable evolution of the Plugin APIs
+
+## Status
+
+Approved
+
+## Context
 
 The Plugin API is currently stable. However, we have noticed that there are some areas that could be improved.
-To prevent the risk that we introduce a new API that does not fully meet our needs and our users needs we need to introduce an API for unstable and private hook points.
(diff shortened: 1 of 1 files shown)


Geal (approved):

o0Ignition0o (approved):
This looks great!

Are there implications re Rhai / Coprocessors when these APIs graduate to unstable / public?

BrynCooke:
> This looks great!
> 
> Are there implications re Rhai / Coprocessors when these APIs graduate to unstable / public?

Let's add something to the DR that explains graduation criteria.

Gut feeling is that to move from unstable to public we have to make sure it integrates with everything, but beyond that we can accept piecemeal integration.

dev-docs/decision-records/unstable-plugins.md
```diff
@@ -0,0 +1,36 @@
...
+* pub (crate) PluginPrivate
+
+Demo here: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b4022dfeaeb920bc7ae94a94b7b9ea23
+
+This doesn't give us permission to not think about APIs we introduce, but it does enable us to unblock ourselves and our users in a safe way.
+
+## Acceptance criteria
+* Existing plugins should compile without modification.
+* Plugins external to the Router codebase cannot use the private API.
+* Plugins external to the router codebase can use the unstable API.
```
bnjjj, line 35: Should we gate it under a feature called `router-unstable` like what we can find in `tracing` or `tokio` ?

dev-docs/decision-records/unstable-plugins.md
```diff
@@ -0,0 +1,36 @@
...
+The order of introduction of new hook points would be:
+* Private - Does this work and make the router code cleaner?
+* Unstable - Available for use by plugins, but not guaranteed to be stable.
+* Public - Usable by all, and won't change.
+
+We will promote hook points from private to unstable to public as we gain confidence in them.
+
+## Hook points that could be introduced
+
+1. Query planning - Allows plugins tom implement caching, or other optimizations.
```
garypen, line 15:
```suggestion
1. Query planning - Allows plugins to implement caching, or other optimizations.
```

garypen (changes requested):

garypen:
> This looks great!
> 
> Are there implications re Rhai / Coprocessors when these APIs graduate to unstable / public?

I think there would be. We should perhaps add to the DR that before graduating (to unstable? or to public?) from private we should have considered and decided what details (if any) would be accessible from coprocessors and Rhai. I think we'd also need to have good reasons to not expose such details via these mechanisms.

dev-docs/decision-records/unstable-plugins.md
BrynCooke: Don't think we'll feature flag it as much of the benefit is that we get feedback from not only users but ourselves. Our plugins would not be able to use the unstable APIs if we feature gate it.

BrynCooke:
I added some discussion for graduation that can be discussed.

garypen (approved):

bnjjj (approved):

PR #4102, opened 2023-10-26
Add support for custom attributes for spans and logs

### LINKED ISSUE #4096: Add `DynAttribute` functionality to spans and logs
We need to be able to create spans and logs with custom attributes defined in yaml.

Acceptance criteria:

* We can create a span with dynamic attributes that are output to tracing and logs.
* We can create a log message with custom attributes that are output to tracing and logs.

bnjjj:
Fixes #4096

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/plugins/telemetry/dynamic_attribute.rs
```diff
@@ -0,0 +1,159 @@
...
+                        }
+                    }
+                };
+            } else {
+                eprintln!("no Registry, this is a bug");
+            }
+        });
+    }
+
+    fn set_dyn_attributes(&self, attributes: HashMap<String, String>) {
```
Geal, line 137: maybe taking a `Iterator<Item = (String, String)>` instead of a `HashMap`

apollo-router/src/plugins/telemetry/dynamic_attribute.rs
```diff
@@ -0,0 +1,159 @@
...
+use super::ROUTER_SPAN_NAME;
+use super::SUBGRAPH_SPAN_NAME;
+use super::SUPERGRAPH_SPAN_NAME;
+
+#[derive(Debug)]
+pub(crate) enum LogAttributes {
+    Router(HashMap<String, String>),
+    Supergraph(HashMap<String, String>),
+    Subgraph(HashMap<String, String>),
+}
```
Geal, line 17: if all variants have the same content, it may be nicer with a flag:

```suggestion
pub(crate) enum SpanKinf {
  Router,
  Supergraph,
  Subgraph
}
pub(crate) struct LogAttributes {
  kind: SpanKind,
  attributes: HashMap<String, String>,
}
```

apollo-router/src/plugins/telemetry/dynamic_attribute.rs
```diff
@@ -0,0 +1,159 @@
...
+
+impl DynAttributeLayer {
+    pub(crate) fn new() -> Self {
+        Self {}
+    }
+}
+
+pub(crate) trait DynAttribute {
+    fn set_dyn_attribute(&self, key: String, value: String);
```
Geal, line 109: nit: it's likely that keys will stay the same for the entire time config and schema will be used, so you could avoid allocations by using `Arc<String>`

apollo-router/src/plugins/telemetry/formatters/text.rs
```diff
@@ -153,6 +154,50 @@ impl TextFormatter {
...
+                    //     let style = Style::new().dimmed();
+                    //     write!(writer, "{}", style.prefix())?;
+                    //     write!(writer, "[trace_id={trace_id}]")?;
+                    //     write!(writer, "{}", style.suffix())?;
+                    // } else {
+                    let attrs: Vec<String> = attributes
+                        .iter()
+                        .map(|(key, val)| format!("{key}={val}"))
+                        .collect();
+                    write!(writer, "[attributes=[{}]]", attrs.join(", "))?;
```
Geal, line 188: this would avoid allocating a temporary string:
```suggestion
                    write!(writer, "[attributes=[")?;
                    let mut it = attributes.iter();
                    if let Some((key, value)) = it.next() {
                      write!(writer, "{key}={value}")?;
                    }
                    for (key, value) in it {
                      write!(writer, ", {key}={value}")?;
                    }
                    write!(writer, "]]")?;
```

Geal (approved):

apollo-router/src/plugins/telemetry/dynamic_attribute.rs
bnjjj: Yes, this part of the API has to change, especially I would like to not require `String` but something more dynamic when we can avoid allocations

apollo-router/src/plugins/telemetry/dynamic_attribute.rs
bnjjj: Yes related to the same comment [here](https://github.com/apollographql/router/pull/4102#discussion_r1373298243)

apollo-router/src/plugins/telemetry/formatters/text.rs
bnjjj: I know, this part is a test and I think I'll delete that part and only support it in json formatter. But if not I'll take your suggestion. Thanks

apollo-router/src/plugins/telemetry/config_new/attributes.rs
```diff
@@ -792,6 +832,600 @@ pub(crate) struct HttpClientAttributes {
...
+                operation_name,
+                default,
+                ..
+            } => {
+                let op_name = request.context.get(OPERATION_NAME).ok().flatten();
+                match operation_name {
+                    OperationName::String => {
+                        op_name.or_else(|| default.clone().map(AttributeValue::String))
+                    }
+                    OperationName::Hash => todo!(),
```
Geal, line 1125: what's the plan here?

apollo-router/src/plugins/telemetry/formatters/json.rs
```diff
@@ -3,18 +3,277 @@ use std::fmt;
...
+    fn format_event(
+        &self,
+        ctx: &FmtContext<'_, S, N>,
+        mut writer: Writer<'_>,
+        event: &Event<'_>,
+    ) -> fmt::Result
+    where
+        S: Subscriber + for<'a> LookupSpan<'a>,
+    {
+        let timestamp = chrono::Utc::now().to_rfc3339_opts(SecondsFormat::Millis, true);
```
Geal, line 200: maybe let's use https://docs.rs/time/latest/time/serde/rfc3339/index.html instead of adding the entire chrono crate

apollo-router/src/services/subgraph.rs
```diff
@@ -39,6 +39,8 @@ pub struct Request {
 
     pub context: Context,
 
+    /// Name of the subgraph, it's an Option to not introduce breaking change
+    pub(crate) subgraph_name: Option<String>,
```
garypen, line 43: This struct is `non-exhaustive` so does this need to be an `Option`?

apollo-router/src/plugins/telemetry/config_new/attributes.rs
bnjjj: Good question ! Haha, I forgot about that part, @BrynCooke Do you know what kind of hash we should provide ?

apollo-router/src/plugins/telemetry/config.rs
```diff
@@ -453,14 +452,30 @@ pub(crate) enum AttributeValue {
     Bool(bool),
     /// i64 values
     I64(i64),
+    /// i64 values
+    U128(u128),
     /// f64 values
     F64(f64),
     /// String values
     String(String),
+    // TODO add static string type support
```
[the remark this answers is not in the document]
bnjjj: Not at all, that was because I thought we would have a lot of static attribute values but at the end not that much. So I think I'll delete that comment because for now we won't benefit from having that static support

apollo-router/src/plugins/telemetry/config_new/spans.rs
```diff
@@ -61,7 +90,8 @@ pub(crate) struct RouterAttributes {
     #[serde(flatten)]
     common: HttpCommonAttributes,
     /// Span http server attributes from Open Telemetry semantic conventions.
-    #[serde(flatten)]
+    // TODO unskip it
```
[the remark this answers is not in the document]
bnjjj: Yes It's basically part of the big umbrella issue for this telemetry revamp

apollo-router/src/plugins/telemetry/formatters/text.rs
```diff
@@ -153,6 +153,47 @@ impl TextFormatter {
         writer.write_char(' ')
     }
 
+    // #[inline]
+    // fn format_dyn_attributes<S, N>(
+    //     &self,
+    //     ctx: &FmtContext<'_, S, N>,
```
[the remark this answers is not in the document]
bnjjj: I prefer to keep it because once we tackle the new logging config we will need this I think

apollo-router/src/services/layers/query_analysis.rs
```diff
@@ -105,13 +107,16 @@ impl QueryAnalysisLayer {
 
                 let context = Context::new();
 
-                let operation_name = doc
-                    .executable
-                    .get_operation(op_name.as_deref())
-                    .ok()
+                let operation = doc.executable.get_operation(op_name.as_deref()).ok();
+                let operation_name = operation
```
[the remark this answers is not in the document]
bnjjj: Why ?

apollo-router/src/services/subgraph.rs
bnjjj: Hmm not for the struct itself but for the builder. What am I supposed to put in the builder ? If I'm not using an option it will break all the existing builders because it will require the subgraph_name. And so if it's not an option, I will still have to use an Option in the builder

apollo-router/src/services/layers/query_analysis.rs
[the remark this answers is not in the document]
garypen: My mistake, I missed you were using `operation` at line 116.

apollo-router/src/services/subgraph.rs
garypen: Does that mean `buildstructor` means there's no point marking a struct as `non-exhaustive`?

bnjjj: I think so... Unless I'm missing something ?! cc @BrynCooke

BrynCooke: We could default the name in the builder if it is not supplied. That way in the struct is can be mandatory.

apollo-router/src/plugins/telemetry/config.rs
```diff
@@ -514,13 +531,26 @@ impl From<AttributeValue> for opentelemetry::Value {
         match value {
             AttributeValue::Bool(v) => Value::Bool(v),
             AttributeValue::I64(v) => Value::I64(v),
+            AttributeValue::U128(v) => Value::I64(v as i64),
```
o0Ignition0o, line 534: This is odd 🤔

apollo-router/src/services/subgraph.rs
garypen: That seems nicer.
Not this PR: but does it mean we should put comments on any other structs which are marked `non-exhaustive` and use `buildstructor` so future us know how to handle this situation? We could raise an issue for that.

apollo-router/src/plugins/telemetry/formatters/json.rs
```diff
@@ -122,16 +380,58 @@ impl<'a> tracing_subscriber::field::VisitOutput<fmt::Result> for JsonVisitor<'a>
...
+
     /// Visit a string value.
     fn record_str(&mut self, field: &Field, value: &str) {
-        if field.name() == TRACE_ID_FIELD_NAME {
-            self.values
-                .insert(field.name(), serde_json::Value::from(value));
+        let field_name = field.name();
+        match field_name {
+            "code.filepath" | "code.namespace" | "code.lineno" | "thread.id" | "thread.name" => {
+                println!("!!!!!!!!!!!!!!!!!!!!!");
```
o0Ignition0o, line 412: ?!

apollo-router/src/plugins/telemetry/mod.rs
```diff
@@ -298,9 +308,8 @@ impl Plugin for Telemetry {
                     .and_then(|h| h.to_str().ok())
                     .unwrap_or("");
                 let span = ::tracing::info_span!(ROUTER_SPAN_NAME,
-                    "http.method" = %router_request.method(),
                     "http.route" = %router_request.uri(),
-                    "http.flavor" = ?router_request.version(),
```
o0Ignition0o, line 303: oh do we remote http 1.1 now ?

o0Ignition0o (approved):

apollo-router/src/plugins/telemetry/formatters/json.rs
bnjjj: Oops, fixed

apollo-router/src/plugins/telemetry/config.rs
bnjjj: Ah yes, otel Value doesn't support u128 unfortunately

apollo-router/src/plugins/telemetry/mod.rs
bnjjj: I removed that part because it's now part of the customized attributes we can add on spans, and it's called `network.protocol.version`

apollo-router/src/plugins/telemetry/mod.rs
```diff
@@ -256,116 +267,166 @@ impl Plugin for Telemetry {
...
-                    );
+                    Self::plugin_metrics(&config);
 
+                    async move {
+                        let span = Span::current();
+                        span.set_dyn_attributes(custom_attributes);
+                        let trace_id = TraceId::maybe_new()
+                            .map(|t| t.to_string())
+                            .unwrap_or_default();
+                        span.record("trace_id", trace_id);
```
bnjjj, line 366: I think we can delete that part

apollo-router/src/plugins/telemetry/mod.rs
```diff
@@ -256,116 +267,166 @@ impl Plugin for Telemetry {
...
-                            span.record("otel.status_code", "Error");
-                        } else {
-                            span.record("otel.status_code", "Ok");
+                            if response.response.status() >= StatusCode::BAD_REQUEST {
+                                span.record(OTEL_STATUS_CODE, "Error");
+                            } else {
+                                span.record(OTEL_STATUS_CODE, "Ok");
+                            }
+                        } else if let Err(err) = &response {
+                            span.set_dyn_attributes(config.spans.router.attributes.on_error(err));
```
bnjjj, line 406: Should we also add `span.record(OTEL_STATUS_CODE, "Error");` here ?

apollo-router/src/plugins/telemetry/mod.rs
```diff
@@ -546,6 +627,20 @@ impl Plugin for Telemetry {
...
+impl Callsite for LateCallsite {
+    fn set_interest(&self, interest: Interest) {
+        self.callsite.get().unwrap().set_interest(interest);
+    }
+
+    fn metadata(&self) -> &Metadata<'_> {
+        self.callsite.get().unwrap().metadata()
+    }
+}
+
```
bnjjj, line 643: I think we can delete that part

apollo-router/src/plugins/telemetry/reload.rs
```diff
@@ -207,7 +210,8 @@ where
 
         // we only make the sampling decision on the root span. If we reach here for any other span,
         // it means that the parent span was not enabled, so we should not enable this span either
-        if meta.name() != REQUEST_SPAN_NAME {
+        // TODO take the conf option to check the root span name
```
bnjjj, line 213: We can delete that comment

apollo-router/src/plugins/telemetry/mod.rs
BrynCooke: Yes

apollo-router/src/plugins/telemetry/config.rs
```diff
@@ -67,14 +66,12 @@ pub(crate) struct Conf {
...
-    #[cfg(feature = "telemetry_next")]
+    #[serde(skip)]
     /// Event configuration
     pub(crate) events: config_new::events::Events,
-    #[cfg(feature = "telemetry_next")]
     /// Span configuration
     pub(crate) spans: config_new::spans::Spans,
-    #[cfg(feature = "telemetry_next")]
+    #[serde(skip)]
     /// Instrument configuration
```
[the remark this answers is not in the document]
BrynCooke: I could go either way.
Having it as skip means that we don't have to deal with stuff falling out of sync when people change the code, but if you feel strongly it should be a feature gate then we can.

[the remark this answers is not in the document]
bnjjj: Hmm no that's a good point, let's keep it ungated. Do we still need that feature gate in cargo.toml then ?

apollo-router/src/plugins/telemetry/fmt_layer.rs
```diff
@@ -0,0 +1,54 @@
...
+pub(crate) fn create_fmt_layer(
+    config: &config::Conf,
+) -> Box<dyn Layer<LayeredTracer> + Send + Sync> {
+    let config = &config.new_logging.stdout;
+    match &config.format {
+        Format::Json(config) => {
+            let format = Json::default()
+                .with_level(config.display_level)
+                .with_target(config.display_target)
+                .with_level(config.display_level)
```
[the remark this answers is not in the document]
BrynCooke: I don't think our formatter supports flatten event. Let's remove for now.

apollo-router/src/plugins/telemetry/mod.rs
BrynCooke: Hmm, not sure. For legacy mode the `trace_id` is still needed?

BrynCooke: Unless you added the functionality elsewhere?

bnjjj: I deleted it because the trace_id is already given at the creation of the deprecated span in span_factory

apollo-router/src/plugins/telemetry/mod.rs
BrynCooke: Yes

apollo-router/src/plugins/telemetry/config.rs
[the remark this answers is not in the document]
BrynCooke: Yeah, let's remove it.

PR #4114, opened 2023-10-30
New telemetry config tweaks

BrynCooke:
Fix telemetry attribute schema generation
Add a couple of missing attributes.

The new config is skipped right now, but be tested by existing config tests once the config is enabled.

Existing tests will check that the router yaml examples are valid with respect to schema. As the new documentation covers the news schema, it will be tested.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/plugins/telemetry/config_new/attributes.rs
+++ b/apollo-router/src/plugins/telemetry/config_new/attributes.rs
@@ -1,3 +1,4 @@
+use std::any::type_name;
 use std::collections::HashMap;
 
 use schemars::gen::SchemaGenerator;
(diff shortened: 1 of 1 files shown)


apollo-router/src/plugins/telemetry/config_new/attributes.rs
```diff
@@ -37,7 +38,11 @@ where
     E: JsonSchema,
 {
     fn schema_name() -> String {
-        "extendable_attribute".to_string()
+        format!(
```
BrynCooke, line 41: The attributes were conflicting in the json schema, so make sure they have unique names.

bnjjj (approved):

PR #4119, opened 2023-10-30
Lb/record replay

lennyburdette:
Request recording plugin and replay test runner

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/plugins/record_replay/record.rs
```diff
@@ -0,0 +1,350 @@
...
+/// Request recording configuration.
+#[derive(Debug, Clone, serde::Deserialize, schemars::JsonSchema)]
+#[serde(deny_unknown_fields)]
+struct RecordConfig {
+    /// The recording plugin is disabled by default.
+    enabled: bool,
+    /// The path to the directory where recordings will be stored. Defaults to
+    /// the current working directory.
+    #[serde(default = "default_storage_path")]
+    storage_path: String,
```
Geal, line 47: if this is made an option, then the snapshot will not have different results per machine

apollo-router/src/plugins/record_replay/record.rs
```diff
@@ -0,0 +1,350 @@
...
+                                  .operation_name
+                                  .clone()
+                                  .unwrap_or("UnnamedOperation".to_string());
+
+                              let unix_time = std::time::SystemTime::now()
+                                  .duration_since(std::time::UNIX_EPOCH)?
+                                  .as_secs();
+
+                              let filename =
+                                  format!("{}-{}.json", operation_name, unix_time).to_string();
```
Geal, line 145: it might be better to generate files where names can't collide, just to be safe. A lot of queries can happen in a second. Maybe some uuid v7?

apollo-router/src/plugins/record_replay/record.rs
```diff
@@ -0,0 +1,350 @@
...
+                let supergraph_sdl = supergraph_sdl.clone();
+
+                async move {
+                    let res: router::Response = future.await?;
+                    let (parts, stream) = res.response.into_parts();
+
+                    let headers = parts.headers.clone();
+                    let context = res.context.clone();
+
+                    let stream = stream.after_complete(move || {
```
Geal, line 115: the recording task should happen in a `spawn_blocking`, since serializing and writing are synchronous and could block

apollo-router/src/plugins/record_replay/after_complete.rs
```diff
@@ -0,0 +1,70 @@
...
+        })
+    }
+}
+
+impl<T: ?Sized> AfterCompleteStreamExt for T where T: Stream {}
+
+pub(crate) trait AfterCompleteStreamExt: Stream {
+    /// Executes the given async closure after the stream completes. The stream
+    /// will not yield Poll::Ready(None) until the closure completes.
+    fn after_complete<F, Fut>(self, callback: F) -> AfterComplete<Self, F, Fut>
```
Geal, line 62: that helper looks useful!

apollo-router/src/plugins/record_replay/record.rs
BrynCooke, line 115: Possibly we could get away with using the existing stream combinators.

Something like:

`mystream.chain(once(async{ do something; None}).filter_map(|a|a)`

`chain` tacks a new stream on the end of the first stream.
The second stream gets polled once and but returns None which gets filtered out by `filter_map`.

lennyburdette: thank you! i _knew_ there must be some way to do this with existing combinators but this didn't click with me

apollo-router/src/spec/query.rs
```diff
@@ -279,7 +279,7 @@ impl Query {
         Arc::new(ParsedDocumentInner { ast, executable })
     }
 
-    #[cfg(test)]
+    // #[cfg(test)]
```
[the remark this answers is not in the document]
lennyburdette: i wanted to write
```rs
let query = Query::parse(query, &schema, &Configuration::default()).expect("query must valid");
query.contains_introspection()
```
but `::parse` was accessible. is there another way?

apollo-router/Cargo.toml
```diff
@@ -265,10 +267,10 @@ tikv-jemallocator = "0.5"
 
 [dev-dependencies]
 axum = { version = "0.6.20", features = [
```
garypen, line 269: Not this PR, but I keep wondering why we duplicated entries in `dependencies` into `dev-dependencies`. We shouldn't unless the features set is different.

apollo-router/src/plugins/record_replay/record.rs
```diff
@@ -0,0 +1,356 @@
...
+        .unwrap_or(None)
+        .unwrap_or(false)
+}
+
+fn is_introspection(query: String, schema: Arc<Schema>) -> bool {
+    let query = Query::parse(query, &schema, &Configuration::default()).expect("query must valid");
+    query.contains_introspection()
+}
+
+fn externalize_header_map(
```
garypen, line 346: This looks like a copy of the co-processor `externalize_header_map()`. Maybe better to move the original `fn` into the src/services/external.rs source and have both plugins use that function?

apollo-router/src/plugins/record_replay/runner.rs
```diff
@@ -0,0 +1,46 @@
+#[cfg(test)]
+mod tests {
+    use console::style;
```
garypen, line 3: This seems like it would be better named as `replay_tests.rs` and it should be a direct submodule or `replay.rs`. Also, do something like this to restructure things a little:
```
Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
        renamed:    apollo-router/src/plugins/record_replay/runner.rs -> apollo-router/src/plugins/record_replay/replay_tests.rs

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   apollo-router/src/plugins/record_replay/mod.rs
        modified:   apollo-router/src/plugins/record_replay/replay.rs
        modified:   apollo-router/src/plugins/record_replay/replay_tests.rs
```

The code changes would look like:
```
garypen@Garys-MacBook-Pro router % git diff -b
diff --git a/apollo-router/src/plugins/record_replay/mod.rs b/apollo-router/src/plugins/record_replay/mod.rs
index 893ec15fe..b3f42fffa 100644
--- a/apollo-router/src/plugins/record_replay/mod.rs
+++ b/apollo-router/src/plugins/record_replay/mod.rs
@@ -1,4 +1,3 @@
 mod record;
 mod recording;
 mod replay;
-mod runner;
diff --git a/apollo-router/src/plugins/record_replay/replay.rs b/apollo-router/src/plugins/record_replay/replay.rs
index d3234d21e..d4f7e39a8 100644
--- a/apollo-router/src/plugins/record_replay/replay.rs
+++ b/apollo-router/src/plugins/record_replay/replay.rs
@@ -412,3 +412,7 @@ fn render_invisible(s: &str, newlines_matter: bool) -> Cow<'_, str> {
         Cow::Borrowed(s)
     }
 }
+
+#[cfg(test)]
+#[path = "replay_tests.rs"]
+mod tests;
diff --git a/apollo-router/src/plugins/record_replay/replay_tests.rs b/apollo-router/src/plugins/record_replay/replay_tests.rs
index efc13011e..6b79aa379 100644
--- a/apollo-router/src/plugins/record_replay/replay_tests.rs
+++ b/apollo-router/src/plugins/record_replay/replay_tests.rs
@@ -1,13 +1,11 @@
-#[cfg(test)]
-mod tests {
-    use console::style;
-    use tower::ServiceExt;
+use console::style;
+use tower::ServiceExt;
 
-    use super::super::replay::Replay;
-    use crate::TestHarness;
+use super::Replay;
+use crate::TestHarness;
 
-    #[tokio::test]
-    async fn replay_recording() {
+#[tokio::test]
+async fn replay_recording() {
     let recording_file = if let Ok(file) = std::env::var("RECORDING_FILE") {
         file
     } else {
@@ -42,5 +40,4 @@ mod tests {
             println!();
         }
     }
-    }
 }
```

garypen (changes requested):
This looks nice, I look forward to having this. I've made some suggestions that are just about structuring code.

apollo-router/src/plugins/record_replay/runner.rs
lennyburdette: i like it!

apollo-router/src/plugins/record_replay/recording-readme.md
```diff
@@ -0,0 +1,44 @@
+# Apollo Router Request Recordings
```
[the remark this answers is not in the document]
lennyburdette: My intention wasn't for this to be docs, but a file that's included in the output of the recordings. I can make a separate PR with actual docs for this plugin. (I also changed the content so it's a little more clear what it's for.)

apollo-router/src/plugins/record_replay/record.rs
```diff
@@ -0,0 +1,342 @@
...
+use crate::spec::query::Query;
+use crate::spec::Schema;
+use crate::Configuration;
+
+const RECORD_HEADER: &str = "x-apollo-router-record";
+const RECORD: &str = "record";
+const CLIENT_REQUEST: &str = "client_request";
+const CLIENT_RESPONSE: &str = "client_response";
+const QUERY_PLAN: &str = "query_plan";
+const SUBGRAPHS: &str = "subgraphs";
```
bnjjj, line 35: Can we namespace these context keys by `experimental_record::` or something like this to avoid collision ?

apollo-router/src/plugins/record_replay/record.rs
```diff
@@ -0,0 +1,342 @@
...
+
+                    let headers = parts.headers.clone();
+                    let context = res.context.clone();
+
+                    let after_complete = once(async move {
+                        if recording_enabled(&context) {
+                            let client_request =
+                                context.get::<_, RequestDetails>(CLIENT_REQUEST)?;
+                            let client_response =
+                                context.get::<_, ResponseDetails>(CLIENT_RESPONSE)?;
```
bnjjj, line 122: Maybe we can just use the private context

apollo-router/src/plugins/record_replay/record.rs
```diff
@@ -0,0 +1,342 @@
...
+                                    .operation_name
+                                    .clone()
+                                    .unwrap_or("UnnamedOperation".to_string());
+
+                                let unix_time = std::time::SystemTime::now()
+                                    .duration_since(std::time::UNIX_EPOCH)?
+                                    .as_secs();
+
+                                let filename =
+                                    format!("{}-{}.json", operation_name, unix_time).to_string();
```
bnjjj, line 143: I think in the future it might be better to hash the query and use that hash in the filename instead of unix_time. If we do that then for the same operation/request it will override the previous file

apollo-router/src/plugins/record_replay/record.rs
```diff
@@ -0,0 +1,342 @@
...
+                    };
+
+                if recording_enabled {
+                    let query = req.supergraph_request.body().query.clone();
+                    let operation_name = req.supergraph_request.body().operation_name.clone();
+                    let variables = req.supergraph_request.body().variables.clone();
+                    let headers = externalize_header_map(req.supergraph_request.headers())
+                        .expect("failed to externalize header map");
+                    req.context
+                        .upsert::<_, RequestDetails>(CLIENT_REQUEST, |_value| RequestDetails {
```
bnjjj, line 216: private context might be more convenient

apollo-router/src/plugins/record_replay/record.rs
```diff
@@ -0,0 +1,342 @@
...
+                }
+                req
+            })
+            .map_response(|res: supergraph::Response| {
+                if recording_enabled(&res.context) {
+                    let context = res.context.clone();
+
+                    return res.map_stream(move |chunk| {
+                        context
+                            .upsert::<_, ResponseDetails>(CLIENT_RESPONSE, |mut value| {
```
bnjjj, line 232: Same for private context

apollo-router/src/plugins/record_replay/record.rs
```diff
@@ -0,0 +1,342 @@
...
+            .service(service)
+            .boxed()
+    }
+
+    fn execution_service(&self, service: execution::BoxService) -> execution::BoxService {
+        ServiceBuilder::new()
+            .map_request(|req: execution::Request| {
+                if recording_enabled(&req.context) {
+                    req.context
+                        .insert(QUERY_PLAN, req.query_plan.formatted_query_plan.clone())
```
bnjjj, line 251: private context

apollo-router/src/plugins/record_replay/replay.rs
```diff
@@ -0,0 +1,418 @@
...
+
+#[derive(Debug)]
+pub(crate) struct Replay {
+    recording: Recording,
+    pub(crate) report: Arc<Mutex<Vec<ReplayReport>>>,
+}
+
+#[allow(dead_code)]
+impl Replay {
+    pub(crate) async fn from_file(recording_file: String) -> Result<Self, BoxError> {
```
bnjjj, line 36: IMHO `recording_file` should be a `&Path`

apollo-router/src/plugins/record_replay/replay.rs
```diff
@@ -0,0 +1,418 @@
...
+        let recording = fs::read_to_string(recording_file).await?;
+        let recording: Recording = serde_json::from_str(&recording)?;
+        Ok(Self::new(recording))
+    }
+
+    pub(crate) fn new(recording: Recording) -> Self {
+        Self {
+            recording,
+            report: Arc::from(Mutex::from(vec![])),
```
bnjjj, line 45: NIT: I think it should work
```suggestion
            report: Arc::default(),
```

apollo-router/src/plugins/record_replay/replay.rs
```diff
@@ -0,0 +1,418 @@
...
+            }
+        }
+
+        let req = supergraph::Request::builder()
+            .query(client_request.query.unwrap().clone())
+            .and_operation_name(client_request.operation_name.clone())
+            .variables(client_request.variables.clone())
+            .headers(request_headers)
+            .context(Context::default())
+            .uri(Uri::from_static("http://localhost:8080"))
```
bnjjj, line 66: Are we overriding the original value ? Do we care about URI and method ?

apollo-router/src/plugins/record_replay/replay.rs
```diff
@@ -0,0 +1,418 @@
...
+
+    fn supergraph_service(&self, service: supergraph::BoxService) -> supergraph::BoxService {
+        let report = self.report.clone();
+        let recorded_chunks = self.recording.client_response.chunks.clone();
+
+        ServiceBuilder::new()
+            .map_request(|req: supergraph::Request| {
+                // TODO - nothing? request is coming from elsewhere
+                req
+            })
```
bnjjj, line 132: to delete

apollo-router/src/plugins/record_replay/replay.rs
```diff
@@ -0,0 +1,418 @@
...
+            .map_response(|res: supergraph::Response| {
+                let mut i = 0;
+                res.map_stream(move |chunk| {
+                    let recorded_chunk = &recorded_chunks[i];
+                    let chunk = chunk.clone();
+
+                    // TODO - json string equality is sufficient?
```
bnjjj, line 139: I don't know, do we have cases when we have an HashMap in json data ? That might be annoying because depending on the machine it's running it might have different orders. We will discover it I guess

apollo-router/src/plugins/record_replay/replay.rs
```diff
@@ -0,0 +1,418 @@
...
+                .replace('\x7f', "␡"),
+        )
+    } else {
+        Cow::Borrowed(s)
+    }
+}
+
+#[cfg(test)]
+#[path = "replay_tests.rs"]
+mod tests;
```
bnjjj, line 418: I think it's more idiomatic to get rid of that part and start the `replay_tests.rs` file with a global attribute like this 
```rust
#![cfg(test)]
// Content of the test file
```

apollo-router/src/plugins/record_replay/replay_tests.rs
```diff
@@ -0,0 +1,43 @@
+use console::style;
```
bnjjj, line 1:
```rust
#![cfg(test)]
```

apollo-router/src/plugins/record_replay/replay_tests.rs
```diff
@@ -0,0 +1,43 @@
...
+    if has_items == 0 {
+        println!("{}", style("Replay matched the recording 🎉").green());
+    } else {
+        println!();
+        for item in report.iter() {
+            item.print();
+            println!();
+        }
+    }
+}
```
bnjjj, line 43: Is it a real test ? If yes, it's missing some assert I think

lennyburdette: it's not, it's just a way to replay a recording and print some output. seemed cheaper than making an xtask task or something, but i'm open to ideas

bnjjj: You can just ignore that test to not run it on CI 
```
#[ignore]
```

lennyburdette: if i do that, i can't run it with cargo test at all. in CI it does nothing

apollo-router/src/plugins/record_replay/record.rs
```diff
@@ -0,0 +1,342 @@
...
+                .storage_path
+                .unwrap_or_else(default_storage_path),
+        );
+
+        write_file(
+            storage_path.clone().into(),
+            "README.md",
+            include_str!("recording-readme.md").as_bytes(),
+        )
+        .await?;
```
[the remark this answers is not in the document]
lennyburdette: good catch!

apollo-router/src/plugins/record_replay/replay.rs
lennyburdette: i moved the cfg(test) to the test file

apollo-router/src/plugins/record_replay/record.rs
lennyburdette: i'm not sure i want to overwrite files — i guess i could if i included other aspects of the request (variables, headers) in the hash too

lennyburdette: done!

apollo-router/src/plugins/record_replay/replay.rs
lennyburdette: done!

bnjjj (approved):

PR #4122, opened 2023-10-31
configurable error location

### LINKED ISSUE #4116: Authorization: more error configuration
follow up to #4076 and #4079

the dry-run option is supposed to ease migration, but the default behaviour of clients is to raise an exception on errors, so that would still make it a huge step. We could instead make that step easier with more configurable errors. #4076 gives the ability to remove errors from logs. But we could also have the following modes:
* put errors in the response's `extensions` field so it would not raise exceptions
* log errors but do not return them with the response

Geal:
Fix #4116

This makes authorization errors reporting more flexible, by configuring response errors in one of two ways:
- authorization errors in the errors fields
- authorization errors in the extensions fields
- not returning errors to the client

This will work nicely with the dry-run option, allowing a test of authorization that will not trigger exceptions in the client (due to the `errors` field), and so an easier integration path.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


docs/source/configuration/authorization.mdx
```diff
@@ -639,12 +639,17 @@ authorization:
...
+      log: true # default: true
+      response: "errors" # possible values: "errors" (default), "extensions", "disabled"
 ```
 
-By default, when part of a query is filtered by authorization, the list of filtered paths is added to the response and logged by the router. Filtering parts of a query according to the client's rights may be seen as normal operation, and not warrant investigation by platform operators, so those logs can be disabled.
\ No newline at end of file
+By default, when part of a query is filtered by authorization, the list of filtered paths is added to the response and logged by the router. We may need more flexibility in reporting errors, for various reasons:
+- filtering parts of a query according to the client's rights may be seen as normal operation, and not warrant investigation by platform operators, so logs should be disabled
+- when testing authorization, we may not want the clients to raise exceptions if authorization errors are present in the response's `errors` field, so we can put them in `extensions` instead
+- we may not want to give the clients any information about authorization errors, so they can be disabled entirely
```
lennyburdette, line 655: maybe a note about combining `response: extensions` or `response: disabled` with `dry_run: true`?

lennyburdette: or maybe a non-normative section about the adoption story (add directives, dry_run+extensions/logging, validate behavior, dry_run: false/errors)

lennyburdette (approved):

docs/source/configuration/authorization.mdx
BrynCooke, line 655: Suggestion that doesn't render nicely as a suggest:

By default, when part of a query is filtered by authorization, the list of filtered paths is added to the response and logged by the router. This is behavior can be customized for your needs.

#### `log`
By setting the enabled option you can choose if query filtering will result in a log event being output.
  
```yaml title="router.yaml"
authorization:
  preview_directives:
    log_errors: true
    errors:
      log: false # default: true
```

If filtering parts of a query according to the client's rights may be seen as normal operation, and not warrant investigation by platform operators, logging should be disabled.

#### `response`

By default filtered are places in the errors part of the graphql response.

You can configure `response` to be:
* `errors` (default) - to place filtered paths in graphql errors.
* `extensions` - to place filtered paths in extensions. Useful to suppress exceptions on the client side while still giving information that parts of the query were filtered.
* `disabled` - to suppress all information that the query was filtered.


```yaml title="router.yaml"
authorization:
  preview_directives:
    log_errors: true
    errors:
      response: errors # errors (default) | extensions | disabled
```

BrynCooke (commented):
I made a docs suggestion though.
There's no test for the logging functionality, maybe add that to the test harness as I'm sure this comes up many times, but we never test.....

.changesets/feat_geal_authz_log_errors.md
```diff
@@ -1,13 +1,15 @@
...
   preview_directives:
-    log_errors: true
+    errors:
+      log: true # default: true
+      response: "errors" # possible values: "errors" (default), "extensions", "disabled"
 ```
 
-By [@Geal](https://github.com/Geal) in https://github.com/apollographql/router/pull/4076
\ No newline at end of file
+Logging can be disabled if platform operators do not want to see the logs polluted by common authorization errors. And errors in responses can be moved to extensions to avoid exceptions raised by the clients. Or they can be disabled entirely if the clients should not learn anything about the authorization errors.
```
garypen, line 13:
```suggestion
Logging can be disabled if platform operators do not want to see the logs polluted by common authorization errors.
Errors in responses may be:
 - moved to extensions, to avoid raising exceptions in clients
 - or disabled entirely, in which case clients will not receive any authorization errors.
```

garypen (approved):

docs/source/configuration/authorization.mdx
Geal: yup, I think we should have a section on how to migrate easily. But maybe that should live in odyssey?

Geal: @shorgi ^

Geal:
> There's no test for the logging functionality, maybe add that to the test harness as I'm sure this comes up many times, but we never test.....

I looked a bit into it, that looks too large to add in this PR, but I can commit to it as a folow up. Maybe something like [tracing-mock](https://github.com/tokio-rs/tracing/tree/master/tracing-mock) or [tracing-test](https://docs.rs/tracing-test/latest/tracing_test/) could help

Geal:
I added a test for logs in 506e043b

Geal:
I'm removing the test, it has issues with a global subscriber being set somewhere. I'll look at that later

PR #4124, opened 2023-10-31
docs: add caution admonition in Rhai docs for handling non-existent header

shorgi:
In Rhai docs, added Caution admonition for non-existent header. See [section with admonition in deploy preview](https://deploy-preview-4124--apollo-router-docs.netlify.app/customizations/rhai#manipulating-headers-and-the-request-context).


docs/source/customizations/rhai.mdx
```diff
@@ -375,6 +375,29 @@ fn process_response(response) {
 }
 ```
 
+<Caution>
+
+**Accessing a non-existent header throws an exception**. 
+
+To safely check for the existence of a header before reading it, use `request.headers.contains("header-name")`. 
+
+When using `contains` within `subgraph_service`, you must assign your headers to a temporary local variable. Otherwise, `contains` throws an exception that it "cannot mutate" the original request. For example:
```
Meschreiber, line 384: When I first read this, I wasn't sure if the example I was about to read was an incorrect example that would throw the previously mentioned exception, or if it would be a correct example. Additional spacing and being explicit could help.

```suggestion
When using `contains` within `subgraph_service`, you must assign your headers to a temporary local variable. Otherwise, `contains` throws an exception that it "cannot mutate" the original request.

For example, the following `subgraph_service` function checks whether the `x-custom-header` exists before processing it.
```

Meschreiber (approved):
One suggestion for extra clarity, but otherwise lgtm!

Geal (approved):

Issue #4121, opened 2023-10-31
Evaluate the use of threads within the router


garypen:
It's difficult to debug memory issues in the router because:
 - We are not always disciplined about `join`-ing our threads.
 - Interactions between Threads and Tokio Runtimes are unclear.

We should carefully examine our use of threads and only use them where they are required to handle non-async blocking tasks. Mostly this is in the area of Telemetry where dropping metrics or tracing providers is blocking and synchronous.

Explicitly managed means that if the thread has a carefully managed lifespan, scoped or joined at a point guaranteed to occur before `main` exits, then we can leave that alone for now.

```[tasklist]
### Examine Source
- [x] apollo-router/src/plugins/authentication/subgraph.rs (Test only, ignore.)
- [x] apollo-router/src/plugins/telemetry/mod.rs (Threads are spawned and never joined.)
- [x] apollo-router/src/plugins/rhai/mod.rs (Thread is joined in Drop.)
```

PR #4127 addresses the second and third use cases by making use of a new utility struct `DropWatch` which centralises our use of threads into a single implementation. I've not done this for `subgraph.rs` since it's a special situation used in a test.

PR #4127, opened 2023-11-01
Tidy up various thread interactions

garypen:
Introduce the DropWatch utility struct which can be used to watch a closure and either:
 - execute the closure and then wait for the DropWatch struct to be dropped
 - or wait for the struct to be dropped and then execute the closure

It's primarily useful when dealing with Drops that are triggered from an async context, but could be useful wherever we have a thread that we'd like to have a managed lifetime associated with a Drop.

fixes: #4121

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/drop_watch.rs
+++ b/apollo-router/src/drop_watch.rs
@@ -69,3 +69,72 @@ impl Drop for DropWatch {
         wh.join().expect("drop watcher thread terminating");
     }
 }
+
+#[cfg(test)]
+mod test {
+    use std::sync::atomic::{AtomicU8, Ordering};
+    use std::sync::mpsc;
+    use std::sync::Arc;
+
+    use super::DropWatch;
+
+    #[test]
+    fn it_runs_and_waits_when_dropped() {
+        let value = Arc::new(AtomicU8::new(0));
+        let shared_value = value.clone();
+        let (tx, rx) = mpsc::channel();
+
+        let _ = DropWatch::run_and_wait(move || {
+            shared_value.fetch_add(1, Ordering::Relaxed);
+            tx.send(()).unwrap();
+        });
+        let _ = rx.recv();
+        assert_eq!(value.load(Ordering::Relaxed), 1);
+    }
+
+    #[test]
+    fn it_runs_and_waits_when_alive() {
+        let value = Arc::new(AtomicU8::new(0));
+        let shared_value = value.clone();
+        let (tx, rx) = mpsc::channel();
+
+        let watcher = DropWatch::run_and_wait(move || {
+            shared_value.fetch_add(1, Ordering::Relaxed);
+            tx.send(()).unwrap();
+        });
+        let _ = rx.recv();
+        assert_eq!(value.load(Ordering::Relaxed), 1);
+        drop(watcher);
+        assert_eq!(value.load(Ordering::Relaxed), 1);
+    }
+
+    #[test]
+    fn it_waits_and_runs_when_dropped() {
+        let value = Arc::new(AtomicU8::new(0));
+        let shared_value = value.clone();
+        let (tx, rx) = mpsc::channel();
+
+        let _ = DropWatch::wait_and_run(move || {
+            shared_value.fetch_add(1, Ordering::Relaxed);
+            tx.send(()).unwrap();
+        });
+        let _ = rx.recv();
+        assert_eq!(value.load(Ordering::Relaxed), 1);
+    }
+
+    #[test]
+    fn it_waits_and_runs_when_alive() {
+        let value = Arc::new(AtomicU8::new(0));
+        let shared_value = value.clone();
+        let (tx, rx) = mpsc::channel();
+
+        let watcher = DropWatch::wait_and_run(move || {
+            shared_value.fetch_add(1, Ordering::Relaxed);
+            tx.send(()).unwrap();
+        });
+        assert_eq!(value.load(Ordering::Relaxed), 0);
+        drop(watcher);
+        let _ = rx.recv();
+        assert_eq!(value.load(Ordering::Relaxed), 1);
+    }
+}
--- a/apollo-router/src/drop_watch.rs
+++ b/apollo-router/src/drop_watch.rs
@@ -1,0 +1,71 @@
+//! Provide a [`DropWatch`] utility.
+//!
+//! Accept a FnOnce and carefully watch over it until we are dropped.
+//!
+//! Useful when we want to join threads which will ultimately be terminated within an asynchronous
+//! context.
+
+use std::mem::ManuallyDrop;
+use std::sync::atomic::AtomicBool;
+use std::sync::atomic::Ordering;
+use std::sync::Arc;
+use std::thread::JoinHandle;
+
+pub(crate) struct DropWatch {
+    watcher_handle: ManuallyDrop<JoinHandle<()>>,
+    park_flag: Arc<AtomicBool>,
+}
+
+impl DropWatch {
+    /// Create a DropWatch which will execute the supplied closure then wait for notification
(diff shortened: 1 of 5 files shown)


BrynCooke:
I think I'm going to need some more explanation as to why this is going to work.
It'll park one of the tokio threads while doing its thing, but why is this OK?

apollo-router/src/drop_watch.rs
```diff
@@ -0,0 +1,141 @@
...
+
+pub(crate) struct DropWatch {
+    watcher_handle: ManuallyDrop<JoinHandle<()>>,
+    park_flag: Arc<AtomicBool>,
+}
+
+impl DropWatch {
+    /// Create a DropWatch which will execute the supplied closure then wait for notification
+    pub(crate) fn run_and_wait<F: FnOnce() + Send + 'static>(watched: F) -> Self {
+        let park_flag = Arc::new(AtomicBool::new(false));
```
BrynCooke, line 22: How about using a tokio mutex? Create an owned lock guard and pass it into the spawned thread.
Then try to lock it in the original thread. This would remove the thread park stuff.

garypen: Because you can't `join` it. You have to `await`. This is specifically for code which is not async and which we would like to know it completed.

apollo-router/src/drop_watch.rs
```diff
@@ -0,0 +1,141 @@
...
+        let watching_flag = park_flag.clone();
+        let watcher_handle = std::thread::spawn(move || {
```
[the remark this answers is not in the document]
garypen: Because you can't `join` it. You have to `await`. This is specifically for code which is not async and which we would like to know it completed.

apollo-router/src/drop_watch.rs
```diff
@@ -0,0 +1,141 @@
...
+//! within an asynchronous context.
+
+use std::mem::ManuallyDrop;
+use std::sync::atomic::AtomicBool;
+use std::sync::atomic::Ordering;
+use std::sync::Arc;
+use std::thread::JoinHandle;
+
+pub(crate) struct DropWatch {
+    watcher_handle: ManuallyDrop<JoinHandle<()>>,
```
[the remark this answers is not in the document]
garypen: That's what I had originally, but the ManuallyDrop approach is cleaner. Like I said yesterday, `take` is only marked as `unsafe` because it's unsafe to do unless you know what you are doing. It's the best way to do it from a `Drop`, so that's what the comment and link to the documentation are recording in the `Drop` implementation.

BrynCooke (dismissed):
Approved, but we should have a couple more reviewers before merging.

apollo-router/src/drop_watch.rs
```diff
@@ -0,0 +1,141 @@
...
+}
+
+impl Drop for DropWatch {
+    fn drop(&mut self) {
+        // Safety: watcher_handle is taken, and must not be accessed again
+        // Since this is in Drop::drop, it is not possible to access it afterwards.
+        // https://doc.rust-lang.org/stable/std/mem/struct.ManuallyDrop.html#method.into_inner
+        let wh = unsafe { ManuallyDrop::<JoinHandle<()>>::take(&mut self.watcher_handle) };
+        self.park_flag.store(true, Ordering::Release);
+        wh.thread().unpark();
```
BrynCooke, line 68: Are we still not effectively blocking a tokio thread here?

garypen: At line 69 we have to wait for the join() to occur, but we are only waiting for our own thread to terminate. That's safe as long as our thread terminates. If our thread never terminated (infinite loop or other logic error), then we would have that problem anyway.
Summary: We want to join our threads to make sure that they execute correctly. That's part of the purpose of this struct and this is how we enforce it.

Another way to think about what this code is doing is to compare it to `spawn_blocking()`. Spawn blocking is waiting for an `async` task to complete. If we `await` the spawned task, then we are blocked until the spawned task completes. This code is doing exactly the same thing, expect it's:
 - for synchronous tasks, not async tasks (e.g. dropping file watchers, shutting down tracing)
 - synchronous itself, so we can join where we can't `await`.

Running it from an async context would only have the same impact as running `spawn_blocking()` in an async context and the awaiting that task.

apollo-router/src/plugins/rhai/mod.rs
```diff
@@ -146,10 +146,7 @@ impl Plugin for Rhai {
         )?));
         let watched_block = block.clone();
 
-        let park_flag = Arc::new(AtomicBool::new(false));
-        let watching_flag = park_flag.clone();
-
-        let watcher_handle = std::thread::spawn(move || {
+        let watcher_handle = move || {
             let watching_path = watched_path.clone();
```
Geal, line 150: instead of a thread, that one could be moved into a spawn_blocking

apollo-router/src/drop_watch.rs
```diff
@@ -0,0 +1,141 @@
...
+        wh.join().expect("drop watcher thread terminating");
```
Geal, line 69: I am not sure this is the behaviour we want. We need the various threads to be joined before stopping the process, but here that would be at the cost of an executor thread blocked waiting for the `DropWatch` to execute its stack. Which means that every time we reload schema or config, dropping the old tracer provider and meter provider would affect the async runtime

garypen: It's hard to say which is worse: spawning unbound tasks/threads (complicated debugging) or possible synchronisation delays (are these going to be long in duration?). I think the former, but I can see there's a case for the latter and we should benchmark out the impact.

If we are trying to address the former, then the runtime would be affected however we do it. The constraint is that we wish to "join" (either `join` for sync context or `await` for async context) a spawned `thread`/`task`. Right now we don't do that and it is a source of debugging confusion and (potentially) a source of future bugs.

The main problem is trying to execute `sync` code from a `sync` fn which "lives within" an `async` call stack. We can't change the behaviour of `Drop`, so we need a way to be able to address the problem. This is the least intrusive mechanism I can concoct.

You could try to concoct something which enters the async runtime during `Drop`, but that might be executing after tokio has finished and you still won't have a way to `await` the spawned task. (You could work around the "runtime has finished" problem using `try_current()` but you still wouldn't have synchronisation point if there was a runtime.

Geal: the main goal of the code before was to execute that synchronous task outside of the async executor. So joining on that thread in the async executor is a regression

garypen: My point still stands. The existing code (telemetry) just spawned a thread and didn't join it in. The new code does the same thing but joins. We could take out the join, but then we'd be back where we were before but with a slightly tidier utility class.

Geal: To rephrase my concerns here:
- dropping tracer providers in the async code was causing issues because it is a potentially long synchronous task, so it was moved to a thread
- I am not attached to the thread solution, and moving that pattern in a specific structure that encodes a correct behaviour sounds fine
- reintroducing the synchronous task in the async runtime is not a correct behaviour
- I agree that we need a way to track the threads we start, and we have to join on them at some point
- when to join on them is unclear to me at this point. Do you want to make sure that all threads have stopped before closing the process? Or is it during before switching to a new conf during a reload?
- as a side point, it is unclear to me how threads cause issues in tracking memory leaks with heaptrack, even with frequent reloads, but we can look into that elsewhere

garypen:
Following further discussion with Geoffroy and Bryn, let's ignore this PR and consider an alternative approach that uses channels and:
 - drops the tracer 
 - reloads metrics
at the other end of the channel.

PR #4131, opened 2023-11-02
Provide unsupported primitives for custom error handling + A layer that can sit on top of each endpoints

o0Ignition0o:
Steps before this can be merged:
- [ ] discuss exposed structures
- [ ] discuss gate name
- [ ] provide documentation for the structures
- [ ] describe what may change
- [ ] thoroughly test it


apollo-router/src/graphql.rs
```diff
@@ -160,6 +161,106 @@ impl Error {
...
+
+        Self {
+            message,
+            locations,
+            path,
+            extensions,
+        }
+    }
+
+    pub(crate) fn from_value(service_name: &str, value: Value) -> Result<Error, FetchError> {
```
xuorig, line 218: @o0Ignition0o forgot to mention that this function would be very useful to have `pub` if possible. It's super useful to handle responses for subgraphs in a custom client plugin.

o0Ignition0o: I don't expect this function to change anytime soon so I'll make it public right now, and discuss it with the team early next week.

Issue #4103, opened 2023-10-26
umbrella: futures channels aren't as bounded as we would like


garypen:
Based on this comment: https://github.com/rust-lang/futures-rs/issues/2287#issuecomment-743096329

it's apparent that a "bounded" channel isn't really "bounded" (at least not as I understand a bound), when you consider that cloning a `Sender` means that you will get a parked `SenderTask`.

To confirm my suspicion, I ran a test with a modified router which never read any of the messages from a futures channel and sure enough, the senders never failed and the router's memory grew at a rapid rate as memory was consumed on the sender side of the channel.

The main problem are where this first manifested is in the Apollo Exporter code and I've made the changes to convert the `futures` channel to a tokio channel. I then looked at doing this for all the`futures` channels we are using, but quickly realised that it wasn't a straightforward port because of the complexities of interactions with streams in various places in the router.

This umbrella issue lists all the locations where we would need to do more conversion work:
  
```[tasklist]
### Tasks
- [ ] https://github.com/apollographql/router/issues/4109
- [ ] https://github.com/apollographql/router/issues/4117
- [ ] https://github.com/apollographql/router/issues/4108
- [ ] apollo-router/src/query_planner/execution.rs
- [ ] apollo-router/src/query_planner/subscription.rs
- [ ] apollo-router/src/query_planner/tests.rs (NOT URGENT: test only)
- [ ] apollo-router/src/router/mod.rs (NOT URGENT: test only)
- [ ] apollo-router/src/services/execution_service.rs
- [ ] apollo-router/src/services/subgraph.rs
- [ ] apollo-router/src/services/subgraph_service.rs (NOT URGENT: test only)
- [ ] apollo-router/src/services/supergraph_service.rs
- [ ] https://github.com/apollographql/router/issues/4110
```

Note: I generated that list from a grep through the source code and so some of them might not require conversion when properly evaluated and some may be quite simple to do (like the apollo exporter was). However, I already know the first one that I looked at was tricky (files.rs), so YMMV. I'm going to update the urgency of these as and when I get time.

PR #4138, opened 2023-11-03
Replace final group of futures channels with tokio channels

garypen:
This group of channels had to be replaced in one go because they are all inter-related.

fixes: #4103

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/query_planner/execution.rs
```diff
@@ -3,7 +3,8 @@ use std::sync::Arc;
 
 use futures::future::join_all;
 use futures::prelude::*;
-use tokio::sync::broadcast::Sender;
+use tokio::sync::broadcast;
```
Geal, line 6: why the broadcast? The defer implementation only needed an mpsc before. Andthe broadcast channel has a very different performance profile

garypen: That was already there. I've just preserved the existing behaviour. If I haven't, then it's a mistake, but I just checked and all my changes look consistent to me.

BrynCooke (approved):

Geal (approved):

bnjjj (approved):

PR #4139, opened 2023-11-03
Add support for `telemetry.spans.mode` 

bnjjj:
Add support for `mode` option in `telemetry.spans` in order to keep the request root span if needed.

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [x] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/configuration/snapshots/apollo_router__configuration__tests__schema_generation.snap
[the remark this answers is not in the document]
bnjjj: I think @BrynCooke will know

apollo-router/src/plugins/telemetry/mod.rs
```diff
@@ -567,31 +615,6 @@ impl Plugin for Telemetry {
                         );
                         result
                     }
-                    // f.map(move |result: Result<SubgraphResponse, BoxError>| {
```
BrynCooke, line 570: Does this need deleting?

apollo-router/src/configuration/snapshots/apollo_router__configuration__tests__schema_generation.snap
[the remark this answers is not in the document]
BrynCooke: I don't know, this is specific to this branch. @bnjjj Did you delete it rather than accepts the snapshot?
If you like I can regen the snapshot on the base branch and then you can rebase?

apollo-router/src/axum_factory/utils.rs
```diff
@@ -135,7 +145,7 @@ impl PropagatingMakeSpan {
         ) {
             tracing::error_span!(
                 REQUEST_SPAN_NAME,
-                "http.method" = %request.method(),
+                "http.request.method" = %request.method(),
```
o0Ignition0o, line 148: Is this a breaking change ? 🤔

apollo-router/src/configuration/snapshots/apollo_router__configuration__tests__schema_generation.snap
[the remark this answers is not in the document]
o0Ignition0o: does this mean the test got removed? 🤔

apollo-router/src/plugins/telemetry/config_new/spans.rs
```diff
@@ -31,7 +31,6 @@ use crate::tracer::TraceId;
 #[derive(Deserialize, JsonSchema, Clone, Default, Debug)]
 #[serde(deny_unknown_fields, default)]
 pub(crate) struct Spans {
-    #[serde(skip)]
     /// Whether to create a `request` span. This will be removed in future, and users should set this to false.
     pub(crate) legacy_request_span: bool,
```
o0Ignition0o, line 35: should this be set to true by default to avoid a breaking change?

o0Ignition0o (approved):
LGTM overall, still a fair amount of work but we're getting closer!

apollo-router/src/axum_factory/utils.rs
bnjjj: Hmm maybe, we should add both then

apollo-router/src/plugins/telemetry/config_new/spans.rs
bnjjj: I think I discussed previously about this with @BrynCooke but I don't remember what was the outcome

BrynCooke: We are between a rock and hard place here.

We don't want new users to rely on the old span. We don't want to break existing users.

Suggest that for now it is true, and we log a warning if the user has not set it to false.
At some point in the future we flip the default value.

How does this sound?

bnjjj: Ok we have now a configuration called `mode` which can be set to `legacy` or `new` and should be fully compliant with what we had previously. That's a better approach. And by default we set the `new` mode. Which we should document in the router changelog

PR #4140, opened 2023-11-03
Introduce private and unstable plugin traits

Geal:
The goal of those traits is to allow experimentation and development on new plugin hook points without committing to a public API right away.

We add to new traits:
- PluginUnstable: contains experimental plugin methods that can be used publicly but for which there's no committment to stability, they can change at any moment
- PluginPrivate: internal trait that is not exposed publicly

This implements the decision record introduced in https://github.com/apollographql/router/commit/43c56400302ee754944866a092aef96957e0e480

*Description here*

Fixes #**issue_number**

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/plugin/mod.rs
```diff
@@ -360,13 +626,19 @@ pub(crate) trait DynPlugin: Send + Sync + 'static {
 
     /// Support downcasting
     fn as_any_mut(&mut self) -> &mut dyn std::any::Any;
+
+    /// unstable
+    fn unstable_method(&self);
+
+    /// internal
+    fn internal_method(&self, i: InternalType);
```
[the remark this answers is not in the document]
Geal: (I'll remove them before merging)

bnjjj (approved):

apollo-router/src/test_harness.rs
```diff
@@ -165,6 +167,45 @@ impl<'a> TestHarness<'a> {
         self
     }
 
+    /// Adds an extra, already instanciated unstable plugin.
```
garypen, line 170:
```suggestion
    /// Adds an extra, already instantiated unstable plugin.
```

apollo-router/src/test_harness.rs
```diff
@@ -165,6 +167,45 @@ impl<'a> TestHarness<'a> {
...
+                self.extra_plugins.len(),
+                std::any::type_name::<P>(),
+            ),
+        };
+
+        self.extra_plugins.push((name, Box::new(plugin)));
+        self
+    }
+
+    /// Adds an extra, already instanciated private plugin.
```
garypen, line 189:
```suggestion
    /// Adds an extra, already instantiated private plugin.
```

apollo-router/src/plugin/mod.rs
```diff
@@ -437,6 +747,42 @@ macro_rules! register_plugin {
     };
 }
 
+/// Register a private plugin with a group and a name
+/// Grouping prevent name clashes for plugins, so choose something unique, like your domain name.
+/// Plugins will appear in the configuration as a layer property called: {group}.{name}
+#[macro_export]
+macro_rules! register_private_plugin {
```
garypen, line 754: Are we missing `register_unstable_plugin` ?

apollo-router/src/plugin/mod.rs
```diff
@@ -187,8 +187,36 @@ impl PluginFactory {
     pub(crate) fn is_apollo(&self) -> bool {
         self.name.starts_with("apollo.") || self.name.starts_with("experimental.")
     }
+
+    /// Create a plugin factory.
+    pub fn new<P: PluginUnstable>(group: &str, name: &str) -> PluginFactory {
```
garypen, line 192: Perhaps better named as `new_unstable` ?

dev-docs/decision-records/unstable-plugins.md
```diff
@@ -23,14 +23,13 @@ Part of the motivation for this decision, was the recognition that we already ha
```
garypen, line 23: The changes to this document don't seem consistent with the changes to the code. For instance, the code has a `PluginPrivate`, but it's removed from the DR. Are these changes not properly synced up?

garypen (commented):
Looks good, but I can't reconcile the code changes with the DR changes.

dev-docs/decision-records/unstable-plugins.md
BrynCooke: Suspect this file just needs to be reverted

apollo-router/src/plugin/mod.rs
BrynCooke: What's the reason we need `new` and `new_private` at all?

Geal: it is used in the `register_plugin` macro

dev-docs/decision-records/unstable-plugins.md
Geal: it's reverted now

Geal:
merging this without changelog entries, as the point of those traits is internal work

PR #4144, opened 2023-11-04
Remove the first element from the Vec, not the last

### LINKED ISSUE #4143: fix ordering of batch responses
If a batch of requests is received and processed then the response to the final request is returned first.

i.e.:
```
Request   Response
   a          c
   b          a
   c          b
```
Ordering should be preserved since the client expects to get responses back in the same order

garypen:
This addresses the bug found in-house testing of experimental batching support.

Fixes #4143

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/services/router_service.rs
+++ b/apollo-router/src/services/router_service.rs
@@ -402,13 +402,16 @@ impl RouterService {
         if results.len() == 1 {
             Ok(results.pop().expect("we should have at least one response"))
         } else {
-            let first = results.remove(0);
+            let mut results_it = results.into_iter();
+            let first = results_it
+                .next()
+                .expect("we should have at least one response");
             let (parts, body) = first.response.into_parts();
             let context = first.context;
             let mut bytes = BytesMut::new();
             bytes.put_u8(b'[');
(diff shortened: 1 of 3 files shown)


apollo-router/src/services/router_service.rs
```diff
@@ -402,7 +402,7 @@ impl RouterService {
         if results.len() == 1 {
             Ok(results.pop().expect("we should have at least one response"))
         } else {
-            let first = results.pop().expect("we should have at least one response");
+            let first = results.remove(0);
```
Geal, line 405: `remove()` will shift all the array elements. Inset you can do something like:

```rust
let mut it = results.into_iter();
let first = it.next();

for result in it {
```

garypen: done: 2958874eda037951a5798e7c7f4fd0e9e743039f

Geal (approved):

lrlna (approved):

PR #4148, opened 2023-11-05
Centralise telemetry resource cleanup

garypen:
Use a new fn `check_spawned_task()` to utilise `spawn_blocking()` if tokio runtime is available, else just run the task in the current thread.
    
fixes: #4121

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/plugins/telemetry/mod.rs
```diff
@@ -235,6 +252,16 @@ impl Plugin for Telemetry {
...
 
+        // We may need 4 slots during a drop, so make it at least 4
+        let (tx, mut rx) = tokio::sync::mpsc::channel::<Box<dyn FnOnce() + Send + 'static>>(5);
+
+        // This task accepts blocking synchronous tasks and executes them
+        tokio::task::spawn_blocking(move || {
+            while let Some(msg) = rx.blocking_recv() {
+                msg();
+            }
+        });
```
Geal, line 263: what happens when the process is closed? Is this task canceled before it has flushed all data?

garypen: It depends what you mean by "closed". If it's a normal shutdown, then no, because (from tokio docs):

```
When you shut down the executor, it will wait indefinitely for all blocking operations to finish. You can use [shutdown_timeout](https://docs.rs/tokio/latest/tokio/runtime/struct.Runtime.html#method.shutdown_timeout) to stop waiting for them after a certain timeout. Be aware that this will still not cancel the tasks — they are simply allowed to keep running after the method returns. It is possible for a blocking task to be cancelled if it has not yet started running, but this is not guaranteed.
```

Of course if you `kill -15` a process then it's going to terminate in a fairly disorganised fashion, but that's already the case with or without this change.

BrynCooke: I've been thinking about this and think we can simplify the code and eliminate the channel.

If we are in an async context do the shutdown from within a tokio spawn blocking. If as we asserted above this is forced to finish before the tokio runtime is shut down then we're good.

If we are not in an async context we do the shutdown inline. As we are not async we are allowed to block.

WDYT?

garypen: I added the channel to make it more familiar than the previous thread notification approach. I'm fine with removing it. I don't think we need the code you suggest because I had that in place in an earlier revision of this code and couldn't trigger the "sync" code path. It just doesn't exist in our codebase. I deleted it to remove clutter.

BrynCooke: OK, let's do this, but chat one last time sync to be sure.

garypen: @BrynCooke + @Geal This is re-worked again. I think this is pretty close to what we want now.

BrynCooke (approved):

Geal (approved):

PR #4161, opened 2023-11-08
Run both JS & Rust graphql validations

### LINKED ISSUE #4159: experimental_graphql_validation to default to "both" (rust & js)
Switch `experimental_graphql_validation` default to `Both`, so it runs rust and js validation for everyone.

lrlna:
We've had `compiler@1.0.0` running in the in-house router safely for a week, which means it's time to switch graphql validation to be run in both rust & js for everyone. This changes the default of the validation config.

Fixes #4159 

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


Geal (approved):

garypen (approved):

Geal:
that test failure feels weird, why wouldn't it return the error?

Geal:
I found the issue, the code was replacing the query planner errors with validation errors without following the expected structure with usage reporting. I changed it to merge the errors inside the existing structure, so there will be some redundant errors, which make sense for the "Both" mode

goto-bus-stop:
The usage reporting is from query planning and doesn't exist for the rust-based validation. For validation failures it always has this result (indicating no usage) because it's the first thing that the query planner does.

By the time we're returning the errors from rust-based validation we already checked that the result matched the JS implementation, so appending them to planner errors always means we return duplicate errors.

Given that, is it important to keep the usage reporting in there?

bnjjj (approved):

Geal:
it's important to keep it because it is used in a lot of places in the router. The convention used in the planner was to have some special info in usage reporting in case of validation errors, in additions to the list of errors. I think we should revisit that behaviour soon, but for this PR to activate both JS and Rust validations, we should either:
- keep errors from both validations
- replace JS errors with rust ones, but keeping the usage reporting structure

goto-bus-stop:
OK I understand the usage reporting bit now :)

Rust errors have different error messages from JS, which I think people are currently relying on sometimes. Perhaps for now we should either just throw away the rust errors (by 1.0 we should have stable error codes which we can then tell people to rely on when we switch entirely to rust validation), or keep the duplicates like now

goto-bus-stop:
We should wait for the apollo-compiler 1.0.0-beta.6 update to land (which requires a federation-next release) as it fixes a possible DOS in operation validation

goto-bus-stop:
Updated status is to upgrade to 1.0.0-beta.8 in the router first, as it fixes a false positive validation error in apollo-rs.

SimonSapin (approved):

.changesets/feat_lrlna_graphql_validation_both.md
```diff
@@ -0,0 +1,6 @@
+### Use both Rust and JS Validation as default([issue#4159](https://github.com/apollographql/router/issues/4159))
+
+As part of the process to remove JavaScript validation in the Router, we are enabling the Router to run both validations as a default. This allows us to definitively assess reliability before completely removing JS validation.
```
lrlna, line 3:
```suggestion
As part of the process to replace JavaScript validation with a more performant Rust validation in the Router, we are enabling the Router to run both validations as a default. This allows us to definitively assess reliability and stability of Rust validation before completely removing JavaScript validation. As before, it's possible to toggle between implementations using `experimental_graphql_validation_mode` config key. Possible values are: `new` (runs only Rust-based validation), `legacy` (runs only JS-based validation), `both` (runs both in comparison, logging errors if a difference arises).
```

lrlna:
We are removing the Rust errors from the errors returned to the user in order to keep the output consistent. We can change that output all in one go with switching completely to Rust validation.

Issue #3947, opened 2023-10-02
Support JWT expiration with subscription


Geal:
If a JWT provided with a subscription request expires while the subscription is live, then the subscription should not be authorized anymore, and should be closed (there is no mechanism right now to renew a token while in a subscription).
The authentication plugin should store the expiration date, and add an execution plugin that checks the expiration date.

PR #4166, opened 2023-11-08
Enforce JWT expiration for subscriptions

garypen:
If a JWT expires whilst a subscription is executing, the subscription should be terminated.

fixes: #3947

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [x] Manual Tests

**Exceptions**

It's tricky to automatically test token expiration, so I've tested this feature manually. I used:
https://www.unixtimestamp.com/
https://jwt.io/
to generate timestamps and then JWT tokens using those timestamps as `exp` claims.

I ran a router configured to support JWT authn and subscription and then subscribed to a feed using `curl` and ensured my subscription timed out. Sample output snippet:
```
--graphql

{"payload":{"data":{"reviewAdded":{"id":4}}}}
--graphql

{}
--graphql

{"payload":null,"errors":[{"message":"subscription closed because the JWT has expired","extensions":{"code":"SUBSCRIPTION_JWT_EXPIRED"}}]}
--graphql--
``` 

**Notes**


apollo-router/Cargo.toml
```diff
@@ -72,6 +72,7 @@ base64 = "0.21.4"
 bloomfilter = "1.0.12"
 buildstructor = "0.5.4"
 bytes = "1.5.0"
+chrono = "0.4.31"
```
bnjjj, line 75: In the new telemetry PR we introduced https://docs.rs/time, do you think you can use this one instead of chrono ? It's lighter. And if not could you add  `default-features = false` and only add features you need on chrono ?

apollo-router/src/services/execution_service.rs
```diff
@@ -174,21 +179,74 @@ impl ExecutionService {
...
+                                let claims = x.as_object().expect("claims should be an object");
+                                let exp = claims.get("exp")?;
+                                if !exp.is_number() {
+                                    tracing::error!("JWT 'exp' (expiry) claim should be a number");
+                                    return None;
+                                }
+                                exp.as_i64()
+                            })
+                            .and_then(|seconds_since_epoch| {
+                                let dt = chrono::DateTime::from_timestamp(seconds_since_epoch, 0);
```
Geal, line 203: there's no need to add the chrono dependency to do this, we can do that with std:
https://github.com/apollographql/router/pull/4170/files#diff-f12f8cb5b38073d09c64f5d9739794bbd9a8882f807d809b02e44b36f0d4109eR49-R54

apollo-router/src/services/execution_service.rs
```diff
@@ -174,21 +179,74 @@ impl ExecutionService {
...
-                        &mut nullified_paths,
-                        response,
-                    )
-                }))
-            })
-            .boxed();
+        let stream =
+            stream
+                .map(move |mut response: Response| {
+                    // Enforce JWT expiry for deferred and subscription responses
```
Geal, line 185: could it be enforced on the request side, before executing the subscription?

bnjjj: Oh yes good catch ! I think you can move your logic right [here](https://github.com/apollographql/router/blob/dev/apollo-router/src/services/supergraph_service.rs#L515) gary to not try to fetch data from subgraphs if your jwt is expired

apollo-router/src/services/execution_service.rs
garypen: done: d30a606f91ae4400f256f98f9781e44f666997d4

apollo-router/Cargo.toml
garypen: I'm using @Geal 's suggestion and using SystemTime.

apollo-router/src/services/execution_service.rs
garypen: I've split the logic for enforcement and now enforce subscription expiration in the request as @bnjjj suggested. I've left the defer enforcement in the response. We could remove that I suppose, but it's probably better to have it than not.

bnjjj (approved):

apollo-router/src/services/supergraph_service.rs
```diff
@@ -424,11 +429,62 @@ async fn subscription_task(
...
+                .expect("we should not run before EPOCH")
+                .as_secs() as i64;
+            if now < ts {
+                Duration::from_secs((ts - now) as u64)
+            } else {
+                Duration::ZERO
+            }
+        }
+        None => Duration::MAX,
+    };
```
Geal, line 464: could we get that as a function exported from the authentication plugin?

Geal (approved):

bnjjj (approved):

apollo-router/src/services/supergraph_service.rs
garypen: done: bcd0bdad0db912564c2add1ac5b4d0197f433ee9

PR #4167, opened 2023-11-08
Update apollo-compiler with a fix to enum recursions

### LINKED ISSUE #4142: Parser recursion issue when compiling schema on startup
**Describe the bug**
When running the v1.33.x router with a subgraph that has an enum with over 500 values, the router fails to start with a recursion limit error:

```text
could not create router: schema error: GraphQL parser error: at index 22590: parser recursion limit reached
```

**To Reproduce**
1. Create a subgraph with an enum with 501 values
2. Compose the subgraph into a supergraph, e.g. `rover compose ...`
3. Run a vanilla router with the supergraph
4. Observe the error

Alternative MRE for reproduction:
1. Download supergraph that exhibits the issue: [gist](https://gist.github.com/bsaltz/29be7d38011a0b3085ecc0b95f8ed9c3)
2. Use that supergraph to reproduce, e.g. with Docker:

```shell
docker run -p 4000:4000 \
  --mount "type=bind,source=/absolute/path/to/supergraph.graphql,target=/dist/schema/local.graphql" \
  --rm \
  ghcr.io/apollographql/router:v1.33.2 -c config/router.yaml -s schema/local.graphql
```

3. Observe the error

Running with v1.32.0 does not break.

**Expected behavior**
Either a) there are no recursion errors on startup or b) the recursion limit for the startup parser is configurable.

**Output**

```text
could not create router: schema error: GraphQL parser error: at index 22590: parser recursion limit reached
```

 - OS: MacOS w/ M1
 - Version: 13.6

**Additional context**
So I believe I tracked this down to an `apollo-parser` change, a fix for which was merged literally a few hours ago: https://github.com/apollographql/apollo-rs/issues/666. Judging from the description and changes here, the fix changes the parsing for enums from recursion to looping.

The root cause of the issue came from https://github.com/apollographql/apollo-rs/pull/662. That PR was included in parser 0.6.3+ (the version in the router since 1.33.x), and it put recursion tracking in all the relevant bits of the parser, including the recursive enum parsing.

While that is being packaged into a release, I could get around this if I had a startup config option that let me control the recursion limit for the supergraph schema parser, which is initialized here: [schema.rs](https://github.com/apollographql/router/blob/79ba162b444c75bbd555eb294fe31ebbc556fa4c/apollo-router/src/spec/schema.rs#L63)

```rust
pub(crate) fn make_compiler(sdl: &str) -> Result<apollo_compiler::schema::Schema, SchemaError> {
    let mut parser = apollo_compiler::Parser::new(); // <- no way to pass in configs
    let ast = parser.parse_ast(sdl, "schema.graphql");
```

lrlna:
This updates the Router with the latest `apollo-compiler` version, which contains a fix for unnecessarily recursive enums and unions.

Fixes #4142

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


goto-bus-stop (approved):
technically the counting was accurate, it just uses recursion less now!

SimonSapin (approved):
I pushed another commit with a changelog entry.

SimonSapin:
Ugh sorry, not sure how I missed that you’d already added a changelog. I’ll deduplicate now.

Geal (approved):

bnjjj (approved):

Issue #4173, opened 2023-11-09
Redis configuration should accept the redis-cluster and rediss-cluster schemes


Geal:
Right now the client constructor bails if the scheme is not redis or rediss

https://github.com/apollographql/router/blame/65f97e11fbd38fb44bab75f910c8a2ffa29f6fbb/apollo-router/src/cache/redis.rs#L183-L199

PR #4174, opened 2023-11-09
Add missing schemas for redis connections

garypen:
We have supported additional schemas in our redis configuration for a while, but we never updated our redis connection logic to process the new schemas. This is now fixed.

fixes: #4173

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/cache/redis.rs
+++ b/apollo-router/src/cache/redis.rs
@@ -172,13 +172,16 @@ impl RedisCacheStorage {
     }
 
     fn preprocess_urls(urls: Vec<Url>) -> Result<Url, RedisError> {
-        match urls.get(0) {
+        let url_len = urls.len();
+        let mut urls_iter = urls.into_iter();
+        let first = urls_iter.next();
+        match first {
             None => Err(RedisError::new(
                 RedisErrorKind::Config,
                 "empty Redis URL list",
             )),
             Some(first) => {
-                if urls.len() == 1 {
+                if url_len == 1 {
                     return Ok(first.clone());
                 }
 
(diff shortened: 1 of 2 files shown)


apollo-router/src/cache/redis.rs
```diff
@@ -180,22 +189,16 @@ impl RedisCacheStorage {
...
-                        return Err(RedisError::new(
-                            RedisErrorKind::Config,
-                            format!(
-                                "invalid Redis URL scheme, expected 'redis' or 'rediss', got: {}",
-                                other
-                            ),
-                        ))
-                    }
+                if SUPPORTED_REDIS_SCHEMES.contains(&scheme) {
+                    let _ = result.set_scheme(scheme);
```
Geal, line 193: this will be a regression: in the previous code, if we set multiple `redis://` URLs, it resulted in a single `redis-cluster://` URL

garypen: Good catch. I'll push a change which preserves the behaviour of converting `redis` to `redis-cluster` and `rediss` to `rediss-cluster`

done: 8a6e8bcb838ffeba661076bd8d214e21304e9c29

bnjjj (approved):

apollo-router/src/cache/redis.rs
garypen: I thought some more about this overnight and concluded that change would still be a regression, since it couldn't handle incompatibilities with Sentinel. So, I've done a more comprehensive change which (I hope) will now address all of the various complexities.
see: 92ec1c187dda3dfb1fe6106fe82d196112e100ca

Geal (approved):

PR #4175, opened 2023-11-09
Entity caching: hash queries according to their usage of the schema

### LINKED ISSUE #4192: Entity caching: cache entries should survive schema updates if possible
Often, a schema update will not affect most of the existing queries (example: adding a field to a type changes nothing). This means that we could keep some cache entries across schema reloads too.

Geal:
Fix #4192

If a schema update does not affect an existing query (example: adding a field to a type), then we could detect it and keep using cached entities across reloads

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/spec/query/change.rs
```diff
@@ -0,0 +1,611 @@
...
+            return Err("invalid number of selections for _entities query".into());
+        }
+
+        match node.selection_set.first() {
+            Some(Selection::Field(field)) => {
+                if field.name.as_str() != "_entities" {
+                    return Err("expected _entities field".into());
+                }
+
+                "_entities".hash(self);
```
bnjjj, line 212: I think we already have a constant containing this value. If not, it could be interesting to have it and replace all existing occurences

apollo-router/src/spec/query/change.rs
```diff
@@ -0,0 +1,611 @@
...
+            }
+            _ => Err("expected _entities field".into()),
+        }
+    }
+}
+
+impl<'a> Hasher for QueryHashVisitor<'a> {
+    fn finish(&self) -> u64 {
+        //self.hasher.finalize()
+        todo!()
```
bnjjj, line 237: Is it a mistake ?

apollo-router/src/spec/query/change.rs
```diff
@@ -0,0 +1,611 @@
+use std::collections::HashMap;
```
garypen, line 1: This all feels like it might be something that is generally useful within `apollo-rs`. It's fine to have it in the router right now, but consider adding to appropriate `apollo-rs` crate at some point.

garypen (approved):

apollo-router/src/spec/query/change.rs
Geal: yes, it will be generally useful elsewhere. Probably in federation-next

apollo-router/src/spec/query/change.rs
Geal: fixed in 5d8fc84c5

At some point we should have a central place for all those consts

apollo-router/src/spec/query/change.rs
Geal: it is not used, the real hash is generated elsewhere

PR #4179, opened 2023-11-09
Remove unused crates from the project

garypen:
Remove some unused crates and convert our redis tests to use the `fred` crate. That means we can remove our dev dependency on the `redis` crate.

This yielded a 1.5% reduction in test build time on my laptop.

Note: This is an exploratory PR, since it's fairly tricky to get unused crates information across multiple platforms with various conditional compilations in play.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -372,7 +372,6 @@ dependencies = [
  "proteus",
  "rand 0.8.5",
  "rand_core 0.6.4",
- "redis",
  "regex",
  "reqwest",
  "rhai",
@@ -388,10 +387,9 @@ dependencies = [
  "serde_json_bytes",
  "serde_urlencoded",
  "serde_yaml",
- "sha1 0.10.6",
+ "sha1",
  "sha2",
  "shellexpand",
- "similar-asserts",
  "static_assertions",
  "strum_macros",
  "sys-info",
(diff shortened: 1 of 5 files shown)


lrlna (approved):

bnjjj (approved):

SimonSapin:
I don’t find the lint in the diff. Is it meant to be a new file?

garypen:
> I don’t find the lint in the diff. Is it meant to be a new file?

https://github.com/apollographql/router/pull/4179/commits/fd4b54b7ea4b27d355aeca49eeef185ef3bcba02

Geal:
the goal of using a different client was to make sure that there's nothing specific to fred that would mess with the data, but we can remove it yes

Geal (approved):

garypen:
> the goal of using a different client was to make sure that there's nothing specific to fred that would mess with the data, but we can remove it yes

I did wonder about that, but concluded that we should be testing `fred` since it's our primary (in fact, only) client in the router. We could add back the old test if it's useful.

PR #4186, opened 2023-11-10
docs: AuthZ

shorgi:


Geal (approved):

docs/source/configuration/authorization.mdx
```diff
@@ -417,19 +417,18 @@ If _every_ requested field requires authentication and a request is unauthentica
...
+At the [`RouterService` level](../customizations/overview#the-request-lifecycle), the Apollo Router extracts the list of policies relevant to a request from the schema and then stores them in the request's context in `apollo_authorization::policies::required` as a map `policy -> null|true|false`.
 
-You need to provide a Rhai script or a coprocessor at the `SupergraphService` level that evaluates the map, setting the value to `true` if the policy is validated and `false` if not. After that, the router filters the requests' types and fields to only those where the policy is `true`.
+At the `SupergraphService` level, you need to provide a Rhai script or a coprocessor that evaluates the map. If the policy is validated, set its value to `true`, otherwise set it to `false`. Afterwards, the router filters the requests' types and fields to only those where the policy is `true`.
```
Meschreiber, line 429:
```suggestion
At the `SupergraphService` level, you must provide a Rhai script or coprocessor to evaluate the map. If the policy is validated, the script or coprocessor should set its value to `true` or otherwise set it to `false`. Afterward, the router filters the requests' types and fields to only those where the policy is `true`.
```

docs/source/configuration/authorization.mdx
```diff
@@ -417,19 +417,18 @@ If _every_ requested field requires authentication and a request is unauthentica
...
 
-Like the efficiencies gained via the other authorization directives, if every field on a particular subgraph query is unauthorized because none of its policies pass, this can eliminate entire subgraph requests.
+If no field of a subgraph query passes its authorization policies, the router stops further processing of the query and precludes unauthorized subgraph requests. This efficiency gain is a key benefit of the `@policy` directive and other authorization directives.
```
Meschreiber, line 431:
```suggestion
If no field of a subgraph query passes its authorization policies, the router stops further processing of the query and precludes unauthorized subgraph requests. This efficiency gain is a key benefit of the `@policy` and other authorization directives.
```

docs/source/configuration/authorization.mdx
```diff
@@ -823,26 +814,26 @@ With introspection turned off, you can use GraphOS's [schema registry](/graphos/
 
 ## Configuration options
 
-The authorization plugin behaviour can be modified with various options.
+The behavior of the authorization plugin can be modified with various options.
 
 ### reject_unauthorized
 
+The `reject_unauthorized` option configures whether to reject an entire query if any authorization directive failed, where any part of it is filtered by authorization directives. When enabled, a response contains the list of paths that are affected.
```
Meschreiber, line 821:
```suggestion
The `reject_unauthorized` option configures whether to reject an entire query if any authorization directive failed, or any part of the query was filtered by authorization directives. When enabled, a response contains the list of paths that are affected.
```

docs/source/configuration/authorization.mdx
```diff
@@ -851,14 +842,17 @@ authorization:
...
+<Note>
+
+The `log` option should be disabled if filtering parts of queries according to the client's rights is approved as normal operation by platform operators.
+
+</Note>
 
 #### response
 
-By default, filtered paths are placed in the errors part of the graphql response.
+You can configure `response` to set where to place filtered paths in a GraphQL response:
```
Meschreiber, line 853:
```suggestion
You can configure `response` to define what part of the GraphQL response should include filtered paths:
```

Meschreiber (approved):
Just some small wording suggestions, otherwise LGTM!

Perhaps this will happen in a separate PR for GA, but will the configuration name `preview_directives` change and will directives become enabled by default? E.g.:

```router.yaml
authorization:
  directives:
    enabled: true
```

docs/source/configuration/authorization.mdx
```diff
@@ -417,19 +417,18 @@ If _every_ requested field requires authentication and a request is unauthentica
 ### `@policy`
```
dylan-apollo, line 417: Is it too late to change the name? 😅 Seems like `requiresPolicies` would be more consistent with the other `requiresScopes` directive.

docs/source/configuration/authorization.mdx
```diff
@@ -417,19 +417,18 @@ If _every_ requested field requires authentication and a request is unauthentica
...
 
 The `@policy` directive marks fields and types as restricted based on authorization policies evaluated in a [Rhai script](../customizations/rhai/) or [coprocessor](../customizations/coprocessor). This enables custom authorization validation beyond authentication and scopes.
-The directive should include a `policies` argument that defines an array of the required policies.
+
+The `@policy` directive includes a `policies` argument that defines an array of the required policies. The following example shows a policy that requires all roles from the claims object to contain the string `"support"`:
```
dylan-apollo, line 421: I think it'd be better here to use a generic policy string, not Rhai code (I'm not sure we should be suggesting Rhai code in general, more on that later). This paragraph & example make it seem like the default behavior of policies is to execute some code, which isn't the case.

docs/source/configuration/authorization.mdx
dylan-apollo: What happens if the policy is still `null` (e.g., because it's not implemented in the coprocessor yet)?

docs/source/configuration/authorization.mdx
```diff
@@ -442,46 +441,38 @@ extend schema
     import: [..., "@policy"])
```
dylan-apollo, line 441: I assume this will need to be fed 2.6 since 2.5 is out and doesn't support this yet

docs/source/configuration/authorization.mdx
```diff
@@ -442,46 +441,38 @@ extend schema
...
 @requiresScopes(policies: [["policy1", "policy2", "policy3"]])
 ```
 
-For the preceding example, a request would need `policy1` **AND** `policy2` **AND** `policy3` to be authorized.
-
-You can use nested arrays to introduce **OR** logic:
+Alternatively, to introduce **OR** logic you can use nested arrays. For the following example, a request would need `policy1` **OR** `policy2` **OR** `policy3` to be authorized:
 
 ```graphql
 @requiresScopes(policies: [["policy1"], ["policy2"], ["policy3"]])
```
dylan-apollo, line 464:
```suggestion
@policy(policies: [["policy1"], ["policy2"], ["policy3"]])
```

docs/source/configuration/authorization.mdx
```diff
@@ -442,46 +441,38 @@ extend schema
...
 ```graphql
 scalar federation__Policy
 directive @policy(scopes: [[federation__Policy!]!]!) on OBJECT | FIELD_DEFINITION | INTERFACE | SCALAR | ENUM
 ```
 
-Using the `@policy` directive requires a [Supergraph plugin](../customizations/overview) to evaluate the authorization policies. You can do this with a [Rhai script](../customizations/rhai/), [coprocessor](../customizations/coprocessor), or [native plugin](../customizations/native). Refer to the following Rhai script and coprocessor examples for more information.
-
+Using the `@policy` directive requires a [Supergraph plugin](../customizations/overview) to evaluate the authorization policies. You can do this with a [Rhai script](../customizations/rhai/) or [coprocessor](../customizations/coprocessor). Refer to the following [Rhai script](#usage-with-a-rhai-script) and [coprocessor](#usage-with-a-coprocessor) examples for more information. (Although a [native plugin](../customizations/native) can also evaluate authorization policies, we don't recommend using it.)
 
 #### Combining policies with `AND`/`OR` logic
```
dylan-apollo, line 453: Should we combine this with the `@requiresScopes` description of AND/OR? I believe it works the same way, but maybe it's better to keep the definition separate, unsure.

docs/source/configuration/authorization.mdx
```diff
@@ -442,46 +441,38 @@ extend schema
...
 @policy(policies: [["policy", "policy2"], ["policy3"]])
 ```
 
-This syntax requires requests to have either (`policy1` **AND** `policy2`) **OR** just `policy3` to be authorized.
-
-
 ##### Usage with a Rhai script
 
-The `policies` argument contains a list of strings with no formatting constraints. That means you can use them to store Rhai code.
+The `policies` argument contains a list of strings with no formatting constraints, meaning you can use them to store Rhai code.
```
dylan-apollo, line 475: I don't think we should recommend that folks do `eval` of arbitrary Rhai code from the subgraphs. There are a couple of problems:

1. You're suddenly giving developers who have access to subgraphs access to run code at the router level, which could let them get data they shouldn't have somehow? Maybe not the hugest issue, but worries me.
2. If someone writes invalid Rhai code in _any_ subgraph, they could break **every** operation which tests the policy—it wouldn't just cause that new field to get omitted, it'd cause entire requests to be aborted with an error.

docs/source/configuration/authorization.mdx
```diff
@@ -554,7 +545,7 @@ coprocessor:
       context: true
 ```
 
-then a coprocessor can then receive a request with this format:
+A coprocessor can then receive a request with this format:
```
dylan-apollo, line 548: The coprocessor example feels like one that could use `requiresScopes`. Can we reframe it to make it more obvious why you'd use `policy` in some cases and when you _should_ use `requiresScopes` instead? This might be a bigger point we want to make anyway, why does `@requiresScopes` exist if `@policy` is more powerful?

dylan-apollo (changes requested):
Left some comments based on my testing this feature. Not really related to the changes in this PR, but more the overall docs for `@policy`

docs/source/configuration/authorization.mdx
shorgi: cc @Geal, for this and other comments from Dylan

dylan-apollo (approved):

PR #4193, opened 2023-11-13
Entity cache: influence the cache key through a context entry

### LINKED ISSUE #4190: Entity caching: use a context key as part of the cache key
we need a way to influence the cache key depending on information outside of the query. For the query plan cache, this was done through an entry in the context, we can reuse the same approach here

Geal:
Fix #4190

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/plugins/cache/entity.rs
+++ b/apollo-router/src/plugins/cache/entity.rs
@@ -34,6 +34,7 @@ use crate::Context;
 
 const ENTITIES: &str = "_entities";
 pub(crate) const REPRESENTATIONS: &str = "representations";
+pub(crate) const CONTEXT_CACHE_KEY: &str = "entity_cache_key";
 
 register_plugin!("apollo", "experimental_entity_cache", EntityCache);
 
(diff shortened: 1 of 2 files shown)


apollo-router/src/plugins/cache/entity.rs
```diff
@@ -34,6 +34,7 @@ use crate::Context;
 
 const ENTITIES: &str = "_entities";
 pub(crate) const REPRESENTATIONS: &str = "representations";
+pub(crate) const CONTEXT_CACHE_KEY: &str = "entity_cache_key";
```
bnjjj, line 37: Can we namespace it with `apollo.cache::entity_cache_key` for example ?
```suggestion
pub(crate) const CONTEXT_CACHE_KEY: &str = "apollo.cache::entity_cache_key";
```

Geal: do we use such a convention elsewhere?

bnjjj: Ok I'm wrong, if it's an exposed key for the user don't namespace it. But put it in public and also provide it in rhai globals please. I did the same for subscription context key

bnjjj (approved):
Don't forget to document it in #4195

PR #4196, opened 2023-11-13
Entity caching: do not store entities with errors in cache

### LINKED ISSUE #4191: Entity caching: do not store entities with errors
~Some entities might have errors associated with them, so we have to split them out of the subgraph response~ after more discussion, the expected behaviour is to avoid storing entries that would have errors

Geal:
Fix #4191

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/plugins/cache/entity.rs
```diff
@@ -536,45 +544,96 @@ fn filter_representations(
...
+                        error.path = Some(
+                            Path(vec![
+                                PathElement::Key("_entities".to_string()),
+                                PathElement::Index(new_entity_idx),
+                            ])
+                            .join(path),
+                        );
+                    }
+                    new_errors.push(error);
+                }
```
bnjjj, line 588: I'm wondering if it would not be more efficient to filter_map on v.errors and then extend new_errors directly to avoid re allocation of the new_errors vec. I might be wrong

bnjjj (approved):

Geal:
there's another opinion here: let's not store in cache the entities for which we had errors

apollo-router/src/plugins/cache/entity.rs
```diff
@@ -614,49 +621,81 @@ fn filter_representations(
...
-                            data: value.clone(),
-                        }),
-                    ));
+                    let mut has_errors = false;
+                    for error in errors.iter().filter(|e| {
+                        e.path
+                            .as_ref()
+                            .map(|path| {
+                                path.starts_with(&Path(vec![
+                                    PathElement::Key("_entities".to_string()),
```
bnjjj, line 674: Could you use a constant for `_entities` please ?

PR #4197, opened 2023-11-14
Update `opentelemetry` to version `0.21`

jtescher:
Updates the other otel ecosystem crates as well.

- [x] Changes are compatible
- [x] Documentation completed
- Tests added and passing
    - [x] Unit Tests
    - [x] Integration Tests
    - [x] Manual Tests


Geal (approved):

abernix:
@jtescher Very appreciative of your work on this!  Thank you so much!  We're going to get this reviewed a bit more and then into our testing environments for observation as soon as we possibly can.  Bear with us while we get some more eyes on this — hopefully by the end of the week.  Cheers!

BrynCooke (approved):

apollo-router/src/plugins/telemetry/tracing/apollo_telemetry.rs
```diff
@@ -128,12 +128,19 @@ struct LightSpanData {
...
 impl From<SpanData> for LightSpanData {
     fn from(value: SpanData) -> Self {
+        let attr_len = value.attributes.len();
         Self {
             span_id: value.span_context.span_id(),
             name: value.name,
             start_time: value.start_time,
             end_time: value.end_time,
-            attributes: value.attributes,
+            attributes: value.attributes.into_iter().fold(
```
garypen, line 137: This looks like an optimization to save us just copying across all the attributes. Is that correct?

apollo-router/src/plugins/telemetry/tracing/apollo_telemetry.rs
```diff
@@ -703,7 +710,10 @@ impl SpanExporter for Exporter {
         // We may get spans that simply don't complete. These need to be cleaned up after a period. It's the price of using ftv1.
         let mut traces: Vec<(String, proto::reports::Trace)> = Vec::new();
         for span in batch {
-            if span.attributes.get(&APOLLO_PRIVATE_REQUEST).is_some()
+            if span
+                .attributes
+                .iter()
+                .any(|attr| attr.key == APOLLO_PRIVATE_REQUEST)
```
garypen, line 716: This looks like you've found and fixed a bug. Thanks!

apollo-router/src/plugins/telemetry/tracing/datadog.rs
```diff
@@ -73,8 +73,10 @@ impl TracingConfigurator for Config {
                     .with_resource_mapping(|span, _model_config| {
                         SPAN_RESOURCE_NAME_ATTRIBUTE_MAPPING
                             .get(span.name.as_ref())
-                            .and_then(|key| span.attributes.get(&Key::from_static_str(key)))
-                            .and_then(|value| match value {
+                            .and_then(|key| {
+                                span.attributes.iter().find(|kv| kv.key.as_str() == *key)
```
garypen, line 77: Thanks again!

garypen (commented):
Thanks very much for creating this PR. I've got a few comments, which it would be nice to have my beliefs confirmed.

apollo-router/src/plugins/telemetry/otlp.rs
```diff
@@ -89,7 +88,6 @@ impl Config {
                 let http = self.http.clone();
                 let exporter = opentelemetry_otlp::new_exporter()
                     .http()
-                    .with_env()
```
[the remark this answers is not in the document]
BrynCooke: Yes, this had already happened with other exporters, but is now in otlp as well.

apollo-router/src/plugins/telemetry/tracing/apollo_telemetry.rs
garypen: @jtescher Could you confirm my assumption? We'd like to merge your PR, but this is the one thing we need to be sure we understand.
I believe you have changed this from the previous `get` because there are multiple attributes which might have the key `APOLLO_PRIVATE_REQUEST`, but would appreciate your confirmation.

apollo-router/src/plugins/telemetry/tracing/apollo_telemetry.rs
jtescher: `value.attributes` is now a `Vec`, so all the `get` calls we do would be much less effecient as `.iter().find(...)`. Could consider a different solution but this was the smallest change.

apollo-router/src/plugins/telemetry/tracing/apollo_telemetry.rs
jtescher: yeah `span.attributes` is now an un-deduped vec, so all need to be checked in `0.21`

garypen (approved):

garypen:
I have created a new PR which incorporates all the work from this PR plus some extra bits which were in transit when this PR was created. I couldn't update the original PR (didn't have the permission to update it), so this was the best option.

New PR is #4248 

I'll close this one. I'll make sure authoring credits are preserved into the new PR.

PR #4211, opened 2023-11-15
🐛 use major versioned v2 instead of _latest which we no longer publish to

mayakoneval:
We recently stopped publishing to the `_latest` tag from the [embeddable-explorer repo](https://github.com/apollographql/embeddable-explorer) so that existing folks on Apollo Server [won't get changes without upgrading](https://github.com/apollographql/embeddable-explorer/pull/280). Router / rover dev still is using `_latest`, but it should switch to use the versioned folder we publish to the cdn now. 

Triggered from Watson's [report on slack](https://apollograph.slack.com/archives/CR62ZA5E0/p1699906382881519)


o0Ignition0o (approved):

mayakoneval:
@o0Ignition0o I can't merge this - will you be able to changelog & merge? Is there a how to for changelog?

o0Ignition0o:
I don't think this warrants a changelog entry, let's go with the merge!

PR #4212, opened 2023-11-15
JWKS Poll Interval

### LINKED ISSUE #4185: make the JWKS refresh interval configurable
It is set to 60s by default, but some IDPs might still throttle it if it checks so often

lleadbet:
*Adding `poll_interval` to JWKS configuration*

Fixes #4185.

Adds a new `poll_interval` configuration option under each JWKS entry to avoid becoming rate-limited per endpoint. 

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

None.


apollo-router/src/plugins/authentication/mod.rs
```diff
@@ -128,6 +128,9 @@ struct JWTConf {
 struct JwksConf {
     /// Retrieve the JWK Set
     url: String,
+    /// Polling interval in seconds for each JWKS endpoint; defaults to 60
+    #[serde(default="default_poll_interval")]
+    poll_interval: u32,
```
bnjjj, line 133: Can we use the humantime duration like we do for other duration we have to fill in the configuration ?

bnjjj (changes requested):

apollo-router/src/plugins/authentication/mod.rs
lleadbet: Sure! Let me look at that and port it.

garypen (approved):

Geal (approved):

bnjjj (approved):

Issue #4214, opened 2023-11-16
Valid non intersecting fragment query panics due to assertion


tinnou:
**Describe the bug**
This assertion in [selection.rs](https://github.com/apollographql/router/blob/ed8ef610a63fa47e570d172406a8b1fc87c5573d/apollo-router/src/spec/selection.rs#L125) does not seem to handle valid cases of a query having non intersecting fragments. It panics in debug builds.

**To Reproduce**
I setup a repro unit test in the [router](https://github.com/apollographql/router/compare/dev...Netflix-Skunkworks:router:non-intersecting-fragments) that panics when parsing the valid query.

**Expected behavior**
There should be no panic. The assertion prevents certain use cases to be executed in non optimized builds. 

**Additional context**
This issue was found when debugging this other [issue](https://github.com/apollographql/federation/issues/2862) in the router. In release builds this assertion is not enabled and the query in that issue would then fail at query planning.

PR #4224, opened 2023-11-17
Chore: Refactor selection set

o0Ignition0o:
Recent compiler updates make a lot of workarounds we have written not necessary anymore.

This change removes them, so we can tidy the code up a bit.

No changelog since this has no user impact.

### DIFF
--- a/apollo-router/src/spec/selection.rs
+++ b/apollo-router/src/spec/selection.rs
@@ -9,7 +9,6 @@ use crate::json_ext::PathElement;
 use crate::spec::query::subselections::DEFER_DIRECTIVE_NAME;
 use crate::spec::query::DeferStats;
 use crate::spec::FieldType;
-use crate::spec::Schema;
 use crate::spec::SpecError;
 use crate::spec::TYPENAME;
 
@@ -44,7 +43,6 @@ impl Selection {
     pub(crate) fn from_hir(
         selection: &executable::Selection,
         current_type: &str,
-        schema: &Schema,
         mut count: usize,
         defer_stats: &mut DeferStats,
     ) -> Result<Option<Self>, SpecError> {
@@ -80,7 +78,6 @@ impl Selection {
                                 Selection::from_hir(
                                     selection,
                                     field_type.0.inner_named_type(),
-                                    schema,
                                     count,
                                     defer_stats,
                                 )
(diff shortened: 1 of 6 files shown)


apollo-router/src/spec/selection.rs
```diff
@@ -113,41 +110,12 @@ impl Selection {
...
-                    fragment_type
-                };
-
                 let selection_set = inline_fragment
                     .selection_set
                     .selections
                     .iter()
                     .filter_map(|selection| {
-                        Selection::from_hir(selection, relevant_type, schema, count, defer_stats)
+                        Selection::from_hir(selection, type_condition.as_str(), count, defer_stats)
```
Geal, line 118: that can't be right. The code we had previously was designed to avoid using the type condition if we already knew a more precise type. How is this avoiding that bug?

o0Ignition0o: sorry it's still a draft while i prepare a test

o0Ignition0o: fwiw the tests are all fully passing, and we did write tests for the type_condition bug we saw, so unless there's a failing test here I don't see it being an issue

SimonSapin: Why is it important or useful to know a more precise type?

---

This PR is not just a chore/refactor, it’s a fix for the bug reported in https://github.com/apollographql/router/issues/4214. The validity rule for fragment spreads or inline fragments is:

https://spec.graphql.org/October2021/#sec-Fragment-spread-is-possible
> * Let _applicableTypes_ be the intersection of [GetPossibleTypes](https://spec.graphql.org/October2021/#GetPossibleTypes())(_fragmentType_) and [GetPossibleTypes](https://spec.graphql.org/October2021/#GetPossibleTypes())(_parentType_)
> * _applicableTypes_ must not be empty.

There are two relevant sets of types, but neither is necessarily a subset of the other. They only need to have some intersection. The logic of this `debug_assert` and of `fn most_precise` makes incorrect assumptions about what’s valid.

SimonSapin:
The code being removed here was added in PR https://github.com/apollographql/router/pull/3718, as a fix for issue https://github.com/apollographql/router/issues/2587. I reproduced the issue in Router 1.29.0 by back-porting the test in the PR. Passing a more precise type affects the `known_type` field in the `spec::selection::Selection`.

This field in turn is used when processing response data in `Query::apply_selection_set` to compute an `is_apply` boolean to decide if a fragment should be skipped because its type condition does not apply. I’m not sure why we’re doing that at that point. In cases like https://github.com/apollographql/router/issues/4214, without requesting `__typename` from subgraphs, I’m not sure the Router always has all the information needed to do this correctly. Shouldn’t the responsibility be left to subgraphs to handle type conditions?

Geal:
should we close this now?

Geal:
closing this because there's been no activity in a while and those parts of the code went through multiple refactorings recently

PR #4226, opened 2023-11-17
Entity cache: configuration

### LINKED ISSUE #4194: Entity caching: configuration
support global and per subgraph configuration. In particular, TTL should be modifiable per subgraph

Geal:
Fix #4194

```yaml title="router.yaml"
experimental_entity_cache:
  redis:
    urls: [ "redis://localhost:6379" ]
    ttl: 60s # Optional, by default no expiration
  enabled: true
  subgraphs:
    accounts:
      enabled: false
    products:
      ttl: 120s
```
**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


bnjjj (approved):
Maybe we should now rename `timeout` to `connection_timeout` to avoid confusion with `ttl` ?

BrynCooke (commented):
Looks like config metrics are missing. 
Will this be in a follow up PR?

Geal:
>Looks like config metrics are missing.
Will this be in a follow up PR?

good point, it comes in https://github.com/apollographql/router/pull/4226/commits/c9c865b39dceb2899b1684bb2e6863dba01e0e8d. Would it be worth it to expose the usage of the ttl field?

Geal:
adding the `ttl` and `enabled` options to metrics in ccce76daa

BrynCooke:
I think we should have some way of specifying a key prefix. Users may be using a redis cluster for several purposes, and being able to have a custom prefix would be beneficial? I could be wrong, and if there is another way of doing this then disregard this.

Geal:
you're right, a key prefix would make sense (for APQ and query planning caches too), I even opend an issue about it yesterday https://github.com/apollographql/router/issues/4247 :)
I'll implement that shortly

apollo-router/src/plugins/cache/entity.rs
```diff
@@ -40,15 +41,44 @@ register_plugin!("apollo", "experimental_entity_cache", EntityCache);
...
+    enabled: Option<bool>,
+    /// Per subgraph configuration
+    #[serde(default)]
+    subgraphs: HashMap<String, Subgraph>,
 }
 
+/// Per subgraph configuration for entity caching
+#[derive(Clone, Debug, JsonSchema, Deserialize)]
+#[serde(rename_all = "snake_case", deny_unknown_fields)]
+struct Subgraph {
```
BrynCooke, line 64: Nit could we get away with serde default at the struct level and implement `Default` for subgraph?

A long TTL and enabled true?

BrynCooke (approved):

apollo-router/src/plugins/cache/entity.rs
Geal: This is one of those cases where the subgraph config has to override the global config, but only for the fields specified in the subgraph, so I can't set default values

PR #4228, opened 2023-11-18
fix(telemetry): fix incomplete traces for apollo_telemetry when combined with a rhai script

bnjjj:
Fix trace report for apollo telemetry when you configured the Router with a rhai script.

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/plugins/telemetry/tracing/apollo_telemetry.rs
+++ b/apollo-router/src/plugins/telemetry/tracing/apollo_telemetry.rs
@@ -1,5 +1,6 @@
 use std::borrow::Cow;
 use std::collections::HashMap;
+use std::collections::HashSet;
 use std::io::Cursor;
 use std::num::NonZeroUsize;
 use std::sync::Arc;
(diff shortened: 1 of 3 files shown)


Geal (approved):

BrynCooke:
@bnjjj 
I'm not sure I understand this, the code should already skip over nodes that it it doesn't understand: 

https://github.com/apollographql/router/blob/7428b772116ef14c48ffd841151018106e3a41fa/apollo-router/src/plugins/telemetry/tracing/apollo_telemetry.rs#L592.

Is it doing something else?

o0Ignition0o:
Can we try to snapshot test it or something? 🥶

bnjjj:
@o0Ignition0o We already have snapshot test, I just updated the configuration to catch that kind of behavior. But you don't see any snapshot difference because it's fixed

o0Ignition0o:
I see so if the code is reverted the snapshot looks different right ? SGTM! 🚀

o0Ignition0o (approved):

bnjjj:
@BrynCooke Yes when returning `child_nodes` at that step it sometimes returns the child_nodes in the wrong order because it's a recursive algorithm. That's why the fix is to either add in front or at the end of the child_nodes if it's an unknown span

BrynCooke (approved):

PR #4254, opened 2023-11-22
refactor(telemetry): use LinkedList instead of HashMap for custom attributes

### LINKED ISSUE #4251: Switch to `Vec` instead of `HashMap` in traits for telemetry custom attributes
We might have better performances by doing this. I think it's worth a try

bnjjj:
Only use `LinkedList` and get rid of `HashMap` for the custom attributes.

Fixes #4251 

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [x] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/plugins/telemetry/formatters/mod.rs
+++ b/apollo-router/src/plugins/telemetry/formatters/mod.rs
@@ -2,7 +2,6 @@
 pub(crate) mod json;
 pub(crate) mod text;
 
-use std::collections::HashMap;
 use std::fmt;
 
 use opentelemetry_sdk::Resource;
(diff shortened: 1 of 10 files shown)


apollo-router/src/plugins/telemetry/formatters/mod.rs
```diff
@@ -109,7 +109,7 @@ pub(crate) fn filter_metric_events(event: &tracing::Event<'_>) -> bool {
     })
 }
 
-pub(crate) fn to_map(resource: Resource) -> HashMap<String, serde_json::Value> {
+pub(crate) fn to_vec(resource: Resource) -> LinkedList<(String, serde_json::Value)> {
```
garypen, line 112: I know you mean `to_list`...

garypen (approved):
One suggested naming change, but looks nice.

apollo-router/src/plugins/telemetry/formatters/mod.rs
bnjjj: Oops, it has been merged, here is the fix https://github.com/apollographql/router/pull/4257

PR #4255, opened 2023-11-22
test the impact of merging integration tests

Geal:
integration tests require linking an entire router with each of them, which can significantly increase build times, along with CPU and RAM usage.
This is an experiment where I merge:
- metrics_tests and tracing_test in telemetry_test
- rhai_test and validation_test in integration_test

**This is the impact of building the tests after a modification deep in the router, before those changes:**

timing:
Screenshot from 2023-11-22 15-06-35

CPU and RAM usage:
Screenshot from 2023-11-22 14-52-03

**Now with the changes in this PR:**

timing:

Screenshot from 2023-11-22 15-07-10

CPU and RAM usage:
Screenshot from 2023-11-22 15-01-50

**Significant differences:**
- build time drops from 72s to 58s, that's 19%
- RAM usage was rising from ~40% to more than 90% during build (at times some process had to be OOM killed), now it barely gets above 75% (I have 64GB of RAM)
- All CPU threads were used during the entire build, now some of them get some quiet time

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


Geal:
with [7e38f59](https://github.com/apollographql/router/pull/4255/commits/7e38f5945d93465f5efdc5115899d8a90cd39db5) I get the following:

timing:
Screenshot from 2023-11-22 15-58-57

CPU and RAM:
Screenshot from 2023-11-22 15-57-27

By moving the files to a submodule, I can preserve their shape and history

Geal:
(I force pushed to rewrite the initial test merge commits as moving under submodules)

.circleci/config.yml
```diff
@@ -63,7 +63,7 @@ executors:
   windows_test: &windows_test_executor
     machine:
       image: "windows-server-2019-vs2019:stable"
-    resource_class: windows.xlarge
+    resource_class: windows.large
```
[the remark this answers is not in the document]
Geal: I was checking if this change could reduce the size of our CI runners, since we use less RAM to build all the tests. That only worked for Windows though, the other runners still OOM'd when linking

Geal:
>On the flip side though, the main issue that at least hurts my development is compiling the unit tests. For this cargo seems to be smart enough to not bother with the integration tests at all.

yup, I have no great solution for unit tests so far. https://github.com/apollographql/router/pull/4204 and related PRs have the side effect that changing the tests does not require recompiling the entire service (and checking with everything that depends on it), but the improvement was barely noticeable.
We could gain some time by not compiling some doc tests though, since each doc tests is a new binary to build

bnjjj (approved):

o0Ignition0o (approved):

.circleci/config.yml
Geal, line 66:
```suggestion
    resource_class: windows.xlarge
```

apollo-router/tests/integration/subscription.rs
```diff
@@ -1,15 +1,15 @@
...
 use futures::StreamExt;
 use http::HeaderValue;
 use serde_json::json;
 use tower::BoxError;
 
-use crate::common::IntegrationTest;
-use crate::common::Telemetry;
-
+#[path = "../common.rs"]
 mod common;
```
SimonSapin, line 10: why not move this file too?

Issue #4215, opened 2023-11-16
TLS test (tls_self_signed) often fails in CI


garypen:
For instance: https://app.circleci.com/pipelines/gh/apollographql/router/17280/workflows/0383afbf-7eb5-4fa2-88a0-21d33ff4b5d0/jobs/112957

I think the cause is when we deserialize a Response. Something like in these example is happening:

```
{ "data" = null, "errors" = [] }
```
vs
```
{ "errors" = [] }
```

The first (when deserialized) would give `Response.data=Some(Value::Null}`, the latter would give `Response.data=None`.

I don't know what is causing this and it's odd that it's an intermittent fail.

I think this is quite a high priority issue because the root cause of the intermittent fail could be quite serious.

PR #4297, opened 2023-11-28
Identify and remove race condition in subgraph tests

garypen:
Tokio TcpListener enables SO_REUSEADDR which means our tests are not as well isolated as we would like.

The fix is to ensure that tests using Tokio TcpListener are executed serially.

fixes: #4215

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/services/subgraph_service.rs
+++ b/apollo-router/src/services/subgraph_service.rs
@@ -1207,6 +1207,7 @@ mod tests {
     use rustls::ServerConfig;
     use serde_json_bytes::ByteString;
     use serde_json_bytes::Value;
+    use serial_test::serial;
     use tokio::sync::mpsc;
     use tokio_stream::wrappers::ReceiverStream;
     use tower::service_fn;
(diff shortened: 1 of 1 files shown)


Geal:
those tests do not use a static listener port, they all bind on 127.0.0.1:0 to get a new random port for the test, I don't see how SO_REUSEADDR would affect them

garypen:
> those tests do not use a static listener port, they all bind on 127.0.0.1:0 to get a new random port for the test, I don't see how SO_REUSEADDR would affect them

My theory is based on observations that:
 - these tests do fail intermittently in CI and it looks like the wrong data is being read (e.g. as though from the wrong server).
 - it is always a Tokio TcpListener based test that fails and usually involving TLS
 - it's possible that something like [this](https://stackoverflow.com/questions/3229860/what-is-the-meaning-of-so-reuseaddr-setsockopt-option-linux) is going on (especially point 5)

However, I agree that it's not a categorical root cause of the failure.

Geal:
this needs more digging, because binding multiple TCP listeners on `127.0.0.1:0` ending up on the same address feels like it breaks a fundamental assumption of how these sockets work. Could you try wrapping the creation of the sockets so it does not have SO_REUSEADDR and SO_REUSEPORT? That would show more clearly if that's the issue

garypen:
I don't have time to do the digging you suggested. Especially because it usually only manifests on CI/linux and I don't have a local linux box for testing.

I don't think there's a downside to merging this, but if you want to spend more time working on this, then you should. I'm happy to accept that bad decisions have been made about TCP/TLS design in the past which are causing us testing issues that this may mitigate against.

garypen:
I'm going to close this since it seems unlikely this is the cause and if there is an issue, then we should try and fix it.

PR #4299, opened 2023-11-28
Execution level coprocessor

### LINKED ISSUE #3948: Support the policy directive with subscriptions
`@policy` relies on executing policies externally in coprocessors, at the supergraph service level. Some of the tested policies might return a result that varies with time (example: is user X part of team Y), so they should be tested every time a subscription is triggered.
Subscriptions restart from the execution service level, so a coprocessor at the execution service level that checks the policy results in context are still valid. There is no execution service coprocessor available now though

Geal:
Fix #3948

This introduces a new coprocessor stage intevening at the execution service level: after the supergraph stage, after query planning is done, before executing the query plan.

This exposes the query plan to the coprocessor but does not make it modifiable

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


docs/source/customizations/coprocessor.mdx
```diff
@@ -39,16 +39,17 @@ flowchart TB;
...
+  subgraphService <-."<b>8.</b> Can send request<br/>details to coprocessor<br/>and receive modifications".-> coprocessing;
+  subgraphService -- "<b>9</b>"--> subgraphs;
 
   class client,subgraphs,coprocessing secondary;
 ```
 
 This diagram shows request execution proceeding "down" from a client, through the router, to individual subgraphs. Execution then proceeds back "up" to the client in the reverse order.
 
-As shown in the diagram above, the `RouterService`, `SupergraphService` and `SubgraphService` steps of the [request-handling lifecycle](./rhai/#router-request-lifecycle) can send these POST requests (also called **coprocessor requests**).
+As shown in the diagram above, the `RouterService`, `SupergraphService`, `ExecutionService` and `SubgraphService` steps of the [request-handling lifecycle](./rhai/#router-request-lifecycle) can send these POST requests (also called **coprocessor requests**).
```
shorgi, line 52:
```suggestion
As shown in the diagram above, the `RouterService`, `SupergraphService`, `ExecutionService`, and `SubgraphService` steps of the [request-handling lifecycle](./rhai/#router-request-lifecycle) can send these POST requests (also called **coprocessor requests**).
```

apollo-router/src/services/external.rs
```diff
@@ -168,6 +173,47 @@ where
             method,
             service_name: None,
             has_next,
+            query_plan: None,
+        }
+    }
+
+    #[builder(visibility = "pub(crate)")]
+    /// This is the constructor (or builder) to use when constructing a Supergraph
```
garypen, line 181:
```suggestion
    /// This is the constructor (or builder) to use when constructing an Execution
```

docs/source/customizations/coprocessor.mdx
```diff
@@ -298,7 +299,115 @@ Properties of the JSON body are divided into two high-level categories:
 {
   // Control properties
   "version": 1,
-  "stage": "SupergraphRequest",
+  "stage": "SupergraphResponse",
```
garypen, line 302: I can't tell from the diff if this is correct, but it seems odd that `SupergraphRequest` is now `SupergraphResponse`. Is this fixing an existing problem?

garypen (dismissed):
NIT and question

docs/source/customizations/coprocessor.mdx
Geal: the section title is about the `SupergraphResponse` stage, but it looks like the response is wrong, because we would not have a `control` field, right?

apollo-router/src/plugins/coprocessor/execution.rs
```diff
@@ -0,0 +1,1011 @@
...
+                    )
+                    .await
+                    .map_err(|error| {
+                        succeeded = false;
+                        tracing::error!(
+                            "external extensibility: execution request stage error: {error}"
+                        );
+                        error
+                    });
+                    tracing::info!(
```
BrynCooke, line 110: Please use the new metrics macros rather than the tracing macros. I'm going to add a safelist to the tracing metrics macros to prevent further use within our codebase.

BrynCooke: In addition using the new macros will allow you to add tests for the metrics.

BrynCooke (changes requested):

apollo-router/src/plugins/coprocessor/execution.rs
BrynCooke: Docs for new metrics macros are here: https://github.com/apollographql/router/blob/dev/dev-docs/metrics.md#using-metrics-macros

Geal: why is the dev docs using the format `u64_counter!("test", "test description", 1, "attr" => "val");` while the macro is using `u64_counter!("test", "test description", 1, "attr" = "val");` (with `=` instead of `=>`)?
Could you check the docs again for inconsistencies?

Geal: fixed in [40711d1](https://github.com/apollographql/router/pull/4299/commits/40711d15f38811ccde2bff23412fa8354e620339)
It would be nice to have the annotations tracing provided, like the % prefix when adding attribute values

docs/source/customizations/coprocessor.mdx
Geal: there's actually a control field on the supergraph response from coprocessor, to set the status

Geal:
if there isn't anything else to report, I will manually dismiss the "changes requested" reviews and merge this, as this is blocking further work on authorization with subscriptions

PR #4309, opened 2023-11-30
chore(deps): update apollo-parser

goto-bus-stop:
Contains a fix for a critical regression in 0.7.3. That version was yanked from crates.io but it already made it to our `Cargo.lock` for 1.34.1 and the 1.35.0 release candidate.

This should be part of 1.35.0


goto-bus-stop:
Or should i target this to 1.35.0 release branch so it doesn't have to be cherrypicked onto that?

abernix (approved):
LGTM! 🙇 

@goto-bus-stop That's the right question to ask!  In this case, there's nothing else on `dev` right now.  If our timing is right with merging this, we'll be good keeping it where we are.  To facilitate that, I'll merge this immediately! 😄

PR #4278, opened 2023-11-25
Removing from template the `/` as it will always come with the `/` as…

juancarlosjr97:
Removing from the template the `/` as it will always come with the `/` as gives and incorrect health check URL the logs

*Description here*

Fixes https://github.com/apollographql/router/issues/4270

---

**Checklist**

Complete the checklist (and note appropriate exceptions) before the PR is marked ready-for-review.

- [x] Changes are compatible[^1]
- [x] Documentation[^2] completed
- [ ] Performance impact assessed and acceptable
- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[^1]: It may be appropriate to bring upcoming changes to the attention of other (impacted) groups. Please endeavour to do this before seeking PR approval. The mechanism for doing this will vary considerably, so use your judgement as to how and when to do this.
[^2]: Configuration is an important part of many changes. Where applicable please try to document configuration examples.
[^3]: Tick whichever testing boxes are applicable. If you are adding Manual Tests, please document the manual testing (extensively) in the Exceptions.

Issue #4270, opened 2023-11-24
Apollo health check returns in the logs double // as http://127.0.0.1:8088//health on v1.34.1


juancarlosjr97:
**Describe the bug**

When updating from v1.34.0 to v1.34.1 the health check works as expected on the endpoint `/health` but the logs returns the health check in the log as `http://127.0.0.1:8088//health` with the `/`

**To Reproduce**

1. Dockerfile as follow
  ```dockerfile
  FROM ghcr.io/apollographql/router:1.34.1@sha256:5ea0267179104f79ef0b4109be90b16504a40efe0030b1d4957782542c7da682
  
  WORKDIR /dist/config  
  COPY ./router.yaml ./router.yaml
  
  EXPOSE 4000
  ```

2. Build the image and run federated apollo graphql

```bash
docker build -t router .

export APOLLO_KEY="..."
export APOLLO_GRAPH_REF="..."

docker run -e APOLLO_GRAPH_REF=${APOLLO_GRAPH_REF} -e APOLLO_KEY=${APOLLO_KEY} -p 4100:4100 -p 8088:8088 router
```

3. See the logs returning the health check with an additional `/`

```bash
*router  | {"timestamp":"2023-11-24T14:50:55.547861Z","level":"INFO","message":"Apollo Router v1.34.1 // (c) Apollo Graph, Inc. // Licensed as ELv2 (https://go.apollo.dev/elv2)","target":"apollo_router::executable"}
*router  | {"timestamp":"2023-11-24T14:50:55.548101Z","level":"INFO","message":"Anonymous usage data collection is disabled.","target":"apollo_router::executable"}
*router  | {"timestamp":"2023-11-24T14:50:55.737117Z","level":"INFO","message":"Health check exposed at http://127.0.0.1:8088//health"}
*router  | {"timestamp":"2023-11-24T14:50:55.737722Z","level":"INFO","message":"GraphQL endpoint exposed at http://0.0.0.0:4100/ 🚀"}
```

**Expected behavior**

The health check in the logs to return `http://127.0.0.1:8088/health`

**Desktop (please complete the following information):**
 - OS: Docker - MacOS M1
 - Version: 1.34.1

PR #4322, opened 2023-12-04
fix the changelog for the 1.35.0 release

garypen:
Add an entry for the health check logging contribution.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,12 @@ By [@bnjjj](https://github.com/bnjjj), [@bryncooke](https://github.com/bryncooke
 
 ## 🐛 Fixes
 
+### Remove surplus '/' from health check in logs ([Issue #4270](https://github.com/apollographql/router/issues/4270))
+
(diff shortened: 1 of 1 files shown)


abernix:
For posterity, I'm adding this comment to create the reference to the PR which introduced the change (https://github.com/apollographql/router/pull/4278) and the issue it closed, https://github.com/apollographql/router/issues/4270.  Thanks for your contribution, @juancarlosjr97.  The omission of the contribution in the release notes was purely because the log wasn't added initially in the PR and depending on the change, we don't always insist on changelogs from contributors.  We have a set of instructions on [how to add changelogs in case you're interested](https://github.com/apollographql/router/tree/dev/.changesets), but I recognize the contribution guidelines didn't naturally lead you to that and the failing commit on the PR didn't do much to tell you _how_ to fix it. 😄 .

Thanks again though!

juancarlosjr97:
Awesome! Already know it for the next time! 

What do you think of also updating the PR template by adding CHANGELOG to the list @abernix?

PR #4328, opened 2023-12-05
Fix generation of subscription traces for Apollo reporting

bonnici:
Fixes the generation of subscription traces that are sent to Apollo usage reporting.

Fixes https://github.com/apollographql/router/issues/4339

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [x] Manual Tests

**Exceptions**

I performed manual testing by starting up a subscription-enabled subgraph and publishing a schema with a subscription operation. I then sent a subscription request using Curl:
```
curl 'http://localhost:4000/' -v \
  -H 'accept: multipart/mixed; boundary="graphql"; subscriptionSpec=1.0, application/json' \
  -H 'content-type: application/json' \
  --data-raw '{"query":"subscription ExampleSubFixedRouter { numberIncremented }"}'
```

And the subscription trace was visible in studio: https://studio-staging.apollographql.com/graph/Nick-Subscription-Test/variant/current/insights?insightsTab=operations&query=a4ccc61b2546065c4c3c26748b3913d81c395a72&queryName=ExampleSubFixedRouter&range=lastHour&tab=traces&trace=Nick-Subscription-Test~2023%3A12%3A06T05%3A44~efa80f651708d467b6060a0e4496cdef963fa90c31d65f4095e3e65cf4489093&traceBucket=115

SCR-20231206-nxjf

There is still an open question about whether or not the trace should end after the initial subscription registration rather than ending when the subscription was terminated. The subgraph span in the subscription request also is currently missing FTv1 data even with tracing set to `always_on`, so we are also missing the subgraph timing data in the trace. But with this fix, we at least get a visible trace in studio and it matches the OTel version of the trace.

**Notes**


bonnici:
I'm putting this back in WIP since I just noticed we don't get FTV1 spans for subscriptions even with this change.

bonnici:
Making the PR ready for review again because I confirmed that the `apollo_private.ftv1` attribute is missing on the subgraph fetch span with subscriptions, so the missing subgraph timing is a larger issue that isn't related to the construction of apollo protobuf traces and that's another thing we'd need to fix in a separate issue.

bnjjj (approved):

bnjjj:
FYI @bonnici the ftv1 attribute can be easily added [here](https://github.com/apollographql/router/blob/dev/apollo-router/src/query_planner/execution.rs#L205:L205) and to fetch it I think you can add code [here](https://github.com/apollographql/router/blob/dev/apollo-router/src/query_planner/subscription.rs#L272C1-L272C1) to get data from `response.body().extensions` and directly add it in span attributes using something like `Span::current().record("apollo_private.ftv1", ftv1.as_str());` . I think you can even re-use this function https://github.com/apollographql/router/blob/dev/apollo-router/src/plugins/telemetry/mod.rs#L1882:L1882

bonnici:
@bnjjj thank you! I'll add that info as a note to the follow-up story. I timeboxed a bit of time to try to add it to this PR but got a bit lost in the code.

PR #4332, opened 2023-12-05
Follow up to the active session count metric fix

Geal:
Follow up to #3787 

Fix atomic ordering when calculating the active session count
~we need to figure out why the active session metric does not appear in the tests, and why it keeps increasing in production deployments~

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


garypen:
Did I share that I figured out that both counts disappear (for some reason) from the prom stats if the value returns to 0? I saw that when I was running manual testing last week.

Geal:
that would make sense. Then maybe I can make a test where I start a query then right in the middle of it I check the metrics

Geal:
I'm removing the failing test again. This should still be merged, because updating the ordering will likely fix the issues we saw with that metric when running in production

bnjjj (approved):

PR #4353, opened 2023-12-08
Make observable metrics clean up once they are dropped

### LINKED ISSUE #4352: Aggregated gauges leak memory
When a gauge is registered via meter provider any callback will be registered with the underlying pipeline. This can never be dropped, so in a reload scenario where the gauge is reconstructed the existing callback is still in the pipeline.

BrynCooke:
When observable metrics are created any callbacks are registered with the metrics pipeline. They are never cleaned up. This means that we may slowly leak memory if this feature is ever used.

Currently the metrics layer does NOT use this and instead calls `observe` directly on gauges. This means that we don't currently leak, but this shouldn't actually work as according to the rust docs for `observe`:

```
Records the state of the instrument.
It is only valid to call this within a callback. If called outside of the registered callback it should have no effect on the instrument, and an error will be reported via the error handler.
```

Fixing this means that we get rid of the thread that is constantly refreshing gauges between exports.

Fixes #4352

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/metrics/aggregation.rs
+++ b/apollo-router/src/metrics/aggregation.rs
@@ -214,12 +214,12 @@ impl<T: Copy> SyncCounter<T> for AggregateCounter<T> {
 }
 
 pub(crate) struct AggregateObservableCounter<T> {
-    delegates: Vec<ObservableCounter<T>>,
+    delegates: Vec<(ObservableCounter<T>, Option<DroppingUnregister>)>,
 }
 
 impl<T: Copy> AsyncInstrument<T> for AggregateObservableCounter<T> {
     fn observe(&self, value: T, attributes: &[KeyValue]) {
-        for counter in &self.delegates {
+        for (counter, _) in &self.delegates {
             counter.observe(value, attributes)
         }
     }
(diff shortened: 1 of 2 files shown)


apollo-router/src/metrics/aggregation.rs
```diff
@@ -214,12 +214,12 @@ impl<T: Copy> SyncCounter<T> for AggregateCounter<T> {
 }
 
 pub(crate) struct AggregateObservableCounter<T> {
-    delegates: Vec<ObservableCounter<T>>,
+    delegates: Vec<(ObservableCounter<T>, Option<DroppingUnregister>)>,
```
bnjjj, line 217: Maybe a very dumb question but why is it not included directly in the `ObservableCounter` ? Like only having a single type responsible to the drop impl ?

BrynCooke: We can take the suggestion back to Otel rust. I agree that it seems hard to clean observables.
Also would want to suggest that noop instrument not give the cryptic error message when passed to register_callback.

o0Ignition0o (approved):

PR #4360, opened 2023-12-11
fix: update telemetry config in helm chart

lennyburdette:
This patch updates the generated router config to avoid a "router configuration contains deprecated options" warning on router startup.

I tested this change against the in house router's chart in three ways:

* As it currently is, using the deprecated shape. The results are the same and the router can start up with it.
* Without any telemetry at all. The resulting config no longer contains `telemetry`.
* After updating telemetry configuration to the new structure. The resulting config is correct and the router no longer spits out deprecation warnings.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/helm/chart/router/templates/configmap.yaml
+++ b/helm/chart/router/templates/configmap.yaml
@@ -1,6 +1,6 @@
 {{- if .Values.router.configuration }}
 {{- $routerFullName := include "router.fullname" .  -}}
-{{- $configuration := dict "telemetry" (dict "metrics" (dict "common" (dict "resources" (dict "service.name" $routerFullName)))) -}}
+{{- $configuration := dict "telemetry" (dict "exporters" (dict "metrics" (dict "common" (dict "resource" (dict "service.name" $routerFullName))))) -}}
 {{- $_ := mustMergeOverwrite $configuration .Values.router.configuration  -}}
 
 apiVersion: v1
--- a/helm/chart/router/templates/configmap.yaml
+++ b/helm/chart/router/templates/configmap.yaml
@@ -1,9 +1,11 @@
 {{- if .Values.router.configuration }}
 {{- $routerFullName := include "router.fullname" .  -}}
 {{/* NOTE: metrics configuration moved under telemetry.exporters in Router 1.35.0 */}}
-{{- $configuration := dict "telemetry" (dict "metrics" (dict "common" (dict "resources" (dict "service.name" $routerFullName)))) -}}
-{{- if (and .Values.router.configuration.telemetry .Values.router.configuration.telemetry.exporters) }}
+{{- $configuration := dict }}
+{{- if (.Values.router.configuration.telemetry).exporters }}
 {{- $configuration = dict "telemetry" (dict "exporters" (dict "metrics" (dict "common" (dict "resource" (dict "service.name" $routerFullName))))) -}}
+{{- else }}
+{{- $configuration := dict "telemetry" (dict "metrics" (dict "common" (dict "resources" (dict "service.name" $routerFullName)))) -}}
 {{- end }}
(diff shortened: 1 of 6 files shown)


lennyburdette:
@garypen if someone has legacy config under `telemetry.metrics` (without `.exporters`), the results will be 
```yaml
telemetry:
  exporters:
    metrics:
      common:
        resource:
          service.name: router
  metrics:
    common:
       # my own config
```
that's not going to work right? how do we version this change to the helm chart?

garypen (changes requested):
My concern is that this change this will break everyone using the chart who hasn't got updated configuration.

We've been modifying configuration for the router without concern to compatibility using the "migrations" capability, but helm doesn't have such a capability.

We'd either have to:
 - Modify the chart so that it supports both "flavours" of configuration
 - Make this an incompatible change.

I think the former is the least worse solution for now. If we go with option 2 we'd have to clearly document that in our release notes.

helm/chart/router/templates/configmap.yaml
```diff
@@ -1,6 +1,10 @@
 {{- if .Values.router.configuration }}
```
garypen, line 1: I'm unsure if this set of changes is correct. If I run 
```
helm template --dry-run --debug helm/chart/router
```
then I get a set of templates which include a ConfigMap which looks like this:
```
data:
  configuration.yaml: |
    health_check:
      listen: 0.0.0.0:8088
    supergraph:
      listen: 0.0.0.0:4000
    telemetry:
      metrics:
        common:
          resources:
            service.name: release-name-router
```
That surprised me a little, since I wouldn't expect to get any telemetry configuration if I haven't specified any.

I modified the ConfigMap template like this:
```
{{/* NOTE: metrics configuration moved under telemetry.exporters in Router 1.35.0 */}}
{{- $configuration := dict }}
{{- if (.Values.router.configuration.telemetry).exporters }}
{{- $configuration = dict "telemetry" (dict "exporters" (dict "metrics" (dict "common" (dict "resource" (dict "service.name" $routerFullName))))) -}}
{{- else }}
{{- $configuration := dict "telemetry" (dict "metrics" (dict "common" (dict "resources" (dict "service.name" $routerFullName)))) -}}
{{- end }}
```
and that gave me a ConfigMap that looks like this:
```
data:
  configuration.yaml: |
    health_check:
      listen: 0.0.0.0:8088
    supergraph:
      listen: 0.0.0.0:4000
```
which is what I would expect.

What I don't fully understand is why your change is always generating a dict which contains a "service.name". 

My theory is that this logic was always broken but "worked" because we always specified some telemetry information in the values.yaml file (which you have now removed).

Let's talk this through when you've had a chance to read through this.

garypen (changes requested):
Most of this looks good. I think we just need to understand the impact on the configmap template of removing the default telemetry data from values.yaml.

helm/chart/router/templates/configmap.yaml
lennyburdette: This all makes sense, I've included your suggested change.

I tested this change against the in house router's chart in three ways:
* As it currently is, using the deprecated shape. The results are the same and the router can start up with it
* Without any telemetry at all — the resulting config no longer contains `telemetry`.
* After updating telemetry configuration to the new structure — the resulting config is correct and the router no longer spits out deprecation warnings

> My theory is that this logic was always broken but "worked" because we always specified some telemetry information in the values.yaml file (which you have now removed).

I think this is correct

garypen (approved):
I don't think this needs a changeset, but if you think one is required, then I guess it's a "maintenance" changeset.

It's worth extracting your comments about how you tested and documenting that in the PR checklist.

I'll approve it as is.

PR #4372, opened 2023-12-13
Move comparative validation metric into `apollo.router.operations` namespace

goto-bus-stop:
Here, it'll actually be handled by our analytics infrastructure.

Follow-up for #4159


abernix (approved):
LGTM.  No changelog required.  No docs required.

lrlna (approved):

PR #4389, opened 2023-12-14
prep release: v1.36.0-rc.3

lennyburdette:
> **Note**
>
> When approved, this PR will merge into **the `1.36.0-rc.3` branch** which will — upon being approved itself — merge into `main`.
>
>  - Version bumps
>  - That it targets the right release branch (`1.36.0-rc.3` in this case!).
>
---


abernix:
I don't think we need this anymore.

PR #4394, opened 2023-12-15
perf(telemetry): small improvements for telemetry

bnjjj:


apollo-router/src/plugins/telemetry/dynamic_attribute.rs
```diff
@@ -113,6 +113,10 @@ impl DynAttribute for ::tracing::Span {
     }
 
     fn set_dyn_attributes(&self, attributes: impl IntoIterator<Item = KeyValue>) {
+        let mut attributes = attributes.into_iter().peekable();
+        if attributes.peek().is_some() {
```
o0Ignition0o, line 117: is none ? ^^'

Geal (approved):

PR #4395, opened 2023-12-15
[multipart-subscriptions-protocol.md] Let the server decide the boundary

martinbonnin:
Remove the multipart boundary from the client `Accept` header. Typically, the boundary is decided by the server in order to avoid nameclashes with the parts contents. It just happens that `--graphql` works all the time for us because the part contents are JSON.

Keep `--graphql` as a "may" for convenience but mandate that clients support any boundary sent by the server. I _think_ this is already the case for web/ios/kotlin? Ping @alessbell @calvincestari 

ping @bnjjj @abernix

### DIFF
--- a/dev-docs/multipart-subscriptions-protocol.md
+++ b/dev-docs/multipart-subscriptions-protocol.md
@@ -4,10 +4,10 @@ Instead of relying on WebSockets, the subscriptions protocol supported by the ro
 
 ## Communication
 
-When sending a request containing a subscription to the router, clients MUST support the `multipart/mixed; subscriptionSpec="1.0"` content-type in addition to the `application/graphql-response+json` content-type:
+When sending a request containing a subscription to the router, clients MUST support the `multipart/mixed; subscriptionSpec="1.0"` content-type in addition to the `application/json` content-type:
 
 ```
-Accept: multipart/mixed; subscriptionSpec="1.0", application/graphql-response+json
+Accept: multipart/mixed; subscriptionSpec="1.0", application/json
 ```
 
 The router will then respond with a stream of body parts, following the [definition of multipart content specified in RFC1341](https://www.w3.org/Protocols/rfc1341/7_2_Multipart.html).
--- a/dev-docs/multipart-subscriptions-protocol.md
+++ b/dev-docs/multipart-subscriptions-protocol.md
@@ -4,12 +4,11 @@ Instead of relying on WebSockets, the subscriptions protocol supported by the ro
 
 ## Communication
 
-When sending a request containing a subscription to the router, clients should include the following `Accept` header to indicate their support for the multipart subscriptions protocol:
-```
-Accept: multipart/mixed; boundary="graphql"; subscriptionSpec="1.0", application/json
-```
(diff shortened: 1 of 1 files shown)


dev-docs/multipart-subscriptions-protocol.md
```diff
@@ -4,13 +4,12 @@ Instead of relying on WebSockets, the subscriptions protocol supported by the ro
 
 ## Communication
 
-When sending a request containing a subscription to the router, clients should include the following `Accept` header to indicate their support for the multipart subscriptions protocol:
+When sending a request containing a subscription to the router, clients MUST support the `multipart/mixed; subscriptionSpec="1.0"` content-type in addition to the `application/graphql-response+json` content-type:
+
 ```
-Accept: multipart/mixed; boundary="graphql"; subscriptionSpec="1.0", application/json
+Accept: multipart/mixed; subscriptionSpec="1.0", application/graphql-response+json
```
martinbonnin, line 10: I also changed the non-subscription mime type to `application/graphql-response+json` as recommended by [GraphQL over HTTP](https://github.com/graphql/graphql-over-http/blob/main/spec/GraphQLOverHTTP.md). Happy to revert that change and make it part of another discussion.

bnjjj: BTW we have to still be backward compatible with what's existing right now to not break existing implementations

martinbonnin: > why not both 

I'd this spec should be as strict as possible to avoid confusion. If the router supports `application/graphql-response+json` then use that one. If not, let's stick with `application/json` and wait before changing it.

Clients can always decide to send more if they want.

abernix: `application/graphql-response+json` isn't official yet, so let's hold off.

bnjjj: In any cases I think we won't be able to be that strict if we don't want to introduce breaking changes

martinbonnin: > let's hold off.

Sone in https://github.com/apollographql/router/pull/4395/commits/8cdd3b081830bf421a4d9fe4d3a7dfef2cd10e27

calvincestari:
> mandate that clients support any boundary sent by the server. I think this is already the case for web/ios/kotlin?

This is already the case with the iOS client.

alessbell:
Yes, this is already the case for the web client 👍

abernix (changes requested):
Thanks everyone for jumping in here.  We need to update assertions internal to the Router in addition to this change to the `dev-docs/` before landing this.   @bnjjj Can you action that and re-request ♻️  me for review please?

bnjjj (approved):

bnjjj:
cc @abernix , following a discussion with @martinbonnin I suggested to push his changes to another custom branch to let me update internal assertions of the router in the same commit on `dev` branch. I also suggested to update the requirements and docs for `@defer` as well.

bnjjj:
Merged in another branch containing code changes #4447

PR #4401, opened 2023-12-18
fix fragment spread response formatting with aliased typename #4348

### LINKED ISSUE #4348: Router not GraphQL spec compliant with Abstract spreads in object scopes
The functionality described in [section 5.5.2.3.2 of the GraphQL Spec](https://spec.graphql.org/October2021/#sec-Abstract-Spreads-in-Object-Scope), Abstract spreads within an object scope do not appear to work in Apollo Router. All examples shown below are with the published Apollo Router executable **v1.35.0**

## Unions with no nested types
Using a schema as follows, a union is made on two objects, Cat and Dog. Both objects are similar to one another, and a field under query is created to return a Dog object

```graphql
type Query {
  dog: Dog
}

type Dog {
  name: String!
  nickname: String!
  barkVolume: Int
}

type Cat {
  name: String!
  nickname: String!
  meowVolume: Int
}

union CatOrDog = Cat | Dog
```

The following is the response from a query to a [gqlgen](https://gqlgen.com/getting-started/) server, that specifies the root level dog field and spreads the union CatOrDog under it.

image

The Apollo Router returns a similar response for the same schema, but adds on the field unique to the Cat type, meowVolume to the response payload even though it does not exist on type Dog

image

**Expected behavior**
The router should not be sending back `meowVolume` in this case.

## Union with one nested type
The schema is modified to instead nest Dog under an arbitrary wrapper object called NestedDog. The union is now on Cat and NestedDog

```graphql
type Query {
  nestedDog: NestedDog
}

type NestedDog {
  dog: Dog!
}

type Dog {
  name: String!
  nickname: String!
  barkVolume: Int
}

type Cat {
  name: String!
  nickname: String!
  meowVolume: Int
}

union CatOrDog = Cat | NestedDog
```

The query was modified accordingly, and making the query below to a gqlgen server results in the following response:

image

However, the Apollo Router seems to completely null out the Dog type for the same query:

image

Removing the spread on the Cat object fixes this issue and allows the response to pass through successfully

image

## Additional Context
The request sent to the subgraph (gqlgen server) and response received from the subgraph were logged using a rhai script (this holds for both queries tested under the union with nested type):

**Subgraph Request:**
```graphql
query test__example__0 {
  nestedDog {
    __isCatOrDog: __typename
    dog {
      name
      nickname
      barkVolume
    }
  }
}

```
**Subgraph Response:**
```json
{
 "data": {
   "nestedDog": {
     "__isCatOrDog": "NestedDog",
     "dog": {
       "barkVolume": 7,
       "name": "Fido",
       "nickname": "Fi"
     }
   }
 }
}

```

bnjjj:
When you had an aliased `__typename` (example `myAlias: __typename`) in a fragment spread, the response formatting was wrong.

Fixes #4348

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/.changesets/fix_bnjjj_fix_4348.md
+++ b/.changesets/fix_bnjjj_fix_4348.md
@@ -1,5 +1,5 @@
-### Fix fragment spread response formatting with aliased typename ([Issue #4348](https://github.com/apollographql/router/issues/4348))
+### Fix fragment spread response formatting ([Issue #4348](https://github.com/apollographql/router/issues/4348))
 
-When you had an aliased `__typename` (example `myAlias: __typename`) in a fragment spread, the response formatting was wrong.
+When you had an inline fragment on an union type in a fragment spread, the response formatting was wrong.
 
 By [@bnjjj](https://github.com/bnjjj) in https://github.com/apollographql/router/pull/4401
--- a/.changesets/fix_bnjjj_fix_4348.md
+++ b/.changesets/fix_bnjjj_fix_4348.md
@@ -1,0 +1,5 @@
+### Fix fragment spread response formatting with aliased typename ([Issue #4348](https://github.com/apollographql/router/issues/4348))
+
+When you had an aliased `__typename` (example `myAlias: __typename`) in a fragment spread, the response formatting was wrong.
+
+By [@bnjjj](https://github.com/bnjjj) in https://github.com/apollographql/router/pull/4401
--- a/apollo-router/src/query_planner/tests.rs
+++ b/apollo-router/src/query_planner/tests.rs
@@ -1458,9 +1458,7 @@ async fn typename_propagation2() {
                         "representations": [{
                             "__typename": "Book",
                             "bookId": "book1",
-                            "author": {
-                                "fullName": "Ada"
-                            }
+                            "author": null
                         }]
                     }
                 }},
--- a/apollo-router/src/query_planner/tests.rs
+++ b/apollo-router/src/query_planner/tests.rs
@@ -1426,10 +1426,7 @@ async fn typename_propagation() {
     insta::assert_json_snapshot!(serde_json::to_value(&response).unwrap());
 }
 
-// this test cannot pass as it is right now because we cannot fill in __typename yet without
-// large manipulations using the compiler
 #[tokio::test]
-#[should_panic]
 async fn typename_propagation2() {
     let subgraphs = MockedSubgraphs(
         [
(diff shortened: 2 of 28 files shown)


abernix:
Just noting that https://github.com/apollographql/router/pull/4402 targets this right now. (creating context)

apollo-router/src/query_planner/tests.rs
```diff
@@ -1458,9 +1455,7 @@ async fn typename_propagation2() {
                         "representations": [{
                             "__typename": "Book",
                             "bookId": "book1",
-                            "author": {
-                                "fullName": "Ada"
-                            }
+                            "author": null
```
Geal, line 1458: why this change?

.changesets/fix_bnjjj_fix_4348.md
```diff
@@ -0,0 +1,5 @@
+### Fix fragment spread response formatting with aliased typename ([Issue #4348](https://github.com/apollographql/router/issues/4348))
+
+When you had an aliased `__typename` (example `myAlias: __typename`) in a fragment spread, the response formatting was wrong.
```
Geal, line 3: the changeset should probably reflect the actual issue, which is that response formatting was confused by fragments on unions

apollo-router/src/spec/fragments.rs
```diff
@@ -25,17 +26,24 @@ impl Fragments {
...
                     selection_set: fragment
                         .selection_set
                         .selections
                         .iter()
                         .filter_map(|selection| {
+                            if let Some(field) = selection.as_field() {
+                                if field.name.as_str() == TYPENAME && field.alias.is_some() {
+                                    aliased_typename = field.alias.as_ref().map(|f| f.to_string());
+                                }
+                            }
```
Geal, line 41: the aliased typename field should be removed now, we don't need it anymore

Geal (changes requested):

apollo-router/src/query_planner/tests.rs
bnjjj: Because apparently that's the query planner is asking for. If I don't do this I have an error telling me that no mock request is matching

Geal: right, I saw that, it's ok

Geal (approved):

PR #4418, opened 2023-12-21
fix(metrics): record api schema & license metrics to apollo.router.lifecycle

### LINKED ISSUE #4406: metrics: trace api schema generation errors under apollo.router.operations
We currently set up difference in results from `experimental_api_schema_generation_mode` to be traced under `apollo.router.api_schema`. In order for these metrics to be ingressed properly, they should be stored under `apollo.router.operations.api_schema`.

lrlna:
This commit reorganises some of our metrics: `api_schema` and `license` now get stored under the new `lifecycle` key. This allows us to represent metrics pertaining to response lifecycles and router reloads rather than storing them under `config`.

Not adding a changeset, as this doesn't need to be in the changelog.

fixes #4406

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/.changesets/fix_lrlna_api_schema_metrics_4406.md
+++ b/.changesets/fix_lrlna_api_schema_metrics_4406.md
@@ -1,6 +1,0 @@
-
-### Trace api schema generation mismatches & license metrics under apollo.router.lifecycle ([Issue #4406](https://github.com/apollographql/router/issues/4406))
-
-Store any difference between generated JavaScript and Rust API schemas under `apollo.router.lifecycle` metrics in order for them to be properly ingested. Trace license state under `apollo.router.lifecycle`.
-
-By [@lrlna](https://github.com/lrlna) in https://github.com/apollographql/router/pull/4418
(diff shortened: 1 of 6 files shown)


o0Ignition0o (approved):

.changesets/fix_lrlna_api_schema_metrics_4406.md
```diff
@@ -0,0 +1,6 @@
+
+### Trace api schema generation errors under apollo.router.operations ([Issue #4406](https://github.com/apollographql/router/issues/4406))
```
goto-bus-stop, line 2: do we really need a public facing changeset here? IMO it may be useful for users to know that a new toAPISchema implementation is being tested but not so much how its reported to us.

goto-bus-stop (approved):

.changesets/fix_lrlna_api_schema_metrics_4406.md
lrlna: i think this changeset is likely to get dropped in the release? or do we just not write one in that case?

o0Ignition0o (approved):
Love it!

.changesets/fix_lrlna_api_schema_metrics_4406.md
o0Ignition0o: I tend to not write changesets in this situation, and explain why i didnt write any in the PR body

PR #4423, opened 2023-12-22
Target a federation branch that has an input rewrites fix

o0Ignition0o:
Kept as draft because it is for testing purposes. This PR will be superseded by an official router-bridge bump.


Geal:
closing this as we have integrated the fix in router-bridge and the router

PR #4441, opened 2024-01-05
fix: build entity representations with inline fragments correctly

lennyburdette:
This uses a more flexible abstract type / concrete type check when applying a selection set to an entity reference before it's used in a fetch node. Previous to this change, we would drop data from the reference when it selected using an inline fragment (e.g. `@requires(fields: "... on Foo { a } ... on Bar { b }")`).

This addresses a customer issue.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/.changesets/fix_lb_requires_selection_set_abstract_types.md
+++ b/.changesets/fix_lb_requires_selection_set_abstract_types.md
@@ -1,0 +1,5 @@
+### fix: build entity representations with inline fragments correctly ([PR #4441](https://github.com/apollographql/router/pull/4441))
+
+This uses a more flexible abstract type / concrete type check when applying a selection set to an entity reference before it's used in a fetch node. Previous to this change, we would drop data from the reference when it selected using an inline fragment (e.g. `@requires(fields: "... on Foo { a } ... on Bar { b }")`).
+
+By [@lennyburdette](https://github.com/lennyburdette) in https://github.com/apollographql/router/pull/4441
--- a/apollo-router/src/query_planner/selection.rs
+++ b/apollo-router/src/query_planner/selection.rs
@@ -1,3 +1,4 @@
+use apollo_compiler::schema::ExtendedType;
 use serde::Deserialize;
 use serde::Serialize;
 use serde_json_bytes::ByteString;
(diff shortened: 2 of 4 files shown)


lennyburdette:
i went back to see if my test fails on dev and it doesn't. maybe this issue is actually related to changes coming in router bridge? i'm a bit confused

lennyburdette:
Ok this bug is related to the query planner changes. Some printlns from the test on dev:

```
current_type: Entity | condition: Entity | matches: true
current_type: Union1 | condition: Union1 | matches: true
current_type: Union1 | condition: Union2 | matches: false
current_type: NestedObject | condition: NestedObject | matches: true
current_type: NestedObject | condition: NestedObject2 | matches: false
```

The same printlns for the test on https://github.com/apollographql/router/pull/4423:

```
current_type: Entity | condition: Entity | matches: true
current_type: NestedUnion | condition: Union1 | matches: false
current_type: NestedUnion | condition: Union2 | matches: false
current_type: WrapperNestedInterface | condition: NestedObject | matches: false
current_type: WrapperNestedInterface | condition: NestedObject2 | matches: false
```

So the "requires" selection set we're using to create the entity references has changed and the implementation on `dev` will not work.

.changesets/fix_lb_requires_selection_set_abstract_types.md
```diff
@@ -0,0 +1,8 @@
+### fix: build entity representations with inline fragments correctly ([PR #4441](https://github.com/apollographql/router/pull/4441))
+
+This uses a more flexible abstract type / concrete type check when applying a selection set to an entity reference before it's used in a fetch node. Previous to this change, we would drop data from the reference when it selected using an inline fragment (e.g. `@requires(fields: "... on Foo { a } ... on Bar { b }")`).
+
+<!-- start metadata -->
+---
```
bnjjj, line 6:
```suggestion
```

Geal (approved):

PR #4446, opened 2024-01-09
[WIP] docs: add progressive_override config option

shorgi:


shorgi:
config is outdated, not needed

PR #4447, opened 2024-01-09
refactor(multipart): do not enforce boundary in accept header

bnjjj:
Related to #4395

Remove the multipart boundary from the client `Accept` header. Typically, the boundary is decided by the server in order to avoid nameclashes with the parts contents. It just happens that `--graphql` works all the time for us because the part contents are JSON.

Keep `--graphql` as a "may" for convenience but mandate that clients support any boundary sent by the server. I _think_ this is already the case for web/ios/kotlin? Ping @alessbell @calvincestari 

ping @martinbonnin  @abernix 

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [x] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/services/mod.rs
+++ b/apollo-router/src/services/mod.rs
@@ -71,10 +71,8 @@ pub(crate) fn apollo_graph_reference() -> Option<String> {
 // set the supported `@defer` specification version to https://github.com/graphql/graphql-spec/pull/742/commits/01d7b98f04810c9a9db4c0e53d3c4d54dbf10b82
 pub(crate) const MULTIPART_DEFER_SPEC_PARAMETER: &str = "deferSpec";
 pub(crate) const MULTIPART_DEFER_SPEC_VALUE: &str = "20220824";
-pub(crate) const MULTIPART_DEFER_CONTENT_TYPE: &str =
-    "multipart/mixed;boundary=\"graphql\";deferSpec=20220824";
+pub(crate) const MULTIPART_DEFER_CONTENT_TYPE: &str = "multipart/mixed;deferSpec=20220824";
 
-pub(crate) const MULTIPART_SUBSCRIPTION_CONTENT_TYPE: &str =
-    "multipart/mixed;boundary=\"graphql\";subscriptionSpec=1.0";
+pub(crate) const MULTIPART_SUBSCRIPTION_CONTENT_TYPE: &str = "multipart/mixed;subscriptionSpec=1.0";
 pub(crate) const MULTIPART_SUBSCRIPTION_SPEC_PARAMETER: &str = "subscriptionSpec";
 pub(crate) const MULTIPART_SUBSCRIPTION_SPEC_VALUE: &str = "1.0";
--- a/apollo-router/src/services/mod.rs
+++ b/apollo-router/src/services/mod.rs
@@ -71,8 +71,12 @@ pub(crate) fn apollo_graph_reference() -> Option<String> {
 // set the supported `@defer` specification version to https://github.com/graphql/graphql-spec/pull/742/commits/01d7b98f04810c9a9db4c0e53d3c4d54dbf10b82
 pub(crate) const MULTIPART_DEFER_SPEC_PARAMETER: &str = "deferSpec";
 pub(crate) const MULTIPART_DEFER_SPEC_VALUE: &str = "20220824";
-pub(crate) const MULTIPART_DEFER_CONTENT_TYPE: &str = "multipart/mixed;deferSpec=20220824";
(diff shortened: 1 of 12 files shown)


docs/source/executing-operations/defer-support.mdx
```diff
@@ -40,6 +40,19 @@ To learn the basics of the `@defer` directive and how you can use it with your s
 
 The remainder of this article covers the Apollo Router's defer implementation in greater depth.
 
+## Executing a `@defer` query
+
+To execute a deferred query on the Apollo Router, a GraphQL client sends an HTTP request with _almost_ the exact same format that it uses for usual query and mutation requests.
+
+The only difference is that the request should include the following `Accept` header:
```
martinbonnin, line 47: _Should_ we make this a "must"?
```suggestion
The only difference is that the request must include the following `Accept` header:
```

docs/source/executing-operations/defer-support.mdx
```diff
@@ -40,6 +40,19 @@ To learn the basics of the `@defer` directive and how you can use it with your s
...
+
+```text title="Example header"
+Accept: multipart/mixed; deferSpec=20220824, application/json
```
martinbonnin, line 50: Nit: [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept) uses space after `,` but no space after `;`:

```suggestion
Accept: multipart/mixed;deferSpec=20220824, application/json
```

I kind of like it because it's easier to "see" the different list values. But either works with me.

martinbonnin (approved):
🚀

calvincestari:
> Keep --graphql as a "may" for convenience but mandate that clients support any boundary sent by the server. I think this is already the case for web/ios/kotlin? Ping @alessbell @calvincestari

Correct for Apollo iOS.

docs/source/executing-operations/defer-support.mdx
calvincestari: Yes, I agree. If you don't send the header you won't get back a multipart response. Which essentially is not executing the defer query.

apollo-router/src/services/mod.rs
```diff
@@ -71,10 +71,8 @@ pub(crate) fn apollo_graph_reference() -> Option<String> {
 // set the supported `@defer` specification version to https://github.com/graphql/graphql-spec/pull/742/commits/01d7b98f04810c9a9db4c0e53d3c4d54dbf10b82
 pub(crate) const MULTIPART_DEFER_SPEC_PARAMETER: &str = "deferSpec";
 pub(crate) const MULTIPART_DEFER_SPEC_VALUE: &str = "20220824";
-pub(crate) const MULTIPART_DEFER_CONTENT_TYPE: &str =
-    "multipart/mixed;boundary=\"graphql\";deferSpec=20220824";
+pub(crate) const MULTIPART_DEFER_CONTENT_TYPE: &str = "multipart/mixed;deferSpec=20220824";
```
Geal, line 74: that string is also used to generate the `Content-Type` header of the response to defer queries, so the boundary should be present. Maybe make separate constants, one for the media type in the `Accept` header and error logs, one for the `Content-Type`?
There's the same issue for subscriptions AFAIK

Geal (changes requested):

apollo-router/src/test_harness/http_client.rs
```diff
@@ -123,8 +123,7 @@ where
                 let defer_spec = mediatype::Name::new("deferSpec").unwrap();
                 assert_eq!(media_type.subty, "mixed");
                 assert_eq!(media_type.get_param(defer_spec).unwrap(), "20220824");
-                let boundary = media_type.get_param(mediatype::names::BOUNDARY).unwrap();
-                let boundary = format!("\r\n--{}", boundary.unquoted_str());
+                let boundary = "\r\n--graphql".to_string();
```
Geal, line 126: it should still use the boundary given in the content-type

dev-docs/multipart-subscriptions-protocol.md
```diff
@@ -27,6 +26,9 @@ Content-Type: application/json
 --graphql--
 ```
 
+> Note: because the parts are always JSON, it is never possible for `\r\n--graphql` to appear in the contents of a part. For convenience, servers MAY use `graphql` as a boundary.
```
Geal, line 29: is that warning necessary? Boundary parsing is standard in multipart

martinbonnin: I added this as a non-normative note for implementers more than a warning. This saves the time to implementers trying to think of a correct boundary. I'm fine skipping it if we think it doesn't have value.

Geal (approved):

PR #4451, opened 2024-01-09
fix(log): set info log level by default for the entire application

bnjjj:
By default the log level of the application if `RUST_LOG` is not set is `info` and if you provide `--log` or `APOLLO_RUST_LOG` value then it overrides the default `info` log level with `apollo_router=...` because it only impacts the `apollo_router` crate, not external custom plugin and so one.

By doing this fix, by default if you have a custom plugin with info logs or metrics you won't have to enforce `RUST_LOG=info` everytime, it will work as expected.

> Note: it doesn't impact the behavior of `RUST_LOG` if you set it.

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/.changesets/fix_bnjjj_fix_log_level.md
+++ b/.changesets/fix_bnjjj_fix_log_level.md
@@ -1,0 +1,9 @@
+### Set log level to `info` by default for the entire application ([PR #4451](https://github.com/apollographql/router/pull/4451))
+
+By default the log level of the application if `RUST_LOG` is not set is `info` and if you provide `--log` or `APOLLO_RUST_LOG` value then it overrides the default `info` log level with `apollo_router=...` because it only impacts the `apollo_router` crate, not external custom plugin and so one.
+
+By doing this fix, by default if you have a custom plugin with info logs or metrics you won't have to enforce `RUST_LOG=info` everytime, it will work as expected.
+
+> Note: it doesn't impact the behavior of `RUST_LOG` if you set it.
+
+By [@bnjjj](https://github.com/bnjjj) in https://github.com/apollographql/router/pull/4451
--- a/apollo-router/src/executable.rs
+++ b/apollo-router/src/executable.rs
@@ -265,9 +265,9 @@ fn add_log_filter(raw: &str) -> Result<String, String> {
             let res = rgx.replace_all(&lowered, |caps: &Captures| {
                 // The default level is info, then other ones can override the default one
                 // If the pattern matches, we must have caps 1 and 2
-                format!("info,{}apollo_router={}", &caps[1], &caps[2])
+                format!("{}apollo_router={}", &caps[1], &caps[2])
             });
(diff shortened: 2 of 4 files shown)


o0Ignition0o (approved):

.changesets/fix_bnjjj_fix_log_level.md
```diff
@@ -0,0 +1,9 @@
+### Set log level to `info` by default for the entire application ([PR #4451](https://github.com/apollographql/router/pull/4451))
+
+By default the log level of the application if `RUST_LOG` is not set is `info` and if you provide `--log` or `APOLLO_RUST_LOG` value then it overrides the default `info` log level with `apollo_router=...` because it only impacts the `apollo_router` crate, not external custom plugin and so one.
+
+By doing this fix, by default if you have a custom plugin with info logs or metrics you won't have to enforce `RUST_LOG=info` everytime, it will work as expected.
```
o0Ignition0o, line 5:
```suggestion
Customizing the router's [Log level](https://www.apollographql.com/docs/router/configuration/telemetry/exporters/logging/overview/#log-level) now applies to the router *and* native plugins.
```

bnjjj: That's wrong, it doesn't apply to the plugins if you set `debug`. This change introduces a better default behavior. To not have to enforce `RUST_LOG=info`

BrynCooke (changes requested):
The docs need to updated:
https://www.apollographql.com/docs/router/configuration/telemetry/exporters/logging/overview/#log-level

Also @garypen do you agree with the overall change?

bnjjj:
@BrynCooke I think the documentation is still accurate, I didn't change the behavior. I just fixed it

garypen:
> The docs need to updated: https://www.apollographql.com/docs/router/configuration/telemetry/exporters/logging/overview/#log-level
> 
> Also @garypen do you agree with the overall change?

The motivation of the last adjustment to APOLLO_ROUTER_LOG was to limit its scope to "just router stuff". We restored RUST_LOG functionality to what a rust developer would expect (i.e.: you can do whatever is supported by the RUST_LOG filtering). I think there was an oversight in that fix which mean that APOLLO_ROUTER_LOG didn't apply to plugins, so if you wanted to see logs from plugins you had to use RUST_LOG.

If this fix is addressing that shortcoming, i.e.: plugins observe the same logging behaviour as the main router, then that is correct.

It looks to me like it might also change the value for non-plugin crates, e.g.: `hyper`, and, if it does, then I don't think that would be desirable.

Geal (approved):

garypen (approved):

BrynCooke (approved):

PR #4453, opened 2024-01-10
feat(redis): add options to set username and password separately

### LINKED ISSUE #4346: add options to set the username or password separately for Redis
Right now, the Redis username and password are set in the URL, but due to that we run into issues like https://github.com/aembke/fred.rs/issues/186. We could add an option to set the username and password directly. This would also allow the password to be set from environment variables

bnjjj:
Example of configuration:

```yaml title="router.yaml"
supergraph:
  query_planning:
    experimental_cache:
      redis: #highlight-line
        urls: ["redis://..."] #highlight-line
        username: admin/123 # Optional, can be part of the urls directly, mainly useful if you have special character like '/' in your password that doesn't work in url
        password: admin # Optional, can be part of the urls directly, mainly useful if you have special character like '/' in your password that doesn't work in url
```

Fixes #4346

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/cache/redis.rs
+++ b/apollo-router/src/cache/redis.rs
@@ -136,6 +136,14 @@ impl RedisCacheStorage {
         let url = Self::preprocess_urls(config.urls)?;
         let mut client_config = RedisConfig::from_url(url.as_str())?;
 
+        if let Some(username) = config.username {
+            client_config.username = Some(username);
+        }
+
+        if let Some(password) = config.password {
+            client_config.password = Some(password);
+        }
+
         if let Some(tls) = config.tls.as_ref() {
             let tls_cert_store = tls.create_certificate_store().transpose()?;
             let client_cert_config = tls.client_authentication.as_ref();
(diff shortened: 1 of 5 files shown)


apollo-router/src/cache/redis.rs
```diff
@@ -136,6 +136,14 @@ impl RedisCacheStorage {
         let url = Self::preprocess_urls(config.urls)?;
         let mut client_config = RedisConfig::from_url(url.as_str())?;
 
+        if let Some(username) = config.username {
+            client_config.username = Some(username);
```
Geal, line 140: maybe check if the URL does not have already a username or password? Who takes precedence in that case?

bnjjj: It's documented

bnjjj: Fixed

Geal (approved):

PR #4460, opened 2024-01-11
Aggregate meter provider also delegates to global meter provider

### LINKED ISSUE #4313: Add the ability to use the regular opentelemetry global meter provider for metrics
Currently we have a special meter provider that allows us to hot reload prometheus.
If the apollo meter provider is not set then metrics should be sent via the regular otel meter provider.

Changes should be made here:
https://github.com/apollographql/router/blob/dev/apollo-router/src/metrics/mod.rs#L714-L770
The instrument should still be cached, but they will NEVER be updated config reload.

BrynCooke:
The `AggregateMeterProvider` is used by the Router to allow us to have several metrics subsystems that forward metrics to different places. In addition it has enhanced testability.

However, some users may be trying to set up otel by themselves outside of the regular router lifecycle, and they will receive no metrics.

This PR also makes the `AggregateMeterProvider` forward to the `GlobalMeterProvider`. The setup for this happens at creation time of `AggregateMeterProvider` and setting the regular otel meter provider after this happens will have no effect.

This functionality is not reccommeneded for general users, and may change in breaking ways in the future.

Fixes #4313

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/metrics/filter.rs
+++ b/apollo-router/src/metrics/filter.rs
@@ -9,7 +9,6 @@ use opentelemetry::metrics::Counter;
 use opentelemetry::metrics::Histogram;
 use opentelemetry::metrics::InstrumentProvider;
 use opentelemetry::metrics::Meter;
-use opentelemetry::metrics::MeterProvider;
 use opentelemetry::metrics::ObservableCounter;
 use opentelemetry::metrics::ObservableGauge;
 use opentelemetry::metrics::ObservableUpDownCounter;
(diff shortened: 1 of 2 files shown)


apollo-router/src/metrics/filter.rs
```diff
@@ -177,17 +206,22 @@ impl InstrumentProvider for FilteredInstrumentProvider {
...
         name: impl Into<Cow<'static, str>>,
         version: Option<impl Into<Cow<'static, str>>>,
         schema_url: Option<impl Into<Cow<'static, str>>>,
         attributes: Option<Vec<KeyValue>>,
     ) -> Meter {
-        let delegate = self
-            .delegate
-            .versioned_meter(name, version, schema_url, attributes);
+        let delegate = match &self.delegate {
+            MeterProvider::Regular(provider) => {
```
o0Ignition0o, line 218: Why not put it in `impl MeterProvider {` above?

o0Ignition0o (approved):

bnjjj (approved):

apollo-router/src/metrics/filter.rs
BrynCooke: Nice catch: [36cf7a4](https://github.com/apollographql/router/pull/4460/commits/36cf7a42f5e41bb28761ed48fba04e79e5bd077e)

PR #4465, opened 2024-01-11
JWKS parse improvements

### LINKED ISSUE #4463: JWK parse failure prevents processing of entire JWKS
During parsing of a JWKS we wish to skip any JWK that can't be parsed. Currently there is some logic to deal with algorithms, but there may be other issues. 

Make the parsing safer and add tests for log messages.

BrynCooke:
When parsing a JWKS the Router should ignore any JWKs that fail to parse, rather than failing the entire JWKS parse. 
This can happen when the JWK is malformed, or when a JWK uses an unknown algorithm. When this happens a warning will be output to the logs, for example: 

```
2024-01-11T15:32:01.220034Z WARN fetch jwks{url=file:///home/bryn/git/router/jwks.json,}  ignoring a key since it is not valid, enable debug logs to full content err=unknown variant `UnknownAlg`, expected one of `HS256`, `HS384`, `HS512`, `ES256`, `ES384`, `RS256`, `RS384`, `RS512`, `PS256`, `PS384`, `PS512`, `EdDSA` alg="UnknownAlg" index=2
```

Log messages have the following attributes:
* `alg` The JWK algorithm if known or `<unknown>`
* `index` The index of the JWK within the JWKS.
* `url` The URL of the JWKS that had the issue.

Fixes #4463

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/plugins/authentication/jwks.rs
```diff
@@ -199,33 +206,23 @@ pub(super) async fn get_jwks(url: Url) -> Option<JwkSet> {
...
-                let alg_json = key.get("alg").unwrap_or_else(|| &Value::Null);
-                // Retention rules are, retain if:
-                //  - we can't find an alg field
-                //  - alg field isn't a string
-                //  - jsonwebtoken supports the algorithm (determined by creating an enum member)
-                if alg_json == &Value::Null || !alg_json.is_string() {
-                    true
-                } else {
-                    let alg_name = alg_json.as_str().expect("we checked it's a string");
-                    match Algorithm::from_str(alg_name)
```
Geal, line 214: why remove the algorithm checking?

Geal:
since https://github.com/apollographql/router/commit/fd5285a5d477db7e3f84eaf1afa4206cfebcfc38 it should already be the case that we don't reject the entire JWKS file, but instead we keep only the keys that we can parse. What is failing right now?

apollo-router/src/plugins/authentication/jwks.rs
BrynCooke: It's checked by virtue of it trying to parse the individual key.

The error message can be seen in the tests.

BrynCooke:
> since [fd5285a](https://github.com/apollographql/router/commit/fd5285a5d477db7e3f84eaf1afa4206cfebcfc38) it should already be the case that we don't reject the entire JWKS file, but instead we keep only the keys that we can parse. What is failing right now?

Yeah, there are various other ways to trigger the issue other than incorrect algorithm. For instance other keys with the wrong JSON type (make an `key_ops` a string rather than an array). The idea behind this PR is that we just let the JWK parsing deal with all these eventualities rather than dealing with them ourselves.

apollo-router/src/plugins/authentication/jwks.rs
Geal: nice!

apollo-router/src/logging/mod.rs
```diff
@@ -84,7 +84,14 @@ macro_rules! assert_snapshot_subscriber {
         $crate::logging::test::SnapshotSubscriber::create_subscriber(|yaml| {
             insta::with_settings!({sort_maps => true}, {
                 // the tests here will force maps to sort
-                insta::assert_yaml_snapshot!(yaml);
+                let mut settings = insta::Settings::clone_current();
```
Geal, line 87: could you explain the goal of this change and the snapshot- files renaming? It does not look related to the JWKS parsing issue

BrynCooke: It's because the experience of using the logging tests that also uses insta wasn't great as you end up with snapshots suffixed with an index number. It only manifested itself in this issue as I had not used this pattern before, but can pull out into a separate PR if you like.

Geal: it's ok, let's merge it

Geal (approved):

PR #4466, opened 2024-01-11
Fix format_response for statically skipped root selection set 

### LINKED ISSUE #4397: Inconsistent/invalid null data response when using variables
So we are seeing an issue with nullable fields being skipped. A query like
```
query($skip: Boolean = false) {
  viewer @skip(if: $skip) {
    login
  }
}
```
returns (with variables `{"skip": true}`)
```
{
  "data": null
}
```
while 
```
query {
  viewer @skip(if: true) {
    login
  }
}
```
returns
```
{
  "data": {}
}
```
Per https://spec.graphql.org/October2021/#sec-Data:
> If an error was raised during the execution that prevented a valid response, the data entry in the response should be null.

but this does not seem to be an error case (or vars vs no vars would be equivalent). The spec is a bit vague about what happens if there is no error, but also nothing to return.
It also describes other error cases that would result in an `errors` array, and no `data` field being returned.
Not sure how all the cases would present, but certainly passing a variable shouldn't impact the ultimate return shape, I would think.

_Originally posted by @carldunham in https://github.com/apollographql/router/issues/1308#issuecomment-1855230208_

bnjjj:
Either you have query like this:

```graphql
{
    get @skip(if: true) {
        id
        name
    }
}
```

or 

```graphql
{
    get($skip: Boolean = true) @skip(if: $skip) {
        id
        name
    }
}
```

you'll receive the same response

```json
{ "data": {}}
```

Fixes #4397

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [x] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


.changesets/fix_bnjjj_fix_4397.md
```diff
@@ -0,0 +1,31 @@
+### Fix format_response for statically skipped root selection set  ([Issue #4397](https://github.com/apollographql/router/issues/4397))
+
+Either you have query like this:
```
Geal, line 3: that changeset is not very clear, could you give a bit more details?

apollo-router/src/services/query_planner.rs
```diff
@@ -79,7 +79,7 @@ pub(crate) struct Response {
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub(crate) enum QueryPlannerContent {
     Plan { plan: Arc<QueryPlan> },
-    Introspection { response: Box<graphql::Response> },
+    Response { response: Box<graphql::Response> },
```
Geal, line 82: nice change, that will be easier to follow

apollo-router/src/query_planner/execution.rs
```diff
@@ -354,6 +354,10 @@ impl PlanNode {
                                     .await;
                                 value.deep_merge(v);
                                 errors.extend(err.into_iter());
+                            } else if current_dir.is_empty() {
+                                // If the condition is on the root selection set and it's the only one
+                                // For queries like {get @skip(if: true) {id name}}
+                                value.deep_merge(Value::Object(Default::default()));
```
Geal, line 360: is this because the initial value passed to `execute_recursively` can be `Value::Null`, as set on line 75 `&initial_value.unwrap_or_default(),`?
If that's the case, I think it would be better to do `&initial_value.unwrap_or_else(|| Value::Object(Default::default())),`, that would avoid those additional checks and possibly prevent future issues in other parts of query plan execution, since we should always work on an object

Geal (changes requested):

apollo-router/src/query_planner/execution.rs
bnjjj: No it's not related, it's a specific case that can only happen with `skip` and `include` directive on a root selection_set

Geal (approved):

BrynCooke (approved):

Geal (approved):

PR #4490, opened 2024-01-16
test: change test log subscriber to capture trace/debug

lennyburdette:
---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/logging/mod.rs
+++ b/apollo-router/src/logging/mod.rs
@@ -86,7 +86,7 @@ pub(crate) mod test {
 /// You can also use subscriber::with_default(assert_snapshot_subscriber!(), || { ... }) to assert the logs in non async code.
 macro_rules! assert_snapshot_subscriber {
     () => {
-        $crate::assert_snapshot_subscriber!(tracing_core::LevelFilter::TRACE, {})
+        $crate::assert_snapshot_subscriber!(tracing_core::LevelFilter::INFO, {})
     };
 
     ($level:expr) => {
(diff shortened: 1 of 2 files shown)


apollo-router/src/logging/mod.rs
```diff
@@ -97,7 +102,21 @@ macro_rules! assert_snapshot_subscriber {
...
+                settings.set_snapshot_suffix("logs");
+                settings.set_sort_maps(true);
+                settings.bind(|| {
+                    insta::assert_yaml_snapshot!(yaml, $redactions);
+                });
+            });
+        })
+    };
+
+    ($redactions:tt, $level:tt) => {
```
BrynCooke, line 118: Let's put level before redactions in the parameter list.

apollo-router/src/logging/mod.rs
```diff
@@ -97,7 +102,21 @@ macro_rules! assert_snapshot_subscriber {
...
+        $crate::logging::test::SnapshotSubscriber::create_subscriber($level, |yaml| {
```
BrynCooke, line 119: I think this can be DRYed up a bit with just one implementation and the other macro calls delegating to it.

lennyburdette: i think i did this, please let me know if you meant something else!

BrynCooke (approved):

PR #4515, opened 2024-01-18
follow-up to router darwin deliverables changes

garypen:
We need to:
 - update doc referece to x86_86 to Apple Silicon
 - update our xtask packaging so it knows about aarch64-apple-darwin


o0Ignition0o (approved):

abernix:
see additional follow-up in https://github.com/apollographql/router/pull/4518

PR #4484, opened 2024-01-16
update macos builds in CI to use new m1 machines

garypen:
Update the build pipeline to produce apple silicon binaries for macos. Additionally, update the installer script to reflect the changes in binary architecture for macOS.

Fixes #3902 

also see: #4483 

---

**Checklist**

Complete the checklist (and note appropriate exceptions) before the PR is marked ready-for-review.

- [ ] Changes are compatible[^1]
- [x] Documentation[^2] completed
- [x] Performance impact assessed and acceptable
- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [x] Manual Tests

**Exceptions**

The installer script was tested manually to make sure it did the "right thing" with respect to no longer modifying _cputype.

**Notes**

[^1]: It may be appropriate to bring upcoming changes to the attention of other (impacted) groups. Please endeavour to do this before seeking PR approval. The mechanism for doing this will vary considerably, so use your judgement as to how and when to do this.
[^2]: Configuration is an important part of many changes. Where applicable please try to document configuration examples.
[^3]: Tick whichever testing boxes are applicable. If you are adding Manual Tests, please document the manual testing (extensively) in the Exceptions.

Issue #4483, opened 2024-01-16
macos: removal of support for intel based systems


garypen:
Our [CI provider](https://circleci.com/) announced a while ago that support for [mac os intel build was soon to be deprecated](https://discuss.circleci.com/t/macos-intel-support-deprecation-in-january-2024/48718).

As we draw close to that date we have decided to update our build pipeline to use the new Apple Silicon build machines. This is good news for anyone using a recent M1/M2/M3 based Apple system, since they no longer need to run the router through the Rosetta2 emulator and will get a nice performance boost, but it's problematic for older intel based systems, since they won't be able to run the router binary at all.

This issue is for tracking macos intel users who would like us to continue to provide a router binary that works on those systems.

Let us know here if you fall into that category. Also: consider these alternatives
 - docker images will still work fine on your macos intel system
 - building from source is still an option

It's a non-trivial amount of work to produce/test/support a universal binary so we'll be relying on feedback to this issue to schedule that work.

**Update: while from now on we will only use ARM workers to build and test on CircleCI, we will build router binaries for both x86_64 and aarch64 (ARM) as part of our release process**

BoD:
FWIW we were relying on these builds when running integration tests with Apollo Kotlin on GitHub CI `macos-11` runners (to test our MacOS target). For now we'll switch to linux runners (we could certainly use Docker images too, but that's a less trivial change).

destumme:
If the json schema of each router version configuration was hosted somewhere and didn't require running the binary to generate, this change won't have a major effect on my workflows (tend to run everything in docker locally regardless). Without that though, it's pretty irritating to get the updated schema to validate the router config file, that was the only thing I was using the binary download directly for.

Issue #4448, opened 2024-01-09
Improve JWKS logging to allow debugging of issues during download


BrynCooke:
Currently it is difficult to figure out what has gone wrong when a JWKS fails to download.
Enabling hyper trace logs is not an acceptable solution as this produces a ton of noise, and because we don't use spans they are difficult to correlate.

1. Add a debug message to indicate what was retrieved from the JWKS url.
2. Add a span to add URL context to the entire retrieval process.

PR #4516, opened 2024-01-18
prep release: v1.38.0-rc.1

garypen:
> **Note**
>
> When approved, this PR will merge into **the `1.38.0` branch** which will — upon being approved itself — merge into `main`.
>
>  - Version bumps
>  - That it targets the right release branch (`1.38.0` in this case!).
>
---
### Redis key namespace ([Issue #4247](https://github.com/apollographql/router/issues/4247))

This implements key namespacing in redis caching. The namespace, if provided, will be added as a prefix to the key: `namespace:key`.

By [@Geal](https://github.com/Geal) in https://github.com/apollographql/router/pull/4458

### Promote HTTP request size limit from experimental to general availability ([PR #4442](https://github.com/apollographql/router/pull/4442))

By default Apollo Router limits the size of the HTTP request body it will read from the network to 2 MB. In this version, the YAML configuration to change the limit is promoted from experimental to general availability.

For more information about launch stages, please see the documentation here: https://www.apollographql.com/docs/resources/product-launch-stages/

Before increasing this limit significantly consider testing performance in an environment similar to your production,
especially if some clients are untrusted. Many concurrent large requests could cause the Router to run out of memory.

Previous configuration will warn but still work:

```yaml
limits:
  experimental_http_max_request_bytes: 2000000 # Default value: 2 MB
```

The warning can be fixed by removing the `experimental_` prefix:

```yaml
limits:
  http_max_request_bytes: 2000000 # Default value: 2 MB
```

By [@SimonSapin](https://github.com/SimonSapin) in https://github.com/apollographql/router/pull/4442

### Add options to set username and password separately for Redis ([Issue #4346](https://github.com/apollographql/router/issues/4346))

Example of configuration:

```yaml title="router.yaml"
supergraph:
  query_planning:
    experimental_cache:
      redis: #highlight-line
        urls: ["redis://..."] #highlight-line
        username: admin/123 # Optional, can be part of the urls directly, mainly useful if you have special character like '/' in your password that doesn't work in url. This field takes precedence over the username in the URL
        password: admin # Optional, can be part of the urls directly, mainly useful if you have special character like '/' in your password that doesn't work in url. This field takes precedence over the password in the URL
```

By [@bnjjj](https://github.com/bnjjj) in https://github.com/apollographql/router/pull/4453

### Fix the Datadog default tracing exporter URL ([Issue #4415](https://github.com/apollographql/router/issues/4416))

The default URL for the Datadog exporter was incorrectly set to `http://localhost:8126/v0.4/traces` which caused issues for users that were running different agent versions.
This is now fixed and matches the exporter URL of `http://127.0.0.1:8126`.

By [@BrynCooke](https://github.com/BrynCooke) in https://github.com/apollographql/router/pull/4444

### Use the right conventions for batching metrics ([PR #4424](https://github.com/apollographql/router/pull/4424))

Rename `apollo_router.operations.batching` to  `apollo.router.operations.batching` because it doesn't follow our conventions with `apollo.router.` prefix.

By [@bnjjj](https://github.com/bnjjj) in https://github.com/apollographql/router/pull/4424

### Fix format_response for statically skipped root selection set ([Issue #4397](https://github.com/apollographql/router/issues/4397))

If in your GraphQL operation you have a root selection set skipped by `@skip` or `@include` directive, before this fix the results you got if you hardcoded the value of the parameter in `@skip` directive like this for example:

```graphql
{
    get @skip(if: true) {
        id
        name
    }
}
```

or if you used a variable like this:

```graphql
{
    get(: Boolean = true) @skip(if: ) {
        id
        name
    }
}
```

you received different response. Now it both answers:

```json
{ "data": {}}
```

By [@bnjjj](https://github.com/bnjjj) in https://github.com/apollographql/router/pull/4466

### Improve JWKS parse error handling ([Issue #4463](https://github.com/apollographql/router/issues/4463))

When parsing a JWKS the Router should ignore any JWKs that fail to parse, rather than failing the entire JWKS parse. 
This can happen when the JWK is malformed, or when a JWK uses an unknown algorithm. When this happens a warning will be output to the logs, for example: 

```
2024-01-11T15:32:01.220034Z WARN fetch jwks{url=file:///tmp/jwks.json,}  ignoring a key since it is not valid, enable debug logs to full content err=unknown variant `UnknownAlg`, expected one of `HS256`, `HS384`, `HS512`, `ES256`, `ES384`, `RS256`, `RS384`, `RS512`, `PS256`, `PS384`, `PS512`, `EdDSA` alg="UnknownAlg" index=2
```

Log messages have the following attributes:
* `alg` The JWK algorithm if known or `<unknown>`
* `index` The index of the JWK within the JWKS.
* `url` The URL of the JWKS that had the issue.

By [@BrynCooke](https://github.com/BrynCooke) in https://github.com/apollographql/router/pull/4465

### update router-bridge to v0.5.14+v2.6.3 ([PR #4468](https://github.com/apollographql/router/pull/4468))

This federation update contains query planning fixes for:
* invalid typename used when calling into a subgraph that uses `@interfaceObject`
* performance issue when generating planning paths for union members that use `@requires`

By [@Geal](https://github.com/Geal) in https://github.com/apollographql/router/pull/4468

### Set log level to `info` by default for the entire application ([PR #4451](https://github.com/apollographql/router/pull/4451))

By default the log level of the application if `RUST_LOG` is not set is `info` and if you provide `--log` or `APOLLO_RUST_LOG` value then it overrides the default `info` log level with `apollo_router=...` because it only impacts the `apollo_router` crate, not external custom plugin and so one.

By doing this fix, by default if you have a custom plugin with info logs or metrics you won't have to enforce `RUST_LOG=info` everytime, it will work as expected.

> Note: it doesn't impact the behavior of `RUST_LOG` if you set it.

By [@bnjjj](https://github.com/bnjjj) in https://github.com/apollographql/router/pull/4451

### fix: build entity representations with inline fragments correctly ([PR #4441](https://github.com/apollographql/router/pull/4441))

This uses a more flexible abstract type / concrete type check when applying a selection set to an entity reference before it's used in a fetch node. Previous to this change, we would drop data from the reference when it selected using an inline fragment (e.g. `@requires(fields: "... on Foo { a } ... on Bar { b }")`).

By [@lennyburdette](https://github.com/lennyburdette) in https://github.com/apollographql/router/pull/4441

### Improve logging for JWKS download failures ([Issue #4448](https://github.com/apollographql/router/issues/4448))

To enable users to debug JWKS download and parse failures more easily, we've added more detailed logging to the router. The router now logs the following information when a JWKS download or parse fails:

```
2024-01-09T12:32:20.174144Z ERROR fetch jwks{url=http://bad.jwks.com/,}  could not create JSON Value from url content, enable debug logs to see content e=expected value at line 1 column 1
```
Enabling debug logs via `APOLLO_LOG=debug` or `--logs DEBUG` will show the full JWKS content being parsed:
```
2024-01-09T12:32:20.153055Z DEBUG fetch jwks{url=http://bad.jwks.com/,}  parsing JWKS data="invalid jwks"
```

By [@BrynCooke](https://github.com/BrynCooke) in https://github.com/apollographql/router/pull/4449

### make OperationKind public ([Issue #4410](https://github.com/apollographql/router/issues/4410))

`OperationKind` was already used in a public field but the type itself was still private.

By [@Geal](https://github.com/Geal) in https://github.com/apollographql/router/pull/4489

### Pre-built binaries are only available for Apple Silicon ([Issue #3902](https://github.com/apollographql/router/issues/3902))

Prior to this release, macOS binaries were produced on Intel build machines in our CI pipeline. The binaries produced would also work on Apple Silicon (M1, etc.. chips) through the functionality provided by [Rosetta2](https://support.apple.com/en-gb/HT211861).

Our [CI provider has announced the deprecation of the macOS Intel build machines](https://discuss.circleci.com/t/macos-intel-support-deprecation-in-january-2024/48718) and we are updating our build pipeline to use the new Apple Silicon based machines.

This will have the following effects:
 - Older, Intel based, macOS systems will no longer be able to execute our macOS router binaries
 - Newer, Apple Silicon based, macOS systems will get a performance boost due to no longer requiring Rosetta2 support.

We have raised [an issue](https://github.com/apollographql/router/issues/4483) that describes options for Intel based macOS users. Please let us know in that issue if the alternatives we suggest (e.g., Docker, source build) don't work for you so we can discuss alternatives.

By [@garypen](https://github.com/garypen) in https://github.com/apollographql/router/pull/4484


abernix:
This PR isn't needed for an RC.  It's my fault the release checklist isn't intuitive enough for this to be clear, but if this merged right now it would break the release, for prerelease we just:

- run xtask prepare rc
- throw away `./.changesets/` and `CHANGELOG.md` changes
- `git tag -a "v${APOLLO_ROUTER_RELEASE_VERSION}" -m "${APOLLO_ROUTER_RELEASE_VERSION}" && git push "${APOLLO_ROUTER_RELEASE_GIT_ORIGIN}" "v${APOLLO_ROUTER_RELEASE_VERSION}"`

I'll fix it up.

PR #4517, opened 2024-01-18
log rate limiting

### LINKED ISSUE #4479: Error log rate limiting
in some cases we want to log an error, linked to a request, but we do not want to spam the logs with that error if the request is done in a loop. We have code in the telemetry plugin that does rate limiting for error messages, maybe we could reuse or extend it:

https://github.com/apollographql/router/blob/4027132ded0c401dd5b51d2c3cf7d944a5feb3f8/apollo-router/src/plugins/telemetry/mod.rs#L1808-L1867

A possible issue here is that if we don't send that error event, then it will not appear in the trace of the request that triggered it, which might make it harder to debug. So would it be a better idea to rate limit specific logs at the stdout logger level?

Geal:
Fix #4479 

we want to add more error logging, but worry that we will end up spamming the  log collectors if the errors happen on every request. There is also the worry that traces would become hard to debug if events are rate limited: we could end up with traces that are obviously wrong, but without the event that could give some context.

This implements a rate limiter for logs, at the formatter level, so that traces get their own events properly (traces can get sampled independently). I tested that behaviour by combining with the fix in #4486.

Todo: how do we configure it?
- [ ]  filter per log level?
- [x] configurable delay between logs?
- [ ] do we allow different filtering rates per module?
- [ ] the rate limiter would reset  on every configuration or schema update, do we consider that an issue?

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/plugins/telemetry/formatters/mod.rs
```diff
@@ -91,14 +101,32 @@ where
...
+            std::collections::hash_map::Entry::Occupied(mut last) => {
+                if now - *last.get() < Duration::from_secs(10) {
+                    return false;
+                }
+                last.insert(now);
+            }
+            std::collections::hash_map::Entry::Vacant(entry) => {
+                entry.insert(now);
+            }
+        }
```
bnjjj, line 125: Using callsite is the way to go I think.IIRC using entry is a less performant api for hashmap. I'm just worried about perfs here

Geal: I could do a get() and insert() instead, it's not a big difference

Geal: fixed in [346593a](https://github.com/apollographql/router/pull/4517/commits/346593ad690cb9db2bf86a6f1e9077fd88e77970)

Geal:
ok so now we have a config option to activate it, wt configurable amount and interval. This is not a rate limiter with a sliding window so it can be a bit bursty, but I wanted to keep it very simple since that could apply on every log line.
If we are rate limiting a log message, the next time it is written, a log attribute named `skipped_messages` is added, indicating how many logs were ignored:

```json
{
    "timestamp": "2024-01-18T17:34:44.011761506Z",
    "level": "ERROR",
    "message": "subgraph log",
    "target": "apollo_router::services::subgraph_service",
    "spans": [{
        "http.method": "POST",
        "http.request.method": "POST",
        "http.route": "/",
        "http.flavor": "HTTP/1.1",
        "name": "request"
    }, {
        "http.method": "POST",
        "http.request.method": "POST",
        "http.route": "/",
        "http.flavor": "HTTP/1.1",
        "trace_id": "c83f59a30e17341c8da0f3425dd56664",
        "url.path": "/",
        "client.name": "",
        "client.version": "",
        "name": "router"
    }, {
        "graphql.document": "query CachedQuery { me { username } topProducts { name reviews { body author { name } }  } }",
        "graphql.operation.name": "CachedQuery",
        "graphql.operation.name": "CachedQuery",
        "name": "supergraph"
    }, {
        "graphql.operation.type": "query",
        "name": "execution"
    }, {
        "name": "parallel"
    }, {
        "name": "sequence"
    }, {
        "apollo.subgraph.name": "products",
        "skipped_messages": 18541,
        "name": "fetch"
    }],
    "resource": {
        "service.name": "unknown_service:router",
        "service.version": "1.37.0",
        "process.executable.name": "router"
    }
}
```

apollo-router/src/plugins/telemetry/config_new/logging.rs
```diff
@@ -104,12 +105,36 @@ pub(crate) struct StdOut {
     pub(crate) enabled: bool,
     /// The format to log to stdout.
     pub(crate) format: Format,
+    /// Log rate limiting. The limit is set per type of log message
+    pub(crate) rate_limit: Option<RateLimit>,
```
BrynCooke, line 109: Ideally don't use Option here, you can see if the count is zero instead. Option in config often leads to weirdness that I think is documented in the config guidelines.

apollo-router/src/plugins/telemetry/config_new/logging.rs
```diff
@@ -104,12 +105,36 @@ pub(crate) struct StdOut {
...
+            rate_limit: None,
+        }
+    }
+}
+
+#[derive(Deserialize, JsonSchema, Clone, Debug)]
+#[serde(deny_unknown_fields, default)]
+pub(crate) struct RateLimit {
+    /// Number of log lines allowed in interval per message
+    pub(crate) count: u32,
```
BrynCooke, line 126: `count` is sort of non-descript.
Maybe `max_events` ? Still not great, but....

apollo-router/src/plugins/telemetry/config_new/logging.rs
```diff
@@ -104,12 +105,36 @@ pub(crate) struct StdOut {
...
+    /// Interval for rate limiting
+    #[serde(deserialize_with = "humantime_serde::deserialize")]
+    #[schemars(with = "String")]
+    pub(crate) interval: Duration,
```
BrynCooke, line 130: I think this should be `period` rather than `interval` because the events happen within a period.

BrynCooke (changes requested):
Looking nice!
Some minor config things to modify.

apollo-router/src/plugins/telemetry/config_new/logging.rs
Geal: in the traffic shaping's rate limiting we use "capacity", I'll use that to keep consistency

apollo-router/src/plugins/telemetry/config_new/logging.rs
Geal: in the traffic shaping's rate limiting we use "interval" too, I'll use that to keep consistency

apollo-router/src/plugins/telemetry/config_new/logging.rs
Geal: fixed in [5dc5aa2](https://github.com/apollographql/router/pull/4517/commits/5dc5aa25977a25d9164b9b2bc4628ce23bc924d5)

apollo-router/src/plugins/telemetry/config_new/logging.rs
Geal: I added an enabled field in [4318aaa](https://github.com/apollographql/router/pull/4517/commits/4318aaabb2854f672e3911e1b14bae231cc65f2c)

apollo-router/src/plugins/telemetry/config_new/logging.rs
BrynCooke: Suggest that we use the right terminology and create a migration for traffic shaping. No point in making things confusing for consistency.

Geal: let's leave that migration to a follow up PR then. I'm not sure I agree that changing it is that useful, because I have seen both terms used in rate limiters, but that discussion should not block this PR (I'd like to get this landed so we can add way more logs in the next release)

docs/source/configuration/telemetry/exporters/logging/stdout.mdx
```diff
@@ -42,6 +42,23 @@ telemetry:
          format: text #highlight-line
 ```
 
+### `rate_limit`
+
+Log messages ca become too frequent, especially when reporting errors when processing a request. To prevent the router from filling the logs with similar messages, it is possible to limit their rate. The rate is set per log location, which means that two different log messages are limited independently, but log lines with the same message with different sets of attributes fal l under the same rate. This rate limiting only applies to logging to stdout, and does not affect event sent to trace exporters like OTLP, which have their own sampling configuration.
```
shorgi, line 47:
```suggestion
The rate at which log messages are produced can become too high, especially for request processing errors. To prevent the router from filling its logs with redundant messages, you can use the `rate_limit` option to set the logging rate limit.

You can set the logging rate limit for each log location, where different log messages are rate limited independently, and log lines with the same message but different sets of attributes are limited under the same rate. This rate limiting only applies to logging to stdout and doesn't affect events sent to trace exporters like OTLP, which have their own sampling configuration.
```

o0Ignition0o (approved):
LGTM.

@Geal @BrynCooke I'm happy to see a followup that updates naming, should you both deem it appropriate. It could be done before we cut the next release.

Issue #4464, opened 2024-01-11
Helm VirtualService should support multiple Gateway names


marcantoine-bibeau:
**Describe the solution you'd like**
Helm VirtualService SHOURL support multiple Gateway names

**Describe alternatives you've considered**
in values file:
virtualservice:
  gatewayName: ""

should be: 

virtualservice:
  gatewayNames: []

**Additional context**

Geal:
would you have an idea of how to do that with our current chart? Is that something you would consider sending a PR for?

marcantoine-bibeau:
Sure! I can do that :)

marcantoine-bibeau:
@Geal I do not find how I can become a contributor. :(  I have the template changes but I cannot create a PR

Geal:
you can fork the project, make the change in your fork then create a PR

PR #4520, opened 2024-01-18
Helm update to allow a list of gateways to VirtualService

marcantoine-bibeau:
Allow to configure 0..N gateways in Istio VirtualService 

Fixes [#4464](https://github.com/apollographql/router/issues/4464)

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [x] Manual Tests

**Notes**

Results with no gateway configured:
```
...
spec:
  hosts:
  - "*"
```

Results with multiple gateways configured:
```
...
spec:
  hosts:
  - "*"
  gateways:
    - gateway-1
    - gateway-2
```

Results with previous (single) gateway configured:
```
...
spec:
  hosts:
  - "*"
  gateways:
    - gateway-1
```


marcantoine-bibeau:
FYI @Geal

garypen:
@marcantoine-bibeau This is a nice contribution, so thank you.

I only have once concern about the change. It would break anyone currently using `gatewayName`.

I think we could have a solution which checked if the existing mechanism was in use and, if not, then checked for the new mechanism. We could mark the old mechanism (via a comment) as deprecated and aim to remove it sometime in the future.

Would you be ok making that change? (Or a better idea if you have one).

marcantoine-bibeau:
@garypen Updated. I added a comment in the values file, I hope it is what you had in mind?

garypen:
> @garypen Updated. I added a comment in the values file, I hope it is what you had in mind?

That was exactly what I had in mind.

I just need to figure out why `helm check` is failing. In the meantime could you add a changeset entry (as per the description in .changesets/README.md) so that you are correctly credited for your contribution?

helm/chart/router/values.yaml
```diff
@@ -158,9 +158,12 @@ ingress:
 
 # set to true to enable istio's virtualservice
 virtualservice:
-  enabled: false
-  # namespace: ""
-  # gatewayName: ""
+  enabled: true
+  namespace: "default"
```
garypen, line 162: @marcantoine-bibeau I figured out that the reason our helm check is failing is that it attempts to validate all the components which are enabled in the `values.yaml` file. We don't want `virtualservice` to be enabled by default, since most users of this chart won't be using it. Suggested modification to your PR:
```suggestion
  enabled: false
  # namespace: ""
```

garypen (changes requested):

helm/chart/router/values.yaml
marcantoine-bibeau: Ha! My bad!

PR #4521, opened 2024-01-18
Progressive override support

trevor-scheer:
This change brings support for progressive `@override`, which allows dynamically overriding root fields and entity fields in the schema. This feature is enterprise only and requires a license key to be used.

Support for this is implemented via plugin and support for a new context key whose value is provided as options to query planning.

We expect users to interact with and resolve labels themselves via coprocessors, so we're providing an interface via a known context key in order to communicate to and from. The coprocessors need to know which labels are available in the schema and they need to tell the router which labels to set.

We also expect users to use a "special-case" label format `percent(x)` which the plugin will resolve out of the box.

All of this is to say, the plugin has a few responsibilities:
1) At the beginning of a request (`router_service`), the plugin will add all (non-percent) labels from the schema to the known context key `apollo_override::unresolved_labels`.
2) Coprocessors (during `supergraph_request`) may read from this context entry and append labels to be overridden to the `apollo_override::labels_to_override` context entry.
3) The plugin (during its `supergraph_request`) will resolve percentage-based labels and append to the `apollo_override::labels_to_override` context entry.
4) Additionally, the plugin will traverse the operation to obtain only the relevant labels for the given operation. It will then take the intersection of those labels and all labels in the `apollo_override::labels_to_override` context entry to ensure the set of labels is minimized, sorted, etc.

It's important to note that the set of overridden labels is an input to query planning, and thus becomes part of the cache key. For every relevant label in an operation, the number of potential QP cache entries doubles. Step 4 is important to ensure we don't create any more cache entries per operation than is necessary.

TODO:
- [x] Add launchdarkly coprocessor example in examples folder
- [x] More tests

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/spec/schema.rs
```diff
@@ -237,22 +238,45 @@ impl Schema {
...
+                    false
+                }
             })
     }
 
     pub(crate) fn directive_name(
         schema: &apollo_compiler::schema::Schema,
-        url: &str,
+        base_url: &str,
+        min_version: &str,
```
Geal, line 279: why only `min_version`? If we change a directive to add a field and expect a meaningful behaviour change from it, a router that does not know how to interpret that directive should reject it

apollo-router/src/uplink/license_enforcement.rs
```diff
@@ -271,7 +317,7 @@ impl Display for LicenseEnforcementReport {
             let restricted_schema = self
                 .restricted_schema_in_use
                 .iter()
-                .map(|v| format!("* {}\n  {}", v.name, v.url))
+                .map(|v| format!("{}", v))
```
Geal, line 320:
```suggestion
                .map(|v| v.to_string())
```

apollo-router/src/query_planner/caching_query_planner.rs
```diff
@@ -293,7 +312,9 @@ where
                 .get::<CacheKeyMetadata>()
                 .cloned()
                 .unwrap_or_default(),
+            plan_options,
         };
+        tracing::info!("CachingQueryPlanner.plan: caching_key: {:?}", &caching_key);
```
Geal, line 317:
```suggestion
```

apollo-router/src/query_planner/caching_query_planner.rs
```diff
@@ -282,6 +288,19 @@ where
...
+                .context
+                .get(LABELS_TO_OVERRIDE)
+                .unwrap_or_default()
+                .unwrap_or_default(),
+        };
+
+        tracing::info!(
+            "CachingQueryPlanner.plan: plan_options: {:?}",
+            &plan_options
+        );
```
Geal, line 302:
```suggestion
```

apollo-router/src/query_planner/bridge_query_planner.rs
```diff
@@ -553,13 +556,23 @@ impl Service<QueryPlannerRequest> for BridgeQueryPlanner {
...
             }
 
+            let plan_options = PlanOptions {
+                override_conditions: context
+                    .get(LABELS_TO_OVERRIDE)
+                    .unwrap_or_default()
+                    .unwrap_or_default(),
+            };
+
+            tracing::info!("BridgeQueryPlanner: plan_options: {:?}", &plan_options);
```
Geal, line 566:
```suggestion
```

apollo-router/src/plugins/progressive_override/mod.rs
```diff
@@ -0,0 +1,246 @@
...
+    tracing::debug!("static_percentages: {:?}", &static_percentages);
+    (static_percentages, other_labels)
+}
+
+#[async_trait::async_trait]
+impl Plugin for ProgressiveOverridePlugin {
+    type Config = Config;
+
+    async fn new(init: PluginInit<Self::Config>) -> Result<Self, BoxError> {
+        let schema = Schema::parse(&*init.supergraph_sdl, "schema.graphql").expect(
```
Geal, line 132: not an issue with this PR, but we should make the parsed schema available to plugins, at the very least for internal plugins

apollo-router/src/plugins/progressive_override/mod.rs
```diff
@@ -0,0 +1,246 @@
...
+                    .get::<_, Vec<Arc<String>>>(LABELS_TO_OVERRIDE)
+                    .unwrap_or_default()
+                    .unwrap_or_default();
+
+                if let Some(parsed_doc) = request.context.extensions().lock().get::<ParsedDocument>() {
+                    // we have to visit the operation to find out which subset
+                    // of labels are relevant unless we've already cached that
+                    // work
+                    let relevant_labels = labels_per_operation_cache
+                        .entry(Arc::clone(parsed_doc))
```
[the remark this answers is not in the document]
trevor-scheer: sgtm!
b2e22a8

apollo-router/src/spec/schema.rs
trevor-scheer: Let me know what you think here
[9803b4b](https://github.com/apollographql/router/pull/4521/commits/9803b4bc9ad07b7783f362b66d148fd4500bb567)

apollo-router/src/uplink/license_enforcement.rs
trevor-scheer: [8efecd3](https://github.com/apollographql/router/pull/4521/commits/8efecd3da1b77c7887b1edc8c9b0f8a038a57c11)

apollo-router/src/plugins/progressive_override/mod.rs
trevor-scheer: I took a stab at this, but happy to revert if it belongs in a different PR. Let me know what you think.
[5fd4a48](https://github.com/apollographql/router/pull/4521/commits/5fd4a488437b01f3eccdf9a870797985fda761f6)

apollo-router/src/plugins/progressive_override/mod.rs
```diff
@@ -0,0 +1,273 @@
...
+    // We have to visit each operation to find out which labels from the schema
+    // are relevant for any given operation. This allows us to minimize the
+    // number of labels we ultimately send to the query planner. Since these
+    // labels are a component of the query plan cache key, it's important we
+    // don't "overprovide" any labels, since doing so can explode the number of
+    // cache entries per operation.
+    labels_per_operation_cache: Arc<DashMap<String, Vec<Arc<String>>>>,
+}
+
+type LabelsFromSchema = (HashMap<Arc<String>, Arc<f64>>, HashSet<Arc<String>>);
```
Geal, line 51:
```suggestion
type LabelsFromSchema = (Arc<HashMap<Arc<String>>, Arc<f64>>, Arc<HashSet<Arc<String>>>);
```
more refcounting, because `HashMap` and `HashSet` can make a lot of allocations for their internal accounting

examples/coprocessor-override-launchdarkly/src/index.ts
```diff
@@ -0,0 +1,47 @@
...
+    return Math.random() * 100 < flagValue;
+  });
+
+  context.entries[LABELS_TO_OVERRIDE_CONTEXT_KEY] = [
+    ...(context.entries[LABELS_TO_OVERRIDE_CONTEXT_KEY] ?? []),
+    ...labelsToOverride,
+  ];
+
+  res.json({ context, ...rest });
+});
```
Geal, line 43: really nice, that's a great example to support this

apollo-router/src/uplink/license_enforcement.rs
```diff
@@ -89,6 +90,7 @@ impl LicenseEnforcementReport {
         !self.restricted_config_in_use.is_empty() || !self.restricted_schema_in_use.is_empty()
     }
 
+    // TODO: trevor to add
```
Geal, line 93: ?

apollo-router/src/uplink/license_enforcement.rs
```diff
@@ -240,14 +281,19 @@ impl LicenseEnforcementReport {
...
-                .name("@authenticated")
-                .url("https://specs.apollo.dev/authenticated/v0.1")
-                .build(),
-            SchemaRestriction::builder()
-                .name("@requiresScopes")
-                .url("https://specs.apollo.dev/requiresScopes/v0.1")
-                .build(),
+            SchemaRestriction::Directive {
+                name: "@authenticated".to_string(),
+                url: "https://specs.apollo.dev/authenticated/v0.1".to_string(),
```
Geal, line 286: a topic for a follow up PR: should the schema restrictions also work on a version range? Or a URL prefix like `https://specs.apollo.dev/authenticated/`?

Geal (approved):

apollo-router/src/spec/schema.rs
abernix: Is this implementation meant to align with the [Versioning](https://specs.apollo.dev/core/v0.2/#sec-Versioning) section of the Core specification?  (I started an internal conversation about [`link`](https://specs.apollo.dev/link/v1.0/#sel-EADHNCAACKBrkP) vs `core`, but for the purposes of this PR, I am mostly just questioning whether — for review purposes — the _intent_ is for this to match that implementation as defined in the `core` spec, irregardless of the overall spec-status itself.)

apollo-router/src/plugins/progressive_override/mod.rs
trevor-scheer: This was reverted based on a conversation we had out of thread

apollo-router/src/plugin/mod.rs
```diff
@@ -66,6 +67,8 @@ pub struct PluginInit<T> {
     pub config: T,
     /// Router Supergraph Schema (schema definition language)
     pub supergraph_sdl: Arc<String>,
+    /// Parsed Supergraph Schema
+    pub supergraph_schema: Arc<Schema>,
```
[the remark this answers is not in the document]
trevor-scheer: This addition was reverted altogether [eb9fa1a](https://github.com/apollographql/router/pull/4521/commits/eb9fa1a908ed8892bc760c426ac523e3214ba980)

apollo-router/src/plugin/mod.rs
```diff
@@ -125,9 +132,13 @@ where
         supergraph_sdl: Arc<String>,
         notify: Notify<String, graphql::Response>,
     ) -> Self {
+        let supergraph_schema = Arc::new(
```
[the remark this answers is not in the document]
trevor-scheer: Reverted this change in [eb9fa1a](https://github.com/apollographql/router/pull/4521/commits/eb9fa1a908ed8892bc760c426ac523e3214ba980)

apollo-router/src/uplink/license_enforcement.rs
trevor-scheer: Probably just a very early comment from before I had implemented any enforcement

apollo-router/src/uplink/license_enforcement.rs
trevor-scheer: Thanks for pointing this out. I resolved a couple related bugs and implemented the version range in a separate PR.
https://github.com/apollographql/router/pull/4570

examples/coprocessor-override-launchdarkly/src/index.ts
```diff
@@ -0,0 +1,47 @@
...
+
+let flagValues: Record<string, number> = {};
+listenForFlagUpdates((name, value) => {
+  flagValues[name] = value;
+});
+
+const app = express();
+app.use(express.json());
+app.post("/", async (req, res) => {
+  const { context, body, ...rest } = req.body;
```
trevor-scheer, line 21:
```suggestion
  const { context, ...rest } = req.body;
```

PR #4528, opened 2024-01-20
Carry static page in Bytes structure to avoid expensive allocation per request

xuorig:
I noticed the `StaticPageLayer` causes a non-insignificant amount of memory to be allocated on every request. The service stack gets cloned on every request, and so does the `CheckpointService` created by the `StaticPageLayer`, even on non-static-page requests.

This is only expensive because templates are pre-rendered and kept on the service struct itself, this relatively large string then gets cloned on every request.

~This pull request addresses this issue by instead moving the rendering only on static page requests. Of course, this carries a cost at every request, but only on less important (relative to service router requests) requests. Instead of the full string, a small enum is cloned that contains the necessary configuration to render the pages.~

To @Geal 's suggestion, I've update the PR to store the rendered template in a `Bytes` struct instead.

~I did not investigate a way to cache those outside the services yet, although it could certainly be possible. Looking to get some feedback on the report / idea before I investigate further.~ Thanks!


xuorig:
@Geal Modified to use `Bytes`. Not sure what is going on with the CI checks. added a changelog too.

Geal (approved):

goto-bus-stop (approved):
sweet, nice find

PR #4195, opened 2023-11-13
Entity caching: docs

Geal:
Fix #4478

---

**Checklist**

Complete the checklist (and note appropriate exceptions) before the PR is marked ready-for-review.

- [ ] Changes are compatible[^1]
- [ ] Documentation[^2] completed
- [ ] Performance impact assessed and acceptable
- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[^1]: It may be appropriate to bring upcoming changes to the attention of other (impacted) groups. Please endeavour to do this before seeking PR approval. The mechanism for doing this will vary considerably, so use your judgement as to how and when to do this.
[^2]: Configuration is an important part of many changes. Where applicable please try to document configuration examples.
[^3]: Tick whichever testing boxes are applicable. If you are adding Manual Tests, please document the manual testing (extensively) in the Exceptions.

PR #4534, opened 2024-01-23
Redis: add a fail open option when starting the Router

### LINKED ISSUE #4334: Entity cache: option to allow starting without a valid Redis connection
right now, if we get an error when inserting or retrieving data from redis, the query will still be executed normally. But if we can't contact redis on startup, the router won't start. It could be useful to have an option for a fail open start, where the router will still start and serve traffic, even if redis is not available

Geal:
Fix #4334

This option allows the router to start even if we cannot connect to Redis.

Needs opinions on the option name

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/configuration/snapshots/apollo_router__configuration__tests__schema_generation.snap
```diff
@@ -1322,6 +1322,11 @@ expression: "&schema"
           "type": "boolean",
           "nullable": true
         },
+        "fail_open": {
```
bnjjj, line 1325: `required_to_start` ?

bnjjj (commented):
Missing documentation and a configuration example in the changelog

apollo-router/src/router_factory.rs
```diff
@@ -150,11 +153,80 @@ impl RouterSuperServiceFactory for YamlRouterFactory {
...
+        Self.inner_create(
+            configuration,
+            schema,
+            previous_router,
+            initial_telemetry_plugin,
+            extra_plugins,
+        )
+        .instrument(router_span)
+        .await
+    }
```
bnjjj, line 207: Is it part of this PR ?

bnjjj (commented):
I think it includes code from another PR (span for plugin init)

apollo-router/src/configuration/snapshots/apollo_router__configuration__tests__schema_generation.snap
lleadbet: ++ I like this since `fail_open` could also apply to post-connection settings (e.g. not cause errors when inserting on entity cache updates or APQs or the like).

Geal: fixed in  8648a87e1

apollo-router/src/configuration/snapshots/apollo_router__configuration__tests__schema_generation.snap
```diff
@@ -2034,6 +2034,11 @@ expression: "&schema"
...
+          "description": "Allow the router to start without a connection to Redis. Option name TBD",
+          "default": false,
+          "type": "boolean"
+        },
```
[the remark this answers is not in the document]
Geal: ok with the suggestion in https://github.com/apollographql/router/pull/4534#pullrequestreview-1838831936? (it was changed just before your comment)

Geal:
I think I won't add a changeset, as it will be released along with https://github.com/apollographql/router/pull/4195, but I'll add a doc entry once #4195 is merged

bnjjj (approved):

Geal:
thinking a bit more: should we make this a global option on redis configuration, instead of just the entity cache? Right now the APQ and Query plan cache will log an error but still start if Redis is not available, do we want the option to have this block the router startup as well?

lleadbet:
@Geal - we absolutely should imo

Geal:
ok, this is now an option for all Redis uses. I confirmed that entity caching will not fail requests if it cannot connect to Redis (both at startup, and later if redis is suddenly down)

apollo-router/src/configuration/snapshots/apollo_router__configuration__tests__schema_generation.snap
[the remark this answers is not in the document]
Geal: the default is false because this is how it works right now for APQ and query planning

PR #4535, opened 2024-01-23
do not propagate Accept, Accept-Encoding and Content-Encoding headers

### LINKED ISSUE #4312: bug: multiple header propagation rules cause duplicate header values
**Describe the bug**

When using a router configuration that matches more than one on a header name for propagation (e.g. named and matching), the Router will duplicate those headers as many times as there are matching rules.

**To Reproduce**
1. Run a router with the following config:

```
headers:
  all:
    request:
    - propagate:
        named: x-my-header
    - propagate:
        matching: ".*"
telemetry:
  experimental_logging:
    when_header:
      - name: x-my-header
        match: .*
        headers: true
```

3. Send a request that contains one `x-my-header` value
4. See that the subgraph receives two `x-my-header` entries with the same value.

**Expected behavior**

The Router should not duplicate header values, but should conserve multi-value headers when configured to propagate them to a subgraph, no matter if a header name matches mutliple entries.

**Output**

N/A

 - OS: macOS
 - Version 1.34.1

**Additional context**

The [current logic](https://github.com/apollographql/router/blob/dev/apollo-router/src/plugins/headers.rs#L296) goes through all header propagation first, then scan all headers in a request, which causes the duplication.

While this applies for propagation, we would need to decide on a strategy for prioritising propagation vs insertion vs removal.

### LINKED ISSUE #4398: headers set by the subgraph service should not be propagated from the client query by the headers plugin
**Describe the bug**
Requests that already have the accept header set to the value `application/json, application/graphql-response+json` on the inbound request to the router are getting multiple accept headers appended to the request incorrectly. 

In other words if the original request contains this:

```
accept: application/json, application/graphql-response+json
```

The router ends up sending this to the subgraph:

```
accept: application/json, application/graphql-response+json
accept: application/json
accept: application/graphql-response+json
```

Unfortunately, the default behavior of the C# GraphQL.Client is to set the accept headers exactly this way : https://github.com/graphql-dotnet/graphql-client/blob/fbf37dfff2830e36a45296079bac27ec60fad550/src/GraphQL.Client/GraphQLHttpRequest.cs#L40

This behavior started in v1.35.0 of the router, and seems to me to be attributable to this changeset: https://github.com/apollographql/router/commit/9a44bedd070dcf66dc0c1e63b4645033bc5262e8#diff-6ef5a208ca8622f30eef88f75c18566e0304d59856b66293dcd6811555e6382e which changed the method used to set the Accept header from an .insert() to an .append(). The older behavior of just using .insert() avoided the issue because it simply overwrote the entire accept header, which is probably not the desired behavior either.

**To Reproduce**
1. Setup services normally (using router or cloud router v1.35.0)
2. Submit a request including the header `Accept: application/json, application/graphql-response+json`
3. View the request coming out of the router, it is incorrectly appending additional headers when it should not be.

**Expected behavior**
The router should see that the headers it is trying to set are already there and should do nothing in that scenario.

**Additional context**
This issue causes breakage with subgraphs running inside of lambdas behind ALBs in AWS because of how request headers are marshaled when they are converted from an http request to a lambda request. AWS lambda does not properly marshal the requests when they contain both multivalue headers AND multiple of the same header key (it works with either or, but not both) . 

We are currently working around this issue by overriding the default behavior of the C# GraphQL.Client and stripping the Accept header off before the request goes out. Not a great solution obviously.

Geal:
Fix #4312
Fix #4398

Potential regression caused by this PR: the only way to activate subgraph request compression right now is to have the `Content-Encoding` header set. This header was not part of the reserved headers before, and it is now, so if a router was configured to propagate that header (either explicitely or by overly large matching), then now the subgraph request will not be compressed.
I do not think it is a good idea to have subgraph compression driven by the headers. That header should instead be set by the subgraph service directly, from an option set in configuration

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/.changesets/fix_geal_headers_propagation_filter.md
+++ b/.changesets/fix_geal_headers_propagation_filter.md
@@ -1,0 +1,18 @@
+### Fix header propagation issues ([Issue #4312](https://github.com/apollographql/router/issues/4312)), ([Issue #4398](https://github.com/apollographql/router/issues/4398))
+
+This fixes two header propagation issues:
+* if a client request header has already been added to a subgraph request due to another header propagation rule, then it is only added once
+* `Accept`, `Accept-Encoding` and `Content-Encoding` were not in the list of reserved headers that cannot be propagated. They are now in that list because those headers are set explicitely by the Router in its subgraph requests
+
+There is a potential regression: if a router deployment was accidentally relying on header propagation to compress subgraph requests, then it will not work anymore because `Content-Encoding` is not propagated anymore. Instead it should be set up from the `traffic_shaping` section of the Router configuration:
+
+```yaml
+traffic_shaping:
+  all:
+    compression: gzip
+  subgraphs: # Rules applied to requests from the router to individual subgraphs
+    products:
+      compression: identity
+```
+
+By [@Geal](https://github.com/Geal) in https://github.com/apollographql/router/pull/4535
--- a/apollo-router/src/configuration/snapshots/apollo_router__configuration__tests__schema_generation.snap
+++ b/apollo-router/src/configuration/snapshots/apollo_router__configuration__tests__schema_generation.snap
@@ -8017,6 +8017,13 @@ expression: "&schema"
                   "enum": [
                     "br"
                   ]
+                },
+                {
+                  "description": "identity",
+                  "type": "string",
+                  "enum": [
+                    "identity"
+                  ]
                 }
               ],
               "nullable": true
@@ -8185,6 +8192,13 @@ expression: "&schema"
                     "type": "string",
                     "enum": [
                       "br"
+                    ]
+                  },
+                  {
+                    "description": "identity",
+                    "type": "string",
+                    "enum": [
+                      "identity"
                     ]
                   }
                 ],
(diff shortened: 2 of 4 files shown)


bnjjj (commented):
I think we should add a changelog and mention the potential breaking change

.changesets/fix_geal_headers_propagation_filter.md
```diff
@@ -0,0 +1,18 @@
...
+
+There is a potential regression: if a router deployment was accidentally relying on header propagation to compress subgraph requests, then it will not work anymore because `Content-Encoding` is not propagated anymore. Instead it should be set up from the `traffic_shaping` section of the Router configuration:
+
+```yaml
+traffic_shaping:
+  all:
+    compression: gzip
+  subgraphs: # Rules applied to requests from the router to individual subgraphs
+    products:
+      compression: identity
```
bnjjj, line 15: does it work ?

Geal: now it does :innocent: [864f55c](https://github.com/apollographql/router/pull/4535/commits/864f55c15ffac8362b60e33be36133d6570178e9)

o0Ignition0o (approved):

o0Ignition0o:
> I think we should add a changelog and mention the potential breaking change

It is a bugfix IMO, I don't see it as a /breaking/ change.

PR #4536, opened 2024-01-23
use nix for deterministic builds

nicholascioli:
This is a draft PR to explore using nix for deterministic builds and developer environments. This should allow for new developers to onboard much faster and have an overall more consistent developer experience while also acting as a sanity check that builds can be made successfully deterministically.

Further research will need to be done into the following:
- Use of more advanced nix flakes, such as [flakebox](https://github.com/rustshop/flakebox), to generate cross-compiled versions of the router for all supported architectures
- Use of a binary cache, such as [cachix](https://www.cachix.org/), for supply chain attack mitigation
- CI/CD that uses the same environment as defined in the nix flake.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


flake.nix
```diff
@@ -0,0 +1,76 @@
...
+          # TODO: Why are the following frameworks required when we aren't hooking into
+          # MacOS internals?
+          Security
+          SystemConfiguration
+        ]);
+    in {
+      # Expose the main router as a buildable package
+      packages.default = rustPlatform.buildRustPackage {
+        pname = "router";
+        version = "1.38.0";
```
bnjjj, line 39: If we specify the version here then we should also add something into xtask for the release prep because we're doing some batch edits when upgrading the version to be sure it's upgraded too

nicholascioli: Yeah, I've seen other projects move the version number into its own file and do some shenanigans to include it in Cargo for this reason.

abernix:
I'll close this for now as I don't think we're able to commit to this at this time.  We can always re-open this if we want to revisit it.

PR #4547, opened 2024-01-25
Add endpoint callback

BrynCooke:
Although part of the public API, this is not intended for use by end users, and may change at any time.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/axum_factory/axum_http_server_factory.rs
+++ b/apollo-router/src/axum_factory/axum_http_server_factory.rs
@@ -458,11 +458,12 @@ static ENDPOINT_CALLBACK: OnceLock<Arc<dyn Fn(Router) -> Router + Send + Sync>> 
 /// Set a callback that may wrap or mutate `axum::Router` as they are added to the main router.
 /// Although part of the public API, this is not intended for use by end users, and may change at any time.
 #[doc(hidden)]
-pub fn set_axum_router_callback(callback: impl Fn(Router) -> Router + Send + Sync + 'static) {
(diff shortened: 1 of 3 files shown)


apollo-router/src/axum_factory/axum_http_server_factory.rs
```diff
@@ -433,14 +434,37 @@ where
...
     let listener = configuration.supergraph.listen.clone();
     Ok(ListenAddrAndRouter(listener, route))
 }
 
+static ENDPOINT_CALLBACK: OnceLock<Arc<dyn Fn(Router) -> Router + Send + Sync>> = OnceLock::new();
+
+/// Set a callback that may wrap or mutate `axum::Router` as they are added to the main router.
+/// Although part of the public API, this is not intended for use by end users, and may change at any time.
+#[doc(hidden)]
+pub fn set_axum_router_callback(callback: impl Fn(Router) -> Router + Send + Sync + 'static) {
```
o0Ignition0o, line 461: can we add `unchecked` or `unsafe` to the function name? I think we should also document it may panic if called twice, etc...

BrynCooke: [6fd7990](https://github.com/apollographql/router/pull/4547/commits/6fd79909e40548c9b81b61ce73ded2de526362e4)

o0Ignition0o (approved):

PR #4510, opened 2024-01-17
Update apollo-rs and apollo-federation with validation fix and experimental API schema

goto-bus-stop:
Closes https://github.com/apollographql/router/pull/4525
Closes https://github.com/apollographql/router/pull/4413

apollo-rs now stores objects and interfaces separately in `implementers_map`, I added a `.implementors()` method in the visitors that rely on the implementers map so they don't need further code changes. Spelling matching the respective convention in apollo-rs vs apollo-router :)

apollo-rs also includes several fixes to operation validation (namely field merging and enum values as input to custom scalars).

`to_api_schema` is now fully implemented. Two changes on the router side:
- It can error, so that bubbles up through `ServiceBuildError` now.
- The `experimental_api_schema: both` option now handles the case where Rust API schema fails while JS API schema succeeds, previously it caused the router to exit, now it emits a metric and uses the JS result.

**Todo:**
- [x] apollo-federation API schema needs an option to add `@defer`
- [x] Merge API schema in apollo-federation
- [x] Release apollo-federation
- [x] Tweak the API schema error cases so we report a metric when both JS and RS produced an error.

---

**Checklist**

Complete the checklist (and note appropriate exceptions) before the PR is marked ready-for-review.

- [ ] Changes are compatible[^1]
- [ ] Documentation[^2] completed
- [ ] Performance impact assessed and acceptable
- Tests added and passing[^3]
    - [ ] Unit Tests
    - [x] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[^1]: It may be appropriate to bring upcoming changes to the attention of other (impacted) groups. Please endeavour to do this before seeking PR approval. The mechanism for doing this will vary considerably, so use your judgement as to how and when to do this.
[^2]: Configuration is an important part of many changes. Where applicable please try to document configuration examples.
[^3]: Tick whichever testing boxes are applicable. If you are adding Manual Tests, please document the manual testing (extensively) in the Exceptions.

PR #4551, opened 2024-01-26
Remove legacy validation

### LINKED ISSUE #4407: validation: default experimental_graphql_validation_mode to "new"
This should switch all router users to use only Rust validation. This is last step before entirely removing `experimental_graphql_validation_mode` config.

### LINKED ISSUE #4409: validation: remove experimental_graphql_validation_mode config
Completely transition to Rust validation in the router. These changes to include:
- [ ] set `graphql_validation`  to false when using `QueryPlannerConfig`
- [ ] remove metrics and traces that compare js vs rust validation errors as js errors do not exist anymore
- [ ] simplify validation paths for executable documents:
    - [ ] `ParsedDocumentInner` to work directly with `Valid<ExecutableDocument>` after an operation is received as part of the response
    - [ ] `QueryAnalysisLayer` should run overall validation
    - [ ] transition parsed schema to also use `Valid<Schema>`
    - [ ] verify the extra reparsing in `bridge_query_planner.rs` for filtered_query (is it necessary? can we reuse the existing `Valid<ExecutableDocument>`

Geal:
Fix #4407 
Fix #4409
This removes GraphQL validation from the query planner, to use the Rust version instead. Validation has now moved to the query analysis layer, which means we can remove a lot of code that was there to accommodate parsing and validating the query but not rejecting it outright, because we needed to compare the validation result with the planner's.
This will greatly reduce the load on the planner, which will now only be used for planning queries, not validating.
This new validation process has been running in production for months concurrently with the JavaScript version, allowing us to detect and fix any discrepancies in the new implementation. We now have enough confidence in the new Rust-based validation to entirely switch off the less performant, JavaScript validation.

Remaining issues:

- [x] usage reporting: since invalid queries are rejected in query analysis, inside the router service, then usage reporting for invalid queries is no longer reported, because the telemetry plugins handles usage reporting at the supergraph response level
- [x] persisted queries: apparently persisted queries use the query analysis transformation of supergraph requests, now one test fails on an invalid query, I have not yet looked at why
- [x] `@defer` directive: a test is failing because the directive is used in the query but not found by the schema. @lrlna what is apollo-compiler expecting here? Is `@defer` considered builtin, or does it need to be declared in the schema?

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


lrlna:
> @defer directive: a test is failing because the directive is used in the query but not found by the schema. @lrlna what is apollo-compiler expecting here? Is @defer considered builtin, or does it need to be declared in the schema?

@Geal we do not define it as a built-in in apollo-compiler. The supergraphs that support defer, will have a defer definition already defined. And it looked like a bunch of tests <sup>[[1]](https://github.com/apollographql/router/blob/2c8115d7b67253686871473c4c4fe93b53476ad1/apollo-router/src/plugins/authorization/scopes.rs#L1180)</sup> <sup>[[2]](https://github.com/apollographql/router/blob/2c8115d7b67253686871473c4c4fe93b53476ad1/apollo-router/src/plugins/authorization/scopes.rs#L1180)</sup> <sup>[[3]](https://github.com/apollographql/router/blob/2c8115d7b67253686871473c4c4fe93b53476ad1/apollo-router/src/plugins/authorization/policy.rs#L1168)</sup> in the router already had them defined as part of the schema. I assume this test also needs it defined.

Geal:
alright, that's an easy fix then

.changesets/feat_geal_remove_legacy_validation.md
```diff
@@ -0,0 +1,6 @@
+### Remove legacy validation ([PR #4551](https://github.com/apollographql/router/pull/4551))
+
+GraphQL query validation was initially performed by the query planner, which resulted in some performance issues. We designed a new validation process using ApolloCompiler from the apollo-rs project, that runs earlier in the request handling pipeline, at the router service level instead of the supergraph service.
+That validtion system has been running in production for months concurrently with the Javascript version, to detect any discrepancy between their behaviors. This built enough confidence in the new implementation that we are now entirely moving to it and removing the Javascript validation.
```
lrlna, line 4:
```suggestion
GraphQL query validation was initially performed by the query planner in JavaScript, which caused some performance issues. Here, we are introducing a new Rust-based validation process using `apollo-compiler` from the `apollo-rs` project.

This new validation process has been running in production for months concurrently with the JavaScript version, allowing us to detect and fix any discrepancies in the new implementation. We now have enough confidence in the new Rust-based validation to entirely switch off the less performant, JavaScript validation.
```

apollo-router/src/error.rs
```diff
@@ -582,6 +562,37 @@ impl IntoGraphQLErrors for ValidationErrors {
...
+                    .build()
+            })
+            .collect())
+    }
+}
+
+/// Collection of schema validation errors.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub(crate) struct ValidationErrors {
+    pub(crate) errors: Vec<apollo_compiler::execution::GraphQLError>,
```
lrlna, line 575: Should this use `GraphQLError` or `DiagnosticsList`? 🤔  cc @SimonSapin

lrlna (commented):
yesssss! tysm @Geal

apollo-router/src/error.rs
Geal: we need something that implements `Serialize` and `Deserialize` so I used `GraphQLError`

SimonSapin: For that purpose either works. `DiagnosticList` can be converted to `Vec<GraphQLError>` either when constructing this struct or later when using it. Doing it later allows printing with nice CLI formatting in `dbg!`, but requires carrying a `SourceMap` around somehow.

apollo-router/src/query_planner/bridge_query_planner.rs
```diff
@@ -657,17 +527,12 @@ impl BridgeQueryPlanner {
         if let Some((unauthorized_paths, new_doc)) = filter_res {
             key.filtered_query = new_doc.to_string();
             let executable = new_doc
-                .to_executable(&self.schema.api_schema().definitions)
+                .to_executable_validate(&self.schema.api_schema().definitions)
                 // Assume transformation creates a valid document: ignore conversion errors
-                .unwrap_or_else(|invalid| invalid.partial);
+                .unwrap();
```
BrynCooke, line 532: Unwrap here, is this safe?

apollo-router/src/services/layers/query_analysis.rs
```diff
@@ -104,68 +108,89 @@ impl QueryAnalysisLayer {
...
+                    }
+                    Ok(doc) => {
+                        let context = Context::new();
+
+                        let operation = doc.executable.get_operation(op_name.as_deref()).ok();
+                        let operation_name = operation.as_ref().and_then(|operation| {
+                            operation.name.as_ref().map(|s| s.as_str().to_owned())
+                        });
+
+                        context.insert(OPERATION_NAME, operation_name).unwrap();
```
BrynCooke, line 133:
```suggestion
                        context.insert(OPERATION_NAME, operation_name).expect("cannot insert operation name into context; this is a bug");
```

apollo-router/src/query_planner/bridge_query_planner.rs
Geal: we assume that the operations we generate are valid, but just in case I'll raise an error here

Geal: bf7669828

BrynCooke (approved):

lrlna (approved):

lrlna:
As an FYI, I know @goto-bus-stop had a change they need to run in our dual mode for a couple of days before we strip this out completely.

Geal:
alright, I'm holding off on merging it for now

goto-bus-stop:
We still need to approve & release the apollo-compiler validation fix and get it out in a public router release (just a patch version). Simon would be the best to review it but he's out today, I expect we'll have to wait a week before we can remove legacy validation entirely

Geal:
updated to the latest dev now that #4510 is merged

lrlna:
@Geal let me take a look at the metrics real quick

lrlna:
@geal we haven't had any hiccups in the metrics since `1.40.1` released on monday. @goto-bus-stop the harness was matching with both implementations, yea?

goto-bus-stop:
I am not sure if we have metrics from the new version yet specifically from the operations that were affected before (it's at least not in the nightly analytics rollup data yet). If we do, then we are good to go

goto-bus-stop:
The option should also be removed from documentation: https://github.com/apollographql/router/blob/23656eac8beba4458f3e994538cf297403edc73c/docs/source/configuration/overview.mdx#L801-L811

Geal:
@goto-bus-stop nice catch, thanks [746af65](https://github.com/apollographql/router/pull/4551/commits/746af65f505a88169700f18eb4667c5f6884c687)

apollo-router/src/query_planner/tests.rs
```diff
@@ -420,7 +421,12 @@ async fn defer_if_condition() {
           }"#;
 
     let schema = include_str!("testdata/defer_clause.graphql");
-    let schema = Arc::new(Schema::parse_test(schema, &Default::default()).unwrap());
+    // we need to use the planner here instead of Schema::parse_test because that one uses the router bridge's api_schema function
+    // does not keep the defer directive definition
```
SimonSapin, line 425: Is this still relevant after https://github.com/apollographql/router/pull/4701 ?

apollo-router/src/query_planner/bridge_query_planner.rs
```diff
@@ -579,18 +452,15 @@ impl Service<QueryPlannerRequest> for BridgeQueryPlanner {
                     )))
                 }
                 Ok(modified_query) => {
-                    let executable_document = modified_query
-                        .to_executable(schema)
-                        // Assume transformation creates a valid document: ignore conversion errors
-                        .unwrap_or_else(|invalid| invalid.partial);
+                    let executable_document =
+                        modified_query.to_executable_validate(schema).map_err(|e| {
```
SimonSapin, line 456: Validation is hopefully faster now but still, do we want to run it 2~3 times per query?

goto-bus-stop: Note also that some transformations do not produce valid documents today (an example in https://github.com/apollographql/router/pull/4938)

apollo-router/src/query_planner/tests.rs
goto-bus-stop: Yeah it should be possible to go back to using `parse_test` here as it uses rust api schema now

apollo-router/src/plugins/demand_control/basic_cost_calculator.rs
```diff
@@ -127,9 +127,17 @@ impl BasicCostCalculator {
...
+            return Err(DemandControlError::QueryParseFailure(format!(
+                "Cannot cost {} operation because the schema does not support this root type",
+                operation.operation_type
+            )));
+        };
+
         cost += BasicCostCalculator::score_selection_set(
             &operation.selection_set,
-            operation.name.as_ref(),
+            root_type_name,
```
goto-bus-stop, line 140: this was just a small mistake, the operation name isn't a type name, the type is always the root type (eg. for any `query` it's always `Query`, or whatever the schema has specified)

apollo-router/src/plugins/telemetry/metrics/snapshots/apollo_router__plugins__telemetry__metrics__apollo__test__apollo_metrics_multiple_operations.snap
```diff
@@ -2,43 +2,4 @@
 source: apollo-router/src/plugins/telemetry/metrics/apollo.rs
 expression: results
 ---
-[
-  {
-    "request_id": "[REDACTED]",
-    "stats": {
-      "## GraphQLValidationFailure\n": {
```
SimonSapin, line 9: Do we want to replace this metric with something else?

apollo-router/src/plugins/progressive_override/tests.rs
```diff
@@ -136,16 +133,10 @@ async fn assert_expected_and_absent_labels_for_supergraph_service(
...
-    // it ourselves for testing purposes
-    let schema = Schema::parse_and_validate(SCHEMA, "").unwrap();
-    let document = Document::parse(query, "query.graphql").unwrap();
-    let executable = document.to_executable(&schema).unwrap();
-    let parsed_doc: ParsedDocument = Arc::from(ParsedDocumentInner {
-        ast: document,
-        executable: Arc::new(executable),
-        ..Default::default()
-    });
+    let schema = crate::spec::Schema::parse(include_str!("./testdata/supergraph.graphql")).unwrap();
```
goto-bus-stop, line 136:
```suggestion
    let schema = crate::spec::Schema::parse_test(include_str!("./testdata/supergraph.graphql"), Default::default()).unwrap();
```
We should use `parse_test` everywhere in tests to ensure that the queries are pointing to the API schema rather than the supergraph schema, as they do in a running router

apollo-router/src/plugins/progressive_override/tests.rs
```diff
@@ -211,14 +202,10 @@ async fn plugin_supergraph_service_trims_0pc_label() {
...
 async fn get_json_query_plan(query: &str) -> serde_json::Value {
-    let schema = Schema::parse_and_validate(SCHEMA, "").unwrap();
-    let document = Document::parse(query, "query.graphql").unwrap();
-    let executable = document.to_executable(&schema).unwrap();
-    let parsed_doc: ParsedDocument = Arc::from(ParsedDocumentInner {
-        ast: document,
-        executable: Arc::new(executable),
-        ..Default::default()
-    });
+    let schema = crate::spec::Schema::parse(include_str!("./testdata/supergraph.graphql")).unwrap();
```
goto-bus-stop, line 205:
```suggestion
    let schema = crate::spec::Schema::parse_test(include_str!("./testdata/supergraph.graphql"), &Default::default()).unwrap();
```

apollo-router/src/plugins/progressive_override/tests.rs
```diff
@@ -286,16 +273,10 @@ async fn query_with_labels(query: &str, labels_from_coprocessors: Vec<&str>) {
...
-    // it ourselves for testing purposes
-    let schema = Schema::parse_and_validate(SCHEMA, "").unwrap();
-    let document = Document::parse(query, "query.graphql").unwrap();
-    let executable = document.to_executable(&schema).unwrap();
-    let parsed_doc: ParsedDocument = Arc::from(ParsedDocumentInner {
-        ast: document,
-        executable: Arc::new(executable),
-        ..Default::default()
-    });
+    let schema = crate::spec::Schema::parse(include_str!("./testdata/supergraph.graphql")).unwrap();
```
goto-bus-stop, line 276:
```suggestion
    let schema = crate::spec::Schema::parse(include_str!("./testdata/supergraph.graphql"), &Default::default()).unwrap();
```

apollo-router/src/query_planner/testdata/defer_clause.graphql
```diff
@@ -30,6 +30,8 @@ directive @link(
   import: [link__Import]
 ) repeatable on SCHEMA
 
+directive @defer(label: String, if: Boolean! = true) on FRAGMENT_SPREAD | INLINE_FRAGMENT
+
```
goto-bus-stop, line 34:
```suggestion

```
defer should not be declared in the supergraph but added by API schema, if there's a test failure the test should use `Schema::parse_test`

SimonSapin (approved):
Looks good overall, but I still think we should use `.to_json()` over `.to_string()` for parse errors: https://github.com/apollographql/router/pull/4551/files#r1525124171

The latter formats errors for a CLI-like output with a monospace font

apollo-router/src/services/supergraph/tests.rs
```diff
@@ -1057,7 +1057,7 @@ async fn subscription_callback_schema_reload() {
 
     let new_schema = format!("{SCHEMA}  ");
     // reload schema
-    let schema = Schema::parse(&new_schema, &configuration).unwrap();
+    let schema = Schema::parse(&new_schema).unwrap();
```
goto-bus-stop, line 1060:
```suggestion
    let schema = Schema::parse_test(&new_schema, &configuration).unwrap();
```
I think it doesn't technically make a difference for this test, but lets use `Schema::parse_test` for consistency? especially since `schema.api_schema()` can silently do the wrong thing at the moment and we should catch those cases in tests

SimonSapin:
Marked as approved to mean I don’t need to approve again after this change, but I do think the parse error thing should be changed.

goto-bus-stop (changes requested):
Awesome 🤩  everything LGTM except a few places in tests where we should use `Schema::parse_test`

apollo-router/src/query_planner/bridge_query_planner.rs
goto-bus-stop: We should probably keep this for now so we don't pass invalid documents to the query planner. When #4938 is fixed we can consider only validating if debug assertions are on and use `assume_valid` in prod?

yanns:
We are using a plugin defining a  `supergraph_service` . Before this change, this method was called, we could do some request validation in it. After this change, it seems that this method is not called when the GraphQL query is invalid.
Can you confirm that this PR could change this behavior? Is it intended?

abernix:
@yanns What was the request validation that you were doing at the supergraph service?  Something GraphQL-specific?

yanns:
> @yanns What was the request validation that you were doing at the supergraph service? Something GraphQL-specific?

There are 2 use-cases:
- (1) as we have a public GraphQL API, we need to translate some error code. So we need a plugin that translates the error codes emitted by the router to "our" error codes. If the router rejects invalid GraphQL queries without calling plugins, we cannot modify the response.
- (2) we are using feature toggles. One GraphQL query does not use exactly the same GraphQL schema as another GraphQL query. So we need to validate requests depending on the GraphQL schema that is valid for this particular request. For this, I don't think that this PR is an issue. We configure the router with a GraphQL schema that has all possible fields. After the router has accepted the query, we validate it again with another GraphQL that is a subset of the one used by the router.

So I guess our main issue is that our plugin cannot adapt the response in case of invalid GraphQL queries.

sushant3524:
Helllo Team, 
Really appreciate the work you guys are doing.
I have a use-case where I don't want any basic validation on router level. For that I did some changes in apollo-compiler and router to ignore validation error on rust , and for the most part it is working.
But for some queries, fragment types in particular are still giving validation error.
After exploring I have found out that these validation checks are coming from js (legacy).
I even tried using `experimental_graphql_validation_mode: new` in my yaml, but still it persists and validation errors are coming from Legacy.

To reproduce
Query is:
```
query getCall1($cursor: String, $size: Int) {
  call1: getCall1(cursor: $cursor, size: $size) {
    results {
      id
      payload {
        ...Payload1
        __typename
      }
      sender
      __typename
    }
    __typename
  }
}

fragment Payload1 on Payload1 {
  field1
  type
  ...Type1
  ...Type2
  __typename
}

fragment Type1 on Type1 {
  field2
  __typename
}

fragment Type2 on Type2 {
    field3
  __typename
}

```
and say `superschema doesn't have Type2 info.`
The error I am getting is 
```
{
	"errors": [
		{
			"message": "Cannot find fragment name \"Type2\" (provided fragments are: [Type1, Type3])",
			"extensions": {
				"code": "GRAPHQL_VALIDATION_FAILED"
			}
		}
	]
}
```
Any help is appreciated, thank you.

goto-bus-stop:
@sushant3524 Accepting invalid queries in the router may cause all kind of unsupported glitches in the router and in the query planner. The router does not do validation in JS but I think the query planner may validate its own *outputs* at the end of query planning. Subgraphs also usually validate queries and if they return an error the router will propagate it. Everything the router does relies on the query being valid so trying to pass through something invalid is asking for trouble.

sushant3524:
Thanks @goto-bus-stop  for the quick revert. Have reviewed much of your code in the router repository and it is a pleasure to finally engage with you.
I understand and accept the risks and will very well work on it. I just want that router doesn't validate through JS and as this PR removes it, I am not sure how and why JS is validating.

Is it because of [#4964](https://github.com/apollographql/router/pull/4964) ?
image
Or is it because router uses JS for some other usecase like maybe query_planner and there in JS if the query looks fishy they validate it again?

sushant3524:
after searching I have found out that this is coming in bridge_query_planner from `PlannerMode`.
Are we working on removing dependency from JS for it?

PR #4568, opened 2024-01-30
HTTP client service

Geal:
This moves HTTP request handling to its own service, separated from the subgraph service. This will allow us to better organize code working strictly at the HTTP level, like compression and request signing.
This service is not publicly exposed for now, it is internal only, so we can iterate on its API and improve it before making it public.
There will be follow up PRs to move some plugins to the HTTP service level.
This service will potentially be reused for other HTTP client needs in the router, like Uplink or Orbiter, which will give us a unique way to configure TLS, retries, etc.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/services/http/service.rs
```diff
@@ -0,0 +1,390 @@
...
+
+        let http_req_span = tracing::info_span!("http_request",
+            "otel.kind" = "CLIENT",
+            "net.peer.name" = %host,
+            "net.peer.port" = %port,
+            "http.route" = %path,
+            "http.url" = %schema_uri,
+            "net.transport" = "ip_tcp",
+            //"apollo.subgraph.name" = %service_name,
+            //"graphql.operation.name" = %operation_name,
```
garypen, line 226: Should this be deleted or should we be using the `service_name` from line 236? Does it matter that the operation name is lost from the span?

garypen (approved):
Looks good. Just the one question about the span attributes.

apollo-router/src/services/http/service.rs
```diff
@@ -0,0 +1,390 @@
...
+}
+
+pin_project! {
+    pub(crate) struct BodyStream<B: hyper::body::HttpBody> {
+        #[pin]
+        inner: DecompressionBody<B>
+    }
+}
+
+//struct BodyStream<B: hyper::body::HttpBody>(DecompressionBody<B>);
```
o0Ignition0o, line 373:
```suggestion
```

o0Ignition0o (approved):

apollo-router/src/services/http/service.rs
Geal: I think we should remove it, because the goal is to reuse that service in other parts of the router, like uplink

PR #4570, opened 2024-01-30
Improve schema enforcement

trevor-scheer:
The initial implementation of directive argument schema enforcement is lacking:
* It doesn't handle renames of the `join` spec via `as`, meaning the restriction could be circumvented via a `join` rename.
* It only looked for the directive argument on the definition rather than actual usages. If the arg is defined but unused, we shouldn't enforce anything. It might exist in the definition simply as a result of using the latest versions of subgraph libs/composition.

Additionally, this PR introduces the requirement of specifying a `semver::VersionReq` for enforcement.

### DIFF
--- a/apollo-router/src/uplink/license_enforcement.rs
+++ b/apollo-router/src/uplink/license_enforcement.rs
@@ -116,7 +116,9 @@ impl ParsedLinkSpec {
                 let spec_name = captures.get(2).unwrap().as_str().to_string();
                 let version_string = captures.get(3).unwrap().as_str().to_string();
 
-                let Ok(parsed_version) = semver::Version::parse(&version_string) else {
+                let Ok(parsed_version) =
+                    semver::Version::parse(format!("{}.0", &version_string).as_str())
+                else {
                     return None;
                 };
                 let imported_as = link_directive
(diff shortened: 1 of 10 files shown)


apollo-router/src/uplink/license_enforcement.rs
```diff
@@ -82,7 +85,69 @@ where
...
+impl ParsedLinkSpec {
+    fn from_link_directive(link_directive: &Node<Directive>) -> Option<ParsedLinkSpec> {
+        link_directive
+            .argument_by_name(LINK_URL_ARGUMENT)
+            .and_then(|value| {
+                let Some(url) = value.as_str() else {
+                    return None;
+                };
+                // capture url + spec name, spec name, and version from a link spec url
+                let re = Regex::new(r"^(.*/([^/]+))/v(\d+\.\d+)$").unwrap();
```
Geal, line 109: why not use a URL parser?

apollo-router/src/uplink/license_enforcement.rs
```diff
@@ -82,7 +85,69 @@ where
...
+                let Some(captures) = re.captures(url) else {
+                    return None;
+                };
+
+                let full_url = captures.get(0).unwrap().as_str().to_string();
```
Geal, line 114: with this unwrap the router would panic on a schema with an invalid spec URL. That function should probably return a `Result<Option<ParsedLinkSpec>, SomeError>`, to account for both cases of no spec found, or parsing error

Geal (changes requested):

Geal (approved):

o0Ignition0o (approved):
LGTM now your semver updates have shipped! :shipit:

PR #4571, opened 2024-01-31
Add analytics for progressive override usage

trevor-scheer:
Holding off on landing this pending some discussion.


o0Ignition0o (approved):

Geal (approved):

apollo-router/src/plugins/progressive_override/mod.rs
```diff
@@ -209,6 +209,23 @@ impl Plugin for ProgressiveOverridePlugin {
                         })
                         .clone();
 
+                    if !relevant_labels.is_empty() {
+                        u64_counter!(
+                            "apollo.router.usage.override.query",
```
BrynCooke, line 217: Suggest that this prefix is changed to either:
`apollo.router.config.`
or
`apollo.router.supergraph.`

Adding a new prefix `usage` doesn't give any extra info over the similarly generic `config`, but you could argue that me make `config` yaml related things and `supergraph` supergraph related.

BrynCooke (changes requested):

apollo-router/src/plugins/progressive_override/mod.rs
Geal: fixed in [0738dee](https://github.com/apollographql/router/pull/4571/commits/0738deef68141834a781b9c1da76c243b5ce548e)

PR #4572, opened 2024-01-31
feat(telemetry): add configurable histogram buckets per metric

### LINKED ISSUE #4543: configurable histogram buckets per metric
there is an ask for configurable histogram buckets per metric (currently we only have a global configuration). This would let users get more precision for some metrics while keeping the amount of data low for other metrics

bnjjj:
Add support for opentelemetry views (a mechanism to override instrument settings) for metrics. It will let you override default histogram buckets for example.

Example of configuration:

```yaml
telemetry:
  apollo:
    client_name_header: name_header
    client_version_header: version_header
  exporters:
    metrics:
      common:
        service_name: apollo-router
        views:
          - instrument_name: apollo_router_http_request_duration_seconds
            aggregation:
              histogram:
                buckets:
                - 1
                - 2
                - 3
                - 4
                - 5
      prometheus:
        enabled: true
```

Fixes #4543

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [x] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/plugins/telemetry/metrics/otlp.rs
```diff
@@ -63,6 +67,23 @@ impl MetricsConfigurator for super::super::otlp::Config {
...
                             .build(),
                     );
+                if let Some(custom_buckets) = metrics_config.buckets.get_custom() {
+                    for (instrument_name, buckets) in custom_buckets {
+                        let name: &'static str =
+                            Box::leak(instrument_name.clone().into_boxed_str());
+                        // Add this here to be able to clean it once we reload the MeterProvider
+                        builder.instrument_names_to_clean.push(name);
+                        let view = new_view(
+                            Instrument::new().name(name),
```
Geal, line 77: I mentioned it in slack, but adding context here too: the `name()` method takes a `Cow<'static, str>` as argument, but we don't have to jump through hoops to make a static str, since `Cow` implements `From<String>` (which will make a `Cow::Owned` directly

apollo-router/src/plugins/telemetry/config.rs
```diff
@@ -119,13 +129,73 @@ impl Default for MetricsCommon {
...
             ],
         }
     }
 }
 
+#[derive(Clone, Debug, Deserialize, JsonSchema)]
+#[serde(deny_unknown_fields)]
+pub(crate) struct MetricView {
+    /// The instrument name you're targeting
+    pub(crate) instrument_name: String,
```
BrynCooke, line 144:
```suggestion
    pub(crate) name: String,
```

This is called `name` in the spec: https://opentelemetry.io/docs/specs/otel/metrics/sdk/#instrument-selection-criteria

In addition it should be defaulted to "*".

docs/source/configuration/telemetry/exporters/metrics/overview.mdx
```diff
@@ -183,6 +184,33 @@ OpenTelemetry includes many [standard attributes](https://opentelemetry.io/docs/
...
+You can override default attributes and default buckets for specific metrics thanks to this configuration.
+
+```yaml title="router.yaml"
+telemetry:
+  exporters:
+    metrics:
+      common:
+        service_name: apollo-router
+        views:
+          - instrument_name: apollo_router_http_request_duration_seconds # Metric name you want to edit
```
BrynCooke, line 198:
```suggestion
          - name: apollo_router_http_request_duration_seconds # Instrument name you want to edit. default: *
```

apollo-router/src/plugins/telemetry/metrics/prometheus.rs
```diff
@@ -131,10 +132,14 @@ impl MetricsConfigurator for Config {
             .with_registry(registry.clone())
             .build()?;
 
-        let meter_provider = MeterProvider::builder()
+        let mut meter_provider_builder = MeterProvider::builder()
```
BrynCooke, line 135: Not on this line, but further up in the file we check to see if the prometheus registry can be reused. If any of the views change then I think that we need to say it can't be reused.

BrynCooke (changes requested):

Geal (approved):

apollo-router/src/plugins/telemetry/config.rs
bnjjj: Name updated, I think we should keep the `name` mandatory, if not then every fields would be optional and that could be weird. I'll document it to let people know they can use wildcard in names

Geal (approved):

BrynCooke (approved):

abernix:
As I posted on https://github.com/apollographql/router/issues/4543#issuecomment-2150541597:

> Note while this feature first appeared in Router https://github.com/apollographql/router/releases/tag/v1.40.0, we don't recommend using it until https://github.com/apollographql/router/releases/tag/v1.46.0.  We've updated the CHANGELOG to reflect this note.
>
> The version recommendation is based on the first stable version (note, this excludes versions which are not recommended in the [changelog](https://github.com/apollographql/router/blob/dev/CHANGELOG.md)) after https://github.com/apollographql/router/pull/4771 landed.

PR #4578, opened 2024-02-01
fix: Progressive override - use `.operations` instead of `.schema` for operations-related metrics

trevor-scheer:
These were accidentally named under the impression that they were tracking schema-related config rather than operation-based behavior.


BrynCooke (approved):

Geal (approved):

Geal:
I'm merging this directly to the 1.39.0 branch so the release can ship with this change

bnjjj (approved):

PR #4579, opened 2024-02-01
Add selectors to extract body elements using JSONPath

### LINKED ISSUE #4443: Selectors that extract body elements use JSONQuery rather than JSONPath
We've documented selectors that extract from json as using JSONPath but it looks like we use JSONQuery.
We should use one or the other. My initial inclination is to switch to JSONPath as we have documented because it is more powerful, and the library is maintained.

Related, the error message for non parseable paths is really bad as it implies that custom attributes are not allowed.

bnjjj:
Deprecated `subgraph_response_body` in favor of `subgraph_response_data` and `subgraph_response_errors` which is a [selector](https://www.apollographql.com/docs/router/configuration/telemetry/instrumentation/selectors/) and use a JSON Path to fetch data/errors from the subgraph response. 

Example:

```yaml
telemetry:
  instrumentation:
    spans:
      subgraph:
        attributes:
          "my_attribute":
            subgraph_response_data: "$.productName"
            subgraph_response_errors: "$.[0].message"
```

Fixes #4443

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/plugin/serde.rs
+++ b/apollo-router/src/plugin/serde.rs
@@ -6,6 +6,7 @@ use std::str::FromStr;
 use access_json::JSONQuery;
 use http::header::HeaderName;
 use http::HeaderValue;
+use jsonpath_rust::JsonPathInst;
 use regex::Regex;
 use serde::de;
 use serde::de::Error;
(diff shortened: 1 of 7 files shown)


apollo-router/src/plugin/serde.rs
```diff
@@ -209,3 +210,27 @@ where
...
+    deserializer.deserialize_str(JSONPathVisitor)
+}
+
+struct JSONPathVisitor;
+
+impl<'de> serde::de::Visitor<'de> for JSONPathVisitor {
+    type Value = JsonPathInst;
+
+    fn expecting(&self, formatter: &mut Formatter) -> std::fmt::Result {
+        write!(formatter, "a string that begins with '$'")
```
Geal, line 227: maybe a more explicit error message, like saying we expect a JSON path here?

Geal (approved):

apollo-router/src/plugin/serde.rs
bnjjj: fixed. Thank you

BrynCooke (approved):

Issue #842, opened 2022-04-13
Upload file mutation failed


suryatresna:
**Describe the bug**
I try to upload files via graphql
```
curl -v localhost:4000/ \
  -F operations='{ "query": "mutation ($file: Upload!) { singleUpload(file: $file) { location } }", "variables": { "file": null } }' \
  -F map='{ "0": ["variables.file"] }' \
  -F 0=@/File/location/to/add.jpg
```

and get this response
```
*   Trying ::1:4000...
* connect to ::1 port 4000 failed: Connection refused
*   Trying 127.0.0.1:4000...
* Connected to localhost (127.0.0.1) port 4000 (#0)
> POST / HTTP/1.1
> Host: localhost:4000
> User-Agent: curl/7.77.0
> Accept: */*
> Content-Length: 1269305
> Content-Type: multipart/form-data; boundary=------------------------f6d06ed6d2fc8b81
> Expect: 100-continue
>
* Mark bundle as not supporting multiuse
< HTTP/1.1 415 Unsupported Media Type
< content-type: text/plain; charset=utf-8
< content-length: 43
< date: Wed, 13 Apr 2022 06:15:28 GMT
* HTTP error before end of send, stop sending
<
* Closing connection 0
The request's content-type is not supported
```

is there any way to add configuration/plugin in apollo router to support feature upload?

**Expected behavior**

The request Upload (`Content-type=multipart/form-data`) should be accepted by the Router and forwarding the request to the subgraph.

reference: [https://github.com/jaydenseric/graphql-multipart-request-spec](https://github.com/jaydenseric/graphql-multipart-request-spec)

Geal:
this is not possible in the router right now, and would require significant changes. Do you currently have a gateway that supports it?
I'm asking about the gateway because the expected behaviour might not be obvious. Would the router buffer the entire file before sending to the subgraph? Or the subgraph supports multipart upload too and data is streamed?

suryatresna:
> Do you currently have a gateway that supports it?

I used https://github.com/nautilus/gateway for gateway before, and this library supports multipart upload.

> the subgraph supports multipart upload too and data is streamed?

and the subgraph for upload, I used https://gqlgen.com/reference/file-upload/

> Would the router buffer the entire file before sending it to the subgraph?

If I'm not mistaken, they kept the entire files on the gateway and then sent them to the subgraph.

[...]

PR #4584, opened 2024-02-02
File uploads via "restricted" GraphQL Multipart Spec

IvanGoncharov:
Added a plugin that implements "restricted version" of [jaydenseric/graphql-multipart-request-spec](https://github.com/jaydenseric/graphql-multipart-request-spec)

Fixes https://github.com/apollographql/router/issues/842

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/services/subgraph_service.rs
```diff
@@ -975,8 +976,7 @@ async fn do_fetch(
     FetchError,
 > {
     let _active_request_guard = context.enter_active_request();
-    let response = client
-        .call(request)
+    let response = file_uploads::wrap_http_client_call(client, request)
```
Geal, line 979: FYI this conflicts with https://github.com/apollographql/router/pull/4568 which was just merged, but that should not be a huge change, it keeps working on a hyper Body

nicholascioli:
I've added some integration tests. I just need to make one for testing incompatible ordering, but the ones written should serve as a good base for testing. Note that some of the tests currently fail due to unimplemented features and the fact that we don't have error handling in the plugin yet.

apollo-router/src/plugins/file_uploads/mod.rs
```diff
@@ -245,9 +269,188 @@ type MapPerVariable = HashMap<String, HashMap<String, Vec<Vec<String>>>>;
...
+                formatted_query_plan: req.query_plan.formatted_query_plan.clone(),
+                query: req.query_plan.query.clone(),
+            }),
+            ..req
+        };
+    }
+    req
+}
+
+fn rearrange_plan_node<'a>(
```
BrynCooke, line 306: This logic may need to be moved to allow the resulting query plan to be cached.

apollo-router/src/plugins/file_uploads/mod.rs
```diff
@@ -0,0 +1,726 @@
...
+}
+
+const APOLLO_REQUIRE_PREFLIGHT: http::HeaderName =
+    HeaderName::from_static("apollo-require-preflight");
+const TRUE: http::HeaderValue = HeaderValue::from_static("true");
+
+pub(crate) async fn http_request_wrapper(
+    mut req: http::Request<hyper::Body>,
+) -> http::Request<hyper::Body> {
+    let supergraph_result = req.extensions_mut().remove();
```
Geal, line 559: what happens if a request is retried? 2 examples:
- the request fails, but request retry was set up in traffic shaping
- using APQ to the subgraph, a first call is made with the hash, the subgraph rejects the request, the router retries with the query to register it

apollo-router/src/plugins/file_uploads/mod.rs
```diff
@@ -82,6 +89,17 @@ impl Plugin for FileUploadsPlugin {
...
     }
 
+    fn execution_service(&self, service: execution::BoxService) -> execution::BoxService {
+        ServiceBuilder::new()
+            .oneshot_checkpoint_async(|req: execution::Request| {
+                rearange_execution_plan(req)
+                    .map(|req| Ok(ControlFlow::Continue(req)))
+                    .boxed()
+            })
+            .service(service)
```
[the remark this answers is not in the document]
BrynCooke: We discussed and we're not going to do this for preview as there are other things that need to be taken into account such as the contents of the file map on the second part of the stream. We will tackle for GA though.

apollo-router/src/plugins/file_uploads/mod.rs
BrynCooke: We discussed and we're not going to do this for preview as there are other things that need to be taken into account such as the contents of the file map on the second part of the stream. We will tackle for GA though.

apollo-router/src/plugins/file_uploads/mod.rs
```diff
@@ -0,0 +1,644 @@
...
+            .extensions()
+            .lock()
+            .insert(RouterLayerResult { multipart });
+
+        let content_type = operations_stream
+            .headers()
+            .get(CONTENT_TYPE)
+            .cloned()
+            .unwrap_or_else(|| HeaderValue::from_static("application/json"));
+        request_parts.headers.insert(CONTENT_TYPE, content_type);
```
bnjjj, line 201: Just to make sure, do we want to `insert` or `append` ?

apollo-router/src/plugins/file_uploads/mod.rs
```diff
@@ -0,0 +1,644 @@
...
+            .insert(SupergraphLayerResult {
+                multipart,
+                map_per_variable,
+                files_order,
+            });
+    }
+    Ok(req)
+}
+
+fn try_path<'a>(
```
bnjjj, line 282: It would be nice to add a comment to document what this function does

apollo-router/src/plugins/file_uploads/mod.rs
```diff
@@ -0,0 +1,644 @@
...
+        let SubgraphHttpRequestExtensions {
+            multipart,
+            map_field,
+        } = supergraph_result;
+
+        let (mut request_parts, operations) = req.into_parts();
+        let form = MultipartFormData::new(operations, map_field, multipart);
+        request_parts
+            .headers
+            .insert(CONTENT_TYPE, form.content_type());
```
bnjjj, line 407: Same question here for `insert` vs `append`

apollo-router/src/plugins/file_uploads/mod.rs
```diff
@@ -0,0 +1,644 @@
...
+    }
+}
+
+impl MultipartRequest {
+    fn new(request_body: hyper::Body, boundary: String, limits: MultipartRequestLimits) -> Self {
+        let multer = Multipart::with_constraints(
+            request_body,
+            boundary,
+            Constraints::new().size_limit(
+                SizeLimit::new().for_field("map", 10 * 1024), // hardcoded to 10kb
```
bnjjj, line 462: NIT: If it's a hardcoded limit, maybe it would make sense to put it in a constant at the beginning of the file with a comment

apollo-router/src/plugins/file_uploads/mod.rs
```diff
@@ -0,0 +1,644 @@
...
+    async fn operations_field(&mut self) -> UploadResult<multer::Field<'static>> {
+        Ok(self
+            .state
+            .lock()
+            .await
+            .multer
+            .next_field()
+            .await?
+            .filter(|field| field.name() == Some("operations"))
+            .ok_or_else(|| FileUploadError::MissingOperationsField)?)
```
bnjjj, line 486:
```suggestion
        self
            .state
            .lock()
            .await
            .multer
            .next_field()
            .await?
            .filter(|field| field.name() == Some("operations"))
            .ok_or_else(|| FileUploadError::MissingOperationsField)
```
if it works it's easier to read IMHO

apollo-router/src/plugins/file_uploads/multipart_form_data.rs
```diff
@@ -0,0 +1,100 @@
+use bytes::Bytes;
```
bnjjj, line 1: For consistency, in the future it might be interesting to refactor that part of the code to mirror what we have in `protocols` directory. And directly move the code there

apollo-router/src/plugins/file_uploads/rearange_query_plan.rs
```diff
@@ -0,0 +1,244 @@
...
+
+use super::error::FileUploadError;
+use super::MapPerVariable;
+use super::UploadResult;
+use crate::query_planner::DeferredNode;
+use crate::query_planner::FlattenNode;
+use crate::query_planner::PlanNode;
+use crate::services::execution::QueryPlan;
+
+pub(super) fn rearange_query_plan(
```
bnjjj, line 17: I would be happy to have a little bit of docs here. To know what's the goal of this function

apollo-router/src/plugins/file_uploads/rearange_query_plan.rs
```diff
@@ -0,0 +1,244 @@
...
+    query_plan: &QueryPlan,
+    files_order: &IndexSet<String>,
+    map_per_variable: &MapPerVariable,
+) -> UploadResult<QueryPlan> {
+    let root = &query_plan.root;
+    let mut variable_ranges = HashMap::new();
```
bnjjj, line 23:
```suggestion
    let mut variable_ranges = HashMap::with_capacity(map_per_variable.len());
```

apollo-router/src/plugins/file_uploads/rearange_query_plan.rs
```diff
@@ -0,0 +1,244 @@
...
+    Ok(QueryPlan {
+        root,
+        usage_reporting: query_plan.usage_reporting.clone(),
+        formatted_query_plan: query_plan.formatted_query_plan.clone(),
+        query: query_plan.query.clone(),
+    })
+}
+
+// Recursive, and recursion is safe here since query plan is executed recursively.
+fn rearrange_plan_node<'a>(
```
bnjjj, line 45: This function needs more documentation I think to explain what's going on inside each match arms

apollo-router/src/services/subgraph_service.rs
BrynCooke: This got resolved

apollo-router/src/plugins/file_uploads/mod.rs
BrynCooke: Clippy also got this: [c0a03c6](https://github.com/apollographql/router/pull/4584/commits/c0a03c6a3e392bc6274685f6e0ea3e05a5be2db4)

apollo-router/src/plugins/file_uploads/mod.rs
BrynCooke: 3a141e6

apollo-router/src/plugins/file_uploads/multipart_form_data.rs
```diff
@@ -0,0 +1,89 @@
...
+    boundary: String,
+    operations: hyper::Body,
+    map: MapField,
+    multipart: MultipartRequest,
+}
+
+impl MultipartFormData {
+    pub(super) fn new(operations: hyper::Body, map: MapField, multipart: MultipartRequest) -> Self {
+        let boundary = format!(
+            "------------------------{:016x}",
```
Geal, line 28: if a custom boundary is used, it should be called out as part of the `Content-Type` header: https://www.w3.org/Protocols/rfc1341/7_2_Multipart.html#z0

```
Content-Type: multipart/mixed; boundary=gc0p4Jq0M2Yt08jU534c0p
```

Geal: also, why the `------------------------` prefix? It is not necessary

apollo-router/src/plugins/file_uploads/multipart_form_data.rs
```diff
@@ -0,0 +1,89 @@
...
+        let mime = MediaType::from_parts(MULTIPART, FORM_DATA, None, &params);
+        mime.to_string()
+            .try_into()
+            .expect("mime should be valid header value")
+    }
+
+    pub(super) async fn into_stream(mut self) -> impl Stream<Item = UploadResult<Bytes>> {
+        let map_bytes = serde_json::to_vec(&self.map).expect("map should be serializable to JSON");
+        let field_prefix = |name: &str| {
+            tokio_stream::once(Ok(Bytes::from(format!(
```
Geal, line 52: generally we use `once` from the `futures` crate instead of reaching for a transitive dependency https://docs.rs/futures/latest/futures/stream/fn.once.html

apollo-router/src/plugins/file_uploads/multipart_form_data.rs
```diff
@@ -0,0 +1,89 @@
...
+            let mut prefix = BytesMut::new();
+            prefix.extend_from_slice(b"\r\n--");
+            prefix.extend_from_slice(boundary.as_bytes());
+            prefix.extend_from_slice(b"\r\n");
+            for (k, v) in headers.iter() {
+                prefix.extend_from_slice(k.as_str().as_bytes());
+                prefix.extend_from_slice(b": ");
+                prefix.extend_from_slice(v.as_bytes());
+                prefix.extend_from_slice(b"\r\n");
+            }
```
Geal, line 78: this way to write it will result in each `chain` call sending a separate chunk, even the small ones. We should try to send larger chunks at once, to be mindful of the subgraph side, that will need to reparse the accumulated chunks until they get to the boundary. If the entire file is small enough to send as one chunk with the boundaries, great, if it is larger it can be cut up in pieces

apollo-router/src/plugins/file_uploads/mod.rs
```diff
@@ -0,0 +1,668 @@
...
+    let supergraph_result = req
+        .context
+        .extensions()
+        .lock()
+        .get::<SupergraphLayerResult>()
+        .cloned();
+    if let Some(supergraph_result) = supergraph_result {
+        let SupergraphLayerResult { map, .. } = supergraph_result;
+
+        let query_plan = Arc::new(rearange_query_plan(&req.query_plan, &map)?);
```
Geal, line 290:
```suggestion
        let query_plan = Arc::new(rearrange_query_plan(&req.query_plan, &map)?);
```

apollo-router/src/plugins/file_uploads/mod.rs
BrynCooke: The content type needs to replace the existing `multipart` content type, so I think it's insert. @IvanGoncharov Do you agree?

IvanGoncharov: Yes we need to replace it, otherwise we will see error in "content_negotition" layer since router don't know what to do with Multipart

apollo-router/src/plugins/file_uploads/mod.rs
IvanGoncharov: I changed code to clone here in 1a852be922514b8f54e4e2d0227c785e5f1c4ab9
But I think we have a deeper problem here; if the subgraph rejects the request after some of the files were transferred, we can't replicate this file in retry.
I think we need to disable APQ for requests using file uploads.

apollo-router/src/plugins/file_uploads/mod.rs
IvanGoncharov: We need to rewrite it from `application/json` to `multipart/form-data`.
I added comments in both places explaining this is intentional

apollo-router/src/plugins/file_uploads/rearange_query_plan.rs
IvanGoncharov: Makes sense algorithm is pretty complex 👍 
I added docs.

apollo-router/Cargo.toml
```diff
@@ -123,7 +123,7 @@ maplit = "1.0.2"
 mediatype = "0.19.18"
 mockall = "0.11.4"
 mime = "0.3.17"
-multer = "3.0.0"
+multer = "2.1.0"
```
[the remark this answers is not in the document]
BrynCooke: The new version of multer requires http 1.0. This is not what the rest of our codebase is using, and needs to be tackled separately.

apollo-router/src/plugins/file_uploads/rearrange_query_plan.rs
```diff
@@ -0,0 +1,243 @@
...
+
+use super::error::FileUploadError;
+use super::ParsedMap;
+use super::UploadResult;
+use crate::query_planner::DeferredNode;
+use crate::query_planner::FlattenNode;
+use crate::query_planner::PlanNode;
+use crate::services::execution::QueryPlan;
+
+pub(super) fn rearange_query_plan(
```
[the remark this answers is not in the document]
IvanGoncharov: fixed

apollo-router/src/plugins/file_uploads/multipart_form_data.rs
IvanGoncharov: > if a custom boundary is used, it should be called out as part of the `Content-Type` header: https://www.w3.org/Protocols/rfc1341/7_2_Multipart.html#z0
> 
> ```
> Content-Type: multipart/mixed; boundary=gc0p4Jq0M2Yt08jU534c0p
> ```

Correct, I provide it here: https://github.com/apollographql/router/pull/4584/files#diff-250dfae3c4048ea64ae9528e5286bfbbc8af08b7bec7d8bdb279d97385995d1eR43

apollo-router/src/plugins/file_uploads/multipart_form_data.rs
o0Ignition0o: fixed in [367ea94](https://github.com/apollographql/router/pull/4584/commits/367ea94546ff3f4db9c41d7e0f81e9884beddc10)

apollo-router/src/plugins/file_uploads/mod.rs
o0Ignition0o: ++, I know we planned on documenting it and calling it out, not sure where we are atm but if anything it will need to be a followup.

apollo-router/src/plugins/file_uploads/mod.rs
o0Ignition0o: https://github.com/apollographql/router/commit/367ea94546ff3f4db9c41d7e0f81e9884beddc10#diff-7067e2dd8c31cd5cd84b593d83c0002ce9a256d8d251abdf738bc83cc65e1bfeR250


we now have:

```rust
// Replaces value at path with the provided one.
// Returns the provided path if the path is not valid for the given object
fn replace_value_at_path<'a>(
    variables: &'a mut json_ext::Object,
    path: &'a [String],
    value: serde_json_bytes::Value,
) -> std::result::Result<(), &'a [String]> {
```

and 

```rust
// Removes value at path.
fn remove_value_at_path<'a>(variables: &'a mut json_ext::Object, path: &'a [String]) {
```

which both use Ivan's new `get_value_at_path` function

o0Ignition0o (approved):

BrynCooke (approved):
Approved with caveat that unit tests fast follow.

Issue #4473, opened 2024-01-12
Query Plan Caching Improvements 


lleadbet:
**Describe the solution you'd like**

The current functionality addresses most needs, but there are a couple of improvements that could be made to help make it more useful. 

* Include the federation version as part of the cache key
* Refresh TTLs on recently accessed query plans (optionally)
* Make the TTL a sane default vs. indefinite (in combination with the above)

The first allows for multiple different versions to use the same query plan cache, as different federation versions would generate different query plans in theory. 

The second would ensure that stale entries are evicted, but hot query plan paths are retained as long as they are being used in the current schema. For distributed caches, this can minimize the amount of data being used in memory without manual pruning or having to take a performance hit every time the TTL expires. 

**Describe alternatives you've considered**

n/a 

**Additional context**

n/a

Geal:
repopening ecause we're still missing https://github.com/apollographql/router/pull/4588

PR #4588, opened 2024-02-02
set a default TTL for query plans

Geal:
Fix https://github.com/apollographql/router/issues/4473

This sets a default TTL for query plan caches, because right now the default is sto store query plans indefinitely, which does not make sense because they change with schema updates.
The PR is a bit verbose because to add that new default, we need to create a redundant structure for redis configuration (I tried const generics but serde won't support them)

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


BrynCooke:
I think there should be some sort of documentation update associated with this. @shorgi can we add a table for the options with defaults?

Geal:
there is some documentation now

Geal:
actually maybe it would make sense to have a default expiration for APQ too :thinking:

Geal:
@BrynCooke @bnjjj @o0Ignition0o could I get a review for this PR? This has waited for 3 weeks already

BrynCooke (approved):

PR #4604, opened 2024-02-06
refresh query plan expiration on GET

Geal:
Fix https://github.com/apollographql/router/issues/4473
We want the query plan cache to act like a LRU, so if a TTL is set in its Redis configuration, we will reset the plan cache key's current expiration when reding it

~unclear to me here: should we instead make it an option on the common redis configuration type? We may want to control that behaviour for APQ and introspection too~ This is now part of the common redis configuration and defaults to true. For entity caching, this option is disabled, as the cache will manage TTL directly

Tests now require REdis 7. Otherwise you will get the error "ERR unknown command `EXPIRETIME`"

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


lleadbet:
> unclear to me here: should we instead make it an option on the common redis configuration type? We may want to control that behaviour for APQ and introspection too

If possible. The "TTL reset" mechanism is helpful for this use-case imo.

garypen (commented):
Is there any way we can test this functionality by extending our existing redis integration tests?
also: config feature => changelog?

Geal:
Probably yes, we can check a key's expiration by querying redis in the test

bnjjj (approved):
Missing the changelog but LGTM

Geal:
there is now a test in f08b3ad90. Integration tests will require Redis 7 from now on, to have access to the `EXPIRETIME` command

PR #4607, opened 2024-02-06
docs: SupergraphRequest example

shorgi:
From [user request](https://apollograph.slack.com/archives/C0721M2F6/p1705952702706759), add (empty) `body.variables` to SupergraphRequest example


Meschreiber (approved):
Seems fine to me, though I wonder if there's more information we can/should give about `body.variables`? I see it's also in the Subgraph Request body, but not sure if it's already clear to what they're for?

Geal (approved):

PR #4615, opened 2024-02-08
Allow OSX x86_64 binaries in the install script

Geal:
Related:
* https://github.com/apollographql/router/pull/4484
* https://github.com/apollographql/router/pull/4515 
* https://github.com/apollographql/router/pull/4518
* https://github.com/apollographql/router/pull/4605

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -97,15 +97,6 @@ get_architecture() {
         fi
     fi
 
-    # If our OS is Darwin, we only have binaries for arm64 machines.
-    # Check and exit early
-    if [ "$_ostype" = Darwin ] && [ "$_cputype" != arm64 ]; then
-        say "No precompiled binaries available on macOS for CPU architecture: $_cputype"
-        say "The most recent macOS release which supports Intel is 1.37.0"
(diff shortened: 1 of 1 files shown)


Cargo.lock
```diff
@@ -258,7 +258,7 @@ dependencies = [
 
 [[package]]
 name = "apollo-router"
-version = "1.39.0"
+version = "1.39.1-rc.0"
```
abernix, line 261: it's targeting `dev` right now, so this is showing up. 😄

abernix (changes requested):

garypen (commented):
Looks like you've got some extra changes (packaging versions) in there as well.

garypen (approved):

abernix (approved):
This change will work!  It's worth flagging that it will cause the current versions 1.38.0 and 1.39.0 to display this message, but that's a fine trade off for now. 

**We should update [that issue](https://github.com/apollographql/router/issues/4483) to surface the new state of the world!**

PR #4619, opened 2024-02-08
Fix active_session_count when future is dropped

xuorig:
We've noticed the `active_session_count` was increasing over time and never coming back to 0. It seems like this happens because it's possible for the future to be dropped before we ever get to decrement this static. Instead, a guard-like struct can be used to decrement on `drop`.

I believe the behavior should be roughly the same (I know there was a comment on waiting for the response to be fully transmitted before decrementing), let me know if you have any thoughts.

### DIFF
--- a/apollo-router/src/axum_factory/axum_http_server_factory.rs
+++ b/apollo-router/src/axum_factory/axum_http_server_factory.rs
@@ -64,6 +64,23 @@ use crate::uplink::license_enforcement::LICENSE_EXPIRED_SHORT_MESSAGE;
 
 static ACTIVE_SESSION_COUNT: AtomicU64 = AtomicU64::new(0);
 
+struct SessionCountGuard;
+
+impl SessionCountGuard {
+    fn start() -> Self {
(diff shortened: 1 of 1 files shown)


Geal (approved):

xuorig:
Result:

Screenshot 2024-02-08 at 19 48 52

bnjjj (approved):

abernix:
👋  So, we released this in https://github.com/apollographql/router/releases/tag/v1.40.0 today, but i'm not convinced it's fixed — yet.

Full transparency, I opened https://github.com/apollographql/router/issues/4650 without realizing that _this_ PR (and the issue, which was also new to me) was actually called out in the changelog.

Anyhow, that's just me saying that I'm still seeing this on 1.40.0, **I think**.

abernix:
Following up on my previous comment, I'm wrong and **this is — indeed — fixed**, thanks to @xuorig's fix. (🙇 ).  This manifestation was a problem with deployments that I was driving.

PR #4625, opened 2024-02-09
use a non blocking stdout

### LINKED ISSUE #4612: investigate standard IO locking
when writing logs to stdout and stderr, we use the standard library's `StdOut`, which requires [locking](https://doc.rust-lang.org/std/io/struct.Stdout.html#method.lock) to make sure multiple threads don't try to write lines at the same time.

There's a performance issue in there, but also, there's the potential to deadlock the router. (I am not entirely clear on the details here) if for some reason one thread takes the stdout lock and does not release it, every other thread that would try to write a log line at that moment would be blocked too, and then the router would be unable to process any more requests.

It should be possible to take the lock once on startup, then keep it the entire time the format layer is using it. But then, what happens on reloads?

Geal:
If the router's output is piped into another process, and that process does not consume that output, it can entirely lock up the router.

In Rust, stdout access is protected by a lock, to prevent multiple threads from interleaving their writes. And inside that lock, there is a LineWriter wrapping a BufWriter to batch line writes.

When the process receiving the output from the router does not consume, then the BufWriter's writes to the stream start to block, which means the current thread is blocked while holding the stdout lock. And then any other thread that might want to log something will end up blocked too, waiting for that lock to be released.

In this PR we mark stdout as non blocking, which means that writing logs will fail silently when the buffer is full. This is done in `executable.rs` to avoid disrupting tests, as any use of println could panic now.
We will have to document that native plugins should not use println to avoid that

Fixes #4612 

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


Geal:
An example of what can happen if we use a `println!` directly in the router:
```stack backtrace:
   0: rust_begin_unwind
             at /rustc/79e9716c980570bfd1f666e3b16ac583f0168962/library/std/src/panicking.rs:597:5
   1: core::panicking::panic_fmt
             at /rustc/79e9716c980570bfd1f666e3b16ac583f0168962/library/core/src/panicking.rs:72:14
   2: std::io::stdio::print_to
             at /rustc/79e9716c980570bfd1f666e3b16ac583f0168962/library/std/src/io/stdio.rs:1019:9
   3: std::io::stdio::_print
             at /rustc/79e9716c980570bfd1f666e3b16ac583f0168962/library/std/src/io/stdio.rs:1095:5
   4: apollo_router::services::layers::query_analysis::QueryAnalysisLayer::supergraph_request::{{closure}}
             at ./apollo-router/src/services/layers/query_analysis.rs:69:9
   5: apollo_router::services::router::service::RouterService::process_supergraph_request::{{closure}}
             at ./apollo-router/src/services/router/service.rs:242:88
   6: <futures_util::future::maybe_done::MaybeDone<Fut> as core::future::future::Future>::poll
             at /home/geal/.cargo/registry/src/index.crates.io-6f17d22bba15001f/futures-util-0.3.30/src/future/maybe_done.rs:95:38
   7: <futures_util::future::join_all::JoinAll<F> as core::future::future::Future>::poll
             at /home/geal/.cargo/registry/src/index.crates.io-6f17d22bba15001f/futures-util-0.3.30/src/future/join_all.rs:143:24
   8: apollo_router::services::router::service::RouterService::call_inner::{{closure}}
             at ./apollo-router/src/services/router/service.rs:401:14
   9: <apollo_router::services::router::service::RouterService as tower_service::Service<apollo_router::services::router::Request>>::call::{{closure}}
             at ./apollo-router/src/services/router/service.rs:222:53
  10: <core::pin::Pin<P> as core::future::future::Future>::poll
             at /rustc/79e9716c980570bfd1f666e3b16ac583f0168962/library/core/src/future/future.rs:125:9
  11: <core::pin::Pin<P> as core::future::future::Future>::poll
             at /rustc/79e9716c980570bfd1f666e3b16ac583f0168962/library/core/src/future/future.rs:125:9
  12: <core::pin::Pin<P> as core::future::future::Future>::poll
             at /rustc/79e9716c980570bfd1f666e3b16ac583f0168962/library/core/src/future/future.rs:125:9
  13: <apollo_router::plugins::telemetry::Telemetry as apollo_router::plugin::Plugin>::router_service::{{closure}}::{{closure}}
             at ./apollo-router/src/plugins/telemetry/mod.rs:361:80
```

While a panic can be scary, I would argue that in that case it is better to have a loud failure that can be detected (the client gets a connection reset which any load balancer could easily interpret as an unhealty node), than a router that accepts connection and then hangs without responding. Because then we have to wait for the client timeout to trigger, we may not detect the failure as easily.

I think I could add a log to stderr when we see the issue of blocked stdout, maybe with a rate limit like other logs

o0Ignition0o (approved):

PR #4631, opened 2024-02-12
fix http client service plugins

Geal:
This refactors parts of the `RouterFactory ` to accommodate plugins at the HTTP service level.
A `SubgraphService` needs to contain the `HttpClientFactory`, which needs to hold a `Arc<Plugins>`. Unfortunately, when creating the subgraph service, the plugin list is not yet in that state because we need to first create all the subgraphs, making sure we do not encounter any error, then we go through the plugins list to find the telemetry plugin and activate it. Only then we can put plugins in an `Arc` and pass that around.
Since we know the `TrafficShaping` plugin already wraps the subgraph service directly (except for mocks that we
handle separately), we add functions to pass the plugins list across the various layers.
Since the last `MapRequest` layer is cumbersome to abstract away here, we move them to the HTTP client level, making `TrafficShaping` the first service to exploit that new level

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


Geal:
closed in favor of #4666

PR #4640, opened 2024-02-13
[multipart-subs] clarifications about `"payload": null`

martinbonnin:
Attempt at clarifying what the behaviour should be when a `"payload": null` message is received. This PR models them as fatal errors because it is what Apollo Kotlin is currently expecting but anything else would work too as long as there is consensus on what the expected behaviour is.

### DIFF
--- a/dev-docs/multipart-subscriptions-protocol.md
+++ b/dev-docs/multipart-subscriptions-protocol.md
@@ -43,13 +43,11 @@ Content-Type: application/json
 --graphql--
 ```
 
-## Messages
+## GraphQL Responses
 
-The protocol differentiates between transport-level concerns and the GraphQL response payloads themselves. One reason for this is that [the response format is part of the GraphQL spec](https://spec.graphql.org/draft/#sec-Response-Format), and additional fields might be confusing or could even break client typing.
+When a GraphQL response is received, it is returned in the `payload` field using the [format from the GraphQL spec](https://spec.graphql.org/draft/#sec-Response-Format).
 
-Except for [heartbeats](#heartbeats), every message will therefore include a `payload`, but the payload may be `null`.
-
-Some errors are part of an execution result. Results may include partial data, and these errors are not fatal (meaning the subscription stream should be kept open). They are therefore delivered within the payload:
+A GraphQL response may contains [errors](https://spec.graphql.org/October2021/#sec-Errors). These errors are not fatal (meaning the subscription stream should be kept open) and are delivered within the payload:
 
 ```json
 {
@@ -61,7 +59,9 @@ Some errors are part of an execution result. Results may include partial data, a
 }
 ```
 
-When the router encounters an error that is fatal and should lead to termination of the subscription, it will instead send a message with a top-level `errors` field, and then close the connection:
+## Fatal Errors
+
+When the router encounters an error that is fatal and should lead to termination of the subscription, `payload` is null:
 
 ```json
 {
@@ -70,5 +70,6 @@ When the router encounters an error that is fatal and should lead to termination
 }
 ```
 
-Both types of `errors` will follow the [GraphQL error format](http://spec.graphql.org/draft/#sec-Errors.Error-Result-Format) (but top-level `errors` will never have `locations` or `path`).
+The top-level `errors` may be present to give more details about the errors. It follows the [GraphQL error format](http://spec.graphql.org/draft/#sec-Errors.Error-Result-Format) but will never contain `locations` or `path` fields.
+After a fatal error, the router closes the connection.
 


abernix:
I'm going to mark this as a draft **just for now**, particularly since this is just an attempt at clarifying and we don't have an issue with any previous discussion on it.

dev-docs/multipart-subscriptions-protocol.md
```diff
@@ -43,13 +43,11 @@ Content-Type: application/json
...
-## Messages
+## GraphQL Responses
 
-The protocol differentiates between transport-level concerns and the GraphQL response payloads themselves. One reason for this is that [the response format is part of the GraphQL spec](https://spec.graphql.org/draft/#sec-Response-Format), and additional fields might be confusing or could even break client typing.
+When a GraphQL response is received, it is returned in the `payload` field using the [format from the GraphQL spec](https://spec.graphql.org/draft/#sec-Response-Format).
 
-Except for [heartbeats](#heartbeats), every message will therefore include a `payload`, but the payload may be `null`.
-
-Some errors are part of an execution result. Results may include partial data, and these errors are not fatal (meaning the subscription stream should be kept open). They are therefore delivered within the payload:
+A GraphQL response may contains [errors](https://spec.graphql.org/October2021/#sec-Errors). These errors are not fatal (meaning the subscription stream should be kept open) and are delivered within the payload:
```
alessbell, line 50: Only saw this on the fifth read through 🙃

```suggestion
A GraphQL response may contain [errors](https://spec.graphql.org/October2021/#sec-Errors). These errors are not fatal (meaning the subscription stream should be kept open) and are delivered within the payload:
```

calvincestari:
For more context, this is the behaviour in Apollo iOS:
* `"payload": null` is effectively ignored and that alone will not throw an error - [test](https://github.com/apollographql/apollo-ios-dev/blob/469866288e7e4fc194e9ad9696c5f6e8c35756b6/Tests/ApolloTests/Interceptors/MultipartResponseSubscriptionTests.swift#L206)
* If `"errors" : []` is present an error will be thrown - [test](https://github.com/apollographql/apollo-ios-dev/blob/469866288e7e4fc194e9ad9696c5f6e8c35756b6/Tests/ApolloTests/Interceptors/MultipartResponseSubscriptionTests.swift#L49)
* We don't have a test that combines them but I put one together quickly and the behaviour is the same with `"payload": null` having no effect.

@alessbell has got a draft PR up at https://github.com/apollographql/apollo-client/pull/11594 which matches that behavior.

Personally I think this spec change and ignoring `"payload": null` are the right changes to make. I don't see any need to take action on an empty payload. I believe the context is which it's sent is to end a subscription and we don't take any particular action for that elsewhere.

dev-docs/multipart-subscriptions-protocol.md
martinbonnin: Thanks

dev-docs/multipart-subscriptions-protocol.md
```diff
@@ -61,7 +59,9 @@ Some errors are part of an execution result. Results may include partial data, a
 }
 ```
 
-When the router encounters an error that is fatal and should lead to termination of the subscription, it will instead send a message with a top-level `errors` field, and then close the connection:
+## Fatal Errors
+
+When the router encounters an error that is fatal and should lead to termination of the subscription, `payload` is null:
```
alessbell, line 64: `payload: null` without `errors` seems semantically more similar to a heartbeat. Can we treat it as a heartbeat instead?

piclemx:
> For more context, this is the behaviour in Apollo iOS:
> 
> * `"payload": null` is effectively ignored and that alone will not throw an error - [test](https://github.com/apollographql/apollo-ios-dev/blob/469866288e7e4fc194e9ad9696c5f6e8c35756b6/Tests/ApolloTests/Interceptors/MultipartResponseSubscriptionTests.swift#L206)
> * If `"errors" : []` is present an error will be thrown - [test](https://github.com/apollographql/apollo-ios-dev/blob/469866288e7e4fc194e9ad9696c5f6e8c35756b6/Tests/ApolloTests/Interceptors/MultipartResponseSubscriptionTests.swift#L49)
> * We don't have a test that combines them but I put one together quickly and the behaviour is the same with `"payload": null` having no effect.
> 
> @alessbell has got a draft PR up at [apollographql/apollo-client#11594](https://github.com/apollographql/apollo-client/pull/11594) which matches that behavior.
> 
> ~Personally I think this spec change and ignoring `"payload": null` is the right change to make. I don't see any need to take action on an empty payload. I believe the context is which it's sent is to end a subscription and we don't take any particular action for that elsewhere.~
> 
> Update for ~strikeout~: @martinbonnin may have won me over with his thoughts on not introducing another way to terminate the subscription when we already have one. Which makes me reconsider whether we should not be treating this as another heartbeat (ignoring it) either. 🤔

Still what I found weird about this issue is that even if the fix is made on the client side (iOS,Kotlin,Javascript) the router is currently terminating the connection when receiving that kind of payload. Still think that the router shouldn't kill the connection.

When we were doing test on that feature, we saw that UI --> Subgraph everything was good, and we never got that payload mention in the issue.

When adding the router into the equation, this weird payload is now there and seems to be added by the router.

bnjjj:
When you receive this payload from the router it often means that the stream between router - subgraph has been closed. If it's not expected then there might be another issue somewhere. And maybe related to the missing ping message sent from the router to your subgraph ? It's an issue that has been recently used. What do you use as subgraph ?

anthonyvallee:
> When you receive this payload from the router it often means that the stream between router - subgraph has been closed.

If that can help, we reproduced the issue while capturing the traffic between (1) the client and router and (2) the router and subgraph.

The session between the router and subgraph is closed gracefully, as noted by the last exchange ("complete") between the router and subgraph. The router should gracefully close the subscription on the client side, but Apollo Client panics when handling the last message from the router.

[1-client-router-cap.txt](https://github.com/apollographql/router/files/14285784/1-client-router.txt)
[2-router-subgraph-cap.txt](https://github.com/apollographql/router/files/14285786/2-router-subgraph.txt)

bnjjj:
> The router should gracefully close the subscription on the client side, but Apollo Client panics when handling the last message from the router.
Yes @anthonyvallee that's the reason why we want to improve our specs docs to explain that could happen and it's not an error so the client should not throw an error. That being said is it expected on your side from your subgraph that the connection is gracefully close ? Were you able to understand why ?

anthonyvallee:
> Is it expected on your side from your subgraph that the connection is gracefully close ? Were you able to understand why ?

Yes, we designed a Ping subscription for our tests, similar to the `ping` system command. The subscription completes after N replies are sent to the client.

bnjjj:
On more detail:
`{"payload": null}` should not be considered as an error, it's not an error. It can happen and should just be ignored/skipped. The only thing you should check to know if it's a terminal error or not is the closing boundary `--graphql--`. If it's still unclosed `--graphl` it means the subscription is still alive and should not be closed. It's not a change in the specs, it's just a more detailed explanation

Issue #4019, opened 2023-10-11
context id is not preserved from router to supergraph stage in a batch


garypen:
When we create a batch, we create a new `Context` for each `Supergraph` that we generate from the `Router`. This means that the ID will be different for `Supergraph` stages to at the `Router` stage.

It's difficult to decide if this is the right thing to do or if we should clone the id from the `Router` onto the `Supergraph`. That also has some nice properties.

garypen:
Following careful review of this decision, within the context of #4661 , we've decided to continue with the current solution (creating a new `Context` for each `SupergraphRequest` that we generate from the batched `RouterRequest`).

In the future, this mechanism could be extended so that we preserve a reference to the `RouterRequest` context, or at least parts of that context (e.g.: context id). That work is not being performed as part of the current batching solution.

PR #4661, opened 2024-02-15
Subgraph support for query batching

### LINKED ISSUE #2002: Allow batched requests to subgraphs
For graph consumers is a very common scenario to have data fetching logic with the component that use it, causing many requests to a supergraph, to fix this we have the option of using batched requests to supergraphs, then the supergraph will concurrently and independently execute all operations in the batch. 

The problem with this is that, given that operations are executed independently, query planner will create a plan for each operation potentially causing a lot of requests to same subgraph

Example
image

This batched request to the supergraph runtime with 2 operations will cause 4 separated network requests to subgraphs.

At Expedia Group we found a solution for this problem in the gateway-js by adding the dataloader pattern between the communication of supergraph and subgraphs by pretty much having a queue on the request context and take advantage of the javascript microtask queue to dispatch data loader

image

Now for the router things are more interesting given the fact that there is no event loop so a custom dispatch mechanism will need to be implemented.

We believe that this feature is a must on supergraph given the way some clients write and execute requests against the supergraph -- performance, network optimization but also consistency are some of the reasons why we added this feature in our gateway implementation.

Now, in order to write a custom dispatch mechanism, i was thinking that maybe we can inspect each Query Plan for each operation in a batch, and have some some of expected requests vs current fired requests 

we will have n Query Plans and each Query Plan AST will have m levels
so it would be matter of check for each level how many requests we need to fire vs how many requests we have fired, i guess rust has a concept of multithreading, but applying a semaphore lock over a variable that holds the count of fired requests would help.

for graphql-kotlin we had to do something very similar to dispatch data loaders by tracking the synchronous execution of a graphQL operation, once the synchronous execution is exhausted, then we need to dispatch. 

Additional information

https://opensource.expediagroup.com/graphql-kotlin/docs/server/data-loader/data-loader-instrumentation#dispatching-by-synchronous-execution-exhaustion

garypen:
This project is an extension of the existing work to support [client side batching in the router](https://github.com/apollographql/router/issues/126).
The current implementation is experimental and is publicly [documented](https://www.apollographql.com/docs/router/executing-operations/query-batching/).
The additional work to enable batching requests to subgraphs is captured in [this issue](https://github.com/apollographql/router/issues/2002).
Currently the concept of a batch is preserved until the end of the RouterRequest processing. At this point we convert each batch request item into a separate SupergraphRequest. These are then planned and executed concurrently within the router and re-assembled into a batch when they complete. It's important to note that, with this implementation, the concept of a batch, from the perspective of an executing router, now disappears and each request is planned and executed separately.
This extension will modify the router so that the concept of a batch is preserved, at least outwardly, so that multiple subgraph requests are "batched" (in exactly the same format as a client batch request) for onward transmission to subgraphs. The goal of this work is to provide an optimisation by reducing the number of round-trips to a subgraph from the router.
Additionally, the work will address an [unresolved issue](https://github.com/apollographql/router/issues/4019) from the existing experimental implementation and promote the existing implementation from experimental to fully supported.

Fixes #2002 

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/query_planner/plan.rs
```diff
@@ -188,22 +188,45 @@ impl PlanNode {
         }
     }
 
-    pub(crate) fn subgraph_fetches(&self) -> usize {
+    /// Count the number of fetches
+    ///
+    pub(crate) fn subgraph_fetches(&self, include_requires: bool) -> usize {
```
samuelAndalon, line 193: i suspect that counting fetches is not going to be enough.
consider this use case: 

image

the query 
```graphql
query Astronaut {
  id
  name
  missions {
    designation
  }
} 
```
The query plan of above query consists in:
a SequenceNode with 2 children: a FetchNode (astronauts subgraph) and a FlattenNode,
the FlattenNode with a FetchNode (missionsSubgraph).

if we "count" the fetches, we have a grand total of 2, however, fetch #2 will never fire if fetch #1 does not complete first.

now imagine we send a batch of 2 graphqlRequests using above query.
the # of fetches will be 4, while realistically we won't be able to fire fetch #3 and fetch #4 before #1 and #2 completes.

IMHO this is only solvable with dataloaders and stop using async/await in query plan executor

garypen: @samuelAndalon Thanks for the feedback!

I wrote about this limitation (and the assumptions) in Section 2.1.1 of the shared design documentation and it is one of the major assumptions that we aim to test in the prototype.

I believe the prototype accounts for that by excluding fetches which have "required" fields from the total.

I'm finding this works fine in the prototyping graphs/queries I'm working with, but I'm not 100% sure it will work for all use cases and that's one of the things which will require more testing.

We'll need to move onto alternative approaches (or further refinements) if this assumption is proved invalid during prototyping.

smyrick:
Merge request to log the request body in debug mode: https://github.com/apollographql/router/pull/4840

.changesets/feat_garypen_2002_subgraph_batching.md
```diff
@@ -0,0 +1,27 @@
...
+  mode: batch_http_link
+  subgraph:
+    all:
+      enabled: true
+    subgraphs:
+      subgraph_1:
+        enabled: true
+      subgraph_2:
+        enabled:true
+````
```
shorgi, line 23: This is invalid to have both `all` and `subgraphs`, correct? If yes, should replace with separate configs for the two. Also, what's logged/returned by the router if this invalid config is set?

shorgi (commented):
WIP docs review comments

.changesets/feat_garypen_2002_subgraph_batching.md
garypen: I split the config sample into 2 so it was valid. If you did try to use this configuration, you'd see something like:
```
2024-04-03T08:48:41.981376Z ERROR  Failed to read configuration: configuration had errors: 
1. at line 13

  
  batching:
    enabled: true
    mode: batch_http_link
┌   subgraph:
|     all:
|       enabled: true
|     subgraphs:
|       subgraph_1:
|         enabled: true
|       subgraph_2:
|         enabled:true
└-----> {"all":{"enabled":true},"subgraphs":{"subgraph_1":{"enabled":true},"subgraph_2":"enabled:true"}} is not valid under any of the schemas listed in the 'oneOf' keyword


2024-04-03T08:48:41.981728Z INFO  stopped
2024-04-03T08:48:41.981984Z ERROR  no valid configuration was supplied
no valid configuration was supplied
```
which is the usual kind of error your would see for an "all" vs "specific" in router config.

fixed: https://github.com/apollographql/router/pull/4661/commits/7c138693a2f713ee9c0c404dbacc53f6a02cec30

docs/source/executing-operations/query-batching.mdx
```diff
@@ -11,24 +11,28 @@ Learn about query batching and how to configure the Apollo Router to receive que
...
+The Apollo Router supports client and subgraph query batching.
 
-Once configured, Apollo Client automatically combines multiple operations into a single HTTP request. The number of operations within a batch is client-configurable, including the maximum number in a batch and the maximum duration to wait for operations to accumulate before sending the batch. 
+If you’re using Apollo Client, you can leverage the built-in support for batching to reduce the number of individual operations sent to the router.
+
+Once configured, Apollo Client automatically combines multiple operations into a single HTTP request. The number of operations within a batch is client-configurable, including the maximum number in a batch and the maximum duration to wait for operations to accumulate before sending the batch.
 
 The Apollo Router must be configured to receive query batches, otherwise it rejects them. When processing a  batch, the router deserializes and processes each operation of a batch independently, and it responds to the client only after all operations of the batch have been completed. Each operation executes concurrently with respect to other operations in the batch.
```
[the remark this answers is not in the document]
garypen: I've added a paragraph. If it needs improvements, feel free to suggest them.

fixed: https://github.com/apollographql/router/commit/7c138693a2f713ee9c0c404dbacc53f6a02cec30

docs/source/executing-operations/query-batching.mdx
```diff
@@ -38,6 +49,121 @@ experimental_batching:
...
+
+| Subgraph | Fetch Count (without)| Fetch Count (with) |
+|----------|----------------------|--------------------|
+| accounts | 6                    | 2                  |
+| products | 1                    | 1                  |
+| reviews  | 2                    | 2                  |
+
+##### Example 2 Simple Subgraph Batching
+
+For queries which don't have required fetch constraints, subgraph batching is extremely efficient:
```
shorgi, line 119:
```suggestion
This example shows how the router can batch subgraph requests in the most efficient scenario, where the queries of a batch don't have required fetch constraints.

Assume the federated graph contains three subgraphs: `accounts`, `products`, and `reviews`.

The input client query to the federated graph:
```

docs/source/executing-operations/query-batching.mdx
```diff
@@ -38,6 +49,121 @@ experimental_batching:
...
+    {"query":"query MeQuery10 {\n  me {\n    name\n }\n}"},
+    {"query":"query MeQuery11 {\n  me {\n    id\n }\n}"}
+    {"query":"query MeQuery12 {\n  me {\n    name\n }\n}"},
+    {"query":"query MeQuery13 {\n  me {\n    id\n }\n}"}
+    {"query":"query MeQuery14 {\n  me {\n    name\n }\n}"},
+    {"query":"query MeQuery15 {\n  me {\n    id\n }\n}"}
+]
+```
+
+For that input and with the knowledge that our federated graph has 3 subgraphs, accounts, products and reviews, the router will generate a set of queries which look like this:
```
shorgi, line 141:
```suggestion
From the input query, the router generates a set of subgraph queries:
```

docs/source/executing-operations/query-batching.mdx
```diff
@@ -38,6 +49,121 @@ experimental_batching:
...
+"query MeQuery8__accounts__0{me{name}}",
+"query MeQuery9__accounts__0{me{id}}",
+"query MeQuery10__accounts__0{me{name}}",
+"query MeQuery11__accounts__0{me{id}}",
+"query MeQuery12__accounts__0{me{name}}",
+"query MeQuery13__accounts__0{me{id}}",
+"query MeQuery14__accounts__0{me{name}}",
+"query MeQuery15__accounts__0{me{id}}",
+```
+All of the queries may be batched. So, in this case, 15 fetches would be reduced to 1.
```
shorgi, line 159:
```suggestion
All of the queries can be combined into a single batch. So instead of 15 (non-batch) subgraph fetches, the router only has to make one fetch.
```

apollo-router/src/query_planner/plan.rs
```diff
@@ -192,6 +193,64 @@ impl PlanNode {
...
+    /// Populate a Vec of QueryHashes representing Fetches in this plan.
+    ///
+    /// Do not include any operations which contain "requires" elements.
+    ///
+    /// Think about the impact of Defer/Subscription/Condition on this function. In
+    /// particular, Condition processing might be causing us to mis-report the number of fetches
+    /// and that would be a big problem. Perhaps we should Abort Batches which contain
+    /// Defer/Subscription/Condition....?
+    ///
+    pub(crate) fn query_hashes(&self, query_hashes: &mut Vec<Arc<QueryHash>>) {
```
garypen, line 205: Note: I spoke with Derek about this and we don't need to accommodate Condition processing for this implementation of batching. I need to decide to convert this to return an error if Defer/Subscription/Condition is encountered (and make it iterative instead of recursive).

apollo-router/src/services/supergraph/service.rs
```diff
@@ -612,19 +613,40 @@ async fn plan_query(
...
+            plan.root.query_hashes(&mut query_hashes);
+            batch_query
+                .set_query_hashes(query_hashes)
+                .await
+                .map_err(|e| CacheResolverError::BatchingError(e.to_string()))?;
+            tracing::debug!("batch registered: {}", batch_query);
+        }
+
+        context.extensions().lock().insert(batch_query);
+    }
```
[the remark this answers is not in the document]
garypen: I know what you mean. It makes me nervous too. I *believe* that there are no races as currently written, since nothing should be interacting with this extension when this process is performed.
I suppose it would be possible to cause issue if you've spawned a long running task in RouterRequest processing that is monitoring the value of this extension. If you are doing things like that, then you probably have a lot of issues.

I could restructure this so that we don't make the call to `set_query_hashes()` under the lock. Something like:

```
    let mut query_hashes = vec![];
    if let Some(mut batch_query) = context.extensions().lock().get_mut::<BatchQuery>() {
        if let Some(QueryPlannerContent::Plan { plan, .. }) = &qpr.content {
            plan.root.query_hashes(&mut query_hashes);
        }
    }
    if !query_hashes.is_empty() {
            batch_query
                .set_query_hashes(query_hashes)
                .await
                .map_err(|e| CacheResolverError::BatchingError(e.to_string()))?;
            tracing::debug!("batch registered: {}", batch_query);
    }
```

apollo-router/src/services/subgraph_service.rs
```diff
@@ -408,11 +423,25 @@ impl tower::Service<SubgraphRequest> for SubgraphService {
             match get_apq_error(gql_response) {
                 APQError::PersistedQueryNotSupported => {
                     apq_enabled.store(false, Relaxed);
-                    call_http(request, body, context, client, &service_name).await
+                    call_batched_http(
```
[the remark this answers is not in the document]
garypen: How about
`call_batched_http` => `call_http`
`call_http` => `call_single_http`

?

apollo-router/src/services/subgraph_service.rs
```diff
@@ -761,11 +725,417 @@ async fn call_http(
...
+    client_factory: HttpClientServiceFactory,
+    service: String,
+    context: Context,
+    mut request: http::Request<hyper::Body>,
+    listener_count: usize,
+) -> Result<Vec<SubgraphResponse>, FetchError> {
+    // Now we need to "batch up" our data and send it to our subgraphs
+    // We need our own batch aware version of call_http which only makes one call to each
+    // subgraph, but is able to decode the responses. I'll probably need to break call_http
+    // down into sub-functions, and I've started this, but it's not finished.
```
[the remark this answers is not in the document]
garypen: done: eabebee4de8e14d9930c4aba85c9ff04102353d4

apollo-router/src/query_planner/plan.rs
garypen: done: be7da7877d57977d0f0425355b3a7380e48fdfa3

apollo-router/src/services/subgraph_service.rs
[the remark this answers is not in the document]
bnjjj: LGTM

apollo-router/src/batching.rs
```diff
@@ -0,0 +1,694 @@
...
+
+    /// A channel sender for sending updates to the entire batch
+    sender: Option<mpsc::Sender<BatchHandlerMessage>>,
+
+    /// How many more progress updates are we expecting to send?
+    remaining: usize,
+
+    /// Batch to which this BatchQuery belongs
+    batch: Arc<Batch>,
+}
```
Geal, line 49: could this structure be made Clone, Sync and Send? That would avoid the frequent danse of removing it from extensions then putting it back

apollo-router/src/services/subgraph_service.rs
```diff
@@ -761,11 +725,417 @@ async fn call_http(
...
+    // we have the context extensions lock held. That would be very bad...
+    // We grab the (potential) BatchQuery and then operate on it later
+    let opt_batch_query = {
+        let mut extensions_guard = context.extensions().lock();
+
+        // We need to make sure to remove the BatchQuery from the context as it holds a sender to
+        // the owning batch
+        extensions_guard
+            .get::<Batching>()
+            .and_then(|batching_config| batching_config.batch_include(service_name).then_some(()))
```
Geal, line 1023: `batch_include` is iterating over a list of strings. We have often seen in the router that this pattern creates performance issues. Could this list be converted to a HashSet?

apollo-router/src/services/supergraph/service.rs
```diff
@@ -619,19 +620,39 @@ async fn plan_query(
...
             "otel.kind" = "INTERNAL"
         ))
-        .await
+        .await?;
+
+    // We need to operate on the BatchQuery without holding on to the extension lock, but using the
+    // batched query requires awaiting it, which would hold the lock for too long.
+    //
+    // We remove the BatchQuery here and then reinsert it after we've done our call to fix that.
+    let batching = context.extensions().lock().remove::<BatchQuery>();
```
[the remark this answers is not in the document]
garypen: The model is:

[1, 2, 3] => Batch
1 => BatchQuery
2 => BatchQuery
3 => BatchQuery

(where `n` == a single query operation in a batch)

Each BatchQuery thus belongs to a single `SupergraphRequest/Response` (and is associated via `Context` extensions). Each `BatchQuery` is related to its owning `Batch` via an `Arc` shared reference.

I don't think this is racy because the only thing that *should* be interacting with that `BatchQuey` is not racing itself, but, [as I noted above](https://github.com/apollographql/router/pull/4661#discussion_r1555783780), it does seem clunky.

I will try to pursue your suggestion about Clone+Send+Sync to see if I can improve this, but I think there may be downsides to that as well.

apollo-router/src/services/subgraph_service.rs
```diff
@@ -761,11 +725,417 @@ async fn call_http(
...
+        // Map our Vec<u8> into Bytes
+        // Map our serde conversion error to a FetchError
+        let body = Some(
+            serde_json::to_vec(&object)
+                .map(|v| v.into())
+                .map_err(|error| FetchError::SubrequestMalformedResponse {
+                    service: service.to_string(),
+                    reason: error.to_string(),
+                }),
+        );
```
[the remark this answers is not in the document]
garypen: That code is doing the following:

809: Create a JSON object (hopefully an array) from our returned bytes.

821: Make sure it's a JSON array.

826: Iterate over our JSON array.

827: Make sure each thing we iterate over is a JSON object

835: Convert each individual object into individual response bytes for each batch member now that we know it really is a JSON object (single line in a batch)

So, we have to deserialize/serialize since we are breaking down an array of responses into individual responses.

We could try and validate that the raw array bytes "look" like a JSON array, but it seems cleaner/safer to use the serde_json array functionality to do this for us.

.changesets/feat_garypen_2002_subgraph_batching.md
```diff
@@ -0,0 +1,38 @@
...
+  subgraph:
+    # Configure batching support per subgraph
+    subgraphs:
+      subgraph_1:
+        enabled: true
+      subgraph_2:
+        enabled: true
+```
+
+Note: `all` and `subgraphs` are mutually exclusive. This applies in general for all router subgraph configuration options.
```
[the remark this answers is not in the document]
garypen: I did not know that. I also didn't realise we already had a `SubgraphConfiguration` struct that looked after that for us. I'll sort that out.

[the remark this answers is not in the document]
garypen: done: 204961edba5ace2abdba5dc8d041811d81abee84

apollo-router/src/services/subgraph_service.rs
garypen: I've changed `batch_include()` because subgraph batching config is now using the existing `SubgraphConfiguration` code.
done: 204961edba5ace2abdba5dc8d041811d81abee84

apollo-router/src/services/subgraph_service.rs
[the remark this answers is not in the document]
garypen: done: 204961edba5ace2abdba5dc8d041811d81abee84

apollo-router/src/services/supergraph/service.rs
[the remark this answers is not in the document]
garypen: I've raised a PR against this PR to explore this: https://github.com/apollographql/router/pull/4941

Please let me know what you think about that.

apollo-router/src/services/supergraph/service.rs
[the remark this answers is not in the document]
garypen: I've raised a PR against this PR to explore this: https://github.com/apollographql/router/pull/4941

Please let me know what you think about that.

apollo-router/src/batching.rs
garypen: I've raised a PR against this PR to explore this: https://github.com/apollographql/router/pull/4941

Please let me know what you think about that.

apollo-router/src/services/supergraph/service.rs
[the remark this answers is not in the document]
garypen: done: 86855eebe883f211274a51d9449eeea5e51a0aaf

apollo-router/src/services/supergraph/service.rs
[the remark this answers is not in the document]
garypen: done: 86855eebe883f211274a51d9449eeea5e51a0aaf

apollo-router/src/batching.rs
garypen: done: 86855eebe883f211274a51d9449eeea5e51a0aaf

docs/source/executing-operations/query-batching.mdx
```diff
@@ -38,6 +49,138 @@ experimental_batching:
...
+        enabled: true
+```
+
+<Note>
+
+- The router can be configured to support batching for either all subgraphs or individually enabled per subgraph.
+
+- There are limitations on the ability of the router to preserve batches from the client request into the subgraph requests. In particular, certain forms of queries will require data to be present before they are processed. Consequently, the router will only be able to generate batches from queries which are processed which don't contain such constraints. This may result in the router issuing multiple batches or requests. 
+
+- If [query deduplication](/graphos/configuration/traffic-shaping/#query-deduplication) is enabled, it will not apply to batched queries. Batching will take precedence over query deduplication. Query deduplication will still be performed for non-batched queries.
```
garypen, line 88: @shorgi Could you please confirm that this `query deduplication` link is correct? I can't get the docs stuff to run on my laptop any more (I get a SIGSEGV) so it's hard for me to confirm the URL is correct.

Meschreiber: It should be this:

```suggestion
- If [query deduplication](../configuration/traffic-shaping/#query-deduplication) is enabled, it will not apply to batched queries. Batching will take precedence over query deduplication. Query deduplication will still be performed for non-batched queries.
```

bnjjj (approved):

apollo-router/src/services/subgraph_service.rs
```diff
@@ -761,11 +725,410 @@ async fn call_http(
...
+        let extensions_guard = context.extensions().lock();
+
+        // We need to make sure to remove the BatchQuery from the context as it holds a sender to
+        // the owning batch
+        extensions_guard
+            .get::<Batching>()
+            .and_then(|batching_config| batching_config.batch_include(service_name).then_some(()))
+            .and_then(|_| extensions_guard.get::<BatchQuery>().cloned())
+            .and_then(|bq| (!bq.finished()).then_some(bq))
+        //TODO: I'm not sure that we need to check if bq is finished.
```
o0Ignition0o, line 1024: TODONE ?

apollo-router/src/services/router/service.rs
```diff
@@ -36,9 +36,12 @@ use tracing::Instrument;
 
 use super::ClientRequestAccepts;
 use crate::axum_factory::CanceledRequest;
+use crate::batching::Batch;
+use crate::batching::BatchQuery;
 use crate::cache::DeduplicatingCache;
 use crate::configuration::Batching;
 use crate::configuration::BatchingMode;
+// use crate::configuration::SubgraphBatchingConfig;
```
o0Ignition0o, line 44:
```suggestion
```

o0Ignition0o (approved):
Couple of nits, nothing that prevents a merge, lgtm.

apollo-router/src/plugins/traffic_shaping/deduplication.rs
```diff
@@ -73,6 +74,18 @@ where
...
     ) -> Result<SubgraphResponse, BoxError> {
+        // Check if the request is part of a batch. If it is, completely bypass dedup since it
+        // will break any request batches which this request is part of.
+        // This check is what enables Batching and Dedup to work together, so be very careful
+        // before making any changes to it.
+        if request
+            .context
+            .extensions()
+            .lock()
+            .contains_key::<BatchQuery>()
```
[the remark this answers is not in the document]
o0Ignition0o: Not quite, only if there's an else block https://github.com/rust-lang/rust/issues/12033 (but it's super confusing so I'm all for erring on the side of caution here)

apollo-router/tests/integration/batching.rs
```diff
@@ -0,0 +1,1026 @@
...
+
+#[tokio::test(flavor = "multi_thread")]
+async fn it_supports_single_subgraph_batching() -> Result<(), BoxError> {
+    const REQUEST_COUNT: usize = 5;
+
+    let requests: Vec<_> = (0..REQUEST_COUNT)
+        .map(|index| {
+            Request::fake_builder()
+                .query(format!(
+                    "query op{index}{{ entryA(count: {REQUEST_COUNT}) {{ index }} }}"
```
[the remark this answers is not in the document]
garypen: It works with federated queries. I've raised an issue to address this.

apollo-router/src/services/subgraph_service.rs
```diff
@@ -761,11 +725,410 @@ async fn call_http(
             .to_graphql_error(None),
         )
     }
+    graphql_response
+}
+
+/// Process a single subgraph batch request
+#[instrument(skip(client_factory, context, request))]
+pub(crate) async fn process_batch(
```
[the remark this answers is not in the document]
garypen: We did a lot of work to break common functionality out into new subsidiary functions (e.g.: `get_uri_details(), http_response_to_graphql_response()`. At some point I hoped that we'd do enough re-factoring that it would be easy to re-combine the main controlling functions (`process_batch`, `call_single_http`), but it's still not done.

I'll add this to the follow-up issues list.

apollo-router/src/services/subgraph_service.rs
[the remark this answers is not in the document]
garypen: Added to follow-up issue.

apollo-router/src/services/subgraph_service.rs
garypen: done: c33d0ca30e621185c1232c4fbcf43da6b0f060c4

apollo-router/src/plugins/traffic_shaping/deduplication.rs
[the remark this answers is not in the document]
garypen: done: c33d0ca30e621185c1232c4fbcf43da6b0f060c4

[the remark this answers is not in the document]
garypen: I'll revert this out.

docs/source/executing-operations/query-batching.mdx
```diff
@@ -1,34 +1,45 @@
 ---
 title: Query batching
-description: Receive query batches with the Apollo Router  
+description: Receive query batches with the Apollo Router
 ---
 
-<ExperimentalFeature />
```
chandrikas, line 6: We need to add the Enterprise tag to this feature

garypen: done: feb1769b41c3f652d27701ff94b08fcd0da19f3d

Geal (approved):

Issue #4544, opened 2024-01-25
trace sampling headers are not sent if the trace is not sampled


Geal:
It appears that if the router does not sample a trace (either because the client sets the headers or because there's a ratio sampling configured), then it won't set headers in subgraph requests to indicate if the trace should be sampled or not. It may be related to https://github.com/apollographql/router/issues/3226#issuecomment-1833658947

PR #4670, opened 2024-02-16
prep release: v1.40.1

abernix:
> **Note**
>
> When approved, this PR will merge into **the `1.40.1` branch** which will — upon being approved itself — merge into `main`.
>
>  - Version bumps
>  - That it targets the right release branch (`1.40.1` in this case!).
>
---
### Propagate tracing headers even when not sampling a trace ([Issue #4544](https://github.com/apollographql/router/issues/4544))

When the router was configured to sample only a portion of the trace, either through a ratio or using parent based sampling, and when trace propagation was configured, if a trace was not sampled, the router did not send the propagation headers to the subgraph. The subgraph was then unable to decide whether to record the trace or not. Now we make sure that trace headers will be sent even when a trace is not sampled.

By [@Geal](https://github.com/Geal) in https://github.com/apollographql/router/pull/4609


abernix:
Marking as a draft until the [`-rc.1` finishes compiling](https://app.circleci.com/pipelines/github/apollographql/router/19109/workflows/58d5e1e4-ccde-4677-8245-3c44fd0fb603) and makes its way through QA testing.

Once that QA is finished, this PR can be marked as ready for review, reviewed, and released.

lrlna (approved):

PR #4673, opened 2024-02-16
chore(ci): use latest Windows 2019 image 2024.02.21

Geal:
Updates our notion of `current` to the latest version `windows-server-2019-vs2019:2024.02.21` but _pins_ it, so it doesn't change randomly.


.circleci/config.yml
```diff
@@ -59,12 +59,12 @@ executors:
     resource_class: macos.m1.medium.gen1
   windows_build: &windows_build_executor
     machine:
-      image: "windows-server-2019-vs2019:2022.08.1"
+      image: "windows-server-2019-vs2019:current"
```
Geal, line 62:
```suggestion
      image: "windows-server-2022-vs2022:current"
```

.circleci/config.yml
```diff
@@ -59,12 +59,12 @@ executors:
...
     resource_class: windows.xlarge
     shell: bash.exe --login -eo pipefail
   windows_test: &windows_test_executor
     machine:
-      image: "windows-server-2019-vs2019:2022.08.1"
+      image: "windows-server-2019-vs2019:current"
```
Geal, line 67:
```suggestion
      image: "windows-server-2022-vs2022:current"
```

.circleci/config.yml
```diff
@@ -58,12 +58,12 @@ executors:
     resource_class: macos.m1.medium.gen1
   windows_build: &windows_build_executor
     machine:
-      image: "windows-server-2019-vs2019:2022.08.1"
+      image: "windows-server-2019-vs2019:current"
```
abernix: @Geal According to [the build output](https://app.circleci.com/pipelines/github/apollographql/router/20932/workflows/a3610262-3fed-46a9-b39c-cf12a9364be3/jobs/142038/parallel-runs/0/steps/0-0) when you ran this initial commit **(not the 2022 commit)**, the `current` value was `windows-server-2019-vs2019:2024.02.21`.   Since it worked, let's use **that** image explicitly, rather than `current`?  (And move on?)

I don't think we need to try and go for 2022, as this page from CircleCI doesn't indicate that the `2024.02.21` image is being deprecated.  A 2022 upgrade seems like a 2.0 thing.

abernix: Plus,  it looks like 2022 is already failing, so let's go for low hanging fruit?

Geal: sure,; the 2019 image is recent enough to be usable for a while

.circleci/config.yml
```diff
@@ -59,12 +59,12 @@ executors:
     resource_class: macos.m1.medium.gen1
   windows_build: &windows_build_executor
     machine:
-      image: "windows-server-2019-vs2019:2022.08.1"
+      image: "windows-server-2022-vs2022:current"
```
Geal, line 62:
```suggestion
      image: "windows-server-2019-vs2019:2024.02.21"
```

.circleci/config.yml
```diff
@@ -59,12 +59,12 @@ executors:
...
     resource_class: windows.xlarge
     shell: bash.exe --login -eo pipefail
   windows_test: &windows_test_executor
     machine:
-      image: "windows-server-2019-vs2019:2022.08.1"
+      image: "windows-server-2022-vs2022:current"
```
Geal, line 67:
```suggestion
      image: "windows-server-2019-vs2019:2024.02.21"
```

PR #4681, opened 2024-02-19
Fix chunk formatting in multipart protocol

### LINKED ISSUE #4634: Router (1.38.0) emitting a last {"payload":null} message to the client before completing Subscription
Customer is facing an issue when their subgraph (DGS) sends a subscription complete message. They observe Apollo Router (1.38.0) sometimes emits a last {"payload":null} message to the client before completing the subscription.
When this null payload is emitted, Apollo Client throws an error.

**Example: frames from start to complete, when null payload is emitted**

```
--graphql

{}
--graphql

{"payload":{"data":{"toolingPing":"Pong -> 1"}}}
--graphql

{"payload":null}
--graphql--`

**Example: frames from start to complete, when null payload is not emitted**
`--graphql

{}
--graphql

{"payload":{"data":{"toolingPing":"Pong -> 1"}}}
--graphql--
```

**For Router, they are using v1.38.0.**

**They are also on Apollo Client 3.8.10.**

bnjjj:
This PR changes the way we're sending chunks in the stream. Instead of finishing the chunk with `\r\n` we don't send this at the end of our current chunk but instead at the beginning of the next one. For the end users nothing changes but it let us to close the stream with the right final boundary by appending `--\r\n` directly to the last chunk.

Fixes #4634

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [x] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/.changesets/fix_bnjjj_multipart_test.md
+++ b/.changesets/fix_bnjjj_multipart_test.md
@@ -1,0 +1,6 @@
+### Fix chunk formatting in multipart protocol ([Issue #4634](https://github.com/apollographql/router/issues/4634))
+
+This PR changes the way we're sending chunks in the stream. Instead of finishing the chunk with `\r\n` we don't send this at the end of our current chunk but instead at the beginning of the next one. For the end users nothing changes but it let us to close the stream with the right final boundary by appending `--\r\n` directly to the last chunk.
+
+
+By [@bnjjj](https://github.com/bnjjj) in https://github.com/apollographql/router/pull/4681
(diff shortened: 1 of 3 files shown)


calvincestari:
Is it correct to assume that this change will affect the `@defer` messages too since they are also multipart protocol messages?

bnjjj:
@calvincestari yes

alessbell (approved):
Tested with Apollo Client (both defer and multipart subscriptions over HTTP), lgtm.

calvincestari (approved):
I've got a local router running on this branch and Apollo iOS looks all good, receiving HTTP subscription messages as expected.

martinbonnin (approved):

o0Ignition0o (approved):

Geal (approved):

PR #4696, opened 2024-02-21
Set a limit of simultaneous client subscription connections to the router by default

### LINKED ISSUE #4405: subscription: put a default value for `subscription.max_opened_subscriptions`
In order to harden the configuration for subscription we should put a default value for that limit of opened subscriptions set to `8000` by default., it should also support `disabled` as value to not set a limit if we want. And document it properly [here](https://www.apollographql.com/docs/router/executing-operations/subscription-support#limiting-the-number-of-client-connections)

bnjjj:
In order to harden our default configuration I set a limit on opened subscriptions to 8000 for now. This number can be modify we can find a better default number.

Fixes #4405 

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/configuration/snapshots/apollo_router__configuration__tests__schema_generation.snap
+++ b/apollo-router/src/configuration/snapshots/apollo_router__configuration__tests__schema_generation.snap
@@ -2215,12 +2215,21 @@ expression: "&schema"
           "type": "boolean"
         },
         "max_opened_subscriptions": {
-          "description": "This is a limit to only have maximum X opened subscriptions at the same time. By default if it's not set there is no limit.",
-          "default": null,
-          "type": "integer",
-          "format": "uint",
-          "minimum": 0.0,
-          "nullable": true
+          "description": "This is a limit to only have maximum X opened subscriptions at the same time. By default it's set to 8000. If you want to disable the limit, just override it to `null`",
+          "default": 8000,
+          "anyOf": [
+            {
+              "type": "string",
+              "enum": [
+                "disabled"
+              ]
+            },
+            {
+              "type": "integer",
+              "format": "uint",
+              "minimum": 0.0
+            }
+          ]
         },
         "mode": {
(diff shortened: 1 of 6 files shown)


apollo-router/src/configuration/snapshots/apollo_router__configuration__tests__schema_generation.snap
```diff
@@ -2215,12 +2215,21 @@ expression: "&schema"
           "type": "boolean"
         },
         "max_opened_subscriptions": {
-          "description": "This is a limit to only have maximum X opened subscriptions at the same time. By default if it's not set there is no limit.",
-          "default": null,
-          "type": "integer",
-          "format": "uint",
-          "minimum": 0.0,
-          "nullable": true
+          "description": "This is a limit to only have maximum X opened subscriptions at the same time. By default it's set to 8000. If you want to disable the limit, just override it to `null`",
```
o0Ignition0o, line 2218:
```suggestion
          "description": "This is a limit to only have maximum X opened subscriptions at the same time. By default it's set to 8000. Set it to `null` to disable the limit.",
```

o0Ignition0o (approved):
Small suggestion, otherwise lgtm.

apollo-router/src/configuration/snapshots/apollo_router__configuration__tests__schema_generation.snap
bnjjj: Fixed

abernix:
@bnjjj Should we discuss this with the team and figure out next steps?

abernix:
@bnjjj You can certainly feel free to re-open this, it seemed reasonable, maybe even specifically for 2.x, but I'll close it for now. :)

Issue #4649, opened 2024-02-14
Enable `experimental_api_schema_generation_mode: both` by default


goto-bus-stop:
Following up on https://github.com/apollographql/router/pull/4510

We now have a Rust implementation of API schema. To test it further, we should run it side by side with the JS implementation and compare the results. The code for this is partially in the router already, but it is not robust enough to be enabled by default.

To address:
- [x] Sort API schema types and directives similarly to the JS implementation
- [x] When comparing JS and Rust result, normalize even more--TBD exactly what is needed but copying from our internal apollo-rs testing harness should work. We have to re-parse and re-serialize the API schema with apollo-rs to ensure we can do a string comparison. Luckily the performance impact is negligible compared to the time it takes to call the JS API schema implementation 😄 
- [x] https://github.com/apollographql/federation-next/pull/223

Then we can turn it on by default.

PR #4701, opened 2024-02-22
Update experimental API schema and implement robust comparison

goto-bus-stop:
This adds the standardisation code that I used successfully for harness runs, so we should get quite robust results.

I tested this manually by adding a `panic!()` in the mismatch case and intentionally adding a bug in the fed-next implementation locally.

Ref #4649

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


goto-bus-stop:
Wanna do one more apollo-federation release to **really** get this in order.

apollo-router/src/router_factory.rs
```diff
@@ -822,7 +822,7 @@ mod test {
             serde_json::from_value::<crate::plugins::telemetry::config::Conf>(config).unwrap();
         assert_eq!(
             &config.apollo.schema_id,
-            "ba573b479c8b3fa273f439b26b9eda700152341d897f18090d52cd073b15f909"
+            "6af283f857f47055b0069547a8ee21c942c2c72ceebbcaabf78a42f0d1786318"
```
goto-bus-stop, line 825: This changed because the previous implementation of `Schema::parse_test` (through `do_api_schema` in router-bridge) did not specify `include_defer` based on configuration. `@defer` was never included. The new implementation does pass through `include_defer` based on provided configuration, so `@defer` is included in API schema in tests just like it would be in a real environment. I inspected the schemas and this is the only difference.

apollo-router/src/spec/schema.rs
```diff
@@ -587,7 +579,7 @@ mod tests {
             assert_eq!(
                 schema.api_schema().schema_id,
                 Some(
-                    "ba573b479c8b3fa273f439b26b9eda700152341d897f18090d52cd073b15f909".to_string()
+                    "6af283f857f47055b0069547a8ee21c942c2c72ceebbcaabf78a42f0d1786318".to_string()
```
goto-bus-stop, line 582: Same as https://github.com/apollographql/router/pull/4701/files#r1514211589

apollo-router/src/spec/schema.rs
```diff
@@ -601,14 +593,9 @@ mod tests {
...
-Input field "InputObject.privateField" is @inaccessible but is used in the default value of "@foo(someArg:)", which is in the API schema.
-
-GraphQL request:42:1
-41 |
-42 | input InputObject {
-   | ^
-43 |   someField: String"#
+                    r#"The supergraph schema failed to produce a valid API schema: The following errors occurred:
+
+  - Input field `InputObject.privateField` is @inaccessible but is used in the default value of `@foo(someArg:)`, which is in the API schema."#
```
goto-bus-stop, line 598: The formatting is a bit worse here but we would follow up on that once QP itself is getting to a usable shape

docs/source/configuration/overview.mdx
```diff
@@ -810,6 +810,16 @@ experimental_graphql_validation_mode: new
 
 This is an experimental option while we are still finding edge cases in the new implementation, but it will become the default in the future.
 
+### API Schema Generation Mode
```
goto-bus-stop, line 813: I documented this option but I don't think we should really expect users to use this, the visible impact of the change and the risk of something being wrong is *much* lower than it was with `experimental_graphql_validation_mode`. We could just do it silently?

.changesets/feat_renee_api_schema.md
```diff
@@ -0,0 +1,11 @@
+## Experimental support for API Schema generation in Rust
+
+As part of the process of replacing JavaScript code in the Router with more reliable and performant Rust implementations, the API schema can now be generated in Rust. The new `experimental_api_schema_generation_mode` option controls which implementation is used.
+
+```yaml {4,8} title="router.yaml"
```
lrlna, line 5: TIL you can do this in markdown formatting?? very cool

goto-bus-stop: It's just copy pasted from the docs code which does something special, I don't think it does anything in github markdown :)

apollo-router/src/query_planner/bridge_query_planner.rs
```diff
@@ -191,45 +191,37 @@ impl BridgeQueryPlanner {
...
+                                    monotonic_counter.apollo.router.lifecycle.api_schema = 1u64,
+                                    generation.is_matched = false,
+                                    "API schema generation mismatch: apollo-federation and router-bridge write different schema"
+                                );
+
+                                let differences = diff::lines(&left, &right);
+                                tracing::debug!(
+                                    "different API schema between apollo-federation and router-bridge:\n{}",
+                                    render_diff(&differences),
+                                );
```
lrlna, line 220: how do we get to see the diff for customer schemas? if we spot a mismatch in our metrics, should we reach out to people to give us this debug log? i don't mind this as it does get us talking to people who could be potentially helpful when we start swapping out the query planner, just want to make sure we are on the same page on how we will get to this log.

lrlna (approved):

docs/source/configuration/overview.mdx
lrlna: I think we should do it silently.

apollo-router/src/query_planner/bridge_query_planner.rs
goto-bus-stop: Yeah we won't see this log. But since this is about supergraph schemas, we can very easily collect them from Studio and run the comparison ourselves

apollo-router/src/query_planner/bridge_query_planner.rs
```diff
@@ -819,6 +811,153 @@ impl QueryPlan {
...
+                for value in list {
+                    standardize_value_for_comparison(value.make_mut());
+                }
+            }
+            _ => {}
+        }
+    }
+
+    fn standardize_directive_for_comparison(directive: &mut apollo_compiler::ast::Directive) {
+        for arg in &mut directive.arguments {
```
Geal, line 836: should arguments be sorted or are we sure they are generated in the same way on both js and rust?

Geal (approved):

goto-bus-stop:
Filing a separate PR to change the default to `both` that we can land according to the release calendar, meanwhile this can land with the apollo-federation update to support early experimentation with QP integration

PR #4710, opened 2024-02-22
Provide unsupported access to the parsed executable document.

BrynCooke:
This is unsupported because apollo_rs is not at 1.0 and we may change things later.

If you decide to use `unsupported_executable_document` it is at your own risk.

We WILL be changing this later once apollo-rs reaches 1.0.

ExecutableDocument has to be moved into an `Arc` so that we can expose it.


goto-bus-stop (approved):
I'm content with the level of discouragement in this implementation :)

SimonSapin (approved):

PR #4713, opened 2024-02-22
Add a metric for federation version

goto-bus-stop:
To get a better understanding of how many users are still on fed v1, this PR adds a metric.

I put it under the `lifecycle` key, is that appropriate?

The version is inferred based on the version of the join spec that is used, as this is closely tied to the federation version. The federation minor version is not available in a supergraph as far as I know (it could only be inferred based on features that are in use, but that's not reliable if someone doesn't use every feature).

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


goto-bus-stop:
I think schema metrics processing in our backend is currently hardcoded by full metric name, not by `apollo.router.schema.` prefix, so adding a new metric under it doesn't get processed.

PR #4718, opened 2024-02-22
feat: ignore other auth prefixes 

lleadbet:
This PR enables the JWT plugin to ignore other auth prefixes other than the one defined in the configuration. As to not break existing uses, this is defaulting to be false. 

This enables multiple Authorization schemes to be supported, although other schemes would need to be handled in either Rhai or a coprocessor to be used with the AuthN plugin. 

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

Empty prefixes will not be affected by this change as there is no prefix to ignore. 

**Notes**


.changesets/feat_lleadbet_ignore_other_auth_prefixes.md
```diff
@@ -0,0 +1,20 @@
+### Ability to ignore other auth prefixes in the JWT plugin
+
+You can now ignore whether to ignore other header prefixes with the JWT plugin. As many applications will use the format of `Authorization: <scheme> <token>`, this will enable the use other schemes within the `Authorization` header.
```
garypen, line 3:
```suggestion
You can now choose whether to ignore other header prefixes with the JWT plugin. Many applications will use the format of `Authorization: <scheme> <token>` and this will enable the use of other schemes within the `Authorization` header. 
```

apollo-router/src/plugins/authentication/mod.rs
```diff
@@ -181,6 +184,10 @@ fn default_poll_interval() -> Duration {
     DEFAULT_AUTHENTICATION_DOWNLOAD_INTERVAL
 }
 
+fn default_ignore_other_prefixes() -> bool {
+    false
+}
+
```
garypen, line 190: You can delete this if you make the change above.

apollo-router/src/plugins/authentication/mod.rs
```diff
@@ -532,11 +539,13 @@ fn authenticate(
 
     // Make sure the format of our message matches our expectations
     // Technically, the spec is case sensitive, but let's accept
-    // case variations
-    //
+    // case variations. Furthermore, if the user has configured to ignore
```
garypen, line 542: It would be simpler as:
```
if jwt_value.len() < prefix_len
            || !&jwt_value[..prefix_len].eq_ignore_ascii_case(&config.header_value_prefix {
    if config.ignore_other_prefixes {
        return ControlFlow::Continue(request);
    } else {
        return failure_message(
            request.context,
        );
    }
}
```

apollo-router/src/plugins/authentication/tests.rs
```diff
@@ -36,13 +36,14 @@ fn create_an_url(filename: &str) -> String {
 }
 
 async fn build_a_default_test_harness() -> router::BoxCloneService {
-    build_a_test_harness(None, None, false).await
+    build_a_test_harness(None, None, true, None).await
```
garypen, line 39: I assume changing `false` to `true` doesn't affect any existing tests.

apollo-router/src/plugins/authentication/tests.rs
```diff
@@ -36,13 +36,14 @@ fn create_an_url(filename: &str) -> String {
...
 }
 
 async fn build_a_test_harness(
     header_name: Option<String>,
     header_value_prefix: Option<String>,
     multiple_jwks: bool,
+    ignore_other_prefixes: Option<bool>,
```
garypen, line 46: I don't think this needs to be an `Option`.
```suggestion
    ignore_other_prefixes: bool,
```
Anywhere you check `None` can be replaced with a check for `false`.

docs/source/configuration/authn-jwt.mdx
```diff
@@ -133,6 +133,23 @@ The default value is `Bearer`.
...
 
+<tr>
+<td style="min-width: 150px;">
+
+##### `ignore_other_prefixes`
+
+</td>
+<td>
+
+Whether to ignore other prefixes in the `Authorization` header. If set to `false`, the router will only accept tokens with the prefix specified in `header_value_prefix`. If set to `true`, the router will ignore any requests that don't start with the prefix specified in `header_value_prefix`.
```
garypen, line 144:
```suggestion
Whether to ignore other prefixes in the `Authorization` header. If set to `false`, or unspecified, the router will only accept tokens with the prefix specified in `header_value_prefix`. If set to `true`, the router will ignore any requests that don't start with the prefix specified in `header_value_prefix`.
```

garypen (commented):
Most of my comments are NITS or advisory. One thing that bothers me though.

If this option exists and is set to `true`, doesn't that mean that an attacker could bypass JWT authentication by simply setting another prefix? I know that the burden of responsibility then lies on the user to perform that authentication, but it still bothers me that the config is providing this open door as part of our JWT scheme. I feel like we need to document somewhere in big RED LETTERS what the implications of using this are.

lleadbet:
> If this option exists and is set to true, doesn't that mean that an attacker could bypass JWT authentication by simply setting another prefix? I know that the burden of responsibility then lies on the user to perform that authentication, but it still bothers me that the config is providing this open door as part of our JWT scheme. I feel like we need to document somewhere in big RED LETTERS what the implications of using this are.

I agree. It's definitely a fine line to tread, but some folks do `bearer` for jwts, `basic` for token-based, or any mix. 

If we could in the future have the plugin do auth schemes based on prefixes (e.g. bearer is now the JWT plugin) that might solve for some of the concern. 

And to be honest- the changing of headers would still be applicable for any other system that does auth. We (SAs) still highly recommend validating the token in the subgraph regardless. 

Going to flag up to sec and see what they say, however.

apollo-router/src/plugins/authentication/mod.rs
```diff
@@ -128,6 +128,9 @@ struct JWTConf {
     /// Header value prefix
     #[serde(default = "default_header_value_prefix")]
     header_value_prefix: String,
+    /// Whether to ignore any mismatched prefixes
+    #[serde(default)]
+    ignore_other_prefixes: bool,
```
Geal, line 133: I'm not sure we actually need an option. The router looks for the value with the prefix it needs, it ignores the rest, we still have the same behavior as before: if it does not find the value it wants, then the request will not be authenticated

apollo-router/src/plugins/authentication/tests.rs
```diff
@@ -424,9 +428,44 @@ async fn it_accepts_when_auth_prefix_has_correct_format_and_valid_jwt() {
     assert_eq!(expected_mock_response_data, response.data.as_ref().unwrap());
 }
 
+#[tokio::test]
+async fn it_accepts_when_auth_prefix_does_not_match_config_and_is_ignored() {
+    let test_harness = build_a_test_harness(None, None, false, true).await;
+    // Let's create a request with our operation name
+    let request_with_appropriate_name = supergraph::Request::canned_builder()
+        .operation_name("me".to_string())
+        .header(http::header::AUTHORIZATION, "Basic dXNlcjpwYXNzd29yZA==")
```
Geal, line 437: What happens when the `Authorization` header has multiple values?

apollo-router/src/plugins/authentication/mod.rs
Geal: the important part is whether `require_authentication` is enabled

apollo-router/src/plugins/authentication/mod.rs
```diff
@@ -571,7 +573,7 @@ fn authenticate(
 
     let mut jwt = None;
     for source in &config.sources {
-        match extract_jwt(source, request.router_request.headers()) {
+        match extract_jwt(source, config.ignore_other_prefixes, request.router_request.headers()) {
```
Geal, line 576:
```suggestion
        match extract_jwt(
            source,
            config.ignore_other_prefixes,
            request.router_request.headers()
        ) {
```

apollo-router/src/plugins/authentication/tests.rs
Geal: ok, this is something I'll look at as a follow up

Geal (approved):

PR #4722, opened 2024-02-22
Make integration test router bind to random port

### LINKED ISSUE #4720: Fix integration test port binding
The current integration test always binds the router to port 4000, but it would be better to have the port randomly assigned so that we can avoid port collisions.

Alternatives would be to not use networking for integration tests, but part of testing the whole stack means testing the network aspects so this cannot be avoided.

**Additional context**
By having the router always bind to port 4000 and all other portions of the test to random open ports we risk having tests fail non-deterministically due to port collisions.

nicholascioli:
Previously, the integration test would spin up the router with its listener always binding to port 4000, which might cause tests to fail if a different process were to bind to the same port.

This commit makes the test router always bind to an available port and corrects all calls to the router to use this new port.

Fixes #4720 

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


Geal:
did you see tests failing because they were conflicting with each other?

nicholascioli:
We had some flaky tests that failed to bind to port 4000, so I figured that it might be due to the fact that each subgraph binds to a random port and might collide with the router since it starts after.

Geal (approved):
this looks fine but maybe we need a different approach here in the future, because as you point out, another process could still try to reserve the listen port.
We could make the router listen on a unix socket instead. It is supported, and it would be easy to make sure that there is no conflict there. The issue here is that we really want to test the router over TCP in integration tests since that's the common use case.
Another solution would be to create the listen socket from inside the test and then give it to the router on startup. There are various mechanisms for that like [sd_listen_fds](https://manpages.debian.org/testing/libelogind-dev-doc/sd_listen_fds.3.en.html).

abernix:
Are we sure that another process could try and listen on the same port?  0 would be assigned by the kernel, wouldn't it?

abernix:
Ah, nevermind.  I see why this is the case, given how it's wired up.  Still, this is probably better.

PR #4726, opened 2024-02-23
Modify fmt_layer on_event() processing to help macOS

garypen:
Empirically determine that it's not a good idea to try to write more than 1_024 bytes to a non-blocking fd on macOS.

Modify the on_event() processing to take account of this.

### DIFF
--- a/apollo-router/src/plugins/telemetry/fmt_layer.rs
+++ b/apollo-router/src/plugins/telemetry/fmt_layer.rs
@@ -1,10 +1,10 @@
+use std::cell::RefCell;
 use std::collections::HashMap;
 use std::collections::HashSet;
 use std::io::IsTerminal;
 use std::io::Write;
 use std::marker::PhantomData;
 
-use bytes::BytesMut;
 use opentelemetry::Key;
 use opentelemetry::KeyValue;
 use tracing::field;
@@ -30,8 +30,6 @@ use crate::plugins::telemetry::formatters::text::Text;
 use crate::plugins::telemetry::formatters::FilteringFormatter;
 use crate::plugins::telemetry::reload::LayeredTracer;
 use crate::plugins::telemetry::resource::ConfigResource;
-
-const WRITER_BUF_SIZE: usize = 8_192;
 
 pub(crate) fn create_fmt_layer(
     config: &config::Conf,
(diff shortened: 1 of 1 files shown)


apollo-router/src/plugins/telemetry/fmt_layer.rs
```diff
@@ -167,28 +167,56 @@ where
...
+                    while index < limit {
+                        // NOTE: Do not raise this limit about 1_024 for macOS without testing it works
+                        // correctly. This figure was determined experimentally by observing pauses in
+                        // logging during startup. If the figure goes above 1_024, you will get a 5s
+                        // pause during logging at startup on macOS. This may change in the future or
+                        // depending on your version of macOS, so we'll need to verify it a couple of
+                        // times a year to see if anything has changed.
+                        #[cfg(target_os = "macos")]
+                        let upper = index + std::cmp::min(limit - index, 1_024);
+                        #[cfg(not(target_os = "macos"))]
```
Geal, line 186: why not keep the `write_all` for Linux and Windows?

Geal (approved):

apollo-router/src/plugins/telemetry/fmt_layer.rs
```diff
@@ -167,28 +167,56 @@ where
...
+                            Err(err) => {
+                                if !carry_on || err.kind() == std::io::ErrorKind::BrokenPipe {
+                                    break;
+                                }
+                                if err.kind() != std::io::ErrorKind::WouldBlock {
+                                    carry_on = false;
+                                    let _ = std::io::stderr().write_all(
+                                        format!("error during write: {err:?}\n").as_bytes(),
+                                    );
+                                    let _ = std::io::stderr().flush();
```
Geal, line 205: maybe add a break here. If we could not write that log we should not insist

garypen: We try twice (controlled by `carry_on`) and then break whilst ignoring any errors that may occur during writes to `stderr`.
I'm trying to trade off between immediately failing for a temporary out of pipe space type problem and trying too many times. Twice seems reasonable, but it's only an opinion.

goto-bus-stop:
Hmm, I still get test failures with this, for example with:
```
cargo run --test telemetry_test bad_query -- --nocapture
```
but that might be a second issue?

garypen:
> Hmm, I still get test failures with this, for example with:
> 
> ```
> cargo run --test telemetry_test bad_query -- --nocapture
> ```
> 
> but that might be a second issue?

I ran `cargo test --test telemetry_test bad_query -- --nocapture` and sure enough it failed. That's because I've only fixed the fmt_layer (i.e.: logging) part of the router IO. The test exercises `stderr` and breaks as you would expect.

If you make this change:

garypen@Garys-MacBook-Pro router % git diff
diff --git a/apollo-router/src/executable.rs b/apollo-router/src/executable.rs
index 1be97f684..3e10a4d60 100644
--- a/apollo-router/src/executable.rs
+++ b/apollo-router/src/executable.rs
@@ -443,7 +443,7 @@ impl Executable {
         #[cfg(not(target_os = "windows"))]
         {
             let _ = set_blocking(libc::STDOUT_FILENO, false);
-            let _ = set_blocking(libc::STDERR_FILENO, false);
+            // let _ = set_blocking(libc::STDERR_FILENO, false);
         }
then the test will pass.

garypen:
Maybe a better approach is to forget about non-blocking IO on macOS as well as Windows?

goto-bus-stop:
> Maybe a better approach is to forget about non-blocking IO on macOS as well as Windows?

I think that makes sense

garypen:
We reverted the router change so it isn't using non-blocking IO, so we don't need this. If we do have non-blocking IO in the router at some point, it will be more comprehensive than the reverted change, so we won't need this.

PR #4731, opened 2024-02-26
Wrap `UsageReporting` in `Arc` to avoid expensive clone

xuorig:
Cloning the `UsageReporting` structure to use in extensions (https://github.com/apollographql/router/blob/dev/apollo-router/src/query_planner/caching_query_planner.rs#L390) has a fairly higher cost that I would expect. I didn't realize it held a map of referenced fields which can grow large (https://github.com/apollographql/federation-rs/blob/35052d8064e7ca50d09226febdbbe0d55cc5d18b/federation-2/router-bridge/src/planner.rs#L313)

The change has less impact than I thought on the rest of the code, sounds like it may be worth it to wrap into an `Arc`. Thoughts?


xuorig:
Ah crap, it's the extensions that are broken (`get::<UsageReporting>()`) Not sure how to proceed without breaking changes 😢

Geal (approved):

Geal:
looks like the CI is not running, is it because the branch is on your repository?

xuorig:
> looks like the CI is not running, is it because the branch is on your repository?

Yeah no idea 🤔 

@Geal this is a breaking change to the type map. Not sure if we can go ahead with this unfortunately. Maybe bundle that with the `Arc<Query>` one in the next major version?

Geal:
`UsageReporting` is not exposed in the public API of the router, so it's fine. Anything in the context extensions that uses a private type will only be accessible from inside the router's code

xuorig:
The problem is the type is public I believe: https://github.com/apollographql/federation-rs/blob/35052d8064e7ca50d09226febdbbe0d55cc5d18b/federation-2/router-bridge/src/planner.rs#L306

I don't think there's a private wrapper for it in the router?

Geal:
the type is public in router-bridge, but the apollo-router crate does not reexport it: https://docs.rs/apollo-router/latest/apollo_router/?search=usagereporting

xuorig:
huh TIL, didn't know it had to be reexported as part of router. Thought the type ID just had to match. Thanks @Geal

Geal:
technically, if someone uses the apollo-router crate and imports router-bridge manually too, then yes they would be able to access that context key. But I'd be willing to disregard that as a breaking change

xuorig:
>  technically, if someone uses the apollo-router crate and imports router-bridge manually too, then yes they would be able to access that context key. But I'd be willing to disregard that as a breaking change

Ohhh I see, we're on the same page then 👍

xuorig:
Ready to run CI on this one 🤖 , changelog added.

abernix:
@xuorig We have our repo setup to run on forked repos, so I'm not sure why it won't run.  Our CircleCI logs are, for whatever reason, telling US, to tell you to log out and log back in to CircleCI.  I do wonder if this has something to do with both of our organizations having SSO setups or something, but to see how this goes, I've temporarily adopted your PR and pushed it to https://github.com/apollographql/router/tree/abernix/xuorig/arc-usage-reporting which has kicked off [this CircleCI run](https://app.circleci.com/pipelines/github/apollographql/router/19510/workflows/ee55f209-7e5f-4207-9c84-3e877ed11cea), if you want to watch it.  I did have to merge in our `dev` trunk, but it's running now.  If goes through, I'll open a PR and be sure to give you commit credit.

abernix:
It _does_ seem to be doing things now, yes! 😸

abernix:
(Oh, don't worry about the Netlify Deploy preview.  That definitely won't work for you because we don't pay for it to work.)

xuorig:
image

xuorig:
Screenshot 2024-03-08 at 08 10 52

apollo-router/src/plugins/telemetry/mod.rs
```diff
@@ -1311,6 +1312,8 @@ impl Telemetry {
                 }
             }
         } else {
+            println!("MISSING REPORTING");
```
bnjjj, line 1315: Does it still need to be there ? If you want to log something could you use our logging macros instead and use the right log level please

xuorig: Oof thanks for the catch @bnjjj, this was me just debugging, will remove.

Geal:
Closing this as it was merged in https://github.com/apollographql/router/pull/4752

PR #2097, opened 2022-11-14
different status codes from rhai

garypen:
enhance the existing rhai error handling implementation so that specific status codes can be set.

The change is fully backwards compatible (i.e.: the existing method still works), but if a user wishes to specify the HTTP status code set in the client response, they need only `throw` an objectmap as per the examples in NEXT_CHANGELOG.md.

fixes: #2023

PR #4733, opened 2024-02-28
docs: router HTTP response status codes

shorgi:
Document router response status codes in [new errors page](https://deploy-preview-4733--apollo-router-docs.netlify.app/errors)

Resolves https://github.com/apollographql/router/issues/4740 and https://apollographql.atlassian.net/browse/DOC-34

### DIFF
--- a/docs/source/errors.mdx
+++ b/docs/source/errors.mdx
@@ -1,7 +1,7 @@
 ---
 title: Errors
 subtitle: Error and status codes returned by the Apollo Router
-description: Error codes and response status codes returned by the Apollo Router
+description: Error codes and HTTP response status codes returned by the Apollo Router
 ---
 
 Learn about error codes and HTTP response status codes returned by the Apollo Router.
@@ -11,14 +11,14 @@ Learn about error codes and HTTP response status codes returned by the Apollo Ro
 <PropertyList kind="errCodes">
 <Property name="400" short="Bad request">
 
-The request failed GraphQL validation or failed to be parsed.
+A request failed GraphQL validation or failed to be parsed.
 
 </Property>
 <Property name="401" short="Unauthorized">
 
 Requests may receive this response in two cases:
 
-- For requests that requires authentication, the client's JWT failed verification.
+- For a client request that requires authentication, the client's JWT failed verification.
 - For a non-client subscription endpoint calling a subscription callback URL, the router couldn't find a matching subscription identifier between its registered subscriptions and a subscription event.
 
 </Property>
@@ -35,7 +35,7 @@ Both mutations and subscriptions must use POST.
 </Property>
 <Property name="406" short="Not acceptable">
 
-A query's HTTP `Accept` header didn't contain any of the router's supported mime-types: 
+A request's HTTP `Accept` header didn't contain any of the router's supported mime-types: 
 - `application/json`
 - `application/graphql-response+json`
 - `multipart/mixed;deferSpec=20220824`
--- a/docs/source/errors.mdx
+++ b/docs/source/errors.mdx
@@ -26,7 +26,7 @@ Used an unallowed HTTP method. Both mutations and subscriptions must use POST.
 </Property>
 <Property name="406" short="Not acceptable">
 
-A query's Accept header doesn't match one of the supported content types: 
+A query's HTTP `Accept` header doesn't contain any of the Router's supported mime-types: 
 - `application/json`
 - `application/graphql-response+json`
 - `multipart/mixed;deferSpec=20220824`
--- a/docs/source/errors.mdx
+++ b/docs/source/errors.mdx
@@ -11,14 +11,15 @@ Learn about error codes and HTTP response status codes returned by the Apollo Ro
 <PropertyList kind="errCodes">
 <Property name="400" short="Bad request">
 
-A request failed GraphQL validation, or a request failed to be parsed.
+The request failed GraphQL validation or failed to be parsed.
 
 </Property>
 <Property name="401" short="Unauthorized">
 
-For a client making a request that requires authentication, a JWT failed to be verified.
+Requests may receive this response in two cases:
 
-For a non-client subscription endpoint calling a subscription callback URL, the router couldn't find a matching subscription identifier between its registered subscriptions and a subscription event.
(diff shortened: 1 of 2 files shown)


docs/source/errors.mdx
```diff
@@ -0,0 +1,46 @@
...
+
+</Property>
+<Property name="429" short="Too many requests">
+
+Exceeded configured rate limits. See [client side traffic shaping](/configuration/traffic-shaping/#client-side-traffic-shaping).
+
+</Property>
+<Property name="500" short="Internal server error">
+
+An unexpected issue. Report this [possible bug](https://github.com/apollographql/router/issues/new?assignees=&labels=raised+by+user&projects=&template=bug_report.md&title=) to the router team.
```
abernix, line 43: It is also possible that the user has custom configured this HTTP status by throwing a particular Rhai error — I can't find this in the docs explicitly, but

- https://github.com/apollographql/router/pull/2097 introduced it.
- [This page mentions it](https://www.apollographql.com/docs/router/customizations/rhai/#:~:text=Headers-,Status%20codes,-Request%20context)

docs/source/errors.mdx
```diff
@@ -0,0 +1,46 @@
...
+
+<PropertyList kind="errCodes">
+<Property name="400" short="Bad request">
+
+A problem when validating GraphQL or parsing an HTTP request.
+
+</Property>
+<Property name="401" short="Unauthorized">
+
+Sent an event to a non-existent subscription or used an invalid hash. Returned by either the authorization plugin to clients or the subscription endpoint.
```
abernix, line 19: @Geal I think I need more crispness on what you had meant when explaining this.  My **nuanced** take on what you had wrote was that this applied to ***subscription callback URL*** — which is _distinctly_ different from what Router clients would send to the Router.  In other words, two separate things:

1. When a client (e.g., Android, iOS) is making a request to the Route, if there is not a satisfactory answer from the JWT negotiation, then a 401 will be returned.
2. When using the subscription callback URL (i.e., when calling it from a subgraph implementation or pub/sub implementation; **not** a client-sourced request) if a:
   - Non-existent subscription identifier is used.
   - (I'm actually not sure what the difference is between an invalid hash and non-existent subscription — either way it's non-existent, I think?)

abernix:
Also, worth noting that CircleCI is currently in a brown-out for all of today as one of our base images which we still use is being deprecated in Q4.  That's not on you — we're on it — but just FYI.

docs/source/errors.mdx
Geal: on 1. if there is an issue when verifying the JWT, or if no JWT is present and `require_authentication` is set to true, then a 401 is returned
on 2. yes, it is about the subscription callback URL. It can give the subgraph a key that is used to hash the subscription event. If the hash does not match the event, the router will return a 401 to the callback request. And there's another case in heartbeats where the router wil return a 401 if the id is not in the expected list, but I am not clear on that, maybe @bnjjj could give more context (or I'll dig in the code)

docs/source/errors.mdx
```diff
@@ -0,0 +1,60 @@
+---
+title: Errors
+subtitle: Error and status codes returned by the Apollo Router
+description: Error codes and response status codes returned by the Apollo Router
```
Meschreiber, line 4: Might as well include HTTP here for SEO:

```suggestion
description: Error codes and HTTP response status codes returned by the Apollo Router
```

docs/source/errors.mdx
```diff
@@ -0,0 +1,60 @@
...
+---
+
+Learn about error codes and HTTP response status codes returned by the Apollo Router.
+
+## Status codes
+
+<PropertyList kind="errCodes">
+<Property name="400" short="Bad request">
+
+A request failed GraphQL validation, or a request failed to be parsed.
```
Meschreiber, line 14: Definite articles seem more natural to me? Will just make the suggestion here rather than globally.

```suggestion
The request failed GraphQL validation or failed to be parsed.
```

docs/source/errors.mdx
```diff
@@ -0,0 +1,60 @@
...
+
+</Property>
+<Property name="401" short="Unauthorized">
+
+For a client making a request that requires authentication, a JWT failed to be verified.
+
+For a non-client subscription endpoint calling a subscription callback URL, the router couldn't find a matching subscription identifier between its registered subscriptions and a subscription event.
```
Meschreiber, line 21:
```suggestion
Requests may receive this response in two cases:

- For requests that requires authentication, the client's JWT failed verification.
- For a non-client subscription endpoint calling a subscription callback URL, the router couldn't find a matching subscription identifier between its registered subscriptions and a subscription event.
```

docs/source/errors.mdx
```diff
@@ -0,0 +1,60 @@
...
+<Note> 
+
+Both mutations and subscriptions must use POST.
+
+</Note>
+
+</Property>
+<Property name="406" short="Not acceptable">
+
+A query's HTTP `Accept` header didn't contain any of the router's supported mime-types:
```
Meschreiber, line 37: Is the switch from "request" --> "query" intentional here?

docs/source/errors.mdx
```diff
@@ -0,0 +1,60 @@
...
+
+</Property>
+<Property name="429" short="Too many requests">
+
+Request traffic exceeded configured rate limits. See [client side traffic shaping](/configuration/traffic-shaping/#client-side-traffic-shaping).
+
+</Property>
+<Property name="500" short="Internal server error">
+
+The router encountered an unexpected issue. Report this [possible bug](https://github.com/apollographql/router/issues/new?assignees=&labels=raised+by+user&projects=&template=bug_report.md&title=) to the router team.
```
Meschreiber, line 51: The link seems to fit better with the text "Report":

```suggestion
The router encountered an unexpected issue. [Report](https://github.com/apollographql/router/issues/new?assignees=&labels=raised+by+user&projects=&template=bug_report.md&title=) this possible bug to the router team.
```

Meschreiber (approved):
Some wording suggestions on my end, but otherwise LGTM! (Assuming the router team's accuracy review ofc 🙇‍♀️.)

Geal (approved):

PR #4735, opened 2024-03-04
update our base linux image to avoid CI brown outs

garypen:
Update our linux images
from: `2004:2022.04.1`
to: `2004:2024.01.1`

Still `ubuntu 20.04` based, but with more recent patches and not subject to brown outs.


garypen:
No changeset. This is just fixing up our CI pipeline.

BrynCooke (approved):

PR #4748, opened 2024-03-05
Type conditioned fetching

o0Ignition0o:
### Type conditioned fetching ([PR #4748](https://github.com/apollographql/router/pull/4748))

Type conditioned fetching

When querying a field that is in a path of 2 or more unions, the query planner was not able to handle different selections and would aggressively collapse selections in fetches yielding an incorrect plan.

This change introduces an experimental configuration option to enable type conditioned fetching:

```yaml
experimental_type_conditioned_fetching: true # false by default
```

By [@o0Ignition0o](https://github.com/o0Ignition0o) in https://github.com/apollographql/router/pull/4748


apollo-router/tests/type_conditions_test.rs
```diff
@@ -0,0 +1,182 @@
...
+        id
+      }
+      ... on ArticleResult {
+        id
+        sections {
+          ... on GallerySection {
+            artwork
+          }
+          ... on EntityCollectionSection {
+            artwork
```
tinnou, line 177: to make it more obvious you could pass a value that is specific to the type condition and that value back from the subgraph.
```graphql
artwork(params: "articleResult")
```
and expect the value to be returned as `"blah-articleResult"`

o0Ignition0o: Great idea, thanks!

tinnou (approved):

Geal (approved):

apollo-router/src/configuration/mod.rs
```diff
@@ -191,6 +191,11 @@ pub struct Configuration {
     /// Batching configuration.
     #[serde(default)]
     pub(crate) batching: Batching,
+
+    /// Type conditioned fetching configuration.
+    /// If you don't know what this is about, you probably don't need it.
```
garypen, line 196: I feel like there could be a better comment for this. :)

apollo-router/src/json_ext.rs
```diff
@@ -576,6 +713,7 @@ fn iterate_path_mut<'a, F>(
 
 /// A GraphQL path element that is composes of strings or numbers.
 /// e.g `/book/3/name`
+/// TODO: impl display for PathElement
```
garypen, line 716: Still required?

apollo-router/src/json_ext.rs
```diff
@@ -21,6 +24,11 @@ pub(crate) type Object = Map<ByteString, Value>;
 
 const FRAGMENT_PREFIX: &str = "... on ";
 
+static TYPE_CONDITIONS_REGEX: Lazy<Regex> = Lazy::new(|| {
+    Regex::new(r"(?:\|\[)(.+?)(?:,\s*|)(?:\])")
```
garypen, line 28: NIT: Maybe easier to understand when captures are used if these are named.

apollo-router/src/json_ext.rs
```diff
@@ -765,14 +1036,34 @@ impl Path {
 
     pub fn last_key(&mut self) -> Option<String> {
         self.0.last().and_then(|elem| match elem {
-            PathElement::Key(k) => Some(k.clone()),
+            // Todo: impl tostring
```
garypen, line 1039: Still required?

apollo-router/src/configuration/mod.rs
o0Ignition0o: XD fair, removing the second line

apollo-router/src/json_ext.rs
o0Ignition0o: Not anymore!

apollo-router/src/json_ext.rs
o0Ignition0o: nice catch, not required anymore! fixed in d14d077d0

apollo-router/src/json_ext.rs
o0Ignition0o: fixed in d14d077d0

apollo-router/src/json_ext.rs
o0Ignition0o: fixed in d14d077d0

apollo-router/src/configuration/mod.rs
o0Ignition0o: fixed in d14d077d0

apollo-router/src/query_planner/bridge_query_planner.rs
```diff
@@ -1477,6 +1480,7 @@ mod tests {
             .await
     }
 
+    #[ignore]
```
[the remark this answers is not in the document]
o0Ignition0o: oh leftover of when i was tracking a branch, fixed in d14d077d0

Cargo.lock
```diff
@@ -87,9 +87,9 @@ dependencies = [
 
 [[package]]
 name = "ahash"
-version = "0.8.11"
+version = "0.8.6"
```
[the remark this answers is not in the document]
o0Ignition0o: Fixed in ea54bf912

apollo-router/src/json_ext.rs
garypen: I'm not sure this is really fixed. I can see that you've named one of the groups "<condition>", but I can't see where it's being used. It looks like you are still using the indices of the various captures.

e.g.:
```
        let path = TYPE_CONDITIONS_REGEX.replace(s, |caps: &Captures| {
            type_conditions.extend(
                caps.extract::<1>()
                    .1
                    .map(|s| s.split(',').map(|s| s.to_string()))
                    .into_iter()
                    .flatten(),
            );
            ""
        });
```
You are always using `extract, group 1`, which is a bit opaque. If you want to use the name then it would be something like: `caps.name("condition")` and you'd need to do some handling for not being present.

Anyway, it was a NIT, so I'm fine with the `extract` approach that you are using, in which case no need to name the capture group.

If you do think it's worth naming the various capture groups, then you should use them to make the code a bit less opaque when you are pulling out the conditions as I outlined above.

o0Ignition0o: actually fixed in 6c67cacfe 😁

garypen: Now you are making me feel really picky, but there multiple used of `extract()`. May as well change all of them.

garypen (approved):

PR #4753, opened 2024-03-05
Validate enum values in input variables

### LINKED ISSUE #4633: Different Router vs JS Gateway error response for non-existent enum input reference
**Describe the bug**
If an operation is sent to the Javascript Gateway that references an invalid enum value for an input object, it returns a `400 Bad Request`, with a `GRAPHQL_VALIDATION_FAILED` error. The router returns a `200: OK` with a `BAD_USER_INPUT` error instead.

**Expected behavior**
I believe the Gateway had this validation correct, returning a `400: Bad Request`.

**Output**
Given an operation such as:

```
{
  "query": "mutation SetCampaignDisposition($var_SetCampaignDispositionInput: SetCampaignDispositionInput!) {\n  setCampaignDisposition(input: $var_SetCampaignDispositionInput) {\n    campaign {\n      id\n      description\n      priority\n      disposition\n    }\n  }\n}\n",
  "variables": {
    "var_SetCampaignDispositionInput": {
      "id": "MOBILE2_MIGRATION_MESSAGE",
      "disposition": "badDisposition"
    }
  }
}
```

JS Gateway `400: Bad Request`:

```
{
  "errors": [
    {
      "message": "Unknown type \"SetCampaignDispositionInput\".",
      "locations": [
        {
          "line": 1,
          "column": 40
        }
      ],
      "extensions": {
        "code": "GRAPHQL_VALIDATION_FAILED",
        "stacktrace": [ ... ]
      }
    },
    {
      "message": "Cannot query field \"setCampaignDisposition\" on type \"Mutation\".",
      "locations": [
        {
          "line": 1,
          "column": 70
        }
      ],
      "extensions": {
        "code": "GRAPHQL_VALIDATION_FAILED",
        "stacktrace": [ ... ]
      }
    }
  ]
}
```

Router `200: Ok`:

```
{
  "data": null,
  "errors": [
    {
      "message": "Variable \"$input\" got invalid value \"NOT_AN_ENUM_VALUE\" at \"input.disposition\"; Value \"NOT_AN_ENUM_VALUE\" does not exist in \"CampaignDisposition\" enum.",
      "locations": [
        {
          "line": 1,
          "column": 62
        }
      ],
      "extensions": {
        "code": "BAD_USER_INPUT",
        "serviceName": "graphql-campaign-service",
        "stacktrace": [ ... ]
      }
    }
  ]
}
```

Geal:
Fix #4633

The Router will now validate enum values provided in JSON variables. The error message fomr input variables validation is changed from `VALIDATION_INVALID_TYPE_VARIABLE` to `GRAPHQL_VALIDATION_FAILED`.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


.changesets/fix_geal_input_enum_validation.md
```diff
@@ -0,0 +1,5 @@
+### Validate enum values in input variables ([PR #4753](https://github.com/apollographql/router/pull/4753))
```
Geal, line 1:
```suggestion
### Validate enum values in input variables ([Issue #4633](https://github.com/apollographql/router/issues/4633))
```

apollo-router/src/error.rs
```diff
@@ -160,7 +160,7 @@ impl FetchError {
 impl ErrorExtension for FetchError {
     fn extension_code(&self) -> String {
         match self {
-            FetchError::ValidationInvalidTypeVariable { .. } => "VALIDATION_INVALID_TYPE_VARIABLE",
+            FetchError::ValidationInvalidTypeVariable { .. } => "GRAPHQL_VALIDATION_FAILED",
```
goto-bus-stop, line 163: These should be separate error codes. `GRAPHQL_VALIDATION_FAILED` means that the operation was invalid, which doesn't account for the actual values provided to variables (spec-wise that is part of execution and not validation). In apollo-server, `BAD_USER_INPUT` is used when variable input coercion fails, including wrong types or non-existent enum values, but maybe we should continue to use our value for back compat?

goto-bus-stop (changes requested):

apollo-router/src/error.rs
Geal: apparently the gateway returns `GRAPHQL_VALIDATION_FAILED` here https://github.com/apollographql/router/issues/4633

goto-bus-stop: The gateway raises an entirely different error there (an input type doesn't exist in the schema), it never gets to input validation

goto-bus-stop: Oh, it might be that #4633 should indeed raise a validation error, but because with the JS query planner the router ends up doing input coercion *before* graphql validation, this error is the one that gets returned. So #4551 might solve the difference.

Geal: so which error should the router return exactly?

goto-bus-stop: I was wrong before about #4663, as you said the error on router comes from the subgraph, so there is also something else going on. I still suspect the reproduction may be wrong, or there is something very weird as both operations are validated with JS validation and should thus have the same result.

For this PR, I think we should continue to use the `VALIDATION_INVALID_TYPE_VARIABLE` error code. We could also rename to `BAD_USER_INPUT` to match gateway but people could be relying on our error code already.

apollo-router/src/services/supergraph/tests.rs
```diff
@@ -3549,3 +3549,84 @@ async fn abstract_types_in_requires() {
...
+#[tokio::test]
+async fn invalid_input_enum() {
+    /*let subgraphs = MockedSubgraphs([
+        ("user", MockSubgraph::builder().with_json(
+                serde_json::json!{{"query":"{currentUser{activeOrganization{__typename id}}}"}},
+                serde_json::json!{{"data": {"currentUser": { "activeOrganization": null }}}}
+            ).build()),
+        ("orga", MockSubgraph::default())
+    ].into_iter().collect());*/
+
```
Geal, line 3589:
```suggestion
```

apollo-router/src/error.rs
Geal: agreed, I changed it back

PR #4754, opened 2024-03-05
Add integration tests for nested files

### LINKED ISSUE #4709: File uploads - Nesting integration test
The original file upload spec supports having the uploaded file be nested in either an object or a list. The current tests only test for root-level variables. An example of the kind of request we would want to test is as follows:

```graphql
mutation Upload {
  uploadList($files: [Upload!]! { ... }
  uploadObject($nestedFile: NestedFile) { ... }
}

type NestedFile {
  file: Upload
}
```

nicholascioli:
Some tests were missing to ensure that files nested within other GraphQL types still resolved correctly, so this commit addresses that.

This change has been rebased off of #4726 since some tests fail when the output gets blocked.

Fixes #4709 

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [x] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


IvanGoncharov (approved):
Looks good, except commit list containing unrelated commits 👍

nicholascioli:
I've updated the commit list to just include my changes, but tests might fail now due to the blocking issue. Let me know if it looks good!

PR #4755, opened 2024-03-05
Add integration test for non-nullable files

### LINKED ISSUE #4708: File uploads - Nullability integration test
All of the current tests use a nullable Upload object, but most use-cases will probably make it non-nullable.

nicholascioli:
Existing integration tests for the file uploads plugin did not test for non-nullable uploads. As this is realistically the more common case, this commit adds a test to ensure that non-nullable files work as expected.

This is necessary since the router does some substitution of the variables for the file upload case which might cause validation issues if not carefully tested.

As is the case in #4754, this commit is rebased off of #4726 as certain tests fail when the output blocks.

Fixes #4708 

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [X] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/tests/fixtures/file_upload/schema.graphql
+++ b/apollo-router/tests/fixtures/file_upload/schema.graphql
@@ -82,6 +82,7 @@ type Mutation
   @join__type(graph: UPLOADS_CLONE)
 {
   singleUpload(file: Upload): File @join__field(graph: UPLOADS)
+  singleUploadNonNull(file: Upload!): File! @join__field(graph: UPLOADS)
   singleUploadClone(file: UploadClone): FileClone @join__field(graph: UPLOADS_CLONE)
   multiUpload(files: [Upload!]!): [File!]! @join__field(graph: UPLOADS)
   nestedUpload(nested: NestedUpload): File @join__field(graph: UPLOADS)
--- a/apollo-router/tests/integration/file_upload.rs
+++ b/apollo-router/tests/integration/file_upload.rs
@@ -374,6 +374,59 @@ async fn it_supports_compression() -> Result<(), BoxError> {
 }
 
 #[tokio::test(flavor = "multi_thread")]
+async fn it_supports_non_nullable_file() -> Result<(), BoxError> {
+    use reqwest::multipart::Form;
+    use reqwest::multipart::Part;
+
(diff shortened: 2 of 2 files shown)


Cargo.lock
```diff
@@ -4083,9 +4083,9 @@ dependencies = [
 
 [[package]]
 name = "mio"
-version = "0.8.9"
+version = "0.8.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3dce281c5e46beae905d4de1870d8b1509a9142b62eedf18b443b011ca8343d0"
+checksum = "a4a650543ca06a924e8b371db273b2756685faae30f8487da1b56505a8f78b0c"
```
IvanGoncharov, line 4088: Is this package used inside tests?
Or it is accidential?

apollo-router/src/plugins/telemetry/fmt_layer.rs
```diff
@@ -167,28 +167,58 @@ where
...
-            };
-            if self.fmt_event.format_event(&ctx, &mut buf, event).is_ok() {
-                let mut writer = self.make_writer.make_writer();
-                if let Err(err) = std::io::Write::write_all(&mut writer, buf.as_bytes()) {
-                    if err.kind() != std::io::ErrorKind::WouldBlock {
-                        let _ = std::io::stderr().write_all(
-                            format!("cannot flush the logging buffer, this is a bug: {err:?}")
-                                .as_bytes(),
-                        );
+            let mut buf = &mut *(buf.borrow_mut());
```
IvanGoncharov, line 170: Diff sims to contain unrelated change.
Can you fix this PR?

IvanGoncharov (approved):
Looks good, except commit list containing unrelated commits 👍

Cargo.lock
nicholascioli: Not sure how this got pulled in. I've rebased and removed this change!

apollo-router/src/plugins/telemetry/fmt_layer.rs
nicholascioli: Again, not sure how this got here. Rebased!

Issue #4611, opened 2024-02-07
Missing http.status.response_code when there is a broken pipe error


danpayne17:
**Describe the bug**
If the caller closes the connection before the router is done sending the response (e.g. broken pipe), the router span never gets tagged with `http.status.response_code`. Consequently, these are the only requests that do not show any status code in DataDog.

**To Reproduce**
1. Insert code into a subgraph resolver so that it waits 10 seconds before responding to requests from router
2. Send request to router using curl
3. Cancel request (ctrl-c) while subgraph is waiting to respond which induces a broken pipe error
4. Observe in DataDog that the trace for the above request has no status code

**Expected behavior**
Every trace should have a response status code so that DataDog queries can use the status code as a predicate

**Router:**
 - OS: Amazon Linux 2023
 - Version: 1.37.0

BrynCooke:
Just wondering if status code 499 would be appropriate here? It's non-standard, but widely used because of nginx.

danpayne17:
Yes, I think 499 status code would be a good fit and would certainly work for us

Issue #4590, opened 2024-02-02
Rhai supergraph response callback does not trigger when a broken pipe is encountered


danpayne17:
**Describe the bug**
If the caller closes the connection before the router is done sending the response (e.g. broken pipe), the response callback of the supergraph lifecycle event is never triggered. This is an issue because in this callback, we use Rhai to do a number of things:
1. set response headers (this obviously is irrelevant because the pipe is broken so the response isn't going to be sent so we don't need to set these headers)
2. rewrite the errors array in response body, if present (also irrelevant b/c the response is not going to be sent to caller)
3. set response context variables for use in telemetry instrumentation by `response_context` selectors (these need to be set in order for span tag attributes to be applied as expected; may be related to 
![image](https://github.com/apollographql/router/assets/871193/e17ef112-e2c6-4c8b-9faa-8eed8a90cb64))
4. log the status of the response (e.g. call to n milliseconds to process for this operation name and resulted in these errors) (this is really important as this log statement actually drives Splunk dashboards and this log statement is never made b/c this lifecycle event isn't triggered)
5. log the response body (not super critical but would be nice to still log the response body that would have been sent if the pipe weren't broken)

**To Reproduce**
1. Configure router to execute Rhai scripts during supergraph response callback:
```
fn supergraph_service(service) {
  const response_callback = Fn("process_supergraph_response");
  service.map_response(response_callback);
}

fn process_supergraph_response(response) {
  print('executing fn process_supergraph_response()');
  tracing_handler::set_supergraph_response_headers(response);
  error_handler::rewrite_extensions_in_supergraph_response(response);
  error_handler::set_response_context_selector_values(response);
  logger::log_response_status(response);
  logger::log_response_body(response);
}
```
2. Insert code into subgraph resolver so that it waits 10seconds before responding to requests from router
3. Send request to router using curl
4. Cancel request (ctrl-c) while subgraph is waiting to respond which induces a broken pipe error
5. Observe that the `process_supergraph_response` was never called

**Expected behavior**
The response callback of the supergraph service is still executed when a broken pipe is encountered, just like other non-200 errors the router encounters (e.g. a 400 BAD REQUEST still executes the callback)

**Router:**
 - OS: Amazon Linux 2023
 - Version: 1.37.0

**Additional context**
None

Issue #4589, opened 2024-02-02
response_context selector fails when request results in broken pipe


danpayne17:
**Describe the bug**
If the caller closes the connection before the router has a chance to respond (e.g. broken pipe), the router fails to apply any span tag attributes using the `response_context` selector

**To Reproduce**
1. Configure router with response_context selector:
```
telemetry:
  instrumentation:
    spans:
      mode: spec_compliant
      router:
        attributes:
          graph-operation-type:
            response_context: operation_kind
```
2. Configure subgraph to wait 10s before responding to calls 
3. Send request to router using curl
4. Cancel the request (ctrl-c) while the subgraph is waiting to respond
5. Observe that the graph-operation-type was not applied to the span tag

**Expected behavior**
`response_context` span tag attributes are still applied to the `router` span just as with other non-200 errors (e.g. if there is a 400 BAD REQUEST error, we still see all tag attributes from `response_context` selectors on the router span) even when we have a broken pipe

**Router:**
 - OS: Amazon Linux 2023
 - Version: 1.37.0

**Additional context**
We are ingesting traces into DataDog

Issue #4576, opened 2024-01-31
Missing error.message tag on router span for all traces where there is a broken pipe error


danpayne17:
**Describe the bug**
When the caller closes the connection to the router before it is done sending the response, the trace that is sent to DataDog is missing the `error.message` tag on the router span. This is causing DataDog to be unable to display the cause of the error

**To Reproduce**
1. Connect the router to DataDog
2. Launch router with a single subgraph and in the subgraph, have it wait 10 seconds when it receives the request from the router
3. Use curl to send a request to the router
4. While the request is waiting at the subgraph, cancel the curl request (ctrl-c)
5. Observe the error in DataDog and how there is no error message. Instead it displays "Missing error message and stack trace"

<img width="316" alt="image" src="https://github.com/apollographql/router/assets/871193/26d65379-1844-4c30-b1ca-2130a9ec3340">

**Expected behavior**
An error message indicating the error encountered (e.g. 'Broken pipe') should be displayed in DataDog and seen in the router trace span at `error.message`

**Router:**
 - OS: Amazon Linux 2023
 - Router Version: 1.37.0

**Additional context**
Other errors (e.g. 400) seem to populate error.message just fine. It's only when there's a broken pipe that it doesn't tag set.

Errors view in DataDog:
<img width="406" alt="image" src="https://github.com/apollographql/router/assets/871193/39f19809-002d-4c11-91d2-5b82fc10d494">

Span tags (error key excerpted):
<img width="529" alt="image" src="https://github.com/apollographql/router/assets/871193/f6887c7c-eb7c-4ae7-9301-27e711e475e1">

PR #4770, opened 2024-03-08
Execute the entire request pipeline if the client closed the connection

Geal:
Fix https://github.com/apollographql/router/issues/4569
Fix https://github.com/apollographql/router/issues/4576
Fix https://github.com/apollographql/router/issues/4589
Fix https://github.com/apollographql/router/issues/4590
Fix https://github.com/apollographql/router/issues/4611

When the client closes the connection prematurely, it drops the entire request handling task, which means that it won't go through the entire response pipeline, where we record the operation and handle telemetry. Some users also have additional steps with rhai or coprocessors where they add metadata, and those steps should run even on canceled requests. This moves the request handling to a separate task to make sure it runs, but it also skips subgraph requests if we detected that the client closed the connection, to prevent unneeded traffic.

*Description here*

Fixes #**issue_number**

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/services/router/service.rs
```diff
@@ -265,6 +266,16 @@ impl RouterService {
...
         process_vary_header(&mut parts.headers);
 
+        if context
+            .extensions()
+            .lock()
+            .get::<CanceledRequest>()
+            .is_some()
+        {
+            parts.status = StatusCode::from_u16(499)
+                .expect("499 is not a standard status code but common enough");
```
o0Ignition0o, line 276: shall we document this?

o0Ignition0o (approved):
LGTM

apollo-router/src/services/router/service.rs
Geal: good point: ce890f8b8

PR #4771, opened 2024-03-08
Add support of instruments in configuration for telemetry

### LINKED ISSUE #4319: Custom instruments defined by yaml
Implement custom instruments defined by yaml as documented in:
https://www.apollographql.com/docs/router/configuration/telemetry/instrumentation/instruments

This will allow users to avoid reaching for a custom plugin when they need to collect metrics for when things happen/

bnjjj:
Add support for custom and standard instruments through the configuration file. You'll be able to add your own custom metrics just using the configuration file. It can be conditional, get values from selectors like headers, context, body. And the metrics can have different types like `histogram` or `counter`.

Example:

```yaml title="router.yaml"
telemetry:
  instrumentation:
    instruments:
      router:
        http.server.active_requests: true
        acme.request.duration:
          value: duration
          type: counter
          unit: kb
          description: "my description"
          attributes:
            http.response.status_code: true
            "my_attribute":
              response_header: "x-my-header"
  
      supergraph:
        acme.graphql.requests:
          value: unit
          type: counter
          unit: count
          description: "supergraph requests"
          
      subgraph:
        acme.graphql.subgraph.errors:
          value: unit
          type: counter
          unit: count
          description: "my description"
```

[Documentation](https://www.apollographql.com/docs/router/configuration/telemetry/instrumentation/instruments)

Fixes #4319 

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/plugins/telemetry/config_new/mod.rs
```diff
@@ -27,9 +25,9 @@ pub(crate) mod spans;
 pub(crate) trait Selectors {
     type Request;
     type Response;
-    fn on_request(&self, request: &Self::Request) -> LinkedList<KeyValue>;
-    fn on_response(&self, response: &Self::Response) -> LinkedList<KeyValue>;
-    fn on_error(&self, error: &BoxError) -> LinkedList<KeyValue>;
+    fn on_request(&self, request: &Self::Request) -> Vec<KeyValue>;
```
Geal, line 28: I suspect that these could return iterators (impl traits in traits should work now) instead of allocating lots of vectors, since they're almost always consumed as iterators right away, but I'm not sure there will be a huge benefit in doing that

bnjjj: I tried using Iterator but at the end it won't help because we often have to manipulate vec to mutate and insert attributes in that vec and then convert it again to an iterator

docs/source/configuration/telemetry/instrumentation/conditions.mdx
```diff
@@ -6,7 +6,7 @@ description: Set conditions for when events or instruments are triggered in the
 
 <PremiumFeature linkWithAnchor="https://www.apollographql.com/pricing#observability"/>
 
-You can set conditions for when an [instrument](./instruments) should be mutated or an [event](./events) should be triggered.
+You can set conditions for when an [instrument](./instruments) should be mutated <!--or an [event](./events) -->should be triggered.
```
IvanGoncharov, line 9: You added HTML comments. Is it intentional?

BrynCooke:
One thing that needs to be added are the config metrics in configuration/metrics.rs. It can use the same pattern as we did for spans.

BrynCooke (changes requested):
In general this is looking good.

I think that the code very unit testable so would like to see the individual functions tested in each module.

docs/source/configuration/telemetry/instrumentation/conditions.mdx
bnjjj: Yes, that's how it's done in other document so I just reproduced the same :p and it works

Geal (approved):

BrynCooke:
I had a poke around and didn't manage to spot the issue in the code.
Here's my config:
```
telemetry:
  exporters:
    metrics:
      prometheus:
        enabled: true
  instrumentation:
    instruments:
      router:
        http.server.request.duration:
          attributes:
            network.protocol.version: true
```
I was expecting the reccomended attributes to appear on the standard instrument but there are none:
`http_server_request_duration_bucket{otel_scope_name="apollo/router",le="0.001"} 0`

BrynCooke (approved):

garypen (approved):

PR #4775, opened 2024-03-08
use Object::with_capacity with selection set len in format_response

xuorig:
This is far from a great improvement, I think this procedure probably needs to be changed to make a bigger dent. Maybe re-ordering and modifying in place, which would require much more effort. This change does seem to make a difference comparing flamegraphs, but it's small.


Geal (approved):
I think that will help for queries with a large number of fields

bnjjj (approved):

PR #4779, opened 2024-03-11
upgrade rust version to 1.76.0

bnjjj:


garypen (commented):
Does the scaffold template need to be updated as well?

apollo-router-scaffold/templates/base/Dockerfile

bnjjj:
@garypen good catch ! And it was already broken I think so it's even better

Geal (approved):

PR #4780, opened 2024-03-11
Build upon the recent configuration changes to enable subgraph filtering

garypen:
Now that we have Batching subgraph configuration we can use it to control filtering by batching enabled subgraphs.
    
also:
 - Add a migration from `experimental_batching` -> `batching`
 - Add subgraph filtering utility function to batching configuration
 - Add some subgraph filtering unit tests
 - merge dev

Note: @nicholascioli  I merged in `dev`, so for reviewing purposes, just look at [this commit](https://github.com/apollographql/router/commit/984d2c60d67c2fec9635ea5d50f3ddb335307e13).


apollo-router/src/services/subgraph_service.rs
```diff
@@ -737,40 +738,54 @@ async fn call_batched_http(
...
-                waiters_opt = Some(batching.get_waiters());
+    {
+        let mut extensions_guard = context.extensions().lock();
+        let batching_opt = extensions_guard.get::<Batching>();
+        if let Some(batching) = batching_opt {
+            if batching.batch_include(service_name) {
+                if let Some(batch_query) = extensions_guard.get_mut::<BatchQuery>() {
+                    if !batch_query.finished() {
+                        tracing::debug!(
+                        "in subgraph we have batch_query: {batch_query}, service: {service_name}"
```
nicholascioli, line 761: Should this line and the next be indented once more?

nicholascioli (approved):
Looks good to me!

There seems to be a weird formatting issue, but if the linting test didn't catch it then maybe it's OK? My guess is that the linter doesn't look at macros.

PR #4784, opened 2024-03-11
return a better error when using defer in a subscription

### LINKED ISSUE #4783: Using `@defer` in a Subscription returns `router-bridge` error vs. nicer user-facing erorr
**Describe the bug**

When using a subscription in the router and using a deferred fragment within, you'll get the following error:

```json
{
    "errors": [
        {
            "message": "value retrieval failed: router bridge error: the deno runtime raised an error: `request: couldn't receive response couldn't deserialize payload `3689970476626457495`: `deno: couldn't deserialize response : `Error(\"invalid neither null nor empty object: found Object {\\\"code\\\": String(\\\"QUERY_PLANNING_FAILED\\\"), \\\"exception\\\": Object {\\\"stacktrace\\\": Array [String(\\\"Error: @defer is not supported on subscriptions\\\")]}}\", line: 0, column: 0)``.`",
            "extensions": {
                "code": "INTERNAL_SERVER_ERROR"
            }
        }
    ]
}
```

We do have a nicer error in the logs: `an error happened in the worker runtime Error: @defer is not supported on subscriptions` but the client-facing error is noisy and can be improved. 

**To Reproduce**

Use defer in a subscription request, such as: 

```gql
subscription Subscription {
  realtimeThing {
    id
    ... @defer {
      name
    }
  }
}
```

**Expected behavior**

A nicer user-facing error.

bnjjj:
Fixes #4783

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/services/supergraph/service.rs
+++ b/apollo-router/src/services/supergraph/service.rs
@@ -229,7 +229,7 @@ async fn service_call(
                 return Ok(SupergraphResponse::infallible_builder()
                     .context(context)
                     .errors(vec![graphql::Error::builder()
-                        .message("@defer is not supported on subscriptions")
+                        .message("@defer is not supported within a subscription")
                         .extension_code("DEFER_NOT_SUPPORTED_ON_SUBSCRIPTION")
                         .build()])
                     .status_code(StatusCode::BAD_REQUEST)
--- a/apollo-router/src/services/supergraph/service.rs
+++ b/apollo-router/src/services/supergraph/service.rs
@@ -225,6 +225,16 @@ async fn service_call(
             let operation_name = body.operation_name.clone();
             let is_deferred = plan.is_deferred(operation_name.as_deref(), &variables);
             let is_subscription = plan.is_subscription(operation_name.as_deref());
+            if is_deferred && is_subscription {
+                return Ok(SupergraphResponse::infallible_builder()
+                    .context(context)
+                    .errors(vec![graphql::Error::builder()
+                        .message("@defer is not supported on subscriptions")
+                        .extension_code("DEFER_NOT_SUPPORTED_ON_SUBSCRIPTION")
+                        .build()])
+                    .status_code(StatusCode::BAD_REQUEST)
+                    .build());
+            }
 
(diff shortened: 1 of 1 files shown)


apollo-router/src/services/supergraph/service.rs
```diff
@@ -225,6 +225,16 @@ async fn service_call(
             let operation_name = body.operation_name.clone();
             let is_deferred = plan.is_deferred(operation_name.as_deref(), &variables);
             let is_subscription = plan.is_subscription(operation_name.as_deref());
+            if is_deferred && is_subscription {
+                return Ok(SupergraphResponse::infallible_builder()
+                    .context(context)
+                    .errors(vec![graphql::Error::builder()
+                        .message("@defer is not supported on subscriptions")
```
lleadbet, line 232:
```suggestion
                        .message("@defer is not supported within a subscription")
```

abernix (changes requested):
Needs tests, but otherwise LGTM.

goto-bus-stop (commented):
I think we should do this before calling into QP, and after Rust-based validation. Maybe we can do it as the first thing inside this `service_call` using the `ParsedDocument` on the context? Or even inside `CachingQueryPlanner`

goto-bus-stop:
Per discussion in Slack, the JS QP already does this check, but the error gets mangled on the way up from deno to the final response. A fix for that is here: https://github.com/apollographql/federation-rs/pull/457

abernix:
@bnjjj Did you want to bring this PR back to life? It seems useful intent?

bnjjj:
It's something that will be automatically fixed by the new QP so I think it's ok to just close it now and wait for the proper fix in the new QP

goto-bus-stop:
I don't remember exactly but i think this is already done because of https://github.com/apollographql/federation-rs/pull/457 .

abernix:
thanks folks!

PR #4802, opened 2024-03-14
feat(subscription): add configurable heartbeat for websocket protocol

### LINKED ISSUE #4621: Configure a subgraph websocket ping interval
The [graphql-ws](https://github.com/enisdenjo/graphql-ws) protocol supports sending a [`Ping` message](https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md#ping). This is useful for detecting failed connections, displaying latency metrics or other types of network probing.

If I have opened a passthrough mode subscription to my subgraph the connection is all good. But we're facing issues with longer intervals between messages. Apollo Router logs a TCP connection reset when no messages have been sent beyond our load balancer after a connection idle timeout.

By capturing traffic, we confirmed neither Apollo Router nor the subgraph actively sends pings/heartbeats on the WebSocket session. Both _can_ respond to heartbeats, but both services are just sitting passive and our LB eventually closes the idle connection.

Add an option to enable sending a `Ping` message to a subgraph from Router on a set interval. Having this off by default is ok, but we need an option to turn it on for our subgraphs

```yaml
subscription:
  enabled: true
  mode:
    passthrough:
      all: 
        path: /subscriptions
        protocol: graphql_ws
        ping_interval: 300 # Some option probably in seconds is fine
```        

To keep the connection open we can go and update our subgraph code to also send `Ping` message because this is bi-directional but we may have to do that across many subgraphs and unfortunately some libraries like DGS don't have support to configure a ping interval either

**Additional context**
This is not a concern if we use the [HTTP callback protocol](https://www.apollographql.com/docs/router/executing-operations/subscription-callback-protocol) instead

IvanGoncharov:
Add the ability to enable heartbeat for cases where the subgraph drops idle connections.
For example, https://netflix.github.io/dgs/

Example of configuration:

```yaml
subscription:
  mode:
    passthrough:
      all:
        path: /graphql
        heartbeat_interval: enable #Optional
 ```

Fixes #4621

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/.changesets/feat_subscription_websocket_heartbeat.md
+++ b/.changesets/feat_subscription_websocket_heartbeat.md
@@ -1,0 +1,17 @@
+### feat(subscription): add configurable heartbeat for websocket protocol ([Issue #4621](https://github.com/apollographql/router/issues/4621))
+
+Add the ability to enable heartbeat for cases where the subgraph drops idle connections.
+For example, https://netflix.github.io/dgs/
+
+Example of configuration:
+
+```yaml
+subscription:
+  mode:
+    passthrough:
+      all:
+        path: /graphql
+        heartbeat_interval: enable #Optional
+ ```
+
+By [@IvanGoncharov](https://github.com/IvanGoncharov) in https://github.com/apollographql/router/pull/4802
--- a/.changesets/feat_subscription_websocket_heartbeat.md
+++ b/.changesets/feat_subscription_websocket_heartbeat.md
@@ -1,9 +1,8 @@
-### feat(subscription): add configurable heartbeat for websocket protocol ([Issue #4621](https://github.com/apollographql/router/issues/4621))
+### Subscriptions: Add configurable "heartbeat" to subgraph WebSocket protocol ([Issue #4621](https://github.com/apollographql/router/issues/4621))
 
-Add the ability to enable heartbeat for cases where the subgraph drops idle connections.
-For example, https://netflix.github.io/dgs/
+To account for GraphQL Subscription WebSocket implementations (e.g., [DGS](https://netflix.github.io/dgs/)) which drop idle connections by design, the router adds the ability to configure a heartbeat to keep active connections alive.
 
-Example of configuration:
+An example of configuration:
 
 ```yaml
 subscription:
--- a/apollo-router/src/plugins/subscription.rs
+++ b/apollo-router/src/plugins/subscription.rs
@@ -171,9 +171,11 @@ pub(crate) struct CallbackMode {
 #[derive(Debug, Copy, Clone, PartialEq, Eq, Deserialize, Serialize, JsonSchema)]
 #[serde(rename_all = "snake_case", untagged)]
 pub(crate) enum HeartbeatInterval {
+    /// disable heartbeat
     Disabled(Disabled),
     /// enable with default interval of 5s
     Enabled(Enabled),
+    /// enable with custom interval, e.g. '100ms', '10s' or '1m'
     #[serde(with = "humantime_serde")]
     #[schemars(with = "String")]
     Duration(Duration),
--- a/apollo-router/src/plugins/subscription.rs
+++ b/apollo-router/src/plugins/subscription.rs
@@ -112,7 +112,7 @@ impl SubscriptionModeConfig {
             if callback_cfg.subgraphs.contains(service_name) || callback_cfg.subgraphs.is_empty() {
                 let callback_cfg = CallbackMode {
                     public_url: callback_cfg.public_url.clone(),
-                    heartbeat_interval: callback_cfg.heartbeat_interval.clone(),
+                    heartbeat_interval: callback_cfg.heartbeat_interval,
                     listen: callback_cfg.listen.clone(),
                     path: callback_cfg.path.clone(),
                     subgraphs: HashSet::new(), // We don't need it
@@ -168,7 +168,7 @@ pub(crate) struct CallbackMode {
     pub(crate) subgraphs: HashSet<String>,
 }
 
-#[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize, JsonSchema)]
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Deserialize, Serialize, JsonSchema)]
 #[serde(rename_all = "snake_case", untagged)]
 pub(crate) enum HeartbeatInterval {
     Disabled(Disabled),
@@ -186,22 +186,22 @@ impl HeartbeatInterval {
     pub(crate) fn new_disabled() -> Self {
         Self::Disabled(Disabled::Disabled)
     }
-    pub(crate) fn into_option(&self) -> Option<Duration> {
+    pub(crate) fn into_option(self) -> Option<Duration> {
         match self {
             Self::Disabled(_) => None,
             Self::Enabled(_) => Some(Duration::from_secs(5)),
-            Self::Duration(duration) => Some(*duration),
+            Self::Duration(duration) => Some(duration),
         }
     }
 }
 
-#[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize, JsonSchema)]
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Deserialize, Serialize, JsonSchema)]
 #[serde(rename_all = "snake_case")]
 pub(crate) enum Disabled {
     Disabled,
 }
 
-#[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize, JsonSchema)]
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Deserialize, Serialize, JsonSchema)]
 #[serde(rename_all = "snake_case")]
 pub(crate) enum Enabled {
     Enabled,
--- a/apollo-router/src/plugins/subscription.rs
+++ b/apollo-router/src/plugins/subscription.rs
@@ -43,6 +43,7 @@ use crate::Endpoint;
 use crate::ListenAddr;
 
 type HmacSha256 = Hmac<sha2::Sha256>;
+pub(crate) const DEFAULT_HEARTBEAT_INTERVAL: Duration = Duration::from_secs(5);
 pub(crate) const APOLLO_SUBSCRIPTION_PLUGIN: &str = "apollo.subscription";
 #[cfg(not(test))]
 pub(crate) const APOLLO_SUBSCRIPTION_PLUGIN_NAME: &str = "subscription";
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

**Comment by IvanGoncharov** [comment]
Missing unit tests 😞 
I made a few attempts to add them, but each time they ended up being flaky and unreliable.

File: apollo-router/src/protocols/websocket.rs [hunk 1]
```diff
@@ -414,11 +431,32 @@ where
...
+                }
+                Poll::Pending
+            }
         }
     }
 }
 
+fn send_heartbeat(mut stream: Pin<&mut impl Sink<ClientMessage>>, cx: &mut std::task::Context<'_>) {
+    if stream.as_mut().poll_flush(cx).map(|result| result.is_ok()) == Poll::Ready(true)
+        && stream.as_mut().poll_ready(cx).map(|result| result.is_ok()) == Poll::Ready(true)
```
IvanGoncharov, line 449: I like my original version with pattern matching: https://github.com/apollographql/router/pull/4802/commits/efcb2bfb10d242ab34cca5598e03df73eeb2d6b8#diff-b932e2ce79bcb30a9685d96419b458d13772deb85c850fe19741532cc21f0dbeL448

But Clippy forced me to rewrite it in this form.

bnjjj: I was a little bit confused by the name of the variable. I think you should name it `sink`

apollo-router/src/protocols/websocket.rs
```diff
@@ -414,11 +431,32 @@ where
...
+    {
+        // Ignore error
+        let _ = stream.start_send(ClientMessage::Ping {
+            payload: Some(serde_json_bytes::Value::String(
+                "APOLLO_ROUTER_HEARTBEAT".into(),
```
bnjjj, line 454: Should we provide a payload ? If we can avoid sending more data than necessary I would vote for it

docs/source/executing-operations/subscription-support.mdx
```diff
@@ -101,9 +101,11 @@ subscription:
         reviews: # Overrides settings for the 'reviews' subgraph
           path: /ws # Absolute path that overrides '/subscriptions' defined above
           protocol: graphql_ws # The WebSocket-based subprotocol to use for subscription communication (Default: graphql_ws)
+          heartbeat_interval: enable # Optional (Default: disabled)
```
bnjjj, line 104: Could you add document to explain it also support human readable duration like `10secs`

apollo-router/src/protocols/websocket.rs
IvanGoncharov: you right! will change

apollo-router/src/protocols/websocket.rs
IvanGoncharov: It makes debugging easier, without it. 
It's not clear that this is a heartbeat.
Then I looked in traffic I also see RFC6455 PING/PONG so payload makes this particular message to standout.
I think it would also be helpful for customer's engeeneer trying to look into WS traffic.
With default interval of 5s it doesn't add that much overhead.

But this is not a principal issue for me so can remove it if you still have concerns?

docs/source/executing-operations/subscription-support.mdx
IvanGoncharov: will do 👍

apollo-router/src/protocols/websocket.rs
bnjjj: No I think you're right, if it improves the experience let's keep it

apollo-router/src/protocols/websocket.rs
IvanGoncharov: Rewrote this part completely.

docs/source/executing-operations/subscription-support.mdx
IvanGoncharov: fixed

apollo-router/src/plugins/subscription.rs
```diff
@@ -168,25 +168,43 @@ pub(crate) struct CallbackMode {
...
-#[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize, JsonSchema)]
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Deserialize, Serialize, JsonSchema)]
 #[serde(rename_all = "snake_case", untagged)]
 pub(crate) enum HeartbeatInterval {
     Disabled(Disabled),
+    /// enable with default interval of 5s
+    Enabled(Enabled),
     #[serde(with = "humantime_serde")]
     #[schemars(with = "String")]
     Duration(Duration),
```
bnjjj, line 179: You have to add documentation here

apollo-router/src/plugins/subscription.rs
```diff
@@ -168,25 +168,43 @@ pub(crate) struct CallbackMode {
     pub(crate) subgraphs: HashSet<String>,
 }
 
-#[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize, JsonSchema)]
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Deserialize, Serialize, JsonSchema)]
 #[serde(rename_all = "snake_case", untagged)]
 pub(crate) enum HeartbeatInterval {
     Disabled(Disabled),
```
bnjjj, line 174: Add documentation

apollo-router/src/protocols/websocket.rs
```diff
@@ -361,7 +388,129 @@ where
...
+        Self {
+            inner_stream,
+            close_signal: Some(close_signal),
+        }
+    }
+}
+
+impl<S> Drop for SubscriptionStream<S> {
+    fn drop(&mut self) {
+        if let Some(close_signal) = self.close_signal.take() {
```
Geal, line 464: technically you don't even need to send with `close_signal`. If it is dropped, `close_sentinel` will resolve and return the error `TryRecvError::Closed`: https://docs.rs/tokio/latest/tokio/sync/oneshot/error/enum.TryRecvError.html
but it is good to be explicit

apollo-router/src/protocols/websocket.rs
```diff
@@ -361,7 +388,129 @@ where
...
+                let mut interval =
+                    tokio::time::interval_at(tokio::time::Instant::now() + duration, duration);
+                interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
+                let mut heartbeat_stream = IntervalStream::new(interval)
+                    .map(|_| {
+                        Ok(ClientMessage::Ping {
+                            payload: Some(serde_json_bytes::Value::String(
+                                "APOLLO_ROUTER_HEARTBEAT".into(),
+                            )),
+                        })
```
Geal, line 424: this is fine to leave as is for now, but as a follow up, we should actually check if we got a response to the heartbeat. Right now it makes sure that the websocket connection stays open, but it does not account for the possibility of the backend hanging

Geal (approved):
this sounds very reasonable, nicely done. I'd like to get a second review on this before merging, to make sure I did not miss something (I did not sleep much last night)

BrynCooke:
@IvanGoncharov Is there a test that exercises the new behaviour?

.changesets/feat_subscription_websocket_heartbeat.md
```diff
@@ -0,0 +1,17 @@
+### feat(subscription): add configurable heartbeat for websocket protocol ([Issue #4621](https://github.com/apollographql/router/issues/4621))
+
+Add the ability to enable heartbeat for cases where the subgraph drops idle connections.
+For example, https://netflix.github.io/dgs/
+
+Example of configuration:
```
abernix, line 6:
```suggestion
### Subscriptions: Add configurable "heartbeat" to subgraph WebSocket protocol ([Issue #4621](https://github.com/apollographql/router/issues/4621))

To account for GraphQL Subscription WebSocket implementations (e.g., [DGS](https://netflix.github.io/dgs/)) which drop idle connections by design, the router adds the ability to configure a heartbeat to keep active connections alive.

An example of configuration:
```

docs/source/executing-operations/subscription-support.mdx
```diff
@@ -101,9 +101,11 @@ subscription:
         reviews: # Overrides settings for the 'reviews' subgraph
           path: /ws # Absolute path that overrides '/subscriptions' defined above
           protocol: graphql_ws # The WebSocket-based subprotocol to use for subscription communication (Default: graphql_ws)
+          heartbeat_interval: 10s # Optional, with 'enable' equal to '5s' (Default: disabled)
```
abernix, line 104: This line is a bit confusing to me.   Could you help me understand it better so I can suggest an alternative?

.changesets/feat_subscription_websocket_heartbeat.md
```diff
@@ -0,0 +1,16 @@
...
+
+An example of configuration:
+
+```yaml
+subscription:
+  mode:
+    passthrough:
+      all:
+        path: /graphql
+        heartbeat_interval: enable #Optional
```
BrynCooke, line 13: Why are we introducing `enable`?
The default before was 5s.

abernix: I was under the impression that it was disabled by default before?

docs/source/executing-operations/subscription-support.mdx
IvanGoncharov: Ok, so it's disabled by default. 
If you specify human readable time it will be that time, e.g. `10s`
If you specify `enable` it will be `5s`.

.changesets/feat_subscription_websocket_heartbeat.md
IvanGoncharov: Yes, it's disabled by default.
Benjamin suggested to add `enable` since we already have 5s interval as default for callback protocol.
The idea was to allow customers to set reasonable defaults without trying to guess the numbers.

IvanGoncharov: Enable is equal to 5s, which is our current default for callback protocol.

apollo-router/src/protocols/websocket.rs
```diff
@@ -616,6 +745,26 @@ mod tests {
...
+                       serde_json::to_string(&ClientMessage::Ping { payload: Some(serde_json_bytes::Value::String("APOLLO_ROUTER_HEARTBEAT".into())) }).unwrap(),
+                   ));
+
+                   assert!(
+                       socket.next().now_or_never().is_none(),
+                       "It should be no pending messages"
+                   );
+                   tokio::time::resume();
+                }
+
```
IvanGoncharov, line 767: > @IvanGoncharov Is there a test that exercises the new behaviour? As in it should check the heartbeat from the subgraph perspective.

Yes, it checked here. 
1. we stop timer
1. We check that there are no messages in a socket
1. advance timer for duration (60s)
1. check that we got ping
1. check that there are no messages in a socket
1. resume timer

It's done using a saved timer, so Tokio automatically advances the timer, and we don't need to actually wait for the 60s, so the test is fine.
I chose the 60s, so if the mock timer works for some reason, the test will timeout.

o0Ignition0o (approved):

apollo-router/src/plugins/subscription.rs
IvanGoncharov: added

apollo-router/src/plugins/subscription.rs
IvanGoncharov: added

docs/source/executing-operations/subscription-support.mdx
IvanGoncharov: I rewrote entire comment, hope it's more clear now.

PR #4811, opened 2024-03-15
docs: clarify headers for coproc should be normalized

shorgi:
Clarify that HTTP headers for the coprocessor should be normalized 

Resolves [DOC-92](https://apollographql.atlassian.net/browse/DOC-92)

[DOC-92]: https://apollographql.atlassian.net/browse/DOC-92?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docs/source/customizations/coprocessor.mdx
```diff
@@ -825,6 +825,8 @@ When `stage` is `SupergraphResponse`, if present and `true` then there will be s
 
 An object mapping of all HTTP header names and values for the corresponding request or response.
 
+Headers should be handled like HTTP headers in general and normalized before your coprocessor operates on them.
```
Meschreiber, line 828: We're telling the reader to do this right? Perhaps clearer to use active case then. (Rather than making it ambiguous whether we/Apollo generally handles headers like HTTP headers)

```suggestion
Ensure headers are handled like HTTP headers in general. For example, normalize header case before your coprocessor operates on them. 
```

docs/source/customizations/coprocessor.mdx
```diff
@@ -825,6 +825,8 @@ When `stage` is `SupergraphResponse`, if present and `true` then there will be s
 
 An object mapping of all HTTP header names and values for the corresponding request or response.
 
+Headers should be handled like HTTP headers in general and normalized before your coprocessor operates on them. 
+
 If your coprocessor [returns a _different_ value](#responding-to-coprocessor-requests) for `headers`, the router replaces the existing headers with that value.
 ⚠️ `content-length` sent by Coprocessors are discarded by the router. (Wrong `content-length` values could lead to failure for HTTP requests to be sent)
```
Meschreiber, line 831: Might be worth cleaning this up a little too?

```suggestion
If your coprocessor [returns a _different_ value](#responding-to-coprocessor-requests) for `headers`, the router replaces the existing headers with that value.

> The router discards any `content-length` headers sent by coprocessors because incorrect `content-length` values can lead to HTTP request failures.
```

(Going with a blockquote rather than admonition because it might look a bit cluttered within a table cell.)

Meschreiber (approved):
Just some small suggestions for clarity.

PR #4814, opened 2024-03-15
doc: clarify the resulting protocol of a subgraph connection when configuring experimental_http2

shorgi:
Add table to clarify the resulting HTTP protocol of a subgraph connection when configuring `experimental_http2`.

Resolves [DOC-11](https://apollographql.atlassian.net/browse/DOC-11)

[DOC-11]: https://apollographql.atlassian.net/browse/DOC-11?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/source/configuration/traffic-shaping.mdx
+++ b/docs/source/configuration/traffic-shaping.mdx
@@ -185,6 +185,13 @@ Use the table below to look up the resulting protocol of a subgraph connection, 
 | `experimental_http2: disable` | HTTP/1.1           | HTTP/1.1 with TLS           |
 | `experimental_http2: enable`  | HTTP/1.1           | Either HTTP/1.1 or HTTP/2 with TLS, as determined by the TLS handshake with a subgraph |
 | `experimental_http2: http2only` | h2c              | HTTP/2 with TLS              |
+| `experimental_http2` not set | HTTP/1.1 | Either HTTP/1.1 or HTTP/2 with TLS, as determined by the TLS handshake with a subgraph |
+
+<Note>
+
+Configuring `experimental_http2: http2only` for a subgraph that doesn't support HTTP/2 results in a failed subgraph connection.
+
+</Note>
 
 ### Ordering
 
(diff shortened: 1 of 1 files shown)


docs/source/configuration/traffic-shaping.mdx
```diff
@@ -170,11 +172,19 @@ traffic_shaping:
 
 ### HTTP/2
 
-The router supports subgraph connections over
-- HTTP/2 with TLS. This is the default configuration.
+The router supports subgraph connections over:
+- HTTP/1.1
+- HTTP/1.1 with TLS
+- HTTP/2 with TLS
 - HTTP/2 Cleartext protocol (h2c). This uses HTTP/2 over plaintext connections.
```
Meschreiber, line 179: From what I've seen `H2C` is more common than `h2c`, particularly if we're going with the more formal HTTP capitalization (vs http).

Meschreiber (commented):
Tabular format ftw! Do we need to further emphasize that this just _happens_ w/o any configuration, or is that already in the doc?

docs/source/configuration/traffic-shaping.mdx
shorgi: For whatever reason, the authoritative [IETF RFC](https://datatracker.ietf.org/doc/html/rfc7540#section-3.1) uses `HTTP` and `h2c`.

shorgi:
> Do we need to further emphasize that this just _happens_ w/o any configuration, or is that already in the doc?

Ah, I need to add a row for not configured ...

docs/source/configuration/traffic-shaping.mdx
Meschreiber: Ah, interesting.

Geal (approved):

PR #4825, opened 2024-03-19
Cleanup batching to handle aborted requests

nicholascioli:
This PR cleans up the batching interface to allow for handling cancelled requests or errors during batch construction. 

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


nicholascioli:
The current state of this PR is just to test if we agree on how the batching itself should work outside of the rest of the router. More commits will be added to integrate these changes into the rest of the router later.

I've left TODOs in spots that I think warrant discussion and lengthy comments everywhere else.

garypen (commented):
I've tried to address as many of the TODO comments as I could.

It seems like the main complication is "how do we identify exactly which parts of the batch are dead" when we kill part of the batch. I believe the way to do this is to replace storing subgraph fetch counts with the collection of "FetchNode" operations (or some memo-isation of them) and then simply using the index to remove that collection of operations from the fetches we would perform in our batches.

I might spend some time investigating that today.

apollo-router/src/batching.rs
```diff
@@ -5,129 +5,247 @@
...
-#[derive(Clone, Debug, Default)]
+#[derive(Clone, Debug)]
 pub(crate) struct BatchQuery {
+    /// The index of this query relative to the entire batch
     index: usize,
-    // Shared Batch
-    shared: Arc<Mutex<Batch>>,
+
+    /// A channel sender for sending updates to the entire batch
+    sender: mpsc::Sender<BatchHandlerMessage>,
```
garypen, line 30: If my idea about dropping this once the `BatchQuery` is done talking to the entire batch is a good one, this will need to be an `Option`.

nicholascioli:
I've made it compile again. It also still passes the `assemble_batch` unit test, which is nice.

I still need to address your comments and fix some hacks, but you should be able to run the router now using the new batching messaging.

PR #4828, opened 2024-03-20
Migrate existing integration tests to a single binary

BrynCooke:
Each integration tests takes time to link, so this will help with build times.
It also makes things a bit less chaotic as we can have modules for integration tests.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/tests/telemetry_test.rs
```diff
@@ -1,55 +0,0 @@
...
-use std::result::Result;
-
-use tower::BoxError;
-
-use crate::common::IntegrationTest;
-use crate::common::Telemetry;
-
-#[ignore]
-#[tokio::test(flavor = "multi_thread")]
-async fn test_otlp_tracing() -> Result<(), BoxError> {
```
SimonSapin, line 13: Is removing rather than moving these deliberate?

SimonSapin (approved):
Do we also want [`autotests = false`](https://doc.rust-lang.org/cargo/reference/cargo-targets.html#target-auto-discovery) in `Cargo.toml`?

apollo-router/tests/telemetry_test.rs
BrynCooke: Yeah, these tests are ignored anyway, and they didn't have anything actually set up to verify. They were just placeholders.

BrynCooke:
> Do we also want [`autotests = false`](https://doc.rust-lang.org/cargo/reference/cargo-targets.html#target-auto-discovery) in `Cargo.toml`?

I'm not sure. It's not as if there are extra binaries being created because of bin directories. What's the reason for wanting this?

SimonSapin:
By default (`autotests = true`), every `tests/*.rs` file gets compiled as an integration test executable. From the diff it’s not clear if after this PR there is more than one such file remaining. Setting `autotests = false` would allow having such files (such as to use a modules from the "main" one) but not have them compiled as a separate executable. (Only explicit `[[test]]` TOML sections would create an integration test)

BrynCooke:
| By default (autotests = true), every tests/*.rs file gets compiled as an integration test executable. From the diff it’s not clear if after this PR there is more than one such file remaining. Setting autotests = false would allow having such files (such as to use a modules from the "main" one) but not have them compiled as a separate executable. (Only explicit [[test]] TOML sections would create an integration test)

We're going to have to do this separately. I had a quick go, and the act of doing this for the existing tests changes the mod hierarchy.
Maybe there is some way around this.

Issue #4807, opened 2024-03-14
Tracing service name reported as `unknown_service`


nathanmarcos:
**Describe the bug**
We have tracing set up with zipkin exporter and the `service.name` field from router has the value `unknown_service` despite the name being configured on `router.yaml`.

Root list:
![image](https://github.com/apollographql/router/assets/2166279/af7bcfbf-c8d2-48cd-b601-428e85566462)

Trace:
![image](https://github.com/apollographql/router/assets/2166279/6a6ceae5-f0ca-47a4-9451-94dbfea4c76f)

**To Reproduce**
1. Start zipkin: 
	- `docker run --rm -d -p 9411:9411 openzipkin/zipkin-slim`
2. Start a stock router version `1.42.0` with the following `router.yaml` config:
	- `cargo run -- --supergraph testing_schema.graphql --config router.yaml --dev`
```
telemetry:
  instrumentation:
    spans:
      mode: spec_compliant
  exporters:
    tracing:
      common:
        sampler: 1
        service_name: foo
        service_namespace: bar
        resource:
          "service.name": "foo"
      zipkin:
        enabled: true
        endpoint: "http://localhost:9411/api/v2/spans"
```
3. Shoot some query to the server: 
	- [Example](http://localhost:4000/?explorerURLState=N4IgJg9gxgrgtgUwHYBcQC4QEcYIE4CeABALIEDyADvgIYoCWESRwAOs0QPqcoHVI1E7AL7sQwoA)
4. Check the logs on zipkin:
	- http://localhost:9411/zipkin

**Expected behavior**
The `service.name` field to be carrying the value defined on `router.yaml`.

**Desktop (please complete the following information):**
 - OS: macOS Sonoma
 - Version: 14.2.1

**Additional context**
- Router version: 1.42.0

PR #4757, opened 2024-03-06
Unix socket support for subgraphs

Geal:
Fix #3504

This implements support for Unix sockets in subgraph connections by specifying URLs in the `unix:///path/to/router.sock` format.

Implementation details:
- the Uri type used throughout the router does not actually support the `unix://` scheme, and there is no real standard for it, so the library we use here hides the socket path by hex encoding it in the authority part of the Uri
- this only supports stream sockets
- decompression is supported
- TLS is not supported

Note: I'd like to add it to the doc but can't decide on the right place to put it. Any suggestions?

TODO:
- [x] update the override_subgraph_url plugin to support it

---

**Checklist**

Complete the checklist (and note appropriate exceptions) before the PR is marked ready-for-review.

- [ ] Changes are compatible[^1]
- [ ] Documentation[^2] completed
- [ ] Performance impact assessed and acceptable
- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[^1]: It may be appropriate to bring upcoming changes to the attention of other (impacted) groups. Please endeavour to do this before seeking PR approval. The mechanism for doing this will vary considerably, so use your judgement as to how and when to do this.
[^2]: Configuration is an important part of many changes. Where applicable please try to document configuration examples.
[^3]: Tick whichever testing boxes are applicable. If you are adding Manual Tests, please document the manual testing (extensively) in the Exceptions.

PR #4831, opened 2024-03-21
prep release: v1.43.0

abernix:
> **Note**
>
> When approved, this PR will merge into **the `1.43.0` branch** which will — upon being approved itself — merge into `main`.
>
>  - Version bumps
>  - That it targets the right release branch (`1.43.0` in this case!).
>
---
### Support configurable heartbeat for subscriptions using the WebSocket protocol  ([Issue #4621](https://github.com/apollographql/router/issues/4621))

To support GraphQL Subscription WebSocket implementations such as [DGS](https://netflix.github.io/dgs/) that drop idle connections by design, the router adds the ability to configure a heartbeat to keep active connections alive.

An example router configuration:

```yaml
subscription:
  mode:
    passthrough:
      all:
        path: /graphql
        heartbeat_interval: enable # Optional
```

By [@IvanGoncharov](https://github.com/IvanGoncharov) in https://github.com/apollographql/router/pull/4802

### Unix socket support for subgraphs ([Issue #3504](https://github.com/apollographql/router/issues/3504))

> ⚠️ This is an [Enterprise feature](https://www.apollographql.com/blog/platform/evaluating-apollo-router-understanding-free-and-open-vs-commercial-features/) of the Apollo Router. It requires an organization with a [GraphOS Enterprise plan](https://www.apollographql.com/pricing/).
>
> If your organization doesn't currently have an Enterprise plan, you can test out this functionality by signing up for a free Enterprise trial.

The router now supports Unix sockets for subgraph connections by specifying URLs in the `unix:///path/to/router.sock` format in the schema, in addition to coming a valid URL option within [the existing `override_subgraph_url` configuration](https://www.apollographql.com/docs/router/configuration/overview/#subgraph-routing-urls). The router uses Unix stream-oriented sockets (not datagram-oriented). It supports compression but not TLS.

Due to the lack of standardization of Unix socket URLs (and lack of support in the common URL types in Rust) a transformation is applied to to the socket path to parse it: the host is encoded in hexadecimal and stored in the `authority` part. This will have no consequence on the way the router functions, but [subgraph services](https://www.apollographql.com/docs/router/customizations/overview/#the-request-lifecycle) will receive URLs with the hex-encoded host.

By [@Geal](https://github.com/Geal) in https://github.com/apollographql/router/pull/4757

### Add an option to refresh expiration on Redis GET ([Issue #4473](https://github.com/apollographql/router/issues/4473))

This adds the option to refresh the time-to-live (TTL) on Redis entries when they are accessed. We want the query plan cache to act like an LRU cache (least-recently used), so if a TTL is set in its Redis configuration, it should reset every time it is accessed.

While the option is also available for APQ, it's disabled for entity caching because that cache manages TTL directly.

By [@Geal](https://github.com/Geal) in https://github.com/apollographql/router/pull/4604

### Helm: Support configuring `j` on Horizontal Pod Autoscaler (HPA) targets ([PR #4776](https://github.com/apollographql/router/pull/4776))

The router supports configuration of the [`ContainerResource` type metric](https://kubernetes.io/blog/2023/05/02/hpa-container-resource-metric/) on Horizontal Pod Autoscaler (HPA) targets in the Helm chart with Kubernetes v1.27 or later.

By [@caugustus](https://github.com/caugustus) in https://github.com/apollographql/router/pull/4776

### Fix chunk formatting in multipart protocol ([Issue #4634](https://github.com/apollographql/router/issues/4634))

Previously, when sending a stream of chunks for HTTP multipart, the router finished each chunk by appending it with `
`.
Now, the router doesn't append `
` to the current chunk but instead prepends it to the next chunk. This enables the router to close a stream with the correct final boundary by appending `--
` directly to the last chunk.

This PR changes the way we're sending chunks in the stream. Instead of finishing the chunk with `
` we don't send this at the end of our current chunk but instead at the beginning of the next one. For the end users nothing changes but it let us to close the stream with the right final boundary by appending `--
` directly to the last chunk.

By [@bnjjj](https://github.com/bnjjj) in https://github.com/apollographql/router/pull/4681

### Zipkin service name not populated ([Issue #4807](https://github.com/apollographql/router/issues/4807))

The Zipkin trace exporter now respects service name configuration from YAML or environment variables.

For instance to set the service name to `my-app`, you can use the following configuration in your `router.yaml` file:

```yaml
telemetry:
  exporters:
    tracing:
      common:
        service_name: my-app
      zipkin:
        enabled: true
        endpoint: default
```

By [@BrynCooke](https://github.com/BrynCooke) in https://github.com/apollographql/router/pull/4816

### Preallocate response formatting output ([PR #4775](https://github.com/apollographql/router/pull/4775))

To improve runtime performance, an internal change to the router's `format_response` now preallocates the output object.

By [@xuorig](https://github.com/xuorig) in https://github.com/apollographql/router/pull/4775


abernix:
This is ready for review.  Please take a look!

CHANGELOG.md
```diff
@@ -4,6 +4,86 @@ All notable changes to Router will be documented in this file.
...
+
+## 🐛 Fixes
+
+### Fix chunk formatting in multipart protocol ([Issue #4634](https://github.com/apollographql/router/issues/4634))
+
+This PR changes the way we're sending chunks in the stream. Instead of finishing the chunk with `\r\n` we don't send this at the end of our current chunk but instead at the beginning of the next one. For the end users nothing changes but it let us to close the stream with the right final boundary by appending `--\r\n` directly to the last chunk.
+
+By [@bnjjj](https://github.com/bnjjj) in https://github.com/apollographql/router/pull/4681
+
+### Use `Object::with_capacity` with selection set length in `format_response` ([PR #4775](https://github.com/apollographql/router/pull/4775))
```
BrynCooke, line 62: Would put this under `improvement` rather than `fix`.

BrynCooke (approved):
Approved with a minor comment

CHANGELOG.md
abernix: We don't have an "Improvements" section, but I moved it to "Maintenance".

garypen (commented):
Line 48 in the Changelog. What is `j`?

abernix:
> Line 48 in the Changelog. What is `j`?

That's what happens when you're using vim and don't realize you accidentally a word.

CHANGELOG.md
```diff
@@ -4,6 +4,91 @@ All notable changes to Router will be documented in this file.
...
+
+### Add an option to refresh expiration on Redis GET ([Issue #4473](https://github.com/apollographql/router/issues/4473))
+
+This adds the option to refresh the time-to-live (TTL) on Redis entries when they are accessed. We want the query plan cache to act like an LRU cache (least-recently used), so if a TTL is set in its Redis configuration, it should reset every time it is accessed.
+
+While the option is also available for APQ, it's disabled for entity caching because that cache manages TTL directly.
+
+By [@Geal](https://github.com/Geal) in https://github.com/apollographql/router/pull/4604
+
+### Helm: Support configuring `j` on Horizontal Pod Autoscaler (HPA) targets ([PR #4776](https://github.com/apollographql/router/pull/4776))
```
abernix, line 48:
```suggestion
### Helm: Support configuring `ContainerResource` on Horizontal Pod Autoscaler (HPA) targets ([PR #4776](https://github.com/apollographql/router/pull/4776))
```

abernix: @garypen Pulling your top-level comment into this thread.  This is fixed here in [4e0a580](https://github.com/apollographql/router/pull/4831/commits/4e0a5802525dd24821f29333bf23b712b6763a69).

garypen (approved):

Geal (approved):

PR #4840, opened 2024-03-22
move the request body logging to debug

tapaderster:


garypen (commented):
@tapaderster Thanks for the PR. We'll incorporate that into the next release. We'll also check for any more instances like this which need attention.

garypen:
I'm closing this because we added it to the development branch in this commit: https://github.com/apollographql/router/pull/4849/commits/0ce51eae132bd4f04107f2391f77f6bb936c78d1

PR #4857, opened 2024-03-26
Add map of subgraph schemas to QueryPlan

tninesling:
This change makes the individual subgraph schemas from the query planner accessible to plugins by adding them to the query plan. This enables the demand control plugin to parse and score subgraph operations from each fetch node in the plan.

Precursor to [ROUTER-174](https://apollographql.atlassian.net/browse/ROUTER-174)

---

**Checklist**

Complete the checklist (and note appropriate exceptions) before the PR is marked ready-for-review.

- [ ] Changes are compatible[^1]
- [ ] Documentation[^2] completed
- [ ] Performance impact assessed and acceptable
- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[^1]: It may be appropriate to bring upcoming changes to the attention of other (impacted) groups. Please endeavour to do this before seeking PR approval. The mechanism for doing this will vary considerably, so use your judgement as to how and when to do this.
[^2]: Configuration is an important part of many changes. Where applicable please try to document configuration examples.
[^3]: Tick whichever testing boxes are applicable. If you are adding Manual Tests, please document the manual testing (extensively) in the Exceptions.

[ROUTER-174]: https://apollographql.atlassian.net/browse/ROUTER-174?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #4873, opened 2024-03-27
Surface query planner subgraph schemas to plugins via PluginInit

tninesling:
Adds a new method `PluginInit::unsupported_subgraph_schemas()` which exposes the Query Planner's subgraph schemas to plugins. Passing the schemas through the init struct addresses the plan caching expansion issues discussed in #4857.

Precursor to [ROUTER-174](https://apollographql.atlassian.net/browse/ROUTER-174)

---

**Checklist**

- Tests added and passing[^3]
    - [X] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-174]: https://apollographql.atlassian.net/browse/ROUTER-174?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apollo-router/src/plugin/mod.rs
```diff
@@ -104,51 +110,54 @@ where
     pub fn try_new(
         config: serde_json::Value,
         supergraph_sdl: Arc<String>,
+        subgraph_schemas: Arc<HashMap<String, Valid<Schema>>>,
```
BrynCooke, line 113: TODO undo change to pub signature.

apollo-router/src/plugin/mod.rs
```diff
@@ -81,7 +82,11 @@ where
 {
     #[deprecated = "use PluginInit::builder() instead"]
     /// Create a new PluginInit for the supplied config and SDL.
-    pub fn new(config: T, supergraph_sdl: Arc<String>) -> Self {
+    pub fn new(
+        config: T,
+        supergraph_sdl: Arc<String>,
+        subgraph_schemas: Arc<HashMap<String, Valid<Schema>>>,
```
BrynCooke, line 88: TODO undo change to pub signature.

apollo-router/src/plugin/mod.rs
```diff
@@ -69,8 +68,10 @@ pub struct PluginInit<T> {
     /// Router Supergraph Schema (schema definition language)
     pub supergraph_sdl: Arc<String>,
     /// The supergraph schema (parsed)
-    pub(crate) supergraph_schema:
-        Arc<apollo_compiler::validation::Valid<apollo_compiler::schema::Schema>>,
+    pub(crate) supergraph_schema: Arc<Valid<Schema>>,
+
+    /// The parsed subgraph schemas from the query planner, keyed by subgraph name
+    pub(crate) subgraph_schemas: Arc<HashMap<String, Valid<Schema>>>,
```
Geal, line 74:
```suggestion
    pub(crate) subgraph_schemas: Arc<HashMap<String, Arc<Valid<Schema>>>>,
```
if a subgraph plugin wants to keep a copy of the schema around, then we should avoid copying the entire structure

tninesling: Thanks for the feedback. Updated to use Arc

apollo-router/src/plugin/mod.rs
```diff
@@ -164,13 +165,15 @@ where
     pub(crate) fn new_builder(
         config: T,
         supergraph_sdl: Arc<String>,
-        supergraph_schema: Arc<apollo_compiler::validation::Valid<apollo_compiler::schema::Schema>>,
+        supergraph_schema: Arc<Valid<Schema>>,
+        subgraph_schemas: Arc<HashMap<String, Arc<Valid<Schema>>>>,
```
tninesling, line 169: I removed the updates to the public `new` and `try_new` to preserve backward compatibility, but it occurs to me this will also be an API change, as it affects the derived builder code.

Will just wrapping these parameters in an `Option` be sufficient to make that backward compatible?

tninesling:
> would it make sense to store the subgraph schemas in `spec::Schema` ?

In that case, I don't think we could guarantee a given `spec::Schema` always has subgraph schemas, at least not in this implementation where we fetch them from the JS planner. I think that would make sense if we could synchronously generate them from the supergraph SDL though.

Geal (approved):

PR #4875, opened 2024-03-27
chore: Update issue templates to use headers

smyrick:
All these templates use bold text for the sections which are difficult to parse in long text blocks. Using H2 headers instead add line breaks and cleaner sections


smyrick:
@abernix With out the changeset the checks are failing, so it looks like a manual merge is required

abernix:
@smyrick That test isn't required.  It's the other failing tests that are problematic.  Will look soon.

PR #4881, opened 2024-03-28
Expose error code on false positive metric in Rust-based validation

goto-bus-stop:
Updates the apollo-compiler dependencies to get access to `unstable_error_name()`. As an aside, apollo-compiler no longer uses salsa, improving validation performance.

The hope is that by adding this to the metric we get some better idea of the remaining buggy areas in apollo-compiler validation.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


Geal (approved):

garypen (approved):

goto-bus-stop:
Latest statistics suggest that there are *very* few false positives, so not clear that we'll get a lot of use out of this, but i'll merge it anyway because if there continues to be a longer tail of validation issues then it's better to have this now than in the future (and if we can find even 1 of the remaining issues with this it's worthwhile)

PR #4883, opened 2024-03-28
reuse cached query plans across schema updates if possible

### LINKED ISSUE #4834: Entity cache: query hashing should take requires and provides into account
This method of hashing considers directives, but only their presence and arguments. If we want to extend it outside of subgraph queries, it needs to consider how some arguments work, like `@requires`, because it may introduce more fields in the query, and those fields could have changed

Geal:
Fix #4834

This extends the schema aware hashing already used for subgraph queries in entity caching, to be usable for client queries, and use that hash in the query planner cache to be able to reuse cached entries across schema reloads.
This contains:
- an update of the traverse visitor to use an `ExecutableDocument` (necessary to parse field sets in `key` and `requires` argument
- parses field sets in `@join__type`'s `key` argument and `@join__field`'s `requires` argument, because they can be affected by schema updates
- remove the hack around `_entities` operation
- parse subgraph queries using the subgraph schemas extracted from the supergraph schema
- update query planner cache warm up to check if the query hash changed

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


bnjjj (approved):

Geal:
feedback from @bnjjj off band: we should make it available via an option, so users can roll it back if we find an issue

PR #4885, opened 2024-03-28
Add `generate_query_fragments` configuration option

trevor-scheer:
Closing the loop on https://github.com/apollographql/federation/pull/2958

- [ ] I'm unsure if override + log warning is the appropriate action, do we have a similar pattern in config elsewhere?
- [x] Is there a place where the yaml config options are all documented? I don't see anything existing for `reuse_query_fragments`.
- [ ] How to test the log warning was emitted?


BrynCooke:
Looks like documentation still needs updating.

trevor-scheer:
@BrynCooke I added some docs for the new option as well as the existing option since they're pretty related. Let me know if you have any opinions on where the new section belongs, if you think it should go somewhere that isn't where I put it.

apollo-router/src/configuration/mod.rs
```diff
@@ -610,6 +610,11 @@ pub(crate) struct Supergraph {
     #[serde(rename = "experimental_reuse_query_fragments")]
     pub(crate) reuse_query_fragments: Option<bool>,
 
+    /// Enable QP generation of fragments for subgraph requests
+    /// Default: false
+    #[serde(rename = "generate_query_fragments")]
```
Geal, line 615:
```suggestion
```
renaming is not necessary here

Geal (approved):

Geal:
the failing integration tests could be due to a cache key changing with the option https://app.circleci.com/pipelines/github/apollographql/router/20059/workflows/bf8a774f-1e0d-49eb-a5ec-4324d6ac90f0/jobs/134413

apollo-router/src/configuration/mod.rs
```diff
@@ -610,6 +610,11 @@ pub(crate) struct Supergraph {
...
+    pub(crate) generate_query_fragments: Option<bool>,
```
BrynCooke, line 616: Both `reuse_query_fragments` and `generate_query_fragments` should not be wrapped in an `Option` instead set the default in the implementation of `Default`

BrynCooke (changes requested):

apollo-router/src/configuration/mod.rs
trevor-scheer: I might need a bit of help with what you mean here.

I think `reuse_query_fragments` is intentionally an `Option` for historical reasons. IIUC, the router defers to QP to set the default based on its version, though perhaps that's not really an issue since versions are pretty tightly coupled?

I removed the `Option` from `generate_query_fragments` anyway. But I'm not sure if the change in https://github.com/apollographql/router/pull/4885/commits/4c5ef94cda795b3055513ef9dab21853608f2192 is quite what you're after - there's nothing really _in_ `impl Default for Supergraph`, should I be adding something there?

Something like this perhaps?
```rust
impl Default for Supergraph {
    fn default() -> Self {
        Self::builder()
            .generate_query_fragments(false)
            .reuse_query_fragments(true)
            .build()
    }
}
```

BrynCooke: You got it correct. The `Default` impl delegates to the new contructor.

BrynCooke (approved):

apollo-router/src/configuration/mod.rs
```diff
@@ -710,7 +710,10 @@ impl Supergraph {
 
 impl Default for Supergraph {
     fn default() -> Self {
-        Self::builder().build()
+        Self::builder()
+            .generate_query_fragments(false)
+            .reuse_query_fragments(true)
```
trevor-scheer, line 715: @BrynCooke I didn't mean to push this, but you approved. Based on what you said, I don't think this is what we wanted. Shall I remove it, or is this actually correct?

apollo-router/tests/integration/mod.rs
```diff
@@ -1,8 +1,6 @@
 #[path = "../common.rs"]
 pub(crate) mod common;
 pub(crate) use common::IntegrationTest;
-pub(crate) use common::Telemetry;
-pub(crate) use common::ValueExt;
```
trevor-scheer: I get lint errors locally about these being unused, but they seem to be causing lint to fail in CI. Any idea what the discrepancy is? Going to revert this change.

BrynCooke: Are you running `cargo task lint`?

apollo-router/src/configuration/mod.rs
trevor-scheer: `reuse_query_fragments` is still an `Option<bool>` though (you had suggested I change it to a `bool` in a separate thread, but I wasn't certain that's appropriate - maybe it is, but it seemed that way for a reason), so I think returning just `reuse_query_fragments` in this case is correct.

BrynCooke: Fair enough. Also I think I may have been mixing up `generate_query_fragments` and `reuse_query_fragments` in the statements above.

apollo-router/tests/integration/mod.rs
nicholascioli: @garypen discovered that the tests are only enabled on linux targets, so the includes here seem unused on our MacOS development machines.

nicholascioli: We'll fix this in a separate PR.

trevor-scheer:
Thank you both!

Issue #4852, opened 2024-03-26
subscription: Ping message over graphql-ws protocol contains invalid payload


IvanGoncharov:
**Describe the bug**
According to the `graphql-ws` spec, the payload should be an object, but the router sends a string.

**Expected behavior**
Send correct payload inside Ping

Issue #4827, opened 2024-03-19
More accurately nest spans in tracing


Fluxx:
**Describe the bug**

In the spans that the Router generates, it seems there are some pairs of spans which are currently implemented as siblings, but I think might actually be parent/child? For instance, here is a recent trace I pulled that involved our router. The router is configured to export :

![image](https://github.com/apollographql/router/assets/5892/851bc75e-49b1-465a-8b39-a75f0964bfd6)

The green `POST` spans are our coprocessor. I am wondering, for the red boxes around pairs of spans, should the first span of the pair be the parent of the second span? i.e. should the `router` span be the parent of its `external_plugin` span? Or should the `supegraph` span be the parent of its corresponding `external_plugin` span?

**To Reproduce**

1. Run the route with tracing enabled, using the `otlp` exporter.
2. Record a request involving at least one subgraph.

**Expected behavior**

I would expect for each red pair of spans in the screenshot above, that the first span would be a parent of the second span.

**Additional context**

We're running router version `1.40.2` with this relevant section of our config file:

```
    tracing:
      common:
        service_name: graphql
        sampler: always_on
        parent_based_sampler: true
      otlp:
        enabled: true
        endpoint: <internal-collection-endpoint>
        protocol: http
      propagation:
        baggage: true
        trace_context: true
```

PR #4823, opened 2024-03-19
Enable displaying trace and span id on logs

BrynCooke:
To enable correlation between trace and logs trace_id and span_id should be included on the log messages.

 Json:
 ```
 {"timestamp":"2024-03-19T15:37:41.516453239Z","level":"INFO","trace_id":"54ac7e5f0e8ab90ae67b822e95ffcbb8","span_id":"9b3f88c602de0ceb","message":"Supergraph GraphQL response".....
 ```

 Text:
 ```
 2024-03-19T15:14:46.040435Z INFO trace_id: bbafc3f048b6137375dd78c10df18f50 span_id: 40ede28c5df1b5cc router{
 ```

 Configuration options have been added in logging for text and json formats defaulting to true for json and false for text.

Json:
```
   exporters:
     logging:
       stdout:
         format:
           json:
             display_span_id: true
             display_trace_id: true
```

Text:
```
  exporters:
    logging:
      stdout:
        format:
          text:
            display_span_id: true
            display_trace_id: true
```

---

**Checklist**

Complete the checklist (and note appropriate exceptions) before the PR is marked ready-for-review.

- [ ] Changes are compatible[^1]
- [ ] Documentation[^2] completed
- [ ] Performance impact assessed and acceptable
- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[^1]: It may be appropriate to bring upcoming changes to the attention of other (impacted) groups. Please endeavour to do this before seeking PR approval. The mechanism for doing this will vary considerably, so use your judgement as to how and when to do this.
[^2]: Configuration is an important part of many changes. Where applicable please try to document configuration examples.
[^3]: Tick whichever testing boxes are applicable. If you are adding Manual Tests, please document the manual testing (extensively) in the Exceptions.

Issue #4798, opened 2024-03-13
tracing otlp exporter endpoint configuration for http doesn't append the right path


bnjjj:
In our [documentation](https://www.apollographql.com/docs/router/configuration/telemetry/exporters/tracing/otlp/#endpoint) when configuration http endpoint for tracing otlp exporter we say that users should only include the base address of the otlp endpoint.

So for example you'll have to specify the right path for http:

```yaml
telemetry:
  exporters:
    tracing:
      otlp:
        enabled: true
        endpoint: "http://localhost:4318/v1/traces"
        protocol: http
```

but not for GRPC:

```yaml
telemetry:
  exporters:
    tracing:
      otlp:
        enabled: true
        endpoint: "http://localhost:4317"
        protocol: grpc
```

This inconsistency is due to a bug in `opentelemetry-otlp` crate, I raised an issue [here](https://github.com/open-telemetry/opentelemetry-rust/issues/1618)

PR #4889, opened 2024-03-29
prep release: v1.43.1

### LINKED ISSUE #4856: 400 Bad Request when request.body.extensions=null
**Describe the bug**
The gateway supported a `null` value for `extensions` key in the request body but the router does not/returns a 400

**To Reproduce**
Send a request with `extensions: null` to the router (doesn't matter what query is as the request doesn't get far enough in router lifecycle to matter):
```
curl "http://localhost:4001/router" \
  -H 'Content-Type: application/json' \
  -d $'{"query": "query foo {id}","extensions": null}'
```

**Expected behavior**
Router processes the request instead of throwing a 400 just as the gateway does

**Output**
Here is the exact output from above query:
```
{
  "errors": [
    {
      "extensions": {
        "code": "INVALID_GRAPHQL_REQUEST",
        "details": "failed to deserialize the request body into JSON: invalid type: null, expected a map at line 1 column 46"
      },
      "message": "Invalid GraphQL request"
    }
  ]
}
```

**Router:**
 - OS: Amazon Linux 2023
 - Router Version: 1.40.0

**Additional context**
The router does support `extensions: {}` but not `extensions: null`

We have Xamarin iOS and Android apps that do not use the Apollo Client where the `extensions` key is set to `null` if no extensions to be included in the request. It is not trivial to update all the installs in the wild to use `{}` instead of `null` and there does not appear to be a way to use Rhai to modify the request body before request validation in the router occurs. This is preventing adoption of the router at this time.

BrynCooke:
> **Note**
    >
    > When approved, this PR will merge into **the `1.43.1` branch** which will — upon being approved itself — merge into `main`.
    >
    >  - Version bumps
    >  - That it targets the right release branch (`1.43.1` in this case!).
    >
    ---
### Enable displaying trace and span id on logs ([PR #4823](https://github.com/apollographql/router/pull/4823))

To enable correlation between trace and logs `trace_id` and `span_id` can now be included on the log messages.

 Json:
 ```
 {"timestamp":"2024-03-19T15:37:41.516453239Z","level":"INFO","trace_id":"54ac7e5f0e8ab90ae67b822e95ffcbb8","span_id":"9b3f88c602de0ceb","message":"Supergraph GraphQL response".....
 ```

 Text:
 ```
 2024-03-19T15:14:46.040435Z INFO trace_id: bbafc3f048b6137375dd78c10df18f50 span_id: 40ede28c5df1b5cc router{
 ```

To configure this, use `display_span_id` and `display_trace_id` options in the logging exporter configuration.

Json (defaults to true):
```
telemetry:
  exporters:
    logging:
      stdout:
        format:
          json:
            display_span_id: true
            display_trace_id: true
```

Text (defaults to false):
```
telemetry:
  exporters:
    logging:
      stdout:
        format:
          text:
            display_span_id: false
            display_trace_id: false
```

By [@BrynCooke](https://github.com/BrynCooke) in https://github.com/apollographql/router/pull/4823

### count errors in the apollo.router.graphql_error metrics ([Issue #4749](https://github.com/apollographql/router/issues/4749))

This introduces a new counter metric called `apollo.router.graphql_error` that counts GraphQL errors, split by their code

By [@Geal](https://github.com/Geal) in https://github.com/apollographql/router/pull/4751

### expose the operation signature in the context ([Issue #4558](https://github.com/apollographql/router/issues/4558))

It is available at the key `apollo_operation_signature`.

By [@Geal](https://github.com/Geal) in https://github.com/apollographql/router/pull/4864

### Experimental logging of broken pipe errors ([PR #4870](https://github.com/apollographql/router/pull/4870))

You can now emit a log message each time the client closes the connection early, which can help you debug issues with clients that close connections before the server can respond. 

This feature is disabled by default but can be enabled by setting the `experimental_log_broken_pipe` option to `true`:

```yaml title="router.yaml"
supergraph:
  experimental_log_on_broken_pipe: true
```

Users that have internet facing routers will likely not want to opt in to this log message as they have no control over the clients.

By [@Geal](https://github.com/Geal) in https://github.com/apollographql/router/pull/4770 and [@BrynCooke](https://github.com/BrynCooke) in https://github.com/apollographql/router/pull/4870

### Entity cache: fix support for Redis cluster ([PR #4790](https://github.com/apollographql/router/pull/4790))

in a Redis cluster, entities can be stored in different nodes, and a query to one node should only refer to the keys it manages. This is challenging for the MGET operation which requests multiple entities in the same request from the same node. This splits the MGET query in multiple MGETs calls grouped by key hash, to make sure each one will get to the corresponding node, then merges responses in the correct order.

By [@Geal](https://github.com/Geal) in https://github.com/apollographql/router/pull/4790

### Give spans their proper parent in the plugin stack ([Issue #4827](https://github.com/apollographql/router/issues/4827))

Due to the way plugin spans were created and applied, they would appear as siblings instead of being nested, which creates some issues when displaying traces and accounting for time spent in Datadog. Plugin spans are now correctly nested within each other.

By [@Geal](https://github.com/Geal) in https://github.com/apollographql/router/pull/4877

### Fix(telemetry): keep consistency between tracing otlp endpoint ([Issue #4798](https://github.com/apollographql/router/issues/4798))

In our [documentation](https://www.apollographql.com/docs/router/configuration/telemetry/exporters/tracing/otlp/#endpoint) when configuration http endpoint for tracing otlp exporter we say that users should only include the base address of the otlp endpoint. It was only working for grpc protocol and not http due to [this bug](https://github.com/open-telemetry/opentelemetry-rust/issues/1618). This inconsistency is now fixed with a workaround in the router waiting for the fix in openetelemetry crate.

So for example you'll have to specify the right path for http:

```yaml
telemetry:
  exporters:
    tracing:
      otlp:
        enabled: true
        endpoint: "http://localhost:4318"
        protocol: http
```

By [@bnjjj](https://github.com/bnjjj) in https://github.com/apollographql/router/pull/4801

### Execute the entire request pipeline if the client closed the connection ([Issue #4569](https://github.com/apollographql/router/issues/4569)), [Issue #4576](https://github.com/apollographql/router/issues/4576)), ([Issue #4589](https://github.com/apollographql/router/issues/4589)), ([Issue #4590](https://github.com/apollographql/router/issues/4590)), ([Issue #4611](https://github.com/apollographql/router/issues/4611))

The router is now making sure that the entire request handling pipeline is executed when the client closes the connection early, to let telemetry and any rhai scrit or coprocessor perform their tasks before canceling. Before that, when a client canceled a request, the entire execution was dropped and parts of the router, like telemetry, could not run properly. It now executes up to the first response event (in the case of subscription or `@defer` usage), adds a 499 status code to the response and skips the remaining subgraph requests.

This change will report more requests to Studio and the configured telemetry, which will appear like a sudden increase in errors, because those failing requests were not reported before. To keep the previous behavior of immediately dropping execution for canceled requests, it is possible with the following option:

```yaml
supergraph:
  early_cancel: true
```

By [@Geal](https://github.com/Geal) in https://github.com/apollographql/router/pull/4770

### `null` extensions incorrectly disallowed on request ([Issue #4856](https://github.com/apollographql/router/issues/4856))

The [graphql over http spec](https://graphql.github.io/graphql-over-http/draft/#sel-EALFPCCBCEtC37P) mandates `null` is allowed for request extensions.

We were previously rejecting such payloads, but now we will allow them. For example:

```json
{
  "query": "{ topProducts { upc name reviews { id product { name } author { id name } } } }",
  "variables": {
    "date": "2022-01-01T00:00:00+00:00"
  },
  "extensions": null
}
```

Fixes #4856

By [@BrynCooke](https://github.com/BrynCooke) in https://github.com/apollographql/router/pull/4865

### fix the log lines for external extensibility ([PR #4869](https://github.com/apollographql/router/pull/4869))

The `execution` and `supergraph` responses were mistakenly reporting as `router`. This is now fixed.

By [@garypen](https://github.com/garypen) in https://github.com/apollographql/router/pull/4869

### Remove invalid payload on graphql-ws Ping message ([Issue #4852](https://github.com/apollographql/router/issues/4852))

According to [graphql-ws spec](https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md#ping) `Ping` payload should be an object or null but router was sending a string.
To ensure better compatibility Ping's payload was removed. 

By [@IvanGoncharov](https://github.com/IvanGoncharov) in https://github.com/apollographql/router/pull/4852


CHANGELOG.md
```diff
@@ -4,6 +4,159 @@ All notable changes to Router will be documented in this file.
...
+### fix the log lines for external extensibility ([PR #4869](https://github.com/apollographql/router/pull/4869))
+
+The `execution` and `supergraph` responses were mistakenly reporting as `router`. This is now fixed.
+
+By [@garypen](https://github.com/garypen) in https://github.com/apollographql/router/pull/4869
+
+### Remove invalid payload on graphql-ws Ping message ([Issue #4852](https://github.com/apollographql/router/issues/4852))
+
+According to [graphql-ws spec](https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md#ping) `Ping` payload should be an object or null but router was sending a string.
+To ensure better compatibility Ping's payload was removed.
```
shorgi, line 154:
```suggestion
Previously, the router sent a string as a `Ping` payload, but that was incompatible with the [graphql-ws specification](https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md#ping), which specifies that the payload is optional and should be an object or null.

To ensure compatibility, the router now sends no payload for `Ping` messages. 
```

shorgi (approved):
Reviewed changelog

nicholascioli (approved):
LGTM!

PR #4897, opened 2024-04-03
Experimental: Introduce a pool of query planners

o0Ignition0o:
### Experimental: Introduce a pool of query planners ([PR #4897](https://github.com/apollographql/router/pull/4897))

The router supports a new experimental feature: a pool of query planners to parallelize query planning.

You can configure query planner pools with the `supergraph.query_planner.experimental_available_parallelism` option:

```yaml
supergraph:
  query_planner:
    experimental_available_parallelism: auto # number of available cpus
```

Its value is the number of query planners that run in parallel, and its default value is `1`. You can set it to the special value `auto` to automatically set it equal to the number of available CPUs.

You can discuss and comment about query planner pools in this [GitHub discussion](https://github.com/apollographql/router/discussions/4917).

By [@xuorig](https://github.com/xuorig) and [@o0Ignition0o](https://github.com/o0Ignition0o) in https://github.com/apollographql/router/pull/4897


.changesets/exp_carton_ginger_magnet_beacon.md
```diff
@@ -0,0 +1,19 @@
+### Experimental: Introduce a pool of query planners ([PR #4897](https://github.com/apollographql/router/pull/4897))
+
+This changeset introduces an experimental pool of query planners to parallelize query planning.
```
shorgi, line 3:
```suggestion
The router supports a new experimental feature: a pool of query planners to parallelize query planning.
```

.changesets/exp_carton_ginger_magnet_beacon.md
```diff
@@ -0,0 +1,19 @@
...
+Configuration:
+
+```yaml
+supergraph:
+  query_planner:
+    experimental_available_parallelism: auto # number of available cpus
+```
+
+Note you can also set `experimental_available_parallelism` to a number representing how many planners you want to use in a pool.
+The default is `1`.
```
shorgi, line 17:
```suggestion
This feature is experimental, you can discuss it by following this link: https://github.com/apollographql/router/discussions/4917

You can configure query planner pools by setting `supergraph.query_planner.experimental_available_parallelism`:

```yaml
supergraph:
  query_planner:
    experimental_available_parallelism: auto # number of available cpus
```

Its value is the maximum number of query planners that can run in parallel, and its default value is `1`. You can set it to the special value `auto` to automatically set it equal to the number of available CPUs. 
```

File: .changesets/exp_carton_ginger_magnet_beacon.md [hunk 3]
```diff
@@ -0,0 +1,19 @@
...
+
```
shorgi, line 18:
```suggestion
Its value is the maximum number of query planners that can run in parallel, and its default value is `1`. You can set it to the special value `auto` to automatically set it equal to the number of available CPUs.

You can discuss and comment about query planner pools in this [GitHub discussion](https://github.com/apollographql/router/discussions/4917).

```

.changesets/exp_carton_ginger_magnet_beacon.md
```diff
@@ -0,0 +1,19 @@
+### Experimental: Introduce a pool of query planners ([PR #4897](https://github.com/apollographql/router/pull/4897))
+
+This changeset introduces an experimental pool of query planners to parallelize query planning.
+
+This feature is experimental, you can discuss it by following this link: https://github.com/apollographql/router/discussions/4917
+
+
+Configuration:
```
shorgi, line 8:
```suggestion
You can configure query planner pools with the `supergraph.query_planner.experimental_available_parallelism` option:
```

Geal (approved):
could you mention it in the docs?
I am not sure `auto` will be the best option here. Tokio already spawns threads according to the available capacity, so if at the same time there are as many planner threads as there are cores, then we risk not having any capacity left to handle requests because the planners do blocking work

o0Ignition0o:
> could you mention it in the docs?

I'm going to have a pass at it with @shorgi next week <3

> I am not sure auto will be the best option here. Tokio already spawns threads according to the available capacity, so if at the same time there are as many planner threads as there are cores, then we risk not having any capacity left to handle requests because the planners do blocking work

I am not sure either, but the query planner is not synchronous. all it does is making async sends and receives to the deno runtime. I'm not 100% sure on how isolates work but I wouldn't expect an actual CPU to be used. I'll dig a bit more just in case.

apollo-router/src/query_planner/bridge_query_planner_pool.rs
```diff
@@ -0,0 +1,200 @@
...
+use super::bridge_query_planner::BridgeQueryPlanner;
+use super::QueryPlanResult;
+use crate::error::QueryPlannerError;
+use crate::error::ServiceBuildError;
+use crate::services::QueryPlannerRequest;
+use crate::services::QueryPlannerResponse;
+use crate::spec::Schema;
+use crate::Configuration;
+
+static CHANNEL_SIZE: usize = 10_000;
```
garypen, line 26: Should we consider making this configurable? A user could still end up in a bad place with a router which has no effective pushback.
Also: If not configurable, which is fine, this should be a lot lower than 10_000. Let's imagine it takes 20ms to plan a query. (not untypical). Then a queue of 1_000 means a maximum wait of 20seconds... Even if it was 1ms, that's still a potential 1 second wait. 10_000 could mean a 3 min+ wait...

Perhaps consider a queue size of 1_000 at most.

apollo-router/src/query_planner/caching_query_planner.rs
```diff
@@ -317,11 +317,9 @@ where
...
             qp.plan(request).await.map(|response| {
                 if let Some(usage_reporting) = {
-                    context
-                        .extensions()
-                        .lock()
-                        .get::<Arc<UsageReporting>>()
-                        .cloned()
+                    let lock = context.extensions().lock();
+                    let urp = lock.get::<Arc<UsageReporting>>();
+                    urp.cloned()
```
garypen, line 322: This is a no-op change. It was `cloned()` anyway.

apollo-router/src/query_planner/bridge_query_planner_pool.rs
```diff
@@ -0,0 +1,200 @@
...
+            .iter()
+            .map(|p| p.planner().clone())
+            .collect();
+
+        for (worker_id, mut planner) in bridge_query_planners.into_iter().enumerate() {
+            let receiver = receiver.clone();
+
+            tokio::spawn(async move {
+                while let Ok((request, res_sender)) = receiver.recv().await {
+                    let start = Instant::now();
```
garypen, line 104: Should this be at line 116? Otherwise the metric includes waiting for the planner to become ready.

apollo-router/src/query_planner/bridge_query_planner_pool.rs
```diff
@@ -0,0 +1,200 @@
...
+                    let res = svc.call(request).await;
+
+                    f64_histogram!(
+                        "apollo.router.query_planner.duration",
+                        "Duration of the query planning.",
+                        start.elapsed().as_secs_f64(),
+                        [KeyValue::new("workerId", worker_id.to_string())]
+                    );
+
+                    if res_sender.send(res).is_err() {
```
garypen, line 126: What happens here if the client has cancelled/gone away before the result is available?

garypen (commented):
This seems to be different to the approach of using web workers that you were discussing earlier in the week. Was there a problem with that approach? I'm asking because I'm curious, but I think this is a better approach to that anyway.

One other thing worth asking. Instead of managing a queue using `async_channel`, maybe put some kind of adaptive, load-shedding queuing model in front of the pool? e.g.: https://crates.io/crates/little-loadshedder and expose the configuration so that user's can express a queue wait time in configuration?

apollo-router/src/query_planner/bridge_query_planner_pool.rs
xuorig: Great point, 10_000 was chosen arbitrarily at first. 1000 would make sense. Configurable could potentially be useful, at the same time if you hit the max channel size here it's unlikely configuration is the thing you'd want to address. A sane upper limit sounds good for now?

.changesets/exp_carton_ginger_magnet_beacon.md
```diff
@@ -0,0 +1,17 @@
...
+
+You can configure query planner pools with the `supergraph.query_planner.experimental_available_parallelism` option:
+
+```yaml
+supergraph:
+  query_planner:
+    experimental_available_parallelism: auto # number of available cpus
+```
+
+Its value is the number of query planners that run in parallel, and its default value is `1`. You can set it to the special value `auto` to automatically set it equal to the number of available CPUs.
```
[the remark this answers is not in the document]
o0Ignition0o: Good question!

My mac has 8 cpus available, and if i go back and forth from 8 to auto, it will reload at the moment.

apollo-router/src/query_planner/bridge_query_planner_pool.rs
o0Ignition0o: 1_000 sounds good to me. also because we know of the current queue's length, we could followup to make the service not ready if the length reaches a certain treshold!

apollo-router/src/query_planner/bridge_query_planner_pool.rs
o0Ignition0o: Right, we probably should. otoh as a followup i think we want to make sure a planned query always makes it back at least to the Caching service sitting above.

apollo-router/src/query_planner/caching_query_planner.rs
o0Ignition0o: No.

```
When a temporary is returned in a function call in a match scrutinee, its lifetime lasts until the end of the match block, which may be surprising.
```

Locks will be held through the whole if / else scope (same with a match)

apollo-router/src/query_planner/bridge_query_planner_pool.rs
```diff
@@ -0,0 +1,200 @@
...
+
+        while let Some(task_result) = join_set.join_next().await {
+            let bridge_query_planner =
+                task_result.map_err(|e| ServiceBuildError::ServiceError(Box::new(e)))??;
+            bridge_query_planners.push(bridge_query_planner);
+        }
+
+        let schema = bridge_query_planners
+            .first()
+            .expect("There should be at least 1 service in pool")
```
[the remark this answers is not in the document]
o0Ignition0o: Agree!

o0Ignition0o:
> This seems to be different to the approach of using web workers that you were discussing earlier in the week. Was there a problem with that approach? I'm asking because I'm curious, but I think this is a better approach to that anyway.

It turns out both approaches are equivalent in term of [runtime capabilities](https://choubey.gitbook.io/internals-of-deno/threading-model/default-threads#denos-threading-model-with-web-workers). Thankfully a v8 runtime is initialized in a [static](https://github.com/denoland/deno_core/blob/79e73e500e409063cfbc165b58314d2c66cbf371/core/runtime/setup.rs#L78) :D 

The good news is we should be able to drop in any planner implementation in the future.

> One other thing worth asking. Instead of managing a queue using async_channel, maybe put some kind of adaptive, load-shedding queuing model in front of the pool? e.g.: https://crates.io/crates/little-loadshedder and expose the configuration so that user's can express a queue wait time in configuration?

This could be worth considering as a followup. I'm fairly happy with the MPMC approach since workers decide to pick new jobs as soon as they're ready to deal with them.

apollo-router/src/error.rs
```diff
@@ -292,6 +292,9 @@ pub(crate) enum QueryPlannerError {
 
     /// Unauthorized field or type
     Unauthorized(Vec<Path>),
+
+    /// Query planner pool error: {0}
+    PoolError(String),
```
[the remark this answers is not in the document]
o0Ignition0o: Fixed in [1f7e478](https://github.com/apollographql/router/pull/4897/commits/1f7e478781f24f12f064ca669cddd8688b82db6d)

apollo-router/src/query_planner/bridge_query_planner_pool.rs
o0Ignition0o: Fixed in [1f7e478](https://github.com/apollographql/router/pull/4897/commits/1f7e478781f24f12f064ca669cddd8688b82db6d)

apollo-router/src/query_planner/bridge_query_planner_pool.rs
o0Ignition0o: Fixed in [1f7e478](https://github.com/apollographql/router/pull/4897/commits/1f7e478781f24f12f064ca669cddd8688b82db6d)

apollo-router/src/query_planner/bridge_query_planner_pool.rs
o0Ignition0o: Fixed in [1f7e478](https://github.com/apollographql/router/pull/4897/commits/1f7e478781f24f12f064ca669cddd8688b82db6d)

apollo-router/src/query_planner/bridge_query_planner_pool.rs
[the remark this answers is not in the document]
o0Ignition0o: fixed in 68d637400

apollo-router/Cargo.toml
```diff
@@ -67,6 +67,7 @@ anyhow = "1.0.80"
 apollo-compiler = "=1.0.0-beta.14"
 apollo-federation = "=0.0.9"
 arc-swap = "1.6.0"
+async-channel = "1.9.0"
```
bnjjj, line 70: Could you use [broadcast channels](https://docs.rs/tokio/latest/tokio/sync/broadcast/index.html) from tokio instead ? As we're already using it for subscription and it's already in our deps

apollo-router/src/configuration/mod.rs
```diff
@@ -629,6 +639,88 @@ pub(crate) struct Supergraph {
...
+#[serde(deny_unknown_fields)]
+pub(crate) struct QueryPlanner {
+    /// Set the size of a pool of workers to enable query planning parallelism.
+    /// Default: 1.
+    pub(crate) experimental_available_parallelism: AvailableParallelism,
+}
+
+#[derive(Debug, Clone, JsonSchema)]
+#[serde(untagged)]
+pub(crate) enum AvailableParallelism {
```
bnjjj, line 658: If you want to avoid implementing `Serialize` and `Deserialize` you can do the same trick we did for `HeartbeatInterval` in our codebase

apollo-router/Cargo.toml
bnjjj: NVM it won't work

apollo-router/src/configuration/mod.rs
o0Ignition0o: great idea! done in [a5209a8](https://github.com/apollographql/router/pull/4897/commits/a5209a8a0518e4fd257559b7efb6a44508bb2f2c)

BrynCooke (commented):
Needs a config metric.

garypen (commented):
Does anyone else see this warning:
```
garypen@Garys-MacBook-Pro router % cargo check                         
warning: /Users/garypen/dev/router/apollo-router/Cargo.toml: file `/Users/garypen/dev/router/apollo-router/benches/planner.rs` found to be present in multiple build targets:
  * `example` target `planner`
  * `bench` target `planner`
<etc...>
```
Maybe address this? I don't know if it really matters, but ...

o0Ignition0o:
@garypen this is odd, I don't see several targets in the `Cargo.toml`:

https://github.com/apollographql/router/pull/4897/files#diff-aca654efc6c22bebf4bd167370ab3bf380f3e086befe3d7c6761a8f7eb59d89c

garypen:
> @garypen this is odd, I don't see several targets in the `Cargo.toml`:
> 
> https://github.com/apollographql/router/pull/4897/files#diff-aca654efc6c22bebf4bd167370ab3bf380f3e086befe3d7c6761a8f7eb59d89c

The warning is removed if you add: ```
autobenches = false
```
to the package section in `apollo-router/Cargo.toml`. [Target Autodiscovery](https://doc.rust-lang.org/nightly/cargo/reference/cargo-targets.html#target-auto-discovery) is the cause of the problem.

I think you need to decide if this is the behaviour you want wrt other benches/examples etc...

**Comment by o0Ignition0o** [comment]
@garypen great catch! fixed in [d492701](https://github.com/apollographql/router/pull/4897/commits/d492701bbb618abf50d76a9e1f2909854e781b2d)

File: .changesets/exp_carton_ginger_magnet_beacon.md [hunk 14]
```diff
@@ -0,0 +1,17 @@
+### Experimental: Introduce a pool of query planners ([PR #4897](https://github.com/apollographql/router/pull/4897))
+
+The router supports a new experimental feature: a pool of query planners to parallelize query planning.
+
+You can configure query planner pools with the `supergraph.query_planner.experimental_available_parallelism` option:
+
+```yaml
+supergraph:
+  query_planner:
+    experimental_available_parallelism: auto # number of available cpus
```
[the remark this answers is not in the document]
o0Ignition0o: Fixed in [d492701](https://github.com/apollographql/router/pull/4897/commits/d492701bbb618abf50d76a9e1f2909854e781b2d)

BrynCooke (approved):

garypen (approved):

Issue #4899, opened 2024-04-03
Make `router service call failed` a more specific/helpful error


andrewmcgivery:
**Describe the solution you'd like**
Router currently throws the error `router service call failed` when major something goes wrong. The instances I've seen it in are when a coprocessor is broken or not deployed. The error in it's current form is unhelpful and would be nice if it tried to specify what went wrong (some kind of more specific error).

Currently thrown from two places in the code:

https://github.com/apollographql/router/blob/dev/apollo-router/src/axum_factory/axum_http_server_factory.rs#L650
https://github.com/apollographql/router/blob/dev/apollo-router/src/axum_factory/axum_http_server_factory.rs#L683

PR #4900, opened 2024-04-03
add error details to 'router service call failed'

garypen:
It can be difficult to understand 'router service call failed' messages. Adding the error detail should make them more comprehensible.

fixes: #4899


andrewmcgivery (approved):
Thank you @garypen for the quick turnaround on this!

Issue #3297, opened 2023-06-21
Asynchronous External Coprocessor Requests


lleadbet:
**Describe the solution you'd like**

The ability to have "fire and forget"/asynchronous external coprocessor requests that don't block Router execution, ideally as a new coprocessor stage (e.g. RouterRequestAsynchronous). 

**Describe alternatives you've considered**

- Responding to requests immediately, deferring processing to background tasks/threads. 
- Rhai. 
- Sidecar application (non-coprocessor).

**Additional context**

External coprocessing currently blocks the execution of a request until the coprocessor responds, but for some workloads (e.g. custom metrics/logging) it isn't a blocking task and the request can proceed as normal. 

For example, I need to log in a very specific format ([Bunyan](https://www.npmjs.com/package/bunyan)) that the Router doesn't support today. This log can be quick, but we still will have the overhead of HTTP regardless which can still materially affect performance. Having a way to tell the Router that "hey, this coprocessor request just needs to be sent and not block requests/cause errors on failure" would be ideal.

garypen:
If I understand this request correctly, you are really looking for the router to support a new protocol for communication with coprocessors. An HTTP POST (which is how the router transfers data to coprocessors) requires some kind of response and the library we use for making the POST request will expect and wait for a valid HTTP POST response.

I doubt this is the kind of thing we'll be able to productise in the short term. It doesn't seem like there would be enough demand for it and if we did get a lot of demand, I'd be inclined to investigate other approaches rather than using a coprocessor (which is inherently not high performance).

In the particular use-case you mention, bunyan format logging, we could look at using `https://crates.io/crates/tracing-bunyan-formatter` to add support for bunyan format directly to the router.

PR #4902, opened 2024-04-03
detached coprocessor execution

Geal:
Fix https://github.com/apollographql/router/issues/3297

This adds a new `detached` option at all coprocessor stages to allow the router to continue handling the request or response without waiting for the coprocessor to respond. This implies that the coprocessor response will not be used. This is targeted at use cases like logging and auditing, which do not need to block the router

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


.changesets/feat_geal_async_coprocessor.md
```diff
@@ -0,0 +1,5 @@
+### asynchronous coprocessor execution ([Issue #3297](https://github.com/apollographql/router/issues/3297))
+
+This adds a new `asynchronous` option at all coprocessor stages to allow the router to continue handling the request or response without waiting for the coprocessor to respond. This implies that the coprocessor response will not be used. This is targeted at use cases like logging and auditing, which do not need to block the router
```
bnjjj, line 3: Could you add a full example of configuration with this new option please ?

apollo-router/src/plugins/coprocessor/mod.rs
```diff
@@ -778,6 +806,84 @@ where
...
+                    );
+                });
+            }
+
+            deferred_response
+        });
+
+        // Create our response stream which consists of the bytes from our first body chained with the
+        // rest of the responses in our mapped stream.
+        //let bytes = hyper::body::to_bytes(body).await.map_err(BoxError::from);
```
bnjjj, line 880: to delete ?

docs/shared/coproc-typical-config.mdx
```diff
@@ -4,26 +4,46 @@ coprocessor:
...
     request: # By including this key, the `RouterService` sends a coprocessor request whenever it first receives a client request.
+      asynchronous: false # if set to true, the router will handle the request without waiting for the coprocessor to respond
       headers: true # These boolean properties indicate which request data to include in the coprocessor request. All are optional and false by default.
       body: false
       context: false
       sdl: false
       path: false
       method: false
     response: # By including this key, the `RouterService` sends a coprocessor request whenever it's about to send response data to a client (including incremental data via @defer).
+      asynchronous: false # if set to true, the router will handle the response without waiting for the coprocessor to respond
```
bnjjj, line 15: Maybe you can document it's optional in the comment

Geal: fixed in [db39565](https://github.com/apollographql/router/pull/4902/commits/db395654b60678a5348d065b56615d86c9c0e41c)

apollo-router/src/plugins/coprocessor/mod.rs
Geal: fixed in [db39565](https://github.com/apollographql/router/pull/4902/commits/db395654b60678a5348d065b56615d86c9c0e41c)

.changesets/feat_geal_async_coprocessor.md
Geal: fixed in [db39565](https://github.com/apollographql/router/pull/4902/commits/db395654b60678a5348d065b56615d86c9c0e41c)

docs/shared/coproc-typical-config.mdx
```diff
@@ -32,13 +52,15 @@ coprocessor:
   subgraph:
     all:
       request: # By including this key, the `SubgraphService` sends a coprocessor request whenever it is about to make a request to a subgraph.
+        asynchronous: false # optional. if set to true, the router will handle the request without waiting for the coprocessor to respond
```
o0Ignition0o, line 55: the router will proceed ?

Does asynchronous mean `fire and forget`?

bnjjj (approved):

o0Ignition0o (approved):

garypen (commented):
For some of the reasons I documented in the issue, I don't think this is a good approach to solving this problem.

I'd really prefer us to think about how to handle this from an architectural standpoint rather than shoe-horning behaviour into what is, essentially, an RPC mechanism which expects a response and which is co-processor specific.

One way we could address this kind of requirement is by supporting an externalizable event mechanism to which plugins (or any external program, including coprocessors) could subscribe.

If this is really the way that we think we should implement an external event system, then let's not call it "detached", but maybe call it "event"?

One thing I don't like is that the implementor of a coprocessor doesn't know what the router will do with any response that they provide. There's no indication in the payload that an invocation is "detached" or not. Since the behaviour would only be controlled by router configuration, that seems like a potential footgun in the making.

Another thing, which I allude to above, is that if this is a good idea for coprocessor implementations, then isn't it also a good idea to be able to process events from `Rhai` scripts or indeed in Rust plugins?

Summary: I'd rather we implemented a generalised event mechanism.

Geal:
@garypen sorry, Idid not see your message until after this was merged. While a mechanism specifically focused on events could be interesting, I'd worry about introducing yet another way to extend the router and communicate with other services. Especially since we are currently shipping customizable events in telemetry.
I think you're right though that detached coprocessor requests should have a field to indicate that it does not expect an answer

PR #4903, opened 2024-04-03
Demand Control: Calculate actual cost for a response

tninesling:
Introduces a third method to `CostCalculator` which scores the cost of a GraphQL response.

The first stage of scoring a response is to zip it together with the request. The initial implementation does not use the request fields, but this will be required to support custom cost. In that case, we will need to check the corresponding request field definition to see if a custom cost needs to be applied to a particular response field. This information is stored in a counterpart to `serde_json::Value`, which is called `TypedValue`. This `TypedValue` enum pairs each JSON value with the corresponding `apollo_compiler::executable::Field`.

Once the response values are paired with their corresponding fields, it is simple to traverse the JSON structure, taking query or schema directives into account.

Implements [ROUTER-175](https://apollographql.atlassian.net/browse/ROUTER-175)

**Checklist**

- Tests added and passing[^3]
    - [X] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-175]: https://apollographql.atlassian.net/browse/ROUTER-175?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apollo-router/src/plugins/demand_control/schema_aware_response.rs
```diff
@@ -0,0 +1,197 @@
...
+
+use super::DemandControlError;
+use crate::graphql::Response;
+
+pub(crate) struct SchemaAwareResponse<'a> {
+    pub(crate) value: TypedValue<'a>,
+}
+
+impl<'a> SchemaAwareResponse<'a> {
+    pub(crate) fn zip(
```
BrynCooke, line 19: nit, maybe just call this `new`

BrynCooke (approved):

PR #4904, opened 2024-04-03
ignore alternative authorization header prefixes

Geal:
Follow up to https://github.com/apollographql/router/pull/4718

This PR enables the JWT plugin to ignore other auth prefixes other than the one defined in the configuration.

This enables multiple Authorization schemes to be supported, although other schemes would need to be handled in either Rhai or a coprocessor to be used with the AuthN plugin.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


docs/source/configuration/authn-jwt.mdx
```diff
@@ -139,7 +139,7 @@ The default value is `Authorization`.
 
 The string that will always precede the JWT in the header value corresponding to [`header_name`](#header_name). This value must not include whitespace.
 
-The default value is `Bearer`.
+The default value is `Bearer`. If the router encounters authorization headers with a different prefix, it will ignore them.
```
lleadbet, line 142: The current default does the opposite, no? It will reject the request as far as I can tell.

lleadbet: Hmm. I would guess they may use it as an expected behavior to fail if it isn't formatted correctly, including overall header structure. I can reach out to other SAs to get a gut check.

.changesets/feat_lleadbet_ignore_other_auth_prefixes.md
```diff
@@ -0,0 +1,17 @@
+### Ignore other auth prefixes in the JWT plugin
+
+If the router encounters an authorization header with a different prefix in the value than what it expects, it will now ignore it. If the router was configured without the `require_authentication` option or without the authorization directives, then some requests that came with a different header prefix that were rejected before will now go through the router. If those options were configured, then there will be no change in behaviour.
```
garypen, line 3: Attempted clarification, but make sure it still makes sense.

```suggestion
If the router encounters an authorization header with a different prefix to the configured value, it will now ignore it.

If the router is configured without the `require_authentication` option or without the authorization directives, then requests that have a different header prefix, that were previously rejected, will now go through the router.

If both of those options were configured, then there will be no change in behaviour.
```

.changesets/feat_lleadbet_ignore_other_auth_prefixes.md
```diff
@@ -0,0 +1,17 @@
...
+
+As an example, with a router configure like this:
```
garypen, line 5:
```suggestion
As an example, with a router configured like this:
```

.changesets/feat_lleadbet_ignore_other_auth_prefixes.md
```diff
@@ -0,0 +1,17 @@
...
+
+```yaml title="router.yaml"
+authentication:
+  router:
+    jwt:
+      header_name: authorization
+      header_value_prefix: "Bearer"
+```
+
+In the above, the router will ignore `Authorization: Basic <token>`, but process requests with `Authorization: Bearer <token>` defined.
```
garypen, line 15:
```suggestion
In the above, the router will ignore requests with a `Authorization: Basic <token>` header and process requests with a `Authorization: Bearer <token>` header.
```

garypen (commented):
I think the change in behaviour is ok as long as the docs make the change clear.

PR #4915, opened 2024-04-04
[batching] first pass at fit and finish for production

garypen:
A lot of changes mainly geared around improving production stability. In no particular order:
    
     - Keep a reference to the shared Batch in each BatchQuery
     - Introduce SubgraphBatchingError and use in many places
     - set_query_hashes now returns a Result
     - so does signal_cancelled
     - and signal_progress
     - allow our spawned task to return an error
     - remove many redundant TODO messages
     - Don't break loops on error, but continue looping and report all
       errors
     - Eliminate all expect/unwrap in non test code
     - Store the batch size in the Batch (may still remove that...)
     - assemble_batch() returns a Result
     - fix tests so they don't require TEST_APOLLO_KEY
     - Introduce BatchInfo to make map_filter easier to read in
       process_batches()
     - Process assemble_batch() errors in process_batches()
     - Enforce info.len() == txs.len()
     - Add CacheResolverError::BatchingError (and associated code)


apollo-router/src/services/subgraph_service.rs
```diff
@@ -890,12 +890,15 @@ pub(crate) async fn notify_batch_query(
             for tx in senders {
                 // TODO: What should we do if a single send fails? If we error out then none of the other
                 // potentially valid senders will receive their results.
-                tx.send(Err(Box::new(e.clone()))).map_err(|error| {
+                // Try to notify all waiters. If we can't notify a single sender, then log an error
```
garypen, line 893: If we can't notify and individual sender

apollo-router/src/batching.rs
```diff
@@ -178,15 +175,21 @@ pub(crate) struct BatchQueryInfo {
...
     /// A sender channel to communicate with the batching handler
     senders: Mutex<Vec<Option<mpsc::Sender<BatchHandlerMessage>>>>,
 
     /// The spawned batching handler task handle
     ///
-    /// Note: This _must_ be aborted or else the spawned task may run forever.
-    spawn_handle: JoinHandle<()>,
+    /// Note: We keep this as a failsafe. If the task doesn't terminate _before_ the batch is
+    /// dropped, then we will abort() the task on drop.
+    // spawn_handle: JoinHandle<()>,
```
nicholascioli, line 187: Remove dead code

apollo-router/src/batching.rs
```diff
@@ -178,15 +175,21 @@ pub(crate) struct BatchQueryInfo {
...
+    spawn_handle: JoinHandle<Result<(), BoxError>>,
+
+    /// What is the size (number of input operations) of the batch?
+    #[allow(dead_code)]
```
nicholascioli, line 191: Do we want this to be allowed if nothing is using it?

apollo-router/src/batching.rs
```diff
@@ -353,51 +348,57 @@ impl Batch {
...
             remaining: 0,
+            batch,
         }
     }
 }
 
 impl Drop for Batch {
     fn drop(&mut self) {
-        // Make sure that we kill the background task if the batch itself is dropped
+        // Failsafe: nake sure that we kill the background task if the batch itself is dropped
```
nicholascioli, line 382:
```suggestion
        // Failsafe: make sure that we kill the background task if the batch itself is dropped
```

apollo-router/src/batching.rs
```diff
@@ -406,22 +407,18 @@ pub(crate) async fn assemble_batch(
     let (requests, gql_requests): (Vec<_>, Vec<_>) = request_pairs.into_iter().unzip();
 
     // Construct the actual byte body of the batched request
-    let bytes = hyper::body::to_bytes(
-        serde_json::to_string(&gql_requests).expect("JSON serialization should not fail"),
-    )
-    .await
-    .expect("byte serialization should not fail");
+    let bytes = hyper::body::to_bytes(serde_json::to_string(&gql_requests)?).await?;
```
nicholascioli, line 407: Is there a case where byte serialization could fail if the input is valid JSON?

apollo-router/src/batching.rs
```diff
@@ -478,7 +474,9 @@ mod tests {
             .unzip();
 
         // Assemble them
-        let (op_name, _context, request, txs) = assemble_batch(requests).await;
+        let (op_name, _context, request, txs) = assemble_batch(requests)
+            .await
+            .expect("it can assemble a batch");
```
nicholascioli, line 476: If we expect this to never fail, should `assemble_batch` not return a result?

apollo-router/src/services/router/service.rs
```diff
@@ -757,7 +757,8 @@ impl RouterService {
             context
                 .extensions()
                 .lock()
-                .insert(shared_batch_details.query_for_index(0));
+                // .insert(shared_batch_details.query_for_index(0));
```
nicholascioli: Remove dead code

apollo-router/src/services/subgraph_service.rs
```diff
@@ -890,12 +890,15 @@ pub(crate) async fn notify_batch_query(
             for tx in senders {
                 // TODO: What should we do if a single send fails? If we error out then none of the other
```
nicholascioli, line 891: Should this comment be removed now that we handle errors?

apollo-router/src/services/subgraph_service.rs
```diff
@@ -917,39 +920,69 @@ pub(crate) async fn notify_batch_query(
...
         futures::future::join_all(svc_map.into_iter().map(|(service, requests)| async {
-            let (_op_name, context, request, txs) = assemble_batch(requests).await;
+            let (_op_name, context, request, txs) = assemble_batch(requests).await?;
 
-            ((service, request, context, txs.len()), txs)
+            Ok(((service, request, context, txs.len()), txs))
         }))
         .await
         .into_iter()
+        .filter_map(|x: Result<BatchInfo, BoxError>| x.map_err(|e| errors.push(e)).ok())
```
nicholascioli, line 962: This is a super nit, but do we want this lambda to have side-effects?

apollo-router/src/services/subgraph_service.rs
```diff
@@ -917,39 +920,69 @@ pub(crate) async fn notify_batch_query(
...
+            "assembling batches failed".to_string(),
+        )
+        .into());
+    }
     // Collect all of the processing logic and run them concurrently, collecting all errors
-    // TODO: Is it ok to panic if the length of the txs and responses does not match? It seems like that is literally impossible
     let cf = &client_factory;
+    // It is not ok to panic if the length of the txs and info do not match. Let's make sure they
+    // do
+    if txs.len() != info.len() {
```
nicholascioli, line 980: I think that this might be literally impossible since we're using `unzip` to construct this, but I'm OK with leaving this if we want to be extra sure.

nicholascioli (changes requested):
I'm a little worried that if we have the test keys enforced only when checking the result, then it might fail when running it locally without those keys since the feature requires a key.

Besides that, I left a few nits and suggestions!

apollo-router/src/batching.rs
garypen: For now. I'm going to do another pass before starting code review and I'm trying to make my mind up if this is useful (for debug logs) or not. I feel like it might be useful to have it appear in the `debug` representation of a `Batch`, but `rustc` can't work that out.

garypen:
> I'm a little worried that if we have the test keys enforced only when checking the result, then it might fail when running it locally without those keys since the feature requires a key.
> 

I couldn't think of a good way of enforcing this, so I looked at what file upload does and it, effectively, does the same thing. It's a bit more fiddly here, because we return a `Vec<Response>` so I needed to come up with a slightly more intrusive solution.

If you run `cargo test` without a key you get a whole load of failures and I find that more intrusive. Also: I don't understand this comment:
```
Note: The [IntegrationTest] ensures that these test credentials get
            // set before running the router.
```

I don't get that behaviour, I just get a bunch of test fails.

> Besides that, I left a few nits and suggestions!

apollo-router/src/batching.rs
garypen: Not sure I understand. This is a test, so if it does fail, the test will fail and we have an issue to resolve.

apollo-router/src/services/subgraph_service.rs
garypen: It's the only way I can think of to handle both errors and oks. Alternative approaches that achieve that are fine by me.

apollo-bot2:

PR #4930, opened 2024-04-09
align coprocessor metric creation

Geal:
There are currently 2 different ways to create metrics, with slight incompatibilities. This makes sure that the execution stage coprocessor metrics are generated in the same way as the other stages

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


BrynCooke (changes requested):
Sorry, we're trying to move away from the tracing macros for metrics. They cause a bunch of issues around types, they are slow and also they can't be unit tested. Can we move to the new macros?

Geal:
the issue appeared because I tried to move to the new macros at the exectution stage. Do you want me to update the other stages instead?

BrynCooke:
Yeah let's try and yeet the other metrics.

It's generally bad form to have multiple points where a metric is updated as with both the macros there is a risk of collision. The only reason it didn't hit us harder is that the old tracing macros didn't fill out the metric descriptions. In other areas of the code we also had issues with double counting.

Ideally we pull out a function that updates the metric and use this in all places in this plugin or alternatively use the regular meter provider to create the meter once and store it in the plugin for reuse at all current callsites.

PR #4815, opened 2024-03-16
Initial QP integration

SimonSapin:
(This builds on top of https://github.com/apollographql/router/pull/4551, and so uses it as a target branch for now.)

Still to do:

* Conversion from `apollo_federation` query plan data structures to Router ones
* [Formatted query plan](https://github.com/apollographql/federation/blob/%40apollo/query-planner%402.7.1/query-planner-js/src/prettyFormatQueryPlan.ts) (FED-118)
* [Usage reporting](https://github.com/apollographql/federation-rs/blob/router-bridge%40v0.5.16%2Bv2.7.1/federation-2/router-bridge/js-src/plan.ts#L134) (FED-117)

---

**Checklist**

Complete the checklist (and note appropriate exceptions) before the PR is marked ready-for-review.

- [ ] Changes are compatible[^1]
- [ ] Documentation[^2] completed
- [ ] Performance impact assessed and acceptable
- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[^1]: It may be appropriate to bring upcoming changes to the attention of other (impacted) groups. Please endeavour to do this before seeking PR approval. The mechanism for doing this will vary considerably, so use your judgement as to how and when to do this.
[^2]: Configuration is an important part of many changes. Where applicable please try to document configuration examples.
[^3]: Tick whichever testing boxes are applicable. If you are adding Manual Tests, please document the manual testing (extensively) in the Exceptions.

PR #4931, opened 2024-04-09
Enable Rust-based API schema generation by default

### LINKED ISSUE #4649: Enable `experimental_api_schema_generation_mode: both` by default

goto-bus-stop:
Fixes #4649

Requires an apollo-federation update to feign support for join spec v0.4. The new join spec doesn't affect how the API schema is generated so the only thing the update does is *not error*. The new release also includes a query plan Display implementation for #4815! 

I ran it through our whole harness testing again, which is quite reliable for API schema as the schema data is complete. The only observable differences are in the sorting of keys in input objects in default values which is both very limited, and doesn't make sense in the JS version, so that's good. The *meaning* of schemas did not change in any case.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -192,9 +192,9 @@ checksum = "5ad32ce52e4161730f7098c077cd2ed6229b5804ccf99e5366be1ab72a98b4e1"
 
 [[package]]
 name = "apollo-compiler"
-version = "1.0.0-beta.14"
+version = "1.0.0-beta.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "00ed7af048c0beb66c9201c032b25a81b1b450397ddb2cb277ca57bcae2d9e13"
+checksum = "79df4ab329753d36476653850519fe92d1b34854dd4337f6abf42a64963ac0ce"
 dependencies = [
  "apollo-parser",
  "ariadne",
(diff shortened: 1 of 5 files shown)


Cargo.lock
```diff
@@ -220,28 +220,28 @@ dependencies = [
...
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "45c675747dd20db0f124d07b9764265b3ae67afbdd1044345673c184888cd018"
+checksum = "94e3b0774618a4febe307d2ace6714583e13cd7948cdadb2b4a937ccdc166333"
 dependencies = [
  "apollo-compiler",
  "derive_more",
  "indexmap 2.2.3",
  "lazy_static",
  "petgraph",
- "salsa",
```
goto-bus-stop, line 232: it was already unused but still listed in deps.

SimonSapin (approved):

o0Ignition0o (approved):

garypen (approved):

Geal (approved):

PR #4935, opened 2024-04-10
helm chart - make .Values.router optional

### LINKED ISSUE #4781: Helm chart - cannot unset values.router.configuration safely
**Describe the bug**
I would like to provide my own config.yaml file into the container, and in order to do so i have to make `router.configuration` empty so it will skip setting `--config /app/configuration.yaml`.

The problem is that if I set this value to empty, it will fail on service creation because of this line:
`index .Values.router.configuration "health_check").listen`

With the following error:
`Error: template: router/charts/router/templates/service.yaml:21:31: executing "router/charts/router/templates/service.yaml" at <index .Values.router.configuration "health_check">: error calling index: cannot index slice/array with type string`

**To Reproduce**
set this chart as a depedency of a new chart you create an run `helm template . --set router.router.configuration={}`

**Expected behavior**
should unset configuration safely and let us load our own config file from file system.

dan-catalano-vc:
- discussion #4934 (written by us) 
- get liveness & readiness port from containerPorts
- check if `.Values.router` exists before checking if `.Values.router.configuration` exist
- remove `.Values.router.configuration` & `.Values.router.args` from default `values.yaml` 

*Description here*

Fixes #4781 (found existing issue when looking into problem)

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/helm/chart/router/templates/configmap.yaml
+++ b/helm/chart/router/templates/configmap.yaml
@@ -1,4 +1,4 @@
-{{- if .Values.router.configuration }}
+{{- if and .Values.router .Values.router.configuration }}
 {{- $routerFullName := include "router.fullname" .  -}}
 {{/* NOTE: metrics configuration moved under telemetry.exporters in Router 1.35.0 */}}
 {{- $configuration := dict }}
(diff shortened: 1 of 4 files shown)


garypen:
For anyone else reading this PR: I've put some feedback on the linked discussion.

garypen:
There's been no further activity in the [linked discussion](https://github.com/apollographql/router/discussions/4934) so I'm closing this.

PR #4940, opened 2024-04-10
add sha256 support to rhai

### LINKED ISSUE #4939: Add Rhai sha256 hashing function
## Is your feature request related to a problem? Please describe.**
We want to hash error codes and custom values in the request/response for logging purposes. SHA256 is also used with Persisted Queries and caching so being able to generate the same values in Rhai could help with custom scripting

Base64 encoding doesn't help because anyone who has access to the encoded messages can easily decode them vs sha256 hashing which is one way

## Describe the solution you'd like**
Expose new global functions in rhai, similar to `base64::*`, but for [sha256](https://docs.rs/sha256/latest/sha256/)

```rhai
let text = "alice and bob";
let hash = sha256::digest(text);

print(hash) # 629f44ee052437b149370d37b9810e19a1df618712ccae63593d37c9b96e3237
```

## Describe alternatives you've considered**
We could use a coprocessor to solve this today

## Additional context**
This was brought up in a larger issue as one step to support encryption in Rhai in https://github.com/apollographql/router/issues/2025 but that issue was closed because we added native Sigv4 support. This issue was also opened just for the sha256 function, but it was closed as a duplicate of the larger crypto issue: https://github.com/apollographql/router/issues/2112

Outside of crypto support it would be helpful to have a sha256 hashing function available so we don't have to use a coprocessor for small request/response manipulation

lleadbet:
This adds a new `sha256::digest` function to Rhai to enable hash values. 

Fixes #4939 

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests


apollo-router/src/plugins/rhai/engine.rs
```diff
@@ -173,11 +173,22 @@ mod router_json {
...
+        hex::encode(hash)
+    }
+}
+
 #[export_module]
 mod router_expansion {
     pub(crate) type Expansion = expansion::Expansion;
 
-    #[rhai_fn(name = "get", return_raw)]
+    #[rhai_fn(name = "digest", return_raw)]
```
garypen, line 191: Please revert this change.

docs/source/customizations/rhai-api.mdx
```diff
@@ -236,6 +236,19 @@ fn supergraph_service(service) {
     }
 }
 ```
+
+## sha256 hash strings
+
+Your Rhai customization can use the function `sha256::digest()` and `base64::decode()` to hash strings using the SHA256 hashing algorithm.
```
garypen, line 242:
```suggestion
Your Rhai customization can use the function `sha256::digest()` to hash strings using the SHA256 hashing algorithm.
```

docs/source/customizations/rhai-api.mdx
```diff
@@ -236,6 +236,19 @@ fn supergraph_service(service) {
...
+fn supergraph_service(service){
+    service.map_request(|request|{
+        let sha = sha256::digest("hello world");
+        log_info(sha);
+    });
+}
+```
 <Note>
 
 You don't need to import the "base64" module. It is imported in the router.
```
garypen, line 254: I think this should be at line 239. and this line should be:
```
You don't need to import the "sha256" module. It is imported in the router.
```

garypen (commented):
Thanks for the PR. It looks good, just a few little NITS to sort out.

apollo-router/src/plugins/rhai/engine.rs
lleadbet: I have no idea why I had done this- probably working in the wrong spot. fixed.

garypen (commented):
I think we've lost a `</Note>` somewhere. I think it should be under the base64 section.

docs/source/customizations/rhai-api.mdx
```diff
@@ -240,6 +240,22 @@ fn supergraph_service(service) {
...
```
garypen, line 242: I think there's a missing </Note> here.

lleadbet: gah, added.

.changesets/feat_feat_sha256_in_rhai.md
```diff
@@ -0,0 +1,17 @@
+### Brief but complete sentence that stands on its own ([Issue #ISSUE_NUMBER](https://github.com/apollographql/router/issues/4939))
```
bnjjj, line 1: Bad title

lleadbet: fixed

o0Ignition0o (approved):

PR #4946, opened 2024-04-12
Fix error: failed to select a version for `router-bridge`

SimonSapin:
router-bridge was added as a dendency of router-fuzz in a recent PR, but the dependency in apollo-router had been upgraded in the meantime. This switches to specify common dependencies in the workspace, so that version numbers are de-duplicated.

Note that `Cargo.lock` is not changed by this PR.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


Cargo.toml
```diff
@@ -41,3 +41,34 @@ incremental = false
...
+async-trait = "0.1.77"
+http = "0.2.11"
+once_cell = "1.19.0"
+reqwest = { version = "0.11.24", default-features = false, features = [
+    "rustls-tls",
+    "rustls-native-certs",
+    "gzip",
+    "json",
+    "stream",
+    "blocking",            # in router-fuzz
```
Geal, line 60: could this feature be added only in router-fuzz?

garypen (approved):

Cargo.toml
```diff
@@ -41,3 +41,34 @@ incremental = false
 [profile.release-dhat]
 inherits = "release"
 debug = 1
+
+# Dependencies used in more than one place are specified here in order to keep versions in sync:
+# https://doc.rust-lang.org/cargo/reference/workspaces.html#the-dependencies-table
```
abernix, line 46: cool!

Cargo.toml
SimonSapin: It turns out that yes. I’ve pushed another commit.

Geal (approved):

PR #4948, opened 2024-04-12
Experimental integration with the new Rust query planner

SimonSapin:
This starts the integration into the Router of the new query planner, being ported from TypeScript to Rust. Experimental configuration opts into running the new planner, or running both and checking that they return the same query plan. The latter is an important part of our testing strategy, to ensure correctness of the port.

**Note:** the new planner is not yet ready for testing at this point.

Nothing changes for the default configuration.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


.changesets/exp_simon_fed_112_rust_qp.md
```diff
@@ -0,0 +1,12 @@
+### Experimental integration with the new Rust query planner
```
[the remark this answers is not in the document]
SimonSapin: The note plus the config key not being specified in the changelog somewhat nudge towards not trying it yet, but I’m ok with no changelog too. Reverted.

BrynCooke (changes requested):
Please use the new metrics macros rather than the tracing layer.

apollo-router/src/spec/schema.rs
```diff
@@ -20,30 +20,23 @@ use crate::query_planner::OperationKind;
 use crate::Configuration;
 
 /// A GraphQL schema.
-#[derive(Debug)]
 pub(crate) struct Schema {
     pub(crate) raw_sdl: Arc<String>,
-    pub(crate) definitions: Valid<apollo_compiler::Schema>,
+    pub(crate) federation_supergraph: apollo_federation::Supergraph,
```
Geal, line 25: so this is not just an experimental integration that can be switched on without affecting the rest, it directly modifies core code?

SimonSapin: Yes, as the PR is now fed-next metadata is extracted from the supergraph unconditionally. The default config makes it not used (except for the apollo-compiler schema it contains) but still created.

If we judge the risk of spurious failure in fed-next initialization is too great, this field could be turned into an enum that contains either a `Schema` or a `Supergraph`

goto-bus-stop: hmm, we do use the `@link()` handling in API schema, so basic use cases should be okay. I'm not sure if we are 100% confident in it though.

it would be really nice to have a supergraph available because there are a few places that mimick `@link` functionality in the router

apollo-router/src/spec/query/tests.rs
```diff
@@ -5900,7 +5899,7 @@ fn filtered_defer_fragment() {
         &mut response,
         None,
         Object::new(),
-        &schema,
+        schema.api_schema(),
```
Geal, line 5902: ah, nice catch. Making a separate type for the API schema is a great idea to uncover issues like these

Geal (approved):
approved, with one demand: we let this bake for a while in dev, and everybody must test with new, legacy and both to make sure we find issues quickly

apollo-router/src/spec/schema.rs
Geal: we have some supergraphs in https://github.com/apollographql/router/tree/dev/examples/graphql

apollo-router/src/spec/query/tests.rs
SimonSapin: Yes, renaming Schema::definitions() to supergraph_schema() tries to go in the same direction

apollo-router/src/plugins/telemetry/tracing/apollo_telemetry.rs
```diff
@@ -533,8 +533,8 @@ impl Exporter {
                         .unwrap_or_default()
                         .iter()
                         .map(|d| DeferredNodeDepends {
-                            id: d.id.clone(),
-                            defer_label: d.defer_label.clone().unwrap_or_default(),
+                            id: d.id.to_string(),
+                            defer_label: "".to_owned(),
```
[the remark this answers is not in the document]
SimonSapin: This PR removes the `defer_label` field of the `apollo_router::query_planner::plan::Depends` struct because it was always `None`. It is never set in the JS planner, and does not exist in the corresponding fed-next struct. (The label is present in `DeferredNode::label`)

apollo-router/src/configuration/mod.rs
```diff
@@ -502,6 +550,15 @@ impl Configuration {
             }
         }
 
+        if self.experimental_query_planner_mode == QueryPlannerMode::New
+            && self.experimental_apollo_metrics_generation_mode != ApolloMetricsGenerationMode::New
+        {
+            return Err(ConfigurationError::InvalidConfiguration {
+                message: "`experimental_query_planner_mode: new` requires `experimental_apollo_metrics_generation_mode: new`",
+                error: "either change to some other query planner mode, or change to new metrics generation".into()
```
lrlna, line 558: oooo `experimental_apollo_metrics_generation_mode` will probably not be out of "experimental" by June 17th, eh @BrynCooke? I added this as a note to keep an eye out when writing docs for our June 17th release.

goto-bus-stop (changes requested):
change requesting just to make my concern explicit: https://github.com/apollographql/router/pull/4948/files#r1572329511

I think we should not unconditionally create `Supergraph` instances until we have good feedback from the API schema generation, which will be exercising that code. It hasn't caused any issues with the corpus so the risk is very low and I can maybe be convinced by other opinions but this is what I think right now

goto-bus-stop (approved):

BrynCooke (approved):

PR #4951, opened 2024-04-15
Signature and reference generation tweaks

bonnici:
Follow-up from https://github.com/apollographql/router/pull/4796 which addresses some performance concerns and non-blocking issues from the PR.

Related to https://apollographql.atlassian.net/browse/FED-117.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


bonnici:
@Geal these are some tweaks based on comments that you brought up in the original PR https://github.com/apollographql/router/pull/4796. I'm still not sure why the test isn't being run after I moved it to a separate file so any leads you can give me there would be appreciated.

bnjjj (approved):

garypen (approved):

apollo-router/src/apollo_studio_interop/test.rs
```diff
@@ -0,0 +1,1419 @@
+#[cfg(test)]
+mod tests {
```
garypen, line 2: Delete these two lines and the matching module close at the end. They are already in module `tests` by virtue of the file name.

apollo-router/src/apollo_studio_interop/mod.rs
```diff
@@ -523,1423 +523,3 @@ fn format_value(value: &Value, f: &mut fmt::Formatter) -> fmt::Result {
         rest => f.write_str(&rest.to_string()),
     }
 }
-
-#[cfg(test)]
-mod tests {
```
garypen, line 528: Retain these two lines to include the `tests` module for testing.
e.g.:
```
#[cfg(test)]
mod tests;
```

garypen: also, you need to rename `test.rs` to `tests.rs`.

garypen (commented):
To answer your questions about "why isn't the test working".

You need to make a couple of tweaks to your PR to comply with rust's module expectations.

apollo-router/src/apollo_studio_interop/mod.rs
bonnici: Thank you! I tried some of these but not at the same time and I also was always missing the #[cfg(test)]; mod tests;` part.

Geal (approved):

garypen (approved):

o0Ignition0o (approved):

Issue #4910, opened 2024-04-04
axum_factory::tests::test_supergraph_timeout should not use a hardcoded port number


garypen:
It causes confusion when the test hits the "wrong" router.

PR #4955, opened 2024-04-15
Make test_supergraph_timeout more robust

SimonSapin:
* Use an available port number
* Add an artificial delay to make sure the short timeout is triggered

The latter is an attempt to fix the test sometimes returning HTTP 200 instead of the expect 504 Gateway Timeout. I suspect either that the future wrapped in timeout would return `Ready` immediately (our `plugins::traffic_shaping::timeout` only polls the `sleep` future when the main future returns `Pending`), or the normal flow would complete within a kernel tick (which is likely much less frequent than 1ns), so `sleep` would measure zero elapsed time.

Fixes https://github.com/apollographql/router/issues/4910

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/axum_factory/tests.rs
+++ b/apollo-router/src/axum_factory/tests.rs
@@ -69,6 +69,7 @@ use crate::query_planner::BridgeQueryPlannerPool;
 use crate::router_factory::create_plugins;
 use crate::router_factory::Endpoint;
 use crate::router_factory::RouterFactory;
+use crate::services::execution;
 use crate::services::layers::persisted_queries::PersistedQueryLayer;
 use crate::services::layers::query_analysis::QueryAnalysisLayer;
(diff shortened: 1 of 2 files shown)


SimonSapin:
CI failed on ARM because the test modified here got HTTP 200 instead of the expected 504 Gateway Timeout: https://app.circleci.com/pipelines/github/apollographql/router/20434/workflows/7d0645be-b7ed-4e96-a11a-a0e015dd6491/jobs/137701?invite=true#step-116-101483_61

I’m not sure what to make of this

bnjjj (approved):

Geal (approved):

PR #4956, opened 2024-04-15
feat(telemetry): add support of custom events in telemetry

### LINKED ISSUE #4320: Custom events defined by yaml
Allow users to configure events via yaml as documented in https://www.apollographql.com/docs/router/configuration/telemetry/instrumentation/events
This will allow users to avoid reaching for rhai or a custom plugin when they want to log that something has happened e.g. a request had errors of a particular type.

bnjjj:
Fixes #4320

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [x] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/lib.rs
```diff
@@ -90,6 +90,7 @@ pub use crate::context::Context;
 pub use crate::executable::main;
 pub use crate::executable::Executable;
 pub use crate::notification::Notify;
+pub use crate::plugins::telemetry::otel;
```
Geal, line 93: why is it public?

apollo-router/src/plugins/telemetry/config_new/events.rs
```diff
@@ -11,24 +22,348 @@ use crate::plugins::telemetry::config_new::extendable::Extendable;
...
+                .clone()
+                .into_iter()
+                .filter_map(|(name, val)| Some((name?.to_string(), val)))
+                .collect();
+            #[cfg(test)]
+            headers.sort_keys();
+            #[cfg(not(test))]
+            let headers = request.router_request.headers();
+
+            attrs.insert("http.request.headers".to_string(), format!("{:?}", headers));
```
Geal, line 169: do we have to do that for every request if events are active, even if the configuration does not ask for the headers?

Geal (approved):
this looks very useful! We should make sure that we do not steer users too easily towards the full request or response dumps because they will be very expensive.
Could the subgraph events be activated per subgraph service name too? Instead of only on all subgraphs

apollo-router/src/plugins/telemetry/config_new/events.rs
bnjjj: no, only when you enabled `events.x.request|response` by default it's disabled

apollo-router/src/lib.rs
bnjjj: Hmm I think that's a mistake, I wanted to get rid of tracing_opentelemetry even for tests but for now I'll just keep it, maybe I'll remove it later but that's ok. the otel module is no longer pub. Good call

bnjjj:
@Geal Yes we will need to add these disclaimer in our docs. I think @shorgi will help me to find the right way to achieve that. 

I also created an issue for your suggestion https://github.com/apollographql/router/issues/4984 @Geal

BrynCooke (approved):
I played around with this quite a bit. It's great!

PR #4957, opened 2024-04-15
demand control plugin initial integration

BrynCooke:
This provides an initial implementation for the demand control plugin. It is not complete, but has most of the logic in place.

Changes made:
* Config is now centred around a `Strategy` rather than `Algorithm`. Each strategy may have it's own config.
* On plugin init a strategy factory is created. This stores all the information that is needed to create a concrete strategy for a request.
* `Strategy` is responsible for keeping track of costs and deciding if a request should be rejected or aborted.
* `Strategy` reacts to execution_request/response and subgraph_request/response.
* Demand control may be run in either `measure` or `enforce` mode. Measure will not reject/abort requests. Enforce will reject/abort requests.
* Plugin `execution_service` and `subgraph_service` implemented.
* Tests added.  

Things to follow up on:
* `ExecutableDocument` has been created for subgraph requests is not currently being populated.
* `ExecutableDocument` needs to be populated in the query plan to prevent reparsing.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


tninesling (approved):

apollo-router/src/plugins/demand_control/mod.rs
```diff
@@ -100,9 +150,324 @@ impl Plugin for DemandControl {
...
+                        .get::<Strategy>()
+                        .expect("must have strategy")
+                        .clone();
+                    resp.response = resp.response.map(move |resp| {
+                        // Here we are going to abort the stream if the cost is too high
+                        // First we map based on cost, then we use take while to abort the stream if an error is emmitted.
+                        // When we terminate the stream we still want to emit a graphql error, so the error response is emitted first before a termination error.
+                        resp.flat_map(move |resp| {
+                            match strategy.on_execution_response(req.as_ref(), &resp) {
+                                Ok(_) => stream::iter(vec![Ok(resp)]),
```
Geal, line 190: the `flat_map` then iter of a single element vec feels weird, wasn't there another solution? https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html#method.then maybe?

apollo-router/src/plugins/demand_control/strategy/mod.rs
```diff
@@ -0,0 +1,125 @@
...
+    ) -> Self {
+        Self {
+            config,
+            supergraph_schema,
+            subgraph_schemas,
+        }
+    }
+
+    pub(crate) fn create(&self) -> Strategy {
+        let strategy: Arc<dyn StrategyImpl> = match &self.config.strategy {
```
Geal, line 93: why do a dyn trait? This could be an enum

apollo-router/src/plugins/demand_control/mod.rs
BrynCooke: I don't think this is possible without using the `flat_map`. We have to generate an extra element in the stream for us to be able to use `take_while` for termination.
Without the extra element we can't emit an error to the caller as `take_while` will terminate the stream immediately.

We could write our own layer, but would prefer to see if there is a perf issue with the existing constructs before doing that.

I tried using some other constructs for the single vec e.g. std::iter::once., but the issue would be that we would have to box to obtain the same typeing on the match arms.

apollo-router/src/plugins/demand_control/strategy/mod.rs
BrynCooke: I started off with that pattern but wanted to avoid the enum impl getting big. Each strategy should be in a separate module. I started pulling stuff out and having a match arm in the enum that called these separate modules, but after a while I realised I was reimplementing dynamic dispatch.

apollo-router/src/plugins/demand_control/mod.rs
BrynCooke: [edeade8](https://github.com/apollographql/router/pull/4957/commits/edeade88f742c676829e99ae977c73591a835f67)

PR #2489, opened 2023-01-27
Supports federation 2.3 in the router

pcmanus:
The query planner in Federation 2.3 adds a new concept of "data rewrites" for fetches in order to support both `@interfaceObject` and to support the fix for https://github.com/apollographql/federation/issues/1257.

Those "rewrites" describe simple updates that need to be performed either on the inputs (the "representations" passed to `_entities`; need to rewrite the `__typename` when sending queries to an `@interfaceObject`) or the output of a fetch (needed when a field has been aliased to permit the subgraph query to be valid, but that field needs to be "un-aliased" to its original name after the fetch).

This commit implements those rewrites.

**Checklist**

Complete the checklist (and note appropriate exceptions) before a final PR is raised.

- [ ] Changes are compatible[^1]
- [ ] Documentation[^2] completed
- [ ] Performance impact assessed and acceptable
- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[^1]. It may be appropriate to bring upcoming changes to the attention of other (impacted) groups. Please endeavour to do this before seeking PR approval. The mechanism for doing this will vary considerably, so use your judgement as to how and when to do this.
[^2]. Configuration is an important part of many changes. Where applicable please try to document configuration examples.
[^3]. Tick whichever testing boxes are applicable. If you are adding Manual Tests:
    - please document the manual testing (extensively) in the Exceptions.
    - please raise a separate issue to automate the test and label it (or ask for it to be labeled) as `manual test`

PR #4958, opened 2024-04-15
Apply alias rewrites to arrays

o0Ignition0o:
[#2489](https://github.com/apollographql/router/pull/2489) introduced automatic aliasing rules to support `@interfaceObject`.

These rules now properly apply to lists.


bnjjj (approved):

Geal (approved):
nice catch, thanks

PR #4975, opened 2024-04-17
Parse fetch node operations into executable docs during planning

tninesling:
In query plan fetch nodes, store a lazily-initialized `ExecutableDocument` in addition to its GraphQL string representation. Both the authorization plugin and demand control plugin were parsing operations for each request. This `ExecutableDocument` is skipped in serde (de)serialization, either when reading a JSON query plan from the JS planner or when using the Redis distributed cache.

Fixes [ROUTER-221](https://apollographql.atlassian.net/browse/ROUTER-221)

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-221]: https://apollographql.atlassian.net/browse/ROUTER-221?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apollo-router/src/query_planner/fetch.rs
```diff
@@ -127,6 +128,9 @@ pub(crate) struct FetchNode {
     // authorization metadata for the subgraph query
     #[serde(default)]
     pub(crate) authorization: Arc<CacheKeyMetadata>,
+
+    #[serde(skip)]
+    pub(crate) executable_document: Arc<apollo_compiler::ExecutableDocument>,
```
SimonSapin, line 133: I believe this `serde(skip)` means that deserializing a query plan from distributed Redis cache will produce an empty document.

I could maybe be fixed by adding calls to `parse_subgraph_operations` in more places, but I think a better design is to have the fetch node itself lazily initialize the document on demand.

I’ll make a pull request that targets this branch to show what I have in mind.

apollo-router/src/query_planner/plan.rs
```diff
@@ -382,6 +383,60 @@ impl PlanNode {
...
+            PlanNode::Defer { primary, deferred } => {
+                if let Some(node) = primary.node.as_mut() {
+                    node.parse_subgraph_operations(schemas);
+                }
+                for deferred_node in deferred {
+                    if let Some(node) = deferred_node.node.take() {
+                        let mut new_node = (*node).clone();
+                        new_node.parse_subgraph_operations(schemas);
+                        deferred_node.node = Some(Arc::new(new_node));
+                    }
```
SimonSapin, line 414: This function will go away entirely with my lazy-initialization proposal above, but this pattern can be improved with [`Arc::make_mut`](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.make_mut) which only clones if the reference count is above 1.
```suggestion
                    if let Some(node) = &mut deferred_node.node {
                        Arc::make_mut(node).parse_subgraph_operations(schemas);
                    }
```

SimonSapin:
Please see https://github.com/apollographql/router/pull/4985 which merges into this PR.

SimonSapin (approved):

PR #4981, opened 2024-04-18
Fix a small formatting issue with Rust signature generation

bonnici:
Fixes an issue with signature generation where commas were sometimes not inserted when they should be.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


goto-bus-stop (approved):
just needs a cargo fmt :)

abernix:
> just needs a cargo fmt :)

I took care of the `cargo xtask lint --fmt` and pushed to this branch. 😄

apollo-router/src/query_planner/bridge_query_planner.rs
```diff
@@ -464,26 +464,23 @@ impl BridgeQueryPlanner {
...
                         ) {
-                            tracing::warn!(
-                                monotonic_counter.apollo.router.operations.telemetry.studio.signature = 1u64,
-                                generation.is_matched = false,
-                                "Mismatch between the Apollo usage reporting signature generated in router and router-bridge"
+                            u64_counter!(
+                                "apollo.router.operations.telemetry.studio.signature",
+                                "The match status of the Apollo reporting signature generated by the JS implementation vs the Rust implementation",
+                                1,
+                                "generation.is_matched" = "false"
```
SimonSapin, line 471: Shouldn’t this use a `false` boolean instead of a `"false"` string? Similarly for three other `is_matched` below.

bonnici: Hmm I'm sure I tried that before and it was causing a build error, but now it's not. I can update it in my next PR.

PR #4987, opened 2024-04-19
add conditions on custom attributes for spans + a new selector for graphql error

### LINKED ISSUE #4336: NewRelic error trace and spans dashboard support
NewRelic uses a non-standard format for tracking error dashboards on traces. It requires that you set a span attribute of `otel.status_code` which it uses to track error counts

https://docs.newrelic.com/docs/more-integrations/open-source-telemetry-integrations/opentelemetry/view-your-data/opentelemetry-errors-page/

Add the ability to configure this property to be set or include it in all spans on either HTTP or GraphQL errors

We could customize an Otel collector between the Router and New Relic, however directly configuring Router to send info directly to NR does work minus this one feature:  https://docs.newrelic.com/docs/more-integrations/open-source-telemetry-integrations/opentelemetry/get-started/opentelemetry-set-up-your-app/#review-settings

bnjjj:
Currently the router does not set otel.status_description on spans. This is a special attribute that will also set the span status.

This means that even responses have errors or http status is not 200 spans are not flagged as an error.

To complicate matters further, other languages usually use exceptions and stack unwinding to set span status, both of which are not available to us in rust.

Example:

```yaml
telemetry:
  instrumentation:
    spans: 
      router: 
        attributes:    
          otel.status_description: 
            static: "there was an error"
            condition:
              any:
              - not:
                  eq:
                  - response_status: code
                  - 200
              - eq:
                - on_graphql_error
                - true
```

This will allow the user to set the description to something if there was an issue.

Suggest implementing using the existing condition logic and add two new selectors. This would check http status and if there are 1 or more graphql errors.

Fixes #4336

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [x] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/plugins/telemetry/config_new/conditional.rs
```diff
@@ -0,0 +1,501 @@
...
+{
+    type Request = Request;
+    type Response = Response;
+
+    fn on_request(&self, request: &Self::Request) -> Option<opentelemetry::Value> {
+        match &self.condition {
+            Some(condition) => match condition.lock().evaluate_request(request) {
+                None => {
+                    if let Some(value) = self.selector.on_request(request) {
+                        *self.value.lock() = value.into();
```
Geal, line 108: we should be careful here, because matching on something that was locked means the lock will be held for the entire length of the match

Geal (approved):

PR #4990, opened 2024-04-19
Initial cost metrics

BrynCooke:
This PR adds selector that will be used for displaying cost metrics on the router.

A separate PR will be raised for query depth and height under a different category.
```yaml
telemetry:
  instrumentation:
    instruments:
      supergraph:
        cost.actual: true
        cost.delta: true
        cost.estimated: true
```

In addition the following selectors are introduced:
* cost.estimated
* cost.actual
* cost.delta
* cost.result

This allow you to attach cost related data to spans instruments or events.

There is a lifecycle issue present in this PR which prevents cost.actual and cost delta from functioning correct that we will tackled separately. This is because supergraph response happens before the response payload has been returned.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/plugins/telemetry/config_new/cost/mod.rs
```diff
@@ -0,0 +1,358 @@
...
+
+    fn on_request(&self, _request: &Self::Request) -> Vec<KeyValue> {
+        Vec::default()
+    }
+
+    fn on_response(&self, response: &Self::Response) -> Vec<KeyValue> {
+        let mut attrs = Vec::with_capacity(4);
+
+        if let Some(true) = self.cost_estimated {
+            if let Some(cost_result) = &response.context.extensions().lock().get::<CostResult>() {
```
tninesling, line 61: Is this lock statement repeated just to keep the critical sections small?

apollo-router/src/plugins/telemetry/config_new/cost/mod.rs
```diff
@@ -0,0 +1,358 @@
...
+#[serde(deny_unknown_fields, rename_all = "snake_case")]
+pub(crate) enum CostValue {
+    /// The estimated cost of the operation using the currently configured cost model
+    Estimated,
+    /// The actual cost of the operation using the currently configured cost model
+    Actual,
+    /// The delta between the estimated and actual cost of the operation using the currently configured cost model
+    Delta,
+    /// The result of the cost calculation. This is the error code returned by the cost calculation.
+    Result,
```
tninesling, line 255: Nit: the naming of this is a bit confusing to me because it's very similar `CostResult` and `CostResult.result`. Is there a case where the calculation succeeds and this `Result` is returned? Or is this purely for error values?

tninesling (approved):

apollo-router/src/plugins/telemetry/config_new/cost/mod.rs
BrynCooke: [a28b9ce](https://github.com/apollographql/router/pull/4990/commits/a28b9cefa3dfb1dc3c7ea281a75abbfae1ded8cd)

BrynCooke: Renamed

apollo-router/src/plugins/telemetry/config_new/cost/mod.rs
BrynCooke: Oops, rookie mistake: [c9ca854](https://github.com/apollographql/router/pull/4990/commits/c9ca854a6837c890b19fee0e2c120ecd68802120)

PR #4994, opened 2024-04-22
fix versions for cargo-about, cargo-deny, cargo-edit

Geal:
This fixes the version number for `cargo-about` , `cargo-deny` and `cargo-edit` to make sure that building an old branch in the future will still work

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -350,7 +350,9 @@ commands:
           name: Install cargo deny, about, edit
           command: |
             if [[ ! -f "$HOME/.cargo/bin/cargo-deny$EXECUTABLE_SUFFIX" ]]; then
-              cargo install --locked cargo-deny cargo-about cargo-edit
+              cargo install --locked --version 0.14.21 cargo-deny
+              cargo install --locked --version 0.6.1 cargo-about
+              cargo install --locked --version 0.12.2 cargo-edit
             fi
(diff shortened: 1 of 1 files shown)


goto-bus-stop (approved):

.circleci/config.yml
```diff
@@ -350,7 +350,9 @@ commands:
           name: Install cargo deny, about, edit
           command: |
             if [[ ! -f "$HOME/.cargo/bin/cargo-deny$EXECUTABLE_SUFFIX" ]]; then
-              cargo install --locked cargo-deny cargo-about cargo-edit
+              cargo install --locked --version 0.14.21 cargo-deny
```
abernix, line 353: Silly question, but is there  a way to make this _also_ locked for the person running the tool locally? e.g., a project-specific constraint on the binary toolchain used?

Geal: There's a crate we could use in xtask: https://crates.io/crates/cargo-local-install
But I have not had time to look into it yet.
I suspect the regular changes to licenses.html are due to different versions of cargo-about installed locally

PR #4997, opened 2024-04-22
fix rust version in CI

Geal:
When building an old version of the Router, we have a `rust-toolchain.toml` file that fixes the version used to build, but `rustup.sh` does not actually take that file into account: it installs the last stable, that is then cached, and then when building the router, we install on the fly the required version. This is not a big issue when working on the `dev` branch which always uses the last stable. But for older versions, we should cache the version actually used

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -545,7 +545,7 @@ jobs:
       - checkout
       - setup_environment:
           platform: << parameters.platform >>
-      #- xtask_check_compliance
+      - xtask_check_compliance
 
   test:
     environment:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -320,7 +320,7 @@ commands:
             if [[ ! -d "$HOME/.cargo" ]]; then
               curl https://sh.rustup.rs -sSf -o rustup.sh
               chmod 755 ./rustup.sh
-              ./rustup.sh -y --profile minimal --component clippy --component rustfmt
+              ./rustup.sh -y --profile minimal --component clippy --component rustfmt --default-toolchain $(grep -Po '(?<=^channel = ")[^"]*(?=".*)' rust-toolchain.toml)
             fi
             echo 'export PATH="$HOME/.cargo/bin:$PATH"' >> "$BASH_ENV"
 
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -545,7 +545,7 @@ jobs:
       - checkout
       - setup_environment:
           platform: << parameters.platform >>
-      - xtask_check_compliance
+      #- xtask_check_compliance
```
Geal, line 548:
```suggestion
      - xtask_check_compliance
```

.circleci/config.yml
```diff
@@ -320,7 +320,7 @@ commands:
             if [[ ! -d "$HOME/.cargo" ]]; then
               curl https://sh.rustup.rs -sSf -o rustup.sh
               chmod 755 ./rustup.sh
-              ./rustup.sh -y --profile minimal --component clippy --component rustfmt
+              ./rustup.sh -y --profile minimal --component clippy --component rustfmt --default-toolchain $(grep -Po '(?<=^channel = ")[^"]*(?=".*)' rust-toolchain.toml)
```
Geal, line 323:
```suggestion
              ./rustup.sh -y --profile minimal --component clippy --component rustfmt --default-toolchain $(grep -Eo '(?<=^channel = ")[^"]*(?=".*)' rust-toolchain.toml)
```

SimonSapin:
Rather than parsing `rust-toolchain.toml` with grep, another possibility is installing rustup with `--default-toolchain none` then running `rustc -V` to make rustup read the TOML file itself and install the relevant toolchain. (Or `cargo fetch` if we want to also cache crates.io downloads)

Geal:
@SimonSapin it works, thanks! https://github.com/apollographql/router/pull/4997/commits/4f89cbcc1792a67f95fe950062aa1f22c0b1f533

o0Ignition0o (approved):

bnjjj (approved):

PR #5005, opened 2024-04-23
make the coprocessor URL configurable per stage

Geal:
This makes it possible to have different coprocessor URLs per stage. The initial URL configuration still serves as a global default, but it can be overriden per stage

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/.changesets/feat_geal_coprocessor_urls.md
+++ b/.changesets/feat_geal_coprocessor_urls.md
@@ -12,7 +12,7 @@ coprocessor:
       headers: true
   supergraph::
     request:
-      body; true
+      body: true
       url: http://127.0.0.1:8082
 ```
 
--- a/.changesets/feat_geal_coprocessor_urls.md
+++ b/.changesets/feat_geal_coprocessor_urls.md
@@ -1,0 +1,20 @@
+### make the coprocessor URL configurable per stage ([PR #5005](https://github.com/apollographql/router/pull/5005))
+
+This makes it possible to have different coprocessor URLs per stage. The initial URL configuration still serves as a global default, but it can be overriden per stage.
+
+As an example, we can have a router request and supergraph request coprocessors with different URLs:
+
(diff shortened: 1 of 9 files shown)


apollo-router/src/plugins/coprocessor/supergraph.rs
```diff
@@ -53,6 +55,8 @@ pub(super) struct SupergraphResponseConf {
     pub(super) status_code: bool,
     /// Handles the response without waiting for the coprocessor to respond
     pub(super) detached: bool,
+    /// The url you'd like to offload processing to
+    pub(super) url: Option<String>,
```
bnjjj, line 59:
```suggestion
    #[schemars(with = "String")]
    pub(super) url: Option<url::Url>,
```

bnjjj: To benefit from url check when deserializing

bnjjj (commented):
Missing documentation updates

apollo-router/src/plugins/coprocessor/supergraph.rs
Geal: oh nice

.changesets/feat_geal_coprocessor_urls.md
```diff
@@ -0,0 +1,20 @@
...
+
+```yaml title="router.yaml"
+coprocessor:
+  url: http://127.0.0.1:8081
+  router:
+    request:
+      headers: true
+  supergraph::
+    request:
+      body; true
```
o0Ignition0o, line 15:
```suggestion
      body: true
```

o0Ignition0o (approved):

PR #5008, opened 2024-04-23
Normalize "operation_name" value inside supergraph request

IvanGoncharov:
*Description here*

Fixes #**issue_number**

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


bnjjj (approved):

abernix (changes requested):
I think it may be best if `request.body.operation_name` actually reflects what's passed in, I don't think we need to actually **recommend** people use that, but it's critical that the user have the ability to access it if they needed it.  Let's take an exercise to think through that `request.body.variables` would be here, maybe?  Do we want to change any of those things or keep them as opaque values?  (Again the docs need to be clear about what you should use for a use case!)

But the thing that I think users are trying to accomplish is knowing what operation was executed.  I'm not certain if that needs to be as `rquest.body.operation_name`.  It could just be something else (you tell me — `query.executed_operation_name` or something?)

Geal:
we need some clear documentation on the behaviour, as it's often surprising for customers

PR #5012, opened 2024-04-24
Use subgraph.name attribute not apollo.subgraph.name

garypen:
The DD exporter does some explicit mapping of attributes and was using a value "apollo.subgraph.name" that the latest versions of the router don't use. The correct choice is "subgraph.name".

Update the mapping to reflect this change.


apollo-router/src/plugins/telemetry/tracing/datadog.rs
```diff
@@ -29,7 +29,7 @@ lazy_static! {
         map.insert("request", "http.route");
         map.insert("supergraph", "graphql.operation.name");
         map.insert("query_planning", "graphql.operation.name");
-        map.insert("subgraph", "apollo.subgraph.name");
+        map.insert("subgraph", "subgraph.name");
```
Samjin, line 32: Could have spot this by ourself. Thanks Gray! 👍

PR #5013, opened 2024-04-24
Revert "add conditions on custom attributes for spans + a new selector for graphql error"

BrynCooke:
Reverts apollographql/router#4987

There is an issue with the json schema generation. It seems to affect span custom attribute configuration and prevents custom attributes being configured.

Possible this is an editor thing, but let's not take the chance.


bnjjj (approved):

BrynCooke:
We have a fix in #5015

PR #5016, opened 2024-04-24
Ensure subgraph requests have access to their corresponding ExecutableDocuments

tninesling:
When we apply demand control measurements, we want to measure the cost of an operation at several points. We score the overall operation, the query plan, and each subgraph operation. In order to score these operations, we need to use an ExecutableDocument, which is already provided for the top-level operation but not for the subgraph requests.

In order to achieve this, we could store the ExecutableDocument on each subgraph request. However, this defeats the purpose of having the fetch nodes lazily initialize the documents (and breaks a bunch of query planner tests because they are lacking appropriate subgraph schemas in their setup). To avoid having to define subgraph schemas for all these tests and to preserve the lazy initialization setup, I instead embed a reference to the fetch node that is creating the subgraph request, which can trigger the lazy init when it is being executed.

---

**Checklist**

- Tests added and passing[^3]
    - [X] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


BrynCooke (approved):

apollo-router/src/plugins/demand_control/mod.rs
```diff
@@ -357,11 +368,22 @@ impl Plugin for DemandControl {
...
+                            .ok_or(DemandControlError::QueryParseFailure(
+                                "Cannot parse subgraph request due to missing subgraph schema"
+                                    .to_string(),
+                            ))
+                            .and_then(|(schema, operation)| {
+                                operation
+                                    .as_parsed(schema)
+                                    .map_err(DemandControlError::from)
+                            })
+                            .expect("must have a valid query")
```
BrynCooke, line 385: I think we should convert to an error here rather than using expect. `DemandControlError::QueryParseFailure` could be reused, or we could add something else.

BrynCooke (changes requested):
Let's add some tests for the new code in the demand control plugin.

tninesling:
I'm going to close this for now so we can focus on metrics for the execution service in this plugin. We can revisit this later when we want to actually implement the subgraph service

PR #5019, opened 2024-04-24
Add list size configuration for static demand control

tninesling:
Before we implement a directive to indicate paging in lists, we are using a statically defined assumed size for lists returned in each response. Since there's no real one-size-fits-all value for this, the initial setup will require users to set an assumed list size in the configuration for the static cost estimator.

---

**Checklist**

- Tests added and passing[^3]
    - [X] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-239]: https://apollographql.atlassian.net/browse/ROUTER-239?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apollo-router/src/plugins/demand_control/cost_calculator/static_cost.rs
+++ b/apollo-router/src/plugins/demand_control/cost_calculator/static_cost.rs
@@ -117,7 +117,7 @@ impl StaticCostCalculator {
     }
 
     fn score_fragment_spread(_fragment_spread: &FragmentSpread) -> Result<f64, DemandControlError> {
-        todo!("This needs to read the fragment from the schema and score it")
+        Ok(0.0)
     }
 
     fn score_inline_fragment(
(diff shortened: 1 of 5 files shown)


apollo-router/src/plugins/demand_control/cost_calculator/static_cost.rs
```diff
@@ -26,12 +26,16 @@ use crate::query_planner::Primary;
 use crate::query_planner::QueryPlan;
 
 pub(crate) struct StaticCostCalculator {
+    list_size: u32,
```
BrynCooke, line 29: Let's get a unit test in that exercises two different list sizes.

BrynCooke (changes requested):
Lust need a unit test.

BrynCooke (approved):

PR #5022, opened 2024-04-25
add support of chunk responses for selectors and instrument in telemetry

bnjjj:
Giving the ability to compute new attributes everytime we receive a new chunk in supergraph response. Would be really helpful to create observability for subscriptions and defer.

Naming is not definitive. I need feedback. I would vote for something like `on_event` for traits and for configuration I have no idea, for now it's prefixed with `chunk_` like for example for instrument value it's `chunk_unit` which is not that great IMHO.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/plugins/telemetry/config_new/mod.rs
+++ b/apollo-router/src/plugins/telemetry/config_new/mod.rs
@@ -10,6 +10,7 @@ use super::otel::OpenTelemetrySpanExt;
 use super::otlp::TelemetryDataKind;
 use crate::plugins::telemetry::config::AttributeValue;
 use crate::plugins::telemetry::config_new::attributes::DefaultAttributeRequirementLevel;
+use crate::Context;
 
 /// These modules contain a new config structure for telemetry that will progressively move to
 pub(crate) mod attributes;
@@ -27,17 +28,30 @@ pub(crate) mod spans;
 pub(crate) trait Selectors {
     type Request;
     type Response;
+    type ChunkResponse;
+
     fn on_request(&self, request: &Self::Request) -> Vec<KeyValue>;
     fn on_response(&self, response: &Self::Response) -> Vec<KeyValue>;
+    fn on_chunk_response(&self, _response: &Self::ChunkResponse, _ctx: &Context) -> Vec<KeyValue> {
+        Vec::with_capacity(0)
+    }
     fn on_error(&self, error: &BoxError) -> Vec<KeyValue>;
 }
 
 pub(crate) trait Selector {
     type Request;
     type Response;
+    type ChunkResponse;
 
     fn on_request(&self, request: &Self::Request) -> Option<opentelemetry::Value>;
     fn on_response(&self, response: &Self::Response) -> Option<opentelemetry::Value>;
+    fn on_chunk_response(
+        &self,
+        _response: &Self::ChunkResponse,
+        _ctx: &Context,
+    ) -> Option<opentelemetry::Value> {
+        None
+    }
 }
 
 pub(crate) trait DefaultForLevel {
--- a/apollo-router/src/plugins/telemetry/config_new/attributes.rs
+++ b/apollo-router/src/plugins/telemetry/config_new/attributes.rs
@@ -514,6 +514,7 @@ impl DefaultForLevel for HttpServerAttributes {
 impl Selectors for RouterAttributes {
     type Request = router::Request;
     type Response = router::Response;
+    type ChunkResponse = ();
 
     fn on_request(&self, request: &router::Request) -> Vec<KeyValue> {
         let mut attrs = self.common.on_request(request);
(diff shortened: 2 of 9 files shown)


BrynCooke (approved):

apollo-router/src/plugins/telemetry/config_new/cost/mod.rs
```diff
@@ -49,14 +49,23 @@ pub(crate) struct SupergraphCostAttributes {
...
+    }
+
+    fn on_error(&self, _error: &BoxError) -> Vec<KeyValue> {
+        Vec::default()
+    }
+
+    fn on_event_response(&self, _response: &Self::EventResponse, ctx: &Context) -> Vec<KeyValue> {
         let mut attrs = Vec::with_capacity(4);
-        if let Some(cost_result) = &response.context.extensions().lock().get::<CostContext>() {
+        if let Some(cost_result) = &ctx.extensions().lock().get::<CostContext>() {
```
o0Ignition0o, line 68: I think we're holding the lock for longer than we should.

apollo-router/src/plugins/telemetry/config_new/mod.rs
```diff
@@ -28,17 +29,31 @@ pub(crate) mod spans;
 pub(crate) trait Selectors {
     type Request;
     type Response;
+    type EventResponse;
+
     fn on_request(&self, request: &Self::Request) -> Vec<KeyValue>;
     fn on_response(&self, response: &Self::Response) -> Vec<KeyValue>;
+    fn on_event_response(&self, _response: &Self::EventResponse, _ctx: &Context) -> Vec<KeyValue> {
```
o0Ignition0o, line 36: should this be `on_response_event`? oh maybe it refers to defer and subscription events only? is there a design doc somewhere?

apollo-router/src/configuration/snapshots/apollo_router__configuration__tests__schema_generation.snap
```diff
@@ -1968,6 +1986,38 @@ expression: "&schema"
       ],
       "type": "object"
     },
+    "Event_for_RouterSelector": {
+      "oneOf": [
+        {
+          "description": "For every chunk sent through http multipart connection",
```
o0Ignition0o, line 1992: This throws me off a bit, intuitively I read that as multipart file upload chunks. Also does this actually apply only to http multipart /responses/ ?

apollo-router/src/plugins/telemetry/config_new/mod.rs
bnjjj: It refers to event coming in the supergraph stream so yes mainly subscription and defer events. Maybe your suggestion is still valid. WDYT @BrynCooke ?

apollo-router/src/plugins/telemetry/config_new/cost/mod.rs
BrynCooke: I don't think this is the case.

There's nothing blocking or async in the block below, so we could obtain the lock for each attribute, but the overhead of repeatedly locking could be worse.

apollo-router/src/configuration/snapshots/apollo_router__configuration__tests__schema_generation.snap
bnjjj: Hmm not really, I don't know how to explain it better because it's highly related to our internals. It's basically event coming in the supergraph response stream. Maybe a better description could be `For every supergraph response payload (including subscription events and defer events)` ?

o0Ignition0o: cc @shorgi WDYT?

I'm going to preapprove the PR and let Edward give his opinion / suggest something :D

o0Ignition0o (approved):

apollo-router/src/plugins/telemetry/config_new/cost/mod.rs
o0Ignition0o: we aren't locking repeatedly, we're cloning CostContext and releasing the lock immediately instead.

apollo-router/src/plugins/telemetry/config_new/mod.rs
bnjjj: Fixed with `on_response_event` thanks for the suggestion

apollo-router/src/configuration/snapshots/apollo_router__configuration__tests__schema_generation.snap
BrynCooke: @bnjjj I like your updated suggestion.

bnjjj: I updated with my suggestion and enable auto merge in order to have this in the next release. Once we have feedback from @shorgi I'll open a new PR to change the description

PR #5023, opened 2024-04-25
testing/ci: Use nextext when available and store JUnit tests from CI runs

abernix:
As tends to be the case, our CI testing times have been driven up over time.  While we currently have broad-stroke measurements on the CI pipeline, we don't yet have per-test insights.  This makes it difficult to commit time to specific efforts, or measure their improvement.  Further, the ergonomics over using the failed tests interface on CircleCI have been hindered by the lack of test reporting through well-defined testing output artifacts like Junit, popular with most other languages.

Rust core hasn't yet promoted its structured `cargo test` output into a stable build which — if we wanted to measure tests by producing anything remotely coercible into JUnit — would mean we'd have to run `+nightly` runs of the tests.  Using the `nightly` build of `cargo test` to get almost (but still not) JUnit seems unnecessarily risky.

Given that this feature hasn't promoted from `+nightly` in what must be plural years and the allure of other more feature-complete testing tools like [Nextest](https://nexte.st/) — I decided to give Nextest a quick try.

The quick test is seemingly successful and validating: It did not show any degradation in test execution time, is already showing promising nice wins in CircleCI interface experience, seems like an improvement to the CLI experience (image below, but also by [highlighting specific test failures in the UI](https://app.circleci.com/pipelines/github/apollographql/router/20805/workflows/09ba269e-b917-4d2e-ad52-68e80e14ab94/jobs/140871/tests)), unlocks other capabilities (like automatic retries on a per-test basis), and didn't require passing custom thread flags ([perhaps due to a better model for test execution](https://nexte.st/book/how-it-works)?).

image

### Local vs CI / How to enable nextest locally

Nextest is only used if it is installed and the decision to do so is made inside of `cargo xtask test` based on the presence of `cargo-nextest`.  If `cargo-nextest` is **not** installed, then the existing `cargo test` is used as a fall-back.  On CI, it is installed and thus used.


abernix:
For some reason, `cargo +nightly fuzz` tests — **a separate job entirely**, which runs **after** tests — [are failing on `arm`](https://app.circleci.com/pipelines/github/apollographql/router/20805/workflows/41d72d75-237d-4c1f-8757-5e7c99a04c4d/jobs/140906) with this PR, if anyone has any idea:

```
   Compiling version_check v0.9.4
error: sanitizer is incompatible with statically linked libc, disable it using `-C target-feature=-crt-static`

error[E0463]: can't find crate for `core`
  |
  = note: the `aarch64-unknown-linux-musl` target may not be installed
  = help: consider downloading the target with `rustup target add aarch64-unknown-linux-musl`
  = help: consider building the standard library from source with `cargo build -Zbuild-std`

For more information about this error, try `rustc --explain E0463`.
error: could not compile `cfg-if` (lib) due to 2 previous errors
warning: build failed, waiting for other jobs to finish...
Error: failed to build fuzz script: ASAN_OPTIONS="detect_odr_violation=0" RUSTFLAGS="-Cpasses=sancov-module -Cllvm-args=-sanitizer-coverage-level=4 -Cllvm-args=-sanitizer-coverage-inline-8bit-counters -Cllvm-args=-sanitizer-coverage-pc-table -Cllvm-args=-sanitizer-coverage-trace-compares --cfg fuzzing -Clink-dead-code -Zsanitizer=address -Cllvm-args=-sanitizer-coverage-stack-depth -Cdebug-assertions -C codegen-units=1" "cargo" "build" "--manifest-path" "/home/circleci/project/fuzz/Cargo.toml" "--target" "aarch64-unknown-linux-musl" "--release" "--config" "profile.release.debug=true" "--bins"

Exited with code exit status 1
```

image

.circleci/config.yml
```diff
@@ -359,11 +359,28 @@ commands:
       - run:
           name: Install cargo deny, about, edit
           command: |
+            if [[ ! -f "$HOME/.cargo/bin/cargo-binstall$EXECUTABLE_SUFFIX" ]]; then
+              cargo install --locked --version 1.6.4 cargo-binstall
```
Geal, line 363: I'm unsure about using cargo-binstall here. This adds a third party source of executables that is less scrutinized the other install sources we have (apt and crates.io)

Geal: I think I'll test https://github.com/MaulingMonkey/cargo-local-install which would allow us to add the executables to Cargo.toml where it's more visible than deeply embedded, and this way we could make sure that xtask can rely directly on it

abernix: Super Fair. I'm also happy to remove this if it's the contentious part of the PR, but would rather land the rest of the PR.  We could also try `--only-signed`, but not sure if that's even trusted enough yet?

abernix: I've reverted `binstall` in e623cce.

Geal (approved):

bnjjj (approved):

o0Ignition0o (approved):

Geal (approved):

garypen (approved):

PR #5025, opened 2024-04-25
QueryHash bug: different queries can share the same hash

xuorig:
Hey folks, just hit this issue in production. Two queries with different fields/arguments can potentially share a hash, meaning they can share a query plan.

- First commit offers a repro
- Second commit with a fix, but haven't thought through the implications.


abernix:
Ok, @o0Ignition0o, @IvanGoncharov and some other folks are taking a deeper look at this in https://github.com/apollographql/router/pull/5028.  I understand that the approach will be to completely back out some of the more nuanced hashing until an approach can be re-applied with a deeper consideration for the side-effects you've seen here.

xuorig:
Sounds good, yep that works!

PR #5026, opened 2024-04-26
chore(maint): Update code and snapshots after proto update

bonnici:
Updates code and snapshots so that we store default values in the new proto fields.

PR #5029, opened 2024-04-26
chore(maint,1.45.1): Update code and snapshots after proto update

abernix:
Re-apply https://github.com/apollographql/router/pull/5026 on https://github.com/apollographql/router/pull/5009. While this doesn't currently work on `dev` in the original PR, this looks like it passes on this branch and would free up the 1.45.1 branch for release.


abernix:
I don't think this is necessary any longer since 1.45.1 has landed.

PR #5030, opened 2024-04-26
Update Apollo Protobuf via GitHub Actions PR

abernix:
In the same spirit as the existing Apollo Uplink schema automation which exists in https://github.com/apollographql/router/actions/workflows/update_uplink_schema.yml.  This will only allow one PR at a time for this branch name, and only commit changes (and update that branch) when there is a change.

### DIFF
--- a/.github/workflows/update_apollo_protobuf.yaml
+++ b/.github/workflows/update_apollo_protobuf.yaml
@@ -6,7 +6,7 @@ on:
     # On the 5 of every hour
     - cron:  '5 * * * *'
 jobs:
-  Update-Uplink-Schema:
+  Update-Protobuf-Schema:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4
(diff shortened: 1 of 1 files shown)


Geal (approved):

o0Ignition0o:
Great, can you remove the `check_reports_proto_is_up_to_date` test?

.github/workflows/update_apollo_protobuf.yaml
```diff
@@ -0,0 +1,32 @@
+name: Update Apollo Protobuf
+run-name: ${{ github.actor }} is updating the Apollo Protobuf 🚀
+on:
+  workflow_dispatch:
+  schedule:
+    # On the 5 of every hour
+    - cron:  '5 * * * *'
+jobs:
+  Update-Uplink-Schema:
```
SimonSapin, line 9:
```suggestion
  Update-Protobuf-Schema:
```

abernix:
Separate PR, yes.

abernix:
Let me make sure this lands and run it once or twice, and then remove the test.  I can only test this ON `dev`.

SimonSapin (approved):

PR #5031, opened 2024-04-26
tests: Remove "is proto fresh" test in lieu of GitHub Action

abernix:
We'll have a GitHub Action do the maintenance on this file, rather than
having a test remind us that it's stale.  The test results in all PRs (and
all downstream users of this repository) suddenly getting failing tests
every time our infra-team deploys a new version of the proto, whereas the
GitHub Action can be a much more controlled burn.

The GitHub action that replaces this is in [#5030], and we will merge this test just as soon as that is validated.

[#5030]: https://github.com/apollographql/router/pull/5030

### DIFF
--- a/apollo-router/src/plugins/telemetry/apollo_exporter.rs
+++ b/apollo-router/src/plugins/telemetry/apollo_exporter.rs
@@ -442,17 +442,3 @@ where
         None => serializer.serialize_none(),
     }
 }
-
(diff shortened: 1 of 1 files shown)


SimonSapin (approved):

abernix:
It appears to work! https://github.com/apollographql/router/pull/5032

PR #5038, opened 2024-04-29
generate the JSON schema only once

Geal:
Surprisingly, the most expensive part of running tests in the router is in JSON schema compilation. Every time we validate a configuration, we generate the JSON configuration schema (which is about 31k lines at this point) then compile a JSONSchema structure used to validate the configuration.
This computation can be done only once per execution without losing anything, as it only depends on the configuration schema, which is generated from the router's code and native plugins.

*Description here*

Fixes #**issue_number**

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/configuration/schema.rs
+++ b/apollo-router/src/configuration/schema.rs
@@ -3,6 +3,7 @@
 use std::borrow::Cow;
 use std::cmp::Ordering;
 use std::fmt::Write;
+use std::sync::OnceLock;
 
 use itertools::Itertools;
 use jsonschema::error::ValidationErrorKind;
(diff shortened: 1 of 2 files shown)


apollo-router/src/configuration/schema.rs
```diff
@@ -82,19 +83,28 @@ pub(crate) fn validate_yaml_configuration(
...
+                        .with_draft(Draft::Draft7)
+                        .compile(&schema)
+                        .map_err(|e| ConfigurationError::InvalidConfiguration {
+                            message: "failed to compile schema",
+                            error: e.to_string(),
+                        })
+                })
+        })
+        .as_ref()
+        .map(|schema| schema)
```
Geal, line 106:
```suggestion
```

bnjjj (approved):

BrynCooke (approved):

PR #5043, opened 2024-04-29
start the router on a random port in integration tests

Geal:
This fixes the random port binding in integration tests. Previously, we were trying to allocate a random port then setting it in the router's configuration, which still caused races between routers on listen ports, so tests were still failing.
With this change, we tell the router to start on a random port, then extract the port number from the startup log lines.

TODO:
- the health check does not automatically listen on the same port and uses the address incorrectly (`Health check exposed at http://127.0.0.1:0/health`)
- prometheus automatically listens on the same port but reports the wrong port: `Prometheus endpoint exposed at http://127.0.0.1:0/metrics`

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/tests/common.rs
+++ b/apollo-router/tests/common.rs
@@ -1,4 +1,3 @@
-use parking_lot::Mutex;
 use std::collections::HashMap;
 use std::fs;
 use std::net::IpAddr;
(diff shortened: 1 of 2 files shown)


BrynCooke:
Looks like this duplicates: https://github.com/apollographql/router/pull/5040

Geal:
oh wow :smile: 
I had not seen it. I like your approach better, I had not looked into simplifying it with a regex yet

PR #5047, opened 2024-04-29
Use supergraph schema to extract auth info

tninesling:
Use supergraph schema to extract auth info as auth information may not be available on the query planner's subgraph schemas. This undoes the auth changes made in #4975.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/query_planner/bridge_query_planner.rs
```diff
@@ -1062,11 +1062,11 @@ impl QueryPlan {
 
     fn extract_authorization_metadata(
         &mut self,
-        subgraph_schemas: &SubgraphSchemas,
+        schema: &apollo_compiler::Schema,
```
Geal, line 1065: isn't `supergraph_schema()` returnin a `Valid<Schema>`?

tninesling: It does. This first pass just restores the previous type signature, but I can update to use `Valid<Schema>` so we don't need `Valid::assume_valid_ref` anymore

tninesling: Updated in https://github.com/apollographql/router/pull/5047/commits/cf25e1660cafe6eb323845f831a7eb546907c9e3

Geal (approved):

apollo-router/src/query_planner/fetch.rs
```diff
@@ -627,13 +627,18 @@ impl FetchNode {
 
     pub(crate) fn extract_authorization_metadata(
         &mut self,
-        subgraph_schemas: &SubgraphSchemas,
+        schema: &Valid<apollo_compiler::Schema>,
         global_authorisation_cache_key: &CacheKeyMetadata,
     ) {
-        let doc = self.parsed_operation(subgraph_schemas);
-        let schema = &subgraph_schemas[self.service_name.as_str()];
+        let doc = ExecutableDocument::parse(
```
SimonSapin, line 633: Re parsing here is unfortunate. Would it make sense to call `self.parsed_operation()` here with the subgraph schema, but then pass the supergraph schema to `AuthorizationPlugin::generate_cache_metadata` for finding directives?

garypen: This was resolved by off-GH discussion. I'm going to merge.

PR #5061, opened 2024-05-01
Make json schema small

### LINKED ISSUE #5003: Reducing the configuration schema file
the [generated configuration schema](https://github.com/apollographql/router/blob/dev/apollo-router/src/configuration/snapshots/apollo_router__configuration__tests__schema_generation.snap) is now around 35000 lines. It might be time to reduce it a bit. schemars is supposed to extract object types into references that can be reused when necessary, but it looks like this is not happening, especially in the telemetry conditions. Can we find a way to reduce it further?
This will also have an effect on generated code size (schemars derived code is huge) which will impact linking times

BrynCooke:
Fix #5003

This reworks the json schema generation for the config so that it is reduced in size from approx 100k lines to jsut over 7k.

The fix involves three things:
1. Enable references on json schema generation. This got disabled in the past because there were issues with the generated references, but by adding a schema visitor we can work around this.
2. Adjust the schema generation for Extendable and Conditional. These previously relied on the scheme not using references.
3. Modify orbiter metrics to redact only based on the properties in the schema rather than on validation metadata as this is not possible when using schema refs: https://github.com/Stranger6667/jsonschema-rs/issues/403

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


.changesets/fix_bryn_json_schema.md
```diff
@@ -0,0 +1,7 @@
+### Make json schema small ([PR #5061](https://github.com/apollographql/router/pull/5061))
+
+The json schema for the router.yaml is reduced in size from approx 100k lines to just over 7k.
+
+This reduces the startup time of the Router and a smaller schema is more friendly7 for code editors.
```
Geal, line 5:
```suggestion
This reduces the startup time of the Router and a smaller schema is more friendly for code editors.
```

apollo-router/src/orbiter/snapshots/apollo_router__orbiter__test__visit_config.snap
```diff
@@ -2,20 +2,18 @@
 source: apollo-router/src/orbiter/mod.rs
 expression: usage
 ---
-configuration.headers.all.request.insert.len: 1
 configuration.headers.all.request.insert.name.<redacted>: 1
 configuration.headers.all.request.insert.value.<redacted>: 1
-configuration.headers.all.request.propagate.len: 2
```
Geal, line 8: why are we losing propagate and remove?

Geal (approved):
that sounds reasonable. So you solved the orbiter issue?

apollo-router/src/orbiter/snapshots/apollo_router__orbiter__test__visit_config.snap
BrynCooke: I've improved it a bit: [e096eb1](https://github.com/apollographql/router/pull/5061/commits/e096eb19cf908c8a865b61b3312c181045e0ccd6)

PR #5064, opened 2024-05-03
Update the version of the router-bridge

garypen:
Update our router-bridge version to pull in fixes.


o0Ignition0o:
@garypen you need to update the lockfile

o0Ignition0o (approved):

PR #5065, opened 2024-05-03
Import branch 'main' of apollographql/federation-next into 'dev'

SimonSapin:
**This PR is to be merge with a true merge commit, not squashed or rebased.**
    
The git history of the federation-next repository has been rewritten using git-filter-repo version ae71fad1d03f:

```
../git-filter-repo \
  --path-rename :apollo-federation/ \
  --tag-rename v:apollo-federation@ \
  --replace-message <(echo 'regex:#(\d+)==>https://github.com/apollographql/federation-next/pull/\1')
```

The rewrite is mainly so the history is as if files had always been in the new `apollo-federation/` sub-directory. That way `git blame` will work as expected.

The upstream 'main' branch was at commit https://github.com/apollographql/federation-next/commit/f8904bfebd41ffe6ec8a886fe790dd95a836db54 and was mapped to commit a1d54b64437fc26df191b195cf2d4063094eb9bb in this PR.

Build system integration to make Router use the newly imported code will be in a follow-up PR, so that this one only contains changes that have already been reviewed upstream.


lrlna (approved):
lesgo

o0Ignition0o (approved):

goto-bus-stop (approved):
And most PR references in commit messages Just Work :tm: !

PR #5067, opened 2024-05-03
file based integration tests

Geal:
This defines a way to write integration tests using JSON files, to avoid the long linking times that we are suffering when writing unit or integration tests inside the Router. Writing and refining a test plan is a lot faster and lends itself more to a highly iterative way to work.

The way it works right now:
- you define a `plan.json` file that contains a serie of actions
- actions available:
  - start a router with a configuration, schema and some mocked subgraphs
  - stop the router
  - load a new configuration
  - load a new schema
  - make a request and verify that we get the expected response

This is currently an experimentation (with all tests failing on purpose), that we should adapt to our needs now that we can see it works. We could very well start using it ASAP and evolve it as needed, as we've done with other testing systems.

Things to decide on:
- we expect every test as a folder under `apollo_router/tests/samples`. Do we want to support grouping under subfolders? 'per feature, etc)
- this does not support Enterprise features for now. We could easily require an API key in envirment when we do our local tests (I think most of us already do it?) and have it available in CI
- for now this is done as an integration test under `apollo-router/tests`. I think this could work as a separate executable that could be used outside of the Router's CI. Example: to run a set of request/response for a customer, in a non public environment

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


lrlna:
Love this!

I do like the idea of grouping things by features and/or functionality. I'd love to see a separation of `ok` and `err` scenarios too, so we can easily add cases that we run into to either of the buckets. So something like this:
```
apollo-router/
├─ tests/
│  ├─ query_planner_cache/
│  │  ├─ ok/
│  │  │  ├─ README.md
│  │  │  ├─ configuration.yaml
│  │  │  ├─ supgraph.graphql
│  │  │  ├─ plan.json
│  │  ├─ err/
│  │  │  ├─ README.md
│  │  │  ├─ configuration.yaml
│  │  │  ├─ supergraph.graphql
│  │  │  ├─ plan.json
│  ├─ subscriptions/
```
Maybe in the future we can take it even further, and use snapshots to also store possible responses. And each "scenario" could have files for operations and responses that we can add and tweak as the router changes.

Geal:
having subfolders to separate different kinds of tests sounds nice

Geal:
I added support the subfolder because adding other files like the README interacted weirdly with the test runner, and subfolders definitely feel very nice to run this:

```
   Canceling due to test failure: 1 test still running
        PASS [   2.575s] apollo-router::samples /basic/query2
------------
     Summary [   2.576s] 2 tests run: 1 passed, 1 failed, 0 skipped
        FAIL [   0.256s] apollo-router::samples /basic/query1
```

I added a README.md file to explain what to expect and how to write tests. The tests are now passing (I made them explicitely fail before to check the output locally and in CI), so this is good to review and merge

o0Ignition0o (approved):
Great start! love it!

PR #5070, opened 2024-05-03
chore(federation): Remove Normalized prefix from operation types

goto-bus-stop:
As discussed in standup a few days ago. I know not everyone was there so feel free to raise objections here too. I thought this might be a nice opportunity to do it since we have very little open work due to the repo move.

We have several things that are called normalization, but the structures actually do not guarantee any of them have been done. They are just the Operation / Selection structures that apollo-federation uses internally, so we should simply call them Operation, SelectionSet, FieldSelection, etc, and convert between apollo-compiler and apollo-federation types at public API boundaries.

### DIFF
--- a/apollo-federation/src/query_plan/operation.rs
+++ b/apollo-federation/src/query_plan/operation.rs
@@ -23,22 +23,8 @@ use std::sync::atomic;
 use std::sync::Arc;
 use std::sync::OnceLock;
 
-use apollo_compiler::ast::Argument;
-use apollo_compiler::ast::Directive;
-use apollo_compiler::ast::DirectiveList;
-use apollo_compiler::ast::Name;
-use apollo_compiler::ast::OperationType;
-use apollo_compiler::ast::Type;
-use apollo_compiler::ast::Value;
 use apollo_compiler::executable;
-use apollo_compiler::executable::Field;
-use apollo_compiler::executable::Fragment;
-use apollo_compiler::executable::FragmentSpread;
-use apollo_compiler::executable::InlineFragment;
-use apollo_compiler::executable::Operation;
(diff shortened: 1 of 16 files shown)


apollo-federation/src/query_plan/operation.rs
```diff
@@ -155,24 +145,24 @@ fn same_directives(left: &DirectiveList, right: &DirectiveList) -> bool {
...
-    pub(crate) variables: Arc<Vec<Node<VariableDefinition>>>,
-    pub(crate) directives: Arc<DirectiveList>,
-    pub(crate) selection_set: NormalizedSelectionSet,
+    pub(crate) variables: Arc<Vec<Node<executable::VariableDefinition>>>,
+    pub(crate) directives: Arc<executable::DirectiveList>,
+    pub(crate) selection_set: SelectionSet,
     pub(crate) named_fragments: NamedFragments,
 }
 
 pub(crate) struct NormalizedDefer {
```
goto-bus-stop, line 158: This `Normalized` is an unrelated kind of `Normalized` from the selection and operation types, so I kept it.

lrlna (approved):
i am down with this! makes that operation.rs file and its usages across query_graph and query_plan a lot more readable for me

apollo-federation/src/subgraph/spec.rs
```diff
@@ -166,7 +166,7 @@ pub trait AppliedFederationLink {
 macro_rules! applied_specification {
     ($($t:ty),+) => {
         $(impl AppliedFederationLink for $t {
-            /// ```
+            /// ```graphql
```
goto-bus-stop, line 169: Unrelated change but prevents it from being interpreted as a doc test.

PR #5073, opened 2024-05-03
Remove the potential to panic during operation parse/validate

garypen:
DO NOT MERGE THIS PR. THIS IS A BACKUP PR TO POTENTIALLY FACILITATE A ROUTER RELEASE IF THE MORE COMPLETE FIX IS NOT AVAILABLE IN THE RIGHT TIMEFRAME.

The possibility that an ExecutableDocument::parse_and_validate() will fail, means there is a possibility that a router will panic when `as_parsed()` or `as_serialized()` is invoked.

We remove the potential to panic by logging errors for any detected errors and continuing to use the Valid (but incomplete) document, which is the behaviour we had before this change.

Summary: This restores the preceding behaviour of the router with the additional error logging of detected errors.

### DIFF
--- a/apollo-router/src/query_planner/fetch.rs
+++ b/apollo-router/src/query_planner/fetch.rs
@@ -174,15 +174,20 @@ impl SubgraphOperation {
                 .serialized
                 .get()
                 .expect("SubgraphOperation has neither representation initialized");
-            Arc::new(
-                ExecutableDocument::parse_and_validate(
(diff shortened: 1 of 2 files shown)


garypen:
superceded by #5075

goto-bus-stop:
#5075 is used instead. it also isn't what we ultimately want but a minor improvement over the fallback idea from this PR.

PR #5085, opened 2024-05-04
added dev-docs/federation/type-and-directive-specification.md

duckki:
I added a technical documentation regarding src/schema/type_and_directive_specification.rs implementation.

* This is a developer documentation-only change.


lrlna:
@duckki i know we talked about adding these directly to the repo, but can I ask you to actually move this under [dev docs in confluence here](https://apollographql.atlassian.net/wiki/spaces/F/pages/619839576/Dev+Docs)?

PR #5097, opened 2024-05-07
Add ability for router to deal with query plans with contextual rewrites

clenfest:
Two main things that we're doing in this PR. 
1. We've added a variable to FetchNode called `context_rewrites`. This is a vector of DataRewrite::KeyRenamer that are specifically taking data from their path (which will be relative and can traverse up the data path) and writes the data into an argument that is passed to the selection set.
2. There are two cases. In the most straightforward, the data that is passed to the selection set is the same for every entity. This case is pretty easy and doesn't require any special handling. In the second case, the value of the variable may be different per entity. If that is true, we need to use aliasing and duplication in our query in order to send it to subgraphs. Once https://github.com/graphql/composite-schemas-spec/issues/25 is decided and has subgraph support, this query cloning will be able to go away. 

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [x] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/Cargo.toml
```diff
@@ -185,7 +185,7 @@ regex = "1.10.3"
 reqwest.workspace = true
 
 # note: this dependency should _always_ be pinned, prefix the version with an `=`
-router-bridge = "=0.5.21+v2.7.5"
+router-bridge = { git = "https://github.com/apollographql/federation-rs", rev = "8e8ba0d60f99693e5cf43477261e55a9c426f9c9" }
```
o0Ignition0o, line 188: TODO: update with 2.7.8 once it's out

apollo-router/build/main.rs
```diff
@@ -10,23 +10,23 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+    // let router_bridge_version = router_bridge
+    //     .as_str()
+    //     .or_else(|| {
+    //         router_bridge
+    //             .as_object()
+    //             .and_then(|o| o.get("version"))
+    //             .and_then(|version| version.as_str())
+    //     })
+    //     .expect("router-bridge does not have a version");
+    let router_bridge_version = "0.5.20+v2.8.0";
```
o0Ignition0o, line 29: todo: revert once 2.8.0 is ready

apollo-router/src/query_planner/fetch.rs
```diff
@@ -243,6 +255,189 @@ impl Display for QueryHash {
...
+    };
+
+    if path.len() > 1 {
+        if let Some(val) = v {
+            let remaining_path = path.iter().skip(1).cloned().collect();
+            return data_at_path(val, &remaining_path);
+        }
+    }
+    v
+}
```
lrlna, line 408: I don't think this exists yet, but I think this can perhaps be a (`pub(crate) fn`) getter on the `Path` struct.

apollo-router/src/query_planner/fetch.rs
```diff
@@ -243,6 +255,189 @@ impl Display for QueryHash {
...
+        }
+    }
+
+    let mut return_path: Vec<PathElement> = current_dir.iter().take(j).cloned().collect();
+
+    context_path.iter().skip(i).for_each(|e| {
+        return_path.push(e.clone());
+    });
+    Path(return_path.into_iter().collect())
 }
```
lrlna, line 441: There is already a `path.join()`, I think this can be a `path.merge()`. But I'd also have it as `pub(crate) fn`.

apollo-router/src/query_planner/fetch.rs
```diff
@@ -364,13 +660,23 @@ impl FetchNode {
...
         };
 
+        let query_batched_query = if let Ok(qbq) =
+            query_batching_for_contextual_args(operation.as_serialized(), &contextual_args)
+        {
+            qbq
+        } else {
+            // TODO: do we need to provide an error here ?
+            return (Default::default(), Default::default());
+        };
```
lrlna, line 678: I don't know if we want to do this right in the `fetch_node` implementation. Is this a good idea? Should this live a step up from here? Just feels like we are modifying an operation righttttt before it gets sent off.

apollo-router/src/plugins/expose_query_plan.rs
```diff
@@ -218,7 +218,8 @@ mod tests {
             build_mock_supergraph(serde_json::json! {{
                 "plugins": {
                     "experimental.expose_query_plan": true
-                }
+                },
+                "include_subgraph_errors": { "all": true }
```
garypen, line 222: Is this necessary? I imagine it would be if you added the errors tests suggested below, but right now it seems unnecessary.

apollo-router/src/query_planner/fetch.rs
```diff
@@ -243,6 +255,189 @@ impl Display for QueryHash {
 pub(crate) struct Variables {
     pub(crate) variables: Object,
     pub(crate) inverted_paths: Vec<Vec<Path>>,
+    pub(crate) contextual_args: Option<(HashSet<String>, usize)>,
+}
+
+#[derive(Debug)]
+enum ContextBatchingError {
+    InvalidDocument(WithErrors<Document>),
```
garypen, line 263: If the changes @lrlna suggests about not re-parsing the operation are adopted, then this enum variant isn't required. With only one variant, it may even make sense to convert this to a struct.

apollo-router/src/query_planner/fetch.rs
```diff
@@ -243,6 +255,189 @@ impl Display for QueryHash {
...
+    if let Some((ctx, times)) = contextual_args {
+        let parser = apollo_compiler::Parser::new()
+            .parse_ast(operation, "")
+            .map_err(ContextBatchingError::InvalidDocument)?;
+        if let Some(mut operation) = parser
+            .definitions
+            .into_iter()
+            .find_map(|definition| definition.as_operation_definition().cloned())
+        {
+            let mut new_variables: Vec<_> = Default::default();
```
garypen, line 280: NIT: why not just `vec![]`?

apollo-router/src/query_planner/fetch.rs
```diff
@@ -243,6 +255,189 @@ impl Display for QueryHash {
...
+    pub(crate) contextual_args: Option<(HashSet<String>, usize)>,
+}
+
+#[derive(Debug)]
+enum ContextBatchingError {
+    InvalidDocument(WithErrors<Document>),
+    NoSelectionSet,
+}
+
+fn query_batching_for_contextual_args(
```
garypen, line 267: This whole function could really use a block comment to explain what it is doing. The cloning and make_mut in particular make it a little difficult to follow.

apollo-router/src/query_planner/fetch.rs
```diff
@@ -243,6 +255,189 @@ impl Display for QueryHash {
...
+                        }
+
+                        Ok(s)
+                    })
+                    .collect::<Result<Vec<_>, _>>()?;
+
+                let new_operation = operation.make_mut();
+                new_operation.selection_set = new_selection_set;
+                new_operation.variables = new_variables;
+
```
garypen, line 320: Is it possible to have generated an invalid operation at this point? Do we need to re-validate here?

apollo-router/src/query_planner/fetch.rs
```diff
@@ -243,6 +255,189 @@ impl Display for QueryHash {
...
+    assert_eq!(
+        expected,
+        query_batching_for_contextual_args(old_query, &contextual_args)
+            .unwrap()
+            .unwrap()
+    );
+}
+
+// TODO: There is probably a function somewhere else that already does this
+fn data_at_path<'v>(data: &'v Value, path: &Path) -> Option<&'v Value> {
```
Geal, line 383: we have `select_values_and_paths`, or `get_path`, or the lower level `iterate_path`

apollo-router/src/query_planner/fetch.rs
```diff
@@ -364,13 +660,23 @@ impl FetchNode {
...
+            &self.context_rewrites,
         ) {
             Some(variables) => variables,
             None => {
                 return (Value::Object(Object::default()), Vec::new());
             }
         };
 
+        let query_batched_query = if let Ok(qbq) =
+            query_batching_for_contextual_args(operation.as_serialized(), &contextual_args)
```
garypen, line 672: There's no point making us re-parse the operation by passing it as a string.

```suggestion
            query_batching_for_contextual_args(operation.as_parsed(), &contextual_args)
```

apollo-router/src/query_planner/fetch.rs
garypen, line 267: Why `query_batching_for_contextual_args`? I'm not sure what meaning this name is trying to convey?

apollo-router/src/query_planner/fetch.rs
```diff
@@ -243,6 +255,189 @@ impl Display for QueryHash {
...
+                    })
+                    .collect::<Result<Vec<_>, _>>()?;
+
+                let new_operation = operation.make_mut();
+                new_operation.selection_set = new_selection_set;
+                new_operation.variables = new_variables;
+
+                return Ok(Some(new_operation.serialize().no_indent().to_string()));
+            }
+        }
```
Geal, line 323: it is unclear to me why we would need to parse and edit the operation. And it is worrying that this would be called in `Variables::new`, so on every subgraph request. That's a perf hit that we cannot afford

Geal: is that because different entities might need different values from the context?

garypen (commented):
Hi @clenfest . We've done some synchronised reviewing within the router team today. Four of us have spent a couple of hours looking through the code and the design document and you'll see comments from a bunch of folks.

We want to work with you to help this land, so expect a lot of follow-ups.

apollo-router/src/query_planner/fetch.rs
```diff
@@ -243,6 +255,189 @@ impl Display for QueryHash {
...
+    }
+
+    Ok(None)
```
Geal, line 326: if there are no contextual args, we return None instead of keeping the original query? Why?

apollo-router/src/query_planner/fetch.rs
```diff
@@ -256,32 +451,95 @@ impl Variables {
...
+                    })
+                    .collect();
+
                 let mut value = execute_selection_set(value, requires, schema, None);
                 if value.as_object().map(|o| !o.is_empty()).unwrap_or(false) {
                     rewrites::apply_rewrites(schema, &mut value, input_rewrites);
                     match values.get_index_of(&value) {
                         Some(index) => {
                             inverted_paths[index].push(path.clone());
+                            named_args.push(hash_map);
```
Geal, line 536: @clenfest could we get a weird interaction with entity deduplication here?
Example: `{ topProducts { reviews { author { name } } } }` with `Product` providing context and `User.name` using that context.
We could get in a case where:
- the context from product 1 is used in users 1 and 2
- the context from product 2 is used in users 2 and 3

initially that would have made an entities query with the representations `[{"typename": "User", "id": "1"}, {"typename": "User", "id": "2"}, {"typename": "User", "id": "2"}, {"typename": "User", "id": "3"}]`
but entity deduplication reduces it to   `[{"typename": "User", "id": "1"}, {"typename": "User", "id": "2"}, {"typename": "User", "id": "3"}]`

but here when using context, we would still need two sub operations, one with the contextual variable from product 1 and representations `[{"typename": "User", "id": "1"}, {"typename": "User", "id": "2"}]`, and the other with the contextual variable from product 2 and representations `[{"typename": "User", "id": "2"}, {"typename": "User", "id": "3"}]`

apollo-router/src/query_planner/fetch.rs
```diff
@@ -243,6 +255,189 @@ impl Display for QueryHash {
...
+}
+
+fn query_batching_for_contextual_args(
+    operation: &str,
+    contextual_args: &Option<(HashSet<String>, usize)>,
+) -> Result<Option<String>, ContextBatchingError> {
+    if let Some((ctx, times)) = contextual_args {
+        let parser = apollo_compiler::Parser::new()
+            .parse_ast(operation, "")
+            .map_err(ContextBatchingError::InvalidDocument)?;
```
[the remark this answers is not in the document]
clenfest: I noticed that that function requires me to pass in a apollo_compilter::Schema. Any thoughts on the best way to get that? It doesn't seem to be immediately available to me.

apollo-router/src/query_planner/fetch.rs
clenfest: Even with my invented '..' syntax?

apollo-router/src/query_planner/fetch.rs
clenfest: I think I answered this on the call, but we only need to rewrite the operation when argument values are different across the invocations for each entity. We'll make sure that this is fast when:
1. no context_rewrites are present
2. Argument values are the same for all entities being fetched.

apollo-router/src/query_planner/fetch.rs
clenfest: Will change the name

apollo-router/src/query_planner/fetch.rs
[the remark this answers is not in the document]
lrlna: There is a schema available on the `parameters` of the `fetch_node` function, but it'll need to be converted. That's another reason to not have this happen directly in the fetch_node, but a bit earlier - we pretty much always should be operating on `Valid<ExecutableDocument>` or `Valid<Schema>`, especially if we are modifying an operation.

apollo-router/src/query_planner/fetch.rs
lrlna: You can give the `path.merge()` a `strategy` parameter to accommodate a the `..`. But what does the `..` actually mean? Why is context_path specifically different from other paths in the query planner?

apollo-router/src/query_planner/fetch.rs
```diff
@@ -256,32 +451,95 @@ impl Variables {
...
+                                                    v,
+                                                    &DataRewrite::KeyRenamer({
+                                                        DataKeyRenamer {
+                                                            path: data_path.clone(),
+                                                            rename_key_to: item
+                                                                .rename_key_to
+                                                                .clone(),
+                                                        }
+                                                    }),
+                                                );
```
lrlna, line 503: forgot to comment on this before, but this new function seems to be doing what an existing function does on the `DataRewrite` impl - So I'd create a the `DataRewrite` first, and then run `maybe_apply` (will have to be made `pub(crate)`) on it:

```suggestion
                                                let data_rewrite = DataRewrite::KeyRenamer({
                                                    DataKeyRenamer {
                                                        path: data_path.clone(),
                                                        rename_key_to: item.rename_key_to.clone(),
                                                    }
                                                });
                                                data_rewrite.maybe_apply(schema, v);
```

apollo-router/src/query_planner/fetch.rs
clenfest: I think because there's custom notation in here, I'll leave it where it is and remove the TODO.

apollo-router/src/query_planner/fetch.rs
clenfest: The .. is to indicate that it's a relative path rather than an absolute one. We're indicating that we need to traverse up the data tree from the current position.

Cargo.lock
```diff
@@ -3422,7 +3422,7 @@ dependencies = [
  "httpdate",
  "itoa",
  "pin-project-lite",
- "socket2 0.5.5",
+ "socket2 0.4.9",
```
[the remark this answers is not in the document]
clenfest: I have no idea how this happened.

apollo-router/src/query_planner/fetch.rs
clenfest: I don't think this is an issue as we're only going to perform the query after deduplication is already done if I'm understanding the setup correctly.

apollo-router/src/query_planner/subgraph_context.rs
```diff
@@ -0,0 +1,361 @@
...
+    }
+}
+
+// Take the existing subgraph operation and rewrite it to use aliasing. This will occur in the case
+// where we are collecting entites and different entities may have different variables passed to the resolver.
+pub(crate) fn build_operation_with_aliasing(
+    subgraph_operation: &SubgraphOperation,
+    contextual_arguments: &Option<ContextualArguments>,
+    schema: &Schema,
+) -> Result<Operation, ContextBatchingError> {
```
lrlna, line 237: this should _not_ return a new operation. This should be modifying an `ExecutableDocument` and returning a `Valid<ExecutableDocument>>`. That way we can be sure that the operation is semantically valid before sending the operation to subgraphs. 

This will also make serialization easier - `ExecutableDocument::serialized()` returns a formatted string.

apollo-router/src/query_planner/subgraph_context.rs
```diff
@@ -232,62 +236,65 @@ impl<'a> SubgraphContext<'a> {
...
-    contextual_arguments: &Option<ContextualArguments>,
+    contextual_arguments: &ContextualArguments,
     schema: &Schema,
-) -> Result<Operation, ContextBatchingError> {
+) -> Result<Valid<ExecutableDocument>, ContextBatchingError> {
+    dbg!(
+        &subgraph_operation.to_string(),
+        &contextual_arguments,
+        &schema.raw_sdl
+    );
```
lrlna, line 246: Does this actually need to return a `ContextBatchingError`? It seems like at the end of the day we either modify or return an operation as is. And when we match on the result in `fetch_node` we return an invalid operation error.

o0Ignition0o: I think there's value in knowing the `subgraph_operation` couldn't be parsed, or that due to a bug we couldn't generate a valid operation with aliasing.

That being said I'm not sure what we can do about it, hence the tracing::debug!() for now. but maybe we could add a metric?

as a side note `subgraph_operation.as_parsed(schema.supergraph_schema());` is ... not my favorite structure: https://github.com/apollographql/router/blob/f41065e726b30b52535f7211b5f5ff9db5f4ef20/apollo-router/src/query_planner/fetch.rs#L173

as_parsed takes a schema as parameter, but might not use it if the document has already been parsed (let's hope the schema is the same...)

Cargo.lock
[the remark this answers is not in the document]
clenfest: No idea how, but it appears to have resolved itself. I suspect it may have been because I was using a custom federation-rs?

apollo-router/tests/fixtures/set_context/one.json
```diff
@@ -0,0 +1,271 @@
+{
+  "mocks": [
```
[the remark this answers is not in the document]
clenfest: I added a couple. Let me know if there's anything you were expecting that you didn't see.

apollo-router/src/query_planner/fetch.rs
clenfest: I would say it's unlikely.

garypen (commented):
This is looking a lot better than earlier in the week. Great work!

My main interests now are in checking what happens if a subgraph returns an error and the assumptions about which operation to use in `build_operation_with_aliasing()`.

apollo-router/src/query_planner/fetch.rs
```diff
@@ -364,13 +385,44 @@ impl FetchNode {
...
         ) {
             Some(variables) => variables,
             None => {
                 return (Value::Object(Object::default()), Vec::new());
             }
         };
 
+        let alias_query_string; // this exists outside the if block to allow the as_str() to be longer lived
+        let aliased_operation = if let Some(ctx_arg) = contextual_arguments {
+            // TODO: subgraph_schemas should be a HashMap of NodeStr or Name.
```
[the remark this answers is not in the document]
o0Ignition0o: I think it would be a good followup in general for the router to not work with `String`s but `NodeStr` or `Name` for any structures generated by federation_next or the router-bridge.

It's definitely not in the scope of this PR though!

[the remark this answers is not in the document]
clenfest: That was a @o0Ignition0o comment. Was there something more to do here?

apollo-router/src/query_planner/subgraph_context.rs
```diff
@@ -0,0 +1,426 @@
...
+    contextual_arguments: &ContextualArguments,
+    subgraph_schema: &Valid<apollo_compiler::Schema>,
+) -> Result<Valid<ExecutableDocument>, ContextBatchingError> {
+    let mut selections: Vec<Selection> = vec![];
+    let ContextualArguments { arguments, count } = contextual_arguments;
+    let parsed_document = subgraph_operation.as_parsed(subgraph_schema);
+    if let Ok(document) = parsed_document {
+        // TODO: Can there be more than one named operation?
+        //       Can there be an anonymous operation?
+        if let Some((_, op)) = document.named_operations.first() {
```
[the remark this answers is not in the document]
clenfest: Ok, I fixed this. I think there will only be one operation, but I just clone any anonymous or named operations that may or may not exist.

apollo-router/Cargo.toml
```diff
@@ -188,7 +188,7 @@ regex = "1.10.3"
 reqwest.workspace = true
 
 # note: this dependency should _always_ be pinned, prefix the version with an `=`
-router-bridge = "=0.5.21+v2.7.5"
+router-bridge = "=0.5.23+v2.8.0-alpha.0"
```
garypen, line 191: Are we happy releasing a router which is using a federation alpha release? Is the intent to update this before we release the next version of the router?

garypen (approved):
Great work on this and thanks for adding the extra tests and cleanup.

I have one comment about alpha federation release, which I don't think we've done before in the router, but apart from that I'm happy.

apollo-router/Cargo.toml
clenfest: Yes. This is just to let us bake in a bit more testing before we officially release 2.8.0, but if all goes well we can just promote it. We'll certainly release and bump the pin before we cut a 1.48.0 rc

PR #5100, opened 2024-05-07
Add the query planner options to the query plan cache key

### LINKED ISSUE #5093: distributed Query Plan cache may be poisoned by experimental query plan features
**Describe the bug**
If a router instance is sharing a distributed cache with other router instances, and one router instance used experimental query plan features, the resulting query plans will be cached with the same id as query plans that are not using the experimental features. 

**To Reproduce**
1. Deploy a cluster of router instances sharing a distributed cache
2. Deploy a single instance of the router (sharing the same distributed cache) that is configured to use experimental query plan features such as `generate_query_fragments`
3. Issue a request against the new router instance and then the same query against the other instances
4. All instances will issue a query using the experimental query plan feature

**Expected behavior**
Query plans created with different federation versions and federation options will be cached with a unique key

Geal:
Fix #5093

Some query plan options can result in incompatible generated plans if they are enabled or disabled, so we must take them into account in the cache key, otherwise a router might get a plan from the cache that does not match its configuration.
We cannot cover the entire apollo-federation config in the hash for now, this will have to be done as a follow up

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


BrynCooke:
I added some testing for this which meant extending the integration test runner a little.

I'm not massively keen on the way that we obtain the config for the query planners. Once the rust query planner is hashable maybe we could consider implementing `From<Configuration> for QueryPlannerConfig`

.changesets/fix_geal_cache_key_qp_options.md
```diff
@@ -0,0 +1,32 @@
...
+
+This could cause issue in the following scenarios:
+1. The Router configuration changes and a query plan is loaded from cache which is incompatible with the new configuration.
+2. Routers with differing configuration are sharing the same cache causing them to cache and load incompatible query plans. 
+
+Now a hash for the entire query planner configuration is included in the cache key to prevent this from happening.
+
+If you are running previous versions of the Router and wish to change the query planner options then you must take care to
+change the cache namespace to prevent collisions.
+
```
abernix, line 22: I'm trying to build my own clarity here — is this a changelog entry that's meant to inform previous users of the Router, or is it still somehow relevant after this fix?   My understanding is that this fix makes that no longer a concern, but previous users would have had to leverage a custom cache key prefix.  Is that right?

If my understanding is correct, I would just remove line 20 through the of the example here because it makes it sound like _using_ this cache key prefix is necessary if I was coming from a previous version of the Router.

abernix (changes requested):
I have reviewed some softer bits and requested changes around them on account of looking for answers to my questions, but I will leave the rest of the review to others for now.  Happy to revisit.

apollo-router/src/query_planner/caching_query_planner.rs
```diff
@@ -50,6 +52,15 @@ pub(crate) type Plugins = IndexMap<String, Box<dyn QueryPlannerPlugin>>;
 pub(crate) type InMemoryCachePlanner =
     InMemoryCache<CachingQueryKey, Result<QueryPlannerContent, Arc<QueryPlannerError>>>;
 
+#[derive(Debug, Clone, Hash, PartialEq, Eq, Serialize)]
+pub(crate) enum ConfigMode {
+    //FIXME: add the Rust planner structure once it is hashable and serializable,
```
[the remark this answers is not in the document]
abernix: It appears that @BrynCooke has opened https://github.com/apollographql/router/pull/5125 already.

.changesets/fix_geal_cache_key_qp_options.md
BrynCooke: Yes, it's flagging to users of the old versions that they must change the namespace

apollo-router/src/query_planner/caching_query_planner.rs
```diff
@@ -530,8 +561,11 @@ pub(crate) struct CachingQueryKey {
     pub(crate) hash: Arc<QueryHash>,
     pub(crate) metadata: CacheKeyMetadata,
     pub(crate) plan_options: PlanOptions,
+    pub(crate) config_mode: ConfigMode,
 }
 
+// Update this key every time the cache key or the query plan format has to change
```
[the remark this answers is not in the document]
BrynCooke: Yes fed version is included.

`CACHE_KEY_VERSION` seems like it might be a footgun that will hurt us if we rely on it, as it is pretty much guaranteed that we will forget to update it when we have to.

Federation version takes care of query plan format, so the sorts of things we have left are mutations to the query plan by plugins such as file uploads. This isn't an issue right now as query plan rewrite happens every time, but if we wanted to cache the rewrites then the possibility of us getting this right decreases rapidly. What if the config in file uploads changes the way that query plans are rewritten? This is a rabbit hole we don't want to go down. 

Once the rust query planner is in place and optimized it'd be good to revisit the entire query plan cache store to see how much benefit we get from caching, and if we still need it what sort of investment and limits we will place on it. For instance we could mandate that query plan rewrites will *never* be cached which would significantly reduce the scope of the problem.

.changesets/fix_geal_cache_key_qp_options.md
BrynCooke: [4968a88](https://github.com/apollographql/router/pull/5100/commits/4968a88a5d5d4d351b9df67b1c241bdcfaf07f94)

apollo-router/src/query_planner/caching_query_planner.rs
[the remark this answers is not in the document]
BrynCooke: Maybe I'm misunderstand the purpose here. It could be just for debugging so that given a cache key we know how to interpret it.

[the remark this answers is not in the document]
BrynCooke: On the flip side, maybe we should have self describing keys: https://stackoverflow.com/questions/6320739/does-name-length-impact-performance-in-redis

apollo-router/src/query_planner/caching_query_planner.rs
```diff
@@ -545,23 +579,27 @@ impl std::fmt::Display for CachingQueryKey {
...
+            .update(&serde_json::to_vec(&self.config_mode).expect("serialization should not fail"));
         hasher.update(&serde_json::to_vec(&self.sdl).expect("serialization should not fail"));
         let metadata = hex::encode(hasher.finalize());
 
         write!(
             f,
-            "plan:{}:{}:{}:{}",
-            FEDERATION_VERSION, self.hash, operation, metadata,
+            "plan:{}:{}:{}:{}:{}",
+            CACHE_KEY_VERSION, FEDERATION_VERSION, self.hash, operation, metadata,
```
IvanGoncharov, line 590: It's unrelated to this particular PR, but it took me a while to understand that `operation` is a hash of `self.operation`, which is actually an operation name.

Also, `self.hash` is hash of query and relevent parts of schema, but taking into account only operation matching `operation_name`.
At the same time, `self.hash` doesn't include a hash of "operation_name" itself

IvanGoncharov: Also `self.metadata` is actually "authorization metadata".

P.S. Same as the previous comment, this is unrelated to this PR.
Just comments for others to also understand what is happening in this code.

apollo-router/src/query_planner/caching_query_planner.rs
```diff
@@ -530,8 +561,11 @@ pub(crate) struct CachingQueryKey {
     pub(crate) hash: Arc<QueryHash>,
     pub(crate) metadata: CacheKeyMetadata,
     pub(crate) plan_options: PlanOptions,
+    pub(crate) config_mode: ConfigMode,
 }
 
+// Update this key every time the cache key or the query plan format has to change
+const CACHE_KEY_VERSION: usize = 0;
 const FEDERATION_VERSION: &str = std::env!("FEDERATION_VERSION");
```
IvanGoncharov, line 569: Since this value is coming from environment variable maybe it's worth to sanitize it for `:` symbols before putting it in Redis key.
Not a big issue since version shouldn't contain `:`, but maybe worth to do it just in case.

apollo-router/src/query_planner/caching_query_planner.rs
BrynCooke: I'd like to see this refactored and keys become self describing.

apollo-router/src/query_planner/caching_query_planner.rs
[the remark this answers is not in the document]
Geal: I added `CACHE_KEY_VERSION` following feedback on the other issue with schema hashing, to cover the case where we would change anything related to query plan caching and want to ignore the previously cached entries as a safety measure.

apollo-router/src/query_planner/caching_query_planner.rs
Geal: this environment variable is set by the build script: https://github.com/apollographql/router/blob/dev/apollo-router/build/main.rs#L34

apollo-router/src/query_planner/caching_query_planner.rs
```diff
@@ -545,23 +588,28 @@ impl std::fmt::Display for CachingQueryKey {
         hasher.update(
             &serde_json::to_vec(&self.plan_options).expect("serialization should not fail"),
         );
+        hasher
+            .update(&serde_json::to_vec(&self.config_mode).expect("serialization should not fail"));
         hasher.update(&serde_json::to_vec(&self.sdl).expect("serialization should not fail"));
+        hasher.update([self.introspection as u8]);
```
Geal, line 594: why is introspection added here but not in the Hash implementation below?

apollo-router/tests/integration/mod.rs
```diff
@@ -7,6 +7,8 @@ mod docs;
 mod file_upload;
 mod lifecycle;
 mod operation_limits;
+
+#[cfg(any(not(feature = "ci"), all(target_arch = "x86_64", target_os = "linux")))]
```
Geal, line 11:
```suggestion
```
part of the larger issue here is that nobody on the team is testing the router with a local redis instance. This has to change

Geal:
thank you all for the feedback, I will make another pass today or tomorrow to fix up the small issues. @BrynCooke do you want to see the hash prefxies in this PR or in a follow up? I plan to make another PR right after to fix other things, like removing the JSON usage when possible

apollo-router/tests/integration/mod.rs
BrynCooke: This change does exactly that. The previous logic disabled for platforms, but now it's only disabled for CI on specific platforms.

Developers running locally will run all the tests:

#[cfg(**any(not(feature = "ci"),** all(target_arch = "x86_64", target_os = "linux")))]

BrynCooke:
@Geal Happy for the key change to be a follow up.

apollo-router/src/query_planner/caching_query_planner.rs
BrynCooke: Nice catch: [43ccd9d](https://github.com/apollographql/router/pull/5100/commits/43ccd9d21b7d6db5e4c16d9d6a2107169e202d57)

I am wondering though why we are using a manual implementation of `Hash` rather than using derive? There seems to be nothing special about this structure that would warrant a custom impl.

Geal: I'd have to dig it up, but I think there was something in the key that could not implement `Hash` in the past

apollo-router/tests/integration/mod.rs
Geal: ah, got it, thanks!

apollo-router/tests/common.rs
```diff
@@ -1005,5 +1091,20 @@ fn merge_overrides(
...
+        .and_then(|o| o.get_mut("supergraph"))
+        .and_then(|o| o.as_object_mut())
+        .and_then(|o| o.get_mut("query_planning"))
+        .and_then(|o| o.as_object_mut())
+        .and_then(|o| o.get_mut("cache"))
+        .and_then(|o| o.as_object_mut())
+        .and_then(|o| o.get_mut("redis"))
+        .and_then(|o| o.as_object_mut())
+    {
+        query_plan.insert("namespace".to_string(), redis_namespace.into());
```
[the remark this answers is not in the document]
Geal: fixed in [17addbc](https://github.com/apollographql/router/pull/5100/commits/17addbcb21bb4e3f4aa263fde2788e3cf2b35459)[17addbc](https://github.com/apollographql/router/pull/5100/commits/17addbcb21bb4e3f4aa263fde2788e3cf2b35459)

Geal:
@abernix can we merge this?

apollo-router/src/query_planner/caching_query_planner.rs
abernix: Once we figure that out, is there somewhere we could add a comment to make that more intuitive?

bnjjj (approved):

apollo-router/src/query_planner/caching_query_planner.rs
BrynCooke: If we don't need a custom implementation now let's remove it and use derive. That way we don't have to worry about it in the future. If we do eventually need a custom version would need to comment and possibly fuzz test.

Geal:
yes it supersedes that PR

apollo-router/src/query_planner/caching_query_planner.rs
```diff
@@ -545,23 +589,29 @@ impl std::fmt::Display for CachingQueryKey {
...
         let metadata = hex::encode(hasher.finalize());
 
         write!(
             f,
-            "plan:{}:{}:{}:{}",
-            FEDERATION_VERSION, self.hash, operation, metadata,
+            "plan:{}:{}:{}:{}:{}",
+            CACHE_KEY_VERSION, FEDERATION_VERSION, self.hash, operation, metadata,
```
fotoetienne, line 601: Unrelated to this PR: Why are `operation` and `metadata` separate vs combined in the same hasher?

fotoetienne (approved):

apollo-router/src/query_planner/caching_query_planner.rs
Geal: that was a way to see how many variants there can be in cache for a single query

Geal:
follow up issue with the fixes we discussed: https://github.com/apollographql/router/issues/5160

PR #5102, opened 2024-05-07
apollo-federation: remove unused impl ConditionResolver for CachingConditionResolver

lrlna:
Since `QueryPlanningTraversal` directly implements `ConditionResolver` trait, let's remove the todo stub. Should composition need it in the future (I think it will do the same as `QueryPlanningTraversal` though), we can include it again.

Changelog is not necessary, as this does not affect the router.

### DIFF
--- a/apollo-federation/src/query_graph/condition_resolver.rs
+++ b/apollo-federation/src/query_graph/condition_resolver.rs
@@ -54,19 +54,6 @@ impl ConditionResolution {
 //            But in Rust, `QueryPlanningTraversal` implements `ConditionResolver` trait itself,
 //            using `ConditionResolverCache` struct below.
 //            However, `CachingConditionResolver` may still be used in the composition in the future.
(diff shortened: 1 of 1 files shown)


SimonSapin (approved):

apollo-federation/src/query_graph/condition_resolver.rs
```diff
@@ -54,19 +54,6 @@ impl ConditionResolution {
 //            But in Rust, `QueryPlanningTraversal` implements `ConditionResolver` trait itself,
 //            using `ConditionResolverCache` struct below.
 //            However, `CachingConditionResolver` may still be used in the composition in the future.
```
goto-bus-stop, line 56: We should remove the comment then, or move it to the top of the file ("This file doesn't implement CachingConditionResolver, because ...")

goto-bus-stop (approved):

PR #5103, opened 2024-05-07
Reconcile `dev` after merge to `main` for v1.46.0

BrynCooke:
Follow-up to the v1.46.0 being officially released, bringing version bumps and changelog updates into the `dev` branch.


BrynCooke:
This needs a real PR as the router yaml schema snapshot changed.

PR #5109, opened 2024-05-07
chore(tests): provide more threads to each integration test in nextest config

BrynCooke:
Nextest by default will try to run a test on each CPU core available. However this isn't suitable for integration tests as the router will be using multiple threads. This is leading to many test failures in CI.  

In addition the scaffold test really shouldn't be run at the same time as any other test as it invokes rustc. This manifests as tests taking 3 minutes or more if they are unlucky enough to get scheduled at the same time.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -15,3 +15,8 @@ path = "junit.xml"
 [[profile.default.overrides]]
 filter = 'package(apollo-router) & kind(test)'
 threads-required = 2
+
+# Scaffold test takes all the test threads as it runs rustc.
+[[profile.default.overrides]]
+filter = 'package(apollo-router-scaffold)'
+threads-required = 'num-test-threads'
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -9,3 +9,9 @@ fail-fast = false
 # Write to output for persistence to CircleCI
 [profile.ci.junit]
 path = "junit.xml"
+
+# Integration tests require more than one thread. The default setting of 1 will cause too many integration tests to run
+# at the same time and causes tests to fail.
+[[profile.default.overrides]]
+filter = 'package(apollo-router) & kind(test)'
+threads-required = 2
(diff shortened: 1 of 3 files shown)


.config/nextest.toml
```diff
@@ -9,3 +9,14 @@ fail-fast = false
...
+
+# Integration tests require more than one thread. The default setting of 1 will cause too many integration tests to run
+# at the same time and causes tests to fail.
+[[profile.default.overrides]]
+filter = 'package(apollo-router) & kind(test)'
+threads-required = 2
+
+# Scaffold test takes all the test threads as it runs rustc.
+[[profile.default.overrides]]
+filter = 'package(apollo-router-scaffold)'
```
abernix, line 21: Does this need to be called out as an exception that exists somewhere around the actual `apollo-router-scaffold` file?

.config/nextest.toml
```diff
@@ -9,3 +9,14 @@ fail-fast = false
 # Write to output for persistence to CircleCI
 [profile.ci.junit]
 path = "junit.xml"
+
+# Integration tests require more than one thread. The default setting of 1 will cause too many integration tests to run
+# at the same time and causes tests to fail.
+[[profile.default.overrides]]
+filter = 'package(apollo-router) & kind(test)'
```
abernix, line 16: Does this classify it tightly enough to be _just_ an integration test?

abernix (approved):
Left some comments below, but not blocking.

.config/nextest.toml
BrynCooke: [169faa2](https://github.com/apollographql/router/pull/5109/commits/169faa27ee4b3b1b904a7922f3b1c254a3e0f68b)

.config/nextest.toml
BrynCooke: [169faa2](https://github.com/apollographql/router/pull/5109/commits/169faa27ee4b3b1b904a7922f3b1c254a3e0f68b)

goto-bus-stop (approved):

PR #5110, opened 2024-05-07
Add a query plan snapshot test harness

SimonSapin:
This allows porting tests such as found in `query-planner-js/src/__tests__/buildPlan.test.ts`


apollo-federation/tests/query_plan/build_query_plan_support.rs
```diff
@@ -0,0 +1,169 @@
...
+const IMPLICIT_LINK_DIRECTIVE: &str = r#"@link(url: "https://specs.apollo.dev/federation/v2.6", import: ["@key", "@requires", "@provides", "@external", "@tag", "@extends", "@shareable", "@inaccessible", "@override", "@composeDirective", "@interfaceObject"])"#;
+
+/// Runs composition on the given subgraph schemas and return `(api_schema, query_planner)`
+///
+/// Results of composition are cached in `tests/query_plan/supergraphs`.
+/// When needed, composition is done by starting a Rover subprocess
+/// (this requires a recent-enough version of `rover` to be in `$PATH`)
+/// but only if the `USE_ROVER=1` env variable is set.
+///
+/// Panics if composition is needed but `USE_ROVER` is unset.
```
lrlna, line 20: Can you add a few lines of dev docs on running these snapshots?

SimonSapin: What do you mean by running snapshots? Tests using this functionality are run like any other test with `cargo test` or `cargo nextest run`. If a cached supergraph is missing or out of date, the test will fail with a message that tells you what to do. (That is, run the test again with `USE_ROVER=1`.)

lrlna (approved):

apollo-federation/tests/query_plan/build_query_plan_support.rs
lrlna: What I mean is adding a note about needing rover to be installed and explaining why and when it'll be necessary.

lrlna: It looks like we don't have a lot of docs on existing testing infrastructure, so I guess we can add this later when we document it all together.

PR #5111, opened 2024-05-07
Adjust data structures in fetch dependency graphs to avoid clones

sachindshinde:
This PR adjusts for source-aware fetch dependency graphs and query plans to avoid clones (it adds some `Arcs`s around `Vec`s we expect to clone a lot, and uses `Node` around `Value`)

### DIFF
--- a/apollo-federation/src/source_aware/fetch_dependency_graph/mod.rs
+++ b/apollo-federation/src/source_aware/fetch_dependency_graph/mod.rs
@@ -32,7 +32,7 @@ type FetchDependencyGraphPetgraph =
 
 #[derive(Debug)]
 pub(crate) struct FetchDependencyGraphNode {
-    merge_at: Vec<FetchDataPathElement>,
+    merge_at: Arc<Vec<FetchDataPathElement>>,
     source_entering_edge: EdgeIndex,
     operation_variables: IndexSet<Name>,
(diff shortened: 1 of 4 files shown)


apollo-federation/src/source_aware/fetch_dependency_graph/mod.rs
```diff
@@ -32,7 +32,7 @@ type FetchDependencyGraphPetgraph =
 
 #[derive(Debug)]
 pub(crate) struct FetchDependencyGraphNode {
-    merge_at: Vec<FetchDataPathElement>,
+    merge_at: Arc<Vec<FetchDataPathElement>>,
```
goto-bus-stop, line 35: Note if you have an immutable list here (which I think is the case here), you can instead store a `Arc<[FetchDataPathElement]>`. Smart pointers like Arcs can contain a slice directly, and that means there's one less pointer indirection and one less allocation, and also less unnecessary bookkeeping (since a Vec tracks a capacity and allocates excess space to grow into). You can use `Arc::from(vec)` to convert the types, which doesn't copy, it moves the storage from the vec into the Arc.

goto-bus-stop: FWIW I don't think this is that expensive to clone to begin with so I'm not totally convinced that it needs to be in an Arc. As long as it's immutable there isn't any significant downside to having it be an Arc though.

PR #5116, opened 2024-05-08
Add guidance for adding new metrics

BrynCooke:
Until now there has been little guidance on adding new metrics to the router. This PR expands the dev doc to include this.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


dev-docs/metrics.md
```diff
@@ -179,3 +179,85 @@ Strong references to instruments will be discarded when changes to the aggregate
...
+* Look at the [OTel semantic conventions](https://opentelemetry.io/docs/specs/semconv/general/metrics/) 
+* Notify `#proj-router-analytics` channel in Slack.
+* Add the metrics to the spreadsheet linked in the `#proj-router-analytics` channel in Slack.
+
+#### Operation counts
+Each new feature MUST have an operation count metric that counts the number of requests that the feature has processed.
+
+When defining new operation metrics use the following conventions:
+
+**Name:** `apollo.router.operations.<feature>` - (counter)
```
abernix, line 213: We should get **more** crisp here — (e.g., feature as a singular noun vs what it does — `query_planner` vs `query_planning`).

dev-docs/metrics.md
```diff
@@ -179,3 +179,85 @@ Strong references to instruments will be discarded when changes to the aggregate
...
+> Note that even if a feature is experimental this should not be reflected in the metric name.
+
+**Attributes:**
+* `<feature>.<feature-specific-attribute>` - (usually a boolean or number, but can be a string if the set of possible values is fixed)
+
+> :warning: **Remember that attributes are not to be used to store high cardinality or user specific information. Operation name is not permitted!**
```
abernix, line 219: The recommended Markdown way of doing this is now as follows, per https://github.com/orgs/community/discussions/16925

```suggestion
> [!WARNING]
> **Remember that attributes are not to be used to store high cardinality or user specific information. Operation name is not permitted!**
```

dev-docs/metrics.md
```diff
@@ -179,3 +179,98 @@ Strong references to instruments will be discarded when changes to the aggregate
...
+There are different types of metrics.
+
+* Static - Used by us to monitor feature usage.
+* Dynamic - Defined by the user to monitor the health and performance of their system.
+
+> New features should add BOTH static and dynamic metrics.
+
+> Why are static metrics no longer recommended for users to use directly?
+> 
+> They can, but usually it'll be only a starting point for them. We can't predict the things that users will want to monitor, and if we tried we would blow up the cardinality of our metrics resulting in high costs for our users via their APMs.
```
Geal, line 193: that is not clear to me. What do we mean by "users using static metrics directly?" Is it when they would add that in their custom plugin? (which would not increase cardinality for all users) Or asking us to add a new metric to the router?

dev-docs/metrics.md
```diff
@@ -179,3 +179,98 @@ Strong references to instruments will be discarded when changes to the aggregate
 On the fast path the mutex is locked for the period that it takes to upgrade the weak reference. This is a fast operation, and should not block the thread for any meaningful period of time.
 
 If there is shown to be contention in future profiling we can revisit.
+
+## Adding new metrics
+There are different types of metrics.
+
+* Static - Used by us to monitor feature usage.
```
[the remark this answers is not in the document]
Geal: let's assume router users will end up looking at the dev docs

dev-docs/metrics.md
```diff
@@ -179,3 +179,98 @@ Strong references to instruments will be discarded when changes to the aggregate
...
+* `<feature>` - This should be a noun that describes the feature that the metric is monitoring.
+
+* `<feature>.<verb>` - Sub-metrics are usually a verb that describes the action that the metric is monitoring.
+
+**Attributes**
+* `<feature>.<feature-specific-attribute>` - Are always prefixed with the feature name unless they are standard metrics from the [otel semantic conventions](https://opentelemetry.io/docs/specs/semconv/general/metrics/). 
+
+### Static metrics
+When adding a new feature to the Router you must also add new static metrics to monitor the usage of that feature and users cannot turn them off.
+These metrics must be low cardinality and not leak any sensitive information. Users cannot change these metrics and they are primarily for us to see how our features are used so that we can inform future development.
```
Geal, line 210: a lot of static metrics actually monitor standard router operations and are not for us to collect data, but for users to observe the router.
If we want this to be the defining point, let's maybe not call them `static` VS `dynamic` metrics, but `internal` VS `monitoring` or `user` metrics, something like that?
I'd prefer we keep the distinction between `static` metrics as defined directly with `tracing`, and `dynamic` metrics as the ones defined by custom instruments, that can be activated with runtime conditions, and have another clear separation beween the metrics used for internal reporting (as with the `apollo.router.operations` and `apollo.router.config` prefixes) and the user facing ones.

dev-docs/metrics.md
```diff
@@ -179,3 +179,98 @@ Strong references to instruments will be discarded when changes to the aggregate
...
+These metrics are transmitted to Apollo unless explicitly disabled.
+
+When adding new static metrics and attributes make sure to:
+* Include them in your design document.
+* Look at the [OTel semantic conventions](https://opentelemetry.io/docs/specs/semconv/general/metrics/) 
+* Notify `#proj-router-analytics` channel in Slack.
+* Add the metrics to the spreadsheet linked in the `#proj-router-analytics` channel in Slack.
+
```
Geal, line 218: code example of what is a static metric, to be sure which is which between static and dynamic?

dev-docs/metrics.md
```diff
@@ -179,3 +179,98 @@ Strong references to instruments will be discarded when changes to the aggregate
...
+#### Operation counts
+Each new feature MUST have an operation count metric that counts the number of requests that the feature has processed.
+
+When defining new operation metrics use the following conventions:
+
+**Name:** `apollo.router.operations.<feature>` - (counter)
```
Geal, line 224: some of the `apollo.router.operations` metrics are actually monitored by users. What is the strategy here? Do we keep them available for users?

Geal (changes requested):
a lot of what is encoded in this document is unclear to me, I think we should discuss it a bit more

dev-docs/metrics.md
BrynCooke: This is pretty much what I was trying to say:
* Static metrics defined by macro (not `tracing` as we are going to remove this), 
* Dynamic metrics configurable by yaml.

I think the rub though is that I'm trying to define where static metrics are best used.

The problem with static metrics is that attributes that we think are OK are not, and attributes that we don't include are required. For instance:

1. Feature X needs some metrics. We put some static metrics in using a macro.
2. User A: complains that they need operation name on the metric to identify which operations are causing an issue.
3. We add `graphql.operation.name` to a static metric.
4. User B: complains that they have now had their APM cardinality budget blown because they have unrestricted operation names and therefore they are high cardinality.

Ideal static metrics uses:
* Router lifecycle stuff.
* Global error rates.
* Feature usage.
* Stuff that happens outside of a request/response. Cache clears

Ideal dynamic metrics uses:
* Anything that happens during a request response.

It's not so much a question of internal vs external, so I'll change the text to reflect the above and remove: `Users cannot change these metrics and they are primarily for us to see how our features are used so that we can inform future development.`

BrynCooke: d233cf3bdf553f308d2702f47f3c79312ec0bfdc

dev-docs/metrics.md
BrynCooke: I'm not in favour of hiding them. Apart from anything we are transmitting these metrics to apollo, and therefore we need to be transparent about what they contain.

WDYT?

dev-docs/metrics.md
BrynCooke: [df05add](https://github.com/apollographql/router/pull/5116/commits/df05add809a8dade4a720b338a2cbc1ef082ea1e)

abernix:
This seems like it would be great to land.  Do we think we could continue to talk about this more? Perhaps next week?

BrynCooke:
@timbotnik I'd like to resurrect this PR. I think it needs examples, but the content is still relevant. Can you add your thoughts?

dev-docs/metrics.md
```diff
@@ -234,3 +234,115 @@ Strong references to instruments will be discarded when changes to the aggregate
...
+**Metrics**
+* `<feature>` - This should be a noun that describes the feature that the metric is monitoring.
+
+* `<feature>.<verb>` - Sub-metrics are usually a verb that describes the action that the metric is monitoring.
+
+**Attributes**
+* `<feature>.<feature-specific-attribute>` - Are always prefixed with the feature name unless they are standard metrics from the [otel semantic conventions](https://opentelemetry.io/docs/specs/semconv/general/metrics/). 
+
+### Static metrics
+When adding a new feature to the Router you must also add new static metrics to monitor the usage of that feature and users cannot turn them off.
```
bnjjj, line 264: It's almost true. They could still disable it using `views`

dev-docs/metrics.md
```diff
@@ -234,3 +234,115 @@ Strong references to instruments will be discarded when changes to the aggregate
...
+These metrics must be low cardinality and not leak any sensitive information. Users cannot change the attributes that are attached to these metrics.
+These metrics are transmitted to Apollo unless explicitly disabled.
+
+When adding new static metrics and attributes make sure to:
+* Include them in your design document.
+* Look at the [OTel semantic conventions](https://opentelemetry.io/docs/specs/semconv/general/metrics/) 
+* Notify `#proj-router-analytics` channel in Slack.
+* Add the metrics to the spreadsheet linked in the `#proj-router-analytics` channel in Slack.
```
bnjjj, line 272: Is it still accurate ?

dev-docs/metrics.md
```diff
@@ -234,3 +234,115 @@ Strong references to instruments will be discarded when changes to the aggregate
...
+
+To define a static metric us a macro:
+```rust
+u64_counter!("apollo.router.<feature>.<verb>", "description", 1, "attr" => "val");
```
bnjjj, line 276: We now have a new kind of macros to specify unit, it's `u64_counter_with_unit` I think and we should always provide units

BrynCooke: We should also remove any non unit macros

BrynCooke: [35575dc](https://github.com/apollographql/router/pull/5116/commits/35575dc7daf9e616d2b5e8d86c757857f0c1f375)

dev-docs/metrics.md
BrynCooke: I'm loosened this up as the exact method of discussing new metrics may change over time: [35575dc](https://github.com/apollographql/router/pull/5116/commits/35575dc7daf9e616d2b5e8d86c757857f0c1f375)

dev-docs/metrics.md
BrynCooke: I added clarification [35575dc](https://github.com/apollographql/router/pull/5116/commits/35575dc7daf9e616d2b5e8d86c757857f0c1f375)

bnjjj (approved):

PR #5117, opened 2024-05-08
chore(federation): replace selection_of_element uses by Selection::from_element constructor

goto-bus-stop:
Compare implementations:
https://github.com/apollographql/router/blob/9921dcd8abdf889e69e69b135c169e01b7db1251/apollo-federation/src/query_graph/graph_path.rs#L412-L431
https://github.com/apollographql/router/blob/9921dcd8abdf889e69e69b135c169e01b7db1251/apollo-federation/src/query_plan/operation.rs#L688-L706


goto-bus-stop:
Oh wait, Duckki is already doing this in https://github.com/apollographql/router/pull/5082. nevermind

PR #5126, opened 2024-05-09
add router reload metrics

xuorig:
This PR adds a new metric, `apollo_router_reload` to track when the router **actually** reloads, versus only having state change events.

This metric has tags to verify which event caused the reload, as well as a `success` tag. Let me know what you think!

### DIFF
--- a/apollo-router/src/state_machine.rs
+++ b/apollo-router/src/state_machine.rs
@@ -229,7 +229,7 @@ impl<FA: RouterSuperServiceFactory> State<FA> {
                     .await;
 
                     u64_counter!(
-                        "apollo_router_reload",
+                        "apollo.router.reload",
                         "Amount of times the router reloaded",
                         1,
                         new_schema = schema_reload,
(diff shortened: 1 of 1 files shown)


apollo-router/src/state_machine.rs
```diff
@@ -226,8 +226,19 @@ impl<FA: RouterSuperServiceFactory> State<FA> {
                         &mut guard,
                         signals,
                     )
-                    .await
-                    {
+                    .await;
+
+                    u64_counter!(
+                        "apollo_router_reload",
```
BrynCooke, line 232:
```suggestion
                        "apollo.router.reload",
```
Let's use `.` rather than underscore as this is in line with otel standards.

BrynCooke:
@shorgi Can you create a changelog and docs for this new metric? Once we have these we can merge.

xuorig:
Let me know if I can help with the changelog and docs

Geal (approved):
nice idea, thanks

PR #5135, opened 2024-05-10
chore(federation): add and use some more ergonomic operation types constructors

goto-bus-stop:
Over time we added some shorthand constructors for operation types, but we also still constructed many things manually. This PR uses some of the existing constructors instead of writing out the whole struct, and adds several more.

In particular I think the spots that had a lot of `)))})` nesting are now easier to read.

The constructors on `Selection` types are also aimed towards usage in tests.

This is a bit of a weird PR since it doesn't have a clear end goal but I already made these changes on another branch to write a test that turned out to be unnecessary. It will also help with tests in https://github.com/apollographql/router/pull/5119.


dariuszkuc (approved):

duckki:
nice!

PR #5139, opened 2024-05-10
fix(federation): Fix rebasing, fix creating fetch nodes

goto-bus-stop:
Ooookay. It produces a query plan that does key resolution across multiple subgraphs.

Conditions from `@key(fields:)` and other directives are parsed in the context of their subgraph schema. That means we need to rebase them onto the supergraph schema to add them to input selections for fetch dependency nodes. Additionally there was a bug in rebasing that caused the selection set to reference the old schema.

Producing the final FetchNodes relied on `impl From<SelectionSet> for executable::SelectionSet`, which was marked as `todo!()`. We already have a `TryFrom<&SelectionSet>`, so use that instead.

**There are more bugs**: this does **not** appear to actually fetch the `name` field that I asked for. If I find the cause quickly I'll push a fix here otherwise let's do that in a separate PR

### DIFF
--- a/apollo-federation/src/query_plan/fetch_dependency_graph.rs
+++ b/apollo-federation/src/query_plan/fetch_dependency_graph.rs
@@ -2019,9 +2019,17 @@ fn compute_nodes_for_key_resolution<'a>(
             "missing expected edge conditions",
         ));
     };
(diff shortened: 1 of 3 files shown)


dariuszkuc (approved):

duckki:
🎉

PR #5141, opened 2024-05-10
Dylan/validate connectors

dylan-apollo:
A first crack at validations for `@source`. This new code is really only designed to be consumed in `@apollo/federation-internals` via WASM—and there's a working prototype of that [here](https://github.com/apollographql/federation/blob/connectors/internals-js/wasm/src/lib.rs).


apollo-federation/src/sources/connect/models/validation.rs
```diff
@@ -49,3 +39,225 @@
...
+///
+/// This function attempts to collect as many validation errors as possible, so it does not bail
+/// out as soon as it encounters one.
+pub fn validate(schema: Schema) -> Vec<ValidationError> {
+    let mut source_names = Vec::new();
+    let errors = schema
+        .schema_definition
+        .directives
+        .iter()
+        .filter_map(|directive| {
```
nicholascioli, line 59: Nit: I think that it might be more clear to have a `filter` followed by a `map` here to help break apart the nesting.

apollo-federation/src/sources/connect/models/validation.rs
```diff
@@ -49,3 +39,225 @@
...
+            if directive.name == "source" {
+                match validate_source(directive) {
+                    Ok(source_directive) => {
+                        if let SourceName::Valid(name) = source_directive.name {
+                            source_names.push(name);
```
nicholascioli, line 64: Nit: I would avoid having functional components with side-effects, if possible. You could try using a fold or something and having the map return a tuple, if you prefer.

apollo-federation/src/sources/connect/models/validation.rs
```diff
@@ -49,3 +39,225 @@
...
+    fn try_from(directive: &Component<Directive>) -> Result<Self, Self::Error> {
+        let str_value = directive
+            .arguments
+            .iter()
+            .find(|arg| arg.name == "name")
+            .ok_or(ValidationError::MissingSourceName)?
+            .value
+            .as_str()
+            .ok_or(ValidationError::SourceNameType)?;
+        if str_value.chars().all(|c| c.is_alphanumeric() || c == '_') {
```
nicholascioli, line 165: Does this mean that an empty string is allowed as a name?

apollo-federation/src/sources/connect/models/validation.rs
```diff
@@ -49,3 +39,225 @@
...
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum ValidationError {
+    #[error("baseURL argument for @source \"{source_name}\" was not a valid URL: {inner}")]
+    SourceUrl {
+        inner: url::ParseError,
+        source_name: SourceName,
+    },
+    #[error("baseURL argument for @source \"{source_name}\" be http or https, got {scheme}")]
```
nicholascioli, line 188:
```suggestion
    #[error("baseURL argument for @source \"{source_name}\" must be http or https, got {scheme}")]
```

apollo-federation/src/sources/connect/models/validation.rs
```diff
@@ -49,3 +39,225 @@
...
+#[cfg(test)]
+mod test_validate_source {
+    use super::*;
+
+    #[test]
+    fn unparseable_url() {
+        let schema = include_str!("test_data/invalid_source_url.graphql");
+        let schema = Schema::parse(schema, "test.graphql").unwrap();
+        let errors = validate(schema);
+        assert_eq!(errors.len(), 1);
```
nicholascioli, line 214: Nit: It might be good to check that the inner values of these tests include the right information vs just checking that they are of the right type.

apollo-federation/src/sources/connect/models/validation.rs
dylan-apollo: This will inevitably change when it becomes more complex once connect directives are also added—since there's going to be cross-validation. So I'll wait to clean up this code until we see what it needs.

apollo-federation/src/sources/connect/models/validation.rs
dylan-apollo: It's not! Nice catch, will add in that check as well.

apollo-federation/src/sources/connect/models/validation.rs
dylan-apollo: Right now `=="source"` _looks_ like a filter, but there's going to be a `=="connect"` branch as well—it's really dispatching. That said, like the side-effects discussion, I think we'll best know how to clean this up once most of the top-level logic is in.

apollo-federation/src/sources/connect/models/validation.rs
dylan-apollo: I think I want to enhance these with a snapshot mechanism so it doesn't get too tedious to update the source schema files for tests. I _think_ insta can do inline snapshots, I'll do that as a follow-up.

apollo-federation/src/sources/connect/models/validation.rs
```diff
@@ -49,3 +39,225 @@
...
+
+impl TryFrom<&Component<Directive>> for SourceName {
+    type Error = ValidationError;
+
+    fn try_from(directive: &Component<Directive>) -> Result<Self, Self::Error> {
+        let str_value = directive
+            .arguments
+            .iter()
+            .find(|arg| arg.name == "name")
+            .ok_or(ValidationError::MissingSourceName)?
```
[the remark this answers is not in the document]
dylan-apollo: Sure, we can collapse them at this level. Ideally, once more moves into Rust, we can use the type system to enforce stuff somehow so we _guarantee_ that certain errors can't occur at this stage.

dylan-apollo:
Thank you both—suggestions applied!

apollo-federation/src/sources/connect/models/validation.rs
```diff
@@ -49,3 +55,254 @@
...
+                }
+            } else {
+                None
+            }
+        })
+        .flatten()
+        .collect_vec();
+
+    if source_names.is_empty() && errors.is_empty() {
+        todo!("Check for renamed directives—this code is _only_ called for connectors subgraphs")
```
lennyburdette, line 101: i think we're going to allow using `@connect(http: { GET: "https://whatever" })` without a source! (i'm hoping it makes the initial demo/evaluation a touch easier.) so an empty list of sources is expected.

apollo-federation/src/sources/connect/models/validation.rs
lennyburdette: inline snapshots are easy! `assert_snapshot(whatever, @"")`

apollo-federation/src/sources/connect/models/validation.rs
dylan-apollo: Oh, right. Okay I moved the TODO into the big block of TODOs instead 😅

lennyburdette (approved):

PR #5150, opened 2024-05-13
Try a 1min per-test timeout on CI, with an override for scaffold

SimonSapin:


abernix (commented):
LGTM.  (Seems like a reasonable starting-point, at least!)

abernix (approved):

PR #5151, opened 2024-05-13
fix(federation): fix `FetchDependencyGraph::reduce()`

goto-bus-stop:
This fixes the missing plan nodes in the output in the initial multi-subgraph test.

Consider a fetch dependency graph with nodes A, B, C and edges 1, 2, 3:

```
A -1> B -2> C
 \----3--->/
```

`reduce()` should remove edge `3` as you can already reach `C` by
following edge `1` and `2`.

The bug in this code was instead removing edge `2` in this situation. We
have to go through grandchildren and all nested children of the
grandchildren for each node and remove edges to the original node.


apollo-federation/src/query_graph/graph_path.rs
```diff
@@ -271,7 +271,7 @@ impl std::fmt::Display for OpPath {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
         for (i, element) in self.0.iter().enumerate() {
             if i > 0 {
-                write!(f, ", ")?;
+                write!(f, "::")?;
```
lrlna, line 274: oh, is the delimiter change the one we made for type-conditioned fetching?

lrlna (approved):

apollo-federation/src/query_graph/graph_path.rs
goto-bus-stop: Not sure tbh, I saw the JS use `::` as a path delimiter somewhere, and it is more readable for both operation paths and fetch paths in my opinion. Note this is only every used for println debugging and is not user-facing

apollo-federation/src/query_plan/fetch_dependency_graph.rs
```diff
@@ -770,10 +770,11 @@ impl FetchDependencyGraph {
 
     /// Find redundant edges coming out of a node. See `remove_redundant_edges`.
     fn collect_redundant_edges(&self, node_index: NodeIndex, acc: &mut HashSet<EdgeIndex>) {
+        let mut stack = vec![];
         for start_index in self.children_of(node_index) {
-            let mut stack = self.children_of(start_index).collect::<Vec<_>>();
+            stack.extend(self.children_of(start_index));
             while let Some(v) = stack.pop() {
-                for edge in self.graph.edges_connecting(start_index, v) {
+                for edge in self.graph.edges_connecting(node_index, v) {
```
goto-bus-stop, line 777: This one line is the actual fix, the rest is debug printing improvements and reusing the allocation for `stack`

PR #5154, opened 2024-05-13
chore(federation): Un-arc SelectionMap 

goto-bus-stop:
SelectionMap is not used as a standalone type independent of
SelectionSet. Keeping it in an Arc makes SelectionSet methods harder to
implement, and doesn't save any cloning that I can find. It doesn't really
make sense, I think, to have the selection *map* in particular be shared
between different selection *sets*.

Where cloning and structural sharing is important, `Arc<SelectionSet>`
should be used instead.


goto-bus-stop:
I still had this lying around from working on a different PR a few weeks ago. There is actually now a case where the cloning does matter, which is in `.lazy_map()`. This seems a bit overeager. I'll close this and revisit later. What opening a PR and critically looking over it can do for you 😄

PR #5159, opened 2024-05-13
fix: return internal errors as structured error responses

BrynCooke:
When an http 500 is returned we should not return the details of the error to the client. 
Instead, they are now logged at ERROR level.

In addition, the error is now returned as a graphql error rather than a plaintext error. This gives a better experience in sandbox.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/.changesets/fix_bryn_internal_server_error.md
+++ b/.changesets/fix_bryn_internal_server_error.md
@@ -1,8 +1,6 @@
-### Internal server error handling change ([PR #5159](https://github.com/apollographql/router/pull/5159))
+### 5xx Internal server error responses are structured errors ([PR #5159](https://github.com/apollographql/router/pull/5159))
 
-When an http 500 is returned we should not return the details of the error to the client. 
-Instead, they are now logged at ERROR level.
+An internal server error (5xx class) which occurs as a result of an unexpected/unrecoverable disruption to the GraphQL request lifecycle execution (e.g., a coprocessor failure, etc.) will now result in a structured GraphQL error (i.e., `{"errors": [...]}`) being returned to the client rather than a plain-text error as was the case previously.
 
-In addition, the error is now returned as a graphql error rather than a plaintext error, giving a better experience in sandbox.
-
+When these circumstances occur, the underling error message — which may be any number of internal disruptions — will still be logged at an `ERROR` level to the router logs for the administrator of the router to monitor and use for debugging purposes.
 By [@BrynCooke](https://github.com/BrynCooke) in https://github.com/apollographql/router/pull/5159
--- a/.changesets/fix_bryn_internal_server_error.md
+++ b/.changesets/fix_bryn_internal_server_error.md
@@ -1,0 +1,8 @@
+### Internal server error handling change ([PR #5159](https://github.com/apollographql/router/pull/5159))
+
+When an http 500 is returned we should not return the details of the error to the client. 
+Instead, they are now logged at ERROR level.
+
+In addition, the error is now returned as a graphql error rather than a plaintext error, giving a better experience in sandbox.
+
+By [@BrynCooke](https://github.com/BrynCooke) in https://github.com/apollographql/router/pull/5159
--- a/apollo-router/src/axum_factory/axum_http_server_factory.rs
+++ b/apollo-router/src/axum_factory/axum_http_server_factory.rs
@@ -1,4 +1,5 @@
 //! Axum http server factory. Axum provides routing capability on top of Hyper HTTP.
+use std::fmt::Display;
 use std::pin::Pin;
 use std::sync::atomic::AtomicBool;
 use std::sync::atomic::AtomicU64;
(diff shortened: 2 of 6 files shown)


Geal (approved):

Geal:
that's a great idea

apollo-router/src/axum_factory/axum_http_server_factory.rs
```diff
@@ -644,10 +647,7 @@ async fn handle_graphql(
             .in_current_span();
         let res = match tokio::task::spawn(task).await {
             Ok(res) => res,
-            Err(err) => {
-                let msg = format!("router service call failed: {err}");
-                return (StatusCode::INTERNAL_SERVER_ERROR, msg).into_response();
-            }
+            Err(err) => return internal_server_error(err),
```
bnjjj, line 650: I'm just wondering if we could have that kind of critical errors coming before handle_graphql, if so what's the behavior right now ? Do we also want to change this ? And if it's an error coming from a tower layer or so ?

BrynCooke: I think we should open some followups to investigate this and also deal with other non-graphql errors that are escaping in handle_graphql such as rate limiting.

.changesets/fix_bryn_internal_server_error.md
```diff
@@ -0,0 +1,8 @@
+### Internal server error handling change ([PR #5159](https://github.com/apollographql/router/pull/5159))
+
+When an http 500 is returned we should not return the details of the error to the client. 
+Instead, they are now logged at ERROR level.
+
+In addition, the error is now returned as a graphql error rather than a plaintext error, giving a better experience in sandbox.
+
```
abernix, line 7:
```suggestion
### 5xx Internal server error responses are structured errors ([PR #5159](https://github.com/apollographql/router/pull/5159))

An internal server error (5xx class) which occurs as a result of an unexpected/unrecoverable disruption to the GraphQL request lifecycle execution (e.g., a coprocessor failure, etc.) will now result in a structured GraphQL error (i.e., `{"errors": [...]}`) being returned to the client rather than a plain-text error as was the case previously.

When these circumstances occur, the underling error message — which may be any number of internal disruptions — will still be logged at an `ERROR` level to the router logs for the administrator of the router to monitor and use for debugging purposes.
```

apollo-router/src/axum_factory/axum_http_server_factory.rs
abernix: Can we do that before/as we land?

apollo-router/tests/integration/fixtures/broken_coprocessor.router.yaml
```diff
@@ -0,0 +1,6 @@
+# This coprocessor doesn't point to anything
+coprocessor:
+  url: "http://localhost:10000"
```
abernix, line 3: Would `39000` (an impossible port), still recreate this?  I'm worried that `10000` may actually magically appear at some point in this repository. 🤣 
```suggestion
  url: "http://localhost:10000"
```

apollo-router/tests/integration/snapshots/integration_tests__integration__coprocessor__error_not_propagated_to_client.snap
```diff
@@ -0,0 +1,5 @@
+---
+source: apollo-router/tests/integration/coprocessor.rs
+expression: response.text().await?
+---
+"{\"errors\":[{\"message\":\"internal server error\",\"extensions\":{\"code\":\"INTERNAL_SERVER_ERROR\"}}]}"
```
abernix, line 5: Perfect.

apollo-router/tests/integration/fixtures/broken_coprocessor.router.yaml
Geal: We can use a domain that ends in `.example` or `.invalid` instead of localhost, it's guaranteed to not be routable

garypen (approved):

apollo-router/tests/integration/fixtures/broken_coprocessor.router.yaml
BrynCooke: [56112b1](https://github.com/apollographql/router/pull/5159/commits/56112b12e457d7a185049fa1110a53e4c535beec)

BrynCooke: [b15e4f2](https://github.com/apollographql/router/pull/5159/commits/b15e4f2e9dba6beb82f58a9aab1d15c68b86faf3)

apollo-router/src/axum_factory/axum_http_server_factory.rs
BrynCooke: Added https://github.com/apollographql/router/issues/5171

PR #5161, opened 2024-05-14
docs(logging): put the right default values and available configuration for format

bnjjj:


lleadbet:
Changes look good, but we do need to remove `client.address` as well from default spans as it's currently not supported.

abernix (approved):
It appears as if @bnjjj addressed @lleadbet's suggestion even though it wasn't mentioned.  With that done, LGTM.

PR #5162, opened 2024-05-14
Ensure that batch entry contexts are correctly preserved

garypen:
Batch processing was not using contexts correctly. A representative context was chosen, the first item in a batch of items, and used to provide context functionality for all the generated responses.

The router now correctly preserves request contexts and uses them during response creation.

### DIFF
--- a/apollo-router/src/batching.rs
+++ b/apollo-router/src/batching.rs
@@ -425,7 +425,7 @@ pub(crate) async fn assemble_batch(
 ) -> Result<
     (
         String,
-        Context,
+        Vec<Context>,
         http::Request<Body>,
         Vec<oneshot::Sender<Result<SubgraphResponse, BoxError>>>,
     ),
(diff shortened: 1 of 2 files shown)


bnjjj (commented):
Maybe a dumb question but are we sure the ordering of graphql requests is the same and so it's safe to use a Vec instead of a map or something ?
Also could you add tests please ?

garypen:
> Maybe a dumb question but are we sure the ordering of graphql requests is the same and so it's safe to use a Vec instead of a map or something ? Also could you add tests please ?

We are sure (Or at least I am). I did some manual testing to make sure that the ordering was correct. I should have documented that.

Adding tests is challenging, but I'll see what I can think of.

garypen:
I've enhanced an existing unit test and documented the manual testing that I performed.

bnjjj (approved):

PR #5165, opened 2024-05-14
fix(telemetry): do not create custom events if level is off

bnjjj:
Don't create custom events and attributes if you set the level to `off` 

example of configuration:

```yaml
telemetry:
  instrumentation:
    events:
      router:
        # Standard events
        request: info
        response: info
        error: info

        # Custom events
        my.disabled_request_event:
          message: "my event message"
          level: off # Disabled because we set the level to off
          on: request
          attributes:
            http.request.body.size: true
```

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [x] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/.changesets/fix_bnjjj_fix_custom_events_off.md
+++ b/.changesets/fix_bnjjj_fix_custom_events_off.md
@@ -1,0 +1,26 @@
+### Do not create custom events in telemetry if level is off ([PR #5165](https://github.com/apollographql/router/pull/5165))
+
+Don't create custom events and attributes if you set the level to `off` 
+
+example of configuration:
+
+```yaml
+telemetry:
+  instrumentation:
+    events:
+      router:
+        # Standard events
+        request: info
+        response: info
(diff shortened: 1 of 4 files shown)


garypen (commented):
Should we document it?

What I mean is this. We should either:
 - Explicitly enforce our documented interface by deliberately failing use of `off`
 - …OR… embrace the existence of `off` and document/support it.

Either option seems reasonable, but it probably is better to support `off` and document it.

garypen (approved):

BrynCooke (changes requested):
I think this needs a test

bnjjj:
@BrynCooke It's already tested because I updated configuration test and the related snapshots having all the logs didn't change because these events are indeed disabled

BrynCooke (approved):

PR #5169, opened 2024-05-14
Port some more QP tests

SimonSapin:
Part of FED-221, porting some more tests from `query-planner-js/src/__tests__/buildPlan.test.ts`

Most of them are failing and therefore marked with `#[should_panic]`. This attribute is badly named: they really shouldn’t panic, but we expect panics until we follow up by investigating them one by one and fixing relevant bugs until they don’t.

Drive by:

* Use the operation name from the parsed document
* Add a backtrace to `FederationError`: when a test fails on unwrapping such an error run with `RUST_BACKTRACE=1` to see where the error was constructed, not just where it was unwrapped.

---

**Checklist**

- Tests added and passing[^3]
    - x ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-federation/tests/query_plan/build_query_plan_tests/field_covariance_and_type_explosion.rs
```diff
@@ -0,0 +1,150 @@
...
+//! test shows that error can be reproduced on a pure fed2 example, it's just a bit more
+//! complex as we need to involve a @provide just to force the query planner to type explode
+//! (more precisely, this force the query planner to _consider_ type explosion; the generated
+//! query plan still ends up not type-exploding in practice since as it's not necessary).
+
+use apollo_federation::query_plan::query_planner::QueryPlanner;
+
+#[test]
+#[should_panic] // TODO: does this test make sense if we don’t support fed1 supergraphs?
+fn with_federation_1_supergraphs() {
```
lrlna, line 17: nah, since we have the second test with a fed2 supergraph, let's drop this one.

lrlna (approved):
Simonnnnn!!! So good!!

Drop the fed1 tests unless we can manually make them fed2 supergraph tests. Otherwise we are carrying this over forever.

PR #5170, opened 2024-05-14
Port computeRootSerialDependencyGraph

TylerBloom:
This PR addresses FED-127.

The key function that was ported was [`computeRootSerialDependencyGraph`](https://github.com/apollographql/federation/blob/main/query-planner-js/src/buildPlan.ts#L3600). There are two places where I would like feedback on. I've left comments with my notes. Outside those few lines, everything else is ready for review.

### DIFF
--- a/apollo-federation/src/query_plan/query_planner.rs
+++ b/apollo-federation/src/query_plan/query_planner.rs
@@ -499,7 +499,6 @@ impl QueryPlanner {
     }
 }
 
-#[allow(dead_code, unused, unreachable_code, unconditional_panic)]
 fn compute_root_serial_dependency_graph(
     parameters: &QueryPlanningParameters,
     has_defers: bool,
@@ -509,8 +508,6 @@ fn compute_root_serial_dependency_graph(
         supergraph_schema,
         federated_query_graph,
         operation,
-        head,
-        config,
         ..
     } = parameters;
     let root_type: Option<CompositeTypeDefinitionPosition> = if has_defers {
@@ -580,7 +577,6 @@ fn compute_root_serial_dependency_graph(
     Ok(digest)
 }
 
-#[allow(unreachable_code)]
 fn split_top_level_fields(
     selection_set: SelectionSet,
 ) -> Box<dyn 'static + Iterator<Item = SelectionSet>> {
@@ -601,7 +597,7 @@ fn split_top_level_fields(
                     .into_iter()
                     .flat_map(split_top_level_fields)
                     .map(|_set| todo!("Port mapping that uses selectionSetOfElement")),
-                    // return splitTopLevelFields(selection.selectionSet).map(s => selectionSetOfElement(selection.element, s));
+                // return splitTopLevelFields(selection.selectionSet).map(s => selectionSetOfElement(selection.element, s));
             )
         };
         digest
--- a/apollo-federation/src/query_plan/query_planner.rs
+++ b/apollo-federation/src/query_plan/query_planner.rs
@@ -29,8 +29,6 @@ use crate::query_plan::operation::normalize_operation;
 use crate::query_plan::operation::NamedFragments;
 use crate::query_plan::operation::NormalizedDefer;
 use crate::query_plan::operation::RebasedFragments;
-use crate::query_plan::operation::Selection;
-use crate::query_plan::operation::SelectionKey;
 use crate::query_plan::operation::SelectionSet;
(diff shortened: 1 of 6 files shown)


apollo-federation/src/query_plan/query_planner.rs
```diff
@@ -490,11 +500,179 @@ impl QueryPlanner {
...
+            );
+            compute_root_fetch_groups(
+                operation.root_kind,
+                &mut fetch_dependency_graph,
+                &prev_path,
+            )?;
+        } else {
+            // TODO(@TylerBloom): Verify that this is correct? It could be the case that we just
+            // need to grab `.starting_id_generation`.
+            starting_fetch_id = fetch_dependency_graph.next_fetch_id();
```
TylerBloom, line 567: I was unsure if JS code generated the next fetch ID, used the starting ID

apollo-federation/src/query_plan/query_planner.rs
```diff
@@ -490,11 +500,179 @@ impl QueryPlanner {
...
+            prev_subgraph = new_subgraph;
+        }
+    }
+    digest.push(fetch_dependency_graph);
+    Ok(digest)
+}
+
+fn split_top_level_fields(
+    selection_set: SelectionSet,
+) -> Box<dyn 'static + Iterator<Item = SelectionSet>> {
```
TylerBloom, line 582: I would much rather return an `impl Iterator` here, but that would require writing a recursive iterator for this function. It would avoid a few unnecessary allocations and avoid dynamic dispatch, though.Still unsure which would be better.

lrlna: I guess it's a pick your poison kind of scenario - recursive iterator here might (don't quote me) right into lifetime shenanigans. On the other, it would be nice to get a better handle on clones, because this function does it *a lot*.  We can also postpone this decision to later (make a ticket so we don't forget), and just keep what you have here for a few weeks.

An unrelated question: have you thought about having this directly on the SelectionSet struct rather than a standalone function? Are there considerations that prevent us doing something like that?

TylerBloom: > recursive iterator here might (don't quote me) right into lifetime shenanigans

Sorry for *literally* quoting you, but if everything is owned you should be ok (though, this might cause other problems like moving values out of `Arc`s). Otherwise, lifetimes will need to be thought through. Even this solution ran into 

> On the other, it would be nice to get a better handle on clones, because this function does it a lot

Very much agreed. While most of the clones are relatively cheap arc clones, I would love to avoid them and the v-table overhead. I also think a non-adhoc solution would have significantly fewer allocations. All of that does sound like something that can wait.

> An unrelated question: have you thought about having this directly on the SelectionSet struct rather than a standalone function?

I did not. It probably should be.

apollo-federation/src/query_plan/query_planner.rs
dariuszkuc: yeah confusingly JS code does not increment the ID in `next_fetch_id()` (unless I'm missing something it looks like the only place that actually increments the ID is in `extractChildrenAndDeferredDependencies`) but I don't think it matters much? instead of starting with 0 we would just start with 1?

dariuszkuc (approved):

apollo-federation/src/query_plan/query_planner.rs
TylerBloom: There is a chance that `start_fetch_id` gets called multiple times. Also, because we merge dep fetch graph, I'm not sure if their starting number changes (I don't think so, but I'm not 100% sure). In which case, we might want to start at, say, 10 but instead we start at 12 (should this path be taken more than once).

apollo-federation/src/query_plan/query_planner.rs
TylerBloom: @lrlna , I'm going to leave this as is and make a `TODO` comment about what should change. Do you want me to open a ticket as well?

lrlna: yes, please open a ticket! I'll DM you on slack with details

lrlna (approved):

apollo-federation/src/query_plan/query_planner.rs
TylerBloom: I am going to stick with this approach. I'd left a `PORT_NOTE` comment for reference should this cause issues in the future.

PR #5175, opened 2024-05-15
isolate uses of hyper 0.14 types

Geal:
To prepare the migration to hyper 1.0, this PR isolates uses of hyper's types in the Router, to make sure that we do not get stuck updating it all over the code.
One of the issues with the update is that some types exported by hyper, mainly the `Body`, but also the `http` crate's `Request` and `Response` types, along with related builders, are reexported by the Router. This will potentially create issues for users of native plugins.

Goals of this PR:
- [x] remove direct uses of hyper types, replace `hyper::Body` with the wrapper type `RouterBody` when possible, otherwise use the alias `crate::services::router::Body`
- [x] provide helpers to replace direct usage of hyper, in the Router and native plugins
- [ ] mark `crate::services::router::Body` as deprecated, pointing towards better way to create a router service request or response body
- [ ] 

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/services/http.rs
```diff
@@ -21,13 +21,13 @@ pub(crate) type ServiceResult = Result<HttpResponse, BoxError>;
 
 #[non_exhaustive]
 pub(crate) struct HttpRequest {
-    pub(crate) http_request: http::Request<Body>,
+    pub(crate) http_request: http::Request<RouterBody>,
```
garypen, line 24: Breaking change, but ok because we haven't made this publicly useable.

apollo-router/src/services/router.rs
```diff
@@ -30,9 +31,11 @@ use crate::Context;
 pub type BoxService = tower::util::BoxService<Request, Response, BoxError>;
 pub type BoxCloneService = tower::util::BoxCloneService<Request, Response, BoxError>;
 pub type ServiceResult = Result<Response, BoxError>;
+//#[deprecated]
```
garypen, line 34: If we uncomment this, we get ~164 warnings.

I think we need to resolve all of these before this PR is done.

apollo-router/src/services/router.rs
```diff
@@ -196,6 +202,7 @@ impl Response {
         self.response.body_mut().next().await
     }
 
+    #[deprecated]
```
garypen, line 205: I'm happy to deprecate this, but I wonder what the motivation is.

garypen (commented):
This looks like a really good start on the work and it solves the most interesting parts.

Just the tedious parts now...

There are still many mentions of `hyper::Request|Response` in the code base which should all be `http::Request|Response`. Arguably, that's not part of this PR, but I mention it because you had done some replacing in coprocessor/mod.rs.

There are many places inside the router where it looks like we shouldn't be using the (after this PR) deprecated `crate::services::router::Body` and I think this all needs to be fixed as part of this PR.

We'll also need a quality changelog entry explaining why we are doing all of this.

Should the new `body` stuff live under `services`? It isn't a service and so I'd be inclined to just put it in `src/body.rs` or maybe `src/router/body.rs` or ...

apollo-router/src/services/http.rs
Geal: why would it be a breaking change if it is not public?

apollo-router/src/services/router.rs
Geal: I think that was to remove again the `Body` type from the public API. We could instead expose a stream of `Bytes` to map over

apollo-router/src/services/router/body.rs
```diff
@@ -0,0 +1,86 @@
...
+
+    type Error = <hyper::body::Body as HttpBody>::Error;
+
+    fn poll_data(
+        self: std::pin::Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<Option<Result<Self::Data, Self::Error>>> {
+        let mut pinned = unsafe { self.map_unchecked_mut(|s| &mut s.0) };
+        pinned.as_mut().poll_data(cx)
+    }
```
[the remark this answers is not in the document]
Geal: yup, that works, thanks!

apollo-router/src/services/http.rs
garypen: It was a note to myself to make sure I reviewed usage of HttpRequest carefully. I should have removed it before I finished the review. Sorry for the noise.

Geal:
with https://github.com/apollographql/router/pull/5175/commits/733a3ede0caa863a8a72a13ce40acc5cb8f6f47a and https://github.com/apollographql/router/pull/5175/commits/402aa18b492e867995e7233155d0895735b85906 most of the uses of `hyper::Body` have disappeared. If I uncomment the `#[deprecated]` attribute on it, I still see 65 warnings (70 with tests), but they are mainly under `services::router` and router service plugins (unavoidable there).

Geal:
@garypen do you see anything else to change here?

Geal:
I'd like to get this merged before I get to other changes to the HTTP service and its related plugins

bnjjj (approved):

garypen (commented):
regarding your comment:

===
with https://github.com/apollographql/router/commit/733a3ede0caa863a8a72a13ce40acc5cb8f6f47a and https://github.com/apollographql/router/commit/402aa18b492e867995e7233155d0895735b85906 most of the uses of hyper::Body have disappeared. If I uncomment the #[deprecated] attribute on it, I still see 65 warnings (70 with tests), but they are mainly under services::router and router service plugins (unavoidable there).
===

I thought we wanted to deprecate all of the existing `hyper` references within the router 1.x public interfaces to make it clear that anyone using these (including ourselves, I suppose), should expect them to disappear in a future release.

If we can't clean up all usages within the router, I think we should have the warning because it is a constant reminder that we need to address it (even if we can't do it until 2.x).

We need more opinions for a consensus.

garypen (approved):
Following further discussion about deprecation, this is the right PR for the changes we want to make.

PR #5178, opened 2024-05-16
detached coprocessor execution

Geal:
_Reopening https://github.com/apollographql/router/pull/4902 that was reverted in https://github.com/apollographql/router/pull/5068 because we wanted to ship a more coherent list of coprocessor changes_

Fix https://github.com/apollographql/router/issues/3297

This adds a new detached option at all coprocessor stages to allow the router to continue handling the request or response without waiting for the coprocessor to respond. This implies that the coprocessor response will not be used. This is targeted at use cases like logging and auditing, which do not need to block the router

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


abernix:
We can re-open this if/when needed.

PR #5179, opened 2024-05-16
docs: Update RELEASE_CHECKLIST.md

abernix:
Some small tweaks to improve a couple observations I had last time I reviewed this

### DIFF
--- a/RELEASE_CHECKLIST.md
+++ b/RELEASE_CHECKLIST.md
@@ -121,7 +121,7 @@ Start following the steps below to start a release PR.  The process is **not ful
 8. Now, open a draft PR with a small boilerplate header from the branch which was just pushed:
 
    ```
-   cat <<EOM | gh --repo "${APOLLO_ROUTER_RELEASE_GITHUB_REPO}" pr create --draft -B "main" --title "release: v${APOLLO_ROUTER_RELEASE_VERSION}" --body-file -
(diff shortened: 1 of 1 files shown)


Cargo.lock
```diff
@@ -220,7 +220,7 @@ dependencies = [
 
 [[package]]
 name = "apollo-federation"
-version = "0.0.11"
+version = "1.46.0"
```
abernix, line 223: @SimonSapin The fact that this version doesn't (and will likely never) exist on Crates.io shouldn't be a problem, right?

apollo-router/Cargo.toml
```diff
@@ -68,7 +68,7 @@ askama = "0.12.1"
 access-json = "0.1.0"
 anyhow = "1.0.80"
 apollo-compiler.workspace = true
-apollo-federation = { path = "../apollo-federation", version = "=0.0.11" }
+apollo-federation = { path = "../apollo-federation", version = "=1.46.0" }
```
abernix, line 71: Same question as above — it won't break `dev`, right?

Cargo.lock
SimonSapin: No, I don’t expect it to matter.

apollo-router/Cargo.toml
SimonSapin: It won’t because `path` takes priority when `apollo-router` itself is a local path (as opposed to a crates.io dependency) 
https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#multiple-locations

PR #5181, opened 2024-05-16
Support unix_ms_now in rhai engine

### LINKED ISSUE #5182: Support unix time in ms in rhai scripts
We would like to measure precisie timings using the coprocessor, but it's currently impossible since the only available timing function is `unix_now` which returns seconds since epoch.

a new function `unix_ms_now` that returns exact ms of current time.

shaikatzz:
Fixes #5182 

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/plugins/rhai/tests.rs
```diff
@@ -579,6 +579,21 @@ fn it_can_create_unix_now() {
...
+    let engine = new_rhai_test_engine();
+    let st = SystemTime::now()
+        .duration_since(SystemTime::UNIX_EPOCH)
+        .expect("can get system time")
+        .as_millis() as i64;
+    let unix_ms_now: i64 = engine
+        .eval(r#"unix_ms_now()"#)
+        .expect("can get unix_ms_now() timestamp");
+    // Always difficult to do timing tests. unix_ms_now() should execute within a second of st,
+    // so...
```
garypen, line 593: :)

Perhaps reverse the ordering of the timestamp captures and assert that st >= unix_ms_now ?

garypen (commented):
Thanks for this PR and for the excellent presentation. I particularly appreciate the test and the docs additions.

Could you also create a changeset (so you get appropriate credit when this is released) following the procedure in `.changesets/README.md` ? re-request a review from me when you are ready.

apollo-router/src/plugins/rhai/tests.rs
shaikatzz: Actually just followed the same pattern in the test above `it_can_create_unix_now`.
Why does the ordering matter? anyway we will want to check that the second capture ran within a second of the first one, to see that the results makes sense.

shaikatzz:
> Thanks for this PR and for the excellent presentation. I particularly appreciate the test and the docs additions.
> 
> Could you also create a changeset (so you get appropriate credit when this is released) following the procedure in `.changesets/README.md` ? re-request a review from me when you are ready.

Done!

apollo-router/src/plugins/rhai/tests.rs
garypen: It's all good and, as you observe the previous test should be re-ordered.

I was basically thinking that if we re-ordered the samples, then we'd just need to assert that the second timestamp was `>=` the first one. That would eliminate the flakiness of guesstimating that it should run within a second.

Anyway, ignore this for now and we'll change it if the test ever becomes flaky.

garypen (approved):

shaikatzz:
@IvanGoncharov @garypen do you know why the tests are not passing? what else is required in order to merge that PR?

garypen:
> @IvanGoncharov @garypen do you know why the tests are not passing? what else is required in order to merge that PR?

possibly because the CLA wasn't signed. I've just kicked it back into life and we'll see what happens after this run.

shaikatzz:
@garypen I've signed the CLA, let me know if anything else is needed from my side, i'll be happy to have this change merged :-)

PR #5188, opened 2024-05-16
make match arm exhaustive in fetch_dependency display

garypen:
Future clippy (nightly) complained that the match arms weren't exhaustive, so may as well fix this.

I think it's correct to replace 2.. with _.

### DIFF
--- a/apollo-federation/src/query_plan/fetch_dependency_graph.rs
+++ b/apollo-federation/src/query_plan/fetch_dependency_graph.rs
@@ -1684,7 +1684,7 @@ impl std::fmt::Display for FetchInputs {
                 // We can safely unwrap because we know the len >= 1.
                 self.selection_sets_per_parent_type.values().next().unwrap()
             ),
-            2.. => {
+            _ => {
                 write!(f, "[")?;
                 let mut iter = self.selection_sets_per_parent_type.values();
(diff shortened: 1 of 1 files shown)


goto-bus-stop (approved):
I don't mind it, but I don't understand why it complains about 2.. not being exhaustive. These two ways of writing it are exactly equivalent given that `.len()` returns `usize`, no?

garypen:
> I don't mind it, but I don't understand why it complains about 2.. not being exhaustive. These two ways of writing it are exactly equivalent given that `.len()` returns `usize`, no?

Wouldn't it need to be `2..=` to be exhaustive?

goto-bus-stop:
If `2..` was not exhaustive, then it would be a compile error...

goto-bus-stop:
Did you just run `cargo +nightly clippy` to get the warning? I don't see it when I do it today. Maybe this was a clippy bug?

goto-bus-stop:
@garypen is this (still) reproducible or can we close it? 😇

garypen:
@goto-bus-stop I was running some tooling that required "nightly" and that was how I triggered it. If "nightly" isn't complaining, let's just close it. If it ever complains again in the future, it's easy to fix.

PR #5192, opened 2024-05-16
remove iterations from test_reload_config_with_broken_plugin_recovery

Geal:
that test is not flaky, but still takes 15s to do the same operation 3 times in a row

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


garypen (commented):
Might be worth asking @BrynCooke  why he thought it needed to be repeated 3 times?

BrynCooke:
This test was meant to try and catch issues with telemetry reloading, particularly hangs, maybe we're confident enough now that we don't need the multiple iterations.

garypen (approved):

SimonSapin (approved):

PR #5194, opened 2024-05-16
Support renaming of `@source` in validation.

dylan-apollo:
This also adopts the new name in error messages, with snapshot tests to confirm!

Also tweaked the error handling for invalid/no name so that other errors are still collected. This shouldn't matter in practice _yet_ because TypeScript should be validating the presence and type of the `name` argument... but this feels more future-proof.

TS side is at https://github.com/apollographql/federation/pull/3007

### DIFF
--- a/apollo-federation/src/sources/connect/models/validation.rs
+++ b/apollo-federation/src/sources/connect/models/validation.rs
@@ -231,7 +231,7 @@ impl SourceName {
         }
     }
 
-    fn into_value_or_error(self) -> Result<String, ValidationError> {
+    pub fn into_value_or_error(self) -> Result<String, ValidationError> {
         match self {
             Self::Valid { name, .. } => Ok(name),
             Self::Invalid {
--- a/apollo-federation/src/sources/connect/models/validation.rs
+++ b/apollo-federation/src/sources/connect/models/validation.rs
@@ -238,7 +238,7 @@ impl SourceName {
                 name,
                 directive_name,
             } => Err(ValidationError::InvalidSourceName {
-                name,
+                source_name: name,
                 directive_name,
             }),
             Self::Empty { directive_name } => {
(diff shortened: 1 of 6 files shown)


apollo-federation/src/sources/connect/models/validation.rs
```diff
@@ -132,114 +144,166 @@ fn validate_source(directive: &Component<Directive>) -> Result<SourceDirective,
...
     },
-    #[error("name argument for a @source directive must be a string")]
-    SourceNameType,
-    #[error("name argument to @source can't be empty")]
-    EmptySourceName,
+    #[error("name argument to @{directive_name} can't be empty")]
+    EmptySourceName { directive_name: DirectiveName },
     #[error(
-        "invalid characters in source name {0}, only alphanumeric and underscores are allowed"
+        "invalid characters in @{directive_name} name {source_name}, only alphanumeric and underscores are allowed"
```
lennyburdette, line 289: good thinking about renamed directives in error messages! 

what do you think about using the "schema coordinate" in the error message? i looked around the federation code base and saw messages like `Invalid default value (got: "foo") provided for argument Query.f(a:) of type Int.`

So this could be:
```suggestion
        "Invalid characters (got: "j$on") provided for argument @source(name:), only letters, numbers, and underscores are allowed"
```

lennyburdette (approved):

apollo-federation/src/sources/connect/models/validation.rs
dylan-apollo: I do like that format here. As for more general coordinates, I'm going to need a general-purpose coordinate system _anyway_ to communicate the location of `@connect` errors back up to TypeScript. So, I'll definitely add a `Display` impl for that and include it in error messages.

PR #5195, opened 2024-05-17
FED-32: Port `tryOptimizeSubselectionWithFragments`

duckki:
Implemented `try_optimize_with_fragments` function.

- Added a new file `optimize.rs` under `query_plan`.
- Refactored some SelectionSet methods and moved them to SelectionMap for reuse.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

Unfortunately, testing is very cumbersome without the whole `optimize` implementation (FED-191).
I'll implement `optimize` and test this together.

**Notes**


apollo-federation/src/query_plan/optimize.rs
```diff
@@ -0,0 +1,755 @@
...
+    /// - a new selection set partially optimized by re-using given `fragments`, or
+    /// - a single fragment that covers the full selection set.
+    // PORT_NOTE: Moved from `Selection` class in JS code to SelectionSet struct in Rust.
+    // PORT_NOTE: `parent_type` argument seems always to be the same as `self.type_position`.
+    fn try_optimize_with_fragments(
+        &self,
+        parent_type: &CompositeTypeDefinitionPosition,
+        fragments: &NamedFragments,
+        validator: &mut FieldsConflictMultiBranchValidator,
+        can_use_full_matching_fragment: impl Fn(&Fragment) -> bool,
```
lrlna, line 657: let's just have this as a function on `Fragment` directly.

apollo-federation/src/query_plan/optimize.rs
```diff
@@ -0,0 +1,755 @@
...
+        // that one candidate. Otherwise, we cumulate in `applicable_fragments` the list of fragments
+        // that applies to a subset.
+        let mut applicable_fragments = Vec::new();
+        for candidate in candidates {
+            let at_type = candidate.expanded_selection_set_at_type(parent_type)?;
+            if at_type.is_useless() {
+                continue;
+            }
+            if !validator.check_can_reuse_fragment_and_track_it(&at_type)? {
+                continue;
```
lrlna, line 682: for posterity's sake
```suggestion
                // We cannot use it at all, so no point in adding to `applicable_fragments`.
                continue;
```

apollo-federation/src/query_plan/optimize.rs
```diff
@@ -0,0 +1,755 @@
...
+            if matches!(res, Containment::Equal) && can_use_full_matching_fragment(&candidate) {
+                // Special case: Found a fragment that covers the full selection set.
+                return Ok(candidate.into());
+            }
+            // Note that if a fragment applies to only a subset of the sub-selections, then we
+            // really only can use it if that fragment is defined _without_ directives.
+            if !candidate.directives.is_empty() {
+                continue; // Not eligible as a partial selection; Skip it.
+            }
+            applicable_fragments.push((candidate, at_type));
```
lrlna, line 718: i like how is destructured btw, reads a lot better than JS

lrlna (changes requested):
Wow this is a big one!!

First of all, your Rust has gotten **so good**. I am gobsmacked and super impressed at how quickly you got to writing this level of Rust. Incredible!

Second of all, I think we should keep the optimised functionality in operation.rs for a little while longer, and have a conversation about doing a refactor after we've finished everything with in the rewrite. It'll be easier to reason about the changes holistically once everything is in.

I am also really not certain working on the SelectionMap directly is a good idea. I played around a little bit with your changes to see if it's possible to work off a SelectionSet, and I _think_ it should be. We just have to reason through the `minus` and the `intersection` functionality a little bit. Nothing wrong with doing a `SelectionMap::new()`, but I am more concerned about moving functions that are not reliant on being updated upon modifications. 

There are also a few smaller comments and suggestions along the way.

apollo-federation/src/query_plan/optimize.rs
```diff
@@ -0,0 +1,755 @@
...
+        Ok(true)
+    }
+
+    fn do_merge_with_all<'a>(
+        &self,
+        mut iter: impl Iterator<Item = &'a FieldsConflictValidator>,
+    ) -> Result<bool, FederationError> {
+        iter.try_fold(true, |acc, v| Ok(acc && self.do_merge_with(v)?))
+    }
+}
```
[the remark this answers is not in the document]
duckki: But, `is_ok_and(...)` would treat `FederationError` as `false`.
I used `try_fold`, so I can return `FederationError` to caller.

I looked for `try_all` or something like that. But, there were some discussions at Rust community on this. But, it seems not determined, yet. (https://github.com/rust-lang/rfcs/pull/3233)

apollo-federation/src/query_plan/optimize.rs
```diff
@@ -0,0 +1,755 @@
...
+                            }
+                        }
+                    }
+                }
+            }
+        }
+        Ok(true)
+    }
+
+    fn do_merge_with(&self, other: &FieldsConflictValidator) -> Result<bool, FederationError> {
```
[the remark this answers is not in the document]
duckki: We use FederationError. Details in the next comment.

apollo-federation/src/query_plan/optimize.rs
duckki: This has been improved in FED-191. Please bear with me on this for now.

apollo-federation/src/query_plan/optimize.rs
```diff
@@ -0,0 +1,755 @@
...
+    }
+}
+
+//=============================================================================
+// Field validation
+
+// PORT_NOTE: Not having a validator and having a FieldsConflictValidator with empty
+// `by_response_name` map has no difference in behavior. So, we could drop the `Option` from
+// `Option<FieldsConflictValidator>`. However, `None` validator makes it clearer that validation is
+// unnecessary.
```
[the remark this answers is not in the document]
lrlna: Ok, sounds good. Then whatever you think makes this more readable to you!

apollo-federation/src/query_plan/optimize.rs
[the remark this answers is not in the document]
lrlna: It would treat `FederationError` as `false`, yes. 

When I was reading this functions' usage, I thought the error was ignored entirely, which would make this function a bit redundant. I missed the `?` operand. Makes sense then!

apollo-federation/src/query_plan/operation.rs
```diff
@@ -1219,6 +1230,20 @@ mod normalized_field_selection {
...
+            types_can_be_merged(
+                &self_definition.ty,
+                &other_definition.ty,
+                self.schema().schema(),
+            )
+        }
+
+        pub(crate) fn parent_type_position(&self) -> CompositeTypeDefinitionPosition {
+            self.data().field_position.parent()
+        }
```
lrlna, line 1246: these have to be on the outside of the mod implementation of the `Field` - https://github.com/apollographql/router/blob/1e73b8d2819446cf8eb97805a788ded86ab75f2a/apollo-federation/src/query_plan/operation.rs#L3808

lrlna (approved):

PR #5202, opened 2024-05-20
improve experimental broken pipe logging testing

garypen:
The tests are marked as flaky in CI, so let's try to improve them.

 - Reduce the timeout (nano-seconds rather than micro-seconds)
 - Remove the (seemingly) redundant sleep
 - Make the test multi-threaded (which I think it should be)
 - Add comments to explain what the tests are doing


Geal (approved):
LGTM. Have you seen the flaky test pop up again in a lot of CI runs of this PR?

BrynCooke (approved):

BrynCooke:
This is the number one test that seems to flake for me.

PR #5204, opened 2024-05-20
experiment with restoring moka to see how performance varies

garypen:
First replace the storage cache and see how things look. Then think about the other places where we use LruCache.


Geal:
@garypen any insights from running this? Do you see different behaviour when running the same query often, or a lot of different queries?

garypen:
> @garypen any insights from running this? Do you see different behaviour when running the same query often, or a lot of different queries?

Nope. The perf tests aren't sensitive enough to pick up any changes using our standard tests and I haven't had time to do any custom test yet.

Geal:
we would probably see an effect with tests like https://github.com/apollographql/router/pull/5302 where a lot of different queries are sent

PR #5206, opened 2024-05-20
connect: add path to node

nicholascioli:
This commit adds support for adding a path to a connect allowing for JSON selections to be built as the query traverses the connect-specific graph.

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-federation/src/sources/connect/fetch_dependency_graph/mod.rs
+++ b/apollo-federation/src/sources/connect/fetch_dependency_graph/mod.rs
@@ -17,9 +17,12 @@ use crate::source_aware::federated_query_graph::SelfConditionIndex;
 use crate::source_aware::query_plan::FetchDataPathElement;
 use crate::source_aware::query_plan::QueryPlanCost;
 use crate::sources::connect;
+use crate::sources::connect::json_selection::Alias;
 use crate::sources::connect::json_selection::JSONSelection;
 use crate::sources::connect::json_selection::Key;
+use crate::sources::connect::json_selection::NamedSelection;
 use crate::sources::connect::json_selection::PathSelection;
+use crate::sources::connect::json_selection::StarSelection;
 use crate::sources::connect::json_selection::SubSelection;
 use crate::sources::source;
 use crate::sources::source::fetch_dependency_graph::FetchDependencyGraphApi;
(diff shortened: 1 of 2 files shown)


apollo-federation/src/sources/connect/fetch_dependency_graph/mod.rs
```diff
@@ -757,6 +986,451 @@ mod tests {
...
+            let source::fetch_dependency_graph::Node::Connect(result) = node else {
+                unreachable!()
+            };
+            assert_debug_snapshot!(result, @r###"
+            Node {
+                merge_at: [],
+                source_entering_edge: EdgeIndex(3),
+                field_response_name: "_merge_nested_path_test",
+                field_arguments: {},
+                selection: Some(
```
lennyburdette, line 1323: how hard is it to make the snapshot use something like `to_string_pretty()` (which we'd need to write) so that these tests are possible to read?

apollo-federation/src/sources/connect/fetch_dependency_graph/mod.rs
```diff
@@ -757,6 +986,451 @@ mod tests {
...
+                                                vec![
+                                                    Key::Field("baz".to_string()),
+                                                    Key::Field("buzz".to_string()),
+                                                ],
+                                            ),
+                                        ],
+                                        tail_selection: Some((
+                                            name!("biz"),
+                                            connect::fetch_dependency_graph::PathTailSelection::CustomScalarPathSelection {
+                                                path_selection: connect::fetch_dependency_graph::PathSelection::Selection(connect::SubSelection {
```
lennyburdette, line 1290: should we make `SubSelection::parse` and the like crate-public so these tests are more concise?

apollo-federation/src/sources/connect/fetch_dependency_graph/mod.rs
```diff
@@ -93,10 +96,231 @@ impl FetchDependencyGraphApi for FetchDependencyGraph {
...
+            ));
+        }
+        if source_path_field.arguments != source_data.field_arguments {
+            return Err(FederationError::internal(
+                "expected path and source data to have the same field arguments",
+            ));
+        }
+
+        // Ensure that we have a selection, inserting an initial value if not.
+        let selection =
```
lennyburdette, line 144: is there an opportunity to move JSONSelection manipulation code into the json_selection module and have a bunch of small unit tests?

apollo-federation/src/sources/connect/fetch_dependency_graph/mod.rs
```diff
@@ -93,10 +96,231 @@ impl FetchDependencyGraphApi for FetchDependencyGraph {
...
+            source_data
+                .selection
+                .get_or_insert_with(|| match &source_path_field.selections {
+                    // Construct a new selection from the supplied path properties
+                    PathSelections::Selections {
+                        head_property_path,
```
lennyburdette, line 150: i'm still fuzzy on what these properties mean — are you able to add some documentation comments on the `PathSelections` enum at this point?

apollo-federation/src/sources/connect/fetch_dependency_graph/mod.rs
```diff
@@ -93,10 +96,231 @@ impl FetchDependencyGraphApi for FetchDependencyGraph {
...
+            tail_selection,
+            ..
+        } = source_path_field.selections
+        {
+            // We can short out if there's nothing to select
+            let Some((tail_name, tail_subselection)) = tail_selection else {
+                return Ok(());
+            };
+
+            // If we are adding a path and have a selection, then the selection _must_ have a subselection.
```
lennyburdette, line 183: i'd love to have examples in the comments like you have in tests, is this `"foo" + "{ bar }" = "foo { bar }"`?

apollo-federation/src/sources/connect/fetch_dependency_graph/mod.rs
```diff
@@ -93,10 +96,231 @@ impl FetchDependencyGraphApi for FetchDependencyGraph {
...
+            for (name, keys) in named_selections {
+                // If we have a selection already, we'll need to make sure that it includes the new field,
+                // then we process the next subselection in the path chain.
+                // TODO: This is probably not very performant, but we only have a Vec to work with...
+                subselection_ref = if let Some(matching_selection_position) = subselection_ref
+                    .selections
+                    .iter()
+                    .position(|s| name_matches(s, &name))
+                {
+                    let macthing_selection = subselection_ref.selections.get_mut(matching_selection_position).ok_or(FederationError::internal("matched position does not actually exist in selections. This should not happen"))?;
```
lennyburdette, line 214:
```suggestion
                    let matching_selection = subselection_ref.selections.get_mut(matching_selection_position).ok_or(FederationError::internal("matched position does not actually exist in selections. This should not happen"))?;
```

apollo-federation/src/sources/connect/fetch_dependency_graph/mod.rs
nicholascioli: I'm not sure if this would help, since the `Path` input does not match 1:1 to any of the various `JSONSelection` types. Maybe a helper method to automatically construct a `Path` from a subselection would be useful, but let me know.

apollo-federation/src/sources/connect/fetch_dependency_graph/mod.rs
nicholascioli: Done!

apollo-federation/src/sources/connect/fetch_dependency_graph/mod.rs
```diff
@@ -757,6 +986,451 @@ mod tests {
...
+                                response_name,
+                                arguments,
+                                selections:
+                                    connect::fetch_dependency_graph::PathSelections::Selections {
+                                        head_property_path: vec![
+                                            Key::Field("foo".to_string()),
+                                            Key::Field("bar".to_string()),
+                                        ],
+                                        named_selections: vec![
+                                            // - Node's selection: a b { x z { three } } c: last
```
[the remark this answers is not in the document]
nicholascioli: Whoops, this is an outdated comment. Good catch!

apollo-federation/src/sources/connect/fetch_dependency_graph/mod.rs
nicholascioli: Done!

apollo-federation/src/sources/connect/fetch_dependency_graph/mod.rs
nicholascioli: Done!

apollo-federation/src/sources/connect/fetch_dependency_graph/mod.rs
nicholascioli: Somewhat done!

lennyburdette (approved):

apollo-federation/src/sources/connect/fetch_dependency_graph/mod.rs
nicholascioli: This should be future work so that we aren't blocked for now.

sachindshinde (commented):
Reviewed the logic for `add_path()` here, looks fine to me!

(One thing I've mentioned in DMs is that the way we compare `Value`s currently cares about the order of fields in object values when we don't want it to, but this will be fixed across the codebase in a later PR.)

apollo-federation/src/sources/connect/json_selection/parser.rs
```diff
@@ -280,11 +368,43 @@ impl PathSelection {
...
+        }
+    }
+
+    pub fn pretty_print(&self, indentation: Option<usize>) -> Result<String, std::fmt::Error> {
+        let mut result = String::new();
+
+        match self {
+            PathSelection::Var(var, path) => {
+                let rest = path.pretty_print(indentation)?;
+                write!(result, "${var}{rest}")?;
```
benjamn, line 388: The `var` name here already includes the `$`, and may sometimes be simply `$`. I can fix this, but I have to say: if you wrote unit tests for this code, you would have noticed this.

apollo-federation/src/sources/connect/fetch_dependency_graph/mod.rs
benjamn: @nicholascioli I strongly disagree with not writing unit tests to avoid being blocked, because you're adding risk down the line (when deadlines are even nearer) that we won't be able to ship what we think we've built, because we didn't build the necessary confidence as we built it. Please rethink this tradeoff, because it's really not as pragmatic as you may think.

Here's an example of a case where you would have caught a bug if you'd tested the code in question: https://github.com/apollographql/router/pull/5206#discussion_r1613662011

apollo-federation/src/sources/connect/json_selection/parser.rs
nicholascioli: Fixed in https://github.com/apollographql/router/pull/5248

apollo-federation/src/sources/connect/fetch_dependency_graph/mod.rs
nicholascioli: Thanks. Issue is fixed in https://github.com/apollographql/router/pull/5248

apollo-federation/src/sources/connect/fetch_dependency_graph/mod.rs
nicholascioli: Thanks. Fixed in https://github.com/apollographql/router/pull/5248

apollo-federation/src/sources/connect/fetch_dependency_graph/mod.rs
```diff
@@ -147,28 +372,75 @@ pub(crate) struct PathField {
...
+    /// Simple selection using a chain of keys.
     Selection {
+        /// The chain of [Key]s to traverse
         property_path: Vec<Key>,
     },
-    CustomScalarPathSelection {
-        path_selection: PathSelection,
-    },
+
+    /// Custom selection using a [PathSelection]
```
sachindshinde, line 427: I'm not sure if "custom selection" is the right phrase here, it's more of a selection that returns a GraphQL custom scalar (similarly below).

PR #5207, opened 2024-05-21
FED-191: [Operation Processing] port `operation.optimize`

duckki:
implemented the rest of `Operation::optimize` method

https://apollographql.atlassian.net/browse/FED-191

Depends on:
- https://github.com/apollographql/router/pull/5195
- https://github.com/apollographql/router/pull/5200

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

Have a few simple tests. More tests shall be done by poring existing federation tests.

**Notes**


apollo-federation/src/query_plan/optimize.rs
```diff
@@ -765,3 +825,420 @@ impl SelectionSet {
...
+    ) -> Result<(), FederationError> {
+        let min_usages_to_optimize =
+            min_usages_to_optimize.unwrap_or(Self::DEFAULT_MIN_USAGES_TO_OPTIMIZE);
+        let Some(fragments) = fragments else {
+            return Ok(());
+        };
+        if fragments.is_empty() {
+            return Ok(());
+        }
+        assert!(
```
goto-bus-stop, line 1224: We can avoid this assert either by taking a `Option<NonZeroU32>` parameter, or maybe even removing it entirely (AFAIK JS only used it in tests)

duckki: Makes sense. I'll remove the parameter altogether.

apollo-federation/src/query_plan/optimize.rs
```diff
@@ -765,3 +825,410 @@ impl SelectionSet {
...
+        }
+    }
+}
+
+impl FieldSelection {
+    fn optimize(
+        &self,
+        fragments: &NamedFragments,
+        validator: &mut FieldsConflictMultiBranchValidator,
+    ) -> Result<Selection, FederationError> {
```
lrlna, line 1019: Do you think it'll be easier to reason about if this returns a `FieldSelection` instead? Something like this:

```suggestion
    fn optimize(
        mut self,
        fragments: &NamedFragments,
        validator: &mut FieldsConflictMultiBranchValidator,
    ) -> Result<Self, FederationError> {
```

I am thinking we are walking down the tree from the `SelectionSet` root, modifying and optimising it's potential children. It might be easier if the results bubble up as their own respective types, and we handle converting them into a `Selection` at one of the top level nodes?

apollo-federation/src/query_plan/optimize.rs
```diff
@@ -765,3 +825,410 @@ impl SelectionSet {
...
+        Ok(self.with_updated_selection_set(Some(optimized)).into())
+    }
+}
+
+impl InlineFragmentSelection {
+    fn optimize(
+        &self,
+        fragments: &NamedFragments,
+        validator: &mut FieldsConflictMultiBranchValidator,
+    ) -> Result<Selection, FederationError> {
```
lrlna, line 1063: Same comment as above.

apollo-federation/src/query_plan/optimize.rs
lrlna: Potentially might not work due to borrow checker. But I guess what I am trying to say is that I am finding it difficult to follow that `FieldSelection`/`InlineFragmentSelection`/etc all optimize, theoretically their own values, but they don't actually always modify their own selection_set field _and also_ don't return the `selection_set` but a `selection`

lrlna (commented):
Ok, I did a second pass, and nothing really jumps out at me. I think the main thing to do is to port those optimisation tests. It'll be a huge help here!

apollo-federation/src/query_plan/optimize.rs
duckki: Indeed. In the JS version, `optimze()` function was a virtual function with the same return type (`Selection`). But, in Rust, we don't have do that. The wrinkle here is that `InlineFragmentSelection` can be optimized into `FragmentSpreadSelection`. So, it makes sense to add a new enum `InlineOrFragmentSelection`, which helps to eliminate one `unreachable!()` branch case.

duckki:
I ported a few tests. That detected a bug in optimization and in a few other places. I'd like to merge this PR and hunt the rest (if any) together as we port tests.

lrlna (approved):
Glad you already found some bugs! Let's merge and find some more.

apollo-federation/src/query_plan/operation.rs
```diff
@@ -5320,23 +5346,23 @@ type Foo {
...
+            normalized_operation.named_fragments = Default::default();
+            insta::assert_snapshot!(normalized_operation, @r###"
+                query NamedFragmentQuery {
+                  foo {
+                    id
+                    bar
+                    baz
+                  }
+                }
+            "###);
```
lrlna, line 5365: do we want to have snapshots inside mods? have we done this before? i don't want to us to introduce a new pattern for mod tests

lrlna:
There are some snapshot tests changes, and the new changes produced by these snapshots genuinely look incorrect - spacing for fragment definitions looks off 🤔

duckki:
I fixed the spacing issue and a bug in `reuse_query_fragments` option handling.
I reviewed the status of failing tests and annotated appropriately (some improved and some failed showing different symptoms).
I think it's ok to merge into dev.

PR #5210, opened 2024-05-21
Implemented FederatedQueryGraphBuilderApi::process_subgraph_schema for FederatedQueryGraphBuilder

TylerBloom:
This PR addresses FED-200, which outlines the implementation of `FederatedQueryGraphBuilderApi::process_subgraph_schema` for `FederatedQueryGraphBuilder`.

I want to point out that I added a dependency to the `either` crate. This was already a transitive dependency (used by `itertools`). It was helpful for this implementation. I also noticed that it would be useful in other places as well. For example, it would be nice to have a method like `FederationSchema::possible_runtime_types` but which returned a non-consuming iterator rather than a freshly allocated `IndexSet`. Since `either::Either` implements iterator so long as both sides are iterators, we could copy the existing `possible_runtime_types` method and use `Either`s for each branch. I'm sure there are other places where we could make similar low-hanging-fruit optimizations.

### DIFF
--- a/apollo-federation/src/sources/graphql/federated_query_graph/builder.rs
+++ b/apollo-federation/src/sources/graphql/federated_query_graph/builder.rs
@@ -133,7 +133,7 @@ fn process_type(
             .filter(|field| is_field_ignored(field.clone().into(), subgraph_schema))
             .try_for_each(|subgraph_field| {
                 subgraph_schema
-                    .get_type(subgraph_field.field_name.clone())
+                    .get_type(subgraph_field.type_name.clone())
                     .and_then(OutputTypeDefinitionPosition::try_from)
                     .and_then(|ty| process_type(ty, subgraph_schema, builder, existing_types))
                     .and_then(|next_index| {
--- a/apollo-federation/src/sources/graphql/federated_query_graph/builder.rs
+++ b/apollo-federation/src/sources/graphql/federated_query_graph/builder.rs
@@ -15,13 +15,12 @@ use crate::schema::ValidFederationSchema;
 use crate::source_aware::federated_query_graph::builder::{
     IntraSourceQueryGraphBuilderApi, IntraSourceQueryGraphSubBuilderApi,
 };
+use crate::sources::graphql::federated_query_graph::{ConcreteFieldEdge, TypeConditionEdge};
 use crate::sources::graphql::GraphqlId;
 use crate::sources::source;
 use crate::sources::source::federated_query_graph::builder::FederatedQueryGraphBuilderApi;
 use crate::sources::source::federated_query_graph::{EnumNode, ScalarNode};
 use crate::ValidFederationSubgraph;
-
-use super::{ConcreteFieldEdge, TypeConditionEdge};
 
 #[derive(Default)]
 pub(crate) struct FederatedQueryGraphBuilder;
--- a/apollo-federation/src/sources/graphql/federated_query_graph/builder.rs
+++ b/apollo-federation/src/sources/graphql/federated_query_graph/builder.rs
@@ -1,7 +1,27 @@
+use std::collections::HashMap;
+
+use apollo_compiler::{name, NodeStr};
+use indexmap::IndexSet;
+use itertools::Either;
+use petgraph::prelude::NodeIndex;
+use strum::IntoEnumIterator;
+
 use crate::error::FederationError;
-use crate::source_aware::federated_query_graph::builder::IntraSourceQueryGraphBuilderApi;
+use crate::schema::position::{
+    AbstractTypeDefinitionPosition, FieldDefinitionPosition, ObjectTypeDefinitionPosition,
+    OutputTypeDefinitionPosition, SchemaRootDefinitionKind, SchemaRootDefinitionPosition,
+};
+use crate::schema::ValidFederationSchema;
+use crate::source_aware::federated_query_graph::builder::{
+    IntraSourceQueryGraphBuilderApi, IntraSourceQueryGraphSubBuilderApi,
+};
+use crate::sources::graphql::GraphqlId;
+use crate::sources::source;
 use crate::sources::source::federated_query_graph::builder::FederatedQueryGraphBuilderApi;
+use crate::sources::source::federated_query_graph::{EnumNode, ScalarNode};
 use crate::ValidFederationSubgraph;
+
+use super::{ConcreteFieldEdge, TypeConditionEdge};
 
 #[derive(Default)]
 pub(crate) struct FederatedQueryGraphBuilder;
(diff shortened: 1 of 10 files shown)


apollo-federation/src/sources/graphql/federated_query_graph/builder.rs
```diff
@@ -9,9 +32,183 @@ pub(crate) struct FederatedQueryGraphBuilder;
...
+        }
+    };
+    existing_types.insert(output_type_definition_position, index);
+    match ty {
+        Some(Either::Left(concrete_ty)) => concrete_ty
+            .field_positions(subgraph_schema.schema())?
+            .filter(|field| is_field_ignored(field.clone().into(), subgraph_schema))
+            .try_for_each(|subgraph_field| {
+                subgraph_schema
+                    .get_type(subgraph_field.field_name.clone())
```
lennyburdette, line 136: i think this is a bug — you're calling `get_type` with a field name, not a type name.

TylerBloom: Ah, good catch! I think I've made this mistake before. Maybe we should consider nominal typing for things like this as field names and type names are the same type but are semantically very different?

apollo-federation/src/sources/graphql/federated_query_graph/builder.rs
```diff
@@ -133,7 +133,7 @@ fn process_type(
             .filter(|field| is_field_ignored(field.clone().into(), subgraph_schema))
             .try_for_each(|subgraph_field| {
                 subgraph_schema
-                    .get_type(subgraph_field.field_name.clone())
+                    .get_type(subgraph_field.type_name.clone())
```
lennyburdette, line 136: is the type name the parent type (`Type` in `Type.field(): Output`) or the output type (`Output` in `Type.field(): Output`)?

TylerBloom: It should be the return type of the field.

lennyburdette: maybe i'm missing something, but i don't think that's true: https://github.com/apollographql/router/blob/dev/apollo-federation/src/schema/position.rs#L1776-L1782 — the type_name comes from the parent type.

lennyburdette: ohhhh sorry, thanks for the clarification!

apollo-federation/src/sources/graphql/federated_query_graph/builder.rs
```diff
@@ -9,9 +32,183 @@ pub(crate) struct FederatedQueryGraphBuilder;
...
+        let entities = query_def.field(name!("_entities"));
+        let service = query_def.field(name!("_service"));
+        if [entities, service].contains(position) {
+            return true;
+        }
     }
+
+    subgraph_schema
+        .subgraph_metadata()
+        .unwrap()
```
sachindshinde, line 197: Router folks have told me we want to try to avoid calling code that can panic. I think you can avoid the `.unwrap()` by:
1. Changing `is_field_ignored()` to return a `Result<bool, FederationError>`.
2. Use e.g. `let ... else { return Err(FederationError::internal()) }` here to get metadata.
3. Changing the callsites to use `.process_results()` from `itertools`. E.g. the `filter()` call above becomes
```rust
.map(|field| is_field_ignored(field.clone().into(), subgraph_schema).map(|is_ignored| (!is_ignored).then_some(field)))
.process_results(|iter| {
    iter.flatten().try_for_each(|subgraph_field| {
        // ...
    })
})??,
```

apollo-federation/src/sources/graphql/federated_query_graph/builder.rs
```diff
@@ -9,9 +32,183 @@ pub(crate) struct FederatedQueryGraphBuilder;
...
+        .external_metadata()
+        .is_external(&field_definition_position)
+        .unwrap_or_default()
```
sachindshinde, line 200: With the change to returning `Result<bool, FederationError>` above, you should be able to omit this `.unwrap_or_default()` here.

apollo-federation/src/sources/graphql/federated_query_graph/builder.rs
```diff
@@ -9,9 +32,183 @@ pub(crate) struct FederatedQueryGraphBuilder;
...
+    output_type_definition_position: &OutputTypeDefinitionPosition,
+    subgraph_schema: &ValidFederationSchema,
+) -> bool {
+    let OutputTypeDefinitionPosition::Object(ty) = output_type_definition_position else {
+        return false;
+    };
+    let Ok(mut fields) = ty.field_positions(subgraph_schema.schema()) else {
+        return false;
+    };
+    fields.all(|field| is_field_ignored(field.into(), subgraph_schema))
```
sachindshinde, line 213: Similarly to the above, you'll want to change this function to return `Result<bool, FederationError>`, and if you're using `process_results()`, this line changes to e.g.:
```rust
fields.map(|field| is_field_ignored(field.into(), subgraph_schema))
    .process_results(|mut iter| iter.all(identity))
```

sachindshinde (changes requested):

apollo-federation/src/sources/graphql/federated_query_graph/builder.rs
TylerBloom: To make sure that I have the right idea, I need to replace `subgraph_field.type_name.clone()` with `subgraph_field.get(subgraph_schema.schema())?.ty.inner_named_type()` in order to get the correct type name. Put together, this would be get the inner type name then call `subgraph_schema.get_type(inner_type_name)`, i.e.
```rust
// Old
subgraph_schema
    .get_type(subgraph_field.type_name.clone())
// New
subgraph_field
    .get(subgraph_schema.schema())
    .and_then(|field| subgraph_schema.get_type(field.ty.inner_named_type().clone()))
```

sachindshinde: @TylerBloom Yep, that looks right to me

apollo-federation/src/sources/graphql/federated_query_graph/builder.rs
```diff
@@ -9,9 +33,189 @@ pub(crate) struct FederatedQueryGraphBuilder;
...
+            };
+            (ty, builder.add_abstract_node(type_name, node.into())?)
+        }
+    };
+    existing_types.insert(output_type_definition_position, index);
+    match ty {
+        Some(Either::Left(concrete_ty)) => concrete_ty
+            .field_positions(subgraph_schema.schema())?
+            .try_for_each(|subgraph_field| {
+            if !is_field_ignored(subgraph_field.clone().into(), subgraph_schema)? {
```
sachindshinde, line 137: Need to remove the `!` here (should early return when the field is ignored).
```suggestion
            if is_field_ignored(subgraph_field.clone().into(), subgraph_schema)? {
```

sachindshinde (approved):
Just one fix below, but other than that LGTM

lrlna (commented):
can we make sure this has adds some tests?

PR #5211, opened 2024-05-21
connect: impl `can_reuse_node` for FetchDependencyGraph

nicholascioli:
This commit implements the method `can_reuse_node` of the `FetchDependencyGraphApi` trait for the connect-specific `FetchDependencyGraph`.

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-federation/src/sources/connect/fetch_dependency_graph/mod.rs
+++ b/apollo-federation/src/sources/connect/fetch_dependency_graph/mod.rs
@@ -34,7 +34,7 @@ use crate::sources::source::SourceId;
 pub(crate) struct FetchDependencyGraph;
 
 impl FetchDependencyGraphApi for FetchDependencyGraph {
-    fn can_reuse_node<'path_tree>(
+    fn edges_that_can_reuse_node<'path_tree>(
         &self,
         _query_graph: Arc<FederatedQueryGraph>,
         merge_at: &[FetchDataPathElement],
--- a/apollo-federation/src/sources/connect/fetch_dependency_graph/mod.rs
+++ b/apollo-federation/src/sources/connect/fetch_dependency_graph/mod.rs
@@ -50,10 +50,7 @@ impl FetchDependencyGraphApi for FetchDependencyGraph {
         // If we have distinct merge positions, or if the entering edge is different from the supplied node,
         // then there is nothing in common and thus nothing reusable.
         if source_entering_edge != source_data.source_entering_edge
-            || merge_at
-                .iter()
-                .zip(source_data.merge_at.iter())
-                .any(|(lhs, rhs)| lhs != rhs)
+            || *merge_at != *source_data.merge_at
         {
             return Ok(Vec::new());
         }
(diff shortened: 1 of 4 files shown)


apollo-federation/src/sources/connect/fetch_dependency_graph/mod.rs
```diff
@@ -34,12 +34,67 @@ impl FetchDependencyGraphApi for FetchDependencyGraph {
     fn can_reuse_node<'path_tree>(
```
lennyburdette, line 34: I don't think i like this method name — "can" implies a boolean return value. 

should it be "reusable_edges_for_node"?

what happens to the edges that can't be reused? what happens if we return an empty vec? 

cc @sachindshinde

sachindshinde: It originally returned a bool, though for connectors we don't batch across different aliases/arguments at the moment, so it ends up filtering edges instead.

I wouldn't necessarily call it `reusable_edge_for_node()`, as that would imply the edge is the thing being reused when it's really the node. It's more like those particular edges can reuse the node, so maybe `edges_that_can_reuse_node()`?

lennyburdette: yeah that's helpful!

lennyburdette (approved):

apollo-federation/src/sources/connect/fetch_dependency_graph/mod.rs
nicholascioli: Done!

apollo-federation/src/sources/connect/fetch_dependency_graph/mod.rs
```diff
@@ -31,15 +31,71 @@ use crate::sources::source::SourceId;
...
+            if op_data.response_name() != source_data.field_response_name {
+                continue;
+            }
+
+            // If the arguments differ, then we can't reuse the edge
+            if op_data.arguments.len() != source_data.field_arguments.len()
+                || op_data
+                    .arguments
+                    .iter()
+                    .zip(source_data.field_arguments.iter())
```
sachindshinde, line 86: I don't think there's a guarantee that the argument order is the same between the `Vec` of arguments in `op_data.arguments` and the `IndexMap` in `op_data.field_arguments`, so I'm not sure the `.zip()` will work here. After the length check, you could instead iterate through `op_data.arguments`, lookup the argument by name in the `IndexMap`, and check that the value exists and is the same.

sachindshinde (changes requested):

apollo-federation/src/sources/connect/fetch_dependency_graph/mod.rs
```diff
@@ -31,15 +31,71 @@ use crate::sources::source::SourceId;
...
+        let source::fetch_dependency_graph::Node::Connect(source_data) = source_data else {
+            return Err(FederationError::internal("expected connect node"));
+        };
+
+        // If we have distinct merge positions, or if the entering edge is different from the supplied node,
+        // then there is nothing in common and thus nothing reusable.
+        if source_entering_edge != source_data.source_entering_edge
+            || merge_at
+                .iter()
+                .zip(source_data.merge_at.iter())
```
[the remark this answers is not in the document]
nicholascioli: Good catch! I'll add a check to make sure that the lengths are the same!

apollo-federation/src/sources/connect/fetch_dependency_graph/mod.rs
nicholascioli: Will do!

sachindshinde (approved):

PR #5219, opened 2024-05-22
Implemented FederatedQueryGraphBuilders::process_subgraph_schemas

TylerBloom:
This PR addresses FED-199. Specifically, it provides an implementation of `FederatedQueryGraphBuilders::process_subgraph_schemas`.

### DIFF
--- a/apollo-federation/src/sources/source/federated_query_graph/builder.rs
+++ b/apollo-federation/src/sources/source/federated_query_graph/builder.rs
@@ -1,5 +1,3 @@
-#![allow(unused, unconditional_panic, clippy::diverging_sub_expression)]
-use apollo_compiler::name;
 use enum_dispatch::enum_dispatch;
 use indexmap::IndexMap;
 
--- a/apollo-federation/src/sources/source/federated_query_graph/builder.rs
+++ b/apollo-federation/src/sources/source/federated_query_graph/builder.rs
@@ -7,6 +7,7 @@ use crate::error::FederationError;
 use crate::source_aware::federated_query_graph::builder::IntraSourceQueryGraphBuilder;
 use crate::source_aware::federated_query_graph::builder::IntraSourceQueryGraphBuilderApi;
 use crate::sources::connect;
+use crate::sources::connect::ConnectSpecDefinition;
 use crate::sources::graphql;
 use crate::sources::source::SourceKind;
 use crate::ValidFederationSubgraph;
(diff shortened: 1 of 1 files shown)


sachindshinde (approved):
This looks fine to me. One caveat is that I think folks generally don't like methods that can panic (outside of testing), so it may be preferable to avoid the `unwrap()` and return `Result<&FederatedQueryGraphBuilder, FederationError>` from `get_builder()` instead.

TylerBloom:
> avoid the unwrap() and return Result<&FederatedQueryGraphBuilder, FederationError> from get_builder() instead.

I agree with you, I would also like to remove that `unwrap`. However, I believe this builder should always have a query graph builder of both kinds. IMO, this should be made explicit. If we change the definition of this builder to what I have below, this unwrap would be completely unnecessary (there are also other benefits, but I think those are pretty minimal).
```rust
// Old
pub(crate) struct FederatedQueryGraphBuilders {
    builders: IndexMap<SourceKind, FederatedQueryGraphBuilder>,
}

// New
pub(crate) struct FederatedQueryGraphBuilders {
    graphql: FederatedQueryGraphBuilder, // Always the GraphQL variant
    connect: FederatedQueryGraphBuilder, // Always the Connect variant
}

impl FederatedQueryGraphBuilders {
  fn get_builder(&self, kind: SourceKind) -> &FederatedQueryGraphBuilder {
    match kind { /* Return reference to the corresponding builder */ }
  }
}
```

sachindshinde (approved):
Switching the `IndexMap` to struct fields makes sense to ensure there's always a builder for each kind, changes LGTM 🚀

apollo-federation/src/sources/source/federated_query_graph/builder.rs
```diff
@@ -26,32 +26,70 @@ pub(crate) trait FederatedQueryGraphBuilderApi {
...
+        builder: &mut IntraSourceQueryGraphBuilder,
     ) -> Result<(), FederationError> {
-        todo!()
+        subgraphs.into_iter().try_for_each(|(_name, graph)| {
+            let src_kind = extract_source_kind(&graph);
+            builder.set_source_kind(src_kind);
+            self.get_builder(src_kind)
+                .process_subgraph_schema(graph, builder)
+        })
+    }
```
lrlna, line 45: can this be tested in any way?

TylerBloom: The main invariant that I think this needs to hold is that the `GraphQL` and `Connect` `process_subgraph_schema` methods expect to be passed only their respective versions of the `FederatedQueryGraphBuilder`. This is transitively tested when we test both of those implementations. Should they receive an incorrect builder. Is there something else that needs to be tested for?

TylerBloom: I could add `debug_assert` to make sure the variants line up.

PR #5230, opened 2024-05-23
connect: add static cost to connect fetch

nicholascioli:
This commit implements the `to_cost` method of the `FetchDependencyGraphApi` for the connector-specific fetch graph. For now, that cost is set to a static value mirroring the cost of a fetch in GraphQL so that the query planner can still consistently choose shorter paths when comparing between the two.

Later work might entail looking at the `JSONSelection` itself and calculating a scaling cost based on the selection.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


lennyburdette (approved):

sachindshinde (approved):
This should be fine for the first milestone

PR #5245, opened 2024-05-24
Add `FallibleIterator` extension trait

TylerBloom:
This PR adds a `utils` module at the root of apollo federation. In it is an `Iterator` extension trait to help work with iterators over `Result`s, operations that yield `Result`s, and both at the same time.

The extension trait is by no means complete or designed to cover every usecase. The intention is to cover common usecases now and expand things as they are needed. Of particular interest for me were variant of `any`, `all`, and `filter` (plus `and_then`).

### DIFF
--- a/apollo-federation/src/supergraph/mod.rs
+++ b/apollo-federation/src/supergraph/mod.rs
@@ -101,12 +101,14 @@ pub(crate) fn extract_subgraphs_from_supergraph(
     let filtered_types: Vec<_> = supergraph_schema
         .get_types()
         .fallible_filter(|type_definition_position| {
-            Ok(!join_spec_definition
-                .is_spec_type_name(supergraph_schema, type_definition_position.type_name())?)
+            join_spec_definition
+                .is_spec_type_name(supergraph_schema, type_definition_position.type_name())
+                .map(Not::not)
         })
         .and_then_filter(|type_definition_position| {
-            Ok(!link_spec_definition
-                .is_spec_type_name(supergraph_schema, type_definition_position.type_name())?)
+            link_spec_definition
+                .is_spec_type_name(supergraph_schema, type_definition_position.type_name())
+                .map(Not::not)
         })
         .try_collect()?;
     if is_fed_1 {


TylerBloom:
I am still working on docs and tests.

TylerBloom:
While working on this, it occurred to me that there is nothing specific to Apollo about this. Perhaps this could/should be mirrored and published as a separate crate? I don't know what Apollo's policy is on that.

apollo-federation/src/utils/fallible_iterator.rs
```diff
@@ -0,0 +1,344 @@
+#![deny(rustdoc::broken_intra_doc_links)]
+
+use itertools::Itertools;
+
+#[allow(dead_code, unused)]
```
SimonSapin, line 5: Why not `#[test]` instead?

SimonSapin (commented):
Is there nothing at all in the rest of the code base that can use these utilities? Showing actual use in this PR would help motivate their existence. I’m a bit reluctant to add them "just in case"

goto-bus-stop:
I agree it would be good to update some usage sites in a separate commit.

Is there a reason you picked `fallible_*` over `try_*`? I think a `.try_filter()` that short-circuits when the predicate returns Err is perfectly in line with `.try_collect()` in itertools

lrlna:
@goto-bus-stop @SimonSapin the two of you had comments previously on this PR, do you have time in the next week to review again, or discuss with @TylerBloom if this is something we'd like in the repo? I don't want this branch to get too stale.

apollo-federation/src/operation/mod.rs
```diff
@@ -2325,12 +2327,14 @@ impl SelectionSet {
...
-            if selection.conditions()? != conditions {
-                return Ok(Conditions::Boolean(true));
-            }
+        if selections
+            .map(|selection| selection.conditions())
+            .ok_and_any(|cond| cond != conditions)?
+        {
+            Ok(Conditions::Boolean(true))
+        } else {
+            Ok(conditions)
```
SimonSapin, line 2336: I’m not super convinced this is an improvement over the `for` loop but 🤷

apollo-federation/src/query_graph/mod.rs
```diff
@@ -901,17 +899,15 @@ impl QueryGraph {
...
-
-        Ok(false)
+        schema
+            .possible_runtime_types(interface_field_definition_position.parent().into())?
+            .into_iter()
+            .map(|object_type_definition_position| {
+                object_type_definition_position
+                    .field(interface_field_definition_position.field_name.clone())
+            })
+            .map(|field_pos| field_pos.get(schema.schema()))
```
SimonSapin, line 909:
```suggestion
            .map(|object_type_definition_position| {
                let field_pos = object_type_definition_position
                    .field(interface_field_definition_position.field_name.clone());
                field_pos.get(schema.schema())
            })
```

apollo-federation/src/query_graph/mod.rs
```diff
@@ -901,17 +899,15 @@ impl QueryGraph {
...
+            .ok_and_any(|field| field.directives.has(&provides_directive_definition.name))
+            .map_err(Into::into)
```
SimonSapin, line 911: Is there a nice way to replace this map_err with a `?` operator?

apollo-federation/src/supergraph/mod.rs
```diff
@@ -96,16 +98,19 @@ pub(crate) fn extract_subgraphs_from_supergraph(
...
+        .fallible_filter(|type_definition_position| {
+            join_spec_definition
+                .is_spec_type_name(supergraph_schema, type_definition_position.type_name())
+                .map(Not::not)
+        })
+        .and_then_filter(|type_definition_position| {
+            link_spec_definition
+                .is_spec_type_name(supergraph_schema, type_definition_position.type_name())
+                .map(Not::not)
+        })
```
SimonSapin, line 112: `.map(not)` feels very awkward to me, I think I prefer `?` +  `Ok`:
```suggestion
        .fallible_filter(|type_definition_position| {
            Ok(!join_spec_definition
                .is_spec_type_name(supergraph_schema, type_definition_position.type_name())?)
        })
        .and_then_filter(|type_definition_position| {
            Ok(!link_spec_definition
                .is_spec_type_name(supergraph_schema, type_definition_position.type_name())?)
        })
```

SimonSapin (approved):
Approved with `pub(crate)`, the rest as you see fit

apollo-federation/src/query_graph/mod.rs
TylerBloom: `map_err(Into::into)` can be replaced with `?` and wrapping the whole express in an `Ok`. I generally dislike `Ok(fallible()?)`, but that's kind of bikeshedding. It doesn't really matter.

apollo-federation/src/operation/mod.rs
TylerBloom: Ya, I was probably a little overly zealous here.

apollo-federation/src/supergraph/mod.rs
TylerBloom: I agree with you, but there is a reason for this. The compiler can not infer what the error type is for the `Ok`s. This is true with the `fallible_filter` + `and_then_filter` and if the two filters combined into something like:
```rs
        .fallible_filter(|type_definition_position| {
            Ok(!join_spec_definition
                     .is_spec_type_name(supergraph_schema, type_definition_position.type_name())?
                  ||  !link_spec_definition
                        .is_spec_type_name(supergraph_schema, type_definition_position.type_name())?
        )})
```

PR #5246, opened 2024-05-24
Reduce allocations in demand control plugin

tninesling:
# Overview

We noticed a fairly significant drop in performance with demand control enabled during our load tests. After taking a look at some flamegraphs of the execution, the main two culprits seem to be
1. Parsing @requires directives during static estimation
2. Allocations/inserts with HashMaps when we zip responses with the schema

## Handling requires

As currently implemented, our handling of @requires is not necessary because we're only running the estimation on the query plan. At that point, the query planner has already converted any requirements to fetch nodes. This did bring to my attention that we should be hooking into the request pipeline earlier with the static estimation, which will need the requires implementation. The goal of demand control is to have that estimate block expensive requests _prior_ to query planning, otherwise, we're not really saving that much work. This requires further discussion.

This also means that we probably need two versions of our scoring, or at least to share a little less code so the query plan scoring doesn't do the expensive @requires parsing. In the first pass, I've removed it entirely to test how it affects throughput.

## Reducing allocations

There were a few low-hanging fruits here, mainly, initializing HashMaps and Vecs with capacity and using &str keys instead of String. This seems to have a pretty sizeable impact on the flamegraphs. I will confirm with the load tests.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6037,9 +6037,9 @@ dependencies = [
 
 [[package]]
 name = "schemars"
-version = "0.8.18"
+version = "0.8.21"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0afe01b987fac84253ce8acd5c05af9941975e4dee5b4f2d826b6947be8ec2c7"
+checksum = "09c024468a378b7e36765cd36702b7a90cc3cba11654f6685c8f233408e89e92"
 dependencies = [
  "dyn-clone",
  "schemars_derive",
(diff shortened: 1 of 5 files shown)


tninesling:
Perf tests are failing to run, need to see if these changes are compatible with the original load testing branch

PR #5247, opened 2024-05-24
Switch connectors validation locations to line/column

dylan-apollo:
These changes are to power https://github.com/apollographql/federation-rs/pull/499

Previously, errors attempted to refer to a part of the AST without passing nodes themselves (e.g., the `name` of a source directive), since they had to be translated into the JavaScript version of the AST. We couldn't use positions (offsets/line/column) because we did not have the _original_ source, only one that had been modified by the JavaScript composition code.

Now, we return line/column information to be directly consumed by `supergraph` binaries. This version of the connectors validation code is what's powering `supergraph@v2.8.0-connectors.5`

### DIFF
--- a/apollo-federation/src/sources/connect/models/validation.rs
+++ b/apollo-federation/src/sources/connect/models/validation.rs
@@ -60,12 +60,16 @@
 
 */
 
+use std::collections::HashMap;
+use std::error::Error;
 use std::fmt::Display;
 
+use apollo_compiler::ast::Value;
 use apollo_compiler::schema::Component;
 use apollo_compiler::schema::Directive;
+use apollo_compiler::Node;
+use apollo_compiler::NodeLocation;
 use apollo_compiler::Schema;
-use itertools::Itertools;
 use url::Url;
 
 use crate::link::Link;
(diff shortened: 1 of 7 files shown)


apollo-federation/src/sources/connect/models/validation.rs
```diff
@@ -321,7 +384,7 @@ mod test_validate_source {
         let schema = Schema::parse(schema, "test.graphql").unwrap();
         let errors = validate(schema);
         assert_eq!(errors.len(), 1);
-        assert!(matches!(&errors[0], ValidationError::SourceScheme { .. }));
+        assert_snapshot!(errors[0].to_string());
```
lennyburdette, line 387: for short-ish string assertions, can you use [inline snapshots](https://docs.rs/insta/latest/insta/#inline-snapshots) so it's easier to see the expected result in the test? (or just use `assert_eq`)

dylan-apollo: Seems like the sort of thing we should maybe have an _overall_ style guide for—so the errors are consistent across products. Is there already such a thing somewhere or should we kick that off here and with the consuming work.

PR #5249, opened 2024-05-24
connect: use selection parser for tests

nicholascioli:
This commit uses the `JSONSelection` parsers in tests to make them more concise.

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-federation/src/sources/connect/fetch_dependency_graph/mod.rs
+++ b/apollo-federation/src/sources/connect/fetch_dependency_graph/mod.rs
@@ -765,9 +765,7 @@ mod tests {
         use crate::sources::connect::federated_query_graph::ConcreteNode;
         use crate::sources::connect::federated_query_graph::SourceEnteringEdge;
         use crate::sources::connect::fetch_dependency_graph::FetchDependencyGraph;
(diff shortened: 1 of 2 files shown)


benjamn (changes requested):
This is still manually creating an invalid `PathSelection` that consists of only a single `PathSelection::Selection`, similar to the problem I described in my previous comment https://github.com/apollographql/router/pull/5206#discussion_r1613817398.

You need to use the `PathSelection::parse` function, and you should be parsing the string `.blah { x y }` without surrounding braces (that is, not a `SubSelection`).

Also, for your awareness, the inner syntax you're parsing here (`blah: { x y }`) is not a `PathSelection`, but a `NamedGroupSelection` (because of the `:`), which introduces a new `blah` key for nesting/isolating the `{ x y }` object. In other words, with this grouping syntax, the `blah` key isn't something you're extracting from the data, but something you're adding to the result.

nicholascioli:
This will be closed for now, but should be revisited once we have a source-aware query plan.

Issue #5160, opened 2024-05-14
Query plan cache key follow up


Geal:
as a follow up to #5100, there are more fixes to apply to query plan cache keys:
* [use prefixes for each part of the redis cache key so they become self describing](https://github.com/apollographql/router/pull/5100#discussion_r1596585501)
* [remove the custom `Hash` implementation](https://github.com/apollographql/router/pull/5100#discussion_r1598456345)
* [remove JSON serialization](https://github.com/apollographql/router/pull/5100#issuecomment-2106893375)
* hash the Rust planner's config once a new version has been released ([it implements `Hash` now](https://github.com/apollographql/router/commit/d7aa8ad09601d06012d331702d4d185af6114b63))

PR #5255, opened 2024-05-27
Reintroduce query plan reuse during warmup

Geal:
Fix https://github.com/apollographql/router/issues/5160

TL;DR: This reintroduces the schema aware query hashing algorithm with a configuration option to enable it for query plan reuse in warmup, and some performance and reliability fixes around the cache key generation.

## Problem
The router implements a query hashing algorithm that takes the schema into account, so that if the parts of the schema relevant to the query do not change across a schema update (example: a field was added to an unrelated type), then the hash will not change.
This is useful in 2 ways:
- when updating the schema, query plans must be regenerated for existing queries. We can use the `supergraph.query_planning.warmed_up_queries` option to pregenerate query plans before switching traffic to the new schema. But that can result in a lot of queries being planned, a lot of CPU usage, and schema updates taking a long time. This query hashing algorithm can be used to detect that a schema update would not affect a query, so we can reuse its query plan, and avoid regenrating most of the query plans. This is the goal of the [`experimental_reuse_query_plans` option](https://github.com/apollographql/router/commit/3b08805feaf03873a6f546d3692167c32825e232).
- in entity caching, we hash the query as part of the cache key, but that is not enough, we need to check that relevant parts of the schema do not change. As an example: if the type of a field changes, we don't want to keep serving data with the wrong type. But we also do not want to rebuild the entire cache from scratch every time the schema updates. By using this query hashing algorithm, we can reuse most of the entity cache across schema updates

## Proposed solution
An earlier version of this algorithm shipped in a router release with some significant issues that broke query plan caching, so it was promptly deactivated. This PR reintroduces the algorithm, with the `experimental_reuse_query_plans` set by default to `do_not_reuse`, making its use in the query planner cache optional. It will generate the query hash in all cases, and use it with the schema hash (a hash of the schema SDL as bytes), but when `experimental_reuse_query_plans` is set to "reuse", only the query hash will be employed. When the option is set to `measure`, it measures the number of query plans that could have been reused by comparing old and new query plans, in the same way we do with JS and Rust planners.
It also brings significant improvements to the query planner cache handling, using prefixes to make the Redis key self describing, reducing the amount of data serialization when generating the cache key, and making sure the rust planner's configuration is used in the cache key.

TODO:
- [x] ~configuration option to reactivate the schema hash~ the hash is now always generated, but the cache key will by default use that query hash and the schema hash. If the [`experimental_reuse_query_plans` option](https://github.com/apollographql/router/commit/3b08805feaf03873a6f546d3692167c32825e232) is enabled, then it will not use the schema hash
- [x] metric to measure the number of query plans that could have been reused during warmup
- [x] measure how many of the query plans that could have been reused are actually the same
- [x] ~check with the federation version for the in memory cache~ we don't need to check it because a running router instance will only have one federation version
- [x] [remove the custom](https://github.com/apollographql/router/pull/5100#discussion_r1598456345) Hash implementation for the caching key
- [x] [use prefixes for each part of the redis cache key so they become self describing](https://github.com/apollographql/router/pull/5100#discussion_r1596585501)
- [x]  [remove JSON serialization](https://github.com/apollographql/router/pull/5100#issuecomment-2106893375)
- [x] hash the Rust planner's config once a new version has been released ([it implements ](https://github.com/apollographql/router/commit/d7aa8ad09601d06012d331702d4d185af6114b63)Hash now)
- [ ] configuration migration for the `experimental_reuse_query_plans` option from a boolean to an enum

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-269]: https://apollographql.atlassian.net/browse/ROUTER-269?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apollo-router/src/spec/query/change.rs
```diff
@@ -194,97 +248,127 @@ impl<'a> QueryHashVisitor<'a> {
...
+        "^INPUT_VALUE".hash(self);
+
         self.hash_type(&t.ty)?;
         for directive in &t.directives {
             self.hash_directive(directive);
         }
         if let Some(value) = t.default_value.as_ref() {
             self.hash_value(value);
+        } else {
+            "no default value".hash(self);
```
IvanGoncharov, line 363: Since this edge case is handled in two places maybe it better to change `hash_value` to accept `Option`?

IvanGoncharov: Also `no default value` contrasts with other literals used by hashing code since it doesn't start with `^` and use spaces

Geal: updated in 935ecf91e

o0Ignition0o (approved):
One question, otherwise lgtm.

PS: I wonder if this would warrant a custom hash impl for every Type in `apollo-rs` or somewhere else.

bnjjj (approved):
For the future I think we should provide or fill the context with the `reused` value to know if it has been reused or not and also the cache key itself as a selector if people would like to debug and include this value in logs or traces for example. Same for `reused` value as a selector

apollo-router/src/query_planner/caching_query_planner.rs
```diff
@@ -122,24 +113,34 @@ where
...
             crate::configuration::QueryPlannerMode::Legacy => {
-                ConfigMode::Js(Arc::new(configuration.js_query_planner_config()))
+                configuration.js_query_planner_config().hash(&mut hasher);
+
+                configuration.rust_query_planner_config().hash(&mut hasher);
             }
             crate::configuration::QueryPlannerMode::Both => {
-                ConfigMode::Both(Arc::new(configuration.js_query_planner_config()))
+                configuration.js_query_planner_config().hash(&mut hasher);
             }
```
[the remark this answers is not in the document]
Geal: fixed in 1a462111c

apollo-router/src/spec/query/change.rs
```diff
@@ -33,14 +33,12 @@ pub(crate) const JOIN_TYPE_DIRECTIVE_NAME: &str = "join__type";
 pub(crate) struct QueryHashVisitor<'a> {
```
BrynCooke, line 33: This file needs an extensive rust doc.

Describe the algorithm, and details of why the special literals are needed.

BrynCooke (changes requested):
I'd like to see a test coverage report of `change.rs` using JUST the tests from that file before we merge.

Unions and enums should be covered in their own tests, but I can't see them in the diff. Maybe you already covered them separately.

.changesets/fix_geal_fix_hashing_algorithm.md
```diff
@@ -0,0 +1,26 @@
+### Fix cache key hashing algorithm ([Issue #5160](https://github.com/apollographql/router/issues/5160))
+
+> [!IMPORTANT]
+> If you have enabled [Distributed query plan caching](https://www.apollographql.com/docs/router/configuration/distributed-caching/#distributed-query-plan-caching), this release changes the hashing algorithm used for the cache keys.  On account of this, you should anticipate additional cache regeneration cost when updating between these versions while the new hashing algorithm comes into service.
+
+The Router contains a schema aware query hashing algorithm, designed to return the same hash across schema updates if they do not affect the query. It was used mainly to avoid planning a lot of queries when warming up the query plan cache. It was deactivated due to a regression. This reintroduces this algorithm, with a serie of reliability and performance fixes, along with better observability.
```
garypen, line 6:
```suggestion
The Router contains a schema aware query hashing algorithm, designed to return the same hash across schema updates if they do not affect the query. It was used mainly to avoid planning a lot of queries when warming up the query plan cache. It was deactivated due to a regression. This reintroduces this algorithm, with a number of reliability and performance fixes, along with better observability.
```

.changesets/fix_geal_fix_hashing_algorithm.md
```diff
@@ -0,0 +1,26 @@
...
+
+The hashing algorithm can be activated in configuration, if query plan cache warm up is already enabled:
```
garypen, line 8: I don't understand why cache warm up is mentioned here. There's no connection is there?

```suggestion
The hashing algorithm can be activated in configuration.
```

.changesets/fix_geal_fix_hashing_algorithm.md
```diff
@@ -0,0 +1,26 @@
...
+
+```yaml title="router.yaml"
+supergraph:
+  query_planning:
+    warmed_up_queries: 100
+    experimental_reuse_query_plans: true
+```
```
garypen, line 15: If removing the reference to warm-up, also remove the configuration.

```suggestion
```yaml title="router.yaml"
supergraph:
  query_planning:
    experimental_reuse_query_plans: true
```
```

File: .changesets/fix_geal_fix_hashing_algorithm.md [hunk 7]
```diff
@@ -0,0 +1,26 @@
...
+
+There is a counter metric named `apollo.router.query.planning.warmup.reused` that can be used to track the hashing algorithm benefits:
+- if the `experimental_reuse_query_plans` option is false, the `query_plan_reuse_active` metric attribute will be false. Cache warm up will not reuse query plans according to the algorithm, but it will evaluate if some of them could have been reused and report that in the metric
+- if the `experimental_reuse_query_plans` option is true, then the `query_plan_reuse_active` metric attribute will be true
```
garypen, line 19: Since this isn't the main thrust of the changelog (which is about the fixed key hashing algorithm), we should introduce this. Something like:

```
We've also introduced changed to our metrics that are helpful in tracking the impact of these changes.
```

.changesets/fix_geal_fix_hashing_algorithm.md
```diff
@@ -0,0 +1,26 @@
...
+
+Fixes included in this change:
+- strenghten the hashing algorithm to prevent collisions. In particular, the query string is always hashed in, making sure that different queries cannot get the same hash
```
garypen, line 22:
```suggestion
Summary:
- strengthen the hashing algorithm to prevent collisions. In particular, the query string is always hashed in, making sure that different queries cannot get the same hash
```

apollo-router/src/query_planner/caching_query_planner.rs
```diff
@@ -358,6 +369,11 @@ where
         }
 
         tracing::debug!("warmed up the query planner cache with {count} queries planned and {reused} queries reused");
+
+        ::tracing::info!(
+            monotonic_counter.apollo.router.query.planning.warmup.reused = reused,
+            query_plan_reuse_active = experimental_reuse_query_plans
+        );
```
garypen, line 376: Update to use the new metrics mechanism? `u64_counter!()`

docs/source/configuration/in-memory-caching.mdx
```diff
@@ -102,6 +102,8 @@ supergraph:
     experimental_reuse_query_plans: true
 ```
 
+To check if the Router deployment would benefit from this option, check the counter metric named `apollo.router.query.planning.warmup.reused` that indicates how many query plans were reused in warmup if its `query_plan_reuse_active` is true, or how many query plans would have been reused if the option was active, when `query_plan_reuse_active` is false.
+
```
garypen, line 106:
```suggestion
To decide if a Router deployment would benefit from this option, check the counter metric named `apollo.router.query.planning.warmup.reused`. This metric indicates how many query plans were reused, or would have been reused, depending on  whether `query_plan_reuse_active` is true or false.
```

apollo-router/src/spec/query/change.rs
```diff
@@ -348,16 +454,45 @@ impl<'a> Hasher for QueryHashVisitor<'a> {
     }
 
     fn write(&mut self, bytes: &[u8]) {
+        // FIXME: hack I used to debug my code, remove
+        // if bytes.len() != 1 || bytes[0] != 0xFF {
+        //     println!("{:?}", std::str::from_utf8(bytes).unwrap());
+        // }
+        // byte separator between each part that is hashed
+        self.hasher.update(&[0xFF][..]);
```
garypen, line 462: To be removed before merge.

apollo-router/src/query_planner/caching_query_planner.rs
```diff
@@ -626,48 +647,105 @@ impl std::fmt::Display for CachingQueryKey {
...
+        self.plan_options.hash(&mut hasher);
+        self.config_mode.hash(&mut hasher);
+        self.introspection.hash(&mut hasher);
         let metadata = hex::encode(hasher.finalize());
 
         write!(
             f,
-            "plan:{}:{}:{}:{}:{}",
+            "plan:cache:{}:federation:{}:{}:opname:{}:metadata:{}",
             CACHE_KEY_VERSION, FEDERATION_VERSION, self.hash, operation, metadata,
```
IvanGoncharov, line 660: `FEDERATION_VERSION` is coming from an environment variable meaning it can contain `:`
Maybe we need to check if it contains `:` and panic if it does.
It's not possible to exploit it on its own just a suggestion to be extra careful.

apollo-router/src/query_planner/caching_query_planner.rs
```diff
@@ -122,24 +113,32 @@ where
         let enable_authorization_directives =
             AuthorizationPlugin::enable_directives(configuration, &schema).unwrap_or(false);
 
-        let config_mode = match configuration.experimental_query_planner_mode {
+        let mut hasher = StructHasher::new();
+        match configuration.experimental_query_planner_mode {
```
IvanGoncharov, line 117: If we hashing different structures we need to add some prefix otherwise we can have collision where values are matching in different structures (note: property names are not hashed only values).

IvanGoncharov (changes requested):

apollo-router/src/spec/query/change.rs
```diff
@@ -132,58 +165,79 @@ impl<'a> QueryHashVisitor<'a> {
...
+                "^object{".hash(self);
                 for (k, v) in o.iter() {
+                    "^key".hash(self);
+
                     k.hash(self);
-                    ":".hash(self);
+                    "^:".hash(self);
                     self.hash_value(v);
                 }
                 "}".hash(self);
```
IvanGoncharov, line 183:
```suggestion
                "^}".hash(self);
```

For consistency, since we use `^]` to end the list.

IvanGoncharov (changes requested):

.changesets/fix_geal_fix_hashing_algorithm.md
Geal: it needs the warmup to be usable

.changesets/fix_geal_fix_hashing_algorithm.md
Geal: same as above, it is not usable without the warmup

apollo-router/src/query_planner/caching_query_planner.rs
Geal: done in df8e6fc32

apollo-router/src/query_planner/caching_query_planner.rs
Geal: do you see a scenario where this would cause a collision?

apollo-router/src/query_planner/caching_query_planner.rs
Geal: done in 52e27df18

docs/source/configuration/in-memory-caching.mdx
Geal: that makes it less clear what `query_plan_reuse_active` does. Is there another way to rephrase it?

apollo-router/src/spec/query/change.rs
Geal: done in 66c9a0001

Geal:
now with  bf734b92c  we are now at a coverage of 99.40% (1826 lines out of 1837)
That should be enough for now

apollo-router/src/spec/query/change.rs
```diff
@@ -366,30 +544,45 @@ impl<'a> Visitor for QueryHashVisitor<'a> {
         field_def: &ast::FieldDefinition,
         node: &executable::Field,
     ) -> Result<(), BoxError> {
+        "^VISIT_FIELD".hash(self);
+
         if !self.seen_introspection && (field_def.name == "__schema" || field_def.name == "__type")
```
IvanGoncharov, line 549: That fields only have a special meaning for the query root type.
```
{
   foo {
     __schema # <- not an introspection field here
   }
}
```
Technically it should be forbided to have such names:
https://spec.graphql.org/draft/#sec-Names.Reserved-Names

But I can't find code that implements this validation in `apollo-rs`
https://github.com/search?q=repo%3Aapollographql%2Fapollo-rs++%5C%22__%5C%22&type=code

Also, it's not present here:
https://github.com/apollographql/apollo-rs/blob/main/crates/apollo-compiler/src/validation/field.rs#L155

I will report it to apollo-rs but I also think we need to be sure those are at the query root to count as introspection.

IvanGoncharov (changes requested):

apollo-router/src/spec/query/change.rs
```diff
@@ -348,16 +501,41 @@ impl<'a> Hasher for QueryHashVisitor<'a> {
...
+                "^VISIT_OPERATION-NO_DEFAULT".hash(self);
+            }
+
+            for directive in &variable.directives {
+                self.hash_directive(directive);
+            }
+        }
+        for directive in &node.directives {
+            self.hash_directive(directive);
+        }
```
IvanGoncharov, line 534: It's vulnerable to collisions, I wrote test for it:
```rust
    #[test]
    fn it_change_directive_location() {
        let schema: &str = r#"
        directive @foo on QUERY | VARIABLE_DEFINITION

        type Query {
          field(arg: String): String
        }
        "#;

        let query_one = r#"
        query Test ($arg: String @foo) {
          field(arg: $arg) 
        }
        "#;
        let query_two = r#"
        query Test ($arg: String) @foo {
          field(arg: $arg) 
        }
        "#;

        assert_ne!(
            hash(schema, query_one).from_hash_query,
            hash(schema, query_two).from_hash_query
        );
        assert_ne!(
            hash(schema, query_one).from_visitor,
            hash(schema, query_two).from_visitor
        );
    }

```

IvanGoncharov (changes requested):

apollo-router/src/spec/query/change.rs
IvanGoncharov: Some places have similar code where stuff is hashed in a cycle, but they are not prefixed by anything as a group.
I can't trigger collision in those places because they are not followed by stuff that can cause collision but that could change in future.
For example:
https://github.com/apollographql/router/blob/81604c235b8c7e601204e804194bb381df576d71/apollo-router/src/spec/query/change.rs#L319-L324
Currently you can't cause a collision, but it's better to still wrap it in a "input field definition" tag.

IvanGoncharov: so idea is if you have has of multiple stuff in a cycle those should be prefixed as a group.

IvanGoncharov (changes requested):
This test is failing:
```rust
        let schema_one: &str = r#"
        interface SomeInterface {
          value: String
        }

        type Foo implements SomeInterface {
          value: String
        }

        type Bar implements SomeInterface {
          value: String
        }

        union FooOrBar = Foo | Bar

        type Query {
          fooOrBar: FooOrBar
        }
        "#;
        let schema_two: &str = r#"
        interface SomeInterface {
          value: String
        }

        type Foo {
          value: String # <= This field shouldn't be a part of query plan anymore
        }

        type Bar implements SomeInterface {
          value: String
        }

        union FooOrBar = Foo | Bar

        type Query {
          fooOrBar: FooOrBar
        }
        "#;

        let query = r#"
        {
          fooOrBar {
            ... on SomeInterface {
              value
            }
          } 
        }
        "#;

        assert_ne!(
            hash(schema_one, query).from_hash_query,
            hash(schema_two, query).from_hash_query
        );
        assert_ne!(
            hash(schema_one, query).from_visitor,
            hash(schema_two, query).from_visitor
        );
```

It's either reference to interface should hash all implementors or fragments should have some special handling

apollo-router/src/spec/query/change.rs
Geal: I am opening https://github.com/apollographql/apollo-rs/issues/903 to track it. This will not cause issue for this algorithm, because if `__schema` appears in the wrong place, then the only consequence is that the schema string will be entirely hashed in, so that would not add any collision risk

apollo-router/src/spec/query/change.rs
Geal: that's a good point. This should be fixed in [1b76c0a](https://github.com/apollographql/router/pull/5255/commits/1b76c0ad01743272dfaf4497d87760ea9d265b1f)

apollo-router/src/spec/query/change.rs
SimonSapin: Indeed we are missing a validation rule to forbid definitions with a `__` prefix: https://github.com/apollographql/apollo-rs/issues/743

Geal:
@IvanGoncharov good call on hashing the interface implementors, this is done in [555e507](https://github.com/apollographql/router/pull/5255/commits/555e5076321c1a9297044e963f66919b6d7dcdbe)
Now I am wondering about this `//FIXME: hash implemented interfaces?` https://github.com/apollographql/router/pull/5255/files#diff-b21dc8758c509f67655b816231a0ac2c2fd978c90b0ced9ce88a52fc74324e12R285
I think we should hash the interfaces implemented by an object, but don't see any failure case right now. Maybe with things like authorization, where there's some behaviour that needs to match between interface and implementor

apollo-router/src/query_planner/caching_query_planner.rs
```diff
@@ -649,42 +745,101 @@ impl std::fmt::Display for CachingQueryKey {
...
 
-        let mut hasher = Sha256::new();
-        hasher.update(serde_json::to_vec(&self.metadata).expect("serialization should not fail"));
-        hasher
-            .update(serde_json::to_vec(&self.plan_options).expect("serialization should not fail"));
-        hasher
-            .update(serde_json::to_vec(&self.config_mode).expect("serialization should not fail"));
-        hasher.update(&*self.schema_id);
+        let mut hasher = StructHasher::new();
+        self.metadata.hash(&mut hasher);
```
Geal, line 749: should hash key names, not only values

Geal:
now that https://github.com/apollographql/router/pull/6206 is merged into dev, [59d1e2c](https://github.com/apollographql/router/pull/5255/commits/59d1e2c17f894d5f73b34d089c8de5e56a763611) merges back those changes here.

IvanGoncharov:
Converting to draft due to internal discussion regarding the risks introduced by this PR.
We want to add "Query hash stable to incremental schema changes" as the Router's feature.
However, in its current form, it is implemented outside of Query Planner and without any mechanism to determine what parts of the schema are used by Query Planner.

For example, adding new directives that influence Query planning without updating this code can result in a hash collision where a schema update changes the query plan, but the hash remains the same.

BrynCooke:
Closing as this is a year stale

PR #5259, opened 2024-05-28
Add doc strings for certain confusing parts of source-aware graph pat…

sachindshinde:
This PR adds doc strings to certain fields in source-aware graph path and path tree data structures that were causing confusion, to assist in code for path tree generation.


apollo-federation/src/source_aware/federated_query_graph/graph_path.rs
```diff
@@ -27,9 +27,20 @@ pub(crate) struct FederatedGraphPath {
...
+    /// resolution happens at this edge's head for some later source-entering edge's condition
+    /// during option generation, that resolution's information (including ID) is stored in this
+    /// map, keyed by condition ID.
     source_entering_condition_resolutions_at_head:
         IndexMap<SelfConditionIndex, ConditionResolutionInfo>,
+    /// Conditions are imposed by query graph edges and resolved at earlier query graph nodes. When
+    /// resolution happens at this edge's head for some later non-source-entering edge's condition
+    /// during option generation, that resolution's information (including ID) is stored in this
+    /// map, keyed by condition ID.
     condition_resolutions_at_head: IndexMap<SelfConditionIndex, ConditionResolutionInfo>,
```
goto-bus-stop, line 44: sort of raises a new question: why are source-entering edges and other kind of edges treated separately here, instead of being in a single map?

goto-bus-stop (commented):
This helps me based on our conversation yesterday. I think understanding the design of conditions as a whole is useful to *really* makes sense of what these fields do.

apollo-federation/src/source_aware/federated_query_graph/graph_path.rs
sachindshinde: @goto-bus-stop 
In graph paths, they're stored separately to allow resolution of non-source-entering-edge conditions to be "biased" by the previous resolution of source-entering-edge conditions. The idea is that non-source-entering-edge condition resolution (e.g. for `@requires`) may be different, if it were to know how source-entering conditions were resolved; that's what the current `handleRequires()` code is doing in some sense (though very painfully). E.g. if the source-entering-edge condition were resolved by making API calls to some source `A`, and a `@requires` condition being resolved at the same location needs to get a field `foo` and could do so from either `A` or `B`, then you'd want it to also call source `A`.

In path trees, I'm not 100% sure if we still need the separation there (I've left it there in case a later-milestone algorithm might care). My intention is that we can remove the separation later if it turns out it's not needed in path trees.

lrlna (approved):

apollo-bot2:

PR #5262, opened 2024-05-28
Reduce allocations made for demand control processing

tninesling:
Instead of zipping request schemas to their responses and storing in an intermediate data structure, this implements the traversal of field/value pairs as a visitor without new allocations.

Most of the changes account for the fact that SchemaAwareResponse/TypedValue no longer exist, as we give consumers the apollo_compiler Field or serde_json_bytes Value directly. Additionally, the code previously in SchemaAwareResponse/TypedValue lives in `response.rs`, since it's common to a couple different plugins.

---

**Checklist**

- Tests added and passing[^3]
    - [X] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


BrynCooke (dismissed):

BrynCooke (commented):
Looks like there are some CI failures

apollo-router/src/plugins/telemetry/config_new/graphql/mod.rs
```diff
@@ -222,9 +228,28 @@ struct GraphQLInstrumentsVisitor<'a> {
...
-        self.instruments.on_response_field(value, self.ctx);
+impl<'a> ResponseVisitor for GraphQLInstrumentsVisitor<'a> {
+    fn visit_field(
+        &mut self,
+        request: &ExecutableDocument,
+        ty: &NamedType,
+        field: &Field,
+        value: &Value,
+    ) {
+        self.instruments
```
tninesling, line 239: This is required if we want to capture list length, but it also means `field` gets overcounted when it's also passed into the list items below. We could account for this in the attributes, but I don't love the idea of this implementation detail leaking into the metrics code.

apollo-router/src/plugins/telemetry/config_new/graphql/mod.rs
```diff
@@ -278,7 +303,7 @@ pub(crate) mod test {
                 2.0,
                 "graphql.field.name" = "users",
                 "graphql.field.type" = "User",
-                "graphql.type.name" = "Query"
+                "graphql.type.name" = "User"
```
tninesling, line 306: I still don't totally understand the different between `graphql.field.type` and `graphql.type.name`. For now, they're the same thing because the current implementation of the visitor doesn't share the parent's type. It looks like we have some code that should map types to one of `list`, `object`, `scalar`, etc. Is that meant to be used here?

apollo-router/tests/integration/telemetry/metrics.rs
```diff
@@ -198,21 +198,21 @@ async fn test_graphql_metrics() {
...
+        .assert_metrics_contains(r#"graphql_field_list_length_bucket{graphql_field_name="topProducts",graphql_field_type="Product",graphql_type_name="Product",otel_scope_name="apollo/router",le="5"} 1"#, None)
         .await;
     router
-        .assert_metrics_contains(r#"graphql_field_execution_total{graphql_field_name="name",graphql_field_type="String",graphql_type_name="Product",otel_scope_name="apollo/router"} 3"#, None)
+        .assert_metrics_contains(r#"graphql_field_execution_total{graphql_field_name="name",graphql_field_type="String",graphql_type_name="String",otel_scope_name="apollo/router"} 3"#, None)
         .await;
     router
-        .assert_metrics_contains(r#"graphql_field_execution_total{graphql_field_name="topProducts",graphql_field_type="Product",graphql_type_name="Query",otel_scope_name="apollo/router"} 1"#, None)
+        .assert_metrics_contains(r#"graphql_field_execution_total{graphql_field_name="topProducts",graphql_field_type="Product",graphql_type_name="Product",otel_scope_name="apollo/router"} 4"#, None)
```
tninesling, line 210: There are some type name changes in all of these, for which I've already posted a discussion point. This line specifically changes the count as well, which highlights the overcounting issue for array fields. Right now, the three attributes we're checking will be equivalent for a list field and all it's items, so we need to differentiate in some way.

BrynCooke (approved):
Looks good to me
I have a followup PR to further move things around once we get this merged. https://github.com/apollographql/router/pull/5339

PR #5263, opened 2024-05-28
PoC with experimental updated line/col support in apollo-compiler

dylan-apollo:
Just demonstrating the required changes if https://github.com/apollographql/apollo-rs/pull/861 is accepted. This is based on my latest `connectors` branch so I can test what I need e2e. The only changes relevant to `dev` are in `apollo-router/src/error.rs` (which, maybe should be using the same `GraphQLError` struct as `apollo-compiler`)?


abernix:
Feel free to re-open.

PR #5265, opened 2024-05-29
Set Apollo metrics generation mode to new by default

bonnici:
Changes the default value of `experimental_apollo_metrics_generation_mode` to be `new` since all metrics are showing that identical signatures are references are being generated.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/query_planner/bridge_query_planner.rs
```diff
@@ -581,6 +581,12 @@ impl BridgeQueryPlanner {
                     usage_reporting.stats_report_key = sig;
                 }
 
+                tracing::info!(
+                    "experimental_apollo_metrics_generation_mode: {:?}",
+                    self.configuration
+                        .experimental_apollo_metrics_generation_mode
+                );
```
bnjjj, line 588: if your intent is to create a metric, that won't work like this. You should take a look [here](https://github.com/apollographql/router/blob/dev/apollo-router/src/configuration/metrics.rs#L176)

bonnici: Ah sorry, that was temporary code I should have deleted.

bonnici:
closing in favour of https://github.com/apollographql/router/pull/5330

PR #5266, opened 2024-05-29
Fix running `cargo test` without nextest

SimonSapin:
Reproduction: `cargo test -p apollo-federation -- -q`

## Background

https://github.com/apollographql/router/pull/5157 introduced a new kind of snapshot test for the Rust query planner. Test schemas are specified as sets of subgraph schemas, so composing them is needed. Since we don’t have composition in Rust yet, the tests rely on JS composition through Rover. To avoid a dependency on Rover on CI and for most contributors, the composed supergraph are cached in the repository. Rover use is opt-in and only required when a cached file is missing or out of date. (Composition inputs are hashed.)

The file name is derived (through a macro) from the test function name. To detect name conflicts, a static `OnceLock<Mutex<HashSet<&'static str>>>` is used to ensure no name is used more than once.

## The problem

The static hash set relies on all snapshot tests running in the same process. This is the case with `cargo test`, but `cargo nextest run` as used on CI isolates every test in its own process. This breaks conflict detection of cache file names for composed supergraph schemas, since each test only "sees" itself in the static hash set.

https://github.com/apollographql/router/pull/5240 introduced a name conflict: composition is used in a function called twice with different arguments. Because nextest was used both locally and on CI, the conflict went undectected. As a result, running `cargo test` on dev fails because the conflict is detected.

## This PR

This PR fixes this case of cache file name conflict, but conflict detection is still broken with nextest.

As a result it’s possible that this kind of conflict could happen again and be merged undectected.

## Non-solutions tried

* Nextest has a notion of [test groups](https://nexte.st/book/test-groups), but they don’t appear to let multiple tests run in the same process

* Instead of relying on the runtime side effect of tests, could conflict detection rely on enumerating tests at compile time? The [`linkme` crate](https://crates.io/crates/linkme) is a building block Router used to register Rust plugins. It could be used here in all composition inputs can be made const, but `std::any::type_name` used in a macro to extract the current function name is not `const fn` [yet](https://github.com/rust-lang/rust/issues/63084)

## Potential solutions

* Remove the cache and accept the dependency on Rover for testing. This impacts CI and all contributors.

* Require every `planner!` macro invocation to specify an explicit cache file name instead of relying on the function name. Then conflict detection can use `linkme`.

* Move conflict detection to a separate test that something something parses Rust source files of other tests something

### DIFF
--- a/apollo-federation/tests/query_plan/build_query_plan_tests/named_fragments_preservation.rs
+++ b/apollo-federation/tests/query_plan/build_query_plan_tests/named_fragments_preservation.rs
@@ -290,25 +290,7 @@ fn it_avoid_fragments_usable_only_once() {
     );
 }
 
-#[test]
-#[should_panic(expected = "not yet implemented")]
-// TODO: investigate this failure
-
-fn respects_query_planner_option_reuse_query_fragments_true() {
-    respects_query_planner_option_reuse_query_fragments(true)
-}
-#[test]
-#[should_panic(expected = "not yet implemented")]
-// TODO: investigate this failure
-
-fn respects_query_planner_option_reuse_query_fragments_false() {
-    respects_query_planner_option_reuse_query_fragments(false)
-}
-
-fn respects_query_planner_option_reuse_query_fragments(reuse_query_fragments: bool) {
-    let planner = planner!(
-      config = QueryPlannerConfig {reuse_query_fragments, ..Default::default()},
-      Subgraph1: r#"
+const SUBGRAPH: &str = r#"
             type Query {
               t: T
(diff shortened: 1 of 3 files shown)


lrlna (approved):

lrlna:
> Require every `planner!` macro invocation to specify an explicit cache file name instead of relying on the function name. Then conflict detection can use `linkme`.

I potentially don't mind this. What are the potential downsides to this, except for having to manually write out file names?

SimonSapin:
Right, that might not be so bad. It’s just yet an extra step when porting or adding new tests.

PR #5273, opened 2024-05-29
chore(apollo-federation): port skip and include snapshot tests

lrlna:
Ports all the tests in `query-planner-js/src/__tests__/buildPlan.directiveMerging.test.ts`. The first test should not panic as soon as #5207 is merged. 

There is something iffy going on with the way snapshots are formatted though. Going to investigate the macro that's stringifying the query plan snapshot next.


apollo-bot2:

PR #5283, opened 2024-05-29
Fix a couple of error messages and prevent subscriptions in connectors subgraphs

dylan-apollo:


apollo-federation/src/sources/connect/models/validation.rs
```diff
@@ -225,11 +225,26 @@ struct SourceDirective {
...
+    let source_map = &schema.sources;
+    let is_subscription = schema
+        .schema_definition
+        .subscription
+        .as_ref()
+        .is_some_and(|sub| sub.name == object.name);
+    if is_subscription {
+        return vec![Error {
+            code: ErrorCode::SubscriptionInConnectors,
+            message: "connectors can't be used to create subscriptions".to_string(),
```
lennyburdette, line 242:
```suggestion
            message: "The `Subscription` root type is not supported when using `@{connect_directive_name}`".to_string(),
```
maybe?

lennyburdette (approved):

apollo-federation/src/sources/connect/models/validation.rs
dylan-apollo: I was avoiding the `Subscription` name since it can be renamed, but just using the object type could also be confusing. I'll do something similar without the "proper" noun, though.

PR #5290, opened 2024-05-30
feat(federation): source-aware PathTree creation

goto-bus-stop:
Opening for self-review, needs some work.

### DIFF
--- a/apollo-federation/src/source_aware/federated_query_graph/graph_path.rs
+++ b/apollo-federation/src/source_aware/federated_query_graph/graph_path.rs
@@ -40,14 +40,14 @@ pub(crate) enum OperationPathElement {
     InlineFragment(InlineFragment),
 }
 
-#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+#[derive(Debug, Clone, Copy, Eq, PartialEq, Hash)]
 pub(crate) struct ConditionResolutionId(usize);
 
 #[derive(Debug, Clone)] // TODO(@goto-bus-stop): should this be cloneable?
 pub(crate) struct ConditionResolutionInfo {
-    id: ConditionResolutionId,
-    resolution: Arc<FederatedPathTree>,
-    cost: QueryPlanCost,
+    pub id: ConditionResolutionId,
+    pub resolution: Arc<FederatedPathTree>,
+    pub cost: QueryPlanCost,
 }
 
 #[derive(Debug, Clone)]
--- a/apollo-federation/src/source_aware/federated_query_graph/graph_path.rs
+++ b/apollo-federation/src/source_aware/federated_query_graph/graph_path.rs
@@ -25,12 +25,13 @@ pub(crate) struct FederatedGraphPath {
 
 #[derive(Debug)]
 pub(crate) struct Edge {
-    operation_element: Option<Arc<OperationPathElement>>,
-    edge: Option<EdgeIndex>,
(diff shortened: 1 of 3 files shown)


apollo-federation/src/source_aware/federated_query_graph/graph_path.rs
```diff
@@ -25,12 +25,13 @@ pub(crate) struct FederatedGraphPath {
 
 #[derive(Debug)]
 pub(crate) struct Edge {
-    operation_element: Option<Arc<OperationPathElement>>,
-    edge: Option<EdgeIndex>,
-    self_condition_resolutions_for_edge: IndexMap<SelfConditionIndex, ConditionResolutionId>,
-    source_entering_condition_resolutions_at_head:
+    pub operation_element: Option<Arc<OperationPathElement>>,
```
goto-bus-stop, line 28: It might not be correct for these to be pub.

graph paths and path trees seem quite intertwined (since a path tree is made up of graph paths and a graph path can have conditions expressed as path trees). It might make more sense for both of them to be in the same module so they can freely access each other's internals.

apollo-federation/src/source_aware/federated_query_graph/path_tree.rs
```diff
@@ -27,8 +32,255 @@ pub(crate) struct Child {
...
+    ) -> Result<Self, FederationError> {
+        Self::from_paths_inner(
+            graph,
+            node,
+            paths.map(|(path, selection_set)| (path.edges(), selection_set)),
+            &mut Default::default(),
+        )
+    }
+
+    // Path trees should not normally be cloned. This manual implementation is private
```
goto-bus-stop, line 222: I'm not sure if this is true. I just didn't want to add the public API without being sure.

apollo-federation/src/source_aware/federated_query_graph/path_tree.rs
```diff
@@ -17,18 +22,270 @@ pub(crate) struct FederatedPathTree {
...
+                    // For a "None" edge, stay on the same node
+                    node
+                },
+                by_unique_trigger: IndexMap::new(),
+            });
+
+            match for_edge
+                .by_unique_trigger
+                .entry(edge.operation_element.clone())
+            {
```
goto-bus-stop, line 163: This handles paths having potentially multiple different triggers for the same edge. What does that mean in practice? Two different fields can never share an edge in the query graph, right?

apollo-federation/src/source_aware/federated_query_graph/path_tree.rs
```diff
@@ -17,18 +22,270 @@ pub(crate) struct FederatedPathTree {
...
+fn merge_condition_resolutions(
+    condition_resolutions: &mut IndexMap<SelfConditionIndex, ConditionResolutionInfo>,
+    other_condition_resolutions: &IndexMap<SelfConditionIndex, ConditionResolutionInfo>,
+    remapped_condition_ids: &mut HashMap<ConditionResolutionId, ConditionResolutionId>,
+) {
+    for (&condition_index, resolution) in other_condition_resolutions {
+        condition_resolutions
+            .entry(condition_index)
+            .and_modify(|existing_resolution| {
+                if existing_resolution.cost > resolution.cost {
```
goto-bus-stop, line 56: We think the condition ID should always be the same. I could add a `debug_assert!()` here.

lrlna (approved):

PR #5300, opened 2024-05-31
Enable capability of using http2 cleartext (h2c) when communicating with coprocessor

### LINKED ISSUE #5299: External Coprocessor: Support for connections to be made with h2c (http2 cleartext/prior knowledge)
Router has already implemented support for h2c for subgraph requests: https://github.com/apollographql/router/pull/3852
This request is to provide the same functionality to enable configuring the External Coprocessor plugin to use h2c when making requests to a remote External Coprocessor endpoint. This would be when the endpoint service is deployed as a standalone application vs. as a sidecar container in the same Kubernetes pod.

~In addition, support calling the External Coprocessor over a Unix domain socket when the endpoint is deployed as a sidecar container.~
~https://github.com/apollographql/router/pull/4757~

Expose support for h2c ~and Unix domain socket~ configuration options for the External Coprocessor plugin via the router.yaml file.

**Update 2024-07-29**
Pull Unix Domain Sockets support into a separate feature request.
https://github.com/apollographql/router/issues/5739

theJC:
Add http2 cleartext (h2c) communication support to be able to be used to communicate to a coprocessor.  This aligns with the h2c support that was added to subgraph communications in https://github.com/apollographql/router/pull/3852

Allowing h2c provides flexibility in getting http2 performance without the TLS overhead in applications that do not require (or do not support) the communications being encrypted.

Stole a copy of the Enum Http2Config for CoprocessorHttp2Config and aligned the code to be more aligned to whats is found in services/http/service.rs

Fixes the h2c aspect of #5299

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-351]: https://apollographql.atlassian.net/browse/ROUTER-351?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apollo-router/src/plugins/coprocessor/mod.rs
```diff
@@ -282,6 +301,7 @@ pub(super) struct SubgraphResponseConf {
 struct Conf {
     /// The url you'd like to offload processing to
     url: String,
+    experimental_http2: Option<CoprocessorHttp2Config>,
```
Geal, line 304: could you put the option under a `traffic_shaping` section? Since we will probably extend it with more options, this will allow us to put them in the same place (and then later easily extend to stage specific config).
Also, could this use the [`Http2Config` from the traffic shaping plugin](https://github.com/apollographql/router/blob/b4dda0cf5739fae58313e25dd04ea64159b6a717/apollo-router/src/plugins/traffic_shaping/mod.rs#L77-L85)? This way the config schema would refer to the existing type instead of adding a new one

Geal (changes requested):
this looks good, it just needs a few changes on the configuration side

apollo-router/src/plugins/coprocessor/mod.rs
theJC: ROFL, I had done that first, but then thought you might want to de-couple from that.  I'll return back to not having the separate config.

theJC: traffic_shaping section is now added, now using Http2Config from traffic shaping plugin

apollo-router/src/plugins/coprocessor/mod.rs
```diff
@@ -77,19 +78,26 @@ impl Plugin for CoprocessorPlugin<HTTPClientService> {
...
-        let connector = hyper_rustls::HttpsConnectorBuilder::new()
+        let builder = hyper_rustls::HttpsConnectorBuilder::new()
             .with_tls_config(tls_config)
             .https_or_http()
-            .enable_http1()
-            .enable_http2()
-            .wrap_connector(http_connector);
+            .enable_http1();
+
+        let connector = if init.config.traffic_shaping.is_none() || init.config.traffic_shaping.as_ref().unwrap().experimental_http2 == Some(Http2Config::Enable) {
```
theJC, line 86: Considering the case of new version delivered by customer with, wanting to keep same behavior if no change to their config yaml.... if they havent provided the newly added traffic_shaping, or if they have provided it with but have not used the disabled http2 enum value, then enable http2.

apollo-router/src/plugins/coprocessor/mod.rs
```diff
@@ -77,19 +78,26 @@ impl Plugin for CoprocessorPlugin<HTTPClientService> {
...
-        let connector = hyper_rustls::HttpsConnectorBuilder::new()
+        let builder = hyper_rustls::HttpsConnectorBuilder::new()
             .with_tls_config(tls_config)
             .https_or_http()
-            .enable_http1()
-            .enable_http2()
-            .wrap_connector(http_connector);
+            .enable_http1();
+
+        let connector = if init.config.traffic_shaping.is_none() || init.config.traffic_shaping.as_ref().unwrap().experimental_http2 != Some(Http2Config::Disable) {
```
theJC, line 86: Considering the case of new version of Router delivered to customer with, wanting to keep same behavior if no change to their config yaml.... if they haven't provided the newly added traffic_shaping, or if they have provided it with but have not used the disabled http2 enum value, then enable http2.

apollo-router/src/plugins/coprocessor/mod.rs
```diff
@@ -77,19 +78,26 @@ impl Plugin for CoprocessorPlugin<HTTPClientService> {
...
+        } else {
+            builder.wrap_connector(http_connector)
+        };
 
         let http_client = ServiceBuilder::new()
             .layer(TimeoutLayer::new(init.config.timeout))
             .service(
                 hyper::Client::builder()
+                    .http2_only(
+                        init.config.traffic_shaping.is_some() && init.config.traffic_shaping.as_ref().unwrap().experimental_http2 == Some(Http2Config::Http2Only)
```
theJC, line 97: Maintain default behavior.  Only set http2_only if traffic_shaping is provided and the experimental_http2 is set to http2only

apollo-router/src/plugins/coprocessor/mod.rs
theJC: @Geal  -- Let me know next steps/what else we can do to get this over the finish line.

Geal (approved):
great, thanks

Geal:
@theJC could you run `cargo xtask lint --fmt` in the repository root? This should fix the remaining linting issues

bnjjj (changes requested):
Could you generate a changelog entry please ? It can be done using `cargo xtask changeset create`.

If you also could document your new configuration field in the traffic_shaping docs that would be great :)

theJC:
> If you also could document your new configuration field in the traffic_shaping docs that would be great :)

Hmmmm... at this point I had added a new traffic_shaping section under coprocessor, but now that I read this traffic_shaping.mdx doc, I'm wondering if that was the best place for it, or should there be a new coprocessor section added underneath the top-level traffic_shaping section?

bnjjj:
hmm good question @theJC. I think you're probably right my bad I first thought it was part of the traffic_shaping plugin, not the coprocessor one. So I think it would make sense to add this docs in the coprocessor docs. But still unsure about that configuration shape right now, maybe @BrynCooke will have more thoughts

BrynCooke:
My gut reaction is that we should have a common configuration struct for `Client` which can then be used across the config appropriately. That way the client options can be kept close to the functionality that uses that client.
```yaml
coprocessor:
  client:
    #client options
```

```yaml
subgraph:
  default:
    client:
      #client options
  services:
    products:
      client:
        # client options
```

Originally I had thought that traffic shaping would just be to hold things that shape the traffic, like rate limiting or throttling. So it's unfortunate that we have other stuff in there right now.

Needs a bit more thought though.

theJC:
@BrynCooke -  I've added an initial "shared" Client struct, only using it for coprocessor currently and switched over from TrafficShaping over to client.  Take a look and throw darts please.

abernix:
@bnjjj and @BrynCooke could you please take another look?  It would be nice for this to make the cut for our next release.

BrynCooke:
@theJC This is looking good. Can we move the Http2Config struct also into the shared package.
I've pre-approved though.

@Meschreiber Can we follow up to have a shared block in the docs for the http2 client config so that it can be reused in traffic shaping and in co-processor?

BrynCooke (approved):

Meschreiber:
> Can we follow up to have a shared block in the docs for the http2 client config so that it can be reused in traffic shaping and in co-processor? (Or if you have time then you can push to this branch).

Yup! If I have time before this branch is merged will add it here. Otherwise will start a new branch.

docs/source/customizations/coprocessor.mdx
```diff
@@ -92,6 +92,20 @@ coprocessor:
 
 In this case, the `RouterService` only sends a coprocessor request whenever it receives a client request. The coprocessor request body includes _no_ data related to the client request (only "control" data, which is [covered below](#coprocessor-request-format)).
 
+### Client configuration
+
+You can control the coprocessor client configuration as needed.  At the moment only experimental_http2 is supported.
```
Meschreiber, line 97:
```suggestion
<HttpConnection type="coprocessor" />
```

This should work fine once https://github.com/apollographql/docs/pull/889 merges.

(Tested locally.)

docs/source/customizations/coprocessor.mdx
```diff
@@ -92,6 +92,20 @@ coprocessor:
...
+
+For example to enable h2c (http2 cleartext) communication:
```
Meschreiber, line 99:
```suggestion
For example, to enable h2c communication with a coprocessor you can use this configuration:
```

docs/source/customizations/coprocessor.mdx
```diff
@@ -92,6 +92,20 @@ coprocessor:
...
+
+```yaml title="router.yaml"
+coprocessor:
+  url: http://127.0.0.1:8081 # When using an http (not https) url and http2only will result in connections that use h2c
+  client:
```
Meschreiber, line 104: Suggest moving comment to the next line for readability:

```suggestion
  url: http://127.0.0.1:8081
  # Using an HTTP (not HTTPS) URL and experimental_http2: http2only results in connections that use h2c
  client:
```

docs/source/customizations/coprocessor.mdx
Meschreiber, line 97:
```suggestion
<HttpConnection type="coprocessor" />
```

Introduced in https://github.com/apollographql/docs/pull/889 and used in https://github.com/apollographql/router/pull/5536.

(See deploy preview: https://deploy-preview-5536--apollo-router-docs.netlify.app/configuration/traffic-shaping#http2)

docs/source/customizations/coprocessor.mdx
Meschreiber, line 99:
```suggestion
For example, to enable h2c (http2 cleartext) communication with a coprocessor you can use this configuration::
```

docs/source/customizations/coprocessor.mdx
Meschreiber, line 104: Suggest moving comment to the next line for readability:

```suggestion
  url: http://127.0.0.1:8081
  # Using an HTTP (not HTTPS) URL and experimental_http2: http2only results in connections that use h2c
  client:
```

Meschreiber (commented):
Thanks very much for including documentation in your contribution! 🙏  (I'm a technical writer at Apollo and very much appreciate it!)

I've added some suggestions to make things more concise and consistent.

docs/source/customizations/coprocessor.mdx
abernix: Applied via 305a2b53ff8a1e918194f86ff0687235aa82b4a6.  Somehow all suggestions were duplicated.  I tried to accept the right ones. ;)

bnjjj (approved):

PR #5284, opened 2024-05-30
fixed a bug in compute_aliases_for_non_merging_fields

duckki:
- changed to use the originating schema to access the field's type
- ~fixes FED-230: QP internal error: Schema has no type "I"~
- ~Three tests failing with a type-related error are now failing with snapshot errors. But, that's acceptible for now.~

UPDATE: 5/31
- After rebasing on `dev`, the test changes are gone.
- So, this PR is now just a potential fix and refactoring that are reflected in tests.
- I can't get this PR merged due to repeatedly failing test-macos_test CI run). But, a low priority.

---

**Checklist**

Complete the checklist (and note appropriate exceptions) before the PR is marked ready-for-review.

- [X] Changes are compatible[^1]
- [ ] Documentation[^2] completed
- [ ] Performance impact assessed and acceptable
- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[^1]: It may be appropriate to bring upcoming changes to the attention of other (impacted) groups. Please endeavour to do this before seeking PR approval. The mechanism for doing this will vary considerably, so use your judgement as to how and when to do this.
[^2]: Configuration is an important part of many changes. Where applicable please try to document configuration examples.
[^3]: Tick whichever testing boxes are applicable. If you are adding Manual Tests, please document the manual testing (extensively) in the Exceptions.

PR #5301, opened 2024-05-31
Fixed a few normalize issues

duckki:
Added `normalize` call in `normalize_operation`.

Fixed a bug in NamedFragments::reduce

- normalize selection sets after retain_fragments call

Fixed InlineFragmentSelection::normalize

- Normalize sub-selections with the correct type

Sorry for mixing a few different changes in this PR. Unfortunately, those issues are intertwined in tests. Gathering them in one PR to minimize spurious test changes.

This depends on https://github.com/apollographql/router/pull/5284 in order to pass all tests.
This PR will improve a few tests failing in https://github.com/apollographql/router/pull/5214.

---

**Checklist**

- Tests added and passing[^3]
    - [X] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

### DIFF
--- a/apollo-federation/src/query_plan/operation.rs
+++ b/apollo-federation/src/query_plan/operation.rs
@@ -2877,10 +2877,18 @@ impl SelectionSet {
                 // in-place, we eagerly construct the selection that needs to be rebased on the target
                 // schema.
                 let element = OpPathElement::clone(ele);
-                let selection = Selection::from_element(
-                    element,
-                    selection_set.map(|set| SelectionSet::clone(set)),
-                )?;
+                let selection_set = selection_set.map(|selection_set| {
+                        selection_set.rebase_on(
+                            &element.sub_selection_type_position()?.ok_or_else(|| {
+                                FederationError::internal("unexpected: Element has a selection set with non-composite base type")
+                            })?,
+                            &NamedFragments::default(),
+                            &self.schema,
+                            RebaseErrorHandlingOption::ThrowError,
+                        )
+                    })
+                    .transpose()?;
+                let selection = Selection::from_element(element, selection_set)?;
                 let schema = self.schema.clone();
                 let parent_type_position = self.type_position.clone();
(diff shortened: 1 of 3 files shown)


apollo-federation/src/query_plan/operation.rs
```diff
@@ -4125,12 +4130,13 @@ impl InlineFragmentSelection {
...
                     );
-
+                    // Return `... on <rebased condition> { __typename @include(if: false) }`
+                    let rebased_casted_type = rebased_fragment.data().casted_type();
                     return Ok(Some(SelectionOrSet::Selection(
                         Selection::from_inline_fragment(
                             rebased_fragment,
                             SelectionSet::from_selection(
-                                parent_type.clone(),
+                                rebased_casted_type,
```
goto-bus-stop, line 4139: Hmmm! we should make `Selection::from_inline_fragment` should enforce this. The SelectionSet must always have the fragment's type condition as its parent type (or the fragment's parent type if it does not have a type condition).

dariuszkuc: I don't think this is necessary true as inline fragment without type condition is still valid i.e. `.... { x y }` is a valid (useless) fragment.

dariuszkuc: *this was one of the issues i hit with debugging `handle_requires` tests

goto-bus-stop: Yes, then the selection type must match the fragment's parent type.

duckki: inline fragment's sub-selection's type must be the same as its casted type. I'll add this assertion.

duckki:
I just added an assertion.
I also reviewed other `from_inline_fragment()` calls. I'll create a separate PR for that later.

duckki:
Oh no. I merged into the wrong branch. I was going to merge into the dev branch. 🤦

PR #5302, opened 2024-05-31
fuzz test the router VS a monolithic subgraph

Geal:
This repurposes the old router VS gateway fuzz test, that was sending the same query to both router and gateway, and checking that the responses were the same, to instead compare the result of a query sent to a router VS a monolithic subgraph.

The subgraph here merges the code from the usual accounts, products, reviews and inventory subgraphs (taken from the [starstuff repository](https://github.com/apollographql/starstuff/)).
This means that it can answer any query that would be handled by those subgraphs, but since it also has all the types and data available, it can be queried directly too.
The invariant we check here is that we should get the same result by sending the query to the subgraph directly, or through a router that will artificially cut up the query in multiple subgraph requests, according to the supergraph schema.

To execute it:
- start a router using the schema `fuzz/subgraph/supergraph.graphql`
- start the subgraph with `cargo run --release` in `fuzz/subgraph`. It will start a subgraph on port 4005
- start the fuzzer from the repo root with `cargo +nightly fuzz run router`

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


fuzz/Cargo.toml
```diff
@@ -12,8 +12,8 @@ cargo-fuzz = true
 [dependencies]
 libfuzzer-sys = "0.4"
 apollo-compiler.workspace = true
-apollo-smith.workspace = true
-apollo-parser.workspace = true
+apollo-parser =  { version = "0.7.6", git = "https://github.com/apollographql/apollo-rs", ref = "9ff692fcd2669eac5776b1b4a2290bbd03ad12ab" }
+apollo-smith = { version = "0.6.0-beta.1", git = "https://github.com/apollographql/apollo-rs", ref = "9ff692fcd2669eac5776b1b4a2290bbd03ad12ab" }
```
Geal, line 16: This PR can only be merged once a new version of apollo-smith has been released

fuzz/README.md
```diff
@@ -4,9 +4,11 @@
 
 ### Router
 
-This target is especially testing if we have differences between gateway 2 responses and router responses. As soon as it detects a difference it panics and write a report in `router_error.txt`.
-Before launching it you have to spawn the docker-compose for federation 2 with `docker-compose -f docker-compose-federation2.yml up` at the root of the router directory and also run a local router listening on port `4000`.
-And then run it with:
+This target is sending the same query to a backend exposing all the types in a schema, and to the router started with a supergraph schema that points to that same backend for each subgraph, and checks that both answers are the same.
+The invariant tested here is that however the schema is cut up, and however many subgraph requests it will take, the router will return the expected answer, as if it was just one API.
+Before launching it, you have to spawn the subgraph in `fuzz/suvbgraph` with `cargo run --release` and start a Router with the schema in `fuzz/subgrgraph/supergraph.graphql`. You can use whatever router configuration you need, allowing you to fuzz the router with specific features activated.
```
lrlna, line 9:
````suggestion
Before launching the fuzz target, you have to spawn a subgraph server and start a router.

Start the subgraph server with this command:
```bash
cd fuzz/subgraph && cargo run --release
```

Start the router with a supergraph in `fuzz` directory and any router configuration you need to test. Below, the command uses `router.yaml` configuration in the `fuzz` directory:
```bash
 cargo run -- --config fuzz/router.yaml --supergraph fuzz/subgraph/supergraph.graphql --dev --log=trace
```
````

File: fuzz/README.md [hunk 3]
```diff
@@ -4,9 +4,11 @@
...
+
+Run the fuzzer with this command:
 
 ```
 # Only works on Linux
```
lrlna, line 14:
```suggestion
# Only works on Linux and MacOS
```

fuzz/fuzz_targets/router.rs
```diff
@@ -38,21 +38,22 @@ fuzz_target!(|data: &[u8]| {
     debug!("{}", generated_operation);
```
lrlna, line 38: nit: This is hard to leave a "suggestion" on, but the line above this should probably say "operation" or "executable document" instead of "document".

lrlna (approved):
This is great!! I just tried running this, works great. imo let's get this in.

One ask I have is to update the `fuzz/router.yaml` so it can work with this fuzz target, in case anyone trying this needs _some_ sort of config to start with. The current one has an incorrect `listen` and some features that no longer exist. This one can work, for example:

```yaml
 supergraph:
  listen: 0.0.0.0:4000
  introspection: true
sandbox:
  enabled: true
homepage:
  enabled: false
traffic_shaping:
  router: 
    timeout: 300s
```

.changesets/maint_geal_everything_subgraph.md
```diff
@@ -0,0 +1,16 @@
+### fuzz test the router VS a monolithic subgraph ([PR #5302](https://github.com/apollographql/router/pull/5302))
```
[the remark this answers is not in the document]
Geal: users often ask about or testing strategies, so they would probably be interested

fuzz/fuzz_targets/router.rs
Geal: fixed

PR #5307, opened 2024-05-31
Adjust privacy of `JSONSelection` struct internals

benjamn:
Resolves Jira issue CNN-152.
Closes #5249.

Although keeping the `JSONSelection` structures immutable would have been (and still may be) ideal, we currently have a few algorithms that rely on incrementally constructing those structures (specifically populating `SubSelection` with new `NamedSelection` items), so we need to support a limited form of mutability.

To make that mutability safe(r), I decided to replace all direct access to struct fields with public methods, which enforce the invariant that `NamedSelection` items may only be _appended_ to a `&mut SubSelection`. This enforcement works by making the fields of the `SubSelection` struct (and other `JSONSelection`-related structs) `pub(super)`, which means they are accessible and constructible only within the `json_selection/` directory (which is important so the `ApplyTo` trait can still access the fields).

Since `SubSelection` is now guaranteed to be append-only, I was also able to expose a `SubSelection::index_of_named_selection` method, which returns an unforgeable and permanently valid `NamedSelectionIndex` wrapper struct, allowing safe and infallible access via `SubSelection::get_at_index` and `SubSelection::get_at_index_mut`. This indirection is helpful for placating the borrow checker, which refuses to accept any variation of a method like
```rust
SubSelection::get_mut(&mut self, i: usize) -> Option<&mut NamedSelection>
```
as a replacement for `subselection.selections.get_mut(i)`, due to the complexities of working with exclusive mutable references.

It might have been nice to restrict the constructability of enum variants as well (especially for `PathSelection`), but there's no way to do that in Rust without also removing the ability to match against the variants, which seems too onerous (and difficult to replicate with public functions).

### DIFF
--- a/apollo-federation/src/sources/connect/json_selection/parser.rs
+++ b/apollo-federation/src/sources/connect/json_selection/parser.rs
@@ -31,6 +31,13 @@ pub enum JSONSelection {
 }
 
 impl JSONSelection {
+    pub fn empty() -> Self {
+        JSONSelection::Named(SubSelection {
+            selections: vec![],
+            star: None,
+        })
+    }
+
     pub fn parse(input: &str) -> IResult<&str, Self> {
         alt((
             all_consuming(map(
@@ -325,8 +332,8 @@ impl PathSelection {
 
 #[derive(Debug, PartialEq, Clone, Serialize, Default)]
 pub struct SubSelection {
-    pub selections: Vec<NamedSelection>,
-    pub star: Option<StarSelection>,
+    pub(super) selections: Vec<NamedSelection>,
+    pub(super) star: Option<StarSelection>,
 }
 
 impl SubSelection {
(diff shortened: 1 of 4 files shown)


apollo-federation/src/sources/connect/json_selection/parser.rs
```diff
@@ -346,14 +353,69 @@ impl SubSelection {
...
+    pub fn get_at_index_mut(&mut self, index: &NamedSelectionIndex) -> &mut NamedSelection {
+        self.selections
+            .get_mut(index.pos)
+            .expect("NamedSelectionIndex out of bounds")
+    }
+}
+
+pub struct NamedSelectionIndex {
+    // Intentionally private so NamedSelectionIndex cannot be forged.
+    pos: usize,
```
benjamn, line 403: I think it's pretty cool that `get_at_index` and `get_at_index_mut` can return a `NamedSelection` reference without wrapping it in an `Option`, because of the guarantees provided by `NamedSelectionIndex` and the append-only invariant now enforced by `SubSelection`.

lennyburdette (approved):

PR #5309, opened 2024-05-31
feat(federation): implement handle_requires

dariuszkuc:
Implements `handle_requires` method (port of [`handleRequires`](https://github.com/apollographql/federation/blob/main/query-planner-js/src/buildPlan.ts#L4414)).

Resolves [**FED-25**](https://apollographql.atlassian.net/browse/FED-25)

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


dariuszkuc:
See https://github.com/apollographql/router/pull/5214 for code review comments.

lrlna (approved):

apollo-federation/src/query_graph/mod.rs
```diff
@@ -579,6 +587,49 @@ impl QueryGraph {
...
+            let key_value = metadata
+                .federation_spec_definition()
+                .key_directive_arguments(key)?;
+            let selection = parse_field_set(
+                subgraph_schema,
+                composite_type_position.type_name().clone(),
+                &key_value.fields,
+            )?;
+            if !external_metadata.selects_any_external_field(&selection)? {
+                return Ok(Some(selection));
```
lennyburdette, line 627: just curious: it looks like this returns the first local key — what if there are multiple keys, or the first one selects more fields than necessary?

dariuszkuc: Great question -> yep it just picks the first key. I don't think it really matters though -> given an edge between a node (`head`) and requires node (`tail`) we just pick any local key combination from `head` and use it as input for `tail`.

PR #5310, opened 2024-05-31
Allow use of a local persisted query safelist for use with offline licenses

### LINKED ISSUE #4587: Load persisted query list at boot up time from local file
I have an enterprise license with GraphOS and I want to use the [new offline entitlement](https://www.apollographql.com/docs/router/enterprise-features/#offline-enterprise-license) feature to start the Router without a connection to GraphOS. Because I do not want to configure my connection to GraphOS, I can not use the [Persisted Queries and safelisting features](https://www.apollographql.com/docs/router/configuration/persisted-queries) offline.

I would like to be able to also fetch the list of PQs from GraphOS through rover (or manually build them) but then at boot up time of the Router provide them via a start up flag like so

```shell
./router --supergraph schema.graphql --license graphos.jwt --persistedQueries pql.json
```

I could setup my connection to GraphOS or manually push operation into the Redis cluster but I usually don't have to manage the Redis data that Router saves

**Additional context**
This would also be helpful in a fallback scenario, similar to the [supergraph fallback issue](https://github.com/apollographql/router/issues/1282). You could maybe configure the connection to GraphOS but using Rover still fetch everything to have locally at build time for Router so I can still start up just in case my connection fails

lleadbet:
This PR allows users to define a local persisted query manifest as downloaded from GraphOS to be used in place of Uplink. 

Fixes #4587

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Notes**

This will only provide the ability to use in replacement of Uplink, not as a fallback.


apollo-router/src/services/layers/persisted_queries/manifest_poller.rs
```diff
@@ -207,7 +207,69 @@ impl PersistedQueryManifestPoller {
...
+                local_pq_list.clone()
+            );
+            let local_manifest =
+                std::fs::read_to_string(local_pq_list.clone()).map_err(|e| -> BoxError {
+                    format!("could not read local persisted query list file: {}", e).into()
+                })?;
+
+            let mut manifest = PersistedQueryManifest::new();
+
+            let manifest_file: SignedUrlChunk =
```
glasser, line 222: We should probably check the format and version fields?

docs/source/configuration/persisted-queries.mdx
```diff
@@ -109,6 +109,27 @@ To enable safelisting, you _must_ turn off [automatic persisted queries](./in-me
...
+  safelist:
+    enabled: true
+    local_safelist: ./persisted-query-manifest.json
+apq:
+  enabled: false
+```
+
+<Note>
+
+To enable safelisting, you _must_ turn off [automatic persisted queries](./in-memory-caching#caching-automatic-persisted-queries-apq) (APQs). APQs let clients [register arbitrary operations at runtime](/graphos/operations/persisted-queries/#differences-from-automatic-persisted-queries) while safelisting restricts operations to those that have been explicitly registered.
```
glasser, line 129: I'm not sure why this is included in this section.

docs/source/configuration/persisted-queries.mdx
```diff
@@ -109,6 +109,27 @@ To enable safelisting, you _must_ turn off [automatic persisted queries](./in-me
 
 </Note>
 
+
+#### `local_safelist`
+
+Adding `local_safelist` to your `safelist` configuration allows you to use a local persisted query manifest in place of the hosted Uplink version. This is useful when you have an offline Enterprise license and are unable to use Uplink.
```
glasser, line 115: In this implementation, there is no auto-reloading of the manifest file if it changes at runtime (I think?). Should this be called out explicitly in the docs as a limitation?

.changesets/feat_feat_local_safelist.md
```diff
@@ -0,0 +1,17 @@
...
+An example configuration would look like:
+
+```yml
+persisted_queries:
+  enabled: true
+  log_unknown: true
+  safelist:
+    enabled: true
+    require_id: false
+    local_safelist: ./persisted-query-manifest.json
```
glasser, line 14: I'm not sure it makes sense to nest this under `safelist`, which is specifically about the (important! popular!) sub-feature of restricting access. For better or for worse this feature can be used purely as a "you can run things by IDs too" feature and I think your implementation here works with that? I'd suggest moving this up one level to immediately under `persisted_queries` and calling it `local_manifest` or something. I recognize that the vast majority use case for this feature is safelisting (or at least eventually iterating towards safelisting) but for better or for worse, the feature we built does have these different modes of operation.

glasser (changes requested):

.changesets/feat_feat_local_safelist.md
lleadbet: fair! can fix, that's a relatively minor change as far as I can tell.

apollo-router/src/services/layers/persisted_queries/manifest_poller.rs
lleadbet: 🤦 i meant to do that. maybe worth abstracting it out into a validation function.

docs/source/configuration/persisted-queries.mdx
lleadbet: yeah, good call. can add.

docs/source/configuration/persisted-queries.mdx
lleadbet: All of the other safelist sections had it- was just following the same format. But given the first comment- it'll be removed in any case.

docs/source/configuration/persisted-queries.mdx
```diff
@@ -66,6 +66,16 @@ persisted_queries:
 
 If used with the [`safelist`](#safelist) option, the router logs unregistered and rejected operations. With [`safelist.required_id`](#require_id) off, the only rejected operations are unregistered ones. If [`safelist.required_id`](#require_id) is turned on, operations can be rejected even when registered because they use operation IDs rather than operation strings.
 
+#### `local_manifest`
+
+Adding `local_manifest` to your `safelist` configuration allows you to use a local persisted query manifest in place of the hosted Uplink version. This is useful when you have an offline Enterprise license and are unable to use Uplink. This configuration option cause the router to not reload the manifest from the file system, so you will need to restart the router to apply changes.
```
Geal, line 71: Needing to restart is fine for now, but I think it should support reloading in the future.
But I'd be more at ease if it went through the same mechanism as schema and config reload, because filesystem watching is full of fun edge cases

Geal (approved):

apollo-router/src/services/layers/persisted_queries/manifest_poller.rs
```diff
@@ -207,7 +207,76 @@ impl PersistedQueryManifestPoller {
...
     /// and the [`PersistedQueryManifest`] has been fully populated.
     pub(crate) async fn new(config: Configuration) -> Result<Self, BoxError> {
-        if let Some(uplink_config) = config.uplink.as_ref() {
+        if let Some(local_pq_list) = config.persisted_queries.local_manifest {
+            tracing::info!(
+                "Loading persisted query list from local file: {}",
+                local_pq_list.clone()
+            );
+            let local_manifest =
+                std::fs::read_to_string(local_pq_list.clone()).map_err(|e| -> BoxError {
```
bnjjj, line 216: As we're in an async function I would feel more comfortable if we used this [async version](https://docs.rs/tokio_wasi/latest/tokio/fs/fn.read_to_string.html) instead

lleadbet: Updated to use the async variant, although is using await in it; let me know if there's a different approach preferred here.

docs/source/configuration/persisted-queries.mdx
lleadbet: Agree- good future improvement.

docs/source/configuration/persisted-queries.mdx
```diff
@@ -66,6 +66,18 @@ persisted_queries:
 
 If used with the [`safelist`](#safelist) option, the router logs unregistered and rejected operations. With [`safelist.required_id`](#require_id) off, the only rejected operations are unregistered ones. If [`safelist.required_id`](#require_id) is turned on, operations can be rejected even when registered because they use operation IDs rather than operation strings.
 
+#### `experimental_local_manifest`
+
+<ExperimentalFeature />
+
+Adding `experimental_local_manifest` to your `safelist` configuration allows you to use a local persisted query manifest in place of the hosted Uplink version. This is useful when you have an offline Enterprise license and are unable to use Uplink. This configuration option cause the router to not reload the manifest from the file system, so you will need to restart the router to apply changes.
```
glasser, line 73: Should we link to an example of what such a file looks like?

.changesets/feat_feat_local_safelist.md
```diff
@@ -0,0 +1,17 @@
+### Allow use of a local persisted query manifest for use with offline licenses ([Issue #4587](https://github.com/apollographql/router/issues/4587))
```
glasser, line 1: filename `feat_feat_`?

glasser (approved):
This looks good from a PQ perspective. Probably wants an approval from a Router expert.

I'd like to call out a sort of embarrassing fact: the apollo-persisted-query-manifest/1 file format passed to `rover` and the file format downloaded to Routers from Uplink are slightly different! The former needs to have operation names and types, whereas the latter doesn't contain them (Router doesn't use them at all so we left them out for space optimization reasons). So it's a little weird that we have the same format/version specifier for them and that you can provide the former directly to Router here. Ah well, my bad for not changing the format name. It's close enough.

.changesets/feat_feat_local_safelist.md
lleadbet: My branch was titled `feat/local-safelist` and the title is based off the branch name, so, yeah. I can prune it though, good catch.

docs/source/configuration/persisted-queries.mdx
lleadbet: We could, I suppose I would be curious on how we do that/if there's a better resource that we could link to instead for this. @apollographql/docs - thoughts?

Meschreiber: Hmm, AFAIK we don't have a standard protocol for this. I could see us linking to a repo of maintained examples/solutions if something like that exists.

In this case, though the example isn't particularly fleshed out, what about linking to the `apollo-router/tests/fixtures/persisted-queries-manifest.json` that's part of this PR just so readers can see the expected format?

lleadbet: That may be sane- I suppose my only comment is that they shouldn't be crafting it at all, but rather using this here: 
image

Perhaps it's just using a screenshot like that?

Meschreiber: > That may be sane- I suppose my only comment is that they shouldn't be crafting it at all, but rather using this here:

Oh, that's very helpful information to callout! I think that deserves it's own explanation. Will add a suggestion for that below the current example config momentarily.
```

File: docs/source/configuration/persisted-queries.mdx [hunk 9]
```diff
@@ -66,6 +66,18 @@ persisted_queries:
...
+#### `experimental_local_manifest`
+
+<ExperimentalFeature />
+
+Adding `experimental_local_manifest` to your `persisted-queries` configuration lets you use a local persisted query manifest instead of the hosted Uplink version. This is helpful when you're using an offline Enterprise license and can't use Uplink. With the  `experimental_local_manifest`, the router doesn't reload the manifest from the file system, so you need to restart the router to apply changes.
+
+```yaml title="router.yaml"
+persisted_queries:
+  enabled: true
+  experimental_local_manifest: ./persisted-query-manifest.json
```
Meschreiber, line 78:
```suggestion
  experimental_local_manifest: ./path/to/persisted-query-manifest.json
```

docs/source/configuration/persisted-queries.mdx
```diff
@@ -66,6 +66,18 @@ persisted_queries:
...
+```
+
```
Meschreiber, line 80:
```suggestion

You can download a version of your manifest to use locally from [GraphOS Studio](https://studio.apollographql.com/?referrer=docs-content). Open the PQL page for a graph by clicking the **Go to persisted query lists** to the left of the graph's name. Then, click the ••• menu under the **Actions** column to download a PQL's manifest as a JSON file. Save this file locally and update your `experimental_local_manifest` configuration with the path the file.
```

Meschreiber: (Trying to avoid more screenshots.)

docs/source/configuration/persisted-queries.mdx
glasser: One note about that UI control. On the GraphOS -> Router side, we divide a PQL up into "chunks" (each having the format of a manifest JSON) that Router can download separately to limit the size of each individual download (and allow it to only refresh the part of the list that changed).  (Chunk division is not related to the order or structure of how the operations were uploaded.) For better or for worse I believe that the download button in the UI tries to download all of the chunks (in one GraphQL operation, so hopefully that scales well enough) and then concatenates them (without a separator, I think) into the file.

In the common case this is just one JSON object but as folks scale it will eventually become two JSON objects in the same file and so this will fail.

Thoughts on what we should do here? We could change the UI to do something slightly different like parse and join the files or at least insert newlines between them (though at the end of the day I'm not sure how great an idea it is to try to offer in the website a single-click one-file download of something we've explicitly architected as a sequence of bounded-size files).  Or you could change your code to understand that the file might be a "more than one concatenated JSON object" file.  But if we recommend this button and don't make another change, this will start failing for people eventually (either your code won't be able to parse the file, or it'll just ignore part of the file after the first object or something.

lleadbet: That's... interesting. I'd prefer we just concat client-side and send a larger JSON, but if the format is changing- that's another concern altogether since there's not a good way to handle this as far as I know in Rust beyond checking for one format, then another for parsing. 

Anyways, happy to do whatever, just probably should figure out the right path before we dig our own grave.

lleadbet: From our conversation yesterday, I've pushed up a change to make this take an array of manifests vs. just one- should allow us to account for the chunking situation as well as single-file cases.

docs/source/configuration/persisted-queries.mdx
```diff
@@ -66,6 +66,21 @@ persisted_queries:
 
 If used with the [`safelist`](#safelist) option, the router logs unregistered and rejected operations. With [`safelist.required_id`](#require_id) off, the only rejected operations are unregistered ones. If [`safelist.required_id`](#require_id) is turned on, operations can be rejected even when registered because they use operation IDs rather than operation strings.
 
+#### `experimental_local_manifests`
+
+<ExperimentalFeature />
+
+Adding `experimental_local_manifests` to your `persisted-queries` configuration lets you use local persisted query manifests instead of the hosted Uplink version. This is helpful when you're using an offline Enterprise license and can't use Uplink. With the  `experimental_local_manifests`, the router doesn't reload the manifest from the file system, so you need to restart the router to apply changes.
```
abernix, line 73: `s/-/_/`.

```suggestion
Adding `experimental_local_manifests` to your `persisted_queries` configuration lets you use local persisted query manifests instead of the hosted Uplink version. This is helpful when you're using an offline Enterprise license and can't use Uplink. With the `experimental_local_manifests` option, the router doesn't reload the manifest from the file system, so you need to restart the router to apply changes.
```

bnjjj (approved):

.changesets/feat_local_safelist.md
```diff
@@ -0,0 +1,18 @@
...
+```yml
+persisted_queries:
+  enabled: true
+  log_unknown: true
+  experimental_local_manifests: 
+    - ./persisted-query-manifest.json
+  safelist:
+    enabled: true
+    require_id: false
+```
```
abernix, line 16: I know it's in the docs — and also that this PR is now already merged — but I might also change this changelog to lead with the "doesn't hot reload" gotcha.  Basically, this part, again:

> With the `experimental_local_manifests`, the router doesn't reload the manifest from the file system, so you need to restart the router to apply changes.

PR #5315, opened 2024-06-01
Fixed a few normalize issues (take #2)

duckki:
(This is the same as PR #5301. I merged it into the wrong branch. So, I'm trying again.)

Added `normalize` call in `normalize_operation`.

Fixed a bug in NamedFragments::reduce
- normalize selection sets after retain_fragments call

Fixed InlineFragmentSelection::normalize
- Normalize sub-selections with the correct type

Sorry for mixing a few different changes in this PR. Unfortunately, those issues are intertwined in tests. Gathering them in one PR to minimize spurious test changes.

This PR will improve a few tests failing in dev branch.

---

**Checklist**

- Tests added and passing[^3]
    - [X] Unit Tests
    - [X] Integration Tests
    - [ ] Manual Tests

### DIFF
--- a/apollo-federation/src/query_plan/operation.rs
+++ b/apollo-federation/src/query_plan/operation.rs
@@ -2877,10 +2877,18 @@ impl SelectionSet {
                 // in-place, we eagerly construct the selection that needs to be rebased on the target
                 // schema.
                 let element = OpPathElement::clone(ele);
-                let selection = Selection::from_element(
-                    element,
-                    selection_set.map(|set| SelectionSet::clone(set)),
-                )?;
+                let selection_set = selection_set.map(|selection_set| {
+                        selection_set.rebase_on(
+                            &element.sub_selection_type_position()?.ok_or_else(|| {
+                                FederationError::internal("unexpected: Element has a selection set with non-composite base type")
+                            })?,
+                            &NamedFragments::default(),
+                            &self.schema,
+                            RebaseErrorHandlingOption::ThrowError,
+                        )
+                    })
+                    .transpose()?;
+                let selection = Selection::from_element(element, selection_set)?;
                 let schema = self.schema.clone();
                 let parent_type_position = self.type_position.clone();
                 // TODO move the rebasing to add_selection/merge_into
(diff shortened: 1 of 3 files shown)


apollo-federation/src/query_plan/operation.rs
```diff
@@ -701,6 +701,10 @@ impl Selection {
         inline_fragment: InlineFragment,
         sub_selections: SelectionSet,
     ) -> Self {
+        assert_eq!(
```
goto-bus-stop, line 704: we should maybe keep this as a debug_assert so it doesn't bring down any routers.
```suggestion
        debug_assert_eq!(
```
That or making the function return a Result

apollo-federation/src/query_plan/operation.rs
```diff
@@ -2873,10 +2877,18 @@ impl SelectionSet {
                 // in-place, we eagerly construct the selection that needs to be rebased on the target
                 // schema.
                 let element = OpPathElement::clone(ele);
-                let selection = Selection::from_element(
-                    element,
-                    selection_set.map(|set| SelectionSet::clone(set)),
-                )?;
+                let selection_set = selection_set.map(|selection_set| {
```
goto-bus-stop, line 2880: i'm really surprised linting accepts this 😅  the indentation looks so weird. does it work like this?
```suggestion
                let selection_set = selection_set
                    .map(|selection_set| {
```

apollo-federation/src/query_plan/operation.rs
```diff
@@ -2873,10 +2877,18 @@ impl SelectionSet {
...
+                        selection_set.rebase_on(
+                            &element.sub_selection_type_position()?.ok_or_else(|| {
+                                FederationError::internal("unexpected: Element has a selection set with non-composite base type")
+                            })?,
+                            &NamedFragments::default(),
```
goto-bus-stop, line 2885: Are we sure this is right? If `selection_set` contains references to named fragments, they need to exist in this map in order to rebase correctly. I'm not really sure what we *should* provide here. If we're not sure we should at least add a comment, and if we are sure we should mention that too i think.

goto-bus-stop (approved):
A few comments but it looks fine to me :)

apollo-federation/src/query_plan/operation.rs
duckki: Good point!

PR #5328, opened 2024-06-03
FED-247: changed the type of QueryPlanCost to f64

duckki:
- The cost function has an arbitrary precision and can overflow i64.

Fix: https://apollographql.atlassian.net/browse/FED-247

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

It does not change tests right away. But, FED-55 is blocked by this change due to a test failure.

**Notes**


apollo-federation/src/query_plan/mod.rs
```diff
@@ -18,7 +18,7 @@ mod optimize;
 pub mod query_planner;
 pub(crate) mod query_planning_traversal;
 
-pub type QueryPlanCost = i64;
+pub type QueryPlanCost = f64;
```
lrlna, line 21: (by the way, rust also has an i/u128)

duckki: The calculation is really meant to be floating point. One can craft a query that can blow out u128 by making a long `require` chain. 10-require chain breaks i64. 20 chain will surely break i128. Though it's unlikely in practice, it's a security issue.

duckki: Alternatively, we can add a preventive logic. But, f64 is an easy solution and compatible with JS version.

lrlna (approved):

PR #5329, opened 2024-06-03
Require connect directive on all query fields

dylan-apollo:


dylan-apollo:
This is part of #5355

PR #5331, opened 2024-06-04
Add support for experimental extended reference reporting

bonnici:
Adds a new experimental config option `experimental_apollo_metrics_reference_mode: extended` which when enabled will include input object and enum value references in usage reports sent to Studio.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/services/execution/service.rs
```diff
@@ -252,6 +252,16 @@ impl ExecutionService {
...
     ) -> Option<Response> {
+        /*
+        println!("\nin process_graphql_response");
+        println!("query: ${:?}", query);
+        println!("operation_name: ${:?}", operation_name);
+        println!("variables: ${:?}", variables);
+        println!("is_deferred: ${:?}", is_deferred);
+        println!("response: ${:?}", response);
+        println!("nullified_paths: ${:?}", nullified_paths);
+        */
```
bnjjj, line 263: to delete

apollo-router/src/services/layers/query_analysis.rs
```diff
@@ -191,7 +195,38 @@ impl QueryAnalysisLayer {
...
+                                    matches!(
+                                        conf.apollo.experimental_apollo_metrics_reference_mode,
+                                        ApolloMetricsReferenceMode::Extended
+                                    )
+                                }
+                                _ => false,
+                            }
+                        }
+                        None => false,
+                    };
```
bnjjj, line 216: This will create a big perf regression I think. Because we will have to deserialize the config for each request and it's something that can be done by using a downcast and I would suggest to not do this in a hot path like this one. It should be done only when the config is updated and at the creation of plugins/layers.

apollo-router/src/spec/query.rs
```diff
@@ -130,6 +130,15 @@ impl Query {
...
 
         let original_operation = self.operation(operation_name);
+
+        /*
+        println!("\nin format_response");
+        println!("operation_name: ${:?}", operation_name);
+        println!("variables: ${:?}", variables);
+        println!("response: ${:?}", response);
+        println!("original_operation: ${:?}", original_operation);
+        */
```
bnjjj, line 140: to delete

apollo-router/src/spec/query.rs
```diff
@@ -524,6 +533,7 @@ impl Query {
                     Some(ExtendedType::Enum(enum_type)) => {
                         return match input.as_str() {
                             Some(s) => {
+                                // here for enum? or duplicate this logic
```
bnjjj, line 536: ?

apollo-router/src/plugins/telemetry/mod.rs
```diff
@@ -1236,6 +1237,8 @@ impl Telemetry {
         let operation_kind: OperationKind =
             ctx.get(OPERATION_KIND).ok().flatten().unwrap_or_default();
 
+        // here?
```
bnjjj, line 1240: ?

apollo-router/src/apollo_studio_interop/mod.rs
```diff
@@ -301,6 +453,209 @@ impl UsageReportingGenerator<'_> {
...
+        type_name: String,
+        field_name: String,
+        is_referenced: bool,
+        is_null_reference: bool,
+    ) {
+        match self
+            .input_field_references
+            .entry(type_name)
+            .or_default()
+            .entry(field_name)
```
bnjjj, line 493: if we're in a hot path here. We already spotted that using the entry api might be less performant

apollo-router/src/apollo_studio_interop/mod.rs
```diff
@@ -19,12 +21,131 @@ use apollo_compiler::executable::InlineFragment;
...
+        }
+
+        // Merge enum references
+        for (enum_name, enum_values) in other_stats.referenced_enums.iter() {
+            for enum_value in enum_values.iter() {
+                match self
+                    .referenced_enums
+                    .entry(enum_name.to_string())
+                    .or_default()
+                    .entry(enum_value.to_string())
```
bnjjj, line 136: if we're in a hot path here. We already spotted that using the entry api might be less performant

apollo-router/src/apollo_studio_interop/mod.rs
```diff
@@ -19,12 +21,131 @@ use apollo_compiler::executable::InlineFragment;
...
+impl AddAssign<ExtendedReferenceStats> for AggregatedExtendedReferenceStats {
+    fn add_assign(&mut self, other_stats: ExtendedReferenceStats) {
+        // Merge input object references
+        for (type_name, type_stats) in other_stats.referenced_input_fields.iter() {
+            for (field_name, field_stats) in type_stats.iter() {
+                match self
+                    .referenced_input_fields
+                    .entry(type_name.to_string())
+                    .or_default()
+                    .entry(field_name.to_string())
```
bnjjj, line 102: if we're in a hot path here. We already spotted that using the entry api might be less performant

bnjjj (changes requested):

apollo-router/src/services/execution/service.rs
bonnici: Sorry, I should have re-read my PR but i was making changes late last night. I'll clear these up.

apollo-router/src/services/layers/query_analysis.rs
bonnici: I'll figure out a way to pre-calculate this on startup or something like that.

apollo-router/src/apollo_studio_interop/mod.rs
bonnici: I'm not sure if you meant that I shouldn't use the entry API at all or if I should only avoid using it unnecessarily in an inner loop, so I've changed it to never use the entry API. Let me know if I should change it so we still use the entry API but only once in the outer loop.

apollo-router/src/services/layers/query_analysis.rs
bonnici: I've updated it to load the config on startup and tested that it works with hot-reload, but I'm not sure what you meant by using a downcast.

apollo-router/src/apollo_studio_interop/mod.rs
```diff
@@ -92,56 +92,66 @@ impl AggregatedExtendedReferenceStats {
...
         }
 
         // Merge enum references
         for (enum_name, enum_values) in other_stats.referenced_enums.iter() {
+            let enum_name_stats = match self.referenced_enums.get_mut(enum_name) {
+                Some(existing_stats) => existing_stats,
+                None => {
+                    self.referenced_enums
+                        .insert(enum_name.to_string(), HashMap::new());
+                    self.referenced_enums.get_mut(enum_name).unwrap()
```
bnjjj, line 144: Could we change the `unwrap()` with an `expect()` please ? Could be done in a follow up PR

apollo-router/src/apollo_studio_interop/mod.rs
```diff
@@ -473,10 +482,17 @@ impl UsageGenerator<'_> {
...
-            .entry(enum_name)
-            .or_default()
-            .insert(enum_value.to_string());
+        // Not using entry API here due to performance impact
+        let enum_name_stats = match self.enums_by_name.get_mut(&enum_name) {
+            Some(existing_stats) => existing_stats,
+            None => {
+                self.enums_by_name
+                    .insert(enum_name.to_string(), HashSet::new());
+                self.enums_by_name.get_mut(&enum_name).unwrap()
```
bnjjj, line 491: Could we change the `unwrap()` with an `expect()` please ? Could be done in a follow up PR

apollo-router/src/apollo_studio_interop/mod.rs
```diff
@@ -486,28 +502,33 @@ impl UsageGenerator<'_> {
...
-                // The input object field stats are additive in that it represents whether the field was referenced/null/undefined
-                // at least once within this operation.
-                let stats: &mut InputObjectFieldStats = entry.get_mut();
+        // Not using entry API here due to performance impact
+        let type_name_stats = match self.input_field_references.get_mut(&type_name) {
+            Some(existing_stats) => existing_stats,
+            None => {
+                self.input_field_references
+                    .insert(type_name.to_string(), HashMap::new());
+                self.input_field_references.get_mut(&type_name).unwrap()
```
bnjjj, line 511: Could we change the `unwrap()` with an `expect()` please ? Could be done in a follow up PR

apollo-router/src/plugins/telemetry/config.rs
```diff
@@ -682,6 +683,37 @@ impl Conf {
             },
         )
     }
+
+    pub(crate) fn metrics_reference_mode(
+        configuration: &Configuration,
+    ) -> ApolloMetricsReferenceMode {
+        match configuration.apollo_plugins.plugins.get("telemetry") {
+            Some(telemetry_config) => {
+                match serde_json::from_value::<Conf>(telemetry_config.clone()) {
```
bnjjj, line 692: I was talkin about this https://github.com/apollographql/router/blob/dev/apollo-router/src/router_factory.rs#L201:L201

apollo-router/src/plugins/telemetry/config.rs
```diff
@@ -682,6 +683,37 @@ impl Conf {
...
+                }
+            }
+            _ => ApolloMetricsReferenceMode::default(),
+        }
+    }
+
+    pub(crate) fn signature_normalization_algorithm(
+        configuration: &Configuration,
+    ) -> ApolloSignatureNormalizationAlgorithm {
+        match configuration.apollo_plugins.plugins.get("telemetry") {
```
bnjjj, line 704: Same https://github.com/apollographql/router/blob/dev/apollo-router/src/router_factory.rs#L201:L201

apollo-router/tests/snapshots/apollo_otel_traces__client_version.snap.new
bnjjj, line 1: It's a new snapshot file, cargo insta should throw an error if we still have this one.

bnjjj (commented):
All these changes can be in a followup PR, on the other hand I'm not feeling comfortable to merge it if I'm the only one reviewer as I'm not the most experienced one about that part of the codebase

apollo-router/src/plugins/telemetry/config.rs
bonnici: Is this still needed now that I'm only doing the initialisation once on startup and when the config is updated? I'm having trouble figuring out how to get a Box'd telemetry plugin without using the factory to create one like in that example. If it's still needed, could you let me know how I could use it or point me to some docs?

apollo-router/src/apollo_studio_interop/mod.rs
bonnici: I've updated these.

apollo-router/src/services/layers/query_analysis.rs
```diff
@@ -191,7 +199,23 @@ impl QueryAnalysisLayer {
                 request
                     .context
                     .extensions()
-                    .with_lock(|mut lock| lock.insert::<ParsedDocument>(doc));
+                    .with_lock(|mut lock| lock.insert::<ParsedDocument>(doc.clone()));
```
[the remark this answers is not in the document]
bonnici: I've refactored this so that we only do a single lock.

apollo-router/src/uplink/license_enforcement.rs
```diff
@@ -394,6 +394,11 @@ impl LicenseEnforcementReport {
                 .path("$.preview_demand_control")
                 .name("Demand control plugin")
                 .build(),
+            ConfigurationRestriction::builder()
```
BrynCooke, line 397: This needs a related change in https://github.com/apollographql/router/blob/dev/apollo-router/src/uplink/testdata/restricted.router.yaml

BrynCooke (changes requested):

apollo-router/src/uplink/license_enforcement.rs
bonnici: Thanks, I've updated the test and snapshot.

apollo-router/src/plugins/telemetry/config.rs
bnjjj: Ah yes, I thought that in Configuration plugins were already instantiated but unfortunately I was wrong. As it's not longer in the hot path let's keep this code.

bnjjj (approved):

bonnici:
@BrynCooke can you take a look at let me know if any more changes are needed here?

BrynCooke (approved):

.changesets/exp_pencil_hunk_tornado_backyard.md
```diff
@@ -0,0 +1,7 @@
+### Extended references in Apollo usage reports ([Issue #ROUTER-360](https://apollographql.atlassian.net/browse/ROUTER-360))
```
abernix, line 1: Subtle and I've fixed via https://github.com/apollographql/router/pull/5610 but for future review awareness, let's not include the Jira links — they're not useful for the community.  Instead, use our PR numbers in the absence of the GitHub Issue.  cc @bonnici @BrynCooke @bnjjj

PR #5332, opened 2024-06-04
Fix bug where recursive query planning ignored conditionals

sachindshinde:
This PR fixes a bug where recursive query planning would ignore conditionals (`@skip`/`@include`).

### DIFF
--- a/apollo-federation/src/query_plan/query_planning_traversal.rs
+++ b/apollo-federation/src/query_plan/query_planning_traversal.rs
@@ -876,7 +876,6 @@ impl<'a> QueryPlanningTraversal<'a> {
     fn resolve_condition_plan(
         &self,
         edge: EdgeIndex,
-        // PORT_NOTE: The following parameters are not currently used.
         context: &OpGraphPathContext,
         excluded_destinations: &ExcludedDestinations,
         excluded_conditions: &ExcludedConditions,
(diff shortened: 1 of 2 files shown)


apollo-federation/src/query_plan/query_planning_traversal.rs
```diff
@@ -877,7 +877,7 @@ impl<'a> QueryPlanningTraversal<'a> {
         &self,
         edge: EdgeIndex,
         // PORT_NOTE: The following parameters are not currently used.
```
goto-bus-stop, line 879: looks like this is no longer true?
```suggestion
```

apollo-federation/tests/query_plan/build_query_plan_tests/requires/include_skip.rs
```diff
@@ -67,7 +67,6 @@ fn it_handles_a_simple_at_requires_triggered_within_a_conditional() {
 }
 
 #[test]
-#[should_panic(expected = "snapshot assertion")]
 // TODO: investigate this failure - context is not passed, expected [@include(if: $test)] but was []
```
goto-bus-stop, line 70:
```suggestion
```

lrlna (approved):
hmmmmm i really wonder why those conditionals were never in context, I guess we will neverrrrr know 😀

goto-bus-stop (approved):

PR #5335, opened 2024-06-04
apollo-federation: accept supergraphs with join@v0.4

lrlna:
We didn't accept v0.4 join in query planning validation, meaning supergraphs using v0.4 could not be planned. This version of join spec doesn't have an affect on query planning, but is a composition concern, so accepting this version doesn't have an impact on the created plans.

### DIFF
--- a/apollo-federation/src/link/join_spec_definition.rs
+++ b/apollo-federation/src/link/join_spec_definition.rs
@@ -337,19 +337,12 @@ lazy_static! {
             Some(Version { major: 2, minor: 0 }),
         ));
         definitions.add(JoinSpecDefinition::new(
-            Version { major: 0, minor: 5 },
-            Some(Version { major: 2, minor: 8 }),
-        ));
-        definitions
-    };
-
-    /// Versions supported for purpose other than query planning.
-    /// TODO: remove once v0.4 is properly implemented in query planning
-    pub(crate) static ref NON_QUERY_PLANNING_JOIN_VERSIONS: SpecDefinitions<JoinSpecDefinition> = {
-        let mut definitions = JOIN_VERSIONS.clone();
-        definitions.add(JoinSpecDefinition::new(
             Version { major: 0, minor: 4 },
             Some(Version { major: 2, minor: 7 }),
+        ));
+        definitions.add(JoinSpecDefinition::new(
+            Version { major: 0, minor: 5 },
+            Some(Version { major: 2, minor: 8 }),
         ));
         definitions
     };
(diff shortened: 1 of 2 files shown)


apollo-federation/src/link/join_spec_definition.rs
```diff
@@ -336,21 +336,14 @@ lazy_static! {
...
-            Version { major: 0, minor: 5 },
-            Some(Version { major: 2, minor: 8 }),
-        ));
-        definitions
-    };
-
-    /// Versions supported for purpose other than query planning.
-    /// TODO: remove once v0.4 is properly implemented in query planning
-    pub(crate) static ref NON_QUERY_PLANNING_JOIN_VERSIONS: SpecDefinitions<JoinSpecDefinition> = {
-        let mut definitions = JOIN_VERSIONS.clone();
```
lrlna, line 349: I _think_ this shouldn't be necessary if we allow v0.4 to be planned.

apollo-router/src/router_factory.rs
```diff
@@ -237,6 +241,14 @@ impl YamlRouterFactory {
...
             )
             .await?;
+
+        // Don't let the router start in experimental_query_planner_mode and
+        // unimplemented Rust QP features.
+        can_use_with_experimental_query_planner(
+            configuration.clone(),
+            supergraph_creator.schema(),
+        )?;
+
```
lrlna, line 251: I put this in the YamlRouterFactory, it can perhaps also go right on the QueryAnalysisLayer impl. But that seems to be carried around everywhere, so I opted in for this. Let me know if there is a better spot!

apollo-router/src/router_factory.rs
```diff
@@ -724,6 +736,105 @@ fn inject_schema_id(schema_id: Option<&str>, configuration: &mut Value) {
...
+                        }
+                        return false;
+                    }
+                    false
+                });
+            if has_progressive_overrides {
+                return Err(ConfigurationError::InvalidConfiguration {
+                    message: "`experimental_query_planner_mode` cannot be used with progressive overrides",
+                    error: "remove uses of progressive overrides to try the experimental_query_planner_mode in `both` or `new`, otherwise switch back to `legacy`.".to_string(),
+                });
```
lrlna, line 799: Depending on how you look at it, it might not be a `InvalidConfiguration` error. I didn't want to add another variant to the _public_ `ConfigurationError` enum since this is really temporary. Perhaps there is a better suited error type we've got that I can't think of?

dariuszkuc (approved):

PR #5340, opened 2024-06-04
initial stab at making the router prewarm on pqs

### LINKED ISSUE #5334: Warm up query plan cache at boot using persisted queries

With the [current warmup implementation](https://github.com/apollographql/router/blob/dev/apollo-router/src/router_factory.rs#L246), the router will only warm up the query plan cache when it has been reloaded through a schema update (e.g. Uplink) or config update (when using `--hr` or `--dev`). 

Persisted queries affords the opportunity to use known queries to prepopulate the cache prior to serving any traffic on initial boot, and should be used as such in addition to the updating logic. 
 
Running the router in prod with `--hr` and sending a `SIGHUP` command to force a refresh.

lleadbet:
As #5334 outlined, the existing query plan warmup mechanism only warmed up the cache upon reload of either the schema or (if using dev mode/hot-reload) the config. 

This PR allows the warmup function to occur on startup as well, enabling persisted queries to be used to warm up prior to serving traffic, helping reduce latency introduced by the query planner. 

Fixes #5334 

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

The existing setup had no tests, so unsure how/where tests should be written for this.

[ROUTER-346]: https://apollographql.atlassian.net/browse/ROUTER-346?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apollo-router/src/router_factory.rs
```diff
@@ -258,6 +258,17 @@ impl YamlRouterFactory {
...
+            supergraph_creator
+                .prewarm_query_planner(
+                    &query_analysis_layer,
+                    &persisted_query_layer,
+                    configuration
+                        .supergraph
+                        .query_planning
+                        .experimental_reuse_query_plans,
+                )
+                .await;
```
Geal, line 271: maybe something like this would be simpler:

```rust
let previous_cache = previous_router.map(|router| previous_router.previous_cache());
supergraph_creator
                .warm_up_query_planner(
                    &query_analysis_layer,
                    &persisted_query_layer,
                    previous_cache,
                    configuration.supergraph.query_planning.warmed_up_queries,
                    configuration
                        .supergraph
                        .query_planning
                        .experimental_reuse_query_plans,
                )
                .await;
```

While having `warm_up_query_planner` take an `Option<InMemoryCache>` as argument

lleadbet: Yeah, I had considered that so that sounds good; felt like prewarm was just a little clearer for now, but can use the same path for both cases.

lleadbet:
> If we were to mark this feature _in particular_ as experimental for the initial drop, how would we do that? Would we do it with configuration? In other words, how do we make it so we DO NOT use pre-warming from the PQ manifest unless it's explicitly enabled?

We could just add a config option under the query planning chunk if that's preferable. 

I believe we've stated with others before that this is expected behavior, but if we'd like to keep it safe, we can certainly make it opt-in.

abernix:
Ok, understood.  It was really more about making it safe and validating it with a few customers first, but if we can claim it's safe, I'm comfortable.

Geal:
I'd be ok with an option here too

lleadbet:
Updated to use a new setting @Geal / @abernix. Just one note: I stuck it under the PQ settings since it seemed related to that instead, as well as it only applying to warming up on *startup only*, as we don't want to break existing setups.

Geal (approved):

docs/source/configuration/persisted-queries.mdx
```diff
@@ -66,6 +66,18 @@ persisted_queries:
...
+#### `experimental_prewarm_query_plan_cache`
+
+<ExperimentalFeature />
+
+Adding `experimental_prewarm_query_plan_cache: true` to `persisted_queries` configures the router to prewarm the query plan cache on startup using the persisted query list. All subsequent requests benefit from the warmed cache, reducing latency and enhancing performance.
+
+```yaml title="router.yaml"
+persisted_queries:
+  enabled: true
+  experimental_prewarm_query_plan_cache: true
```
bnjjj, line 78:
```suggestion
  experimental_prewarm_query_plan_cache: true # default: false
```

bnjjj: Just a nit, otherwise LGTM

bnjjj (approved):

BrynCooke (approved):

PR #5341, opened 2024-06-04
Move QP "both" mode comparison to a background thread with bounded queue

SimonSapin:
Before this PR on query plan cache miss we would, sequentially:
* Run the Rust QP
* Run the JS QP
* Compare results
* Return the result from JS

After this PR, in the main code path:
* Run the JS QP
* Spawn a background job with that result (shared in `Arc`) and the input document
* Return the result from JS

Meanwhile, the Rust QP and the result comparison runs in the background on a dedicated thread pool. Jobs are dropped if the bounded queue is full. To limit resource consumption at the moment the "pool" is a single thread, and the queue size is 10. Both of these parameters are `const` items in the code (not exposed in Router YAML config at this time).

To try it out, create a `router.yaml` file:

```yaml
```

Then run:

```
cargo run -- --dev -c router.yaml -s examples/graphql/supergraph-fed2.graphql
```

And in a separate terminal:

```
curl --request POST \
  --header 'content-type: application/json' \
  --url 'http://127.0.0.1:4000/' \
  --data '{"query":"{topProducts {name}}"}'
```
Should return a response of:
```
{"data":{"topProducts":[{"name":"Table"},{"name":"Couch"},{"name":"Chair"}]}}
```

Expected in the output of the first terminal (router logs):

```
WARN  JS v.s. Rust query plan mismatch
```

We can add to the Router command line to enable DEBUG level logs and show the diff of formatted query plans:

```
--log=apollo_router::query_planner::bridge_query_planner=debug
```
```
WARN  JS v.s. Rust query plan mismatch
DEBUG  Diff of formatted plans:
 QueryPlan {
   Fetch(service: "products") {
     {
       topProducts {
         name
       }
     }
   },
 }
```
No line has a `+` or `-` prefix, so the formatted strings are identical. This is still considered a mismatch because the comparison function is `derive(PartialEq)`. We can increase the log level to TRACE to see the `derive(Debug)` output, which is likely too verbose for any non-trivial query plan:

```
--log=apollo_router::query_planner::bridge_query_planner=trace
```
```
WARN  JS v.s. Rust query plan mismatch
2024-06-04T20:48:34.805692Z WARN  JS v.s. Rust query plan mismatch
2024-06-04T20:48:34.805801Z DEBUG  Diff of formatted plans:
 QueryPlan {
   Fetch(service: "products") {
     {
       topProducts {
         name
       }
     }
   },
 }

TRACE  JS query plan Debug: Some(
    Fetch(
        FetchNode {
            service_name: "products",
            requires: [],
            variable_usages: [],
            operation: "{topProducts{name}}",
            operation_name: None,
            operation_kind: Query,
            id: None,
            input_rewrites: None,
            output_rewrites: None,
            context_rewrites: None,
            schema_aware_hash: QueryHash(
                "8e0d7c1104748fe6a45011035f772e2b107dd851dc80d0b51c9987b830a9267b",
            ),
            authorization: CacheKeyMetadata {
                is_authenticated: false,
                scopes: [],
                policies: [],
            },
        },
    ),
)
TRACE  Rust query plan Debug: Some(
    Fetch(
        FetchNode {
            service_name: "products",
            requires: [],
            variable_usages: [],
            operation: "{\n  topProducts {\n    name\n  }\n}\n",
            operation_name: None,
            operation_kind: Query,
            id: None,
            input_rewrites: None,
            output_rewrites: None,
            context_rewrites: None,
            schema_aware_hash: QueryHash(
                "",
            ),
            authorization: CacheKeyMetadata {
                is_authenticated: false,
                scopes: [],
                policies: [],
            },
        },
    ),
)
```

Differences are whitespace in the `operation` string, and `schema_aware_hash` being only filled in on one side. Clearly `derive(PartialEq)` is not a good comparison function, but making a better one is for another PR.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -291,6 +291,7 @@ dependencies = [
  "console",
  "console-subscriber",
  "cookie",
+ "crossbeam-channel",
  "dashmap",
  "derivative",
  "derive_more",
--- a/apollo-router/Cargo.toml
+++ b/apollo-router/Cargo.toml
@@ -68,7 +68,7 @@ askama = "0.12.1"
 access-json = "0.1.0"
 anyhow = "1.0.80"
 apollo-compiler.workspace = true
-apollo-federation = { path = "../apollo-federation", version = "=1.48.0"}
+apollo-federation = { path = "../apollo-federation", version = "=1.48.0" }
 arc-swap = "1.6.0"
 async-channel = "1.9.0"
 async-compression = { version = "0.4.6", features = [
@@ -91,6 +91,7 @@ clap = { version = "4.5.1", default-features = false, features = [
 ] }
 console-subscriber = { version = "0.2.0", optional = true }
 cookie = { version = "0.18.0", default-features = false }
+crossbeam-channel = "0.5"
 ci_info = { version = "0.14.14", features = ["serde-1"] }
 dashmap = { version = "5.5.3", features = ["serde"] }
 derivative = "2.2.0"
--- a/apollo-router/src/error.rs
+++ b/apollo-router/src/error.rs
@@ -529,16 +529,21 @@ impl std::fmt::Display for PlanErrors {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         f.write_fmt(format_args!(
             "query validation errors: {}",
-            self.errors
-                .iter()
-                .map(|e| e
-                    .message
-                    .clone()
-                    .unwrap_or_else(|| "UNKNWON ERROR".to_string()))
-                .collect::<Vec<String>>()
-                .join(", ")
+            format_bridge_errors(&self.errors)
         ))
     }
+}
+
+pub(crate) fn format_bridge_errors(errors: &[router_bridge::planner::PlanError]) -> String {
+    errors
+        .iter()
+        .map(|e| {
+            e.message
+                .clone()
+                .unwrap_or_else(|| "UNKNWON ERROR".to_string())
+        })
+        .collect::<Vec<String>>()
+        .join(", ")
 }
 
 /// Error in the schema.
--- a/apollo-router/src/plugins/file_uploads/rearrange_query_plan.rs
+++ b/apollo-router/src/plugins/file_uploads/rearrange_query_plan.rs
@@ -1,6 +1,7 @@
 use std::cmp;
 use std::collections::BTreeMap;
 use std::collections::HashMap;
+use std::sync::Arc;
 
 use indexmap::IndexMap;
 use indexmap::IndexSet;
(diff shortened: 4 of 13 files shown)


lrlna:
I did a tiny bit of load testing. Imo there isn't a huge difference, so this should work!

### rate: 200 requests per second, duration: 1m 
```bash
jq -nscMR 'while(true; .+1) | {
    method: "POST",
    url: "http://127.0.0.1:4000/",
    header: { "Content-type": ["application/json"] },
    body: (
      {
        query: "{ query_\(.): topProducts { field_\(.): name } }"
      } | tojson
    ) | @base64
   }' |
  vegeta attack -lazy -format=json -rate=200/s -duration=1m |
  vegeta report
```

#### `experimental_query_planner_mode: legacy`
```bash
Requests      [total, rate, throughput]         12000, 200.02, 200.01
Duration      [total, attack, wait]             59.998s, 59.995s, 3.346ms
Latencies     [min, mean, 50, 90, 95, 99, max]  3.1ms, 4.412ms, 3.939ms, 4.803ms, 5.341ms, 9.008ms, 149.346ms
Bytes In      [total, mean]                     5483465, 456.96
Bytes Out     [total, mean]                     721786, 60.15
Success       [ratio]                           100.00%
Status Codes  [code:count]                      200:12000  
Error Set:
```
#### `experimental_query_planner_mode:both`
```bash
Requests      [total, rate, throughput]         12000, 200.02, 200.00
Duration      [total, attack, wait]             59.999s, 59.994s, 4.104ms
Latencies     [min, mean, 50, 90, 95, 99, max]  3.725ms, 4.565ms, 4.182ms, 4.871ms, 5.777ms, 10.691ms, 68.271ms
Bytes In      [total, mean]                     8424000, 702.00
Bytes Out     [total, mean]                     721786, 60.15
Success       [ratio]                           100.00%
Status Codes  [code:count]                      200:12000  
Error Set:
```
### rate: 1000 requests per second, duration: 1m 
Success ratio is somewhat identical, and different runs yielded slightly different results, so there isn't anything inherently better about running `both`.

#### `experimental_query_planner_mode: legacy`
```bash
Requests      [total, rate, throughput]         60000, 1000.02, 387.68
Duration      [total, attack, wait]             1m19s, 59.999s, 18.857s
Latencies     [min, mean, 50, 90, 95, 99, max]  43.083µs, 915.765ms, 967.262ms, 1.436s, 1.502s, 12.782s, 19.521s
Bytes In      [total, mean]                     23119884, 385.33
Bytes Out     [total, mean]                     2572920, 42.88
Success       [ratio]                           50.95%
Status Codes  [code:count]                      0:18143  200:30571  500:11286  
Error Set:
500 Internal Server Error
```
#### `experimental_query_planner_mode:both`
```bash
Requests      [total, rate, throughput]         60000, 1000.01, 570.56
Duration      [total, attack, wait]             1m1s, 59.999s, 1.46s
Latencies     [min, mean, 50, 90, 95, 99, max]  1.229ms, 1.002s, 1.601s, 1.901s, 2s, 2.408s, 2.654s
Bytes In      [total, mean]                     28281630, 471.36
Bytes Out     [total, mean]                     3697786, 61.63
Success       [ratio]                           58.44%
Status Codes  [code:count]                      200:35066  500:24934  
Error Set:
500 Internal Server Error
```

lrlna (approved):

PR #5342, opened 2024-06-04
[docs] Add note about expanding query planner heap size

smyrick:
This is an experimental variable and will be removed in the future but it has been used a number of time now and is worth documenting


docs/source/configuration/overview.mdx
```diff
@@ -9,11 +9,11 @@ import RedisTLS from '../../shared/redis-tls.mdx'
 Learn how to customize the behavior of your Apollo Router with environment variables, command-line commands and options, and YAML file configuration.
 
 ## Environment variables
+Most environment variables are also documented with their [Command-line options](#command-line-options) counterpart. However, there are some options that are only available to configure as environment variables.
```
Meschreiber, line 12: This addition confuses me a bit since I don't see the CLI counterpart in the table below.

Also, we should specify which options are only available via env variable.

docs/source/configuration/overview.mdx
```diff
@@ -24,39 +24,51 @@ APOLLO_KEY="..." APOLLO_GRAPH_REF="..." ./router
...
+      </td>
+    </tr>
+    <tr>
+      <td style="min-width: 150px;">
+      
+      ##### `APOLLO_ROUTER_BRIDGE_EXPERIMENTAL_V8_INITIAL_HEAP_SIZE`
+      
+      </td>
+      <td>
+      `EXPERIMENTAL`: The starting size of query planner heap to use in MB. Setting `APOLLO_ROUTER_BRIDGE_EXPERIMENTAL_V8_INITIAL_HEAP_SIZE=4096` is ~4GB.
```
Meschreiber, line 63:
```suggestion
      The initial size to use for query planner heaps in MB. For example, setting this variable to `4096` sets an initial heap size of ~4GB.
```

docs/source/configuration/overview.mdx
```diff
@@ -24,39 +24,51 @@ APOLLO_KEY="..." APOLLO_GRAPH_REF="..." ./router
...
+
+      Apollo Router v1.x uses a JavaScript runtime internally to run the query planner. While this is mostly an internal implementation detail, large supergraphs tend to increase the amount of memory required in the Router. The query planner heap will dynamically expand as needed however, increasing the starting size of the memory allocated to the heap can improve the first planning times for large operations against a large supergraphs.
```
Meschreiber, line 65:
```suggestion
      The Apollo Router v1.x uses a JavaScript runtime internally to run the query planner. Large supergraphs tend to increase the amount of memory required in the router. The query planner heap dynamically expands as needed but increasing the initial size of the memory allocated to the heap can improve the performance of initial large operations against large supergraphs.
```

Meschreiber (commented):
Sorry for the delay @smyrick , this improvement slipped past us. I'm hoping this experimental env variable still exists in its current state.

I have a questions/suggestions about an initial additions, and then just nits for conciseness/clarity for the env variable addition.

Thanks for documenting this!

docs/source/configuration/overview.mdx
smyrick: > I don't see the CLI counterpart in the table below.

That is in the next section below `## Command-line options`

> Also, we should specify which options are only available via env variable.

That is supposed to be this section here.

Meschreiber: Got it, would suggest this to clarify then:

```suggestion

You can set most environment variables via their [command-line option](#command-line-options) counterpart. The following values are only configureable as environment variables:
```

Meschreiber (approved):
Just left one suggestion for clarity. Otherwise LGTM.

Since this is a merge to main, tagging in @shorgi for a second approval.

abernix:
This was a PR that we should have probably landed a while ago.  At this point, we'll opt to leave this out as we're looking at a deprecation route for it.

PR #5343, opened 2024-06-04
Fix equality-checking logic for operation elements

sachindshinde:
This PR fixes equality/hash logic for operation elements. Specifically:

1. Equality shouldn't care about order in map-likes (i.e. argument order in fields and directives, and field order in object values).
    1.  The previous code for key equality used sorted arguments in directives, but it wasn't recursively sorting values in directive arguments.
    2. The previous code for field/inline fragment/fragment spread equality wasn't comparing using sorted map-likes.
    3.  As part of fixing the above, the new code has added a `sorted_arguments` field to `Field`.

2. Equality shouldn't care about directive order (this is specifically a choice of federation, and not GraphQL).
    1. The previous code for key equality wasn't sorting directives.
    2. The previous code for field/inline fragment/fragment spread equality wasn't sorting directives.
    3. Note that sorting directives is more complicated than sorting arguments/object value fields, since a directive may be repeated on a schema element. This in particular means that we need to establish an order between arbitrary GraphQL values. The ordering I've written in this PR isn't set in stone; it just needs to be consistent and ideally fast.
    4. The new code for field/inline fragment/fragment spread equality reuses the directive list in `SelectionKey`.
5. Equality shouldn't care about selection ID unless `@defer` is present (due to delivery groups being created per selection).
    1. The previous code for field/inline fragment/fragment spread equality was always comparing selection IDs, regardless of `@defer` presence.
    2. The new code for field/inline fragment/fragment spread equality reuses the selection ID from `SelectionKey` (which only stores the selection ID if `@defer` is present).
6. Equality shouldn't care about element's schemas, parent types, type kinds, or sibling typename presence.
    1. The previous code for field/inline fragment/fragment spread equality was comparing schemas, parent types, type kinds, and sibling typename presence.
    2. To be clear, I've made these changes because the JS codebase explicitly avoids comparing these properties in its equality logic. My guess is to permit e.g. cross-schema comparison, type covariance, FieldsInSetCanMerge checks, and interface objects, though I'd have to check the callsites to be sure.

### DIFF
--- a/apollo-federation/src/operation/contains.rs
+++ b/apollo-federation/src/operation/contains.rs
@@ -44,24 +44,21 @@ fn same_value(left: &executable::Value, right: &executable::Value) -> bool {
 
 /// Sort an input value, which means specifically sorting their object values by keys (assuming no
 /// duplicates). This is used for hashing input values in a way consistent with [same_value()].
-fn sort_value(value: &executable::Value) -> executable::Value {
+fn sort_value(value: &mut executable::Value) {
     use apollo_compiler::executable::Value;
     match value {
-        Value::List(elems) => Value::List(
+        Value::List(elems) => {
             elems
-                .iter()
-                .map(|value| Node::new(sort_value(value)))
-                .collect(),
-        ),
-        Value::Object(pairs) => {
-            let mut pairs = pairs
-                .iter()
-                .map(|(name, value)| (name.clone(), Node::new(sort_value(value))))
-                .collect::<Vec<_>>();
-            pairs.sort_by(|left, right| left.0.cmp(&right.0));
-            Value::Object(pairs)
+                .iter_mut()
+                .for_each(|value| sort_value(value.make_mut()));
         }
-        _ => value.clone(),
+        Value::Object(pairs) => {
+            pairs
+                .iter_mut()
+                .for_each(|(_, value)| sort_value(value.make_mut()));
+            pairs.sort_by(|left, right| left.0.cmp(&right.0));
+        }
+        _ => {}
     }
 }
 
@@ -160,20 +157,11 @@ fn same_arguments(
 /// Sort arguments, which means specifically sorting arguments by names and object values by keys
 /// (assuming no duplicates). This is used for hashing arguments in a way consistent with
 /// [same_arguments()].
-pub(super) fn sort_arguments(
-    arguments: &[Node<executable::Argument>],
-) -> Vec<Node<executable::Argument>> {
-    let mut arguments = arguments
-        .iter()
-        .map(|arg| {
-            Node::new(executable::Argument {
-                name: arg.name.clone(),
-                value: Node::new(sort_value(&arg.value)),
-            })
-        })
-        .collect::<Vec<_>>();
-    arguments.sort_by(|left, right| left.name.cmp(&right.name));
+pub(super) fn sort_arguments(arguments: &mut [Node<executable::Argument>]) {
     arguments
+        .iter_mut()
+        .for_each(|arg| sort_value(arg.make_mut().value.make_mut()));
+    arguments.sort_by(|left, right| left.name.cmp(&right.name));
 }
 
 /// Compare sorted arguments; see [compare_sorted_name_value_pairs()] for semantics. This is used
@@ -205,23 +193,15 @@ fn same_directives(left: &executable::DirectiveList, right: &executable::Directi
 /// Sort directives, which means specifically sorting their arguments, sorting the directives by
 /// name, and then breaking directive-name ties by comparing sorted arguments. This is used for
 /// hashing arguments in a way consistent with [same_directives()].
-pub(super) fn sort_directives(directives: &executable::DirectiveList) -> executable::DirectiveList {
-    let mut directives = directives
-        .0
-        .iter()
-        .map(|directive| {
-            Node::new(executable::Directive {
-                name: directive.name.clone(),
-                arguments: sort_arguments(&directive.arguments),
-            })
-        })
-        .collect::<Vec<_>>();
+pub(super) fn sort_directives(directives: &mut executable::DirectiveList) {
+    directives
+        .iter_mut()
+        .for_each(|directive| sort_arguments(&mut directive.make_mut().arguments));
     directives.sort_by(|left, right| {
         left.name
             .cmp(&right.name)
             .then_with(|| compare_sorted_arguments(&left.arguments, &right.arguments))
     });
-    executable::DirectiveList(directives)
 }
 
 pub(super) fn is_deferred_selection(directives: &executable::DirectiveList) -> bool {
--- a/apollo-federation/src/operation/contains.rs
+++ b/apollo-federation/src/operation/contains.rs
@@ -1,5 +1,6 @@
 use std::collections::HashMap;
 
+use apollo_compiler::ast::Name;
 use apollo_compiler::executable;
 use apollo_compiler::Node;
 
@@ -25,7 +26,7 @@ fn same_value(left: &executable::Value, right: &executable::Value) -> bool {
         (Value::Float(left), Value::Float(right)) => left == right,
         (Value::Int(left), Value::Int(right)) => left == right,
         (Value::Boolean(left), Value::Boolean(right)) => left == right,
-        (Value::List(left), Value::List(right)) => left
+        (Value::List(left), Value::List(right)) if left.len() == right.len() => left
             .iter()
             .zip(right.iter())
             .all(|(left, right)| same_value(left, right)),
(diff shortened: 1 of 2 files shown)


apollo-federation/src/operation/mod.rs
```diff
@@ -104,6 +104,98 @@ fn same_value(left: &executable::Value, right: &executable::Value) -> bool {
     }
 }
 
+/// Sort an input value, which means specifically sorting their object values by keys (assuming no
+/// duplicates). This is used for hashing input values in a way consistent with [same_value()].
+fn sort_value(value: &executable::Value) -> executable::Value {
```
goto-bus-stop, line 109: The `apollo-compiler` API is really made for mutation rather than creating new instances. All the `sort_*` functions should ideally take a `&mut executable::WhateverTheType` and mutate. It's also easier to implement them that way :)

apollo-federation/src/operation/mod.rs
```diff
@@ -1162,10 +1312,11 @@ mod field_selection {
 
     /// The non-selection-set data of `FieldSelection`, used with operation paths and graph
     /// paths.
-    #[derive(Clone, PartialEq, Eq, Hash)]
+    #[derive(Clone)]
     pub(crate) struct Field {
         data: FieldData,
         key: SelectionKey,
+        sorted_arguments: Arc<Vec<Node<executable::Argument>>>,
```
goto-bus-stop, line 1319: Can we make sure `FieldData` itself always has a sorted argument list, instead of adding a copy here? If not, this is okay.

I had an idea a while back that I already experimented with a little bit. It is to have our own immutable `DirectiveList` and `ArgumentList` wrapper types that guarantee sorted order in their constructor. I can PR that in a followup.

apollo-federation/src/operation/mod.rs
```diff
@@ -1174,10 +1325,29 @@ mod field_selection {
...
+        fn eq(&self, other: &Self) -> bool {
+            self.data.field_position.field_name() == other.data.field_position.field_name()
+                && self.key == other.key
+                && self.sorted_arguments == other.sorted_arguments
+        }
+    }
+
+    impl Eq for Field {}
+
+    impl Hash for Field {
```
goto-bus-stop, line 1338: Not for now, but it would be interesting to use apollo-compiler's `Node<>` type instead of `Arc<>` in select places in this file, as it caches the hash.

goto-bus-stop (changes requested):
Seems good overall!

apollo-federation/src/operation/mod.rs
```diff
@@ -104,6 +104,98 @@ fn same_value(left: &executable::Value, right: &executable::Value) -> bool {
...
+                .zip(right)
+                .map(|(left, right)| compare_sorted_value(left, right))
+                .find(|o| o.is_ne())
+                .unwrap_or(std::cmp::Ordering::Equal)
+        }),
+        (Value::Object(left), Value::Object(right)) => compare_sorted_name_value_pairs(
+            left.iter().map(|pair| (&pair.0, &pair.1)),
+            right.iter().map(|pair| (&pair.0, &pair.1)),
+        ),
+        _ => discriminant(left).cmp(&discriminant(right)),
```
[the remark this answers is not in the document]
sachindshinde: @goto-bus-stop 
`left` and `right` can be different `Value` variants, e.g. we're comparing a `String` to `Null`. This can happen due to e.g. needing to sort a directive list representing:
```graphql
@foo(bar: null) @foo(bar: "bar")
```
In particular, if `bar` is a custom scalar, the values compared can be any valid value.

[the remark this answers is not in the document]
goto-bus-stop: Oh, I misunderstood the context this is in, this is to ensure ordering in that case.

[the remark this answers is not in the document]
goto-bus-stop: We might use `std::mem::discriminant` instead of the manual `discriminant()` implementation though: https://doc.rust-lang.org/std/mem/fn.discriminant.html

apollo-federation/src/operation/mod.rs
sachindshinde: @goto-bus-stop 
The sorting currently is really just for faster equality/hash. It's possible we re-sort on every invocation of those, and that would take less memory since we wouldn't need to store sorted field arguments or element directives, though it'd be slower.

We could sort in-place, though that has some drawbacks:
1. The JS codebase currently leaves arguments/directives/object value fields unsorted, so we'd be diverging from that (that's more of a short-term problem though).
2. Not sorting in-place has the benefit that users basically see the order they provided in the router query, in the subgraph queries. For the case of directives, users may have particularly been relying on us not changing directive order (even though they shouldn't be), and they may view sorting as a breaking change.
3. The sort order we've arbitrarily chosen in this PR suddenly matters more, since it becomes publicly exposed, and changes in sort order influence the order in generated query plans.
4. I'm not sure what the practical memory benefit is of sorting in-place (it's certainly less memory, but I'm not sure if the amount is significant).

The 2nd point is the real issue IMO, but it'll take some time to think through/evaluate.

apollo-federation/src/operation/mod.rs
sachindshinde: Makes sense! Pushed up changes to switch to `&mut` in the latest commit.

apollo-federation/src/operation/mod.rs
sachindshinde: @goto-bus-stop 
Yeah, it would be nice to have automatic cache-hashing, it's certainly something we should consider in the future.

apollo-federation/src/operation/mod.rs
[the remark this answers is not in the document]
sachindshinde: @goto-bus-stop 
It looks like the built-in discriminant doesn't implement `Ord` though, unfortunately.

apollo-federation/src/operation/mod.rs
sachindshinde: Alright, pushed up with merged conflicts resolved.

apollo-federation/src/operation/mod.rs
goto-bus-stop: Thanks, those are good points. AFAIK JS doesn't necessarily maintain all order, because of its lossy representation of object values for example. But directive order is very relevant. The `DirectiveList` and `ArgumentList` structures could account for this though and maintain two sort orders simultaneously, one for display and another one for hashing and comparison.

Anyways, that's a separate issue.

goto-bus-stop: FWIW if the **only** purpose of `sorted_arguments` is equality/hashing, we could store only a hashed `u64` in the structure instead of a whole `Arc<Vec<>>`

apollo-federation/src/operation/contains.rs
```diff
@@ -41,6 +42,95 @@ fn same_value(left: &executable::Value, right: &executable::Value) -> bool {
...
+/// recursively by value. This is intended to compute an ordering quickly for hashing.
+fn compare_sorted_name_value_pairs<'doc>(
+    left: impl ExactSizeIterator<Item = (&'doc Name, &'doc Node<executable::Value>)>,
+    right: impl ExactSizeIterator<Item = (&'doc Name, &'doc Node<executable::Value>)>,
+) -> std::cmp::Ordering {
+    left.len().cmp(&right.len()).then_with(|| {
+        let mut pairs = vec![];
+        for ((left_name, left_value), (right_name, right_value)) in left.zip(right) {
+            match left_name.cmp(right_name) {
+                std::cmp::Ordering::Equal => pairs.push((left_value, right_value)),
```
goto-bus-stop, line 122: Hmm, we are preparing this `pairs` vec with `(left_value, right_value)`, but then afterwards we just iterate through it in order and compare each `left_value` with `right_value`. Can't we just do the `compare_sorted_value` inline here?

goto-bus-stop (approved):
Thanks for answering my concerns 😁

apollo-federation/src/operation/mod.rs
sachindshinde: With later ECMAScript versions, object properties should be in [property creation order](https://262.ecma-international.org/10.0/#sec-ordinaryownpropertykeys) (so if we've been iterating in-order, it might be fine, though not for sure there). It would be nice if `DirectiveList`/`ArgumentList` accounted for ordering concerns.

For hashing, I was curious about storing the hash in the data structure (and using `same_arguments()` for equality), though I wasn't sure how to do that properly in Rust. (It probably won't end up being significant, but definitely down to do it later if it is 😄 )

apollo-federation/src/operation/contains.rs
sachindshinde: @goto-bus-stop 
My thoughts here were that recursing into the values with `compare_sorted_value()` could potentially be expensive, but differences in name/key sets between the pairs would probably be faster to compute, so we should early return if the key sets are different and only recurse if they're the same.

It looks like I can avoid the `vec` creation by just having separate iterators for names and values, so I went ahead and changed the API of `compare_sorted_name_value_pairs()` accordingly in the latest commit, let me know if there's a simpler way and I can make a followup PR.

PR #5345, opened 2024-06-05
FED-55: [Fetch Dependency Graph] optimise portion of reduce_and_optimize 

duckki:
I confirmed several test snapshots have improved, even though most of them didn't change test's status (still have snapshot assertion failures, but they are closer to the expected).

However, two tests that worked before are now failing. I couldn't figure out why. I think we better look at those together.

```
#[should_panic(expected = "Details: Cannot merge field selections from different schemas")]
fn does_not_error_on_some_complex_fetch_group_dependencies() {
```

```
#[test]
#[should_panic(
    expected = "Root nodes should have no remaining nodes unhandled, but got: [1 (missing: [2])]"
)]
fn handles_root_operation_shareable_in_many_subgraphs() {
```

Fix: https://apollographql.atlassian.net/browse/FED-55

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [X] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


goto-bus-stop (commented):
Still working through this, but I had some initial minor comments

goto-bus-stop:
For whoever takes over the review...
I reviewed `remove_empty_nodes` and part of `remove_useless_nodes`. The parts still to review are only in `fetch_dependency_graph.rs`:

- `is_useless_node`
- `merge_child_fetches_for_same_subgraph_and_path`
- `merge_fetches_to_same_subgraph_and_same_inputs`

TylerBloom:
I brought this up during the standup, but this also addresses `tests::query_plan::build_query_plan_tests::merged_abstract_types_handling::handles_types_with_no_common_supertype_at_the_same_merge_at`. That test is marked as `should_panic` but it shouldn't after these changes and rebasing onto `dev`.

apollo-federation/src/query_plan/fetch_dependency_graph.rs
```diff
@@ -964,14 +989,471 @@ impl FetchDependencyGraph {
     ///     was the one with the require and that forced some dependencies. Those fetch should have
     ///     no dependents and we can just remove them.
     ///  2) fetches that are made in parallel to the same subgraph and the same path, and merge those.
-    fn reduce_and_optimize(&mut self) {
+    fn reduce_and_optimize(&mut self) -> Result<(), FederationError> {
         if std::mem::replace(&mut self.is_optimized, true) {
```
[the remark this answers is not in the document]
duckki: Indeed.

apollo-federation/src/query_plan/fetch_dependency_graph.rs
```diff
@@ -955,23 +978,482 @@ impl FetchDependencyGraph {
...
+        let input_selections: Vec<&Selection> = self_inputs
+            .selection_sets_per_parent_type
+            .values()
+            .flat_map(|s| s.selections.values())
+            .collect();
+        // Checks that every selection is contained in the input selections.
+        node.selection_set
+            .selection_set
+            .iter()
+            .try_fold(true, |acc, selection| {
```
sachindshinde, line 1224: Because this `try_fold()` is meant to implement JS `.every()`, you'll want to avoid executing the rest of this function when `acc` is `false`. The easiest thing to do is to just early return `false` here if `acc` is `false`. (Note that this isn't completely short-circuiting since you'll still end up iterating through the rest of the selections in the selection set; one possibility is to extract the code in `try_fold()` into a closure that returns `Result<bool, FederationError>`, and then change that into e.g. `ControlFlow<Result<(), FederationError>>` when calling that closure in the `try_fold()`'s closure, although this is a minor optimization I think.)

apollo-federation/src/query_plan/fetch_dependency_graph.rs
```diff
@@ -955,23 +978,482 @@ impl FetchDependencyGraph {
...
+                if is_interface_type_condition_on_interface_object(selection)? {
+                    return Ok(false);
+                }
+
+                let condition_in_supergraph =
+                    condition_in_supergraph_if_interface_object(selection)?;
+                let Some(condition_in_supergraph) = condition_in_supergraph else {
+                    // We're not in the @interfaceObject case described above. We just check that
+                    // an input selection contains the one we check.
+                    return Ok(acc
```
sachindshinde, line 1236: With the early return above, you can remove the `acc &&` here.

apollo-federation/src/query_plan/fetch_dependency_graph.rs
```diff
@@ -955,23 +978,482 @@ impl FetchDependencyGraph {
...
+                        let Ok(Some(input_selection_set)) = input.selection_set() else {
+                            return false;
+                        };
+                        input_selection_set.contains(sub_selection_set)
+                    }))
+                } else {
+                    Ok(false)
+                }
+            })
+    }
```
sachindshinde, line 1298: You need to set `isKnownUseful` at the end of this function.

apollo-federation/src/query_plan/fetch_dependency_graph.rs
```diff
@@ -955,23 +978,482 @@ impl FetchDependencyGraph {
...
+
+    /// - Calls `on_modification` if necessary.
+    fn recursive_merge_child_fetches_for_same_subgraph_and_path(
+        &mut self,
+        node_index: NodeIndex,
+    ) -> Result<(), FederationError> {
+        // We're traversing the `self.graph` in DFS order and mutate it top-down.
+        // - Assuming the graph is a DAG and has no cycle.
+        let children_nodes: Vec<_> = self.children_of(node_index).collect();
+        if children_nodes.len() < 2 {
```
sachindshinde, line 1317: You need to still recurse into the child nodes if there's one child.

TylerBloom (approved):

apollo-federation/src/query_plan/fetch_dependency_graph.rs
```diff
@@ -912,14 +916,33 @@ impl FetchDependencyGraph {
     }
 
     fn remove_node(&mut self, node_index: NodeIndex) {
+        self.on_modification();
         let edges_to_remove: Vec<EdgeIndex> =
             self.graph.edges(node_index).map(|edge| edge.id()).collect();
         for edge in edges_to_remove {
             self.graph.remove_edge(edge);
         }
+        // Note: Probably the `remove_edge` calls are unnecessary, as `remove_node` should remove
```
sachindshinde, line 925: The API docs say that `remove_node()` should remove edges, so you can indeed remove the calls above to `remove_edge()`.

apollo-federation/src/query_plan/fetch_dependency_graph.rs
```diff
@@ -953,23 +976,482 @@ impl FetchDependencyGraph {
...
+                    // and recursively so), so it does not affect our current iteration.
+                    self.remove_redundant_edges(j_node_index);
+
+                    break; // skip the rest of `j`'s iteration
+                }
+            }
+        }
+
+        // Now we recurse to the sub-nodes.
+        for c in children_nodes {
```
sachindshinde, line 1369: Note that `children_nodes` may contain `NodeIndex`s that no longer exist, so you'll need to remove them before iterating.

sachindshinde (changes requested):

duckki (commented):
Thanks for the review! @sachindshinde 
Follow-up PR: https://github.com/apollographql/router/pull/5401

PR #5347, opened 2024-06-05
chore(deps): update apollo-compiler to v1.0.0-beta.17

### LINKED ISSUE #5313: Extremely slow parse times can cause the router to stop service requests
The version of `apollo-compiler` used by `router` has a perfomance issue in validation, which combined with large schemas and queries with a lot of fragment spreads can cause devastatingly slow parsing times (in seconds). I've opened an issue on [apollo-compiler](https://github.com/apollographql/apollo-rs/issues/862) which contains more details about the problem.

Combined with the fact parsing is executed between `await` points and is blocking (for potentially seconds), this can cause the router's async runtime to completely fall apart and not being able to respond to any queries for minutes. I've opened [a PR](https://github.com/apollographql/router/pull/5235) to run parsing through `spawn_blocking` instead of blocking a worker. This is especially concerning on schema reloads, where contention is high.

Both issues most likely need to be fixed to maintain good performance on schema reloads when it comes to parsing. Additionally, it might be worth thinking about how the `implementers_map` (apollo-compiler issue) can be cached through router reloads.

goto-bus-stop:
Bug fixes and performance improvements

Fixes #5313

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -192,9 +192,9 @@ checksum = "5ad32ce52e4161730f7098c077cd2ed6229b5804ccf99e5366be1ab72a98b4e1"
 
 [[package]]
 name = "apollo-compiler"
-version = "1.0.0-beta.16"
+version = "1.0.0-beta.17"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "175659cea0232b38bfacd1505aed00221cc4028d848699ce9e3422c6bf87d90a"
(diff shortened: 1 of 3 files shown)


xuorig:
That was fast! Thank you!

PR #5355, opened 2024-06-05
More `@connect` validations

dylan-apollo:
Includes:

* Require that all fields on `Query` have a `@connect`
* Check for relative vs absolute URLs in `@connect` are applied correctly, depending on whether a `source` is set
* Provide a warning if it looks like users are trying to use a source, but didn't import the directive (since this is usually a mistake)

Because there is now a warning, there's also refactoring to make this a little closer to fbp-style build messages with severity (instead of only errors).


apollo-federation/src/sources/connect/models/validation.rs
```diff
@@ -606,6 +776,84 @@ mod test_validate_source {
...
+    }
+
+    #[test]
+    fn absolute_connect_url_with_source() {
+        let schema = include_str!("test_data/absolute_connect_url_with_source.graphql");
+        let schema = Schema::parse(schema, "test.graphql").unwrap();
+        let errors = validate(schema);
+        assert_eq!(errors.len(), 1);
+        assert_eq!(errors[0].code, Code::AbsoluteConnectUrlWithSource);
+        assert_snapshot!(errors[0].message, @r###"`GET` in `@connect(http:)` on `Query.resources` contains the absolute URL "http://127.0.0.1/resources" while also specifying a `source`. Either remove the `source` argument or change the URL to a path."###);
```
lennyburdette, line 809: Interesting! I was thinking that if you had an absolute URL in the connect, it would take precedence over the source baseURL. but it's potentially still useful because any source headers still apply. maybe it's better to be strict now and relax this if customers need it?

dylan-apollo: Hmm, yeah inheriting things could still be useful, I didn't know that the router code was / will be capable of that.

Eventually, all base/absolute URLs should probably be coming from Router config/env vars, so this probably won't be a super-useful check long term. In fact, at that point, we might detect absolute URLs and return warnings telling people to put things in config. That's not possible today, though, right?

lennyburdette (approved):

PR #5045, opened 2024-04-29
add support of response status_code in rhai

bnjjj:
Added support of `response.status_code` on `Response` interface in rhai.

Convert response status code to a string.

```rhai
if response.status_code.to_string() == "200" {
    print(`ok`);
}
```

Also useful if you want to convert response status code to a number

```rhai
if parse_int(response.status_code.to_string()) == 200 {
    print(`ok`);
}
```

Fixes #5042

---

**Checklist**

Complete the checklist (and note appropriate exceptions) before the PR is marked ready-for-review.

- [ ] Changes are compatible[^1]
- [ ] Documentation[^2] completed
- [ ] Performance impact assessed and acceptable
- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[^1]: It may be appropriate to bring upcoming changes to the attention of other (impacted) groups. Please endeavour to do this before seeking PR approval. The mechanism for doing this will vary considerably, so use your judgement as to how and when to do this.
[^2]: Configuration is an important part of many changes. Where applicable please try to document configuration examples.
[^3]: Tick whichever testing boxes are applicable. If you are adding Manual Tests, please document the manual testing (extensively) in the Exceptions.

PR #5358, opened 2024-06-06
Rhai: fix 'status_code' on router, supergraph and execution services

### LINKED ISSUE #5357: rhai: `status_code` is not accessible for router, supergraph and execution services
**Describe the bug**
Trying to access `status_code` on router, supergraph or execution services cause exception

**To Reproduce**
1. Create Rhai script:
```
fn supergraph_service(service) {
    let f = |response| {
        print(response.status_code);
        response
    };

    service.map_response(f);
}
```
2. Submit any request
3. See error

**Output**
```
map_response callback failed: Unknown property 'status_code' - a getter is not registered for type 'alloc::sync::Arc<std::sync::mutex::Mutex<core::option::Option<apollo_router::plugins::rhai::engine::RhaiSupergraphResponse>>>' (line 3, position 24)
in closure call
```

IvanGoncharov:
Fix 'status_code' on router, supergraph and execution services

Follows: https://github.com/apollographql/router/pull/5045
Fixes #5357

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-372]: https://apollographql.atlassian.net/browse/ROUTER-372?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Geal:
what changed since https://github.com/apollographql/router/pull/5045/ that made it fails at router, supergraph and execution stages?
It was implemented in `register_rhai_interface` and `register_rhai_router_interface`

Also, I don't think want to give access to the status code at the supergraph and execution now, considering the new pipeline will remove HTTP concerns from there

abernix:
> I don't think want to give access to the status code at the supergraph and execution now, considering the new pipeline will remove HTTP concerns from there

This seems like a fair concern, it may only make sense to have the HTTP status code at HTTP specific points in the pipeline?  Wouldn't the other layers just be inheriting the outer state anyhow?

apollo-router/src/plugins/rhai/tests.rs
```diff
@@ -38,6 +40,78 @@ use crate::services::SupergraphRequest;
...
+    let response = Arc::new(Mutex::new(Some(subgraph::Response::fake_builder().build())));
+
+    // Call our rhai test function. If it doesn't return an error, the test failed.
+    block
+        .engine
+        .call_fn(&mut guard, &block.ast, fn_name, (response,))
+}
+
+async fn call_rhai_function_with_arg<'a, T: Sync + Send + 'a + 'static>(
+    fn_name: &'a str,
```
garypen, line 80: We don't need the `'a` lifetime.
```suggestion
async fn call_rhai_function_with_arg<T: Sync + Send + 'static>(
    fn_name: &str,
```

garypen (commented):
I think @Geal Is right and we probably shouldn't expose the status code at execution and supergraph responses.

IvanGoncharov:
> what changed since https://github.com/apollographql/router/pull/5045 that made it fail at router, supergraph and execution stages?
It was implemented in register_rhai_interface and register_rhai_router_interface

@Geal This code is a mess, but I will try to explain.
So we have this imports:
https://github.com/apollographql/router/blob/3d2269f00223559b10b37cd5bc81fc9995c422d8/apollo-router/src/plugins/rhai/engine.rs#L37-L40

And `register_rhai_router_interface` registered `status_code` here:
https://github.com/apollographql/router/blob/3d2269f00223559b10b37cd5bc81fc9995c422d8/apollo-router/src/plugins/rhai/engine.rs#L1465-L1470
And was called here:
https://github.com/apollographql/router/blob/3d2269f00223559b10b37cd5bc81fc9995c422d8/apollo-router/src/plugins/rhai/engine.rs#L1742

So it registered `status_code` on `router::Response` which coming from here:
https://github.com/apollographql/router/blob/3d2269f00223559b10b37cd5bc81fc9995c422d8/apollo-router/src/plugins/rhai/router.rs#L9

So we basically registered `status_code` on `crate::services::router::Response` but Rhai will get this type, instead it will get 
https://github.com/apollographql/router/blob/3d2269f00223559b10b37cd5bc81fc9995c422d8/apollo-router/src/plugins/rhai/engine.rs#L1265-L1269

Same happening for `register_rhai_interface` it register bunch of stuff for `$base::Response` which becomes:
🔴  `crate::services::supergraph::Response` where Rhai gets `RhaiSupergraphResponse` 
🔴  `crate::services::supergraph::Response` where Rhai gets `RhaiExecutionResponse`
🟢  `crate::services::subgraph::Response` this one is correct

At the same time `register_rhai_interface` can't use `$base:FirstResponse` instead of `$base:Response` because it not present on `subgraph`.

I tried to cut a scope here and fix `status_code` without touching other stuff, but I think both register_rhai_interface and register_rhai_router_interface macros should be removed because they are confusing and also registering dead code in Rhai.

Also, our current tests are unreliable because they are missing some fields (e.g., `response.headers`) completely.

IvanGoncharov:
> Also, I don't think want to give access to the status code at the supergraph and execution now, considering the new pipeline will remove HTTP concerns from there

@Geal @abernix @garypen Original docs said it should be accessible on responses:
https://github.com/apollographql/router/blob/dev/docs/source/customizations/rhai-api.mdx?plain=1#L557
So, I assumed the intention was to expose it on all services.

I will remove it on execution and supergraph.

IvanGoncharov:
@Geal @abernix @bnjjj If the intention for #5045 was to add `status_code` only for a subgraph, that means I also need to change the Changelog entry from being a fix to saying it's a new feature that adds `status_code` to the router.
Right?

bnjjj:
@IvanGoncharov Yes I think it's a good idea to update the existing changelog to specify it's mainly for subgraph. And also change it in the release notes on GH afterwards. Thanks !

bnjjj (approved):

abernix:
Maybe we get one more thoughtful review from someone who has previously engaged on this and see where that takes us?

.changesets/feat_rhai_status_code_on_a_router_response.md
```diff
@@ -0,0 +1,16 @@
...
+        if response.is_primary() {
+            print(response.status_code);
+        }
+    };
+
+    service.map_response(f);
+}
+```
+
+By [@IvanGoncharov](https://github.com/IvanGoncharov) in https://github.com/apollographql/router/pull/5358
```
abernix, line 16:
```suggestion
### Rhai: `status_code` is available at the `router_service` ([Issue #5357](https://github.com/apollographql/router/issues/5357))

Following up on the introduction of [`status_code`](https://www.apollographql.com/docs/router/customizations/rhai-api/#responsestatus_codeto_string)  as an available property in Rhai's `subgraph_service` response, we've now introduced the availability of the property on the `router_service` response as well:

```rust
fn router_service(service) {
    let f = |response| {
        if response.is_primary() {
            print(response.status_code);
        }
    };

    service.map_response(f);
}
```

By [@IvanGoncharov](https://github.com/IvanGoncharov) in https://github.com/apollographql/router/pull/5358
```

File: apollo-router/src/plugins/rhai/tests.rs (hunk 1 above)
Comment (abernix) [reply apollo-router/src/plugins/rhai/tests.rs:80]: @IvanGoncharov Are you still planning on addressing this feedback?

File: docs/source/customizations/rhai-api.mdx [hunk 3]
```diff
@@ -658,4 +658,8 @@ You can also create your own status code from an integer:
 if response.status_code == status_code_from_int(200) {
     print(`ok`);
 }
-```
\ No newline at end of file
+```
+
+<Note>
+The `response.status_code` object is available _only_ for `map_request` callbacks registered in `router_service` or `subgraph_service`.
```
garypen, line 664:
```suggestion
The `response.status_code` object is available _only_ for `map_response` callbacks registered in `router_service` or `subgraph_service`.
```

apollo-router/src/plugins/rhai/engine.rs
```diff
@@ -508,6 +523,12 @@ mod router_context {
         obj.with_mut(|response| response.context = context);
         Ok(())
     }
+    #[rhai_fn(get = "status_code", pure, return_raw)]
+    pub(crate) fn supergraph_first_response_status_code_get(
+        _obj: &mut SharedMut<supergraph::FirstResponse>,
+    ) -> Result<HeaderMap, Box<EvalAltResult>> {
+        Err(CANNOT_ACCESS_STATUS_CODE_ON_A_SUPERGRAPH_RESPONSE.into())
+    }
```
garypen, line 531: I don't think this is the best way to control access to the status code on supergraph/execution. The best approach is to not implement the function.

So delete:
supergraph_first_response_status_code_get
execution_first_response_status_code_get
get_status_code_supergraph_deferred_response
get_status_code_execution_deferred_response

and update tests accordingly.

garypen (commented):
Some additional feedback.

There's no need to implement functionality we don't want people to use and then deny it at run time. So, we can delete the status code implementations (and any tests) on supergraph and execution services.

As an exception, because of deferred responses, we do that for "deferred" headers/status code because people might be surprised between the difference of primary response and later responses, but we don't do it in general.

apollo-router/src/plugins/rhai/engine.rs
IvanGoncharov: Done ✅

apollo-router/src/plugins/rhai/tests.rs
IvanGoncharov: Missed that one.
Done ✅

garypen (approved):

BrynCooke (approved):

PR #5360, opened 2024-06-06
Add Extensions with_lock() to try and avoid timing issues

garypen:
We hit a lot of problems with timing that causes flaky tests because it's not always easy to know when the compiler drops the Extensions lock.

We can side-step this a lot of the time by introducing (and using) a with_lock() function.

Review Notes: This commit (a0963c4883af0a76e837cabdf5e6651c1c175c2f) contains many changes as I replace `lock()` with `with_lock()` over the entire code base and mark `lock()` `deprecated`. You should review the preceding commits to understand the change and then the referenced commit if you want to see the remaining changes that I made.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

The code change is so small/simple, no tests are required.

**Notes**


BrynCooke (approved):

apollo-router/src/context/extensions/sync.rs
```diff
@@ -24,14 +24,29 @@ impl ExtensionsMutex {
     /// It is CRITICAL to avoid holding on to the mutex guard for too long, particularly across async calls.
     /// Doing so may cause performance degradation or even deadlocks.
     ///
+    /// DEPRECATED: prefer with_lock()
+    ///
     /// See related clippy lint for examples: <https://rust-lang.github.io/rust-clippy/master/index.html#/await_holding_lock>
+    #[deprecated]
```
abernix, line 30: Q: Is there a reason we don't use the Deprecated macro's parameters on things like this here?  Doesn't that provide a better user experience?:

```
#[deprecated(since = "1.49.0", note = "Users should instead use with_lock")]
```

abernix: It definitely renders nicer in the documentation!

PR #5361, opened 2024-06-06
chore(federation): refactor RebasedFragments

goto-bus-stop:
This is not urgent, I mostly wanted to see the effect on QueryPlanningParameters.
I think it may be important to *not* clone QueryPlanningParameters for recursive planning. The JS version does create a new QueryPlanningParameters structure but it contains references to all the same data from the original. In Rust we deeply clone it. In JS, there is some mutable state on the structure: changes made to it propagate to the "parent" plan, but in Rust they do not. This could be causing conditions bugs today?

`RebasedFragments` is a cache of fragments rebased on different
subgraphs. It's only used internally in the query planner and not by
any of the operation types, so this moves the structure out of the general
`operation` module and into the `query_plan` module.

`RebasedFragments` would be cloned to do recursive planning for
conditions. This would mean multiple Arc references to the rebased
cache would be alive, and calls to `for_subgraph` would clone the cache.
Different instances of `RebasedFragments` are independent, so the cache
could be out of sync and cause duplicate work.

This changes `RebasedFragments` to be passed by reference, so it's never
cloned. It uses runtime borrow-checking to fill the cache.
`RebasedFragments` instances are only used for one `build_query_plan`
call, which is fully synchronous and single-threaded, so we can
guarantee that conflicting borrows do not happen.

### DIFF
--- a/apollo-federation/src/operation/mod.rs
+++ b/apollo-federation/src/operation/mod.rs
@@ -2407,7 +2407,7 @@ impl SelectionSet {
     pub(crate) fn add_typename_field_for_abstract_types(
         &self,
         parent_type_if_abstract: Option<AbstractType>,
-        fragments: &Option<&mut RebasedFragments>,
+        fragments: Option<&NamedFragments>,
     ) -> Result<SelectionSet, FederationError> {
         let mut selection_map = SelectionMap::new();
         if let Some(parent) = parent_type_if_abstract {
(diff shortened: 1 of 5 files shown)


goto-bus-stop:
Reading through this again I'm not happy with it. The new lifetime is very ad-hoc and adding more lifetime allocations to everything that uses QueryPlanningParameters isn't going to scale. Given our current priorities I think it's better to revisit later, probably fully replacing QueryPlanningParameters. The PR text is still useful documentation :)

PR #5364, opened 2024-06-06
avoid frequent field lookups in the response visitor

Geal:
Since the response value uses an IndexMap to store objects (a map with an ordered list of fields), looking up a specific field can be expensive. At the point where we execute the response visitor, the response has alredy been validated by the response formatting phase, which means:
- it only contains fields present in the client query
- they are already in the required order

This takes advantage of the same order to iterate on both the selections and fields at the same time

FIXME: what happens when a fragment spread or an inline fragment would request a field already present elsewhere?

examples:

```graphql
{
  a {
    b
  }
  ...  on Query {
    a {
      c
    }
  }
}
```

or the reverse:

```graphql
{
  ...  on Query {
    a {
      c
    }
  }
  a {
    b
  }
}
```
  

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/response.rs
+++ b/apollo-router/src/response.rs
@@ -308,18 +308,38 @@ pub(crate) trait ResponseVisitor {
         selection_set: &apollo_compiler::executable::SelectionSet,
         fields: &Map<ByteString, Value>,
     ) {
+        let mut value_it: serde_json_bytes::Iter = fields.iter();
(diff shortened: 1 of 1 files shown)


Geal:
closing this, as this is not robust against fragment usage

Issue #5282, opened 2024-05-29
Include `request` span in spec_compliant mode


smyrick:
## Is your feature request related to a problem? Please describe.
In `deprecated` mode the Router adds the `request` span
https://github.com/apollographql/router/blob/97e7605cdf798b0bb2fcf39c27fd3143988d7fcc/apollo-router/src/plugins/telemetry/span_factory.rs#L70-L72

![Screenshot 2024-05-29 at 1 59 56 PM](https://github.com/apollographql/router/assets/2446877/117bcf5a-bddc-4cec-9875-6900a5a39826)

In the new `spec_compliant` mode this span was removed

![Screenshot 2024-05-29 at 1 59 13 PM](https://github.com/apollographql/router/assets/2446877/172d7b22-1ef4-4048-bf84-e21dc475e6d9)

## Describe the solution you'd like
I would like to include this span so that the service entry point from DataDog is consistent

## Describe alternatives you've considered
I could possibly add a custom collector to rename spans

## Additional context
This will also help keep span name cardinality down even if this span is added back: https://github.com/apollographql/router/issues/5261

bnjjj:
I'm not sure to understand your need. Indeed there is not request span in `spec_compliant` mode because it's not compliant with the specs. That's why we introduced this mode. And it's going to be the default mode in router 2.x because that's how it should work to be specs compliant. If you still want to have a request span then you should use the deprecated mode.

smyrick:
@bnjjj Can you share what part of the Otel spec we are complying with? I linked above another issue but currently in `spec_compliant` mode the service entry span has a different name everytime because it is now the `router` span and that is messing up DataDog dashboards. This may help solve that https://github.com/apollographql/router/issues/5261

But what is the spec that says there can not be a span called `request`?

bnjjj:
The one mentioned by Bryn [here](https://github.com/apollographql/router/issues/5261#issuecomment-2144717668)

Samjin:
The current state of the GraphQL span conventions is `experimental`, and it lacks explicit guidance on whether a root span should adhere to the same standards considering it originates from an HTTP server request rather than a direct GraphQL operation.

Given that the specification for GraphQL spans has not been officially finalized by OTEL, it may not be necessary to strict adherence to the experimental spec. Adopting a more flexible stance could facilitate a smoother transition away from `deprecated` modes. 

[...]

PR #5365, opened 2024-06-06
feat(telemetry): add the ability to override span name with otel.name attribute

### LINKED ISSUE #5261: Keep router span name consistent and do not use operation name
The span name for the service entry point for Apollo Router OpenTelemtry tracing is `<service_name>:<operation_name>`. This causes many issues with my dashboard linking in DataDog as I now have a high cardinality of spans.

Screenshot 2024-05-28 at 2 57 58 PM

Screenshot 2024-05-28 at 2 59 04 PM

It would be nice if instead we used just a standard name like `<serice_name>:router` or if this was configurable so I could set the span name myself. I would also like the operation name to be a configurable span attribute as I have to go to the supergraph to see this today. This is before the GraphQL parsing so what ever logic is being used to get the operation span name today should just become the attribute value

I could maybe use a custom collector and transform all the spans

bnjjj:
It gives you the ability to override the span name by using custom telemetry with any [selectors](https://www.apollographql.com/docs/router/configuration/telemetry/instrumentation/selectors/) you want just by setting the `otel.name` attribute.

Example:

```yaml
telemetry:
  instrumentation:
    spans:
      router:
        otel.name:
           static: router # Override the span name to router 
```

Fixes #5261 

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [x] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


docs/source/configuration/telemetry/instrumentation/spans.mdx
```diff
@@ -157,14 +157,20 @@ The `mode` option will be defaulted to `spec_compliant` in a future release, and
...
+Here is an example if you want to mark the `router` and `supergraph` span in error if you have a graphql error in the payload and you want to enforce the `router` span name to be `graphql_router`.
 
 ```yaml title="router.yaml"
 telemetry:
   instrumentation:
     spans:
       router:
         attributes:
+          otel.name:
+           static: graphql_router # Override the span name to graphql_router
```
BrynCooke, line 173:
```suggestion
            static: graphql_router # Override the span name to graphql_router 
```

BrynCooke (approved):

docs/source/configuration/telemetry/instrumentation/spans.mdx
```diff
@@ -157,14 +157,20 @@ The `mode` option will be defaulted to `spec_compliant` in a future release, and
 By default spans are marked in error only if the http status code is different than 200. If you want to mark a span in error for other reason you can override the `otel.status_code` attribute which is responsible to mark a span in error or not.
 If it's in error then `otel.status_code` = `error`, if not it will be `ok`.
 
-Here is an example if you want to mark the `router` and `supergraph` span in error if you have a graphql error in the payload.
+## Span name
```
abernix, line 160: This page is about Spans, so maybe "Naming" is enough?

```suggestion
## Naming
```

docs/source/configuration/telemetry/instrumentation/spans.mdx
```diff
@@ -157,14 +157,20 @@ The `mode` option will be defaulted to `spec_compliant` in a future release, and
...
+
```
abernix, line 161: The section below this says how to change the names of the spans, but we don't describe what the default span name is (at least right here in this text).  Could we add some text here saying something like:

```
By default, we will use a span naming convention that aligns with the current [blah].

```

Thoughts?

apollo-router/src/plugins/telemetry/config_new/spans.rs
```diff
@@ -556,6 +566,10 @@ mod test {
         assert!(values
             .iter()
             .any(|key_val| key_val.key == opentelemetry::Key::from_static_str("test")));
+
+        assert!(values.iter().any(|key_val| key_val.key
+            == opentelemetry::Key::from_static_str("otel.name")
+            && key_val.value == opentelemetry::Value::String(String::from("new_name").into())));
```
abernix, line 572: Should there also be a test that asserts that when this isn't the case that the default names will be used?

.changesets/feat_bnjjj_feat_otel_name_attribute.md
```diff
@@ -0,0 +1,16 @@
+### Add the ability to override span name with otel.name attribute ([Issue #5261](https://github.com/apollographql/router/issues/5261))
+
```
abernix, line 2: Thoughts? @shorgi 

```suggestion
### Override tracing span names using custom span selectors ([Issue #5261](https://github.com/apollographql/router/issues/5261))

```

apollo-router/src/plugins/telemetry/config_new/spans.rs
bnjjj: There are integration tests that already exist for that

bnjjj:
@abernix I don't think it will entirely fix https://github.com/apollographql/router/issues/5282 so I won't include it here

docs/source/configuration/telemetry/instrumentation/spans.mdx
bnjjj: Added ! Thanks !

Geal (approved):

PR #5366, opened 2024-06-06
Make QP comparison more lenient and print operation name

SimonSapin:
Following up https://github.com/apollographql/router/pull/5341, the query plan comparison now ignores:
* The schema-aware hash in fetch node, which is only populated on one side of the comparison
* Whitespace in subgraph operation, so irrelevant serializer differences don’t cause a mismatch

Additionally, the log message now includes the operation name.

Example:

`router.yaml`:
```yaml
```

Run:

```
cargo run -- \
  --dev \
  -c router.yaml \
  -s examples/graphql/supergraph-fed2.graphql \
  --log=apollo_router::query_planner=debug
```

Note that the `--log` parameter is less specific than given in https://github.com/apollographql/router/pull/5341 because this PR moves some code to a new Rust module.

Run in another terminal:

```
curl --request POST \
  --header 'content-type: application/json' \
  --url 'http://127.0.0.1:4000/' \
  --data '{"query":"query MyExampleQuery{topProducts {name}}"}'
```

Expected in logs back in the first terminal:

```
DEBUG  JS and Rust query plans match in query `MyExampleQuery`! 🎉
```

The default log level is WARN, which logs in case of mismatch but not in case of match.


apollo-router/src/query_planner/dual_query_planner.rs
```diff
@@ -0,0 +1,299 @@
...
+        match (&self.js_result, &rust_result) {
+            (Err(js_errors), Ok(_)) => {
+                tracing::warn!(
+                    "JS query planner error{operation_desc}: {}",
+                    format_bridge_errors(js_errors)
+                );
+                is_matched = false;
+            }
+            (Ok(_), Err(rust_error)) => {
+                tracing::warn!("Rust query planner error{operation_desc}: {}", rust_error);
```
lrlna, line 114:
```suggestion
                tracing::warn!("Rust query planner error in {operation_desc}: {}", rust_error);
```

apollo-router/src/query_planner/dual_query_planner.rs
```diff
@@ -0,0 +1,299 @@
...
+            (Err(_), Err(_)) => {
+                is_matched = true;
+            }
+
+            (Ok(js_plan), Ok(rust_plan)) => {
+                let js_root_node = &js_plan.query_plan.node;
+                let rust_root_node = convert_root_query_plan_node(rust_plan);
+                is_matched = opt_plan_node_matches(js_root_node, &rust_root_node);
+                if is_matched {
+                    tracing::debug!("JS and Rust query plans match{operation_desc}! 🎉");
```
lrlna, line 126:
```suggestion
                    tracing::debug!("JS and Rust query plans match in {operation_desc}! 🎉");
```

apollo-router/src/query_planner/dual_query_planner.rs
```diff
@@ -0,0 +1,299 @@
...
+        && operation_matches(operation, &other.operation)
+}
+
+fn operation_matches(this: &SubgraphOperation, other: &SubgraphOperation) -> bool {
+    operation_without_whitespace(this) == operation_without_whitespace(other)
+}
+
+fn operation_without_whitespace(op: &SubgraphOperation) -> String {
+    op.as_serialized().replace([' ', '\n'], "")
+}
```
lrlna, line 207: i have a feeling we are going to be tweaking these as we go, but it seems like this is good to go from the current results in the corpus 🙏

lrlna (approved):
Providing printing `operation_name` in warn is good, I say let's merge it and keep going 💪

PR #5369, opened 2024-06-06
Fix various bugs in type-explosion logic for implementation fields

sachindshinde:
This PR:
- Makes various bugfixes to query graph methods.
  - The subgraph map was being directly accessed via `QueryGraph.sources`. This generally isn't desirable, as this map contains an entry for `FEDERATED_GRAPH_ROOT_SOURCE` (a dummy source, which is a carryover from the JS codebase's building logic).
    - The field has been returned to being private, and callsites have been updated to use the `sources()` method.
  - The `sources()` method has been updated to filter out the `FEDERATED_GRAPH_ROOT_SOURCE` dummy source.
  - The `sources()` method has been renamed to `source_schemas()`, to make it more clear it will give schemas.
  - The `schema()` method doesn't return anything useful for a federated query graph (it specifically returns a dummy schema, which is again a carryover of JS codebase behavior). This method is mainly used during query graph building and for API schema query graphs, so I've made it private for now.
  - The `nodes_for_type()` method was mistakenly looking at entries for subgraph query graphs instead of the federated query graph.
- Makes various bugfixes to `has_an_entity_implementation_with_shareable_field()`.
  - The initial interface field was being fetched on the wrong schema (the `QueryGraph.schema()`) and ignoring the source argument.
  - A shareable directive application was being created instead of fetching the shareable directive definition.
  - Variable shadowing/reusing simple variables names led to three separate bugs. I've accordingly made the variable names more descriptive. Specifically:
    - `nodes_for_type()` was being called on `ty`, but the intended value had been shadowed.
    - `node_ty.get()` was being called on `schema`, but the code had forgotten to shadow it. 
    - `other_fields` was mistakenly being used in the field subset check, when it should be the field's base type's fields.
  - The field subset check was inverted (we should be checking whether the other subgraph's fields are a subset of this subgraph's).
  - Some logic was updated to use existing helper methods.
- Adds some helper methods to `position.rs` (a way to get field positions from object/interface types, and `try_get()`s for object/interface fields).
- Updates tests.
  - Tests in `handles_fragments_with_directive_conditions.rs` now succeed.
  - The test `it_works_with_type_condition_even_for_types_only_reachable_by_the_at_provides()` in `provides.rs` now succeeds.
  - 7 tests in `interface_object.rs` now fail for different reasons. I've created FED-291 to track it.


apollo-federation/src/schema/position.rs
```diff
@@ -1808,6 +1833,23 @@ impl ObjectTypeDefinitionPosition {
...
     }
 
+    // TODO: Once the new lifetime capturing rules for return position impl trait (RPIT) land in
+    // Rust edition 2024, we will no longer need the "captures" trick here, as noted in
+    // https://rust-lang.github.io/rfcs/3498-lifetime-capture-rules-2024.html
+    pub(crate) fn fields<'a>(
+        &'a self,
+        schema: &'a Schema,
+    ) -> Result<
+        impl Iterator<Item = ObjectFieldDefinitionPosition> + Captures<&'a ()>,
```
goto-bus-stop, line 1843: Does this not work here?
```suggestion
        impl Iterator<Item = ObjectFieldDefinitionPosition> + 'a,
```

apollo-federation/src/query_plan/query_planner.rs
```diff
@@ -341,7 +348,7 @@ impl QueryPlanner {
             // belong to no subgraphs and use a special source named '_'. So we skip that "fake" source.
             let mut subgraphs = self
                 .federated_query_graph
-                .sources()
+                .source_schemas()
                 .filter(|&(name, _schema)| name != "_");
```
goto-bus-stop, line 352: This is already filtered out by `.source_schemas()` now.

goto-bus-stop:
Would it make more sense to remove the `_` source from the `.sources` map entirely? We can special-case it in `.schema_by_source()` if necessary?

`.subgraph_sources()` now creates a map on every call which is not ideal. If we keep the `_` entry on a separate field instead of in the sources map we can go back to just returning the `&IndexMap`.

apollo-federation/src/query_graph/graph_path.rs
```diff
@@ -2249,68 +2248,79 @@ impl OpGraphPath {
...
+                let Some(node_field_base_type_pos): Option<
+                    ObjectOrInterfaceTypeDefinitionPosition,
+                > = node_field_base_type_pos.try_into().ok() else {
+                    // Similar to above, we have a genuine difference.
+                    return Ok(true);
+                };
+
+                if !node_field_base_type_pos
+                    .fields(node_schema)?
+                    .all(|f| implementing_field_base_type_fields.contains(f.field_name()))
```
goto-bus-stop, line 2316:
```suggestion
                    .all(|f| implementing_field_base_type.fields.contains_key(f.field_name()))
```
I think collecting the keys separately is not neccessary here.

apollo-federation/src/schema/position.rs
sachindshinde: @goto-bus-stop 
That was the thing I jumped to at first, but [Rust RFC 3498](https://rust-lang.github.io/rfcs/3498-lifetime-capture-rules-2024.html#the-outlives-trick) noted that this is the opposite lifetime constraint we want.

apollo-federation/src/query_plan/query_planner.rs
sachindshinde: Good catch, removed the extra `filter` in the latest commit

apollo-federation/src/query_graph/graph_path.rs
sachindshinde: It's indeed unnecessary, removed the extraneous hash set.

sachindshinde:
@goto-bus-stop 
> Would it make more sense to remove the `_` source from the `.sources` map entirely? We can special-case it in `.schema_by_source()` if necessary?
> 
> `.subgraph_sources()` now creates a map on every call which is not ideal. If we keep the `_` entry on a separate field instead of in the sources map we can go back to just returning the `&IndexMap`.

I ended up making a copy without the dummy source at `QueryGraph.subgraphs_by_name`, and renaming methods appropriately.

apollo-federation/src/query_plan/query_planner.rs
```diff
@@ -339,10 +339,7 @@ impl QueryPlanner {
         if self.config.debug.bypass_planner_for_single_subgraph {
             // A federated query graph always have 1 more sources than there is subgraph, because the root vertices
             // belong to no subgraphs and use a special source named '_'. So we skip that "fake" source.
```
goto-bus-stop, line 341:
```suggestion
```

goto-bus-stop (approved):
It feels like a big hack that we have this `_` schema and I'd be interested to see if we can just get rid of it in the future. But 🤷🏻‍♀️ 

I didn't check the implementation of `has_an_entity_implementation_etc` against the JS so hopefully the tests are proof enough ;)

PR #5370, opened 2024-06-07
 FED-280 fix: fixed SelectionSet::normalize to merge expanded selections

duckki:
- Previously, overlapping selections could overwrite each other.
- Also, fixed InlineFragmentSelection::normalize to rebase whenever necessary.
- ~Also, fixed two spots to ignore rebase_on errors when they were supposed to be ignored.~

These were fixed together to avoid too much regressions.

Fixes
- https://apollographql.atlassian.net/browse/FED-280
- https://apollographql.atlassian.net/browse/FED-251 (partially)
- https://apollographql.atlassian.net/browse/FED-277
- ~https://apollographql.atlassian.net/browse/FED-278~

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [x] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-federation/src/operation/rebase.rs
```diff
@@ -241,7 +241,10 @@ impl FieldSelection {
             return Ok(Some(Selection::from(self.clone())));
         }
 
-        let Some(rebased) = self.field.rebase_on(parent_type, schema, error_handling)? else {
+        let Some(rebased) =
+            self.field
+                .rebase_on(parent_type, schema, RebaseErrorHandlingOption::IgnoreError)?
```
dariuszkuc, line 246: this is incorrect -> we should pass the original `error_handling` option -> if we cannot rebase and error handling is set to ignore we return `None` otherwise we return an error (see https://github.com/apollographql/federation/blob/main/internals-js/src/operations.ts#L3096)

apollo-federation/src/operation/rebase.rs
```diff
@@ -548,9 +551,11 @@ impl InlineFragmentSelection {
             return Ok(Some(Selection::from(self.clone())));
         }
 
-        let Some(rebased_fragment) =
-            self.inline_fragment
-                .rebase_on(parent_type, schema, error_handling)?
+        let Some(rebased_fragment) = self.inline_fragment.rebase_on(
```
dariuszkuc, line 554: same comment as above

apollo-federation/src/operation/rebase.rs
```diff
@@ -667,12 +672,13 @@ impl NamedFragments {
...
             {
-                if let Ok(mut rebased_selection) = fragment.selection_set.rebase_on(
+                let mut rebased_selection = fragment.selection_set.rebase_on(
                     &rebased_type,
                     &rebased_fragments,
                     schema,
                     RebaseErrorHandlingOption::IgnoreError,
-                ) {
+                )?;
+                if !rebased_selection.is_empty() {
```
dariuszkuc, line 681: this is already handled by `is_selection_set_worth_using` below

duckki: I reverted this change for now, since it's distracting.

apollo-federation/src/operation/mod.rs
```diff
@@ -3292,20 +3342,39 @@ impl FieldSelection {
...
+    }
+
     fn normalize(
         &self,
         parent_type: &CompositeTypeDefinitionPosition,
         named_fragments: &NamedFragments,
         schema: &ValidFederationSchema,
         option: NormalizeSelectionOption,
     ) -> Result<Option<SelectionOrSet>, FederationError> {
+        let field_position = if self.field.parent_type_position() == *parent_type {
```
dariuszkuc, line 3359: in JS side of things those elements hold references to the schema but on RS those are just type names so we need to explicitly check for the schema equality as well

apollo-federation/src/operation/mod.rs
```diff
@@ -2137,7 +2138,23 @@ impl SelectionSet {
             type_position: self.type_position.clone(),
             selections: Arc::new(SelectionMap::new()),
         };
-        expanded.merge_selections_into(expanded_selections.iter())?;
+        for selection in expanded_selections {
```
dariuszkuc, line 2141: couldn't we just `merge_selections_into` and then normalize whole `expanded` set instead of normalizing and merging individual selections?

duckki: Good point. Actually, normalization here is not necessary after all. I'll remove unnecessary changes.

lrlna (approved):

apollo-federation/src/operation/mod.rs
```diff
@@ -2820,27 +2821,28 @@ impl SelectionSet {
         schema: &ValidFederationSchema,
         option: NormalizeSelectionOption,
     ) -> Result<SelectionSet, FederationError> {
-        let mut normalized_selection_map = SelectionMap::new();
+        let mut normalized_selections = Self {
+            schema: schema.clone(),
+            type_position: parent_type.clone(),
```
dariuszkuc, line 2826: good catch! this was done in the JS `lazyMap` itself and not the normalize method

dariuszkuc:
Instead of all the rebasing in individual selections -> can we move it level up to the `Selection::normalize` instead?

apollo-federation/src/operation/mod.rs
```diff
@@ -3347,20 +3363,40 @@ impl FieldSelection {
...
         option: NormalizeSelectionOption,
     ) -> Result<Option<SelectionOrSet>, FederationError> {
+        let field_position =
+            if self.field.schema() == schema && self.field.parent_type_position() == *parent_type {
+                self.field.data().field_position.clone()
+            } else {
+                parent_type.field(self.field.data().name().clone())?
+            };
+
+        let field_element = if self.field.data().field_position == field_position {
```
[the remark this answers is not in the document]
duckki: Yeah, it's so easy to miss. I think we could use a function that checks both. A refactoring idea for later.

duckki:
> Instead of all the rebasing in individual selections -> can we move it level up to the `Selection::normalize` instead?

We are mirroring JS code's `normalizeKnowingItIntersects` workflow, which calls `rebaseOn` here and there. But, in Rust, `makeSelection` (which calls `rebaseOn`) is split into `make_selection` and `add_selection_set`. Otherwise, I think we are mirroring JS code with a good fidelity now. Having said that, there might be opportunities for simplification.

PR #5376, opened 2024-06-07
do not convert the opentelemetry value to a string then an i64

Geal:
this gives a 1-2% performance improvement when using custom instruments

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


tninesling (approved):

BrynCooke (commented):
I think there are several instances of this pattern in this file.

Geal:
@BrynCooke right, there were a few more in there

apollo-router/src/plugins/telemetry/config_new/instruments.rs
```diff
@@ -1186,13 +1186,13 @@ where
         }
 
         if let Some(selected_value) = inner.selector.as_ref().and_then(|s| s.on_request(request)) {
+            let value = match selected_value {
+                opentelemetry::Value::I64(i) => Some(i),
+                _ => None,
+            };
```
bnjjj, line 1192: I'm not sure it would fit what we want to achieve today. If someone would like to provide something that's coming from a header, like a value and would like to use it as an instrument it's now possible but won't be anymore with this change. @BrynCooke don't you think we should keep this ability ?

Geal: can't we parse when we extract from the header?

bnjjj: Header is a selector and so we don't know if this selector will be used for an instrument or for an attribute as a string. I think if we only try to parse it if it's a string it's still an optimization instead of always convert it to a string and then parse it

Geal: Maybe if I provide a `as_i64` method that would directly extract the value if it is a `opentelemetry::Value::I64`, floor it if it is a `f64` and parse if it is a string?

bnjjj: Yes LGTM

bnjjj (approved):

PR #5378, opened 2024-06-07
Fix bugs in selection set locality checking and inconsistent abstract type checking

sachindshinde:
This PR:
- Makes bugfixes to `selection_set_is_fully_local_from_all_nodes()`.
  - It was only checking for inconsistent abstract types at top-level, when it should be checking every element.
    - I've added a method called `any_element()` to help facilitate this (this was called `some()` in the JS codebase).
  - The call to `can_rebase_on()` wasn't checking whether the selection set could be rebased on the node's schema.
  - As a matter of efficiency, we should early return when the node's `has_reachable_cross_subgraph_edges` is `true`, as it's much faster to check than rebasing. I've accordingly rearranged the checks so it's first (mirroring the JS codebase).
  - Similarly for efficiency, the JS codebase had ordered the check for inconsistent abstract runtime types until after rebasing. I've accordingly changed the code to do the same.
- Makes a bugfix to the logic that computes inconsistent abstract runtime types.
  - The `is_inconsistent()` function in particular had inverted its output, and needed a `!`.
- Makes a bugfix to `can_rebase_on()`.
  - This method should accept a schema to rebase upon. (It's the schema containing the provided type.) It was assuming that it was the selection set's schema, but that wasn't the case for the two existing callsites of `can_rebase_on()`.
- Makes bugfixes to `type_at_path()`.
  - While reading the other callsite of `can_rebase_on()` (which is currently `is_node_unneeded()`), I was looking for the appropriate schema, and I noticed `type_at_path()` should have been taking in a schema. After reading its code, I noticed it was using the wrong schema (the schemas of the operation elements). It should only be using the schema of the passed-in type, and the schemas of the operation elements should be ignored. I've updated the code accordingly (it was also missing an assert in the case of type conditions).
- Updates tests.
  - 2 tests in `merged_abstract_types_handling.rs` now succeed.
  - 1 test in `interface_type_explosion.rs` now succeeds, specifically `handles_non_matching_value_types_under_interface_field()`.
  - 1 test in `introspection_typename_handling.rs` now fails for a different reason, specifically `it_does_not_needlessly_consider_options_for_typename()`.

EDIT: For reasons that I don't understand, the original code of this PR repeatedly failed CI specifically on macOS, specifically for 46 router tests with the error message `'GraphQL endpoint exposed' not detected in logs` (most of them integration tests). These tests continued to fail, despite at least 10 re-runs. After stepping through the changes one at a time, moving the inconsistent abstract type check to the end of the loop is what triggered the issue. It appears that moving the inconsistent abstract type check into a closure somehow fixes it. I'm going to check in a later PR whether I can remove the wrapping closure, with the hopes it might be some kind of strange CI caching issue.

### DIFF
--- a/apollo-federation/src/operation/mod.rs
+++ b/apollo-federation/src/operation/mod.rs
@@ -830,6 +830,22 @@ impl Selection {
             Ok(self.clone())
         }
     }
+
+    pub(crate) fn any_element(
+        &self,
+        parent_type_position: CompositeTypeDefinitionPosition,
+        predicate: &mut impl FnMut(OpPathElement) -> Result<bool, FederationError>,
+    ) -> Result<bool, FederationError> {
+        match self {
+            Selection::Field(field_selection) => field_selection.any_element(predicate),
+            Selection::InlineFragment(inline_fragment_selection) => {
+                inline_fragment_selection.any_element(predicate)
+            }
+            Selection::FragmentSpread(fragment_spread_selection) => {
+                fragment_spread_selection.any_element(parent_type_position, predicate)
+            }
+        }
+    }
 }
 
 impl From<FieldSelection> for Selection {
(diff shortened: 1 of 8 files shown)


dariuszkuc (approved):

duckki (approved):
👍

PR #5379, opened 2024-06-07
Rust QP: serialize subgraph operations without line breaks or indentation

SimonSapin:
The default apollo-compiler serialization includes is indented like this:

```graphql
{
  topProducts {
    name
  }
}
```

This change the Rust query planner to use the "no indent" style for subgraph operations in fetch nodes, which looks like this:

```graphql
{ topProducts { name } }
```

There is no line break or indentation, but some whitespace is still used for readability so it still differs from the JS query planner which minimizes whitespace: `{topProducts{name}}`.

This PR merely reduces the amount of bytes on the network (or amount of work given to compression). Before removing Deno entirely we’ll likely want serialization to match that of JS exactly, because tests using `MockSubgraph` rely on a mapping of exact subgraph query string to response to supply.

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/query_planner/bridge_query_planner.rs
+++ b/apollo-router/src/query_planner/bridge_query_planner.rs
@@ -1814,12 +1814,7 @@ mod tests {
 
         let subgraph_queries = subgraph_queries2.lock().await;
         insta::assert_snapshot!(*subgraph_queries, @r###"
-        {
-          topProducts {
-            name
-          }
-        }
-
+        { topProducts { name } }
         "###)
     }
 }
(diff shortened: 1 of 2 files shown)


lrlna (approved):

lrlna:
does look like a test_both_mode test fails though, but only on mac? weird

SimonSapin:
Weird that it’s on specifically on tests semi-related to this PR, but we’ve seen `Test aborted with signal SIGSEGV (code 11)` happen today on mac CI on other PRs.

I cannot reproduce it on my mac.

lrlna:
Really weird then o.o

PR #5380, opened 2024-06-07
add some tests for experimental_query_planner_mode unsupported feature rejections, 

lrlna:
A follow up to #5335

Just to ensure we are not rejecting acceptable supergraphs, along with the correct directive name used for in progressive overrides search.

### DIFF
--- a/apollo-router/src/router_factory.rs
+++ b/apollo-router/src/router_factory.rs
@@ -849,6 +849,7 @@ mod test {
     use tower_http::BoxError;
 
     use crate::configuration::Configuration;
+    use crate::configuration::QueryPlannerMode;
     use crate::plugin::Plugin;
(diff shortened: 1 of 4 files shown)


dariuszkuc (approved):

apollo-router/src/router_factory.rs
```diff
@@ -991,4 +994,122 @@ mod test {
             "8e2021d131b23684671c3b85f82dfca836908c6a541bbd5c3772c66e7f8429d8"
         );
     }
+
+    #[test]
+    fn test_can_use_with_experimental_query_planner() {
```
dariuszkuc, line 999: nit: should this test be split out into individual ones?

dariuszkuc (approved):

PR #5383, opened 2024-06-08
fix(federation): normalize rebased fragment definitions

dariuszkuc:
JS code applies [normalization](https://github.com/apollographql/federation/blob/main/internals-js/src/operations.ts#L1462) on the rebased fragment definitions that eliminates the unnecessary inline spreads.

Fixes #[FED-261](https://apollographql.atlassian.net/browse/FED-261)

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[FED-261]: https://apollographql.atlassian.net/browse/FED-261?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apollo-federation/src/operation/optimize.rs
+++ b/apollo-federation/src/operation/optimize.rs
@@ -847,7 +847,9 @@ impl Selection {
                     // Expand the fragment
                     let expanded_sub_selections =
                         fragment.selection_set.retain_fragments(fragments_to_keep)?;
-                    if *parent_type == fragment.spread.data().type_condition_position.clone() && fragment.spread.data().directives.is_empty() {
+                    if *parent_type == fragment.spread.data().type_condition_position.clone()
+                        && fragment.spread.data().directives.is_empty()
+                    {
                         // The fragment is of the same type as the parent, so we can just use
                         // the expanded sub-selections directly.
                         Ok(expanded_sub_selections.into())
@@ -994,7 +996,7 @@ impl NamedFragments {
                     &fragment.selection_set.type_position,
                     &fragments_to_keep,
                     &fragment.schema,
-                    NormalizeSelectionOption::NormalizeRecursively
+                    NormalizeSelectionOption::NormalizeRecursively,
                 )?;
         }
 
--- a/apollo-federation/src/operation/optimize.rs
+++ b/apollo-federation/src/operation/optimize.rs
@@ -847,7 +847,7 @@ impl Selection {
                     // Expand the fragment
                     let expanded_sub_selections =
                         fragment.selection_set.retain_fragments(fragments_to_keep)?;
-                    if *parent_type == fragment.spread.data().type_condition_position.clone()
+                    if *parent_type == fragment.spread.data().type_condition_position
                         && fragment.spread.data().directives.is_empty()
                     {
                         // The fragment is of the same type as the parent, so we can just use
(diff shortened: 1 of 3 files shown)


lrlna (approved):

apollo-federation/src/operation/optimize.rs
```diff
@@ -989,7 +991,13 @@ impl NamedFragments {
         for (_, fragment) in self.iter_mut() {
             Node::make_mut(fragment).selection_set = fragment
                 .selection_set
-                .retain_fragments(&fragments_to_keep)?;
+                .retain_fragments(&fragments_to_keep)?
+                .normalize(
+                    &fragment.selection_set.type_position,
+                    &fragments_to_keep,
+                    &fragment.schema,
+                    NormalizeSelectionOption::NormalizeRecursively,
```
goto-bus-stop, line 999: Is recursive normalization necessary here?

JS also does it recursively, but I wonder why we even have this option. Neither JS nor Rust QP use `recursive: false` / `NormalizeSingleSelection` anywhere. Should we remove it?

apollo-federation/src/operation/optimize.rs
```diff
@@ -847,7 +847,9 @@ impl Selection {
                     // Expand the fragment
                     let expanded_sub_selections =
                         fragment.selection_set.retain_fragments(fragments_to_keep)?;
-                    if *parent_type == fragment.spread.data().type_condition_position {
+                    if *parent_type == fragment.spread.data().type_condition_position.clone()
```
goto-bus-stop, line 850:
```suggestion
                    if *parent_type == fragment.spread.data().type_condition_position
```

This clone is weird. We are only using the value inline immediately so it can't be necessary to clone if we are using the right reference type.

apollo-federation/src/operation/optimize.rs
dariuszkuc: Indeed! It is actually pretty weird as in JS code we never actually specify it so it defaults to `undefined` which is then always coalesced to true (using `recursive ?? true` check). I'll open a follow up item to clean it up.

dariuszkuc:
>flipped order in parallel fetches is expected i guess.

Yeah so technically that order doesn't matter but IMHO we probably should look into this in the future as it will be harder to compare in the future. Since we already merged similar "fix" I thought it was fine.

apollo-federation/src/operation/optimize.rs
duckki: Hmm, there is a `normalize` call after `retain_fragments`
```rust
        let reduced_selection_set = selection_set.retain_fragments(self)?;

        // Expanding fragments could create some "inefficiencies" that we wouldn't have if we
        // hadn't re-optimized the fragments to de-optimize it later, so we do a final "normalize"
        // pass to remove those.
        reduced_selection_set.normalize(
            &reduced_selection_set.type_position,
            self,
            &selection_set.schema,
            NormalizeSelectionOption::NormalizeRecursively,
        )
```
I wonder why that didn't handle this.

duckki: Or, we need to normalize expanded fragment "definitions"?
In `fn reduce()`:
```rust
        let fragments_to_keep = self.clone();
        for (_, fragment) in self.iter_mut() {
            Node::make_mut(fragment).selection_set = fragment
                .selection_set
                .retain_fragments(&fragments_to_keep)?;
             // <= also normalize here?
        }
```

dariuszkuc: yes it was normalizing selection set but not the fragments (hence the extra normalize after retain fragments)

PR #5384, opened 2024-06-08
Remove what should be unnecessary wrapping closure from `selection_set_is_fully_local_from_all_nodes()`

sachindshinde:
This PR is a followup to https://github.com/apollographql/router/pull/5378 . In that PR, removing this closure triggered 46 router tests to fail specifically on macOS with the message `'GraphQL endpoint exposed' not detected in logs` (and this appeared to be consistent across re-runs). My hope is a separate PR may change this. If not, this PR acts as a repro of this specific issue.


apollo-federation/src/query_plan/query_planning_traversal.rs
```diff
@@ -472,29 +472,7 @@ impl<'a: 'b, 'b> QueryPlanningTraversal<'a, 'b> {
         //   later condition is that `selection` is originally a supergraph selection, but that we're looking to apply "as-is" to a subgraph.
         //   But suppose it has a `... on I` where `I` is an interface. Then it's possible that `I` includes "more" types in the supergraph
         //   than in the subgraph, and so we might have to type-explode it. If so, we cannot use the selection "as-is".
-        let mut has_inconsistent_abstract_types: Option<bool> = None;
```
goto-bus-stop, line 475: We could replace this with `let mut has_inconsistent_abstract_types: OnceCell<bool> = OnceCell::new()`, and then execute the condition with `has_inconsistent_abstract_types.get_or_init()`, which is exactly equivalent to what the closure code is doing. (and imo a little better than having a separate bool to track whether the code has been executed)

lrlna:
@sachindshinde is this still something we need to get in to `dev`? or was this simply to please the CI?

goto-bus-stop:
Actually, can we not just do this once unconditionally at the start of the function? It doesn't look *that* hot...

lrlna:
what do we want to do with this PR @goto-bus-stop @sachindshinde? shall we close it entirely?

goto-bus-stop:
I would close it. The wrapping closure will be inlined by the compiler, so there is no performance impact, and I think it does benefit the readability a bit.

lrlna:
Ok, closing this then.

PR #5386, opened 2024-06-10
Datadog exporter resource behavior and mapping

### LINKED ISSUE #5282: Include `request` span in spec_compliant mode

bnjjj:
### Datadog exporter resource behavior and mapping ([Issue #5282](https://github.com/apollographql/router/issues/5282))

Users of the Datadog trace exporter may have noticed that span and resource naming is not as expected. 
Unlike other APMs, Datadog expects static span names, and then uses resource mapping to provide additional context.

The default behavior of the Datadog exporter has now been changed to support this and give a better user experience.

```yaml
telemetry:
  exporters:
    tracing:
      datadog:
        enabled: true
        # Enables resource mapping, previously disabled by default, but now enabled.
        enable_span_mapping: true
        # Enables fixed span names, defaults to true.
        fixed_span_names: true

  instrumentation:
    spans:
      mode: spec_compliant

```

The following default resource mappings are applied:

| OpenTelemetry Span Name | Datadog Span Operation Name |
|-------------------------|-----------------------------|
| `request`               | `http.route`                |
| `router`                | `http.route`                |
| `supergraph`            | `graphql.operation.name`    |
| `query_planning`        | `graphql.operation.name`    |
| `subgraph`              | `subgraph.name`             |
| `subgraph_request`      | `graphql.operation.name`    |
| `http_request`          | `http.route`                |

You can override the default resource mappings by specifying the `resource_mapping` configuration:

```yaml
  exporters:
    tracing:
      datadog:
        enabled: true
        resource_mapping:
          # Use `my.span.attribute` as the resource name for the `router` span
          router: "my.span.attribute"
```
Fixes #5282 

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/plugins/telemetry/tracing/datadog.rs
```diff
@@ -27,6 +27,7 @@ lazy_static! {
     static ref SPAN_RESOURCE_NAME_ATTRIBUTE_MAPPING: HashMap<&'static str, &'static str> = {
         let mut map = HashMap::new();
         map.insert("request", "http.route");
+        map.insert("router", "http.route");
```
Samjin, line 30: I think based on [this latest comment](https://github.com/apollographql/router/issues/5282#issuecomment-2168849672), we have to `router` as the HTTP and Service entry span. I think this change could go in.

Samjin (approved):

bnjjj:
@BrynCooke Could you see any issues by doing this ? Like breaking existing tracing for our customers on DD ?

BrynCooke:
I pushed some WIP code to add configuration and integration tests to show that span mapping is working.

apollo-router/src/plugins/telemetry/tracing/datadog.rs
```diff
@@ -23,19 +22,22 @@ use crate::plugins::telemetry::tracing::BatchProcessorConfig;
...
-        map.insert("supergraph", "graphql.operation.name");
-        map.insert("query_planning", "graphql.operation.name");
-        map.insert("subgraph", "subgraph.name");
-        map.insert("subgraph_request", "graphql.operation.name");
-        map
-    };
-    static ref DEFAULT_ENDPOINT: Uri = Uri::from_static("http://127.0.0.1:8126");
+fn default_resource_mappings() -> HashMap<String, String> {
+    let mut map = HashMap::new();
+    map.insert("request", "http.route");
```
Samjin, line 27: Do you still need request span for `deprecated` mode if router is doing the same?

apollo-router/src/plugins/telemetry/mod.rs
```diff
@@ -556,17 +558,18 @@ impl Plugin for Telemetry {
...
+                        TraceIdFormat::Decimal => format!("{}", trace_id),
+                        TraceIdFormat::Datadog => trace_id.to_datadog()
                     };
 
                     HeaderValue::from_str(&id).ok()
                 };
                 if let (Some(header_name), Some(trace_id)) = (
                     expose_trace_id_header,
-                    TraceId::current().and_then(format_id),
+                    trace_id().and_then(format_id),
```
BrynCooke, line 572: @bnjjj Can you make sure you are happy with this change?

BrynCooke:
OK, here's how things are shaping up:

We have a new section in the datadog exporter called `resource_mapping`.
It allows you to specify the span attribute that is used as the resource name when viewing in datadog spans.

The key in the resource mapping is the original span name before any renaming for otel or via yaml `otel.name`.
For example, in spec compliant mode the key for the root span is always `router`. 

The value in the resource mapping is the attribute on the span is to be used for the resource name.

Users may with to add graphql.operation.name to the router span to allow this to be used as the resource.


```
telemetry:
  exporters:
    tracing:
      common:
        service_name: router
        resource:
          env: local
      datadog:
        enabled: true
        enable_span_mapping: true
        batch_processor:
          scheduled_delay: 100ms
        resource_mapping:
          router: graphql.operation.name 
          supergraph: override.name 

  instrumentation:
    spans:
      mode: spec_compliant
      router:
        attributes:
          graphql.operation.name:
            operation_name: string
      supergraph:
        attributes:
          override.name: overridden
```

@Samjin Would be great to get your feedback on this.

.changesets/fix_bnjjj_fix_5282.md
bnjjj, line 1: Could you update it with your changes @BrynCooke ?

apollo-router/src/plugins/telemetry/mod.rs
```diff
@@ -337,11 +339,17 @@ impl Plugin for Telemetry {
                             span.record("graphql.operation.name", operation_name);
                         }
                         match (&operation_kind, &operation_name) {
-                            (Ok(Some(kind)), Ok(Some(name))) => {
-                                span.record("otel.name", format!("{kind} {name}"))
+                            (Ok(Some(kind)), Ok(Some(name))) => span.set_span_dyn_attribute(
+                                "otel.name".into(),
```
bnjjj, line 343: NIT: could be a constant

apollo-router/src/plugins/telemetry/mod.rs
bnjjj: yes

apollo-router/src/plugins/telemetry/tracing/datadog.rs
```diff
@@ -19,40 +18,72 @@ use crate::plugins::telemetry::config::GenericWith;
...
-        map.insert("request", "http.route");
-        map.insert("supergraph", "graphql.operation.name");
-        map.insert("query_planning", "graphql.operation.name");
-        map.insert("subgraph", "subgraph.name");
-        map.insert("subgraph_request", "graphql.operation.name");
-        map
-    };
-    static ref DEFAULT_ENDPOINT: Uri = Uri::from_static("http://127.0.0.1:8126");
+fn default_resource_mappings() -> HashMap<String, String> {
+    let mut map = HashMap::new();
```
bnjjj, line 27:
```suggestion
    let mut map = HashMap::with_capacity(7);
```

apollo-router/src/plugins/telemetry/tracing/datadog.rs
```diff
@@ -19,40 +18,72 @@ use crate::plugins::telemetry::config::GenericWith;
...
+const DEFAULT_ENDPOINT: &str = "http://127.0.0.1:8126";
+
+const BUILT_IN_SPAN_NAMES: [&str; 7] = [
+    "request",
+    "router",
+    "supergraph",
+    "subgraph",
+    "subgraph_request",
+    "http_request",
+    "query_planning",
```
bnjjj, line 49: I'm wondering if we could use the existing constant for span names ?

apollo-router/tests/integration/telemetry/mod.rs
```diff
@@ -1,5 +1,9 @@
+#[cfg(any(not(feature = "ci"), all(target_arch = "x86_64", target_os = "linux")))]
```
bnjjj, line 1: Is it related to the collector jaeger thing ? If yes, why not using collector in every cases ?

bnjjj (commented):
I also would like to know what's the behavior if we're using `telemetry.instrumentations.spans` and overriding the span name with `otel.name` attribute ? Could we have a test for this ?

apollo-router/tests/integration/telemetry/mod.rs
BrynCooke: It's just we weren't running these tests on OSX before.

apollo-router/src/plugins/telemetry/mod.rs
BrynCooke: [fbda2bf](https://github.com/apollographql/router/pull/5386/commits/fbda2bfcbb3120f8e0fd74af360fbad140174b7a)

bnjjj:
I can't approve it because I opened that PR but LGTM. Thanks for the refactor around consts it's really nice ! We should do this more often!

BrynCooke (approved):

PR #5392, opened 2024-06-10
add  selector for router service in custom telemetry

bnjjj:
Instead of having to access to the operation_name using the response_context at the router service, we now provide a selector for operation name at the router service in instrumentations.

example:

```yaml
telemetry:
  instrumentation:
    instruments:
      router:
        http.server.request.duration:
          attributes:
            graphql.operation.name:
              operation_name: string
```

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [x] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


docs/source/configuration/telemetry/instrumentation/selectors.mdx
```diff
@@ -32,7 +32,8 @@ The router service is the initial entrypoint for all requests. It is HTTP centri
 
 | Selector           | Defaultable | Values                                    | Description                          |
 |--------------------|-------------|-------------------------------------------|--------------------------------------|
-| `trace_id`         | Yes         | `open_telemetry`\|`datadog`              | The trace ID                         |
+| `trace_id`         | Yes         | `open_telemetry`\|`datadog`               | The trace ID                         |
+| `operation_name`   | Yes         | `string`|`hash`                           | The operation name from the query    |
 | `studio_operation_id` | Yes      | `true`|`false`                            | The Apollo Studio operation id       |
```
goto-bus-stop, line 37: Unrelated but does this make sense? Operation ID as a boolean?

bnjjj: The values listed here are not the value that the selector will have. It's just the configuration value, you can either enable it (`true`)  or disable it (`false`).

BrynCooke (approved):

PR #5394, opened 2024-06-10
Reconcile `dev` after merge to `main` for v1.48.1

abernix:
Follow-up to the v1.48.1 being officially released, bringing version bumps and changelog updates into the `dev` branch.

PR #5395, opened 2024-06-10
Reconcile `dev` after merge to `main` for v1.48.1

abernix:
Supersedes my attempt in https://github.com/apollographql/router/pull/5394, follows #5349 and https://github.com/apollographql/router/releases/tag/v1.48.1


goto-bus-stop (commented):
Hmm, why are the commit hashes on this different than the equivalent commits already in `dev`? That will add duplicates in the commit history 🤔 

I guess it's not an actual problem, but it's a bit fishy. Do we have a rebasing step in the release process or something?

goto-bus-stop:
Rubber ducked it. I guess it makes sense because only *some* commits are added in this release, which means they necessarily have a different parent than the original.

goto-bus-stop (approved):

abernix:
@goto-bus-stop Fair shout and good observation — we **did** actually cherry pick more things on this release than usual (we usually have more clean forks from the main trunk) as `dev` itself had other more substantial changes that are targeted for 1.49.0 than we wanted for a patch release.  Thanks for paying attentiong.

PR #5402, opened 2024-06-11
Fix various bugs in `can_rebase_on()` and `can_add_to()`

sachindshinde:
This PR fixes a few separate issues in `can_rebase_on()` and `can_add_to()`:
- In `FieldSelection.can_add_to()`, when recursing into the selection set, the wrong type position was passed in (it should be the base type of the field's return type, not the field's parent type).
- In `Field.type_if_added_to()`, the result was mistakenly being converted to a `CompositeTypeDefinitionPosition` when it should be an `OutputTypeDefinitionPosition`.
- In `Field.can_rebase_on()`, the wrong schema was being passed into the call to `is_interface_object()` (it should be the field's, based on the JS code).
- In `InlineFragmentData.can_rebase_on()`, the wrong schema was being passed in when getting the casted type and executing `runtime_types_intersect()` (it should be the schema of the given parent type, based on the JS code).
- In `FieldSelection.can_add_to()` and `InlineFragmentSelection.can_add_to()`, the clause before recursing into the selection set (if it exists) should check both the schema and type position, not just the type position.
- Methods have been converted to return `Result<bool, FederationError>`, to more easily permit returning internal errors (previously these were being treated as the selection being unable to rebase). E.g. in `Field.type_if_added_to()`, fetching the field's return type and converting it to an output type should always succeed, and if the parent type/schema matches the field's, looking up the field in the schema should always succeed as well. Similarly, in `Field.can_add_to()`, when the field has a selection set, the field's return type should always be composite.
- Methods have also been cleaned up a bit:
  - Schema/type position checks have been made a bit more consistent structure-wise.
  - In `FieldSelection.can_add_to()` and `InlineFragmentSelection.can_add_to()`, recursing into the selection set has been changed to call `SelectionSet.can_rebase_on()`.
- Update tests.
  - Two tests in `merged_abstract_types_handling.rs` now succeed.
  - One test in `build_query_plan_tests.rs` now fails for a different reason.

### DIFF
--- a/apollo-federation/src/operation/rebase.rs
+++ b/apollo-federation/src/operation/rebase.rs
@@ -20,6 +20,7 @@ use crate::error::FederationError;
 use crate::link::federation_spec_definition::get_federation_spec_definition_from_subgraph;
 use crate::schema::position::CompositeTypeDefinitionPosition;
 use crate::schema::position::ObjectTypeDefinitionPosition;
+use crate::schema::position::OutputTypeDefinitionPosition;
 use crate::schema::ValidFederationSchema;
 
 // TODO(@goto-bus-stop): this is precomputed in the QueryPlanner constructor. Can we expose that
@@ -85,10 +86,13 @@ impl Selection {
         &self,
         parent_type: &CompositeTypeDefinitionPosition,
         schema: &ValidFederationSchema,
-    ) -> bool {
+    ) -> Result<bool, FederationError> {
         match self {
             Selection::Field(field) => field.can_add_to(parent_type, schema),
-            Selection::FragmentSpread(_) => true,
+            // Since `rebaseOn` never fails, we copy the logic here and always return `true`. But as
+            // mentioned in `rebaseOn`, this leaves it a bit to the caller to know what they're
+            // doing.
+            Selection::FragmentSpread(_) => Ok(true),
             Selection::InlineFragment(inline) => inline.can_add_to(parent_type, schema),
         }
     }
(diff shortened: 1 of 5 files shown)


apollo-federation/src/operation/rebase.rs
```diff
@@ -288,26 +288,24 @@ impl FieldSelection {
...
+            return Ok(false);
         };
 
         if let Some(set) = &self.selection_set {
-            if set.type_position != ty {
-                return set
-                    .selections
-                    .values()
-                    .all(|sel| sel.can_add_to(parent_type, schema));
+            let ty: CompositeTypeDefinitionPosition = ty.try_into()?;
```
dariuszkuc, line 303: I think `type_if_added_to` already returns `CompositeTypeDefinitionPosition` so why do we need the extra `try_into`?

dariuszkuc: nvm you changed the method return type 👍

dariuszkuc (approved):

apollo-federation/src/operation/rebase.rs
sachindshinde: The return type of `Field.type_if_added_to()` was previously `CompositeTypeDefinitionPosition`, but this was causing it to return `None` for enums/scalars, which caused `FieldSelection.can_add_to()` to return `false`. I've updated it in this PR to return an `OutputTypeDefinitionPosition` instead, and only try converting to `CompositeTypeDefinitionPosition` when there's a selection set on the field.

PR #5404, opened 2024-06-11
 FED-290: Fixed error: "fragment spread does not have an element"

duckki:
* fixed `make_selection` to handle fragment spreads (only happens during operation optimization)
* ported several more operation optimize tests (not directly related to the fix).

Fixes https://apollographql.atlassian.net/browse/FED-290

~This PR depends on FED-280.~
* ~Two tests are failing after the fix. But, those failures will be gone once FED-280 is fixed.~

---

**Checklist**

- Tests added and passing[^3]
    - [X] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-federation/src/operation/mod.rs
```diff
@@ -596,11 +596,21 @@ pub(crate) enum Selection {
...
 }
 
+/// Element enum that is more general than OpPathElement.
+/// - Used for operation optimization.
+#[derive(Debug, Clone, derive_more::From)]
+pub(crate) enum OperationElement {
+    Field(Field),
+    FragmentSpread(FragmentSpread),
+    InlineFragment(InlineFragment),
+}
```
goto-bus-stop, line 606: I kinda recoiled at first seeing yet another selection type, but now we have this set of data structures:

| Data | Cached key wrapper | Selection wrapper |
|------|----------------|-|
| FieldData | Field | FieldSelection |
| InlineFragmentData | InlineFragment | InlineFragmentSelection |
| FragmentSpreadData | FragmentSpread | FragmentSpreadSelection |
| | OperationElement | Selection |

(plus like `OpPathElement` which only somewhat qualifies, and doesnt fit in the table)

I think that's a pretty acceptable place to be because we can, I think, generalise the middle and right columns into a single ergonomic generic type each. I wrote some notes on it in FED-302.

apollo-federation/src/operation/mod.rs
```diff
@@ -1415,6 +1465,23 @@ impl FragmentSpreadSelection {
         }
     }
 
+    /// Creates a fragment spread selection (in an optimized operation).
+    /// - `named_fragments`: Named fragment definitions that are rebased for the element's schema.
```
dariuszkuc, line 1469: we are bit rebasing here so this comment is somewhat confusing

apollo-federation/src/operation/rebase.rs
```diff
@@ -309,6 +310,41 @@ impl FieldSelection {
...
+            return if let RebaseErrorHandlingOption::ThrowError = error_handling {
+                Err(FederationError::internal(format!(
+                    "Cannot rebase {} fragment if it isn't part of the provided fragments",
+                    self.data().fragment_name
+                )))
+            } else {
+                Ok(None)
+            };
+        };
+        debug_assert_eq!(*schema, self.data().schema);
```
dariuszkuc, line 333: can you add a message to those debug so it is easier to figure out what is the problem?

dariuszkuc (approved):
If possible lets add those debug asserts log messages, otherwise looks fine

apollo-federation/src/operation/mod.rs
duckki: I know. That's the situation for now. I'll follow up on FED-302.

PR #5406, opened 2024-06-11
Fix integer overflow in QP reduce_options_if_needed

SimonSapin:
The first commit shows the test failing.

This method is all about bringing this number down, so we use `usize::MAX` to represent values that overflowed and recompute as needed.

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-federation/tests/query_plan/build_query_plan_tests/debug_max_evaluated_plans_configuration.rs
+++ b/apollo-federation/tests/query_plan/build_query_plan_tests/debug_max_evaluated_plans_configuration.rs
@@ -540,3 +540,32 @@ fn avoid_considering_indirect_paths_from_the_root_when_a_more_direct_one_exists(
     // key) is never exactly a good idea.
     assert_eq!(plan.statistics.evaluated_plan_count.get(), 4);
 }
+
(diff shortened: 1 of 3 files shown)


apollo-federation/tests/query_plan/build_query_plan_tests/debug_max_evaluated_plans_configuration.rs
```diff
@@ -540,3 +540,106 @@ fn avoid_considering_indirect_paths_from_the_root_when_a_more_direct_one_exists(
...
+                f58: id
+                f59: id
+                f60: id
+                f61: id
+                f62: id
+                f63: id
+                f00: id
+                f13: id
+                f01: id
+                f02: id
```
SimonSapin, line 626: I don’t know why the original ordering is not preserved. The overall Router is not affected since `format_response` well reorder data to what the original supergraph client expects, but should we still consider this a bug?

lrlna (approved):

PR #5409, opened 2024-06-11
fix(federation): fix remove_recursive on types with a referencing field

goto-bus-stop:
Reorders removing types, so that `ObjectTypeDefinitionPosition::remove_recursive()` first removes the references to the type, and then removes the type. Else, removing the reference would fail, as it needs to look up the type in the schema to know what kind of reference it has to remove from the `referencers` structure.

This fixes an issue with a schema that, during `extract_subgraphs_from_supergraph`, ends up looks something like this:
```graphql
type User {
  id: ID!
  profile: UserProfile
}
type UserProfile {
}
type Query {
  me: User
}
```
Then, subgraph extraction removes all empty types, which calls `UserProfile.remove_recursive()` and expects to remove the reference to it as well.

This seems very brittle, to be honest. I can add a bunch more tests to cover more cases but it might need a design tweak to be more reliable.


apollo-federation/src/schema/position.rs
```diff
@@ -3339,14 +3368,20 @@ impl InterfaceTypeDefinitionPosition {
...
+    /// Remove this interface from the schema, and recursively remove references to the interface.
     pub(crate) fn remove_recursive(
         &self,
         schema: &mut FederationSchema,
     ) -> Result<(), FederationError> {
-        let Some(referencers) = self.remove_internal(schema)? else {
+        let Some(referencers) = schema
+            .referencers
+            .interface_types
+            .swap_remove(&self.type_name)
```
lrlna, line 3383: TIL `swap_remove()` is a thing. cool!

goto-bus-stop: yeah, with indexmap you must pick either shift or swap. I don't think referencers order matters so swap is faster 🤔

lrlna (approved):
Thank you for all the doc comments and a test too!! I also just ran this for all the graphs that had this issue and none of fail for this anymore.

lrlna:
> This seems very brittle, to be honest. I can add a bunch more tests to cover more cases but it might need a design tweak to be more reliable.

@goto-bus-stop can I ask you make an issue for us to remember to think about design tweaks for this? I got around to make an epic for things like this under [FED-308](https://apollographql.atlassian.net/browse/FED-308).

PR #5412, opened 2024-06-11
Return an error in case of excessive cartesian product size

SimonSapin:
Before https://github.com/apollographql/router/pull/5402, some combinations of schema and query would cause `flat_cartesian_product` to compute `num_options == 1 << 44` and try to allocate 211 terabytes in `Vec::with_capacity`. The heap allocation failed, which made `Vec` abort the process.

Even if we avoided the abort somehow, `flat_cartesian_product` would then go on to try to build trillions of items and take excessive time and/or memory.

This PR makes `flat_cartesian_product` if the number of items exceeds an arbitrary limit of one million items. This is **untested**, since https://github.com/apollographql/router/pull/5402 already fixed our know reproduction case. If another case is found please add a test.

### DIFF
--- a/apollo-federation/src/query_graph/graph_path.rs
+++ b/apollo-federation/src/query_graph/graph_path.rs
@@ -2642,7 +2642,7 @@ impl OpGraphPath {
                         }
                         let all_options = SimultaneousPaths::flat_cartesian_product(
                             options_for_each_implementation,
-                        );
+                        )?;
                         if let Some(interface_path) = interface_path {
                             let (interface_path, all_options) =
                                 if direct_path_overrides_type_explosion {
@@ -2799,7 +2799,7 @@ impl OpGraphPath {
                         }
                         let all_options = SimultaneousPaths::flat_cartesian_product(
                             options_for_each_implementation,
-                        );
+                        )?;
                         Ok((Some(all_options), None))
                     }
                     OutputTypeDefinitionPosition::Object(tail_type_pos) => {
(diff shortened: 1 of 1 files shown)


TylerBloom (approved):
LGTM

apollo-federation/src/query_graph/graph_path.rs
```diff
@@ -3069,8 +3069,14 @@ impl SimultaneousPaths {
         // Pre-allocate `Vec` for the result.
         let num_options = options_for_each_path
             .iter()
-            .map(|options| options.len())
-            .product();
+            .fold(1_usize, |product, options| {
+                product.saturating_mul(options.len())
+            });
+        if num_options > 1_000_000 {
```
dariuszkuc, line 3075: this seems pretty high?

SimonSapin: it’s 24 MiB of memory for the Vec in the next line :shrug:

TylerBloom: Its just under `2^20`, which seems fine for the preview.

dariuszkuc (approved):

apollo-federation/src/query_graph/graph_path.rs
SimonSapin: That said I did pick this number out of thin air, and we can easily change it

TylerBloom: I think refining this number will be a topic of another meeting in a couple of weeks.

PR #5419, opened 2024-06-12
Fix bug in `advance_with_operation_element()` where type explosion should be avoided

sachindshinde:
This PR fixes a bug in `OpGraphPath.advance_with_operation_element()` where, when dealing with a field on an interface type, type explosion was mistakenly performed when it shouldn't have been.

### DIFF
--- a/apollo-federation/src/query_graph/graph_path.rs
+++ b/apollo-federation/src/query_graph/graph_path.rs
@@ -2511,7 +2511,7 @@ impl OpGraphPath {
                                 TypeDefinitionPosition::Scalar(_) | TypeDefinitionPosition::Enum(_)
                             );
                             if is_operation_field_type_leaf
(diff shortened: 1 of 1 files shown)


goto-bus-stop:
awesome! does this fix the `flat_cartesian_product` issue we had? is it easy to write a test for this?

goto-bus-stop (approved):

sachindshinde:
@goto-bus-stop 
I suspect that this was causing most of them, though we won't know for sure until we check.

With integration tests it's a bit difficult to test, since unnecessary type explosion effectively gets pruned when computing the best plan (the added complexity results in a cost function penalty). If we develop testing utilities to help set up inputs to intermediary methods, I think we could write unit tests for this.

lrlna:
Ran this on all graphs that previously had this issue, and it doesn't happen anymore 🎉

PR #5421, opened 2024-06-12
Fix bugs in interface object checking

sachindshinde:
While reviewing #5401, I noticed we had some bugs when checking for interface objects. This PR:
- Fixes a bug in `compute_input_rewrites_on_key_fetch()` where it was passing the wrong schema in when checking whether a type was an interface object.
  - Note that this required adding an argument to `add_post_require_inputs()`, which then caused Clippy to complain about there being too many arguments. I've accordingly split some of the arguments out into a separate struct.
- Fixes a bug in `is_interface_object_type()` where it silently ignores if the type doesn't exist, or if the `@interfaceObject` directive definition is missing. 
  - This requires changing it and some upstream callers to return `Result`s.
- Fixes a bug in `is_useless_node()` where it silently ignores if a node or its subgraph schema doesn't exist. 
- Changes `is_interface_object_type()` to be on `ValidFederationSchema` instead of on schema positions.
  - The JS codebase placed this function on `FederationMetadata` (renamed to `SubgraphMetadata` in Rust). We can't put it on `SubgraphMetadata` since it doesn't have access to the schema (to avoid circular references), but I've hoisted the method up to `ValidFederationSchema` instead.
  - I'm hesitant about adding subgraph-specific logic to schema positions, as it goes against the current separation of concerns (it's also a bit repetitive at the moment, and `position.rs` is already quite large).

### DIFF
--- a/apollo-federation/src/query_plan/fetch_dependency_graph.rs
+++ b/apollo-federation/src/query_plan/fetch_dependency_graph.rs
@@ -3653,21 +3653,16 @@ fn handle_requires(
     let head = dependency_graph
         .federated_query_graph
         .edge_head_weight(query_graph_edge_id)?;
+    let entity_type_schema = dependency_graph
+        .federated_query_graph
+        .schema_by_source(&head.source)?
+        .clone();
     let QueryGraphNodeType::SchemaType(OutputTypeDefinitionPosition::Object(entity_type_position)) =
         head.type_.clone()
     else {
         return Err(FederationError::internal(
             "@requires applied on non-entity object type",
         ));
-    };
-    let entity_type_schema = dependency_graph
-        .federated_query_graph
-        .schema_by_source(&head.source)?
-        .clone();
-    let post_require_inputs_edge_info = PostRequireInputsEdgeInfo {
-        entity_type_position: entity_type_position.clone(),
-        entity_type_schema: &entity_type_schema,
-        query_graph_edge_id,
     };
 
     // In many cases, we can optimize @requires by merging the requirement to previously existing nodes. However,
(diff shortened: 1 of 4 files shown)


apollo-federation/src/query_plan/fetch_dependency_graph.rs
```diff
@@ -1206,21 +1207,20 @@ impl FetchDependencyGraph {
...
+                Ok(self
+                    .parents_of(node_index)
+                    .map(|p| {
+                        let p_node = self.node_weight(p)?;
+                        let p_subgraph_name = &p_node.subgraph_name;
+                        let p_subgraph_schema = get_subgraph_schema(p_subgraph_name)?;
+                        let Ok(type_in_parent) =
+                            p_subgraph_schema.get_type(condition.type_name().clone())
+                        else {
+                            return Ok(false);
```
goto-bus-stop, line 1219: I see the JS does the same, but how intentional is it that this returns `false` instead of an error? Can we add a comment if it is okay if the subgraph schema doesn't have that type?

apollo-federation/src/query_plan/fetch_dependency_graph.rs
```diff
@@ -4125,29 +4134,34 @@ fn inputs_for_require(
     }
 }
 
+struct PostRequireInputsEdgeInfo<'a> {
```
goto-bus-stop, line 4137: I'm not sure this grouping really makes things easier to understand. If it's just to appease clippy, it's okay to use an `#[allow()]` attribute on the function.

goto-bus-stop (approved):
LGTM, agree with not having subgraph stuff on Positions, ideally in the future we will have different types for supergraph / subgraph / API schema so it can live exclusively on the appropriate type of schema.

goto-bus-stop:
As discussed in standup, I replaced the new struct by using `#[allow()]` so we can merge it like that. Feel free to follow up if you disagree with that (maybe it makes more sense than I think)

PR #5425, opened 2024-06-12
Implemented `Operation::without_defer`

TylerBloom:
Currently, the Rust QP does not handle `@defer`. Because of this, the private preview needs to strip out the use of this directive. This is requires implementing `Operation::without_defer` and ensuring that, while building the query plan, this method is called regardless of config.

Fixes FED-293

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-federation/src/operation/mod.rs
+++ b/apollo-federation/src/operation/mod.rs
@@ -143,6 +143,15 @@ impl Operation {
     // PORT_NOTE(@goto-bus-stop): It might make sense for the returned data structure to *be* the
     // `DeferNormalizer` from the JS side
     pub(crate) fn with_normalized_defer(self) -> NormalizedDefer {
+        NormalizedDefer {
+            operation: self,
+            has_defers: false,
+            assigned_defer_labels: HashSet::new(),
+            defer_conditions: IndexMap::new(),
+        }
+        // TODO(@TylerBloom): Once defer is implement, the above statement needs to be replaced
+        // with the commented-out one below.
+        /*
         if self.has_defer() {
             todo!("@defer not implemented");
         } else {
@@ -153,6 +162,7 @@ impl Operation {
                 defer_conditions: IndexMap::new(),
             }
         }
+        */
     }
 
     fn has_defer(&self) -> bool {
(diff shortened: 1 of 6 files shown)


apollo-federation/src/operation/mod.rs
```diff
@@ -2919,6 +2978,26 @@ impl SelectionSet {
...
+    /// Removes the @defer directive from all selections without removing that selection.
+    // PORT_NOTE: This behavior differs from the JS code, which removes anything with @defer.
+    // TODO(@TylerBloom): After the private preview and @defer being impl-ed, check if this logic
+    // should be changed to mirror the JS code.
+    fn without_defer(&mut self) {
+        // TODO: This doesn't seem like the correct way to get the directive name...
+        let Some(defer_name) = self.schema.get_directive_definition(&name!("defer")) else {
+            // TODO: Return an error? Continue? Dunno...
+            return;
+        };
```
TylerBloom, line 2990: If we are unable to fetch the `@defer` directive definition, we should return early, ya?

apollo-federation/src/operation/mod.rs
```diff
@@ -2919,6 +2978,26 @@ impl SelectionSet {
         Ok(normalized_selections)
     }
 
+    /// Removes the @defer directive from all selections without removing that selection.
+    // PORT_NOTE: This behavior differs from the JS code, which removes anything with @defer.
+    // TODO(@TylerBloom): After the private preview and @defer being impl-ed, check if this logic
+    // should be changed to mirror the JS code.
+    fn without_defer(&mut self) {
+        // TODO: This doesn't seem like the correct way to get the directive name...
+        let Some(defer_name) = self.schema.get_directive_definition(&name!("defer")) else {
```
TylerBloom, line 2987: This seems incorrect, but I always forget how to pull out the directive definition from a schema because the users might have renamed the directive when they imported it.

TylerBloom (commented):
There are two things that I would like to point out that I was not sure of.

apollo-federation/src/operation/mod.rs
```diff
@@ -164,11 +164,15 @@ impl Operation {
                 .any(|f| f.has_defer())
     }
 
-    pub(crate) fn without_defer(self) -> Self {
```
lrlna, line 167: there is a `with_normalized_defer` function right above this that has a `todo!()`. As part of this PR can you remove that todo and just return the `NormalizedDefer` struct and add a comment mentioning `FED-95`? I didn't want to do it in #5426 so we could keep all of the defer things together.

apollo-federation/src/operation/mod.rs
goto-bus-stop: `@defer` is a spec name, not an imported one, so it's always `@defer` and you don't have to bother with renames :)

TylerBloom: Excellent! Thanks

dariuszkuc (approved):

PR #5426, opened 2024-06-12
fix(federation): turn todos into SingleFederationError

lrlna:
Instead of panicking with `todo!()`, return an error.

A lot of these were only coming from composition call sites as far as I could tell, but I'd rather not have the todos and have redundant errors in places than a panic. 

I added an explicit `UnsupportedFederationVersion` error kind to `SingleFederationError`. It's `ErrorCode` placement is the very last, so shouldn't mess up the numbering there. I prefer to be extra explicit here rather than returning `Internal`. 

The `MergeFailure` had to be turned to `Vec<String>`, I didn't really feel like dealing with `Vec<&str>` (which we shouldn't have had in the first place) when trying to do a `From`. That entire module's error handling will have to be reworked anyways, this just makes it more straightforward to convert for that todo.

### DIFF
--- a/apollo-federation/src/lib.rs
+++ b/apollo-federation/src/lib.rs
@@ -111,8 +111,7 @@ impl Supergraph {
     pub fn compose(subgraphs: Vec<&ValidSubgraph>) -> Result<Self, MergeFailure> {
         let schema = merge_subgraphs(subgraphs)?.schema;
         Ok(Self {
-            schema: ValidFederationSchema::new(schema)
-                .map_err(|err| Into::<MergeFailure>::into(err))?,
+            schema: ValidFederationSchema::new(schema).map_err(Into::<MergeFailure>::into)?,
         })
     }
 
--- a/apollo-federation/src/lib.rs
+++ b/apollo-federation/src/lib.rs
@@ -112,7 +112,7 @@ impl Supergraph {
         let schema = merge_subgraphs(subgraphs)?.schema;
         Ok(Self {
             schema: ValidFederationSchema::new(schema)
-                .map_err(|err| todo!("missing error handling: {err}"))?,
+                .map_err(|err| Into::<MergeFailure>::into(err))?,
         })
     }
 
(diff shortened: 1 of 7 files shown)


apollo-federation/src/lib.rs
```diff
@@ -112,7 +112,7 @@ impl Supergraph {
         let schema = merge_subgraphs(subgraphs)?.schema;
         Ok(Self {
             schema: ValidFederationSchema::new(schema)
-                .map_err(|err| todo!("missing error handling: {err}"))?,
+                .map_err(|err| Into::<MergeFailure>::into(err))?,
```
lrlna, line 115: we definitely don't run `.compose()` at runtime, but better safe than sorry o/

apollo-federation/src/operation/mod.rs
```diff
@@ -3045,9 +3045,7 @@ impl SelectionSet {
                     path: Vec::new(),
                     field: field.clone(),
                 }),
-                Selection::FragmentSpread(_fragment) => {
-                    todo!()
-                }
+                Selection::FragmentSpread(_fragment) => return fields,
```
lrlna, line 3048: @duckki I could use your input on this one, I know you worked with this code just recently. Is the panic here (`todo!()` or `unreachable!()`) absolutely necessary here? Can we return what would be an empty vec with `return fields`? Or will that mess things up down the line?

duckki: Yes, returning an empty vec is sensible here. But, please add an debug assertion so we can catch it during testing.
BTW, I'd like to move it into `optimize.rs`, where it should belong. But, we can do that refactoring later.

TylerBloom: @duckki , should we return `vec![]` or `fields`?

duckki: Oh, I meant "not add to `fields` and keep going on".

duckki:
```suggestion
                Selection::FragmentSpread(_fragment) => {
                    debug_assert!(false, "unexpected fragment spreads in expanded fetch operation");
                }
```

duckki (approved):

PR #5428, opened 2024-06-12
Planner duration metric with "planner mode" attribute

xuorig:
Mostly to discuss, not sure about the changes.

It would be nice to get metrics about the planning duration with a tag with the actual planner used under the hood. This requires moving the metric closer to the actual planner call.

Problem is we currently tag the `workerId` in the pool, where the metric is actually emitted. In this PR I've currently opted for a new metric for the pool. But alternatives with a bit of a higher effort could be:

  - Passing `workerId` down, which feels a bit odd leaking that concept.
  - Returning the time taken and planner in the response, and then emitting the metric with worker ID. (feels better)
  
  Let me know what you think.

### DIFF
--- a/apollo-router/src/query_planner/bridge_query_planner_pool.rs
+++ b/apollo-router/src/query_planner/bridge_query_planner_pool.rs
@@ -116,12 +116,11 @@ impl BridgeQueryPlannerPool {
                             continue;
                         }
                     };
-                    let start = Instant::now();
 
                     let res = svc.call(request).await;
 
                     f64_histogram!(
-                        "apollo.router.query_planning.plan.duration",
+                        "apollo.router.query_planning.pooled_planner.duration",
                         "Duration of the query planning.",
                         start.elapsed().as_secs_f64(),
                         "workerId" = worker_id.to_string()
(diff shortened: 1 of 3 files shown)


apollo-router/src/query_planner/bridge_query_planner_pool.rs
```diff
@@ -116,12 +116,11 @@ impl BridgeQueryPlannerPool {
                             continue;
                         }
                     };
-                    let start = Instant::now();
 
                     let res = svc.call(request).await;
 
                     f64_histogram!(
-                        "apollo.router.query_planning.plan.duration",
+                        "apollo.router.query_planning.pooled_planner.duration",
```
bnjjj, line 123: We try as much as we can to not rename an existing metric because it could break existing dashboard of our customers and would introduce a breaking change

bnjjj (changes requested):

apollo-router/src/query_planner/bridge_query_planner_pool.rs
bnjjj: Hmm I just read the PR's description more in details. And it looks like this is like a different one and the previous one does still exist... Which is valid.

cc @BrynCooke do you have any opinion ?

BrynCooke: Yeah, we can't rename the existing metric.

BrynCooke (changes requested):
There's more discussion to be had, but at a code level multiple calls to the metrics macros at different callsites isn't good as any difference in description leads to runtime warnings.

lrlna:
@xuorig I am going to close this in favour of #5530, where I made a few small tweaks.

PR #5429, opened 2024-06-12
Fix bug in alias checking where it only looks at top-level

sachindshinde:
This PR fixes a bug in our alias-checking logic for field sets, where it only checked the top-level of the selection set (it should check descendants as well).


apollo-federation/src/operation/mod.rs
```diff
@@ -1537,6 +1553,22 @@ impl FragmentSpreadSelection {
         }
         self.selection_set.any_element(predicate)
     }
+
+    pub(crate) fn for_each_element(
+        &self,
+        parent_type_position: CompositeTypeDefinitionPosition,
+        callback: &mut impl FnMut(OpPathElement) -> Result<(), FederationError>,
+    ) -> Result<(), FederationError> {
+        let inline_fragment = InlineFragment::new(InlineFragmentData {
```
dariuszkuc, line 1562: guessing this is because `OpPathElement` cannot be fragment spread?

sachindshinde: It's more because it's what the JS codebase currently does (which I'm guessing is a reflection of the caller's expectations/requirements).

dariuszkuc (approved):

SimonSapin (approved):

PR #5431, opened 2024-06-13
(FED-305) error: fragment `x` must be used in an operation

duckki:
fixed `NamedFragments::reduce` to loop until fix-point

The first commit ports the rest of operation optimization tests (except for 3 of them that need to port `SelectionSetNode`). Two of them failed due to FED-305.

The second commit fixes the bug where `NamedFragments::reduce` didn't iterate multiple times to reduce fragments all the way. Reduction of fragments and normalization of dropped fragments may have chain reactions caused by each other. Thus, the reduction must iterate until a fix-point is reached. I previously overlooked this and ran it just once.

This also fixes the usage of deeply nested fragments that are referenced from conditional fragments which are eliminated or inlined.

I confirmed this fixes the observed failures from the corpus operations.

Fixes https://apollographql.atlassian.net/browse/FED-305

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-federation/src/operation/optimize.rs
```diff
@@ -855,10 +855,11 @@ impl Selection {
...
                     } else {
                         // Create an inline fragment since type condition is necessary.
-                        let inline = InlineFragmentSelection::from_fragment_spread_selection(
+                        let inline = InlineFragmentSelection::from_selection_set(
                             parent_type.clone(),
-                            fragment,
-                        )?;
+                            expanded_sub_selections,
+                            fragment.spread.data().directives.clone(),
+                        );
```
duckki, line 862: Nice catch!

duckki (commented):
@dariuszkuc Thanks for the fix!

lrlna (approved):

PR #5432, opened 2024-06-13
fix(federation): fix invalid unit test inputs

dariuszkuc:
`@skip/@include` directives require non-nullable `Boolean!` condition. Queries in our tests were invalid. Looks like JS parser is much more lenient on the inputs as it allows passing nullable `Boolean` condition without a default.

Fixes #[FED-253](https://apollographql.atlassian.net/browse/FED-253)

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[FED-253]: https://apollographql.atlassian.net/browse/FED-253?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


duckki (approved):
The change looks good! I've seen a few cases like this before where JS test's query inputs are slightly invalid, but tests themselves are passing.

lrlna (approved):

PR #5433, opened 2024-06-13
Add support for query limit metrics

garypen:
Adds support for four new attributes for the supergraph Query instrument:
 - aliases (the number of aliases in the query)
 - depth (the depth of the query)
 - height (the height of the query)
 - root_fields (the number of root_fields in the query)

You can use this data to understand how your graph is used and to help determine where to set limits, if desired.

For example:

```
instruments:
  supergraph:
    'query.depth':
      description: 'The depth of the query'
      value:
        query: depth
      unit: unit
      type: histogram
```


apollo-router/src/plugins/telemetry/config_new/cost/mod.rs
```diff
@@ -284,7 +284,7 @@ pub(crate) enum CostValue {
 
 pub(crate) fn add_cost_attributes(context: &Context, custom_attributes: &mut Vec<KeyValue>) {
     context.extensions().with_lock(|c| {
-        if let Some(cost) = c.get::<CostContext>().cloned() {
+        if let Some(cost) = c.get::<CostContext>() {
```
garypen, line 287: drive-by: no need to clone

apollo-router/src/plugins/telemetry/dynamic_attribute.rs
```diff
@@ -6,6 +6,7 @@ use tracing_subscriber::registry::LookupSpan;
 use tracing_subscriber::Layer;
 use tracing_subscriber::Registry;
 
+use super::formatters::APOLLO_PRIVATE_PREFIX;
```
garypen, line 9: drive-by: remove redunant const

apollo-router/src/plugins/telemetry/formatters/text.rs
```diff
@@ -24,12 +24,12 @@ use tracing_subscriber::registry::SpanRef;
 
 use super::get_trace_and_span_id;
 use super::EventFormatter;
+use super::APOLLO_PRIVATE_PREFIX;
```
garypen, line 27: drive-by: remove redundant const

apollo-router/src/plugins/telemetry/tracing/apollo_telemetry.rs
```diff
@@ -97,14 +105,6 @@ const APOLLO_PRIVATE_HTTP_RESPONSE_HEADERS: Key =
     Key::from_static_str("apollo_private.http.response_headers");
 pub(crate) const APOLLO_PRIVATE_OPERATION_SIGNATURE: Key =
     Key::from_static_str("apollo_private.operation_signature");
-pub(crate) const APOLLO_PRIVATE_COST_ESTIMATED: Key =
```
garypen, line 100: drive-by: remove redundant consts

apollo-router/src/plugins/telemetry/tracing/mod.rs
```diff
@@ -16,6 +16,7 @@ use serde::Deserialize;
 use tower::BoxError;
 
 use super::config_new::spans::Spans;
+use super::formatters::APOLLO_PRIVATE_PREFIX;
```
garypen, line 19: drive-by: remove redundant const

docs/source/configuration/telemetry/instrumentation/selectors.mdx
```diff
@@ -33,39 +33,39 @@ The router service is the initial entrypoint for all requests. It is HTTP centri
 | Selector           | Defaultable | Values                                    | Description                          |
 |--------------------|-------------|-------------------------------------------|--------------------------------------|
 | `trace_id`         | Yes         | `open_telemetry`\|`datadog`               | The trace ID                         |
-| `operation_name`   | Yes         | `string`|`hash`                           | The operation name from the query    |
-| `studio_operation_id` | Yes      | `true`|`false`                            | The Apollo Studio operation id       |
+| `operation_name`   | Yes         | `string`\|`hash`                          | The operation name from the query    |
```
garypen, line 36: drive-by: fix markup

bnjjj: You know the guilty... :)

apollo-router/src/plugins/telemetry/config_new/selectors.rs
```diff
@@ -804,6 +815,26 @@ impl Selector for SupergraphSelector {
 
     fn on_response(&self, response: &supergraph::Response) -> Option<opentelemetry::Value> {
         match self {
+            SupergraphSelector::Query { query, .. } => {
```
bnjjj, line 818: If it happens in `on_response` we might duplicate this behavior for `on_error` as even on error it could be interesting to have these selectors applied right ?

bnjjj (changes requested):

BrynCooke (changes requested):
Looks generally great, I would have expected this snapshot to have changed though:
https://github.com/apollographql/router/blob/dev/apollo-router/tests/snapshots/apollo_reports__demand_control_trace.snap#L557-L560

garypen:
> Looks generally great, I would have expected this snapshot to have changed though: https://github.com/apollographql/router/blob/dev/apollo-router/tests/snapshots/apollo_reports__demand_control_trace.snap#L557-L560

Good catch. 30 minutes of digging reveal that the issue is in the `LightSpanData` struct. I didn't know that I had to pass the new attribute names into the filtering list here, so they were being thrown away.

I'm not sure how best to document this. Once you know about it, it's obvious, but it's very easy to miss right now.

apollo-router/src/plugins/telemetry/config_new/selectors.rs
garypen: It could be interesting. I've added it: 837aa9a7d3263d060f1e3f92aa2589c733b7442a

garypen: @bnjjj One thing about that. I'm not clear how (or if) I should document this. Any suggestions?

BrynCooke (approved):
Great!

BrynCooke:
I think we need to add something to `test_stats` and `test_batch_stats` to assert that the metrics are also being populated.

garypen:
> I think we need to add something to `test_stats` and `test_batch_stats` to assert that the metrics are also being populated.

I've spent most of today trying to understand what this requirement means and how to implement it and got nowhere. I don't understand:
 - what makes these metrics special?
 - how do we test the existence (or otherwise of other metrics)?

I have tried to look for hints in the way that other `limit` style data is treated, but can't find anything. I don't think these metrics are treated differently to how the existing limits are treated, so I would expect the existing test to cover the new metrics.

I've extended `test_extract_limits` to include the new metrics: (see: c58773889f53c411a3bc4cbd15266eb15f90c72b).

BrynCooke (approved):

bnjjj (approved):

PR #5438, opened 2024-06-13
Rename the scaffold `Cargo.toml` files so that they don't confuse cargo

dylan-apollo:
When consuming this repo as a Git dependency, Cargo throws errors for any file named `Cargo.toml` that isn't actually a valid Cargo manifest (as is the case for our templates). This especially plagues RustRover which just refuses to give you completions/highlighting if Cargo returned an error.

This works around that problem by renaming the files in a post-hook for scaffold instead of naming the templates properly.


SimonSapin (approved):
Let’s assume `mv` is cross-platform enough for our users

PR #5443, opened 2024-06-13
Send demand control costs to Studio via protobuf limits field

tninesling:
If you are an Apollo GraphOS user, field usage and lengths can be sent to apollo studio without requiring subgraphs to support FTV1.
This data is not currently displayable in GraphOS Studio.

This is an experimental feature and must be enabled via:

```yaml
telemetry:
  apollo:
    experimental_local_field_metrics: true
```

The reason for this is there is currently a slight performance impact from using this.

---

**Checklist**

- Tests added and passing[^3]
    - [X] Unit Tests
    - [x] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


BrynCooke (approved):

BrynCooke (changes requested):
Code looks great, but there looks to be perf issues.

BrynCooke (approved):

apollo-router/src/plugins/telemetry/metrics/apollo/list_length_histogram.rs
```diff
@@ -0,0 +1,103 @@
+use std::ops::AddAssign;
+
+use hdrhistogram::Histogram;
+use serde::ser::SerializeSeq;
+use serde::Serialize;
+
+use super::studio::MAX_HISTOGRAM_BUCKETS;
+
+/// A histogram for recording lengths of list fields in GraphQL responses. This implementation is clamped to a maximum
+/// of 386 buckets, a restriction imposed by Studio. The buckets roughly follow order of magnitude of the recorded value,
```
garypen, line 10: I note that duration histogram uses 383 and that was taken from the TS implementation. Which one is correct?

garypen (commented):
Do we need the new histogram types or could we re-purpose DurationHistogram somehow into a general purpose thing?

apollo-router/src/plugins/telemetry/metrics/apollo/list_length_histogram.rs
BrynCooke: I've fixed this in the histogram refactor code.

PR #5444, opened 2024-06-13
Fix bug where field/argument `remove_type_references()` requires referenced types to exist

sachindshinde:
This PR is another take at #5409 for fixing the same issues. To summarize some context here:
- We currently maintain the invariant that we add an entry to `FederationSchema.referencers` for a type/directive before adding the type/directive itself to the `apollo_compiler::Schema`. This means that whenever code observes that a type/directive exists in a schema, it can assert that it always has a corresponding entry in `FederationSchema.referencers`. This also means the contrapositive holds, i.e. whenever code observes that an entry is missing from `FederationSchema.referencers` for a type/directive, then it can assert it doesn't exist in the schema.
- This is why in code for `remove()` and `remove_recursive()`, we always remove the element from the `Schema` and then remove the entry from `FederationSchema.referencers`, as it maintains the invariant.
- In #5409, my understanding is that we'd like to shift toward removing elements that reference a type/directive before removing the type/directive itself. That stance makes sense, as it avoids references for that type/directive becoming temporarily invalid.
- One concern with this approach is that you may end up trying to remove the same type/directive more than once when removing elements recursively. When you recurse, you need some way of being aware of types/directives that are in the process of being removed. To my understanding, #5409 effectively tracks this by removing from `FederationSchema.referencers`, and early-returning if an entry isn't present, which breaks the invariant mentioned earlier. I'd prefer here that we instead track a set of types/directives we're in the midst of removing, rather than trying to reuse `FederationSchema.referencers` for such a purpose.
- The JS codebase takes a different approach, removing the type/directive first and then removing any referencers which may have become invalid. We take a similar approach in the current Rust code, which is why `remove_*()` methods generally  will ignore a type reference if it has no entry in `FederationSchema.referencers`. However, the methods `remove_type_references()` for fields/arguments have a bug where they were asserting that the type exists when finding the type kind, when it may not due to already having been removed (or the type was only pre-inserted).
- The fix I've made in this PR is to only check `FederationSchema.referencers` instead of looking up the type in the schema; this fix makes tests pass, and also seems to fix the original customer issue. I've mirrored the changes to `insert_type_references()` (which technically also had a similar bug, but wasn't elicited due to how the current callsites are structured).

To be clear, I think it would be good to move toward a world where we remove schema elements that reference a type/directive before removing the type/directive. It means our `remove_*()` methods can stop ignoring type/directive references that aren't present, which would be safer. However, that's a bigger design change that we should probably consider after further QP testing.

Regarding the `shift_remove()` vs `swap_remove()`, I agree that `swap_remove()` would be preferable here for `FederationSchema.referencers`. However, because referencers may be used to find e.g. the application sites of a directive, I've been hesitant to use `swap_remove()` since it deviates from insertion order. I'd be fine to switch to `swap_remove()` once QP testing is finished up.


goto-bus-stop:
That's a great explanation of your reasoning, thanks a lot!

goto-bus-stop (approved):

Issue #3208, opened 2023-06-02
trace_id is not always displayed in json logs


bnjjj:
For now only logs in span `router` contains the `trace_id`. We need to fix this in the json formatter or somewhere else

abernix:
This could be a consideration of #3226!

Issue #5387, opened 2024-06-10
Consistent trace id 


BrynCooke:
Users are having trouble correlating their logs due to there being multiple IDs floating around, sampling affecting the presence or not of a trace_id.

The behaviour we would like:

1. Otel Trace ID is exposable on logs. This functionality already exists.
```yaml
telemetry:
  exporters:
    logging:
      stdout:
        format:
          json:
            display_trace_id: true
```
2. Promote attributes from the root span to the json container as a top level attribute. This will allow users to have headers etc that have been attached to the root span be output on logs.
```yaml
telemetry:
  exporters:
    logging:
      stdout:
        format:
          json:
            promote_span_attributes:
              - my.custom.attribute 
```

3. TraceID generated for logs even if the request is not sampled. FIxed in: https://github.com/apollographql/router/pull/5447

4. `experimental_response_trace_id` should use the same logic as in `#3` to enable a consistent trace id to be returned to the client. Fixed in: https://github.com/apollographql/router/pull/5447

Samjin:
Would this method also add custom attribute to custom events as well? For instance, when user does trace.info!(), would the log object include the custom attribute?

abernix:
This issue might be fixed by @bnjjj's #5447.


> Would this method also add custom attribute to custom events as well? For instance, when user does trace.info!(), would the log object include the custom attribute?

Good question! I think @bnjjj would need to answer this part.

Samjin:
Talked to @BrynCooke that this feature would also append to custom `request` logs.

BrynCooke:
I'm going to close this and split out item 2 into a new ticket as we think the rest has been solved already.
https://github.com/apollographql/router/issues/5540

BrynCooke:
The PRs that closed items 3 and 4 are linked in the description, but is also here: https://github.com/apollographql/router/pull/5447

PR #5447, opened 2024-06-14
fix(telemetry): always create trace id even if the trace is not sampled

bnjjj:
Now when using `experimental_response_trace_id` for example, you'll always have a valid trace_id in response headers even if the trace is not sampled. It's following the behavior of the [OpenTelemetry specs](https://opentelemetry.io/docs/specs/otel/trace/sdk/#sampling). 

close https://github.com/apollographql/router/issues/3208

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [x] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/plugins/telemetry/dynamic_attribute.rs
```diff
@@ -223,22 +223,26 @@ impl EventDynAttribute for ::tracing::Span {
...
+                                    // Can't use ::tracing::error! because it could create deadlock on extensions
+                                    eprintln!("no OtelData, this is a bug");
                                 }
-                            },
-                            None => {
-                                // Can't use ::tracing::error! because it could create deadlock on extensions
-                                eprintln!("no OtelData, this is a bug");
                             }
+                        } else {
+                            // FIXME: we should put event attributes somewhere else to make it work even if it's not sampled like we did with LogAttributes
```
bnjjj, line 245: I created the [related issue ](https://github.com/apollographql/router/issues/5448)

apollo-router/src/plugins/telemetry/dynamic_attribute.rs
```diff
@@ -258,23 +262,27 @@ impl EventDynAttribute for ::tracing::Span {
...
```
bnjjj, line 285: Related [issue](https://github.com/apollographql/router/issues/5448) created

BrynCooke (changes requested):
Can we add some tests to verify that the traceID is making it to the logging also?
Either in https://github.com/apollographql/router/blob/dev/apollo-router/tests/integration/telemetry/logging.rs or the jaeger tests.

bnjjj:
@BrynCooke Added 2 tests for json and text format with sampler set to `always_off` to make sure it works when it's sampled and not sampled https://github.com/apollographql/router/pull/5447/commits/e81d7cdfe7e0788499b2537fd78f4fe9bfedda25

bnjjj:
@BrynCooke Correct me if I'm wrong but I think this one will fix https://github.com/apollographql/router/issues/5387 right ?

garypen (approved):

apollo-router/src/plugins/telemetry/otel/layer.rs
```diff
@@ -650,24 +757,53 @@ where
...
+                Some(SampledSpan::Sampled(trace_id, _) | SampledSpan::NotSampled(trace_id, _)) => {
+                    opentelemetry_api::trace::TraceId::from(trace_id.to_u128())
+                }
+                // It's probably the root span
+                None => self.tracer.new_trace_id(),
+            }
+        };
+
+        let span_id = self.tracer.new_span_id();
+        let sampled = if is_sampled {
```
BrynCooke, line 784: Nice!

BrynCooke (approved):

apollo-router/src/plugins/telemetry/config_new/events.rs
```diff
@@ -278,23 +289,36 @@ impl Instrumented
...
+                ),
+            ));
+            attrs.push(KeyValue::new(
+                Key::from_static_str("http.request.body"),
+                opentelemetry::Value::String(
+                    serde_json::to_string(request.supergraph_request.body())
+                        .unwrap_or_default()
+                        .into(),
+                ),
+            ));
```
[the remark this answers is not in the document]
bnjjj: Events will always be displayed even if the span is not sampled. But I think one of improvement we can do here is to check if the user set the level `debug` we can know thanks to a helper in `tracing` I think and directly not collecting anything related to that event.

PR #5453, opened 2024-06-14
Fix bug in interface object type collection during query planner construction

sachindshinde:
This PR fixes a bug where, during query planner construction, the interface types that are interface objects (in at least one subgraph schema) are computed incorrectly. This ends up fixing one particular snapshot test.


duckki (approved):
👍

PR #5463, opened 2024-06-17
Entity cache: fix Cache-Control max-age calculation

Geal:
The header was not properly updated according to the creation date of the cached entries

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-370]: https://apollographql.atlassian.net/browse/ROUTER-370?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apollo-router/src/plugins/cache/cache_control.rs
```diff
@@ -154,16 +160,23 @@ impl CacheControl {
     pub(crate) fn to_headers(&self, headers: &mut HeaderMap) -> Result<(), BoxError> {
         let mut s = String::new();
         let mut prev = false;
+        let now = now_epoch_seconds();
         if let Some(max_age) = self.max_age {
-            write!(&mut s, "{}max-age={}", if prev { "," } else { "" }, max_age)?;
+            //FIXME: write no-store if max_age = 0?
```
bnjjj, line 165: Does it need to be fixed at part of this PR ?

apollo-router/src/plugins/cache/entity.rs
```diff
@@ -372,14 +410,18 @@ impl InnerCacheService {
...
-                        c
-                    };
-                    update_cache_control(&response.context, &cache_control);
+                    let mut cache_control =
+                        if response.response.headers().contains_key(CACHE_CONTROL) {
+                            CacheControl::new(response.response.headers(), self.storage.ttl)?
+                        } else {
+                            let mut c = CacheControl::default();
+                            c.no_store = true;
+                            c
```
bnjjj, line 419:
```suggestion
                            CacheControl::no_store()
```

apollo-router/src/plugins/cache/entity.rs
```diff
@@ -389,12 +431,16 @@ impl InnerCacheService {
...
                         &mut response,
-                        cache_control,
+                        cache_control.clone(),
                         cache_result.0,
                         is_known_private,
                         private_id,
                     )
                     .await?;
+
+                    // FIXME: will the header be duplicated?
```
bnjjj, line 441: Does it need to be fixed as part of this PR ?

bnjjj (changes requested):

apollo-router/src/plugins/cache/cache_control.rs
Geal: yes, because the `max-age` value returned to the client was incorrect

apollo-router/src/plugins/cache/entity.rs
Geal: I just checked, the `HeaderMap::insert` method replaces all existing entries, so it is fine

apollo-router/src/plugins/cache/entity.rs
```diff
@@ -389,12 +429,16 @@ impl InnerCacheService {
...
```
Geal, line 439:
```suggestion
```

garypen (commented):
The changes look good, but this could use a few more tests. In particular, the functionality in CacheControl has one unit test, but it's doing a lot of stuff, and the test uses a test specific function. I'd like to see `can_use()` being tested at least.

The FIXMEs... Are they waiting for the entity cache invalidation project?

Geal:
yes, the FIXMEs were waiting for follow ups, when the project was scheduled again

Geal:
@garypen I added more tests, but stuff like `can_use` is truly validated only when running under integration tests, and I have planned for us to spend time on that

bnjjj (approved):

BrynCooke (commented):
There's been significant changes to `entity.rs` but no test was affected, this points to some missing testing somewhere. Maybe this is a good candidate for your new yaml driven test runner?

Geal:
the yaml test runner won't be able to handle `Cache-Control` calculations without significant changes

Geal:
while the integration tests are currently lacking, I have validated that it works properly with numerous manual tests

PR #5464, opened 2024-06-17
fix events when trace is not sampled, and keep event attribute types

### LINKED ISSUE #5448: telemetry: specific events attributes configured in custom telemetry won't be displayed if the trace is not sampled
I discovered that  bug when working on https://github.com/apollographql/router/pull/5447 I added a FIXME in the codebase and the fix will mainly to reproduce the same behavior than with `LogAttributes` struct in the span extensions when the trace is not sampled.

bnjjj:
Several fixes about events:

+ Keep original attribute type and not convert it to string (also perf improvement)
+ Use `http.response.body.size`  and `http.request.body.size` as a number and not a string
+ Display custom event attributes even if the trace is not sampled 

Fixes #5448 

> :warning: If you had some monitoring enabled on your logs please be cautious because with these changes, now instead of having for example an attribute like `http.response.status_code = "200"` it will be `http.response.status_code = 200` the 200 code will be a number and not a string anymore. That's an example. Same for `http.request|response.body.size` attribute

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [x] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


BrynCooke (approved):
Great! Needs a changelog detailing that the attribute types have changed etc. This may affect people who have set up monitors on their logs so it needs to be called out and also make this a minor bump rather than a patch. 
CC @jesse-apollo

bnjjj:
I added a changelog and add details. I'll merge this PR directly in the other branch with an opened PR https://github.com/apollographql/router/pull/5447

PR #5467, opened 2024-06-17
fix(federation): handle shareable root operations

dariuszkuc:
Fix state update logic for handling processed nodes. We were incorrectly retaining already processed nodes.

JS code `mergeRemaingsAndRemoveIfFound`
```javascript
return inEdges.filter(e => otherEdges.includes(e))
```

RS code `merge_remains_and_remove_if_found`
```rust
- in_edges.retain(|e| !other_node.unhandled_parents.contains(e));
+ in_edges.retain(|e| other_node.unhandled_parents.contains(e));
```

JS code `updateForProcessedGroups`
```javascript
const newEdges = edges.filter((edge) => !processed.includes(edge.group));
```

RS code `update_for_processed_nodes`
```rust
- edges.retain(|edge| processed.contains(&edge.parent_node_id));
+ edges.retain(|edge| !processed.contains(&edge.parent_node_id));
```

Also fixed incorrect parent<->child relationship in handle requires -> when creating new key fetch nodes we should make it a child of newly created nodes and NOT the opposite.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-federation/src/query_plan/fetch_dependency_graph.rs
+++ b/apollo-federation/src/query_plan/fetch_dependency_graph.rs
@@ -416,7 +416,7 @@ impl ProcessingState {
         {
             // Remove any of the processed nodes from the unhandled edges of that node.
             // And if there is no remaining edge, that node can be handled.
-            edges.retain(|edge| processed.contains(&edge.parent_node_id));
+            edges.retain(|edge| !processed.contains(&edge.parent_node_id));
             if edges.is_empty() {
                 if !next.contains(&g) {
                     next.push(g);
--- a/apollo-federation/src/query_plan/fetch_dependency_graph.rs
+++ b/apollo-federation/src/query_plan/fetch_dependency_graph.rs
@@ -3975,6 +3975,15 @@ fn handle_requires(
         let parent_type = new_node.parent_type.clone();
         for created_node_id in &new_created_nodes {
             let created_node = dependency_graph.node_weight(*created_node_id)?;
+            // Usually, computing the path of our new group into the created groups
+            // is not entirely trivial, but there is at least the relatively common
+            // case where the 2 groups we look at have:
+            // 1) the same `mergeAt`, and
+            // 2) the same parentType; in that case, we can basically infer those 2
+            //    groups apply at the same "place" and so the "path in parent" is
+            //    empty. TODO: it should probably be possible to generalize this by
+            //    checking the `mergeAt` plus analyzing the selection but that
+            //    warrants some reflection...
             let new_path =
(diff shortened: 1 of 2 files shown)


dariuszkuc:
note: this change had a side effect that 3 other tests started failing with the `Root nodes should have no remaining nodes unhandled`. It appears that we end up with incorrect parent<->child relationship between nodes. Still looking into that.

new failures:
* query_plan::build_query_plan_tests::interface_object::does_not_rely_on_an_interface_object_directly_if_a_specific_implementation_is_requested
* query_plan::build_query_plan_tests::interface_object::it_avoids_buffering_interface_object_results_that_may_have_to_be_filtered_with_lists
* query_plan::build_query_plan_tests::requires::handles_multiple_requires_involving_different_nestedness

apollo-federation/src/query_plan/fetch_dependency_graph.rs
```diff
@@ -3975,27 +3975,26 @@ fn handle_requires(
...
             let created_node = dependency_graph.node_weight(*created_node_id)?;
+            // Usually, computing the path of our new group into the created groups
+            // is not entirely trivial, but there is at least the relatively common
+            // case where the 2 groups we look at have:
+            // 1) the same `mergeAt`, and
+            // 2) the same parentType; in that case, we can basically infer those 2
+            //    groups apply at the same "place" and so the "path in parent" is
+            //    empty. TODO: it should probably be possible to generalize this by
+            //    checking the `mergeAt` plus analyzing the selection but that
+            //    warrants some reflection...
```
lrlna, line 3986: please make a task for this in Jira so we don't lose this thought in a comment.

lrlna (approved):
That's some really good investigative work, @dariuszkuc!

Would you be able to add a few more snapshots that could potentially run into a similar issue? Having only one that caught this error seems like a bit of a smell 😞

apollo-federation/src/query_plan/fetch_dependency_graph.rs
dariuszkuc: this is old Sylvain's note ported over

dariuszkuc:
> Would you be able to add a few more snapshots that could potentially run into a similar issue? Having only one that caught this error seems like a bit of a smell 😞

Bug in handle requires was triggering this failure for 3 other tests. Fixing only `update_for_processed_nodes` without `merge_remains_and_remove_if_found` was triggering another failure as well. So there is some overlap between those tests testing the condition.

PR #5469, opened 2024-06-18
incremental improvements to merge_subgraphs for connectors

lennyburdette:
This is a small start to support basic connectors:
* Switches to using ValidFederationSubgraph so we have access to link metadata
* Uses the imported name from `@link`
* Ensures that join__field is added to all relevant fields (tho this now adds it more aggressively than it should, but that optimization is a much larger change — this required some snapshot update but doesn't semantically change the result)
* Adds a test

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


dariuszkuc (approved):

apollo-federation/src/merge.rs
```diff
@@ -81,6 +92,33 @@ impl Debug for MergeFailure {
...
+                        errors: vec![e.to_string()],
+                        composition_hints: Default::default(),
+                    }
+                })?,
+            })
+            .map_err(|e| MergeFailure {
+                schema: None,
+                errors: vec![e.to_string()],
+                composition_hints: Default::default(),
+            })?;
```
pubmodmatt, line 114: There is a `From<FederationError>` implementation `MergeFailure`, so you can simply remove the boilerplate `map_err` implementations.

```suggestion
       federation_subgraphs.add(ValidFederationSubgraph {
            name: subgraph.name.clone(),
            url: subgraph.url.clone(),
            schema: ValidFederationSchema::new(subgraph.schema.clone())?,
        })?;
```

apollo-federation/src/merge.rs
```diff
@@ -1259,3 +1350,127 @@ fn merge_directive(
...
+            })
+            .unwrap();
+
+        let result = merge_federation_subgraphs(subgraphs).unwrap();
+
+        let schema = result.schema.into_inner();
+        let validation = schema.clone().validate();
+        assert!(validation.is_ok(), "{:?}", validation);
+
+        assert_snapshot!(schema.serialize(), @r###"
```
pubmodmatt, line 1422: Why inline this here rather than having it in a snapshot file?

lennyburdette: yeah this could definitely be a file, no reason for it to be inline in this case

apollo-federation/src/merge.rs
lennyburdette: good catch, thanks!

pubmodmatt (approved):

PR #5470, opened 2024-06-18
FED-313: ported the remaining operation optimize unit tests involving empty selection sets

duckki:
To test `without_empty_branches` method, we need to test operations with empty selection sets. However, such operations can't be constructed from strings, since the parser will reject them. The JS tests use `astSSet` function to construct queries with empty selection sets using graphql-js's `SelectionSetNode` API. In Rust version, instead of re-creating such API, we first create a valid query with non-empty selection sets and then selectively clear them.

Fixes https://apollographql.atlassian.net/browse/FED-313

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

n/a

**Notes**


lrlna (approved):
More of general musings rather than anything to change in this PR.

It seems like at this point we are writing up a lot of functions specific for testing operations (which is *great*, will make unit testing *heaps* easier), but is getting me to think we might want a more structurally solid organisation for these. Should some of the functions like `parse_operation` be more available? They could potentially live with `apollo_federation::operation` in a function annotated with `#[cfg(test)]` to make sure it can only be called from tests? Basically trying to think about what you've done in `optimize.rs` and how we can do more of this elsewhere.

duckki:
> More of general musings rather than anything to change in this PR.
> 
> It seems like at this point we are writing up a lot of functions specific for testing operations (which is _great_, will make unit testing _heaps_ easier), but is getting me to think we might want a more structurally solid organisation for these. Should some of the functions like `parse_operation` be more available? They could potentially live with `apollo_federation::operation` in a function annotated with `#[cfg(test)]` to make sure it can only be called from tests? Basically trying to think about what you've done in `optimize.rs` and how we can do more of this elsewhere.

Good point. I think `operation::tests` module is good home for these functions. I've moved some functions there.

PR #5474, opened 2024-06-18
Entity cache: reorganize subgraph configuration override

Geal:
**breaking change** this updates the entity cache configuration to reuse the all/subgraphs pattern of configuration override used in other parts of the router.

Since this is a preview feature, do we still make a configuration migration for it?

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-369]: https://apollographql.atlassian.net/browse/ROUTER-369?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bnjjj (approved):
Missing a changelog, but otherwise LGTM

garypen (approved):

IvanGoncharov (approved):

apollo-router/src/plugins/cache/entity.rs
```diff
@@ -67,34 +68,30 @@ pub(crate) struct Config {
     /// activates caching for all subgraphs, unless overriden in subgraph specific configuration
```
BrynCooke, line 68:
```suggestion
    /// Enable or disable the the entity caching feature
```

BrynCooke (changes requested):

BrynCooke:
@Geal Can you make a generic struct to hold `all` and `subgraphs` that can be reused? You've mentioned before that we should have a standard way of doing this stuff and this would help us.

Geal:
> @Geal Can you make a generic struct to hold all and subgraphs that can be reused? You've mentioned before that we should have a standard way of doing this stuff and this would help us.
There is, it is called `SubgraphConfiguration`, it is used in this PR and defined here: https://github.com/apollographql/router/blob/dd9005e5cdca5090af049f2d236e3a3a47a89e34/apollo-router/src/configuration/subgraph.rs#L78-L88

Geal:
I made the following changes in [bfb7fbe](https://github.com/apollographql/router/pull/5474/commits/bfb7fbe0dd81be6886873c9bd8b5a3b7a5d08e2a) and [d36870d](https://github.com/apollographql/router/pull/5474/commits/d36870d968cb1b29209cdc1253ac010c0e158e57):
- the top level `enabled` is explicitely a precondition to enabling entity cache on any subgraph
- uses the `SubgraphConfiguration::get` method to fall back to the `all` configuration if the subgraph is not present (per field override between `all` and `subgraphs` is done in the `Deserialize` implementation)

BrynCooke (approved):

PR #5476, opened 2024-06-18
feat(telemetry): add support of condition on standard events

### LINKED ISSUE #5475: Give to standard events the ability to be displayed conditionally
Instead of only provide the level I really would like to be able to have conditions on these standard events. Example of configuration:

```yaml
telemetry:
  instrumentation:
    events:
      router:
        request:
          level: info
          condition:
            eq:
            - request_header: x-log-request
            - "enabled"
        response: off
        error: error
```

bnjjj:
Now, you can also enable these standard events based on conditions.

For example:

```yaml title="router.yaml"
telemetry:
  instrumentation:
    events:
      router:
        request:
          level: info
          condition: # Only log the router request if you sent `x-log-request` with the value `enabled`
            eq:
            - request_header: x-log-request
            - "enabled"
        response: off
        error: error
        # ...
```

Fixes #5475

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [x] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/plugins/telemetry/config_new/events.rs
```diff
@@ -706,6 +800,28 @@ mod tests {
                 )
                 .await
                 .expect("expecting successful response");
+            test_harness
```
BrynCooke, line 803: Nit separate this out into a different test.

docs/source/configuration/telemetry/instrumentation/events.mdx
```diff
@@ -60,6 +60,26 @@ telemetry:
         # ...
 ```
 
+But you can also enable these standard events based on [conditions](./conditions) (not supported on batched requests).
```
BrynCooke, line 63: Let's add examples for logging when there are graphql errors.

apollo-router/src/plugins/telemetry/config_new/events.rs
```diff
@@ -664,9 +753,14 @@ mod tests {
                         .build()
                         .unwrap(),
                     |_r| {
+                        let context_with_error = Context::new();
```
BrynCooke, line 756: Let's separate this out into a different test, and also add tests for response headers.

BrynCooke (changes requested):
We need to add some more tests. In particuar, conditions that are taken from the response.

BrynCooke (approved):

PR #5481, opened 2024-06-18
fix: misc graph_path display fixes

dariuszkuc:
Fixes
* `SimultaneousPaths` should not wrap single element in redundant brackets `{ }`
* `OpGraphPath` added missing conditions to the edges
* `OpGraphPath` added missing root nodes

Before
```
{ --[key()]--> I(S1) --[__typename]--> String(S1) }
```

After
```
I(S2) --[{ id } ⊢ key()]--> I(S1) --[__typename]--> String(S1)
```

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-federation/src/query_graph/graph_path.rs
+++ b/apollo-federation/src/query_graph/graph_path.rs
@@ -3015,10 +3015,11 @@ impl Display for OpGraphPath {
         // If the path is length is 0 return "[]"
         // Traverse the path, getting the of the edge.
         let head = &self.graph.graph()[self.head];
-        if head.is_root_node() && self.edges.is_empty() {
+        let head_is_root_node = head.is_root_node();
+        if head_is_root_node && self.edges.is_empty() {
             return write!(f, "_");
         }
-        if !head.is_root_node() {
+        if !head_is_root_node {
             write!(f, "{head}")?;
         }
         self.edges
@@ -3029,7 +3030,7 @@ impl Display for OpGraphPath {
                 Some(e) => {
                     let tail = self.graph.graph().edge_endpoints(e).unwrap().1;
                     let node = &self.graph.graph()[tail];
-                    if i == 0 && head.is_root_node() {
+                    if i == 0 && head_is_root_node {
                         write!(f, "{node}")
                     } else {
                         let edge = &self.graph.graph()[e];
--- a/apollo-federation/src/query_graph/graph_path.rs
+++ b/apollo-federation/src/query_graph/graph_path.rs
@@ -3033,7 +3033,10 @@ impl Display for OpGraphPath {
                     let label = edge.transition.to_string();
                     if let Some(conditions) = &edge.conditions {
                         write!(f, " --[{conditions} ⊢ {label}]--> {node}")
-                    } else if !matches!(edge.transition, QueryGraphEdgeTransition::SubgraphEnteringTransition) {
+                    } else if !matches!(
+                        edge.transition,
+                        QueryGraphEdgeTransition::SubgraphEnteringTransition
+                    ) {
                         write!(f, " --[{label}]--> {node}")
                     } else {
                         core::fmt::Result::Ok(())
--- a/apollo-federation/src/query_graph/graph_path.rs
+++ b/apollo-federation/src/query_graph/graph_path.rs
@@ -544,7 +544,7 @@ impl SimultaneousPaths {
         match self.0.as_slice() {
             [] => f.write("<no path>"),
 
-            [first] => f.write_fmt(format_args!("{{ {first} }}")),
+            [first] => f.write_fmt(format_args!("{first}")),
 
             _ => {
                 f.write("{")?;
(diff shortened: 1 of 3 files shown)


dariuszkuc:
There is still a mismatch in how we display graph paths
* JS always adds the head info, e.g. above path should be ` I(S2) --[{ id } ⊢ key()]--> I(S1) --[__typename]--> String(S1)`
* root nodes are not printed correctly, e.g. 
```
Query(S2) --[iFromS2]--> I(S2) --[{ __typename } ⊢ ... on A]--> I(S2) --[y]--> Int(S2)
// vs
[query](_)* --[iFromS2]--> I(S2) --[{ __typename } ⊢ ... on A]--> I(S2) --[y]--> Int(S2)
```

Display for `QueryGraphNode` seems [correct](https://github.com/apollographql/router/blob/v1.49.0/apollo-federation/src/query_graph/mod.rs#L77-L88) but we format the type [incorrectly](https://github.com/apollographql/router/blob/v1.49.0/apollo-federation/src/query_graph/mod.rs#L96-L105) with extra `[]` brackets and lowercase string which is [derived](https://github.com/apollographql/router/blob/dev/apollo-federation/src/schema/position.rs#L1259-L1266).

Questions
* graph node source is `_` -> should it be pointing to S2?
```
QueryGraphNode { type_: FederatedRootType(Query), source: "_", has_reachable_cross_subgraph_edges: true, provide_id: None, root_kind: Some(Query) }
```
* interestingly even though it is root vertex why don't don't we print the `*` in the JS version?

apollo-federation/src/query_graph/graph_path.rs
```diff
@@ -544,7 +544,7 @@ impl SimultaneousPaths {
         match self.0.as_slice() {
             [] => f.write("<no path>"),
 
-            [first] => f.write_fmt(format_args!("{{ {first} }}")),
+            [first] => f.write_fmt(format_args!("{first}")),
```
duckki, line 547: It's my bad. Nice catch!

duckki (changes requested):
The change looks good!

This part doesn't look inconsistent with JS version, which might be related to your question.
```rust
        if head.root_kind.is_some() {
            write!(f, "{head}")?;
        }
```

The JS version (below) does NOT print the head if it is root (see the last line).
If the head is root, it's skipped and the first edge's tail is printed first without the arrow.

```js
  toString(): string {
    const isRoot = isRootVertex(this.root);
    if (isRoot && this.size === 0) {
      return '_';
    }
    const pathStr = this.mapMainPath((edge, idx) => {
      if (edge) {
        if (isRoot && idx == 0) {
          return edge.tail.toString();
        }
        const label = edge.label();
        return ` -${label === "" ? "" : '-[' + label + ']-'}-> ${edge.tail}`
      }
      return ` (${this.props.edgeTriggers[idx]}) `;
    }).join('');
    const deferStr = this.deferOnTail ? ` <defer='${this.deferOnTail.label}'>` : '';
    const typeStr = this.props.runtimeTypesOfTail.length > 0 ? ` (types: [${this.props.runtimeTypesOfTail.join(', ')}])` : '';
    return `${isRoot ? '' : this.root}${pathStr}${deferStr}${typeStr}`;
  }
```

dariuszkuc:
Yeah I was looking into this a bit more and the JS code
- by default it adds the root node IF it is not root -> 
```
return `${isRoot ? '' : this.root}${pathStr}${deferStr}${typeStr}`;
```

There is a difference in closed branches though
- in JS we start at (head) `[query](_)*` (entry to supergraph) so when iterating over the edges (always start with head) -> if it `isRoot` we just use the [tail of the edge](https://github.com/apollographql/federation/blob/%40apollo/subgraph%402.8.1/query-graphs-js/src/graphPath.ts#L891) (without any labels which is just a vertex and results in the `Query(S2)` entry point) 
- in RS side of things we do start at the supergraph entry but it appears that we are missing this extra `Query(S2)` node

apollo-federation/src/query_graph/graph_path.rs
```diff
@@ -3029,9 +3029,23 @@ impl Display for OpGraphPath {
                 Some(e) => {
                     let tail = self.graph.graph().edge_endpoints(e).unwrap().1;
                     let node = &self.graph.graph()[tail];
-                    let edge = &self.graph.graph()[e];
-                    let label = edge.transition.to_string();
-                    write!(f, " --[{label}]--> {node}")
+                    if i == 0 && head.root_kind.is_some() {
+                        write!(f, "{node}")
```
dariuszkuc, line 3033: this still adds extra `*` at the end of the subgraph transition node i.e. `Query(S2)` vs `Query(S2)*`

dariuszkuc: seems like a bug as in JS that node is not marked as root

duckki (approved):
This display fix looks good to me!

apollo-federation/src/query_graph/graph_path.rs
dariuszkuc: we were incorrectly looking at the `root_kind` struct field  vs `is_root_node()` function value

PR #5494, opened 2024-06-20
Try to minimise work in DurationHistogram

garypen:
Currently we maintain a total for each instance, but we only access the total once. Let's not maintain a total. We'll just derive it from the individual buckets when we need it.


garypen:
Hard to assess the impact of the change, but the perf results looked fine and the tests all pass.

bnjjj (approved):

Geal (approved):

BrynCooke:
Please don't merge this as I'm making some big changes to histogram. I'll fold in this change though.

garypen:
I managed to do some meaningful performance testing (with my new testing framework) of this PR and it shows a substantial regression in performance for a test which generates a lot of Apollo Telemetry. The test hits a router configured to send Apollo Telemetry to a local "dump process" and then hits it with 5 clients each making 300 requests/second (1,500 aggregate).

Here's what `dev` looks like for one of the clients (no significant variations between clients):
```
Requests      [total, rate, throughput]         6000, 300.06, 256.13
Duration      [total, attack, wait]             23.425s, 19.996s, 3.429s
Latencies     [min, mean, 50, 90, 95, 99, max]  37.41ms, 2.091s, 251.145ms, 6.674s, 10.167s, 15.85s, 20.772s
Bytes In      [total, mean]                     417608, 69.60
Bytes Out     [total, mean]                     2211000, 368.50
Success       [ratio]                           100.00%
Status Codes  [code:count]                      200:6000  
Error Set:
```
Note: mean latency is around 2 seconds.

The same report for one of the clients against router built against this branch:
```
Requests      [total, rate, throughput]         6000, 300.05, 220.97
Duration      [total, attack, wait]             27.153s, 19.996s, 7.157s
Latencies     [min, mean, 50, 90, 95, 99, max]  62.826ms, 3.934s, 723.954ms, 12.373s, 14.724s, 19.067s, 22.678s
Bytes In      [total, mean]                     408149, 68.02
Bytes Out     [total, mean]                     2266549, 377.76
Success       [ratio]                           100.00%
Status Codes  [code:count]                      200:6000  
Error Set:

```
Note: mean latency is now 3.9 seconds.

Throughput is reduced from 256 to 220.

I'll close this PR since this is conclusive evidence that the change isn't likely to be a good change.

PR #5501, opened 2024-06-21
Add enum value references from responses to extended references

bonnici:
* Updates the enhanced reference experimental feature so that we now also send references to enums from response objects in the reports.
* Adds some usage and config metrics related to the studio interop functionality.
* Changeset is covered in https://github.com/apollographql/router/pull/5331

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-375]: https://apollographql.atlassian.net/browse/ROUTER-375?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apollo-router/src/apollo_studio_interop/mod.rs
```diff
@@ -230,6 +231,7 @@ impl ComparableUsageReporting {
 /// Generate a ComparableUsageReporting containing the stats_report_key (a normalized version of the operation signature)
 /// and referenced fields of an operation. The document used to generate the signature and for the references can be
 /// different to handle cases where the operation has been filtered, but we want to keep the same signature.
+#[tracing::instrument(skip_all, level = "debug")]
```
Geal, line 234: are all the tracing annotations still needed?

Geal: each of those annotations creates a new span, and there's some overhead for each created span even if not at a visible level

apollo-router/src/apollo_studio_interop/mod.rs
```diff
@@ -279,6 +282,117 @@ pub(crate) fn generate_extended_references(
...
+fn add_enum_value_to_map(
+    enum_name: &Name,
+    enum_value: &JsonValue,
+    referenced_enums: &mut ReferencedEnums,
+) {
+    match enum_value {
+        JsonValue::String(val_str) => {
+            let enum_name_str = enum_name.to_string();
+            // Not using entry API here due to performance impact
+            let enum_name_stats = match referenced_enums.get_mut(&enum_name_str) {
```
Geal, line 313:
```suggestion
            // Not using entry API here due to performance impact
            let enum_name_stats = match referenced_enums.get_mut(enum_name.as_str()) {
```

Geal: if we're not using the entry API, then there's no need to preallocate the key

apollo-router/src/apollo_studio_interop/mod.rs
```diff
@@ -279,6 +282,117 @@ pub(crate) fn generate_extended_references(
...
+    }
+}
+
+fn extract_enums_from_selection_set(
+    selection_set: &[SpecSelection],
+    fragments: &Fragments,
+    schema: &Valid<Schema>,
+    selection_response: &Object,
+    result_set: &mut ReferencedEnums,
+) {
```
Geal, line 340: this feels like the visitor created for demand control: https://github.com/apollographql/router/blob/dev/apollo-router/src/graphql/visitor.rs

@BrynCooke we should probably update this to move to the visitor after we merge (to make sure it won't block the release)

apollo-router/src/services/execution/service.rs
```diff
@@ -321,8 +339,32 @@ impl ExecutionService {
...
+            }
+
+            if let (ApolloMetricsReferenceMode::Extended, Some(Value::Object(response_body))) = (metrics_ref_mode, &response.data) {
+                extract_enums_from_response(
+                    query.clone(),
+                    operation_name,
+                    schema.api_schema(),
+                    response_body,
+                    &mut referenced_enums,
+                );
```
Geal, line 361: this should not be done twice. `format_response` happens twice because it first has to go through the filtered query (with fields removed due to authorization), then the client query, to get the expected shape by inserting nulls were appropriate. `extract_enums_from_response` should happen only once, and after those steps

Geal (changes requested):
This looks fine overall, but 2 things should be done before merging:
- removing the tracing annotations that create debug level spans
- extracting enums only once, and after the 2 passes of response formatting have happened

apollo-router/src/apollo_studio_interop/mod.rs
bonnici: I thought they would be useful information to have in the trace view, but can remove if it's a performance issue. Are the spans created even if you're not running the router with `--log=trace`?

apollo-router/src/apollo_studio_interop/mod.rs
bonnici: I tried using a form of the visitor pattern a while ago but I can't remember why it didn't work for me. Maybe it was not available in router source before and just in federation, or maybe I'm using something that's not available from the visitor. In any case I'm happy to revisit this once all the changes are merged.

apollo-router/src/services/execution/service.rs
bonnici: So does the `query` contain all the enums that are also in `filtered_query` at this point in the code?

Geal: if an enum is annotated with `@authenticated` or other authorization directives and the client is not authorized, then the filtered query will not have the fields of that enum type. Do you want to measure what is obtained in query plan execution, or what is returned to the client?

bonnici: The reason for returning the enum usage is mainly so we can see if it's safe to remove an unused enum, so I think it's better to use the query plan execution.

o0Ignition0o (approved):

Geal (approved):

PR #5502, opened 2024-06-21
Added integrity checks on operations and selection sets

duckki:
Operations and selection/selection-sets has internal invariants (before considering its correctness or any other constraints).

I call it "well-formedness" of operation/selection/-set structs. For example:
- Every node has the same schema.
- Every selection set must have a type that is dictated by their parent selection (namely, field's base type and inline fragment's "casted" type).
- All fields must be defined in the schema.
- All fragment spreads must be defined in the NamedFragments.
- Every inline fragment spread's `parent_type_position` must be the type of its parent selection set.

I implemented the check and placed it in several places where we build operations. Those checks are only executed in debug builds.

I found a bug in `rebase_on` and fixed it in this PR. Though, that fix didn't change the status of existing tests.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [x] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[FED-246]: https://apollographql.atlassian.net/browse/FED-246?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


TylerBloom:
I really, really like the listing of invariants and checking them. We should start a doc that lists these things out.

TylerBloom (approved):
It seems we don't want to call the `is_well_formed` method in production, but it might be something we want to enable based on log level as it would provide never helpful information. This isn't something you need to change in this PR, but I wanted to note it.

apollo-federation/src/operation/integrity.rs
```diff
@@ -0,0 +1,202 @@
...
+        }
+        for selection in self.iter() {
+            selection.is_well_formed(schema, named_fragments, &self.type_position, option)?;
+        }
+        Ok(())
+    }
+}
+
+impl Operation {
+    pub fn is_well_formed(
```
sachindshinde, line 188: I'm noticing that:
1. The well-formedness check for fragment spread selections recurses into the fragment's selection sets, each time the fragment spread is referenced (assuming they're not skipped).
2. We don't check the well-formedness of `Operation.named_fragments` here. I'm guessing because we assume they should be exactly the same as the fragments referenced above (although we're not checking this here).

Looking at the codebase, the `FragmentSpreadSelection.selection_set` field, when directly set, comes from existing `Fragment`s. That means that it should always match the corresponding `Fragment`'s selection set in the operation. And not just `Eq` here for `SelectionSet`, but more specifically `SelectionSet.schema` and `SelectionSet.type_position` should be `Eq`, and `SelectionSet.selections` should compare the same using `Arc::ptr_eq()`.

This would suggest that we change the above validation to:
1. Validate the fragments in `Operation.named_fragments` (recursing into their selection sets, but also e.g. checking that `Fragment.type_condition_position` matches the one in the selection set).
2. When validating a fragment spread selection, we should check that the `FragmentSpreadSelection.selection_set` matches in the way described above.

There's a similar case to be made for `FragmentSpreadSelection.spread`, where it always comes from feeding a `Fragment` to `FragmentSpreadData::from_fragment()`. This means we should similarly check matching there, i.e. we should check for `Eq` for `FragmentSpreadData.schema`, `FragmentSpreadData.fragment_name`, and `FragmentSpreadData.type_condition_position`, and check `Arc::ptr_eq()` for `FragmentSpreadData::fragment_directives()`.
- While looking here, I noticed a bug in `FragmentSpreadSelection::new()` where it re-fetches its fragment selection set from `NamedFragments`, but doesn't update the spread data. Glancing at the code, this looks to be used only for `SelectionSet::make_selection()`, specifically in `Selection::from_operation_element()`. The latter function appears to be a rename of `selectionOfElement()` from JS, but the JS code converts fragment spreads to inline fragments, while the Rust code appears to be keeping them as fragment spreads. It would be better I think here to copy the JS code's behavior (or error when fragment spreads are present, if we believe they shouldn't occur).

sachindshinde (changes requested):

goto-bus-stop:
Changing to draft so we don't get pinged to rereview it every day in slack 😇

PR #5508, opened 2024-06-21
AWS Security Token Service: Implement a manual cache on top of assume_role and credentials provider

o0Ignition0o:
While the STS `CredentialsProvider` chain has a cache, it is not the case for `AssumeRoleProvider`.

This changeset introduces a custom CredentialsProvider that operates as a caching layer, with a couple of business rules related to it:

1. When credentials are retrieved, they will be kept in cache for:
     - `credentials.expiry()` if it is set
     - `ever` if not
2. 5 minutes before credentials get removed from cache, we will try to retrieve new ones.
3. Failure to retrieve credentials will trigger a new attempt after `1 minute`
4. `CredentialsProvider` exposes a `refresh_credentials()` function that could be used to manually trigger a refresh, say if the subgraph call yields a `401` (TODO as a followup, not part of this changeset)


apollo-router/src/plugins/authentication/subgraph.rs
```diff
@@ -192,12 +198,100 @@ pub(crate) struct SigningParams {
...
+            credentials_provider.provide_credentials().await?,
+        ));
+        let c2 = credentials.clone();
+        let handle = tokio::spawn(async move {
+            loop {
+                tokio::select! {
+                    _ = tokio::time::sleep(refresh_timer) => {
+                        // Refresh credentials
+                        if let Ok(new_credentials) = credentials_provider.provide_credentials().await {
+                            refresh_timer = next_refresh_timer(&new_credentials);
```
nmoutschen, line 236: `refresh_timer` is a `Duration`, which means that on `Err` we will wait again for `X`, `X` potentially being greater than the validity period of the credentials. Should we handle `Err`s here so we have a shorter refresh timer window?

apollo-router/src/plugins/authentication/subgraph.rs
```diff
@@ -192,12 +198,100 @@ pub(crate) struct SigningParams {
...
-    credentials_provider: Arc<dyn ProvideCredentials>,
+    credentials_provider: CredentialsProvider,
     region: Region,
     service_name: String,
     subgraph_name: String,
 }
 
+#[derive(Clone, Debug)]
+struct CredentialsProvider {
+    credentials: Arc<RwLock<Credentials>>,
```
nmoutschen, line 209: How should we handle failures to refresh credentials for X period of time, exceeding the validity of those credentials?

apollo-router/src/plugins/authentication/subgraph.rs
```diff
@@ -192,12 +198,100 @@ pub(crate) struct SigningParams {
...
+fn next_refresh_timer(credentials: &Credentials) -> Duration {
+    credentials
+        .expiry()
+        .and_then(|expiry| {
+            expiry
+                .duration_since(SystemTime::now())
+                .ok()
+                .and_then(|d| d.checked_sub(MIN_REMAINING_DURATION))
+        })
+        .unwrap_or_else(|| Duration::from_secs(0))
```
nmoutschen, line 276: This will cause the router to try to refresh credentials in a loop if there are no expiry value. Long-lived credentials (e.g. AWS IAM Users credentials) won't have an expiry value, which would cause this to spin in a loop.

Should we treat credentials that never expire differently?

o0Ignition0o: Great question! We could use the default (15 minutes), which shouldn't put too much pressure on the aws systems.

apollo-router/src/plugins/authentication/subgraph.rs
o0Ignition0o: Great idea, i went and used a 1 minute retry timer, we'll see how it goes!

apollo-router/src/plugins/authentication/subgraph.rs
o0Ignition0o: Good question, `refresh_credentials()` can be called manually if we end up in such a situation, the plan is to intercept sts 401s and call refresh_credentials(), so this + the refresh mechanism should give us enough guardrails.

apollo-router/src/plugins/authentication/subgraph.rs
```diff
@@ -192,12 +198,111 @@ pub(crate) struct SigningParams {
...
+    credentials: Arc<RwLock<Credentials>>,
+    _credentials_updater_handle: Arc<JoinHandle<()>>,
+    #[allow(dead_code)]
+    refresh_credentials: Sender<()>,
+}
+
+// Refresh token if it will expire within the next 5 minutes
+const MIN_REMAINING_DURATION: Duration = std::time::Duration::from_secs(60 * 5);
+// If the token doesn't have a validity duration, default to 15 minutes
+const FALLBACK_TOKEN_DURATION: Duration = std::time::Duration::from_secs(60 * 15);
```
nmoutschen, line 218: __Nit__: if a `Credentials` doesn't have an expiry, it means that it never expires. So we could potentially remove the fallback if the `expires_after` value is `None`.

See: https://docs.rs/aws-credential-types/1.2.0/src/aws_credential_types/credentials_impl.rs.html#37

nmoutschen (approved):

apollo-router/src/plugins/authentication/subgraph.rs
o0Ignition0o: I see, and if we manually invalidate the token, subgraph calls will yield 401s so users could still be able to diagnose this. lgtm!

bnjjj (approved):
Should we document this behavior ?

o0Ignition0o:
> Should we document this behavior ?
Great question! the behavior is fairly close to the sts client's default (it just forgets to cache assume role), maybe as a followup?

PR #5513, opened 2024-06-21
docs: Add extended reference reporting config

Meschreiber:
Adds docs for extended reference reporting.


docs/source/configuration/overview.mdx
```diff
@@ -742,6 +742,33 @@ query AliasedQuery {
...
+
+Configure extended reference reporting in `router.yaml` with the `telemetry.apollo.experimental_apollo_metrics_reference_mode` option like so:
+
+```yaml title="router.yaml"
+telemetry: 
+  apollo:
+    experimental_apollo_metrics_reference_mode: extended # Default is legacy
+```
+
+Once you enable this configuration, you can view enum and input usage along with other types of [field usage in GraphOS Studio](/graphos/metrics/field-usage).
```
Meschreiber, line 764: Note: This docs page will be updated in a separate PR to reflect the new UI/capabilities.

docs/source/configuration/overview.mdx
```diff
@@ -742,6 +742,33 @@ query AliasedQuery {
 }
 ```
 
+<MinVersion version="1.50.0">
+
+### Extended reference reporting
+
+</MinVersion>
+
+<ExperimentalFeature />
```
bonnici, line 751: This one is enterprise-only as well.

docs/source/configuration/overview.mdx
```diff
@@ -742,6 +742,33 @@ query AliasedQuery {
...
+  apollo:
+    experimental_apollo_metrics_reference_mode: extended # Default is legacy
+```
+
+Once you enable this configuration, you can view enum and input usage along with other types of [field usage in GraphOS Studio](/graphos/metrics/field-usage).
+
+<Note>
+
+Once you enable `extended` reference reporting, there will be a lag in changes to metrics in Studio.
+This is because historical operations still use the `legacy` reference reporting.
```
bonnici, line 769: There shouldn't be a delay in metrics in Studio. But enabling this also allows us to run "enhanced checks" which is better at reducing the number of false-positives like the removal of unused enums and input fields, and there is a delay there because we can only run these enhanced checks if the flag has been enabled for the entire checks period. I can go into this further with you if you want.

Meschreiber: Ah, that makes sense, ty!

bonnici: They'll be automatic once the config has been turned on for the entire checks duration. As part of the check we run a query to check that the "enhanced reference reporting" flag has been turned on in usage reports, which happens when the flag is turned on.

bonnici: Also to clarify the "why", the new checks work mostly by detecting when enums or input values are not used, so we need to check that the setting is on for a while to be sure that we don't have the stats about them because they're actually not used, and not because they're just not included in "legacy" reports.

abernix:
@bonnici @Meschreiber As this wasn't merged onto `dev` already, should this target our just created `1.50.0` release branch?

Meschreiber:
Thanks @jesse--planning on completing it for review from @bonnici today and will retarget 1.50!

Meschreiber:
@abernix , I'm note seeing said `1.50.0` branch yet, so @bonnici , will you just review on here and then I'll retarget or create a new docs PR once `1.50.0` is available.

(Unless I'm missing something!)

abernix:
apologies. miscommunication upstream! We will make the 1.50.0 branch **tomorrow**.

docs/source/configuration/overview.mdx
```diff
@@ -742,6 +742,96 @@ query AliasedQuery {
...
+</tr>
+<tr>
+<td>
+
+##### Default argument changes for input object fields
+
+</td>
+
+<td>Changing or removing a default argument is generally considered a breaking change, but changing or removing default values for input object fields isn't.</td>
+<td>Changing or removing default values for input object fields is considered a breaking change.</td>
```
bonnici, line 808: Can we maybe add some information here about how there is a checks config option to ignore default value changes (not removals)? See [this slide](https://docs.google.com/presentation/d/19waRBNbd_-62r9uHLg4AHtQPNz34fmH215TKcxIhR7k/edit#slide=id.g28f84195726_0_28) for the UI - it's at the bottom of the checks config page.

bonnici (commented):
This explanation of the new checks functionality is great! Just one thing we could potentially add.

docs/source/configuration/overview.mdx
Meschreiber: Good call! Updated here: https://github.com/apollographql/router/pull/5513/commits/a8d9215c4cefc973ae9b11f97c92e45f54e2d9f1

I'll update the [ignored condition settings docs](https://www.apollographql.com/docs/graphos/delivery/schema-checks/#ignored-conditions-settings) to call out the "not removals" bit.

PR #5525, opened 2024-06-25
add basic support for overrideLabel in extract_subgraphs_from_supergraph

lennyburdette:
---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-federation/src/merge.rs
+++ b/apollo-federation/src/merge.rs
@@ -37,10 +37,15 @@ use itertools::Itertools;
 
 use crate::error::FederationError;
 use crate::link::federation_spec_definition::FEDERATION_EXTERNAL_DIRECTIVE_NAME_IN_SPEC;
+use crate::link::federation_spec_definition::FEDERATION_FIELDS_ARGUMENT_NAME;
+use crate::link::federation_spec_definition::FEDERATION_FROM_ARGUMENT_NAME;
 use crate::link::federation_spec_definition::FEDERATION_INTERFACEOBJECT_DIRECTIVE_NAME_IN_SPEC;
 use crate::link::federation_spec_definition::FEDERATION_KEY_DIRECTIVE_NAME_IN_SPEC;
+use crate::link::federation_spec_definition::FEDERATION_OVERRIDE_DIRECTIVE_NAME_IN_SPEC;
+use crate::link::federation_spec_definition::FEDERATION_OVERRIDE_LABEL_ARGUMENT_NAME;
 use crate::link::federation_spec_definition::FEDERATION_PROVIDES_DIRECTIVE_NAME_IN_SPEC;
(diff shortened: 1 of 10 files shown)


dariuszkuc:
Do we also need a guard that this is only applicable to fed 2.7+ version?

lennyburdette:
i can add a guard but since this is just for the in-memory versions of subgraphs extracted from the supergraph, i don't think it matters?

apollo-federation/src/merge.rs
```diff
@@ -510,6 +515,17 @@ impl Merger {
                     },
                 );
 
+                let overrides_directive_option = Option::and_then(
+                    field.directives.get_all(&override_directive_name).next(),
+                    |p| {
+                        let overrides_from =
+                            directive_string_arg_value(p, &name!("from")).map(|v| v.as_str());
+                        let ovevrrides_label =
+                            directive_string_arg_value(p, &name!("label")).map(|v| v.as_str());
```
dylan-apollo, line 524: Should these use the fed spec consts?

apollo-federation/src/merge.rs
```diff
@@ -510,6 +515,17 @@ impl Merger {
                     },
                 );
 
+                let overrides_directive_option = Option::and_then(
+                    field.directives.get_all(&override_directive_name).next(),
```
dylan-apollo, line 519: Curious why this vs

`field.directives.get_all(&override_directive_name).next().and_then(...)1

apollo-federation/src/merge.rs
```diff
@@ -1088,6 +1105,13 @@ fn join_field_directive_definition() -> DirectiveDefinition {
                 ty: ty!(String).into(),
                 default_value: None,
             }),
+            Node::new(InputValueDefinition {
+                name: name!("overrideLabel"),
```
dylan-apollo, line 1109: Same Q re: const

dylan-apollo (approved):
I don't know anything about join spec, so only reviewed for style 😁

apollo-federation/src/merge.rs
lennyburdette: just how it was already written! i can definitely change it

apollo-federation/src/merge.rs
lennyburdette: it's a fair question but this whole file is filled with strings and i don't really want to make a bigger diff than necessary! i'll replace this one at least

apollo-federation/src/merge.rs
lennyburdette: lol Option::and_then avoided a clippy warning

PR #5527, opened 2024-06-25
fix(federation): pass fragments when normalizing selections

dariuszkuc:
When normalizing the selection sets they might contain references to named fragments. We need to pass valid named fragments in order to normalize.

Note that in the JS codebase, fragments kept reference to named fragments hence when rebasing it would work without explicitly passing the fragments. In RS we don't store those references so when we rebase/normalize we have to explicitly pass them in.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-federation/src/operation/mod.rs
+++ b/apollo-federation/src/operation/mod.rs
@@ -2788,16 +2788,34 @@ impl SelectionSet {
         self.merge_into(std::iter::once(selection_set))
     }
 
-    /// Rebase given `SelectionSet` on self and then inserts it into the inner map. Should any sub
-    /// selection with the same key already exist in the map, the existing selection and the given
-    /// selection are merged, replacing the existing selection while keeping the same insertion index.
+    /// Rebase given `SelectionSet` on self and then inserts it into the inner map. Assumes that given
+    /// selection set does not reference ANY named fragments. If it does, Use `add_selection_set_with_fragments`
+    /// instead.
+    ///
+    /// Should any sub selection with the same key already exist in the map, the existing selection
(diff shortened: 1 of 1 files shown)


apollo-federation/src/operation/mod.rs
```diff
@@ -2788,16 +2788,34 @@ impl SelectionSet {
...
+        self.add_selection_set_with_fragments(selection_set, &NamedFragments::default())
+    }
+
+    /// Rebase given `SelectionSet` on self with the specified fragments and then inserts it into the
+    /// inner map.
+    ///
+    /// Should any sub selection with the same key already exist in the map, the existing selection
+    /// and the given selection are merged, replacing the existing selection while keeping the same
+    /// insertion index.
+    pub(crate) fn add_selection_set_with_fragments(
```
duckki, line 2811: I think this can be private, since it's only used by `fn normalize`.

duckki (approved):
I have a minor suggestion. Otherwise, it looks good to me! Thanks for catching this bug.

apollo-federation/src/operation/mod.rs
dariuszkuc: I'd keep it as public so it pops out when people try to use other `add_selection_set` that does rebase without fragments. Otherwise it is pretty easy to miss that rebasing wont work.

lrlna:
@dariuszkuc can this get follow up unit tests and/or snapshot tests. There was nothing that indicated this bug aside from an operation in the corpus.

PR #5528, opened 2024-06-25
fix(federation): fixed unnecessary sibling typename aliasing

duckki:
In JS version, Field's `siblingTypename` field can be either undefined, empty string or an alias name.
In Rust, the `Name` type can't hold empty string and the empty string case wasn't implemented before. This PR fixed that by adding a new enum type to distinguish the unaliased/aliased cases.

One previously failing snapshot test is now fixed.
Many mismatches from corpus data are fixed.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [x] Integration Tests
    - [ ] Manual Tests

### DIFF
--- a/apollo-federation/src/operation/mod.rs
+++ b/apollo-federation/src/operation/mod.rs
@@ -209,7 +209,6 @@ mod selection_map {
     use std::ops::Deref;
     use std::sync::Arc;
 
-    use apollo_compiler::ast::Name;
     use apollo_compiler::executable;
     use indexmap::IndexMap;
 
@@ -222,6 +221,7 @@ mod selection_map {
     use crate::operation::Selection;
     use crate::operation::SelectionKey;
     use crate::operation::SelectionSet;
+    use crate::operation::SiblingTypename;
 
     /// A "normalized" selection map is an optimized representation of a selection set which does
     /// not contain selections with the same selection "key". Selections that do have the same key
@@ -457,7 +457,7 @@ mod selection_map {
             self.0
         }
 
-        pub(crate) fn get_sibling_typename_mut(&mut self) -> &mut Option<Name> {
+        pub(crate) fn get_sibling_typename_mut(&mut self) -> &mut Option<SiblingTypename> {
             Arc::make_mut(self.0).field.sibling_typename_mut()
         }
 
@@ -1272,11 +1272,11 @@ mod field_selection {
             &mut self.data.directives
         }
 
-        pub(crate) fn sibling_typename(&self) -> Option<&Name> {
+        pub(crate) fn sibling_typename(&self) -> Option<&SiblingTypename> {
             self.data.sibling_typename.as_ref()
         }
 
-        pub(crate) fn sibling_typename_mut(&mut self) -> &mut Option<Name> {
+        pub(crate) fn sibling_typename_mut(&mut self) -> &mut Option<SiblingTypename> {
             &mut self.data.sibling_typename
         }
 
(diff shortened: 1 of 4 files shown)


apollo-federation/src/query_plan/fetch_dependency_graph.rs
```diff
@@ -3325,18 +3325,13 @@ fn compute_nodes_for_op_path_element<'a>(
...
-            None
-        } else {
-            Some(name.clone())
-        };
+        let alias = sibling_typename.alias();
         let typename_field = Arc::new(OpPathElement::Field(Field::new_introspection_typename(
             &dependency_graph.supergraph_schema,
             &operation.parent_type_position(),
-            alias,
+            alias.cloned(),
```
goto-bus-stop, line 3334:
```suggestion
            sibling_typename.alias().cloned(),
```
We don't really need the `alias` binding here anymore

apollo-federation/src/operation/mod.rs
```diff
@@ -2683,16 +2707,11 @@ impl SelectionSet {
...
-                None
-            } else {
-                Some(sibling_typename.clone())
-            };
+            let alias = sibling_typename.alias();
             let field_element = Field::new_introspection_typename(
                 &self.schema,
                 &selection.element()?.parent_type_position(),
-                alias,
+                alias.cloned(),
```
goto-bus-stop, line 2714:
```suggestion
            sibling_typename.alias().cloned(),
```
We don't really need the `alias` binding here anymore

goto-bus-stop (approved):
In theory we could've used `name == name!(__typename)` conditions instead of `name.is_empty()`, but I like that the enum forces you to deal with it!

duckki:
> In theory we could've used `name == name!(__typename)` conditions instead of `name.is_empty()`, but I like that the enum forces you to deal with it!

I thought about something like that. But, enum would be clearer and avoid future bugs like this.

dariuszkuc:
> In Rust, the Name type can't hold empty string and the empty string case wasn't implemented before. This PR fixed that by adding a new enum type to distinguish the unaliased/aliased cases.

Whats the use case for empty string? I was under impression that we either have an alias (which has to be a valid GraphQL name) or we don't (so its None).

duckki:
> > In Rust, the Name type can't hold empty string and the empty string case wasn't implemented before. This PR fixed that by adding a new enum type to distinguish the unaliased/aliased cases.
> 
> Whats the use case for empty string? I was under impression that we either have an alias (which has to be a valid GraphQL name) or we don't (so its None).

The `siblingTypename` field has three cases (in JS):

* `undefined` means the field is not tagged (so, no `__typename` sibling to be added).
* The empty string means the field should add a `__typename` sibling without an alias.
* A non-empty string means the field should add a `__typename` sibling with the string as its alias.

apollo-federation/src/operation/mod.rs
```diff
@@ -1330,14 +1330,32 @@ mod field_selection {
         }
     }
 
+    // SiblingTypename indicates how the sibling __typename field should be restored.
+    // PORT_NOTE: The JS version used the empty string to indicate unaliased sibling typenames.
+    // Here we use an enum to make the distinction explicit.
+    #[derive(Debug, Clone)]
+    pub(crate) enum SiblingTypename {
```
dariuszkuc, line 1337: if we are using custom enum to represent the sibling typename -> maybe we should drop the `Option`? so we would have `None/Unaliased/Aliased`?

dariuszkuc: Technically we could still use just `Option<Name>` without enum -> ie just initialize it with `__typename`

duckki: Using a special name as non-alias would throw someone off at some point. So, I didn't want to do that.

I did consider `None/Unaliased/Aliased`, but I liked the added `Option` after all. First, the sibling typename handling logic for Unaliased/Aliased will take the same branch anyways to add `__typename`. The added `fn alias()` getter function, which returns `Option<Name>`, makes no further case-split necessary. After all, that's just a subtle difference. So, I took the path of least resistance, not changing all the existing `None` cases :)

PR #5534, opened 2024-06-25
`possible_runtime_types` causing perf issues on large complex schemas

xuorig:
`possible_runtime_types` allocates  an `IndexSet` and inserts all possible types on every call which can be quite expensive for complex query plans combined with large schemas.

This PR shows an example fix where I was able to get 4x improvements on complex plans by removing all allocations. I'm happy to discuss a proper fix here, or treat this PR as an issue if you have a fix in mind.

### DIFF
--- a/apollo-federation/src/operation/mod.rs
+++ b/apollo-federation/src/operation/mod.rs
@@ -47,6 +47,7 @@ use crate::schema::position::CompositeTypeDefinitionPosition;
 use crate::schema::position::FieldDefinitionPosition;
 use crate::schema::position::InterfaceTypeDefinitionPosition;
 use crate::schema::position::SchemaRootDefinitionKind;
+use crate::schema::PossibleRuntimeTypes;
 use crate::schema::ValidFederationSchema;
 
 mod contains;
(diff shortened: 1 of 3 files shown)


xuorig:
## Before
flame-slow

## After
flame

apollo-federation/src/operation/mod.rs
```diff
@@ -4878,10 +4879,15 @@ fn runtime_types_intersect(
...
 
     if let (Ok(runtimes_1), Ok(runtimes_2)) = (
-        schema.possible_runtime_types(type1.clone()),
-        schema.possible_runtime_types(type2.clone()),
+        schema.possible_runtime_types_ref(type1.clone()),
+        schema.possible_runtime_types_ref(type2.clone()),
     ) {
-        return runtimes_1.intersection(&runtimes_2).next().is_some();
+        return match (runtimes_1, runtimes_2) {
+            (PossibleRuntimeTypes::Single(_), PossibleRuntimeTypes::Single(_)) => false,
```
dariuszkuc, line 4886: shouldn't this be `a == b`?

xuorig: in practice this is fine since

```
    if type1 == type2 {
        return true;
    }
```

is checked earlier in this function. But yeah ignore the uglyness of the implementation for now 😆

goto-bus-stop:
In theory we could do this without any additional cache or allocations by having an enum like this:

```rust
enum PossibleRuntimesRef<'a> {
    Single(ObjectTypeDefinitionPosition),
    Members(&'a IndexSet<ComponentName>),
    Implementers(&'a IndexSet<ObjectTypeDefinitionPosition),
}
```

I tried to use a pair of owned `PossibleRuntimes` / borrowed `PossibleRuntimesRef<'a>` throughout the code base which was a lot of work due to all the ways the structures are used. Maybe the idea in this PR to use the cache only in the really hot case is a good one. on Monday I can see if the "complete" approach is finishable, or else I can finish this up so we can at least fix the immediate problem.

PR #5536, opened 2024-06-25
docs: use shared `HTTPConnection` component

Meschreiber:
Uses shared `<HttpConnection />` component introduced in https://github.com/apollographql/docs/pull/889

See in deploy preview: https://deploy-preview-5536--apollo-router-docs.netlify.app/configuration/traffic-shaping#http2

### DIFF
--- a/docs/source/configuration/traffic-shaping.mdx
+++ b/docs/source/configuration/traffic-shaping.mdx
@@ -185,26 +185,7 @@ traffic_shaping:
 
 ### HTTP/2
 
-The router supports subgraph connections over:
-- HTTP/1.1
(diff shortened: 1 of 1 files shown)


BrynCooke:
@Meschreiber should the shared component not go here: https://github.com/apollographql/router/tree/dev/docs/shared
That way it's still all in the same repo.

Meschreiber:
@BrynCooke , that is where I started with it, but then the shared content needed to inject a variable (to show either `subgraph` or `coprocessor`). I couldn't find a way to do that nicely in the router repo, though that was already possible in the docs repo.

BrynCooke (approved):

PR #5537, opened 2024-06-25
connector: add visitor for JSONSelection

nicholascioli:
This commit adds a visitor to JSONSelection for walking its AST when building out an expanded subgraph.

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-federation/src/sources/connect/expand/mod.rs
```diff
@@ -145,6 +151,23 @@ fn split_subgraph(
...
+fn filter_directives<'a, D, I, U>(deny_list: &IndexSet<&Name>, directives: D) -> U
+where
+    D: IntoIterator<Item = &'a I>,
+    I: 'a + AsRef<Directive> + Clone,
+    U: FromIterator<I>,
+{
+    directives
+        .into_iter()
+        .filter(|d| {
+            println!("Checking if `{}` is in `{:?}`", d.as_ref().name, deny_list);
```
lennyburdette, line 164:
```suggestion
```

apollo-federation/src/sources/connect/expand/mod.rs
```diff
@@ -145,6 +151,23 @@ fn split_subgraph(
         .try_collect()
 }
 
+/// Filter out directives from a directive list
+fn filter_directives<'a, D, I, U>(deny_list: &IndexSet<&Name>, directives: D) -> U
```
lennyburdette, line 155: this is a good idea, but won't be useful in practice because extract_subgraphs_from_supergraphs doesn't care about directives not involved in query planning so there won't be anything to copy

lennyburdette: oh i see how this is copying over keys/requires/etc — that does make many things work in our current examples. but we'll need to control how we create keys more carefully so it'll be a bit more work, unfortunately!

apollo-bot2:

PR #5538, opened 2024-06-26
Call connector service from fetch service

pubmodmatt:
* Populate the `connectors_by_service_name` map on the `ConnectorServiceFactory`
* Call the connector service from the fetch service based on whether the service name is present in the map

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

**Notes**


apollo-router/src/services/connector_service.rs
```diff
@@ -46,10 +46,9 @@ impl tower::Service<ConnectRequest> for ConnectorService {
             .get(&request.service_name)
             .cloned();
 
-        let http_client_factory = connector
-            .as_ref()
-            .map(|c| c.id.subgraph_name.clone())
-            .and_then(|subgraph_name| self.http_service_factory.get(&subgraph_name.to_string()))
+        let http_client_factory = self
+            .http_service_factory
+            .get(&request.service_name.to_string())
```
lennyburdette, line 51: this makes sense because we haven't changed the code that creates the service factory to know about the service_name/connector/subgraph distinction. but currently the HTTP client configuration is by subgraph name, so we'll need to figure out how to map the config to the appropriate HTTP Service

apollo-router/src/services/fetch_service.rs
```diff
@@ -56,30 +119,30 @@ impl tower::Service<FetchRequest> for FetchService {
...
             output_rewrites,
             id,
             ..
         } = fetch_node;
 
-        let service_name_string = service_name.to_string();
-
-        // TODO: perf
+        // TODO: remove this when older private preview and source-aware code is removed
         let (service_name, subgraph_service_name) = match &*fetch_node.protocol {
```
lennyburdette, line 133: this should be easy to remove — there's no difference between `service_name` and `subgraph_service_name` for GraphQL subgraph requests

pubmodmatt: This was just to keep existing tests running, I think they are using the older directives. Once those are removed, this can be too.

lennyburdette: ah yeah i see the old nullability_formatting test — let's delete that since it's in the old services file. we can test that same behavior in another way

licenses.html
lennyburdette, line 1: we'll reset a lot of files if you want to just merge this

lennyburdette (approved):

PR #5539, opened 2024-06-26
FED-318: Semantic comparison of query plans

duckki:
Summary
* Expose `plan_matches` function via `_private` module, so performance-harness can use it without duplicating comparison code.
* `operation_matches` compares operations via their ASTs.
* Some parts of AST are ordering-tolerant like query variable declaration, parallel fetches.

Known issues:
* There are still false positive patterns that are not handled, yet. (Need more investigation)
* Comparison of parallel fetches could be more optimized.

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [x] Manual Tests

[FED-318]: https://apollographql.atlassian.net/browse/FED-318?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apollo-router/src/query_planner/dual_query_planner.rs
+++ b/apollo-router/src/query_planner/dual_query_planner.rs
@@ -1,6 +1,9 @@
 //! Running two query planner implementations and comparing their results
 
 use std::borrow::Borrow;
+use std::hash::Hash;
+use std::hash::Hasher;
+use std::hash::DefaultHasher;
 use std::sync::Arc;
 use std::sync::OnceLock;
 use std::time::Instant;
@@ -12,6 +15,7 @@ use apollo_compiler::ExecutableDocument;
 use apollo_compiler::NodeStr;
 use apollo_federation::query_plan::QueryPlan;
 use apollo_federation::query_plan::query_planner::QueryPlanner;
+use apollo_federation::subgraph::spec::ENTITIES_QUERY;
 
 use super::fetch::FetchNode;
 use super::fetch::SubgraphOperation;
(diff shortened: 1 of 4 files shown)


bnjjj (approved):
LGTM but I'm not the best one to review this

apollo-router/src/query_planner/dual_query_planner.rs
```diff
@@ -208,16 +213,33 @@ fn subscription_primary_matches(this: &SubscriptionNode, other: &SubscriptionNod
 }
 
 fn operation_matches(this: &SubgraphOperation, other: &SubgraphOperation) -> bool {
-    operation_without_whitespace(this) == operation_without_whitespace(other)
-}
-
-fn operation_without_whitespace(op: &SubgraphOperation) -> String {
-    op.as_serialized().replace([' ', '\n'], "")
+    let this_ast = match ast::Document::parse(this.as_serialized(), "this_operation.graphql") {
```
lrlna, line 216: flagging that reparsing an operation can potentially be expensive, but since this is run in a background thread, we are likely okay.

lrlna (approved):

apollo-router/src/query_planner/dual_query_planner.rs
duckki: JS plan has to be parsed, since it comes as a string, but the rust plan could be directly converted from `executable::Document`. I'm considering to do that.

PR #5541, opened 2024-06-26
chore(federation): implement Deref on keyed operation types

goto-bus-stop:
These are types that encode a selection key together with immutable data for that selection. The purpose of these types is that you can't mutate the things that are encoded in the key. This PR is just an ergonomic improvement :)

To do this the `Field`, `FragmentSpread`, and `InlineFragment` types hid the data behind a `.data()` accessor. But all it does is ensure that the data is immutable: we can do the same with deref, without having to call `.data()` everywhere.

The `.data()` accessor still exists because in some places we specifically clone only the data and not the key.

After this, I think I want to do FED-215. I also found an issue: #5425 introduced a `get_directives_mut()` function, but this breaks the contract of the selection map (there should be no mutable access to directives because they are part of the selection key). Unfortunately that contract was not documented clearly. I couldn't find a simple way to address it so I'm also saving that for FED-215.


dariuszkuc (approved):
Nice cleanup!

PR #5544, opened 2024-06-26
Entity cache: set up basic infrastructure for invalidation

Geal:
This sets up the `Notify` structure to act as pub/sub for invalidation events. A task is started by the entity cache plugin to execute those events. There is apparently no way to mock Redis `SCAN` commands with the `fred` crate, so instead this modifies the integration tests to be able to send back more complex subgraph responses, with headers and extensions. It also adds matching on method, headers, path at the subgraph mock level.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-380]: https://apollographql.atlassian.net/browse/ROUTER-380?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apollo-router/src/plugins/cache/entity.rs
```diff
@@ -389,6 +399,15 @@ impl InnerCacheService {
                             }
                         }
 
+                        if let Some(invalidation_extensions) = response
+                            .response
+                            .body_mut()
+                            .extensions
+                            .remove("invalidation")
```
bnjjj, line 406: Could be part of a constant

apollo-router/src/plugins/cache/invalidation.rs
```diff
@@ -0,0 +1,151 @@
...
+    u64_histogram!(
+        "apollo.router.cache.invalidation.keys",
+        "Number of invalidated keys.",
+        count
+    );
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(tag = "kind")]
+pub(crate) enum InvalidationRequest {
```
bnjjj, line 125: Do we really want to have the kind field capitalized ?

bnjjj (approved):

apollo-router/src/plugins/cache/invalidation.rs
Geal: do I do that with `#[serde(rename_all = "lowercase")]` over the whole enum?

Geal:
noting that this cannot work right now because redis is only running on linux in CI (I think)

apollo-router/src/plugins/cache/invalidation.rs
Geal: done in a189c67bd

PR #5557, opened 2024-06-27
feat(coprocessor): add conditions on coprocessor stages

bnjjj:
Adding support of conditions on coprocessor stages (except for `Execution` service) based on existing [conditions](https://www.apollographql.com/docs/router/configuration/telemetry/instrumentation/conditions/) we already have in the router for custom telemetry. Including usage of existing [selectors](https://www.apollographql.com/docs/router/configuration/telemetry/instrumentation/selectors/)

Example if you would like to execute a coprocessor on supergraph response only on primary response and not subscription events for example:

```yaml title=router.yaml
coprocessor:
  url: http://127.0.0.1:3000 # mandatory URL which is the address of the coprocessor
  timeout: 2s # optional timeout (2 seconds in this example). If not set, defaults to 1 second
  supergraph:
    response: 
      condition:
        eq:
        - true
        - is_primary_response: true
      body: true
```

or another example to not execute the coprocessor at all if it's a subscription:

```yaml title=router.yaml
coprocessor:
  url: http://127.0.0.1:3000 # mandatory URL which is the address of the coprocessor
  timeout: 2s # optional timeout (2 seconds in this example). If not set, defaults to 1 second
  supergraph:
    response: 
      condition:
        not:
          eq:
          - subscription
          - operation_kind: string
      body: true
```

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [x] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-389]: https://apollographql.atlassian.net/browse/ROUTER-389?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.changesets/feat_bnjjj_feat_condition_copro.md
```diff
@@ -0,0 +1,36 @@
+### feat(coprocessor): add conditions on coprocessor stages ([PR #5557](https://github.com/apollographql/router/pull/5557))
+
+Adding support of conditions on coprocessor stages (except for `Execution` service) based on existing [conditions](https://www.apollographql.com/docs/router/configuration/telemetry/instrumentation/conditions/) we already have in the router for custom telemetry. Including usage of existing [selectors](https://www.apollographql.com/docs/router/configuration/telemetry/instrumentation/selectors/)
```
shorgi, line 3:
```suggestion
### feat(coprocessor): Support conditional coprocessor execution per stage of request lifecycle ([PR #5557](https://github.com/apollographql/router/pull/5557))

The router now supports conditional execution of the coprocessor for each stage of the request lifecycle (except for the `Execution` stage).

To configure, define conditions for a specific stage by using selectors based on headers or context entries. For example, based on a supergraph response you can configure the coprocessor not to execute for any subscription:
  
```

BrynCooke (approved):
Approved!
But please can you follow up by moving the selectors to `src/configuration/selectors`.
Also it'd be good to have a separate file for each type of selector, so we would end up with:
`src/configuration/selectors/router.rs`
`src/configuration/selectors/supergraph.rs`

etc.

apollo-router/src/configuration/snapshots/apollo_router__configuration__tests__schema_generation.snap
```diff
@@ -4892,8 +4902,8 @@ expression: "&schema"
           "description": "The trace ID of the request.",
           "properties": {
             "trace_id": {
-              "$ref": "#/definitions/TraceIdFormat2",
```
BrynCooke, line 4895: Looks like there is non-determinism here. I did a check and there are two TraceIdFormat enums. We need to unify them.

PR #5560, opened 2024-06-27
Add ability to override base URL on a connector source

pubmodmatt:
Add the ability override the base URL of a `@source` directive in the supergraph configuration YAML. Any connector referencing that source will use the overridden base URL.

This required adding the source name to the `ConnectId` of the connector, to allow looking up the associated source configuration. This field is optional, since a connector can directly specify the full URL of the REST API, rather than referring to a source directive. In that case, the base URL cannot be overridden.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

**Notes**

[CNN-229]: https://apollographql.atlassian.net/browse/CNN-229?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lennyburdette (approved):
spoke offline about the location of `override_connector_base_urls` but otherwise LGTM!

PR #5561, opened 2024-06-27
Adding http header validation for source and connect directives

tayrrible:
*Description here*

Fixes #**issue_number**

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-federation/src/sources/connect/models/test_data/validation/invalid_connect_http_headers.graphql
```diff
@@ -0,0 +1,19 @@
...
+
+type Query {
+  resources: [String!]!
+    @connect(
+      http: {
+        GET: "http://127.0.0.1:8000/resources"
+        headers: [
+          { name: "Accept", value: "text/html" }
+          { name: "Accept", as: "Accept" }
+          { as: "Accept", value: "text/html" }
```
dylan-apollo, line 12: I think `name` is required, so this will eventually fail GraphQL validation when we start checking it.

apollo-federation/src/sources/connect/models/validation.rs
```diff
@@ -199,6 +204,33 @@ fn validate_source(directive: &Component<Directive>, sources: &SourceMap) -> Sou
     if let Some(url_error) = url_error {
         errors.push(url_error);
     }
+
+    if let Some(http_arg) = directive
```
dylan-apollo, line 208: We already get the http arg up above, since we need it to validate the URL. So maybe we should get it once, earlier, and return early (probably with a `GraphQLError`, since the arg is required) if it's not present.

apollo-federation/src/sources/connect/models/validation.rs
```diff
@@ -514,6 +569,169 @@ fn parse_url(value: &Node<Value>, coordinate: &str, sources: &SourceMap) -> Resu
...
+    source_map: &SourceMap,
+    object: Option<&Name>,
+    field: Option<&Name>,
+) -> Result<&'a Node<Value>, Vec<Message>> {
+    let mut errors = Vec::new();
+
+    headers
+        .as_list()
+        .map(|l| l.iter().filter_map(|o| o.as_object()).collect_vec())
+        .unwrap_or_else(|| headers.as_object().map(|o| vec![o]).unwrap_or_default())
```
dylan-apollo, line 585: Can `headers` be an object? I thought it had to be a list of objects

apollo-federation/src/sources/connect/models/validation.rs
```diff
@@ -514,6 +569,169 @@ fn parse_url(value: &Node<Value>, coordinate: &str, sources: &SourceMap) -> Resu
...
+    headers: &'a Node<Value>,
+    source_map: &SourceMap,
+    object: Option<&Name>,
+    field: Option<&Name>,
+) -> Result<&'a Node<Value>, Vec<Message>> {
+    let mut errors = Vec::new();
+
+    headers
+        .as_list()
+        .map(|l| l.iter().filter_map(|o| o.as_object()).collect_vec())
```
dylan-apollo, line 584: I don't _think_ you need to `collect_vec()` here

apollo-federation/src/sources/connect/models/validation.rs
```diff
@@ -514,6 +569,169 @@ fn parse_url(value: &Node<Value>, coordinate: &str, sources: &SourceMap) -> Resu
...
 }
 
+fn validate_header_arg<'a>(
+    directive_name: &Name,
+    argument_name: &str,
+    headers: &'a Node<Value>,
+    source_map: &SourceMap,
+    object: Option<&Name>,
+    field: Option<&Name>,
+) -> Result<&'a Node<Value>, Vec<Message>> {
```
dylan-apollo, line 579: Looks like you're only using the `Err` variant of this in callers, so this _could_ return just `Vec<Message>`

apollo-federation/src/sources/connect/models/validation.rs
```diff
@@ -514,6 +569,169 @@ fn parse_url(value: &Node<Value>, coordinate: &str, sources: &SourceMap) -> Resu
...
+    headers
+        .as_list()
+        .map(|l| l.iter().filter_map(|o| o.as_object()).collect_vec())
+        .unwrap_or_else(|| headers.as_object().map(|o| vec![o]).unwrap_or_default())
+        .iter()
+        .for_each(|arg_pairs| {
+            let mut has_as_or_value = false;
+            let mut unique_header_set = HashSet::new();
+
+            arg_pairs.iter().for_each(|(key, value)| {
```
dylan-apollo, line 591: I think rather than trying to keep state in `has_as_or_value`, I would `.find()` each of the relevant arguments and then do the checks. Might be easier to read.

Something like this:

```rust
let name_arg = arg_pairs.iter().find_map(|(key, value)| (key == &name!("name")).then_some(value));
let as_arg = arg_pairs.iter().find_map(|(key, value)| (key == &name!("as")).then_some(value));
let value_arg = arg_pairs.iter().find_map(|(key, value)| (key == &name!("value")).then_some(value));
```

apollo-federation/src/sources/connect/models/validation.rs
```diff
@@ -514,6 +569,169 @@ fn parse_url(value: &Node<Value>, coordinate: &str, sources: &SourceMap) -> Resu
...
+    field: Option<&Name>,
+) -> Result<&'a Node<Value>, Vec<Message>> {
+    let mut errors = Vec::new();
+
+    headers
+        .as_list()
+        .map(|l| l.iter().filter_map(|o| o.as_object()).collect_vec())
+        .unwrap_or_else(|| headers.as_object().map(|o| vec![o]).unwrap_or_default())
+        .iter()
+        .for_each(|arg_pairs| {
```
dylan-apollo, line 587: If you return a `Vec<Message>`, then you'll probably want to do `map()` here instead and `.collect()` at the end.

apollo-federation/src/sources/connect/models/validation.rs
lennyburdette: i just tested this — we're working with the AST which doesn't have semantic understanding of definitions. it's valid to provide a single value instead of a list for a list argument, so we have to handle both cases. Here's an example AST 
```
[apollo-federation/tests/main.rs:22:5] &field.directives = [
    89..106 @1 Directive {
        name: "myDirective",
        arguments: [
            92..105 @1 Argument {
                name: "listOfStrings",
                value: 95..105 @1 List(
                    [
                        96..99 @1 String(
                            "a",
                        ),
                        101..104 @1 String(
                            "b",
                        ),
                    ],
                ),
            },
        ],
    },
    107..117 @1 Directive {
        name: "myDirective",
        arguments: [
            110..116 @1 Argument {
                name: "listOfStrings",
                value: 113..116 @1 String(
                    "c",
                ),
            },
        ],
    },
]
```

apollo-federation/src/sources/connect/models/validation.rs
lennyburdette: i like this suggestion, and i think it makes the as/value rule easier to write too 
```rs
if let (Some(as_arg), Some(value_arg)) = (as_arg, value_arg) {
  errors.push(...);
}
```

apollo-federation/src/sources/connect/models/validation.rs
```diff
@@ -514,6 +569,169 @@ fn parse_url(value: &Node<Value>, coordinate: &str, sources: &SourceMap) -> Resu
...
+) -> Result<&'a str, Message> {
+    let s = value.as_str().ok_or_else(|| Message {
+        code: Code::GraphQLError,
+        message: format!("{coordinate} contains an invalid header name type."),
+        locations: Location::from_node(value.location(), source_map)
+            .into_iter()
+            .collect(),
+    })?;
+
+    match HeaderName::try_from(s) {
```
lennyburdette, line 682:
```rs
HeaderName::try_from(s).map_err(|_| Message {...})
```

apollo-federation/src/sources/connect/models/validation.rs
```diff
@@ -571,6 +789,22 @@ fn connect_directive_url_coordinate(
...
 }
 
+fn directive_http_header_coordinate(
+    directive_name: &Name,
+    argument_name: &str,
+    object: Option<&Name>,
+    field: Option<&Name>,
+) -> String {
+    match (object, field) {
+        (Some(obj), Some(fld)) => {
```
lennyburdette, line 799: nit:
```suggestion
        (Some(object), Some(field)) => {
```
shadowing variables is common in idiomatic rust

apollo-federation/src/sources/connect/models/validation.rs
dylan-apollo: I didn't know it was legal in GraphQL to provide a single value to an input that takes a list. What a wild place.

apollo-federation/src/sources/connect/models/validation.rs
tayrrible: I tried removing it but it caused errors

apollo-federation/src/sources/connect/models/validation.rs
```diff
@@ -569,100 +575,92 @@ fn parse_url(value: &Node<Value>, coordinate: &str, sources: &SourceMap) -> Resu
...
+            // validate `as`
+            if let Some(as_value) = as_arg {
+                if let Some(err) = validate_header_name(&name!("as"), as_value, pair_coordinate, source_map).err() {
+                    errors.push(err);
+                }
+            }
+
+            if let (Some(as_arg), Some(name_arg)) = (as_arg, name_arg) {
+                if let (Some(as_value), Some(name_value)) = (as_arg.as_str(), name_arg.as_str()) {
+                    if as_value == name_value {
```
lennyburdette, line 631: i didn't even think of this one! it makes sense as it's probably an accident, but you could argue that `{ name: "x-foo" }` is really just a convenience for `{ name: "x-foo", as: "x-foo" }`.

when i first read this, i thought it was validating that `name`s are unique across header objects.

maybe this is actually a bit more complicated than i thought.

```graphql
@source(
  name: "json",
  http: { 
    baseURL: "http://whatever"
    headers: [
      { name: "x-foo", as: "x-bar" }
      { name: "x-foo", values: ["baz"] }
    ]
)
```
At runtime, this should result in a valid map:
```json
{ "x-bar": req.headers["x-foo"], "x-foo": ["baz"] }
```

now i'm wondering if it should actually be
```
    headers: [
      { name: "x-bar", from: "x-foo" }
      { name: "x-foo", values: ["baz"] }
    ]
```

where `{ name: "x-bar" }` is a convenience for `{ name: "x-bar", from: "x-bar" }`.

this would make the validation rule simpler: all `name` attributes must be unique.

lennyburdette (approved):

apollo-bot2:

PR #5566, opened 2024-06-28
tell jemalloc to use 64-KiB pages

### LINKED ISSUE #3382: Unsupported system page size
**Describe the bug**
router returns error when arm64 version is deployed to Kubernetes
`<jemalloc>: Unsupported system page size
<jemalloc>: Unsupported system page size
memory allocation of 5 bytes failed`

**To Reproduce**
Deploy router v1.22.0 from the Helm Chart to Kubernetes running on arm64 node

yanns:
*Description here*

Fixes #3382

For more detail about the issue:
- https://github.com/jemalloc/jemalloc/issues/467

This change is similar to what [other projects are doing](https://github.com/PingCAP-QE/ci/pull/2409). We've been using that internally as well.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


garypen:
It's probably the right thing to do at this point. I really hoped that they'd fix this inside jemalloc, but ...

garypen (commented):
I wonder if we are happy only having this in CI or if it's something we need to set for local builds...

bnjjj:
Maybe we could use [this](https://doc.rust-lang.org/nightly/cargo/reference/config.html#env) to enforce it for the project, even locally ? cc @garypen. @SimonSapin did you already use this ?

garypen:
@bnjjj That's a good suggestion I think. I suppose it would only apply to people developing/testing on Linux (we don't use jemalloc on macOS/Windows). Could you try it out?

abernix:
@mergify copy dev

aaronArinder:
stale; reopen if needed

PR #5567, opened 2024-06-28
Fetch parity between connector service and subgraph service

pubmodmatt:
* Change return type of the connector service to `graphql::Response` to match subgraph service. This allows the fetch service to treat responses the same as subgraph service responses
* Add `response_at_path` call after connector service responses
* Add deferred fetch handling after connector service responses
* The unit test is updated to include an entity, which now passes

The goal here is to achieve parity between connector service and subgraph service response processing, while minimizing changes from the `dev` branch.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

**Notes**


lennyburdette (approved):
looks great!

PR #5573, opened 2024-07-01
Cheap apollo_compiler::Name to opentelemetry::StringValue conversion

SimonSapin:
Take advantage of the recent change of internal representation of `Name`: https://github.com/apollographql/apollo-rs/pull/868


goto-bus-stop (approved):

apollo-router/src/plugins/telemetry/config_new/graphql/selectors.rs
```diff
@@ -175,6 +175,16 @@ impl Selector for GraphQLSelector {
     }
 }
 
+fn name_to_otel_string(name: &apollo_compiler::Name) -> opentelemetry::StringValue {
+    if let Some(static_str) = name.as_static_str() {
+        static_str.into()
+    } else {
+        // One of `as_static_str` or `to_cloned_arc` always returns `Some`,
+        // so this `unwrap` never panics.
+        name.to_cloned_arc().unwrap().into()
```
bnjjj, line 184:
```suggestion
        name.to_cloned_arc().expect("One of `as_static_str` or `to_cloned_arc` always returns `Some` so this `unwrap` never panics.").into()
```

bnjjj (approved):

apollo-router/src/plugins/telemetry/config_new/graphql/selectors.rs
SimonSapin: I’ve pushed a change from `unwrap` to `expect` as suggested but the comment meant to explain to source code readers why we don’t propagate an error so I’ve rephrased it from the point of view of what the panic message should look like if it happens.

bnjjj (approved):

PR #5578, opened 2024-07-02
Request timeout and rate limited error responses are structured errors

IvanGoncharov:
Request timeout (`408 Request Timeout`) and request rate limited (`429 Too Many Requests`)errors will now result in a structured GraphQL error (i.e., `{"errors": [...]}`) being returned to the client rather than a plain-text error as was the case previously.

Also both errors are now properly tracked in telemetry, including `apollo_router_graphql_error_total` metric. 

 
 
---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [x] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-413]: https://apollographql.atlassian.net/browse/ROUTER-413?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ROUTER-357]: https://apollographql.atlassian.net/browse/ROUTER-357?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apollo-router/src/plugins/traffic_shaping/mod.rs
```diff
@@ -375,16 +379,23 @@ impl TrafficShaping {
...
                 ))
-                    .layer(TimeoutLayer::new(
-                        config.shaping
-                        .timeout
-                        .unwrap_or(DEFAULT_TIMEOUT),
-                    ))
+                    .map_future_with_request_data(
+                        |req: &subgraph::Request| req.context.clone(),
+                        move |ctx, response| {
+                            request_timeout(timeout, response).unwrap_or_else(|error| {
```
Geal, line 390: ah, I think there was some understanding. When I suggested mapping on the response, that was not to replace the entire `Timeout` layer, but to look at the error it returned and converting it

apollo-router/src/plugins/traffic_shaping/mod.rs
```diff
@@ -266,15 +267,7 @@ impl Plugin for TrafficShaping {
...
-        timeout::future::ResponseFuture<
-            Oneshot<
-                Either<
-                    Retry<RetryPolicy, Either<rate::service::RateLimit<S>, S>>,
-                    Either<rate::service::RateLimit<S>, S>,
-                >,
-                subgraph::Request,
-            >,
-        >,
+        BoxFuture<'static, Result<subgraph::Response, BoxError>>,
```
bnjjj, line 270: If it's 
```
Either<
        BoxFuture<'static, Result<subgraph::Response, BoxError>>,
        BoxFuture<'static, Result<subgraph::Response, BoxError>>,
    >
```

then we probably don't need Either at all

apollo-router/src/plugins/traffic_shaping/mod.rs
bnjjj, line 390: I think naming is confusing here. It's not really a `response`  it's a future and `request_timeout` maybe we should rename it for something like `check_timeout` or `execute_with_timeout`. I don't know. But it would make the review easier

apollo-router/src/layers/map_future_with_request_data.rs
```diff
@@ -36,6 +36,7 @@ where
 }
 
 /// [`Service`] for mapping futures with request data. See [`ServiceBuilderExt::map_future_with_request_data()`](crate::layers::ServiceBuilderExt::map_future_with_request_data()).
+#[derive(Clone)]
```
bnjjj, line 39: Is it required ?

bnjjj (changes requested):
I think it might also be interesting to add more tests for metrics, like is it properly accounted with the right status code and so one

apollo-router/src/plugins/traffic_shaping/mod.rs
Geal: @IvanGoncharov could you bring back the previous timeout code? That was extracted from tower's implementation, we tend to align with their way of doing things in traffic shaping. The only thing we need to do here is converting the error response

apollo-router/src/plugins/traffic_shaping/mod.rs
IvanGoncharov: Sadly we can't remove it 😢 
We are using `Either` here: 
https://github.com/apollographql/router/blob/bf66e7ab035be9bcc26a01944710a9199bd3d19e/apollo-router/src/plugins/traffic_shaping/mod.rs#L378
And it forces `Either` on future:
https://docs.rs/tower/latest/src/tower/util/either.rs.html#38

apollo-router/src/layers/map_future_with_request_data.rs
IvanGoncharov: Yes, we use it for subgraph service, which should be clonable since it is shared between subgraphs:
https://github.com/apollographql/router/blob/bf66e7ab035be9bcc26a01944710a9199bd3d19e/apollo-router/src/plugins/traffic_shaping/mod.rs#L334

apollo-router/src/plugins/traffic_shaping/mod.rs
IvanGoncharov: Fixed ✅

IvanGoncharov: Yes, I reverted this code.
I still think `map_future_with_request_data` removes the need for extra boilerplate implemented in the tower's layer, but I understand the argument of staying close to the tower's code.

IvanGoncharov:
>I think it might also be interesting to add more tests for metrics, like is it properly accounted with the right status code and so one

@bnjjj I added metrics into integration tests, please take a look

apollo-router/src/plugins/traffic_shaping/mod.rs
```diff
@@ -311,6 +303,32 @@ impl TrafficShaping {
         <S as Service<supergraph::Request>>::Future: std::marker::Send,
     {
         ServiceBuilder::new()
+            .map_future_with_request_data(
```
bnjjj, line 306: I don't think we need to pass the context here as it's already part of the response

apollo-router/src/plugins/traffic_shaping/mod.rs
bnjjj: I'll take a deeper look but maybe we can completely get rid of either at that place as it's the same kind of future

apollo-router/src/plugins/traffic_shaping/mod.rs
IvanGoncharov: @bnjjj It's not.
If the timeout or rate-limiting code is returning an error; I don't have access to the response at all since it inside `Ok`.
Also, timeout, in particular, triggers before the response is created, so we need to have some code that passes context to the timeout error somewhere.

apollo-router/src/plugins/traffic_shaping/mod.rs
IvanGoncharov: @bnjjj It's the same kind of future, but services are different.
My initial revision removed it all and just used `BoxService.` I also wanted to use proper plugin API.
But @Geal provided some context in the Slack thread.
So, I just reverted my changes to the smallest possible change.

bnjjj (approved):

BrynCooke (approved):

garypen (approved):

bnjjj (approved):

PR #5579, opened 2024-07-02
remove panics in connect/spec

lennyburdette:
to avoid issues with invalid schemas during authoring — the expansion code for satisfiability happens in composition/on edit

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-federation/src/sources/connect/spec/directives.rs
+++ b/apollo-federation/src/sources/connect/spec/directives.rs
@@ -6,6 +6,7 @@ use apollo_compiler::Node;
 use indexmap::map::Entry::Occupied;
 use indexmap::map::Entry::Vacant;
 use indexmap::IndexMap;
+use itertools::Itertools;
 
 use super::schema::ConnectDirectiveArguments;
 use super::schema::ConnectHTTPArguments;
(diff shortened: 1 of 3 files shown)


apollo-federation/src/sources/connect/spec/directives.rs
```diff
@@ -102,38 +113,41 @@ impl TryFrom<&Component<Directive>> for SourceDirectiveArguments {
             let arg_name = arg.name.as_str();
 
             if arg_name == SOURCE_NAME_ARGUMENT_NAME.as_str() {
-                name = Some(
-                    arg.value
-                        .as_str()
-                        .expect("`name` field in `@source` directive is not a string"),
-                );
+                name = Some(arg.value.as_str().ok_or_else(|| {
```
pubmodmatt, line 116: When your closure just returns an error like this, you can use `ok_or` instead of `ok_or_else` to just directly map to that error. Applies to many of the other `ok_or_else` calls as well.

apollo-federation/src/sources/connect/spec/type_and_directive_specifications.rs
```diff
@@ -43,6 +44,12 @@ pub(super) fn check_or_add(
     link: &Link,
     schema: &mut FederationSchema,
 ) -> Result<(), FederationError> {
+    fn internal(s: &str) -> SingleFederationError {
```
pubmodmatt, line 47: nitpick: this seems more like a macro

pubmodmatt (changes requested):

apollo-federation/src/sources/connect/spec/directives.rs
lennyburdette: switched to ok_or with a macro

pubmodmatt (approved):

PR #5580, opened 2024-07-02
FED-332: added the `remove_introspection` step in `normalize_operation`

duckki:
Summary
- ported `withoutIntrospection` JS function

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[FED-332]: https://apollographql.atlassian.net/browse/FED-332?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apollo-federation/src/operation/mod.rs
+++ b/apollo-federation/src/operation/mod.rs
@@ -4920,13 +4920,13 @@ pub(crate) fn normalize_operation(
 fn remove_introspection(selection_set: &mut SelectionSet) {
     // Note that, because we only apply this to the top-level selections, we skip all
     // introspection, including __typename. In general, we don't want to ignore __typename during
-    // query plans, but at top-level, we can let the gateway execution deal with it rather than
+    // query plans, but at top-level, we can let the router execution deal with it rather than
     // querying some service for that.
 
     Arc::make_mut(&mut selection_set.selections).retain(|_, selection| {
         !matches!(selection,
             Selection::Field(field_selection) if
-                field_selection.field.data().field_position.is_introspection_typename_field()
+                field_selection.field.field_position.is_introspection_typename_field()
         )
     });
 }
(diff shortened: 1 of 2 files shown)


apollo-federation/src/operation/mod.rs
```diff
@@ -4915,6 +4916,21 @@ pub(crate) fn normalize_operation(
     Ok(normalized_operation)
 }
 
+// PORT_NOTE: This is a port of `withoutIntrospection` from JS version.
+fn remove_introspection(selection_set: &mut SelectionSet) {
+    // Note that, because we only apply this to the top-level selections, we skip all
+    // introspection, including __typename. In general, we don't want to ignore __typename during
+    // query plans, but at top-level, we can let the gateway execution deal with it rather than
```
goto-bus-stop, line 4923:
```suggestion
    // query plans, but at top-level, we can let the router execution deal with it rather than
```

apollo-federation/src/operation/mod.rs
```diff
@@ -4915,6 +4916,21 @@ pub(crate) fn normalize_operation(
...
+    // querying some service for that.
+
+    Arc::make_mut(&mut selection_set.selections).retain(|_, selection| {
+        !matches!(selection,
+            Selection::Field(field_selection) if
+                field_selection.field.data().field_position.is_introspection_typename_field()
```
goto-bus-stop, line 4929:
```suggestion
                field_selection.field.field_position.is_introspection_typename_field()
```
`.data()` accessor is no longer required

goto-bus-stop (approved):
We *may* also handle this in Rust introspection later, then this would not happen. https://github.com/apollographql/apollo-rs/issues/867

dariuszkuc:
we were already skipping those during the initial step of converting from [apollo-rs selection set to fed selection set](https://github.com/apollographql/router/blob/dev/apollo-federation/src/operation/mod.rs#L3606-L3608)

dariuszkuc:
ah I see - JS logic was stripping anything starting with two `__`, probably should merge those two together

PR #5583, opened 2024-07-02
configure introspection caching

### LINKED ISSUE #4861: Allow disabling introspection responses going to cache
Currently, introspection queries and their response data is saved to the cache, which also populated in Redis.

https://github.com/apollographql/router/blob/615b74d9cec12d2ea94109335b8321fea73d688a/apollo-router/src/introspection.rs#L79

If I have an extremely large schema and I have clients making slightly different introspection queries this can start sending lots of data to my Redis instance and eating up performance, bandwith, and storage

Allow configuring or disabling if introspection responses are cached or not

Enabled cache
```yaml
supergraph:
  introspection:
    enabled: true
    cache: 
      in_memory:
        limit: 1
      redis:
        urls: [ "redis://localhost:6379" ]
```

Disabled cache

```yaml
supergraph:
  introspection:
    enabled: true
    cache: false # Not sure what is better here. I think default should be cache on, so how do we turn it off?
    cache: 
      in_memory:
        enabled: false
      redis:
        enabled: false
```

There is no control over this today. I need to enforce that introspection is all the same or I can disable it entirely

```yaml
supergraph:
  introspection: false
```

**Additional context**
Maybe as part of the issue https://github.com/apollographql/router/issues/4608 we can add more features

OR

We could lock down introspection to only 1 predefined operation: https://github.com/apollographql/router/issues/4871

Geal:
currently, introspection has to go through the query planner, and since that is expensive, the Router caches the introspection responses. This can end up filling the distributed cache, so until we have moved introspection execution entirely out of the planner, we make introspection response caching configurable

Fixes #4861

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [x] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-352]: https://apollographql.atlassian.net/browse/ROUTER-352?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apollo-router/src/query_planner/caching_query_planner.rs
```diff
@@ -881,4 +894,108 @@ mod tests {
             stats_report_key_hash("# IgnitionMeQuery\nquery IgnitionMeQuery{me{id}}")
         );
     }
+
+    #[test(tokio::test)]
+    async fn test_introspection_cache() {
+        let mut delegate = MockMyQueryPlanner::new();
+        delegate
+            .expect_clone()
+            // This is the main point of the test: if ontrospection queries are not cached, then the delegate
```
bnjjj, line 903:
```suggestion
            // This is the main point of the test: if introspection queries are not cached, then the delegate
```

bnjjj (approved):

apollo-router/src/configuration/mod.rs
```diff
@@ -1080,6 +1080,19 @@ pub(crate) struct QueryPlanning {
...
     /// Default: 1.
     pub(crate) experimental_parallelism: AvailableParallelism,
+
+    /// Activates introspection response caching
+    /// Historically, the Router has executed introspection queries in the query planner, and cached their
+    /// response in its cache because they were expensive. This will change soon as introspection will be
+    /// removed from the query planner. In the meantime, since storing introspection responses can fill up
+    /// the cache, this option can be used to deactivate it.
+    /// Default: true
+    #[serde(default = "legacy_introspection_caching")]
```
BrynCooke, line 1090: This struct needs to use `serde_derive_default::Default` rather than just default to prevent this having unexpected values.

apollo-router/src/configuration/mod.rs
```diff
@@ -1080,6 +1080,19 @@ pub(crate) struct QueryPlanning {
...
+    pub(crate) legacy_introspection_caching: bool,
+}
+
+fn legacy_introspection_caching() -> bool {
+    true
```
BrynCooke, line 1095: Do we not want to disable this by default?

BrynCooke (changes requested):
Must have `serde_derive_default::Default`. Question around the actual default value that should be discussed with team.

apollo-router/src/configuration/mod.rs
bnjjj: I think we should keep the previous default behavior now, But that's something we can revisit for router next ?

apollo-router/src/configuration/mod.rs
bnjjj: @BrynCooke I changed the impl. Could you confirm me it's good ?

docs/source/configuration/overview.mdx
```diff
@@ -578,6 +578,20 @@ The default value of `experimental_parallelism` is `1`.
...
+<MinVersion version="1.51.0">
+
+### Introspection response caching
+
+Introspection responses are generated by the query planner for now, so they are expensive to execute and the router stores them in its query planner cache. Unfortunately, they can fill up the cache, so until we move out introspection execution, there is an option to deactivate response caching.
+
+```yaml title="router.yaml"
+supergraph:
+  query_planning:
+    legacy_introspection_caching: true
```
BrynCooke, line 590:
```suggestion
    legacy_introspection_caching: false
```

The text avove talks about deactivating introspection caching, but then the example shows it being enabled.

.changesets/feat_geal_configure_introspection_caching.md
```diff
@@ -0,0 +1,13 @@
+### configure introspection caching ([PR #5583](https://github.com/apollographql/router/pull/5583))
+
+Adds an option to deactivate introspection response caching.
+Currently, introspection has to go through the query planner, and since that is expensive, the Router caches the introspection responses. This can end up filling the distributed cache, so until we have moved introspection execution entirely out of the planner, we make introspection response caching configurable, as follows:
+
+
+```yaml
+supergraph:
+  query_planning:
+    legacy_introspection_caching: true
```
BrynCooke, line 10:
```suggestion
    legacy_introspection_caching: false
```

BrynCooke (approved):

apollo-router/src/configuration/mod.rs
BrynCooke: Fair enough

PR #5584, opened 2024-07-02
chore(federation): refactor error handling in rebasing

goto-bus-stop:
**This is a cleanup PR so reviewing it is not urgent.** However we
may be able to build performance improvements on top of this (by
mutating `Self` instead of creating copies).

---

Rebasing has an error handling option to silently drop selections if
they cannot rebase cleanly. This is only used when rebasing named
fragments onto a different schema.

The way it was implemented is that `.rebase_on()` returns
`Result<Option<T>>`. The `Ok(None)` case would only ever happen if the
`IgnoreError` option was passed. Since it almost never is passed, most
usage sites had to handle a case that would never happen.

This changes the error handling to use a special `FederationError`
branch for rebasing errors. Instead of checking for `Ok(None)`, the
selection set rebasing code checks for errors having `err.is_rebase_error()`,
and ignores them at that point.

This has the same effect, but all the other usage sites no longer have
to check a case that can never happen.

Additionally, since `IgnoreError` was only used in one place, this makes
the error handling option private. I renamed it to `OnNonRebaseableSelection`
which I don't really like but I wanted to make it clearer exactly what case it
handles.


dariuszkuc (approved):
nice cleanup!

PR #5591, opened 2024-07-02
connectors license enforcement with join__directive serialization

lennyburdette:
---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/uplink/license_enforcement.rs
+++ b/apollo-router/src/uplink/license_enforcement.rs
@@ -11,8 +11,10 @@ use std::time::Duration;
 use std::time::SystemTime;
 use std::time::UNIX_EPOCH;
 
+use apollo_compiler::ast;
 use apollo_compiler::ast::Definition;
 use apollo_compiler::schema::Directive;
+use apollo_compiler::Name;
 use apollo_compiler::Node;
 use buildstructor::Builder;
 use displaydoc::Display;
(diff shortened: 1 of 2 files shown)


apollo-router/src/uplink/license_enforcement.rs
```diff
@@ -135,6 +138,43 @@ impl ParsedLinkSpec {
...
+        let spec_name = segments.next()?.to_string();
+        let spec_url = format!(
+            "{}://{}/{}",
+            parsed_url.scheme(),
+            parsed_url.host()?,
+            spec_name
+        );
+        let version_string = segments.next()?.strip_prefix('v')?;
+        let parsed_version =
+            semver::Version::parse(format!("{}.0", &version_string).as_str()).ok()?;
```
pubmodmatt, line 161: Is appending this `.0` always correct? Is it enforced that the version string in the URL is always major.minor before we get here?

pubmodmatt (approved):

apollo-router/src/uplink/license_enforcement.rs
lennyburdette: good question, but yeah — the link spec doesn't support patch versions: https://specs.apollo.dev/link/v1.0/#@link.url

PR #5593, opened 2024-07-02
chore(federation): split out and rename SelectionSet::normalize implementation

goto-bus-stop:
The main point of this PR is to clarify what `SelectionSet::normalize` actually is. I've called it `flatten_unnecessary_fragments` here. In speech we can refer to this as flattening fragments instead of as normalization. I'm open to bikeshedding but please only if you feel strongly.

I also removed non-recursive flattening which was never used.

I recommend reviewing the individual commits.

With this and #5584, it would be great if we could take `&mut self` in these methods and mutate instead of creating loads of copies. I think it requires adding an API to `SelectionMap` to modify a selection and handle the in-place update in case the selection's key was changed by the modification. Once that's figured out we should be able to remove a lot of busywork from these functions.


apollo-federation/src/operation/simplify.rs
```diff
@@ -0,0 +1,537 @@
...
+                typename_selection.insert(non_included_typename);
+
+                normalized_selection.selections = Arc::new(typename_selection);
+                selection.selection_set = Some(normalized_selection);
+            } else {
+                selection.selection_set = Some(normalized_selection);
+            }
+            Ok(Some(SelectionOrSet::Selection(Selection::from(selection))))
+        } else {
+            // JS PORT NOTE: In JS implementation field selection stores field definition information,
```
dariuszkuc, line 117: can you remove this old comment (that is no longer true)

apollo-federation/src/operation/simplify.rs
```diff
@@ -0,0 +1,537 @@
...
+                            schema,
+                            RebaseErrorHandlingOption::ThrowError,
+                        )?
+                    };
+                    Ok(Some(SelectionOrSet::SelectionSet(selection_set)))
+                };
+            }
+        }
+
+        // We preserve the current fragment, so we only recurse within the sub-selection if we're asked to be recursive.
```
dariuszkuc, line 229: this comment is also out of date

dariuszkuc (approved):
approved (but would be good to drop those old comments)

apollo-federation/src/operation/simplify.rs
```diff
@@ -0,0 +1,532 @@
...
+    /// will be discarded). Note that `self.flatten_unnecessary_fragments(self.type_condition)` is always valid and useful, but it is
+    /// also possible to pass a `parent_type` that is more "restrictive" than the selection current type position
+    /// (as long as the top-level fields of this selection set can be rebased on that type).
+    ///
+    // PORT_NOTE: this is now module-private, because it looks like it *can* be. If some place
+    // outside this module *does* need it, feel free to mark it pub(crate).
+    // PORT_NOTE: in JS, this was called "normalize".
+    // PORT_NOTE: in JS, this had a `recursive: false` flag, which would only apply the
+    // simplification at the top level. This appears to be unused.
+    pub(super) fn flatten_unnecessary_fragments(
```
dariuszkuc, line 502: maybe this should just be `flatten`?

goto-bus-stop: when we talked about it about a month ago I got the sense that people preferred being specific about it affecting fragments

PR #5605, opened 2024-07-04
fix(federation): remove temporary string allocations from position `try_get()` errors

goto-bus-stop:
In our schema position code, we have two functions on every position type: `get()` and `try_get()`. `get()` returns a `Result`, and `try_get()` returns an `Option`. `try_get()` is `get()` but it discards the error.

`get()` returns a `FederationError`, which includes a formatted string explaining the error. Both functions can be quite hot. The `try_get()` case called on a position that does not exist allocates a string only to immediately discard it, which makes it much more expensive than expected.

This changes `get()` to use a new error type that formats lazily, and only costs some refcount increments. It converts into a `FederationError` using `?` so usage sites don't need to update.

Ideally it would avoid the refcount increases, but at the moment that would require adding very many branches to the error type for each possible position. With the error type I used here, I just decompose certain kinds of positions into their `Name` parts so we can reuse the same error branch for many types.


lrlna (approved):
let's do it!

PR #5613, opened 2024-07-05
fix(federation): several reuse_fragments() performance improvements

goto-bus-stop:
This does *not* fix our trouble yet, but it addresses a difference in JS and Rust implementations that made Rust substantially slower. After these changes the `FieldsInSetCanMerge` validation doesn't dominate time spent on the queries I'm using.

Also added a cache for `FragmentRestrictionsAtType`, just like the JS has. Plus some miscellaneous smaller improvements.

Finally, renamed the extremely generic `optimize()` to `reuse_fragments()` :)

With one example query:
| Before | After |
|-|-|
| 14.91s | 5.12s |


apollo-federation/src/operation/mod.rs
```diff
@@ -3042,8 +3042,17 @@ impl SelectionSet {
...
     }
 
+    /// In a normalized selection set containing only fields and inline fragments,
+    /// iterate over all the fields that may be selected.
+    ///
+    /// # Preconditions
+    /// The selection set must not contain named fragment spreads.
+    pub(crate) fn field_selections(&self) -> FieldSelectionsIter<'_> {
+        FieldSelectionsIter::new(self.selections.values())
+    }
```
goto-bus-stop, line 3052: We do not use the whole `CollectedFieldInSet` data structure in `optimize()` so we can use an iterator (that has a small temporary vec inside) instead of a big temporary vec.

apollo-federation/src/operation/optimize.rs
```diff
@@ -502,7 +504,7 @@ impl FieldsConflictValidator {
         &self,
         mut iter: impl Iterator<Item = &'a FieldsConflictValidator>,
     ) -> Result<bool, FederationError> {
-        iter.try_fold(true, |acc, v| Ok(acc && self.do_merge_with(v)?))
+        iter.try_fold(true, |acc, v| Ok(acc && v.do_merge_with(self)?))
```
goto-bus-stop, line 507: This shouldn't make a difference but it aligns with what JS does just in case.

apollo-federation/src/operation/optimize.rs
```diff
@@ -894,19 +891,25 @@ impl SelectionSet {
...
-                // Special case: Found a fragment that covers the full selection set.
-                return Ok(candidate.into());
-            }
-            // Note that if a fragment applies to only a subset of the sub-selections, then we
-            // really only can use it if that fragment is defined _without_ directives.
-            if !candidate.directives.is_empty() {
-                continue; // Not eligible as a partial selection; Skip it.
+            match res {
+                Containment::Equal if full_match_condition.check(&candidate) => {
+                    if !validator.check_can_reuse_fragment_and_track_it(&at_type)? {
```
goto-bus-stop, line 896: This is the big difference: `check_can_reuse_fragment` is where `do_merge_with` is called, potentially a lot, and it is potentially expensive for big selection sets. JS only calls it in this rare condition, while we were calling it all the time (and in another spot that I still need to add in Rust).

apollo-federation/src/operation/optimize.rs
```diff
@@ -1403,19 +1406,12 @@ impl SelectionSet {
...
-                let self_selections = Arc::make_mut(&mut self.selections);
-                self_selections.clear();
-                self_selections.insert(optimized.into());
-            }
-
-            InlineOrFragmentSelection::InlineFragment(inline_fragment) => {
-                // Note: `inline_fragment.selection_set` can't be moved (since it's inside Arc).
-                // So, it's cloned.
-                *self = inline_fragment.selection_set.clone();
+        *self = match optimized {
```
goto-bus-stop, line 1409: Replacing the whole selection set avoids potentially cloning a big `self.selections` map only to immediately clear it.

apollo-federation/src/operation/mod.rs
```diff
@@ -3199,6 +3208,36 @@ impl IntoIterator for SelectionSet {
...
+            None => {
+                self.stack.pop();
+                self.next()
+            }
+            Some(Selection::Field(field)) => Some(field),
+            Some(Selection::InlineFragment(frag)) => {
+                self.stack.push(frag.selection_set.selections.values());
+                self.next()
+            }
+            Some(Selection::FragmentSpread(_frag)) => unreachable!(),
```
goto-bus-stop, line 3236: TODO: maybe `optimize()` should check that no fragment spreads are anywhere in the selection set so it can error out early.

apollo-federation/src/operation/mod.rs
```diff
@@ -785,6 +784,10 @@ impl Selection {
         }
     }
 
+    fn sub_selection_type_position(&self) -> Option<CompositeTypeDefinitionPosition> {
```
goto-bus-stop, line 787: Just cleanup: this function exists on, and was used as  `.element().sub_selection_type_position()`, but doing that requires creating an `OperationElement` (in JS the main structures are built on top of `OperationElement`, but in Rust they are not, so it requires conversion Arc copies). The exact same data already exists on the selection set type so we can just return it directly without that conversion.

apollo-federation/src/operation/mod.rs
```diff
@@ -3382,21 +3427,6 @@ fn gen_alias_name(base_name: &Name, unavailable_names: &HashMap<Name, SeenRespon
     }
 }
 
-pub(crate) fn subselection_type_if_abstract(
```
goto-bus-stop, line 3385: Removed because we can just use `AbstractTypeDefinitionPosition::try_from()`.

apollo-federation/src/operation/optimize.rs
```diff
@@ -372,29 +369,27 @@ struct FieldsConflictValidator {
...
+                let entry = at_response_name
+                    .entry(field_selection.field.clone())
+                    .or_default();
+                if let Some(ref field_selection_set) = field_selection.selection_set {
+                    entry.push(field_selection_set);
+                } else {
+                    // Note that whether a `FieldSelection` has a sub-selection set or not is entirely
+                    // determined by whether the field type is a composite type or not, so even if
+                    // we've seen a previous version of `field` before, we know it's guaranteed to have
+                    // no selection set here, either.
```
goto-bus-stop, line 391: TODO(@goto-bus-stop): maybe i need to remove this comment

dariuszkuc (approved):

lrlna (approved):
@goto-bus-stop your PRs are such a joy to read! Thank you **so** much for putting in the effort to elaborate on all the important changes through the code. It really does make a difference in understanding what's currently happening and where you want this to go.

PR #5614, opened 2024-07-05
Create the invalidation endpoint for entity caching

bnjjj:
Create the invalidation endpoint for entity caching

TODO:
- [x] the response to invalidation requests on the endpoint returns the number of invalidated entries. Could we return errors too there? Users would be able to react to errors during invalidation (ex: to retry them)
- [x] the extensions contained a `Vec<InvalidationRequest>` but the endpoint only accept a single `InvalidationRequest`. Do we want to align them?
- [x] the endpoint directly calls the `handle_request` function instead of sending the request to the invalidation task. We may want to have all requests go through that task. That implies passing a channel sender along with the requests, to get notified once the invalidation is finished). It might also requires transmitting the current span too, so the invalidation execution happens under the same trace
- [x] handle the `shared_key` parameter
---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [x] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-109]: https://apollographql.atlassian.net/browse/ROUTER-109?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apollo-router/src/plugins/cache/invalidation.rs
```diff
@@ -36,33 +71,53 @@ impl Invalidation {
...
-    ) -> Result<(), BoxError> {
+    ) -> Result<u64, BoxError> {
         if self.enabled {
             let mut sink = self.handle.clone().into_sink();
-            sink.send((origin, requests)).await.map_err(|e| e.message)?;
-        }
+            let (response_tx, mut response_rx) = broadcast::channel(1);
+            sink.send((requests, origin, response_tx))
+                .await
+                .map_err(|e| e.message)?;
```
garypen, line 96: Could this be the cause of the BAD_REQUEST that we are seeing in performance testing?

bnjjj: I think so

docs/source/configuration/entity-caching.mdx
```diff
@@ -92,6 +92,11 @@ preview_entity_cache:
     timeout: 5ms # Optional, by default: 2ms
     ttl: 24h # Optional, by default no expiration
 
+  # Configure invalidation endpoint
+  invalidation:
+    listen: 127.0.0.1:4000
+    path: /invalidation
+
```
garypen, line 99: Do we want to remove this from the PR?

garypen (changes requested):

apollo-router/src/configuration/snapshots/apollo_router__configuration__tests__schema_generation.snap
```diff
@@ -3518,6 +3523,24 @@ expression: "&schema"
...
+          "description": "#/definitions/ListenAddr"
+        },
+        "path": {
+          "description": "Specify on which path you want to listen for invalidation endpoint. (default: /invalidation)",
+          "type": "string"
+        }
+      },
+      "required": [
+        "listen",
+        "path"
```
IvanGoncharov, line 3540: The path description says it has a default value, but it's required at the same time.
Maybe the default should be documented here:
https://github.com/apollographql/router/pull/5614/files#diff-ea34c7fa4e3f6984aae7f7e0d20d22930d5386763c6be9d9fb2f017adf16f746R5601

bnjjj: I removed the comment and default implementation

BrynCooke (approved):

garypen (approved):

PR #5618, opened 2024-07-08
chore(ci): Enable Codecov on the `dev` branch

abernix:
Enable Codecov on the `dev` branch

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,7 @@ orbs:
   gh: circleci/github-cli@2.3.0
   slack: circleci/slack@4.12.6
   secops: apollo/circleci-secops-orb@2.0.7
-
+  codecov: codecov/codecov@4.0.1
 executors:
   amd_linux_build: &amd_linux_build_executor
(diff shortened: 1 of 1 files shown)


Geal:
we don't record coverage by default, so it needs a step with `cargo-tarpaulin` like in https://github.com/rust-bakery/nom/blob/main/.github/workflows/ci.yml#L189-L193

```
      - name: Run cargo tarpaulin
        uses: actions-rs/cargo@v1
        with:
          command: tarpaulin
          args: --output-dir coverage --out xml --workspace --exclude benchmarks
```

abernix:
I agree, not done with the PR right now (going through the workflow with Codecov) however, I was planning to proceed with what I'd used successfully with `nextest` recently — thoughts? Docs here: https://nexte.st/docs/integrations/test-coverage/#llvm-cov

Geal:
that's worth a try

abernix:
Turns out this was **so** slow it is unusable — Rust code coverage is not quick on our repo.

PR #5621, opened 2024-07-08
Add support for invalidation by subgraph name and/or type name

garypen:
Scan redis using supplied subgraph and entity type name and delete the matching keys.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/plugins/cache/entity.rs
```diff
@@ -539,6 +539,7 @@ async fn cache_lookup_root(
 
     let key = extract_cache_key_root(
         &name,
+        None,
```
garypen, line 542: Question for reviewers: Is there a way to figure out the entity type here?

apollo-router/src/plugins/cache/entity.rs
```diff
@@ -842,14 +845,17 @@ fn extract_cache_key_root(
...
-    // - subgraph name: caching is done per subgraph
+    // - subgraph name: subgraph name
+    // - entity type: entity type
     // - query hash: invalidate the entry for a specific query and operation name
     // - additional data: separate cache entries depending on info like authorization status
     let mut key = String::new();
     let _ = write!(
         &mut key,
-        "subgraph:{subgraph_name}:Query:{query_hash}:{additional_data_hash}"
+        "subgraph:{subgraph_name}:type:{entity_type}:hash:{query_hash}:data:{additional_data_hash}"
```
garypen, line 858: drive-by: fixing to match changes in key structure in spec.

apollo-router/tests/integration/redis.rs
```diff
@@ -416,7 +416,7 @@ async fn entity_cache() -> Result<(), BoxError> {
     insta::assert_json_snapshot!(response);
 
     let s:String = client
-          .get("subgraph:products:Query:0df945dc1bc08f7fc02e8905b4c72aa9112f29bb7a214e4a38d199f0aa635b48:d9d84a3c7ffc27b0190a671212f3740e5b8478e84e23825830e97822e25cf05c")
+          .get("subgraph:products:type:Query:hash:0df945dc1bc08f7fc02e8905b4c72aa9112f29bb7a214e4a38d199f0aa635b48:data:d9d84a3c7ffc27b0190a671212f3740e5b8478e84e23825830e97822e25cf05c")
```
garypen, line 419: drive-by: fixing to match changes in key structure in spec.

apollo-router/tests/integration/redis.rs
```diff
@@ -743,7 +743,7 @@ async fn entity_cache_authorization() -> Result<(), BoxError> {
...
```
garypen, line 746: drive-by: fixing to match changes in key structure in spec.

apollo-router/src/plugins/cache/entity.rs
Geal: * get the schema at plugin creation (there's a different plugin init function signature for that)
* get the API schema which is a `apollo_copiler::Schema`
* `schema.schema_definition.query.as_ref().map(|q| q.name.to_string())`

Geal (changes requested):
could you split that test from the subgraph level invalidation test?

garypen:
> could you split that test from the subgraph level invalidation test?

done: 5256aff5ea5cf1d8ad47b78202bb25277e0554d1

Geal (approved):

PR #5623, opened 2024-07-08
Remove JavaScript-based API schema implementation

goto-bus-stop:
This change removes the experimental configuration to pick the API schema generation mode. Now, only Rust-based API schema is available. We have been running this in `both` mode in all routers for a few versions now, and have observed 0 mismatches, so we can be confident in the new implementation.

Note that the JS query planner still uses its own API schema implementation internally. This introduces some risk of drift between the implementations. We don't intend to change the API schema in JS or in Rust (as far as I know, that would be a breaking change), but after this PR we will not be notified if a difference is introduced.

This change was scheduled for a future breaking change release. We could still wait until then, I don't think this is very pressing, it's mostly a nice cleanup. In particular, we don't need the two-stage initialisation of the `Schema` type anymore, as we do not need to create a query planner to generate the API schema, which removes some error paths.

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [x] Integration Tests
    - [ ] Manual Tests

**Exceptions**

- Documentation is not relevant for this internal change.
- Testing has been completed in production!

**Notes**


apollo-router/src/configuration/mod.rs
```diff
@@ -161,10 +161,6 @@ pub struct Configuration {
     #[serde(default)]
     pub(crate) experimental_chaos: Chaos,
 
-    /// Set the API schema generation implementation to use.
-    #[serde(default)]
-    pub(crate) experimental_api_schema_generation_mode: ApiSchemaMode,
```
SimonSapin, line 166: Yes, old configs using this will (I think) cause a “Additional properties are not allowed” error.

To avoid that please add:
* A migration file under `apollo-router/src/configuration/migrations/`
* A test config file under `apollo-router/src/configuration/testdata/migrations/`

SimonSapin (approved):
Approved assuming a migration is added and maybe a test simplified

apollo-router/src/query_planner/tests.rs
```diff
@@ -433,7 +433,7 @@ async fn defer_if_condition() {
           }"#;
 
     let schema = include_str!("testdata/defer_clause.graphql");
-    // we need to use the planner here instead of Schema::parse_test because that one uses the router bridge's api_schema function
+    // we need to use the planner here instead of Schema::parse because that one uses the router bridge's api_schema function
     // does not keep the defer directive definition
     let planner =
         BridgeQueryPlanner::new(schema.to_string(), Arc::new(Configuration::default()), None)
```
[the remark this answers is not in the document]
goto-bus-stop: Ah, yes! No! 🥳

apollo-router/src/configuration/mod.rs
goto-bus-stop: Thanks!

SimonSapin (approved):

lrlna (approved):

lrlna (approved):

Geal (approved):
great!

Geal:
I think there's a follow up we can make to that PR, where schema parsing does not happen in `BridgeQueryPlanner` but in the state machine, where we check for schema features according to the licence

SimonSapin:
> I think there's a follow up we can make to that PR, where schema parsing does not happen in `BridgeQueryPlanner` but in the state machine, where we check for schema features according to the licence

PR https://github.com/apollographql/router/pull/5736 does this

PR #5627, opened 2024-07-08
Refactoring the `validation` method and moving methods into separate modules

tayrrible:
*Description here*

Fixes #**issue_number**

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-federation/src/sources/connect/models/validation/source_name.rs
```diff
@@ -0,0 +1,67 @@
...
+                        .into_iter()
+                        .collect(),
+                });
+        }
+    }
+
+    messages
+}
+
+fn connect_directive_name_coordinate(
```
dylan-apollo, line 60: We should keep all the coordinates together, I think, for easier discovery. Maybe in their own `schema_coordinates` module. Still tinkering with the idea of making them into structs instead of format strings, which I think could be a bit more efficient 🤔.

apollo-federation/src/sources/connect/models/validation/mod.rs
```diff
@@ -42,16 +42,18 @@
 - Missing $this
 
 */
+mod entities;
```
dylan-apollo, line 45: This probably won't have _all_ entity checks in it (e.g., `@key` stuff), so maybe this should be `entity` to refer to the arg?

apollo-federation/src/sources/connect/models/validation/mod.rs
```diff
@@ -42,16 +42,18 @@
...
+mod http_headers;
+mod http_method;
+mod selection;
+mod source_name;
```
dylan-apollo, line 49: `SourceName` should probably also end up in here.

dylan-apollo (approved):
Definitely a big step in the right direction! We don't need to be too prescriptive about structure, since I'm sure our needs will change more as we continue to add validations, I do want to _generally_ try and think about how we want to organize things, though. Meaning, the point of splitting up the code is to make it easier to find things, so we should make sure we're organizing in a way that achieves that.

To me, it makes the most sense to group things by where in the schema we are looking. So some logical "groups" could be:
1. `@source` directive
2. Types (e.g., validating no `Subscription`
3. Fields (if there are validations other than `@connect`)
4. Then _pieces_ of `@connect`, since it's so big:
    1. `entity`
    2. `selection`
    3. `http`
    
Any other ideas for a logical way to split things up that makes it easy to find what you're looking for?

apollo-federation/src/sources/connect/models/validation/mod.rs
```diff
@@ -370,32 +385,13 @@ fn validate_field(
...
-                "{coordinate} must specify an HTTP method.",
-                coordinate =
-                    connect_directive_http_coordinate(connect_directive_name, object, &field.name),
-            ),
-            locations: Location::from_node(http_arg_location, source_map)
-                .into_iter()
-                .collect(),
-        });
-    }
+    errors.extend(validate_http_method(
```
pubmodmatt, line 388: The code above to gather the HTTP methods seems like it should also be moved into the `http_method` module. Just the fact that this takes in an array of HTTP methods is tied to the logic that validates whether more than one is specified. Separating that logic across two files introduces an unnecessary interdependency there.

apollo-federation/src/sources/connect/models/validation/mod.rs
```diff
@@ -476,241 +451,24 @@ fn validate_field(
         }
     }
 
-    // validate header arg
-    let http_headers_arg = http_arg
+    if let Some(headers) = http_arg
```
pubmodmatt, line 454: Similarly, here the logic to assemble the headers probably belongs in the `http_heasers` module.

apollo-federation/src/sources/connect/models/validation/source_name.rs
pubmodmatt: The way this is structured now is that there is a top level `validate(Schema)` method, that knows how to validate everything. Even if some bits of code are moved into sub-modules, the top level method still calls into those modules directly in a tightly coupled way, passing in the exact arguments they need.

Instead, the way I've generally seen this kind of validation done is to decouple the validation engine from the individual validators. The top level `validate` method just runs a validation engine, that iterates over the Schema nodes and calls out to any validators that have registered an interest in the current node.

So a validator might indicate that it's interested in any field with a `@connect` directive (or maybe more specifically, a `@connect` directive with a `headers` attribute). When the validation engine sees such a node, it calls the validator and passes in that field. The validator implementation then does whatever it needs to do.

Adding in a new validator is then very decoupled - no changes are needed to the engine itself, you just add a new validator and usually a line of code to register the new validator.

This might all be overkill for now, since we're just looking at a couple of directives, but as things grow this kind of structure can be beneficial.

dylan-apollo: That sort of abstraction could definitely be useful if it makes it easier to read the code. It kind of gets at my question around overall structure above. If we organize by individual validation rule, then we probably want some pattern that decouples from the schema structure. If we organize by schema structure, I think that level of decoupling wouldn't provide the same value.

We also only have the one consumer, so whatever organization/abstraction we go for here only needs to be optimized for readability, not flexibility.

apollo-federation/tests/query_plan/build_query_plan_tests.rs
```diff
@@ -541,75 +541,25 @@ fn key_where_at_external_is_not_at_top_level_of_selection_of_requires() {
...
-          Sequence {
-            Fetch(service: "A") {
-              {
-                u {
-                  __typename
-                  k1 {
-                    id
-                  }
-                }
+    QueryPlan {
```
pubmodmatt, line 544: These tests are broken on `dev` and intentionally ignored with `should_panic`. Lenny reached out to the owner but I'm not sure if they have a plan to fix this. But for the `connectors` branch, we need to minimize changes to only things related to connectors, so you should probably remove this diff from your PR. I've just been running 'cargo insta reject` to clear out these snapshots, then running just the connectors tests that changed, then `cargo insta review`. That avoids picking up these snapshots.

dylan-apollo: I do `cargo t validation -p apollo-federation` when working on validations to avoid even compiling router 😁

tayrrible: Definitely didn't mean for this test to get included; I think I had happy fingers when going through skipping and approving 😅

ty for the catch!

apollo-federation/src/sources/connect/models/test_data/validation/disallowed_federation_imports.graphql
tayrrible, line 1: not deleted, just moved

apollo-federation/src/sources/connect/models/snapshots/validation_tests@disallowed_federation_imports.graphql.snap
tayrrible, line 1: not deleted, just moved

pubmodmatt (approved):

PR #5629, opened 2024-07-09
perf: don't re-create meter and instruments on every calls

bnjjj:
The creation of otel instruments using a regex is no longer part of the hot path. Now we create these instruments when starting the telemetry plugin and not in every serives.

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [x] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/plugins/telemetry/config_new/instruments.rs
```diff
@@ -100,8 +116,118 @@ impl InstrumentsConfig {
             .defaults_for_levels(self.default_requirement_level, TelemetryDataKind::Metrics);
     }
 
-    pub(crate) fn new_router_instruments(&self) -> RouterInstruments {
+    pub(crate) fn new_static_router_instruments(&self) -> HashMap<String, StaticInstrument> {
```
BrynCooke, line 119: Can we consider changing the term `static` to `built_in`? Static makes me think that the instruments have been changed are stored in a static somewhere.

apollo-router/src/plugins/telemetry/mod.rs
```diff
@@ -275,6 +281,57 @@ impl Plugin for Telemetry {
             ::tracing::warn!("telemetry.instrumentation.spans.mode is currently set to 'deprecated', either explicitly or via defaulting. Set telemetry.instrumentation.spans.mode explicitly in your router.yaml to 'spec_compliant' for log and span attributes that follow OpenTelemetry semantic conventions. This option will be defaulted to 'spec_compliant' in a future release and eventually removed altogether");
         }
 
+        let graphql_custom_instruments = if cfg!(test) {
```
BrynCooke, line 284: Can we factor out this section of code and the related similar one in activate to a separate `fn` and remove the `cfg!(test)` logic? It doesn't matter if we instruments before activate as no-ops.

BrynCooke (changes requested):

apollo-router/src/plugins/telemetry/config_new/instruments.rs
bnjjj: Fixed https://github.com/apollographql/router/pull/5629/commits/c4a8b14787030df5190bcecc59b4d53fdeb99dfb

apollo-router/src/plugins/telemetry/mod.rs
bnjjj: Fixed https://github.com/apollographql/router/pull/5629/commits/c4a8b14787030df5190bcecc59b4d53fdeb99dfb

BrynCooke (approved):

PR #5633, opened 2024-07-09
Handle subgraph response with -1 values inside error locations

IvanGoncharov:
GraphQL spec require that both "line" and "column" are positive numbers. 
However GraphQL Java and GraphQL Kotlin return `{ "line": -1, "column": -1 }` if they can't determine error location inside query. 
So instead of failing we now just remove such locations.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [x] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-393]: https://apollographql.atlassian.net/browse/ROUTER-393?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apollo-router/src/graphql/mod.rs
+++ b/apollo-router/src/graphql/mod.rs
@@ -123,29 +123,30 @@ impl Error {
         let mut object =
             ensure_object!(value).map_err(|error| FetchError::SubrequestMalformedResponse {
                 service: service_name.to_string(),
-                reason: error.to_string(),
+                reason: format!("invalid error within `errors`: {}", error),
             })?;
 
         let extensions =
             extract_key_value_from_object!(object, "extensions", Value::Object(o) => o)
                 .map_err(|err| FetchError::SubrequestMalformedResponse {
                     service: service_name.to_string(),
-                    reason: err.to_string(),
+                    reason: format!("invalid `extensions` within error: {}", err),
                 })?
                 .unwrap_or_default();
         let message = extract_key_value_from_object!(object, "message", Value::String(s) => s)
             .map_err(|err| FetchError::SubrequestMalformedResponse {
                 service: service_name.to_string(),
-                reason: err.to_string(),
+                reason: format!("invalid `message` within error: {}", err),
             })?
             .map(|s| s.as_str().to_string())
             .unwrap_or_default();
         let locations = extract_key_value_from_object!(object, "locations")
+            .map(skip_invalid_locations)
             .map(serde_json_bytes::from_value)
             .transpose()
             .map_err(|err| FetchError::SubrequestMalformedResponse {
                 service: service_name.to_string(),
-                reason: err.to_string(),
+                reason: format!("invalid `locations` within error: {}", err),
             })?
             .unwrap_or_default();
         let path = extract_key_value_from_object!(object, "path")
@@ -153,7 +154,7 @@ impl Error {
             .transpose()
             .map_err(|err| FetchError::SubrequestMalformedResponse {
                 service: service_name.to_string(),
-                reason: err.to_string(),
+                reason: format!("invalid `path` within error: {}", err),
             })?;
 
         Ok(Error {
@@ -163,6 +164,19 @@ impl Error {
             extensions,
         })
     }
+}
+
+// GraphQL spec require that both "line" and "column" are positive numbers.
+// However GraphQL Java and GraphQL Kotlin return `{ "line": -1, "column": -1 }`
+// if they can't determine error location inside query.
+// This function removes such locations from suplied value.
+fn skip_invalid_locations(mut value: Value) -> Value {
+    value.as_array_mut().map(|array| {
+        array.retain(|location| {
+            location.get("line") != Some(&json!(-1)) || location.get("column") != Some(&json!(-1))
+        })
+    });
+    value
 }
 
 /// Displays (only) the error message.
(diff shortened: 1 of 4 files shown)


apollo-router/src/graphql/mod.rs
```diff
@@ -165,6 +166,19 @@ impl Error {
...
 }
 
+// GraphQL spec require that both "line" and "column" are positive numbers.
+// However GraphQL Java and GraphQL Kotlin return `{ "line": -1, "column": -1 }`
+// if they can't determine error location inside query.
+// This function removes such locations from suplied value.
+fn skip_invalid_locations(mut value: Value) -> Value {
+    if let Some(array) = value.as_array_mut() {
+        array.retain(|location| {
+            location.get("line") != Some(&json!(-1)) || location.get("column") != Some(&json!(-1))
```
bnjjj, line 176: We need to check but I think it would be better to use enum Value directly instead of using the json macro which could parse

bnjjj (changes requested):

apollo-router/src/graphql/mod.rs
IvanGoncharov: Fixed ✅

apollo-router/src/graphql/mod.rs
```diff
@@ -165,6 +166,20 @@ impl Error {
...
 }
 
+/// GraphQL spec require that both "line" and "column" are positive numbers.
+/// However GraphQL Java and GraphQL Kotlin return `{ "line": -1, "column": -1 }`
+/// if they can't determine error location inside query.
+/// This function removes such locations from suplied value.
+fn skip_invalid_locations(mut value: Value) -> Value {
+    if let Some(array) = value.as_array_mut() {
+        array.retain(|location| {
+            location.get("line") != Some(&Value::from(-1))
```
Geal, line 176: this only handles the case where the line or column is "-1", what about any other negative number? What happens if we get a float instead?

garypen: Good point, since spec compliance implies `> 0` to me (since 0 is not a positive number).
Probably better as: 
```
        array.retain(|location| {
            location
                .get("line")
                .map_or(false, |l| l.as_u64().map_or(false, |v| v > 0))
                && location
                    .get("column")
                    .map_or(false, |l| l.as_u64().map_or(false, |v| v > 0))
        })
```
or something like that.

garypen: Probably need to fix the tests below since I see `0` is being used in them.

IvanGoncharov: > this only handles the case where the line or column is "-1", what about any other negative number? What happens if we get a float instead?

@Geal @garypen I don't think we should handle all the cases and just focus on `-1` since it is what graphql-java and graphql-kotlin return.
Those implementations have hardcoded `-1` in their code bases, so there is a 0% chance that we will get other negative values or floats.

At the moment, it's a workaround we added to support customers until they switch to fixed versions of graphql-java and graphql-kotlin. But if we change it to skip all the `<= 0` values or even float, it means we made a conscious decision to relax validation on `locations`.

In my opinion, we shouldn't relax validation and should adhere to the spec. We could add a workaround only in case customers are stuck with a particular graphql implementation to fix this particular case.

IvanGoncharov: > Probably need to fix the tests below since I see 0 is being used in them.

@garypen Thanks 👍 
I will fix the tests.

IvanGoncharov: In my code I check that both `line` and `column` are `-1` so I'm certain that I can safely remove it without alerting users.

But for other invalid values situation is a different, we silently remove `locations` without users being notified and have an option to fix his subgraph.

bnjjj (approved):

apollo-router/src/graphql/mod.rs
abernix: I do think there's something to be said about adhering to the spec and only adding exceptions where exceptions are warranted/known, rather than being too relaxed.  The spec could/should/may be more clear about saying positive integers rather than numbers (e.g., floats), but it is also somewhat implied that they are integers since the values themselves (`column` and `line`) are meant to point to character offsets.

While impacted versions of this are never going away and we'll probably have to keep this exception in forever, do we have any interest in opening an issue or PR upstream with `graphql-java`?   It could very much be welcome.

apollo-router/tests/integration/subgraph_response.rs
```diff
@@ -0,0 +1,214 @@
...
+    assert_eq!(
+        serde_json::from_str::<serde_json::Value>(&response.text().await?)?,
+        json!({
+            "data": { "topProducts": null },
+            "errors": [{
+                "message":"Some error on subgraph",
+                "path":["topProducts"]
+            }]
+        })
+    );
```
Geal, line 165: why are the assert and the mocked response using a different top level operation?

Geal (approved):

garypen (approved):

apollo-router/tests/integration/subgraph_response.rs
IvanGoncharov: @Geal Thank you for catching this!
I will fix it in a minute.

IvanGoncharov: fixed ✅

PR #5634, opened 2024-07-09
fix(federation): fix logic in computing non trivial follow up edges

dariuszkuc:
Fix handling of `SubgraphEnteringTransition` which should skip `RootTypeResolution` edges. This fixes the issue with redundantly processing same edges over and over again.

Fix handling of `KeyResolution` which should compare whole selection sets of the conditions vs just the selections.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


lrlna:
@dariuszkuc can you add a regression test for this?

goto-bus-stop:
do we still want a regression test here? (is it possible to write one, or is it only a performance difference?)

otherwise we should merge it I think

dariuszkuc:
Superseded by https://github.com/apollographql/router/pull/5766

PR #5636, opened 2024-07-09
Use OpTree concatenation in serial query planning 

TylerBloom:
This PR addresses an issue in the `compute_root_serial_dependency_graph` function. While iterating through the top-level fields, sucessive `OpTree`s from the same subgraph were being merged instead of concatenated. This can cause mutations to be merged together and effective forgotten.

Fixes FED-336.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-federation/src/query_graph/path_tree.rs
+++ b/apollo-federation/src/query_graph/path_tree.rs
@@ -283,6 +283,40 @@ where
             })
     }
 
+    pub(crate) fn concat(self: &Arc<Self>, other: &Arc<Self>) -> Arc<Self> {
+        if Arc::ptr_eq(self, other) {
+            return self.clone();
+        }
+        assert!(
+            Arc::ptr_eq(&self.graph, &other.graph),
+            "Cannot merge path tree build on another graph"
+        );
+        assert_eq!(
+            self.node, other.node,
+            "Cannot merge path trees rooted different nodes"
+        );
+        if other.childs.is_empty() {
+            return self.clone();
+        }
+        if self.childs.is_empty() {
+            return other.clone();
+        }
+        let mut childs = Vec::with_capacity(self.childs.len() + other.childs.len());
+        childs.extend_from_slice(&self.childs);
+        childs.extend_from_slice(&other.childs);
+        Arc::new(Self {
+            graph: self.graph.clone(),
+            node: self.node,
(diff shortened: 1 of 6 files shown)


apollo-federation/tests/query_plan/build_query_plan_tests/mutations.rs
```diff
@@ -0,0 +1,78 @@
...
+              }
+            },
+            Fetch(service: "SubgraphB") {
+              {
+                updateInBOne: updateFooInB {
+                  id
+                  baz
+                }
+              }
+            },
```
TylerBloom, line 73: This is still incorrect (or at least not what I intend to happen). The QP mistakes each root field as belonging to the same subgraph.

apollo-federation/src/query_graph/path_tree.rs
```diff
@@ -283,6 +283,40 @@ where
             })
     }
 
+    pub(crate) fn concat(self: &Arc<Self>, other: &Arc<Self>) -> Arc<Self> {
```
goto-bus-stop, line 286: Could we take ownership of `self` here and try to mutate it?
```suggestion
    pub(crate) fn concat(self: Arc<Self>, other: &Arc<Self>) -> Arc<Self> {
```

And below `Arc::make_mut(&mut self).childs.extend_from_slice(&other.childs)`

goto-bus-stop (changes requested):
Looks good! Please document on `concat` that it panics if the graphs or root nodes of the paths are not the same.

apollo-federation/src/query_graph/path_tree.rs
TylerBloom: Ya, that make sense. I just copy-pasted the signature of the `merge` method. I think if that is the route we want to take, the method should take a `&mut self` because there is no case in which we will not need to call `Arc::make_mut`

goto-bus-stop: That works for me! Though then maybe I would rather name it `extend` or something (concat typically produces a copy)

goto-bus-stop (approved):
Thank you!

PR #5638, opened 2024-07-10
Introduce make_fake_batch() to avoid racy caches and redact more otel trace details

garypen:
It transpires that there are multiple race conditions in our tests which makes snapshotting traces problematic. This problem is particularly acute when we are testing batch requests, because of the way we build up our batch requests. The most significant impact has been on the otel traces tests which have been very flaky.

A number of the tests expose subtle races in their expectations around span creation. This is more complex and warrants a detailed explanation.

We already have a mechanism for generating a query from a fake supergraph query builder. When we added support for batching, we leveraged this functionality to convert a simple single request into a batch request by doing something (conceptually at least) like this:

map the request, so that we clone it and wrap it in an array which now contains two identical requests.

That is simple and easy to implement, so we've used it in a few places in our tests.

However, when the router executes....

At various places in the router we check if items are cached to decide whether or not to do some processing. That processing may involve the generation of a tracing span. If we don't do the processing, because, let's say, data is in a cache, then we now have a potential race condition, because not doing the processing means the various snapshots that we rely on in our tests will (or won't) match.

This, in a nutshell, are the problems that this PR fixes.

1. Make items in a batch different enough that caching effects are avoided.
2. Redact various details so that sequencing is not as much of an issue in the otel traces tests

This results in a lot of updated snapshots, but the changes aren't actually that significant.

Unblocks: https://github.com/apollographql/router/pull/5644


timbotnik (approved):
This seems fine - you are more of an expert here wrt batching.  It would be nice if we could have some of the tests not rely on batching and keep a tighter match on important span attributes (the signature is one of the only "must have" pieces of the Apollo OTel trace ingestor, and I noticed in this PR we are now redacting it from snapshots).

garypen:
> This seems fine - you are more of an expert here wrt batching. It would be nice if we could have some of the tests not rely on batching and keep a tighter match on important span attributes (the signature is one of the only "must have" pieces of the Apollo OTel trace ingestor, and I noticed in this PR we are now redacting it from snapshots).

I know it's not ideal. I think we'd have to have a different version of the `assert_report` macro for batching vs non-batching. The non-batching wouldn't redact the additional details. That might be good enough.

I'll try it out later.

IvanGoncharov (approved):

garypen:
I modified the macro to only perform the additional redaction if the test is a batch test. I think that's a nice compromise and keeps the tighter match that @timbotnik requested above.

bnjjj (approved):

PR #5582, opened 2024-07-02
use spawn_blocking for parsing

Geal:
This leverages the work that @xuorig started in https://github.com/apollographql/router/pull/5235 with a number of necessary follow-up fixes to get it into the shape we'll need to land it.

I have checked manually that the span duplication in the snapshot has no impact on traces reported to aggregators, it's more of an artifact of our way to gather spans in the test.

---

**Checklist**

Complete the checklist (and note appropriate exceptions) before the PR is marked ready-for-review.

- [ ] Changes are compatible[^1]
- [ ] Documentation[^2] completed
- [ ] Performance impact assessed and acceptable
- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[^1]: It may be appropriate to bring upcoming changes to the attention of other (impacted) groups. Please endeavour to do this before seeking PR approval. The mechanism for doing this will vary considerably, so use your judgement as to how and when to do this.
[^2]: Configuration is an important part of many changes. Where applicable please try to document configuration examples.
[^3]: Tick whichever testing boxes are applicable. If you are adding Manual Tests, please document the manual testing (extensively) in the Exceptions.

PR #5644, opened 2024-07-10
Reintroduce "use spawn_blocking for parsing"

abernix:
> [!IMPORTANT]
> ## This is ready to review once #5638 lands.

Re-lands apollographql/router#5582 after it was reverted in apollographql/router#5643 as it seems to have been causing tests to flake at a somewhat pronounced rate on `dev` — always on Linux. We believe it's just actually the tests, or the way we tested, but we'll revert this for now so `dev` isn't flaking for everyone trying to develop against it.

And worth mentioning again, this was was originally @xuorig's in https://github.com/apollographql/router/pull/5235 which @Geal took over and @garypen ultimately landed in #5582 . 

Marking this as a draft for someone to look at it.

[ROUTER-426]: https://apollographql.atlassian.net/browse/ROUTER-426?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


garypen:
The code is fine. It was a source of testing fails because it triggered the behaviour described in #5638. As soon as #5638 merges, we should merge this.

garypen (approved):
#5638 must merge before this.

garypen (approved):

PR #5645, opened 2024-07-10
connectors: simplify testing for subgraph expansion

nicholascioli:
This PR makes it easier to add tests for subgraph expansion in a DRY way using macros. 

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-federation/src/sources/connect/expand/tests/mod.rs
```diff
@@ -1,64 +1,72 @@
-use insta::assert_debug_snapshot;
-use insta::assert_snapshot;
+/// Macro to quickly expand a test case based on a test schema name.
+macro_rules! test_expands {
+    ($($name:ident),* $(,)*) => {
+        $(
+            mod $name {
+                #[test]
+                fn it_expands_supergraph() {
```
pubmodmatt, line 7: Nitpick: maybe include the `$name` here so each one ends up with a different test name?

pubmodmatt (approved):

apollo-federation/src/sources/connect/expand/tests/mod.rs
nicholascioli: Using the name in the generated function requires using a procedural macro, like paste. I see that we're already using it in the router crate, so I can try adding it as a workspace dependency if you prefer.

pubmodmatt: Ah right. This is fine, not worth the additional effort.

PR #5646, opened 2024-07-10
Implement `$config` for connectors

dylan-apollo:
Mostly this implements the new `$config` syntax, where values in Router config can be accessed from connectors schema (which, in turn, gives users a way to pass env vars).

Since this was my first time exploring this code, I also made some improvements along the way—in particular getting rid of some errors that were statically not possible, reducing some cloning, etc.

Because config is set during expansion, we _could_, in theory, do some static verification on those values if we start validating router config as part of composition. That's a future goal, though.

[CNN-205]: https://apollographql.atlassian.net/browse/CNN-205?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apollo-router/src/spec/schema.rs
```diff
@@ -84,7 +85,14 @@ impl Schema {
 
         let mut api_schema: Option<_> = None;
         let mut connectors: Option<_> = None;
-        let expansion = expand_connectors(sdl).map_err(SchemaError::Connector)?;
+        let connectors_config = config
+            .apollo_plugins
+            .plugins
+            .get("preview_connectors")
```
lennyburdette, line 91: should we make this a const so it's easier to change when we drop "preview_"?

apollo-router/src/plugins/connectors/tests.rs
```diff
@@ -565,22 +609,31 @@ async fn execute(
...
-          "subgraphs": {
-            "connectors": {
-              "json": {
-                "override_url": connector_uri
-              }
-            }
-          }
-        }),
-    );
+    config_object
```
lennyburdette, line 612: is there a deep_merge fn or something similar this easier to read?

apollo-router/src/plugins/connectors/make_requests.rs
```diff
@@ -18,18 +19,19 @@ const REPRESENTATIONS_VAR: &str = "representations";
 const ENTITIES: &str = "_entities";
 const TYPENAME: &str = "__typename";
 
-#[derive(Debug, Default)]
-struct RequestInputs {
+#[derive(Clone, Debug, Default)]
+pub(crate) struct RequestInputs {
```
lennyburdette, line 23: not a change for this PR, but because we're storing more stuff from various places for use in URL templates and selections, and we end up cloning all of that stuff into a map, i've been wondering:

should we change the URLTemplate::generate_path and Selection::apply_with_vars to take some kind of map-like trait that allows us to store Arc's of config, context, request headers, and other values without cloning?

apollo-router/src/spec/schema.rs
dylan-apollo: Probably, I wasn't sure how this gets set in the first place—where does this config get defined? Is it related to the plugin name, and can that also be a const?

apollo-router/src/plugins/connectors/make_requests.rs
dylan-apollo: I thought the same thing—I'm going to add a public preview issue for generally auditing performance and doing a sweep of any easy clean-ups.

apollo-router/src/spec/schema.rs
lennyburdette: yeah it's the plugin name as registered, but i'm guessing that can't be a const: https://github.com/apollographql/router/blob/connectors/apollo-router/src/plugins/connectors/plugin.rs#L120

apollo-router/src/plugins/connectors/tests.rs
dylan-apollo: Not built in, but I found a crate which does it and added it.

lennyburdette (approved):

PR #5647, opened 2024-07-11
feat(xtask): add a shorthand to profile the planner

goto-bus-stop:
We are working through some known-slow operations in the native query planner, and this is a really useful way to generate a flame graph quickly and investigate specific queries.

```
cargo xtask fed-flame plan /path/to/query.graphql /path/to/supergraph.graphql
```

This compiles the apollo-federation CLI in release mode with debug info enabled, and then runs it with [samply], to produce a call tree and a flame graph. It can even show you exactly where in the code all the callsites are.

image

[samply]: https://github.com/mstange/samply


lrlna (changes requested):
Can you add a small blurb to the `xtask` readme and the `DEVELOPMENT.md` file on running this?

goto-bus-stop:
I was going to update the document in confluence instead of adding it here. fwiw the xtask readme.md is quite outdated and references commands that no longer exist 💀

goto-bus-stop:
updated the readme to list the xtasks that i use regularly 😄

lrlna (approved):
Thank you!!

goto-bus-stop:
Hmmm. I find that this hangs unexpectedly locally 🤔 
I would like to figure that out before I merge this and recommend people use it. It might be an issue with using `samply`.

goto-bus-stop:
Okay, it was probably an issue with how I spawned the subprocess, I've switched to xshell which appears to do the right thing.

PR #5659, opened 2024-07-15
fix(xtask): fix compilation

goto-bus-stop:
Since 0d2e347ffdb28cb7b4a7664e911a538d7616a61e I get a compile error when running anything `xtask`. I think maybe the `reqwest` update changed the set of tokio features that's enabled by `reqwest`?
```
error[E0599]: no function or associated item named `new_multi_thread` found for struct `tokio::runtime::Builder` in the current scope
   --> src/commands/changeset/mod.rs:240:34
    |
240 |         tokio::runtime::Builder::new_multi_thread()
    |                                  ^^^^^^^^^^^^^^^^
    |                                  |
    |                                  function or associated item not found in `Builder`
    |                                  help: there is an associated function with a similar name: `new_current_thread`
```

This patch enables "full" tokio features. That's excessive for what we use right now but I don't think it really hurts...

### DIFF
--- a/xtask/Cargo.lock
+++ b/xtask/Cargo.lock
@@ -419,7 +419,7 @@ checksum = "d4029edd3e734da6fe05b6cd7bd2960760a616bd2ddd0d59a0124746d6272af0"
 dependencies = [
  "cfg-if",
  "libc",
- "redox_syscall",
+ "redox_syscall 0.3.5",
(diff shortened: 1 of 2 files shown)


bnjjj (approved):

bnjjj:
thanks @goto-bus-stop I merged it because I needed that one locally too :)

PR #5667, opened 2024-07-16
fix(federation): reduce strcmp by swapping a couple of conditions

o0Ignition0o:


goto-bus-stop:
If we do this we should add a comment to these cases I think. They read better when the subgraph name is at the start, so just saying "we compare the subgraph name last because on average it improves performance" would be helpful

o0Ignition0o:
> If we do this we should add a comment to these cases I think. They read better when the subgraph name is at the start, so just saying "we compare the subgraph name last because on average it improves performance" would be helpful

true! adding comments

o0Ignition0o:
Sneaky whitespace 😅 rerunning CI!

PR #5671, opened 2024-07-16
docs: Dynatrace otlp port

lleadbet:
Adds a small note to the docs about using Dynatrace with HTTP OTLP, as it uses a non-standard port.


Meschreiber (approved):
Thanks for the addition!

abernix (approved):
Love a good docs contribution like this.  Thank you! This will be easily merged after we do the release — it addresses the tests that are failing here.

PR #5673, opened 2024-07-16
Fix Nightly Clippy lints

SimonSapin:
Clippy from Rust toolchain version `1.81.0-nightly (24d2ac0b5 2024-07-15)`

### DIFF
--- a/apollo-router/src/executable.rs
+++ b/apollo-router/src/executable.rs
@@ -2,8 +2,6 @@
 
 use std::cell::Cell;
 use std::env;
-use std::ffi::OsStr;
-use std::fmt;
 use std::fmt::Debug;
 use std::net::SocketAddr;
 use std::path::PathBuf;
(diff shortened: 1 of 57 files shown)


apollo-router/src/notification.rs
```diff
@@ -34,44 +34,21 @@ static NOTIFY_CHANNEL_SIZE: usize = 1024;
...
-
-impl<K, V> From<TrySendError<Notification<K, V>>> for NotifyError<V>
-where
-    K: Send + Hash + Eq + Clone + 'static,
-    V: Send + Clone + 'static,
-{
-    fn from(error: TrySendError<Notification<K, V>>) -> Self {
-        error.into()
-    }
-}
```
SimonSapin, line 73: Each of these three impls only called itself in infinite recursion

apollo-router/src/executable.rs
```diff
@@ -314,43 +311,6 @@ impl Opt {
     }
 }
 
-/// Wrapper so that clap can display the default config path in the help message.
-/// Uses ProjectDirs to get the default location.
-#[derive(Debug)]
-struct ProjectDir {
```
SimonSapin, line 320: Dead code is removed here and in a few other places

abernix (approved):

PR #5676, opened 2024-07-16
Add validation for abstract types

tayrrible:
*Description here*
image

Fixes #**issue_number**

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[CNN-273]: https://apollographql.atlassian.net/browse/CNN-273?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apollo-federation/src/sources/connect/validation/test_data/validation/disallowed_abstract_types.graphql
tayrrible, line 1: please lmk if it'd make more sense to split out the `union` and `interface` checks into separate files 🙏

apollo-federation/src/sources/connect/validation/mod.rs
```diff
@@ -147,23 +147,38 @@ pub fn validate(schema: Schema) -> Vec<Message> {
...
+                    .into_iter()
+                    .collect(),
+            }),
+            ExtendedType::Interface(interface) => connect_errors.push(Message {
+                code: Code::UnsupportedAbstractType,
+                message: "Abstract schema types, such as `interface`, are not supported when using connectors. You can check out our documentation at https://go.apollo.dev/connectors/best-practices#abstract-schema-types-are-unsupported.".to_string(),
+                locations: Location::from_abstract_type_node(interface, source_map, "interface")
+                    .into_iter()
+                    .collect(),
+            }),
```
dylan-apollo, line 176: Maybe we should put all of this into a single function called `validate_type` or something? Rather than letting this function continue to grow.

Then we could go back to `flat_map` instead of `for_each`, which I like better 😁

apollo-federation/src/sources/connect/validation/mod.rs
```diff
@@ -581,6 +597,23 @@ impl Location {
...
+        sources: &SourceMap,
+        abstract_type: &str,
+    ) -> Option<Range<Self>> {
+        let node_location = node.location()?;
+        let source = sources.get(&node_location.file_id())?;
+        let start = source
+            .get_line_column(node_location.offset())
+            .map(|(line, column)| Location { line, column })?;
+        let end = source
+            .get_line_column(node_location.offset() + abstract_type.chars().count())
```
dylan-apollo, line 612: This is clever! Seems like the "right" thing to do is to add something to `apollo-rs` that gives us the location of the keyword, but that'll take a while. This seems like a good workaround in the meantime.

The only risk, I think, is that `apollo-rs` changes the way they do locations, but then we should see our snapshots change/fail.

apollo-federation/src/sources/connect/validation/test_data/validation/disallowed_abstract_types.graphql
dylan-apollo: I think this is fine, since they're basically the same check.

apollo-federation/src/sources/connect/validation/mod.rs
```diff
@@ -581,6 +597,23 @@ impl Location {
             .map(|(line, column)| Location { line, column })?;
         Some(Range { start, end })
     }
+
+    fn from_abstract_type_node<T>(
```
dylan-apollo, line 601: You could turn this into a top-level `unsupported_abstract_type` function and return the `Message` from up above. Then that'll unify how the error messages are formatted, too.

apollo-federation/src/sources/connect/validation/mod.rs
tayrrible: I made an update to pull this into a new function but def would like some input so we can get it back to `flat_map`

apollo-federation/src/sources/connect/validation/extended_type.rs
```diff
@@ -0,0 +1,283 @@
...
+        ExtendedType::Interface(interface) => connect_errors.push(validate_abstract_type(
+            NodeLocation::recompose(interface.location(), interface.name.location()),
+            source_map,
+            "interface",
+        )),
+        _ => (),
+    }
+
+    connect_errors
+}
```
dylan-apollo, line 66:
```suggestion
   match extended_type {
        ExtendedType::Object(object) => validate_object_fields(
            object,
            schema,
            connect_directive_name,
            source_directive_name,
            all_source_names,
        ),
        ExtendedType::Union(union_type) => vec![validate_abstract_type(
            NodeLocation::recompose(union_type.location(), union_type.name.location()),
            source_map,
            "union",
        )],
        ExtendedType::Interface(interface) => vec![validate_abstract_type(
            NodeLocation::recompose(interface.location(), interface.name.location()),
            source_map,
            "interface",
        )],
        _ => Vec::new(),
    }
```

dylan-apollo (approved):
LGTM!

PR #5679, opened 2024-07-16
Update entity-caching.mdx

mwalker5000:
Updating the page title, moving from 'subgraph entity caching' to 'entity caching'. This aligns with the official feature name (entity caching) and is an attempt to reduce confusion (e.g. this title might lead to the misunderstanding that an entity might somehow be able to exist within a subgraph).

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/docs/source/configuration/entity-caching.mdx
+++ b/docs/source/configuration/entity-caching.mdx
@@ -1,5 +1,5 @@
 ---
-title: Subgraph Entity Caching for the Apollo Router
+title: Entity Caching for the Apollo Router
 subtitle: Configure Redis-backed caching for entities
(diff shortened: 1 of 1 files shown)


BrynCooke (dismissed):

BrynCooke:
Personally think this may be the way to go, as we may get entities from places other than subgraphs going forwards.

abernix:
I'll close this, as this is not the current direction.  If entities do come from different places in the future, we can choose a different name, but for now the name accurately specifies the boundary at which these cache decisions are made, and more closely — though not perfectly — the extent of their re-use.

PR #5682, opened 2024-07-17
Switch connectors fully to `headers.from`,

dylan-apollo:
Removes `headers.as` in favor of the new `headers.from`. In the process, I made a lot of code more consistent:

1. Duplicate values are allowed (as they are with HTTP requests). The implementation already supported this, now validation & parsing do too. We were also already _not_ catching a conflict between `@source` and `@connect`. We may want to add a feature to `@connect` to specify that a header should be overridden vs appended, but that can be a new parameter (I think).
2. Got rid of the shorthand `{name: "blah"}` syntax in favor of requiring an explicit `from` or `value`. We can always add back the shorthand later if we think it's needed, but we can't go the other way.
3. The last two changes make it possible to unify `HTTPHeader` with `HTTPHeaderOption` and use some slightly-simpler `Vec` instead of `HashMap`.

I think we should also unify the validator's parsing of headers with the runtime, but that will take some extra care to not slow down the runtime with extra diagnostic info. I figure that should be a separate PR since this one is already quite large.

Note that _using_ this requires a new version of `supergraph`, but that new version of `supergraph` will depend on this code. It can be tested locally for now using local dependencies until we can release all the pieces.

## TODO

- [x] Put back duplicate header validation
- [x] Implement `@connect` headers overriding `@source` headers
- [x] Make duplicate header detection case-insensitive

[CNN-288]: https://apollographql.atlassian.net/browse/CNN-288?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lennyburdette:
thanks for tackling this!

regarding duplicate names/multiple headers with the same name/source-connector conflicts: i liked [Shane's suggestion](https://docs.google.com/document/d/13-XyFxDjGnLoxPJceuWk-RZegkumtWEkC0bMxPeWPnw/edit?disco=AAABGDV1LTk). the idea is:

1. you can't use a header name multiple times.
2. if you want to provide multiple values for a header, you can use the `value: ["one", "two"]` syntax (allowed because value is actually a list).
3. if the header name in `@connect` conflicts with a header name in `@source`, the connect one always wins.

lennyburdette:
> 1. Combining two from headers into a list
> 2. Adding value(s) to an incoming from

Those are interesting examples, but at that point I wonder if it's better to use a coprocessor/rhai script to construct the header(s)

> > if the header name in @connect conflicts with a header name in @source, the connect one always wins.
> 
> This seems logical, should I fix the implementation to do that here or in a separate PR?

I think I'd just tackle that while you're here, but I'm fine with a separate PR too.

> Also, worth noting I think, that (as far as I can tell), the spec defines multiple header values as being comma separated. So technically, we could also remove the ability to set an array in value and if users want multiple values they can define a comma-separated string.

Which spec are you referring to?

dylan-apollo:
> Which spec are you referring to?

[RFC 2616](https://www.rfc-editor.org/rfc/rfc2616#section-4.2), see the last paragraph of that section

lennyburdette:
Hmm, so i think that's saying that 
```
x-foo: bar
x-foo: baz
```
must be equivalent to
```
x-foo: bar,baz
```

which make sense but _i think_ that's orthogonal to how values are defined in our API. but maybe i'm missing something?

dylan-apollo:
@lennyburdette what I mean is we don't need to offer an array of values as part of our API, in the interest of keeping things minimal. That will simplify the implementation, I think, as well (now that we're not accepting values from multiple sources).

lennyburdette (approved):

PR #5683, opened 2024-07-17
Support `@connect` with no source

dylan-apollo:
Supports `@connect` with absolute URL by putting the base URL right in the `URLPathTemplate` (now called `URLTemplate`). We'll need to support expansion of params in the baseURL eventually (when #5646 lands) so that seemed like the best place for it.

The rest of the implementation doesn't care whether the `baseURL` came from `@source` or `@connect`. We may want to put back some diagnostics in tracing if we think the distinction is important there.

One thing I'm not sure on is if we still need the `append_path` logic. If so, we should add some tests that fail without it. It _looks_ like the intent is to allow global query parameters via a `@source(baseURL:)`, but I'm not sure if that's (still) a feature we're supporting.

Composing new graphs for this will require a new release of `supergraph`.

[CNN-201]: https://apollographql.atlassian.net/browse/CNN-201?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apollo-router/src/plugins/connectors/tests.rs
```diff
@@ -555,10 +556,47 @@ async fn test_nullability() {
...
+async fn test_no_source() {
+    // TODO: make this port dynamic once overriding via config is supported
+    let mock_server = MockServer::builder()
+        .listener(TcpListener::bind("127.0.0.1:37895").expect("Could not bind port"))
+        .start()
+        .await;
+    mock_api::user_1().mount(&mock_server).await;
+
+    let response = execute(
+        NO_SOURCES_SCHEMA,
```
lennyburdette, line 569: you could do `NO_SOURCES_SCHEMA.replace("http://localhost/", mock_server.uri()` as a cheap way to avoid hardcoding the port number

lennyburdette (approved):

PR #5684, opened 2024-07-18
Improves semantic comparison of query plans

duckki:
This PR improves semantic comparison of query plans as following:

- Selection sets are no longer order-sensitive. Selections are sorted by the selection key before comparison.
- Fragment definitions are no longer order-sensitive. Fragment definitions are sorted by the name before comparison.

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [x] Manual Tests

[FED-318]: https://apollographql.atlassian.net/browse/FED-318?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apollo-router/src/query_planner/dual_query_planner.rs
+++ b/apollo-router/src/query_planner/dual_query_planner.rs
@@ -511,24 +511,19 @@ fn hash_selection_key(selection: &ast::Selection) -> u64 {
 }
 
 fn same_ast_selection_set_sorted(x: &[ast::Selection], y: &[ast::Selection]) -> bool {
-    fn sorted_by_selection_key(s: &[ast::Selection]) -> Vec<ast::Selection> {
-        let mut sorted = s.to_owned();
-        sorted.sort_by_key(hash_selection_key);
+    fn sorted_by_selection_key(s: &[ast::Selection]) -> Vec<&ast::Selection> {
+        let mut sorted: Vec<&ast::Selection> = s.iter().collect();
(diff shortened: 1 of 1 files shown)


apollo-router/src/query_planner/dual_query_planner.rs
```diff
@@ -385,26 +372,46 @@ fn flatten_node_matches(this: &FlattenNode, other: &FlattenNode) -> bool {
...
+    assert!(
+        x_ops.len() == y_ops.len(),
+        "Different number of operation definitions"
+    );
+
+    x_ops
+        .iter()
+        .zip(y_ops.iter())
+        .all(|(x_op, y_op)| same_ast_operation_definition(x_op, y_op))
+        && x_frags.len() == y_frags.len()
```
TylerBloom, line 410: Is there a reason that we're panicking if `x_ops.len()` doesn't equal `y_ops.len()` but not for the fragments?

duckki: The number of operations shouldn't really different at all (just one op all the time). So, I thought it was assertion-worthy.

duckki: I'm rethinking this. It's running in production. It shouldn't be using assert. I'll change that.

TylerBloom (approved):

PR #5685, opened 2024-07-18
Apollo Federation Logging

TylerBloom:
This PR includes an initial pass in how logging will be approached in `apollo-federation`. There are two sets of changes:
 - Porting debug logs from JS
 - Added snapshot logging for visualization

Snapshot logs are feature gated (behind the new `snapshot_tracing` feature) as they contain a lot of information that is not desired in production and is largely designed for debugging. Also, the snapshot logging serializes objects, so deriving `Serialize` was added to many types.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-federation/src/operation/mod.rs
```diff
@@ -603,24 +606,33 @@ pub(crate) use selection_map::SelectionValue;
...
 /// * directive arguments order does not matter (they get automatically sorted by their names).
 /// * selection cannot specify @defer directive
-#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize)]
 pub(crate) enum SelectionKey {
     Field {
         /// The field alias (if specified) or field name in the resulting selection set.
         response_name: Name,
         /// directives applied on the field
+        // TODO(@TylerBloom): Add back in once apollo-compiler supports serde or a workaround is
```
dariuszkuc, line 615: can you open a follow up JIRA to track this?

goto-bus-stop: Suggest using a `serde(serialize_with)` that just uses `directives.to_string()`?

dariuszkuc (approved):

apollo-federation/src/operation/mod.rs
TylerBloom: > Suggest using a `serde(serialize_with)` that just uses `directives.to_string()`?

That's a good idea. I'll add replace the relevant `serde(skip)`s will that.

TylerBloom: Here's the Jira ticket for the other `TODO`s.
https://apollographql.atlassian.net/browse/FED-363

apollo-federation/src/operation/mod.rs
```diff
@@ -66,7 +67,7 @@ static NEXT_ID: atomic::AtomicUsize = atomic::AtomicUsize::new(1);
 ///
 /// Note that we shouldn't add `derive(Serialize, Deserialize)` to this without changing the types
 /// to be something like UUIDs.
-#[derive(Clone, Debug, Eq, PartialEq, Hash)]
+#[derive(Clone, Debug, Eq, PartialEq, Hash, Serialize)]
```
abernix, line 70: @TylerBloom @lennyburdette @dariuszkuc Does the comment above this need to be updated after this change?  (The comment says not to add `Serialize` here, but adds `Serialize`? 😄

TylerBloom: Ah, I definitely should have made a comment about this during review. That's my bad. My thought was that serialize was ok but deserialize was not (and both was definitely not a good idea); however, this should probably be feature-gated like the `snapshot!` macro that was added. I need to do some clean up around the `Serialize` implementation of a few types. I can bundle this change into that PR.

abernix: Thank you! that'd be wonderful. 🙇  Consider my curiousity placated. 😄

TylerBloom: Addressed in this PR: https://github.com/apollographql/router/pull/5724

PR #5688, opened 2024-07-19
fix(federation): fix "$var is not defined" in subgraph entity query

goto-bus-stop:
We had a bug where an input like this:
```graphql
fragment x on T {
  __typename @include(if: $var)
  field
}
query ($var: Boolean!) {
  returnsT {
    ...x
  }
  someEntity {
    .. on T { field }
  }
}
```

Could produce a subgraph entity query like this. Note this isn't a precise reproduction, since the exact reproduction involving entities is more difficult, but it is an accurate presentation of the problem:
```graphql
query Example($representations: [_Any!]!) {
  _entities(representations: $representations) {
    ... on T {
      __typename @include(if: $var)
      field
    }
  }
}
```

The entity query references a `$var` that does not exist.

This is because when the final subgraph operation is produced, we first determine what variables the operation needs, and *then* try to reuse fragments from the input. Inline fragments may be replaced by an equivalent fragment spread. Whether a `__typename` field is present does not matter for equivalence. So fragment reuse considers that `... on T { field }` may be reduced to `...x`. Then, it considers that `x` is only used once in the final operation, so using `x` actually requires more code, and re-expands it. We get:
```graphql
... on T {
  __typename @include(if: $var)
  field
}
```
And we're referencing `$var` even though the `someEntity` part of the query never referenced a variable originally.

So. The bug is that `.contains()` has to ignore `__typename`, but *only* if it does not have directives. It's possible that there are other issues when you have a `__typename` with directives. But the JS only considers `__typename` *without directives* specially, and Rust was considering *any* `__typename` specially. This patch aligns JS and Rust.

An alternative solution might have been to collect the used variables *after* reusing fragments. But I think fragment reuse should never introduce new requirements on the operation, so this is a better fix.

The first 3 commits are for debugging help and cleanup (see commit messages). The remaning commits are part of the fix.

### DIFF
--- a/apollo-federation/cli/src/main.rs
+++ b/apollo-federation/cli/src/main.rs
@@ -170,7 +170,7 @@ fn plan(query_path: &Path, schema_paths: &[PathBuf]) -> Result<(), FederationErr
     // TODO: add CLI parameters for config as needed
     let config = QueryPlannerConfig::default();
     let planner = QueryPlanner::new(&supergraph, config)?;
-    print!("{}", planner.build_query_plan(&query_doc, None).unwrap());
+    print!("{}", planner.build_query_plan(&query_doc, None)?);
     Ok(())
 }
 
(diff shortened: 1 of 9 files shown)


lrlna (approved):
A rollercoaster of a read 👀 !

PR #5700, opened 2024-07-22
fix(federation): collect variables nested inside lists and input objects

goto-bus-stop:
When variables are inserted inside a list or an input object, we were not propagating the variable definition onto subgraph queries, so we'd generate invalid output. The variable reference would still exist inside the list or input object but the definition was missing.

This patch recurses into lists and input objects when collecting variable usages.


lrlna (approved):
🎉

PR #5609, opened 2024-07-04
Datadog exporter improvements

BrynCooke:
This PR updates the Datadog exporter with a backport from the latest code. 
The exporter code has been placed in the `apollo-router/src/plugins/telemetry/tracing/datadog_exporter` directory.

It also fixes:
* `span.kind` not being propagated
* Missing span stats. (Users may override configuration for these).

Span metrics are enabled for major built in spans by default.

```yaml
telemetry:
  exporters:
    tracing:
      experimental_response_trace_id:
        enabled: true
        header_name: apollo-custom-trace-id
        format: datadog
      common:
        service_name: router
      datadog:
        enabled: true
        batch_processor:
          scheduled_delay: 100ms
        span_metrics: # New config option to enable or disable span metrics. This is a free format span to boolean map.
          supergraph: false

  instrumentation:
    spans:
      mode: spec_compliant
      supergraph:
        attributes:
          graphql.operation.name: true
```

---

**Checklist**

Complete the checklist (and note appropriate exceptions) before the PR is marked ready-for-review.

- [x] Changes are compatible[^1]
- [x] Documentation[^2] completed
- [ ] Performance impact assessed and acceptable
- Tests added and passing[^3]
    - [ ] Unit Tests
    - [x] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[^1]: It may be appropriate to bring upcoming changes to the attention of other (impacted) groups. Please endeavour to do this before seeking PR approval. The mechanism for doing this will vary considerably, so use your judgement as to how and when to do this.
[^2]: Configuration is an important part of many changes. Where applicable please try to document configuration examples.
[^3]: Tick whichever testing boxes are applicable. If you are adding Manual Tests, please document the manual testing (extensively) in the Exceptions.

[ROUTER-363]: https://apollographql.atlassian.net/browse/ROUTER-363?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #5703, opened 2024-07-22
Datadog sampling priority not set

BrynCooke:
In #5609 The backported code from the datadog exporter used code from the feature `agent-sampling`.

This was an incorrect decision as this requires that the sampler is updated to support such functionality. The result is that traces are not being shown in the Datadog UI even though they were exported.

This PR switches the code with `not(agent-samping)` code which will delegate sampling to the sampler in the router yaml config.

Relates to #5609

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [x] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/tests/integration/telemetry/datadog.rs
+++ b/apollo-router/tests/integration/telemetry/datadog.rs
@@ -427,6 +427,7 @@ impl TraceSpec {
         tracing::debug!("{}", serde_json::to_string_pretty(&trace)?);
         self.verify_trace_participants(&trace)?;
         self.verify_operation_name(&trace)?;
+        self.verify_priority_sampled(&trace)?;
         self.verify_version(&trace)?;
         self.verify_spans_present(&trace)?;
         self.validate_span_kinds(&trace)?;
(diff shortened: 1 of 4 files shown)


apollo-router/tests/integration/telemetry/datadog.rs
```diff
@@ -576,4 +577,17 @@ impl TraceSpec {
...
+
+    fn verify_priority_sampled(&self, trace: &Value) -> Result<(), BoxError> {
+        let binding = trace.select_path("$.._sampling_priority_v1")?;
+        let sampling_priority = binding.first();
+        assert_eq!(
+            sampling_priority
+                .expect("sampling priority expected")
+                .as_f64()
+                .expect("sampling priority must be a number"),
+            1.0
```
BrynCooke, line 589: This was 0.0 but is now 1.0!

bnjjj (approved):

PR #5704, opened 2024-07-22
Remove custom `Location` type in favor of `apollo_compiler::LineColumn`

dylan-apollo:
Also removes a mod/function that aren't used (found while walking through all the code).


lennyburdette:
was there an off-by-one issue with the previous implementation?

dylan-apollo:
@lennyburdette our custom type was exposing zero-indexed locations. The `apollo_compiler` type is 1-indexed. I'm going to update all the consumers to make sure they're converting from the new type properly.

Rover/fbp use 1-indexed. LSP uses 0-indexed.

lennyburdette (approved):

PR #5709, opened 2024-07-23
feat(qp): enable running both query planners by default

lrlna:
This change is enabling running the Native query planner in a background thread in order to compare output and performance differences to the existing JS query planner. Execution continues to be based on query plans produced by the JS query planner, as Native query planner results are discarded after the comparison.

This option has been run in various different environments without an impact to memory or CPU utilisation.

You can go back to the previous behaviour of running only the JS query planner with the following config in your router yaml:

```yml
experimental_query_planner_mode: legacy
```

[ROUTER-458]: https://apollographql.atlassian.net/browse/ROUTER-458?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changesets/feat_lrlna_enable_both_query_planners_by_default.md
+++ b/.changesets/feat_lrlna_enable_both_query_planners_by_default.md
@@ -1,8 +1,7 @@
 ### Enable running both query planners by default ([PR #5709](https://github.com/apollographql/router/pull/5709))
 
-<!-- [ROUTER-458] -->
 
-This change is enabling running the Native query planner in a background thread
+This change enables running the Native query planner in a background thread
 in order to compare output and performance differences to the existing JS query
 planner. Execution continues to be based on query plans produced by the JS query
 planner, as Native query planner results are discarded after the comparison.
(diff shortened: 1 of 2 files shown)


.changesets/feat_lrlna_enable_both_query_planners_by_default.md
```diff
@@ -0,0 +1,19 @@
+### Enable running both query planners by default ([PR #5709](https://github.com/apollographql/router/pull/5709))
+
+<!-- [ROUTER-458] -->
```
garypen, line 3:
```suggestion
```

.changesets/feat_lrlna_enable_both_query_planners_by_default.md
```diff
@@ -0,0 +1,19 @@
...
+
+This change is enabling running the Native query planner in a background thread
```
garypen, line 5:
```suggestion
This change enables running the Native query planner in a background thread
```

garypen (approved):

abernix:
I'll close this for now.

PR #5720, opened 2024-07-24
join__directive support in extract_subgraphs_from_supergraph

lennyburdette:
and some federation constants

this change is mostly useless at the moment because the yet-to-be-implemented connector directives are the only ones eligible for join__directive serialization

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-federation/src/query_graph/extract_subgraphs_from_supergraph.rs
```diff
@@ -2094,10 +2104,181 @@ fn maybe_dump_subgraph_schema(subgraph: FederationSubgraph, message: &mut String
...
+fn join_directive_to_real_directive(directive: &Node<Directive>) -> (Directive, Vec<Name>) {
+    let subgraph_enum_values = directive
+        .argument_by_name("graphs")
+        .and_then(|arg| arg.as_list())
+        .map(|list| {
+            list.iter()
+                .map(|node| {
+                    Name::new(
+                        node.as_enum()
+                            .expect("join__directive(graphs:) value is not an enum")
```
lrlna, line 2238: the message of `.expect()` should be explaining why the Result/Option of the value should have been there ([guidance from rust docs](https://doc.rust-lang.org/std/result/enum.Result.html#recommended-message-style)). All of these are worded in an `.assert()` style when it's not actually an assert. These should read something like "join__directive(graphs:) value is an enum" or something of the sort. There are a bunch of good examples in `apollo-router/src/json_ext.rs` with expect statements.

apollo-federation/src/query_graph/extract_subgraphs_from_supergraph.rs
```diff
@@ -2094,10 +2104,181 @@ fn maybe_dump_subgraph_schema(subgraph: FederationSubgraph, message: &mut String
...
+                    graph_enum_value_name_to_subgraph_name,
+                    &subgraph_enum_value,
+                )?;
+
+                schema_def_pos.insert_directive(
+                    &mut subgraph.schema,
+                    Component::new(link_directive.clone()),
+                )?;
+
+                // TODO: add imported definitions from relevant specs
```
lrlna, line 2165: can you add the todos to jira as well? we lose track of these so easily

lrlna (approved):
A small comment on `.expect()` style. Otherwise this should be good to go!

apollo-federation/src/query_graph/extract_subgraphs_from_supergraph.rs
lennyburdette: https://apollographql.atlassian.net/browse/FED-373

PR #5724, opened 2024-07-24
Fixed serialization panic in snapshot macro

TylerBloom:
The `snapshot!` macro was serializing values to JSON. This caused issues for graphs and maps with nodes and keys (respectively) that could not be readily serialized into strings. The `derive` impl of `Serialize` was also feature gated for a few types that we do not yet to generally be serializable.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-federation/Cargo.toml
+++ b/apollo-federation/Cargo.toml
@@ -14,7 +14,7 @@ autotests = false                                      # Integration tests are m
 # logging statements that capture serialized versions of key data structures.
 # This logging is gated behind a feature to avoid any unnecessary (even if
 # small) runtime costs where this data will not be desired.
-snapshot_tracing = []
+snapshot_tracing = ["ron"]
 
 [dependencies]
 apollo-compiler.workspace = true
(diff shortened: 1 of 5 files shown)


apollo-federation/Cargo.toml
```diff
@@ -36,6 +36,7 @@ strum_macros = "0.26.0"
 thiserror = "1.0"
 url = "2"
 tracing = "0.1.40"
+ron = { version = "0.8.1", optional = true }
```
dariuszkuc, line 39: should this be dev dependency instead?

lrlna (approved):

apollo-federation/Cargo.toml
TylerBloom: No. This is not used for tests. The visualizer uses this for snapshots. If you want to get snapshot logs while testings, you can just enable that feature when you run `cargo test`.

PR #5725, opened 2024-07-25
Evaluate selectors in response stage when possible

bnjjj:
As `events` are triggered at a specific stage (`request`|`response`|`error`) we can only have condition for the related stage and so sometimes you can also have selectors that can be applied at several stages (like `subgraph_name` to get the subgraph name). This change adds more possibilities when creating conditions on events.

Example of an event to log the raw subgraph response only on a subgraph named `products`, this was not working before.

```yaml
telemetry:
  instrumentation:
    events:
      subgraph:
        response:
          level: info
          condition:
            eq:
            - subgraph_name: true
            - "products"
```

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [x] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/.changesets/fix_bnjjj_fix_subgraph_selector_for_event.md
+++ b/.changesets/fix_bnjjj_fix_subgraph_selector_for_event.md
@@ -1,0 +1,20 @@
+### Evaluate selectors in response stage when possible ([PR #5725](https://github.com/apollographql/router/pull/5725))
+
+As `events` are triggered at a specific stage (`request`|`response`|`error`) we can only have condition for the related stage and so sometimes you can also have selectors that can be applied at several stages (like `subgraph_name` to get the subgraph name). This change adds more possibilities when creating conditions on events.
+
+Example of an event to log the raw subgraph response only on a subgraph named `products`, this was not working before.
+
+```yaml
+telemetry:
+  instrumentation:
+    events:
+      subgraph:
+        response:
+          level: info
+          condition:
+            eq:
+            - subgraph_name: true
+            - "products"
+```
+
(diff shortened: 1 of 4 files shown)


.changesets/fix_bnjjj_fix_subgraph_selector_for_event.md
```diff
@@ -0,0 +1,20 @@
+### Evaluate selectors in response stage when possible ([PR #5725](https://github.com/apollographql/router/pull/5725))
+
+As `events` are triggered at a specific stage (`request`|`response`|`error`) we can only have condition for the related stage and so sometimes you can also have selectors that can be applied at several stages (like `subgraph_name` to get the subgraph name). This change adds more possibilities when creating conditions on events.
```
BrynCooke, line 3:
```suggestion
As `events` are triggered at a specific event (`request`|`response`|`error`) we can only have condition for the related event, but sometimes selectors that can be applied at several events (like `subgraph_name` to get the subgraph name). Adds support for various supergraph selectors on response events.
```

apollo-router/src/plugins/telemetry/config_new/selectors.rs
```diff
@@ -1357,6 +1357,39 @@ impl Selector for SubgraphSelector {
                     .canonical_reason()
                     .map(|reason| reason.into()),
             },
+            SubgraphSelector::SubgraphOperationKind { .. } => response
```
BrynCooke, line 1360: I'm pretty sure that this file has unit tests for all these selectors, so we should add to them passing in the response event.

BrynCooke (changes requested):

PR #5727, opened 2024-07-25
chore(qp): add a diff_plan function to display differences between js & rust qps

lrlna:
This specifically compares JS' `QueryPlanResult` and Rust's `QueryPlan` structs, i.e. the ones we use to semantic diff when looking for mismatches. The rendered diff uses `diff::lines` under the hood.

### DIFF
--- a/apollo-router/src/lib.rs
+++ b/apollo-router/src/lib.rs
@@ -114,7 +114,6 @@ pub mod _private {
     pub use crate::plugin::PluginFactory;
     pub use crate::plugin::PLUGINS;
     // For comparison/fuzzing
-    pub use crate::query_planner::bridge_query_planner::render_diff;
     pub use crate::query_planner::bridge_query_planner::QueryPlanResult;
     pub use crate::query_planner::dual_query_planner::diff_plan;
     pub use crate::query_planner::dual_query_planner::plan_matches;
--- a/apollo-router/src/lib.rs
+++ b/apollo-router/src/lib.rs
@@ -114,7 +114,9 @@ pub mod _private {
     pub use crate::plugin::PluginFactory;
     pub use crate::plugin::PLUGINS;
     // For comparison/fuzzing
+    pub use crate::query_planner::bridge_query_planner::render_diff;
     pub use crate::query_planner::bridge_query_planner::QueryPlanResult;
+    pub use crate::query_planner::dual_query_planner::diff_plan;
     pub use crate::query_planner::dual_query_planner::plan_matches;
     // For tests
     pub use crate::router_factory::create_test_service_factory_from_yaml;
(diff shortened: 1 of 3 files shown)


apollo-router/src/lib.rs
```diff
@@ -115,6 +115,7 @@ pub mod _private {
     pub use crate::plugin::PLUGINS;
     // For comparison/fuzzing
     pub use crate::query_planner::bridge_query_planner::QueryPlanResult;
+    pub use crate::query_planner::dual_query_planner::diff_plan;
```
lrlna, line 118: exporting this under the `_private`, so we can use it in other tooling.

apollo-router/src/query_planner/dual_query_planner.rs
```diff
@@ -143,12 +143,10 @@ impl BothModeComparisonJob {
...
-                    if let Some(formatted) = &js_plan.formatted_query_plan {
-                        tracing::debug!(
-                            "Diff of formatted plans:\n{}",
-                            render_diff(&diff::lines(formatted, &rust_plan.to_string()))
-                        );
-                    }
+                    tracing::debug!(
+                        "Diff of formatted plans:\n{}",
+                        diff_plan(&js_plan, &rust_plan)
+                    );
```
lrlna, line 149: The rendered diff wouldn't be as accurate when comparing formatted plans, since both JS and Rust don't show all the information that could have diverged.

apollo-router/src/query_planner/dual_query_planner.rs
```diff
@@ -242,6 +240,31 @@ pub fn plan_matches(js_plan: &QueryPlanResult, rust_plan: &QueryPlan) -> bool {
     opt_plan_node_matches(js_root_node, &rust_root_node)
 }
 
+pub fn diff_plan(js_plan: &QueryPlanResult, rust_plan: &QueryPlan) -> String {
+    let js_root_node = &js_plan.query_plan.node;
+    let rust_root_node = convert_root_query_plan_node(rust_plan);
```
lrlna, line 245: Both of these parameters _can_ be `Option<PlanNode>`, but that would require additionally exporting `convert_root_query_plan_node` as part of `_private`, and I want to limit that amount of things we export, even though they are under `_private`.

apollo-router/src/query_planner/dual_query_planner.rs
duckki: It's good enough for now. We can improve the diff output as we go.

Clippy suggestion:
```suggestion
                    tracing::debug!(
                        "Diff of formatted plans:\n{}",
                        diff_plan(js_plan, rust_plan)
                    );
```

duckki (approved):
Looks good!

PR #5730, opened 2024-07-26
fix(federation): prevent reuse_fragments() from introducing new variable references

goto-bus-stop:
There is a special case where `reuse_fragments()` can introduce a new variable reference, that is not part of the operation definition it is called on.

If you have a fragment `B` that is used inside a different fragment `A`, where `B`'s runtime types do not intersect with the parent type where fragment `A` is used, `B` is essentially a dead branch. It will never be executed. Because most of the federation code works on expanded selection sets, the dead branch is not present at all. However, when we try to reuse fragments, the dead branch may be re-added. Fragment reuse checks also work on expanded selection sets, so it's as if the fragment `B`'s selections do not exist while we consider if fragment `A` should be used. That means fragment `A` may be used, introducing the dead branch `B`. If `B` contains a variable reference, we have just produced an invalid query. It's a bit tricky to follow in text but the test should clarify.

Ideally, we would not re-add any dead branches at all. I tried a few approaches to do that, and it is definitely possible, but I think it's more work than we should give to fragment reuse, as we hope to replace it with a better strategy anyways. This patch simply ignores any fragments that use variables that aren't declared in the operation.

To do that i have to be able to collect all used variables from a fragment, including through named fragment spreads. It turns out that we can actually do that very easily because a `FragmentSpreadSelection` records the selection set of its named fragment. `used_variables()` does not have to return an error anymore so we can simplify that a bit as well.

### DIFF
--- a/apollo-federation/src/operation/mod.rs
+++ b/apollo-federation/src/operation/mod.rs
@@ -4028,17 +4028,17 @@ fn collect_variables_from_directive<'selection>(
     variables: &mut HashSet<&'selection Name>,
 ) {
     for arg in directive.arguments.iter() {
-        collect_variables_from_value(&arg.value, variables)
+        collect_variables_from_value(&arg.value, variables);
     }
 }
 
 impl Field {
     fn collect_variables<'selection>(&'selection self, variables: &mut HashSet<&'selection Name>) {
         for arg in self.arguments.iter() {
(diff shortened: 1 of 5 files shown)


TylerBloom (approved):

duckki (approved):
👍

PR #5734, opened 2024-07-26
Standardize URL creation

dylan-apollo:
User-facing changes:

1. All URL path components & query parameters defined in a `@connect` template will now be percent-encoded. This is done very conservatively to start with the most secure solution.
2. If an invalid base URL somehow makes it through composition (e.g., building supergraph schemas by hand or a bug in those validations), the error should now occur at router startup, not on query execution

Internal changes:

1. Moves all the URL-generation code into the same place (in `apollo-router` instead of `apollo-federation`)
2. Relies much more heavily on the `url` crate instead of custom generation code.
3. Switch `URLTemplate` to the standard `FromStr` trait instead of the custom `parse` method
4. Pass around more `Url` structs to avoid re-parsing at runtime

Note that most of the test changes are moving things from `url_template` to `http_json_transport` (where all of the URL generation now happens).


apollo-router/src/plugins/connectors/configuration.rs
```diff
@@ -64,7 +64,7 @@ pub(crate) fn apply_config(config: &Configuration, mut connectors: Connectors) -
         if let Some(url) =
             source_config.and_then(|source_config| source_config.override_url.as_ref())
         {
-            connector.transport.source_url = Some(url.to_string());
+            connector.transport.source_url = Some(url.clone());
```
dylan-apollo, line 67: This could be pretty easily `Arc`d now I think, if we want

apollo-router/src/plugins/connectors/http_json_transport.rs
```diff
@@ -114,86 +117,36 @@ pub(crate) fn make_request(
...
-            &generated,
-        )
-    } else {
-        Url::parse(&generated).map_err(HttpJsonTransportError::InvalidUrl)
-    }
-}
+    let mut url = source_url
+        .or(template.base.as_ref())
+        .ok_or(HttpJsonTransportError::NoBaseUrl)?
+        .clone();
```
dylan-apollo, line 128: In theory we only have to `.clone()` if we're going to add path &| query params, but I figure that's the vast majority of the time

apollo-router/src/plugins/connectors/http_json_transport.rs
```diff
@@ -272,89 +223,496 @@ pub(crate) enum HttpJsonTransportError {
...
+                    "p1.z" => 3,
+                    "p2.x" => 4,
+                    "p2.y" => 5,
+                    "p2.z" => 6
+                }
+            )
+            .err()
+            .unwrap()
+            .to_string(),
+            r#"Could not generate path from inputs: Missing required variable p1.y in {"p1.x":1,"p1.z":3,"p2.x":4,"p2.y":5,"p2.z":6}"#
```
lennyburdette, line 681: i'm wondering if this is a potential security issue — if these error messages are exposed, they could reveal information that should be kept internal to query planning.

at first i was thinking this won't be a problem, because we should be validating `$args` and `$this` to ensure they don't select things that won't be provided. but because we have support for `$config`, which we can't validate now, it's pretty easy to get this message in a client response:
```
  "errors": [
    {
      "message": "TransportError(TemplateGenerationError(\"Missing required variable $config.x in {\\\"$this.__typename\\\":\\\"Track\\\",\\\"$this.id\\\":\\\"c_0\\\"}\"))",
      "extensions": {
        "code": "FETCH_SERVICE"
      }
    }
  ],
```

lennyburdette (approved):

apollo-router/src/plugins/connectors/http_json_transport.rs
dylan-apollo: Maybe we should get rid of the context, then and just specify the name of the missing variable? I assume you can only see these errors in dev mode, right? Otherwise it'd be easy to leak secrets by showing the current context.

lennyburdette (approved):
i pulled this into my test branch and everything is working, though my examples don't do anything weird with url components

apollo-router/src/plugins/connectors/http_json_transport.rs
dylan-apollo: I went ahead and removed it for now just in case. We can always add back more details if folks struggle to debug this.

lennyburdette: the errors become top-level graphql errors right now, so i think removing this data is the right move

PR #5746, opened 2024-07-30
add more attributes and include invalidation api endpoint as a root trace to be sampled

bnjjj:


BrynCooke (changes requested):
No test has changed in this PR.
As there are existing integration tests for cache invalidation I think some minor modification to the yaml test harness may get us there.

bnjjj:
There is no tracing test for invalidation cache ATM

Geal (approved):

BrynCooke:
@bnjjj What do we want to do with this? I think we should add some tests if we don't have any for cache invalidation.

BrynCooke (approved):

PR #5752, opened 2024-07-30
Remove tracing-fluent-assertions dependency

pubmodmatt:
There was feedback in #5108 about adding a dependency on [tracing-fluent-assertions](https://crates.io/crates/tracing-fluent-assertions) when there are already other dependencies in use to test tracing. Instead, use the existing `test-snap` dependency to create a snapshot of the connectors span.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

**Notes**

[CNN-367]: https://apollographql.atlassian.net/browse/CNN-367?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -628,7 +628,6 @@ dependencies = [
  "tower-test",
  "tracing",
  "tracing-core",
- "tracing-fluent-assertions",
  "tracing-futures",
  "tracing-opentelemetry",
  "tracing-serde",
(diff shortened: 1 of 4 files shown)


dylan-apollo (commented):
I think the goal of trying to keep down dev dependencies is for compilation speed—it's unclear to me if this additional code would have a positive impact compared to the one dependency. It certainly results in harder-to-read code.

However, I figure it's worth mentioning that this test, when it was using `tracing-fluent-assertions`, would occasionally fail for me locally if I ran only this module (as I often do). So if we _don't_ swap to mock, we should still investigate why it's flakey.

dylan-apollo (approved):

PR #5743, opened 2024-07-30
fix(federation): pass on operation directives to subgraph queries

goto-bus-stop:
Requires https://github.com/apollographql/router/pull/5753

This propagates the input operation's `DirectiveList` down to subgraph queries.

When the subgraph operations are built, we don't have access to the input operation. We have access to its selection set and directives list separately. To make sure we propagate variable definitions that are used inside the directive list of the operation, I adjusted the API of `used_variables()`, with all the visitor code in a `VariableCollector` struct that you can use to collect variables from multiple elements. This change also fixes a very, very minor hole in #5730, namely if there is a variable reference in a directive application on a named fragment definition.

PR #5753, opened 2024-07-31
deps: update apollo-compiler

goto-bus-stop:
`query ($var: Int) @directive(arg: $var)` is now accepted - it used to raise an unused variable error for `$var`.

I also moved some remaining apollo-rs dependency version declarations to use the `.workspace = true` syntax, to remove multiple versions of the same crate from the dependency tree.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [x] Integration Tests (federation snapshot test proves the fix)
    - [ ] Manual Tests

### DIFF
--- a/.changesets/fix_renee_operation_variables.md
+++ b/.changesets/fix_renee_operation_variables.md
@@ -1,0 +1,11 @@
+### Fix GraphQL query directives validation bug
+
+GraphQL supports an obscure syntax, where a variable is used in a directive application on the same operation where the variable is declared.
+
+The router used to reject queries like this, but now they are accepted:
+
+```graphql
+query GetSomething($var: Int!) @someDirective(argument: $var) {
+  something
+}
+```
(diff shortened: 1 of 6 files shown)


apollo-federation/tests/query_plan/build_query_plan_tests/handles_operations_with_directives.rs
```diff
@@ -244,7 +244,7 @@ fn test_if_directives_with_arguments_applied_on_queries_are_ok() {
 }
 
 #[test]
-#[should_panic(expected = r#"unused variable: `$some_var`"#)]
+#[should_panic(expected = r#"snapshot assertion"#)]
```
goto-bus-stop, line 247: The expected error changed because apollo-compiler fixed validation. This test still fails for other reasons, which will be addressed in https://github.com/apollographql/router/pull/5743.

.changesets/fix_renee_operation_variables.md
```diff
@@ -0,0 +1,11 @@
+### Fix GraphQL query directives validation bug
```
goto-bus-stop, line 1: I'm not sure if it's worth adding a changeset for this, because as far as we know, none of our users have any query that does this...

lrlna (approved):

.changesets/fix_renee_operation_variables.md
lrlna: I'd keep the changeset. It's still a fix, even if it doesn't actually affect anyone.

PR #5754, opened 2024-07-31
Allow to use progressive override when using federation 2.7 and above.

o0Ignition0o:
A bug in the way we extracted progressive override metadata from the schema would lead to the plugin not being enabled when using join 0.5.


dariuszkuc (approved):

bnjjj (approved):

trevor-scheer (approved):
LGTM, thank you! I am really not sure why I capped it at 0.4, I can't think of a good reason.

PR #5757, opened 2024-08-01
Fix usage reporting of enum values in deferred requests

bonnici:
Fixes reporting of the enum values from responses in extended references for deferred requests. Deferred requests include two passes through the format response function, so we need to add to an existing map of enums rather than overwriting it on each pass. We also clear the map when submitting a report so that subscription reports only include the enum values referenced in responses from the particular query in the subscription request/event.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/apollo_studio_interop/mod.rs
+++ b/apollo-router/src/apollo_studio_interop/mod.rs
@@ -285,18 +285,17 @@ pub(crate) fn extract_enums_from_response(
     operation_name: Option<&str>,
     schema: &Valid<Schema>,
     response_body: &Object,
-) -> ReferencedEnums {
-    let mut result = ReferencedEnums::new();
+    existing_refs: &mut ReferencedEnums,
+) {
     if let Some(operation) = query.operation(operation_name) {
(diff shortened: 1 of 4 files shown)


Geal (approved):

bnjjj (approved):
Does it need a changelog @bonnici ?

bonnici:
> Does it need a changelog @bonnici ?

@bnjjj I can add one if it's needed but I don't really know the policy here. It's basically a bug fix.

bnjjj:
@bonnici I'm just asking if we need it. Do you think it's something that would be relevant for our customers to know ?If so let's create a changelog entry. If not then we don't need to do anything

bonnici:
Ah ok, I don't think it's needed then. This wasn't based on a customer request and the feature hasn't been enabled widely yet so I don't think anyone would know the bug exists yet.

PR #5759, opened 2024-08-01
fix(telemetry): improve support of conditions at the request level, especially for events

### LINKED ISSUE #5702: telemetry: `exists` condition is not triggered with some custom events
In some cases `exists` condition doesn't work properly. For example with a simple use case "I don't want to log for unnamed operation". We would be tempted to create this configuration:

```yaml
telemetry:
  instrumentation:
    events:
      supergraph:
        my.event:
          message: "Auditing Router Event"
          level: info
          on: request
          attributes:
            graphql.operation.name: true
          condition:
            exists:
              operation_name: string
```

but the way it's implemented doesn't fit this use case. Because of custom instruments that can be triggered at any stages (`response`|`request`|`error`|...) we can't just return a boolean. Because for example if we want to create a metric measuring the duration with a condition on `status_code` at router level then first when we will try to compute it we will check if the condition for request matches or not. If the selector is applied at the response level like `status_code` once we will hit `on_request` if we just return a boolean it will return false and then we won't compute it later. The reason is to optimize perf you don't want to compute something that is not going to be useful. That's why in this case we just return `None` because nothing returned from the `on_request` method on selector so we will continue to compute it later.

I can see 2 solutions here:

+ We just remove this optimization and keep the instrument running even if we received `false`  on `on_request` call. Could we have side effects by doing this ?

+ I think something that might be useful for future improvements. We should provide a new method on `Selector` trait to indicate at which stage it can get some values. Something like `stage(&self) -> &[Stage]` which contains all the potential stages at which we can return some data. Thanks to this we will be able to know if we can directly shut down the instruments or even just not compute useless stages.

The main question we should have is about side effects that this could create.  

The current workaround if you're trying to achieve the example with `operation_name` is to use `eq` with the `default` configuration like this:

```yaml
telemetry:
  instrumentation:
    events:
      supergraph:
        my.event:
          message: "Auditing Router Event"
          level: info
          on: request
          attributes:
            graphql.operation.name: true
          condition:
            any:
            - eq:
              - operation_name: string
                default: UNKNOWN # If it doesn't exist it will take this value
              - UNKNOWN
```

bnjjj:
`exists` condition is now properly handled with events, this configuration will now work:

```yaml
telemetry:
  instrumentation:
    events:
      supergraph:
        my.event:
          message: "Auditing Router Event"
          level: info
          on: request
          attributes:
            graphql.operation.name: true
          condition:
            exists:
              operation_name: string
```

Fixes #5702 

### Add warnings for invalid configuration on custom telemetry ([PR #5759](https://github.com/apollographql/router/issues/5759))

For example sometimes if you have configuration like this:

```yaml
telemetry:
  instrumentation:
    events:
      subgraph:
        my.event:
          message: "Auditing Router Event"
          level: info
          on: request
          attributes:
            subgraph.response.status:
              subgraph_response_status: code # This is a first warning because you can't access to the response if you're at the request stage
          condition:
            eq:
            - subgraph_name # Another warning because instead of writing subgraph_name: true which is the selector, you're asking for a comparison between 2 strings ("subgraph_name" and "product")
            - product
```

This configuration is syntaxically correct but wouldn't probably do what you would like to. I put comments to highlight 2 mistakes in this example.
Before it was silently computed, now you'll get warning when starting the router.

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [x] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/plugins/telemetry/config_new/mod.rs
```diff
@@ -51,6 +52,66 @@ pub(crate) trait Selectors {
...
+            Stage::Response
+            | Stage::ResponseEvent
+            | Stage::ResponseField
+            | Stage::Error
+            | Stage::Drop => {
+                if *other != Self::Request {
+                    Some(std::cmp::Ordering::Greater)
+                } else {
+                    Some(std::cmp::Ordering::Less)
+                }
```
Geal, line 109: why is it returning `Greater` whenever non `Request` stages compare amongst themselves?

apollo-router/src/plugins/telemetry/config_new/extendable.rs
```diff
@@ -255,6 +256,23 @@ where
...
+impl<A, E, Request, Response, EventResponse> Extendable<A, E>
+where
+    A: Default + Selectors<Request = Request, Response = Response, EventResponse = EventResponse>,
+    E: Selector<Request = Request, Response = Response, EventResponse = EventResponse>,
+{
+    pub(crate) fn validate(&self, restricted_stage: Option<Stage>) -> Result<(), String> {
+        if let Some(Stage::Request) = &restricted_stage {
+            for (name, custom) in &self.custom {
+                if !custom.is_active(Stage::Request) {
+                    return Err(format!("cannot set the attribute {name:?} because it's using a selector computed in another stage than 'request' because it won't be computed"));
```
Geal, line 268:
```suggestion
                    return Err(format!("cannot set the attribute {name:?} because it is using a selector computed in another stage than 'request' so it will not be computed"));
```

apollo-router/src/plugins/telemetry/config_new/conditions.rs
```diff
@@ -55,28 +56,95 @@ impl<T> Condition<T>
...
-                (None, Some(right)) => {
-                    eq[1] = SelectorOrValue::Value(right.into());
-                    None
-                }
-                (Some(left), None) => {
-                    eq[0] = SelectorOrValue::Value(left.into());
-                    None
+            Condition::Eq(arr) | Condition::Gt(arr) | Condition::Lt(arr) => match (&arr[0], &arr[1]) {
+                (SelectorOrValue::Value(val1), SelectorOrValue::Value(val2)) => {
+                    Err(format!("trying to compare 2 values ('{val1}' and '{val2}'), usually it's a syntax error because you want to use a specific selector and a value in a condition"))
```
BrynCooke, line 64: I'm not sure that it's always true that a condition must be a value on one side.

apollo-router/src/plugins/telemetry/config_new/conditions.rs
```diff
@@ -55,28 +56,95 @@ impl<T> Condition<T>
 where
     T: Selector,
 {
-    pub(crate) fn evaluate_request(&mut self, request: &T::Request) -> Option<bool> {
+    /// restricted_stage is Some if this condiiton will only applies at a specific stage like for events for example
+    pub(crate) fn validate(&self, restricted_stage: Option<Stage>) -> Result<(), String> {
```
BrynCooke, line 60: Let's talk about this. This is to catch selectors that will be triggered for stages after the containing event has taken place right?

bnjjj: Exactly

apollo-router/src/plugins/telemetry/config_new/conditions.rs
```diff
@@ -55,28 +56,95 @@ impl<T> Condition<T>
...
-                    if left == right {
-                        *self = Condition::True;
-                        Some(true)
-                    } else {
-                        Some(false)
+                (SelectorOrValue::Value(_), SelectorOrValue::Selector(sel)) | (SelectorOrValue::Selector(sel), SelectorOrValue::Value(_)) => {
+                    // Special condition for events
+                    if let Some(Stage::Request) = &restricted_stage {
+                        if !sel.is_active(Stage::Request) {
+                            return Err(String::from("can't compare a selector computed in another stage than 'request' because it won't be computed at all"));
```
BrynCooke, line 70: I'd be better if it said something like:
`selector X is only valid for X stages, this log event will never trigger.`

bnjjj: fixed

BrynCooke: Did you push your changes?
I don't see them.

bnjjj: Part of this commit https://github.com/apollographql/router/pull/5759/commits/b96552485e29fa9a1644e494f49aff46e65406f4

BrynCooke (approved):

PR #5761, opened 2024-08-01
Remove changes to FetchNode::response_at_path

pubmodmatt:
Reverts changes to the `FetchNode::response_at_path` function that were made for Connectors. The function is now identical to what it was before - only the visibility has changed so it can be called from `FetchService`. Most notably, this adds back the `&self` parameter. This simplifies the calling code, which already had access to the `FetchNode` and can just pass a reference to it in directly, rather than destructuring out fields to be passed to the function.

This is a minor code refactor - no functional changes were made.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

**Notes**

[CNN-371]: https://apollographql.atlassian.net/browse/CNN-371?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apollo-router/src/query_planner/fetch.rs
+++ b/apollo-router/src/query_planner/fetch.rs
@@ -371,7 +371,6 @@ impl FetchNode {
         &self,
         service: BoxService,
         subgraph_request: SubgraphRequest,
-        service_name: &str,
         current_dir: &Path,
         schema: &Schema,
         paths: Vec<Vec<Path>>,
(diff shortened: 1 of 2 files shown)


apollo-router/src/query_planner/fetch.rs
```diff
@@ -369,12 +368,11 @@ impl Variables {
 impl FetchNode {
     #[allow(clippy::too_many_arguments)]
     pub(crate) async fn subgraph_fetch(
+        &self,
         service: BoxService,
         subgraph_request: SubgraphRequest,
         service_name: &str,
```
dylan-apollo, line 374: Should this use `self.service_name` instead now, like `response_at_path` does? Or can these be intentionally different?

dylan-apollo (approved):
LGTM

PR #5766, opened 2024-08-02
fix(federation): misc fixes

dariuszkuc:
Fixes
* Fix handling of `SubgraphEnteringTransition` which should skip `RootTypeResolution` edges. This fixes the issue with redundantly processing same edges over and over again.
* Fix handling of `KeyResolution` which should compare whole selection sets of the conditions vs just the selections.
* Ensures consistent order of child nodes when computing new key nodes
* Add missing `self.on_modification()` when removing redundant edges
* Returns false instead of throwing an error when calculating if new requires node is useless (i.e. return false if parent relation path is `None`)
* Fix `PlanBuilder` trait to no longer swallow exceptions from query planning

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-federation/src/query_graph/build_query_graph.rs
+++ b/apollo-federation/src/query_graph/build_query_graph.rs
@@ -1888,7 +1888,9 @@ impl FederatedQueryGraphBuilder {
                             followup_edge_weight.transition,
                             QueryGraphEdgeTransition::KeyResolution
                         ) {
-                            if let (Some(conditions), Some(followup_conditions)) = (&edge_weight.conditions, &followup_edge_weight.conditions) {
+                            if let (Some(conditions), Some(followup_conditions)) =
+                                (&edge_weight.conditions, &followup_edge_weight.conditions)
+                            {
                                 if conditions == followup_conditions {
                                     continue;
                                 }
--- a/apollo-federation/src/query_graph/build_query_graph.rs
+++ b/apollo-federation/src/query_graph/build_query_graph.rs
@@ -1888,22 +1888,10 @@ impl FederatedQueryGraphBuilder {
                             followup_edge_weight.transition,
                             QueryGraphEdgeTransition::KeyResolution
                         ) {
-                            let Some(conditions) = &edge_weight.conditions else {
-                                return Err(SingleFederationError::Internal {
-                                    message: "Key resolution edge unexpectedly missing conditions"
-                                        .to_owned(),
+                            if let (Some(conditions), Some(followup_conditions)) = (&edge_weight.conditions, &followup_edge_weight.conditions) {
+                                if conditions == followup_conditions {
+                                    continue;
                                 }
-                                .into());
-                            };
-                            let Some(followup_conditions) = &followup_edge_weight.conditions else {
-                                return Err(SingleFederationError::Internal {
-                                    message: "Key resolution edge unexpectedly missing conditions"
-                                        .to_owned(),
-                                }
-                                .into());
-                            };
-                            if conditions.selections == followup_conditions.selections {
-                                continue;
                             }
                         }
                     }
@@ -1926,7 +1914,7 @@ impl FederatedQueryGraphBuilder {
                         // since we can do "start of query" -> C and that's always better.
                         if matches!(
                             followup_edge_weight.transition,
-                            QueryGraphEdgeTransition::SubgraphEnteringTransition
+                            QueryGraphEdgeTransition::RootTypeResolution { .. }
                         ) {
                             continue;
                         }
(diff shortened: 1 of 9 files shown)


apollo-federation/src/query_graph/build_query_graph.rs
```diff
@@ -1888,22 +1888,10 @@ impl FederatedQueryGraphBuilder {
                             followup_edge_weight.transition,
                             QueryGraphEdgeTransition::KeyResolution
                         ) {
-                            let Some(conditions) = &edge_weight.conditions else {
```
dariuszkuc, line 1891: previously we were failing when there are no conditions on key resolution edges -> should we continue doing this?

apollo-federation/src/query_plan/fetch_dependency_graph.rs
```diff
@@ -668,8 +667,12 @@ impl FetchDependencyGraph {
         //    keeping nodes separated when they have a different path in their parent
         //    allows to keep that "path in parent" more precisely,
         //    which is important for some case of @requires).
-        for existing_id in self.children_of(parent.parent_node_id) {
-            let existing = self.node_weight(existing_id)?;
+        // for existing_id in self.children_of(parent.parent_node_id) {
+        //     let existing = self.node_weight(existing_id)?;
+        // // TODO: REVERSING ORDER OF CHILDREN DOES FIX THE ISSUE BUT SEEMS WRONG
+        let children: Vec<NodeIndex> = self.children_of(parent.parent_node_id).collect();
+        for existing_id in children.iter().rev() {
```
dariuszkuc, line 674: should we sort this?

apollo-federation/src/query_plan/fetch_dependency_graph.rs
```diff
@@ -668,7 +668,9 @@ impl FetchDependencyGraph {
         //    keeping nodes separated when they have a different path in their parent
         //    allows to keep that "path in parent" more precisely,
         //    which is important for some case of @requires).
-        for existing_id in self.children_of(parent.parent_node_id) {
+        for existing_id in
+            FetchDependencyGraph::sorted_nodes(self.children_of(parent.parent_node_id))
+        {
```
goto-bus-stop, line 673: This will probably be slow when a fetch dependency node has very many children. Children are added one by one, so if this function is O(n), the complexity of adding many children is O(n^2). This is already a problem with `add_parent` and technically also with this loop, it will just be more pronounced if we also sort the children over and over.

I think it's okay if we do this here now, but we will have to improve the behaviour for nodes with many children in the near future.

apollo-federation/src/query_graph/build_query_graph.rs
goto-bus-stop: Based on the _must_ language here:
https://github.com/apollographql/router/blob/d2a4b1cbd2756a021ac6f8d94f6cb5ee960e820f/apollo-federation/src/query_graph/mod.rs#L195-L198
I think we should error if it's missing.

goto-bus-stop (commented):
LGTM. We should keep the error but otherwise good to go

apollo-federation/src/query_plan/fetch_dependency_graph.rs
dariuszkuc: Realistically this will be at most in the order of hundreds of nodes (maybe couple thousand?) so don't think this will have much impact but yeah something to keep in mind.

apollo-federation/src/query_plan/fetch_dependency_graph.rs
```diff
@@ -921,6 +949,7 @@ impl FetchDependencyGraph {
         let mut redundant_edges = HashSet::new();
         self.collect_redundant_edges(node_index, &mut redundant_edges);
 
+        self.on_modification();
```
duckki, line 952: Shouldn't we check `is_empty` here?
```suggestion
        if !redundant_edges.is_empty() {
            self.on_modification();
        }
```

duckki (commented):
It looks good to me. I'm running corpus for comparison.

duckki (approved):
Corpus comparison result looks good. No new issues found.

goto-bus-stop (approved):

apollo-federation/src/query_plan/fetch_dependency_graph.rs
goto-bus-stop: We already have a query where we spend multiple seconds in `add_parent` because of this 😄

PR #5767, opened 2024-08-02
Fix race condition when adding debug data to connector context

pubmodmatt:
There was a race condition due to the way debug data was being added to the `ConnectorsContext`. The connectors context was removed from the request context extensions for each query, updated, then put back. If the query planner executed multiple queries in parallel, they would all try to do this at once. Once one of them successfully removed the connectors context, any others running concurrently would get back `None`.

The connectors context is now left in place rather than removed, and is accessed through `Option<Arc<Mutex<ConnectorsContext>>>`. This preserves a single lookup in the `Extensions` for each fetch. If the debug feature is not enabled, the option will be `None`. If it is enabled, the mutex is locked and the context is updated with the debug information.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

**Notes**

[CNN-388]: https://apollographql.atlassian.net/browse/CNN-388?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apollo-router/src/plugins/connectors/handle_responses.rs
+++ b/apollo-router/src/plugins/connectors/handle_responses.rs
@@ -12,6 +12,7 @@ use crate::plugins::connectors::plugin::ConnectorContext;
 use crate::plugins::connectors::plugin::SelectionData;
 use crate::services::connect::Response;
 use crate::services::router::body::RouterBody;
+use crate::Context;
 
 const ENTITIES: &str = "_entities";
 const TYPENAME: &str = "__typename";
@@ -35,7 +36,7 @@ pub(crate) enum HandleResponseError {
 pub(crate) async fn handle_responses(
     responses: Vec<http::Response<RouterBody>>,
     connector: &Connector,
-    debug: &mut Option<ConnectorContext>,
+    context: Context,
     _schema: &Valid<Schema>, // TODO for future apply_with_selection
 ) -> Result<Response, HandleResponseError> {
     use HandleResponseError::*;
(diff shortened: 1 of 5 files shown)


apollo-router/src/plugins/connectors/handle_responses.rs
```diff
@@ -58,9 +59,11 @@ pub(crate) async fn handle_responses(
...
         if parts.status.is_success() {
             let Ok(json_data) = serde_json::from_slice::<Value>(body) else {
-                if let Some(ref mut debug) = debug {
-                    debug.push_invalid_response(&parts, body);
-                }
+                context.extensions().with_lock(|mut lock| {
+                    if let Some(ref mut debug) = lock.get_mut::<ConnectorContext>() {
+                        debug.push_invalid_response(&parts, body);
+                    }
+                });
```
dylan-apollo, line 66: I worry about lock contention—especially since it could slow down _all_ requests even if debug is disabled. Can we get router team eyes on this for a suggestion? Maybe we could make extensions `RwLock`, or maybe we should be storing an `Option<Mutex<ConnectorContext>>` and getting a reference to that (so we can still check for debug disabled).

pubmodmatt: Sure, I can reach out for suggestions. For your second proposal, storing `Option<Mutex<ConnectorContext>>` would still require locking the `ExtensionsGuard` to get that `Option`, so there's still potential contention there, or am I misunderstanding?

pubmodmatt: Actually, I think I see - you mean we could get that `Option` once and then if `Some`, lock the second `Mutex` inside the `Option`. So we'd only lock the extensions guard once instead of several times per request.

dylan-apollo: And because the Option is on the outside of the second mutex, we wouldn't need to lock it at all if debugging is disabled

pubmodmatt: I like this option the best and coded it up.

dylan-apollo (approved):
Looks good to me!

PR #5769, opened 2024-08-05
feat(federation): add query planner config flags to the CLI

goto-bus-stop:
A better way to tinker with options than editing the source.

### DIFF
--- a/apollo-federation/cli/src/main.rs
+++ b/apollo-federation/cli/src/main.rs
@@ -23,15 +23,15 @@ struct QueryPlannerArgs {
     /// Enable @defer support.
     #[arg(long, default_value_t = false)]
     enable_defer: bool,
-    /// Turn off fragment reuse in subgraph queries.
-    #[arg(long = "no-reuse-fragments", default_value_t = true, action = clap::ArgAction::SetFalse)]
+    /// Reuse fragments to compress subgraph queries.
+    #[arg(long, default_value_t = false)]
     reuse_fragments: bool,
-    /// Generate fragments to compress subgraph queries. Implies --no-reuse-fragments.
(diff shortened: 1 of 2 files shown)


apollo-federation/cli/src/main.rs
```diff
@@ -17,6 +18,32 @@ use clap::Parser;
 mod bench;
 use bench::run_bench;
 
+#[derive(Parser)]
+struct QueryPlannerArgs {
+    /// Enable @defer support.
+    #[arg(long, default_value_t = false)]
+    enable_defer: bool,
+    /// Turn off fragment reuse in subgraph queries.
+    #[arg(long = "no-reuse-fragments", default_value_t = true, action = clap::ArgAction::SetFalse)]
```
duckki, line 27: Wouldn't it be more consistent with harness to have "reuse-fragments" with false as default?

goto-bus-stop: my goal here is to match the default configuration of the query planner.
I would've preferred to have `--reuse-fragments` to explicitly turn it on and `--no-reuse-fragments` to explicitly turn it off, and using the QueryPlannerConfig default if neither is given. Maybe that's possible if i make it an Option, actually 🤔

duckki: I had a similar discussion with @lrlna over at performance-harness. I thought the same thing and added `--disable-reuse-fragments`, but I realized it would be another complexity added, while `--reuse-fragments <true/false>` option is used everywhere else.

duckki (approved):
Besides some bikeshedding on the command line option, it looks good otherwise. We can always change the internal cli tool's interface, if we want to.

PR #5770, opened 2024-08-05
Add metrics for cache entry memory size

BrynCooke:
Query planner cache entries may use an significant amount of memory in the Router.
To help users understand and monitor this the Router now exposes a new metric `apollo.router.cache.storage.estimated_size`.

This metric give an estimated size in bytes for the cache entry and has the following attributes:
- `kind`: `query planner`.
- `storage`: `memory`.

As the size is only an estimation, users should check for correlation with pod memory usage to determine if cache needs to be updated.

Usage scenario:
1. Your pods are being terminated due to memory pressure.
2. Add the following metrics to your monitoring system to track:
  * `apollo.router.cache.storage.estimated_size`.
  * `apollo_router_cache_size`.
  * ratio of `apollo_router_cache_hits` - `apollo_router_cache_misses`.
3. Observe the `apollo.router.cache.storage.estimated_size` to see if it grows over time and correlates with pod memory usage.
4. Observe the ratio of cache hits to misses to determine if the cache is being effective.

Remediation:
* Adjust the cache size to lower if the cache reaches near 100% hit rate but the cache size is still growing.
* Increase the pod memory to higher if cache hit rate is low and cache size is still growing.
* Adjust the cache size to lower if the latency of query planning cache misses is acceptable and memory availability is limited.

## Technical info
The estimate is only implemented for query plans and uses serde to create the estimate without actually writing to a string. It is very difficult to provide accurate memory usage, so it's only useful to see a trend.

Providing an estimated size for a cache entry is optional, and only implemented for query planner at this time.
Where no estimated size is set the gauge will not be emitted, e.g. APQ or entity caching.

## Testing
For manual testing we will deploy internally and observe that the metric tops out when the query cache is full.

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/cache/size_estimation.rs
```diff
@@ -0,0 +1,438 @@
...
+        Ok(self)
+    }
+
+    fn serialize_bytes(mut self, v: &[u8]) -> Result<Self::Ok, Self::Error> {
+        self.count += v.len();
+        Ok(self)
+    }
+
+    fn serialize_none(self) -> Result<Self::Ok, Self::Error> {
+        Ok(self)
```
IvanGoncharov, line 136: Just to double check: `None` would still use the size of `Some` in memory, right?
We estimate it as 0 because of serve limitations?

apollo-router/src/cache/storage.rs
```diff
@@ -187,25 +255,24 @@ where
...
+            .fetch_add(value.estimated_size().unwrap_or(0) as i64, Ordering::SeqCst);
         let mut in_memory = self.inner.lock().await;
-        in_memory.put(key, value);
-        let size = in_memory.len() as u64;
-        tracing::info!(
-            value.apollo_router_cache_size = size,
-            kind = %self.caller,
-            storage = &tracing::field::display(CacheStorageName::Memory),
-        );
+        if let Some((_, v)) = in_memory.push(key.clone(), value.clone()) {
```
Geal, line 270: the in memory cache lock is held until the end of the function, it should only be held when inserting the old entry.
Maybe fixed with something like:
```rust
let old_entry = {
  self.inner.lock().await.push(key.clone(), value.clone())
};
if let Some((_, v)) = old_entry {
```

Geal (changes requested):

apollo-router/src/cache/storage.rs
BrynCooke: Fixed in: [e37c87e](https://github.com/apollographql/router/pull/5770/commits/e37c87ebc97cb588dfc1de831760df1d0ce1d68f)

apollo-router/src/cache/size_estimation.rs
BrynCooke: Yeah, I guess in theory we can do something better. But None values won't have strings etc in them so I don't think we should worry about this and maybe we can follow up later to improve things.

Geal (approved):

.changesets/feat_enhanced_observability.md
```diff
@@ -0,0 +1,27 @@
...
+Query planner cache entries may use an significant amount of memory in the Router.
+To help users understand and monitor this the Router now exposes a new metric `apollo.router.cache.storage.estimated_size`.
+
+This metric give an estimated size in bytes for the cache entry and has the following attributes:
+- `kind`: `query planner`.
+- `storage`: `memory`.
+
+As the size is only an estimation, users should check for correlation with pod memory usage to determine if cache needs to be updated.
+
+Usage scenario:
```
BrynCooke, line 12: @shorgi Maybe this info should also be in the docs under cache troubleshooting?

BrynCooke:
Going to wait for a docs review before merging.

.changesets/feat_enhanced_observability.md
shorgi: This is great content for docs. Since it's related to pod memory pressure, moving it to the k8s page.

shorgi, line 12:
```suggestion
```

.changesets/feat_enhanced_observability.md
```diff
@@ -0,0 +1,27 @@
...
+  * `apollo_router_cache_size`.
+  * ratio of `apollo_router_cache_hits` - `apollo_router_cache_misses`.
+
+* Observe the `apollo.router.cache.storage.estimated_size` to see if it grows over time and correlates with pod memory usage.
+* Observe the ratio of cache hits to misses to determine if the cache is being effective.
+
+Remediation:
+* Adjust the cache size to lower if the cache reaches near 100% hit rate but the cache size is still growing.
+* Increase the pod memory to higher if cache hit rate is low and cache size is still growing.
+* Adjust the cache size to lower if the latency of query planning cache misses is acceptable and memory availability is limited.
```
shorgi, line 25:
```suggestion
To learn how to troubleshoot with this metric, see the [Pods terminating due to memory pressure](https://www.apollographql.com/docs/router/containerization/kubernetes#pods-terminating-due-to-memory-pressure) guide in docs.
```

shorgi (commented):
Moved the usage scenario from the changeset into docs

BrynCooke:
ROUTER-669

PR #5771, opened 2024-08-05
Share a single Rust QP instance when using a pool of JS workers

SimonSapin:
Config `supergraph.query_planning.experimental_parallelism` defaults to 1, but when set to something else `BridgeQueryPlannerPool` creates that many instances of `BridgeQueryPlanner` which in turn create a JS worker each.

If config `experimental_query_planner_mode` is also set to `new` or `both` (current default is `legacy`), the each `BridgeQueryPlanner` would also unnecessarily create its own instance of the Rust `QueryPlanner` struct.

Instead, a single `QueryPlanner` is now shared with `Arc`. Unlike JS workers, `QueryPlanner` only contains read-only data and can safely be shared between threads. This should reduce memory use when both `supergraph.query_planning.experimental_parallelism` and `experimental_query_planner_mode` are configured to non-default values.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/plugins/demand_control/cost_calculator/static_cost.rs
+++ b/apollo-router/src/plugins/demand_control/cost_calculator/static_cost.rs
@@ -464,7 +464,7 @@ mod tests {
         let config: Arc<Configuration> = Arc::new(Default::default());
         let (schema, query) = parse_schema_and_operation(schema_str, query_str, &config);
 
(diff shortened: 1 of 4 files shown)


TylerBloom (approved):
LGTM

PR #5772, opened 2024-08-05
feat(federation): minimal fragment generation

goto-bus-stop:
This implements the basic fragment generation algorithm that is currently in use in JS.

Inline fragments that select at least 2 fields are converted to named fragments. This reduces the size of the query if multiple inline fragments can be replaced by spreads of the same named fragment.

Differences from JS:
- JS generates somewhat informative fragment names. Here I generate the shortest possible names.
- JS never changes an inline fragment if it has directives. Here I special-case `@include` and `@skip`, which by spec behave the same on an inline fragment spread and a named fragment spread. This means more fragments can be moved out.

Overall, fragment generation works better than fragment reuse, and is significantly faster. There are still cases where fragment generation produces bigger outputs than reuse, because it is not very smart (it doesn't touch repeated fields in non-fragment selection sets).

This PR is ready for review but to enable it by default we will need further profiling and improvements to the algorithm.


apollo-federation/src/operation/optimize.rs
```diff
@@ -1549,6 +1552,16 @@ impl Operation {
         self.reuse_fragments_inner(fragments, Self::DEFAULT_MIN_USAGES_TO_OPTIMIZE)
     }
 
+    pub(crate) fn generate_fragments(&mut self) -> Result<(), FederationError> {
+        let mut generator = FragmentGenerator::default();
+        generator.visit_selection_set(&mut self.selection_set)?;
+        let fragments = generator.into_inner();
+        // To remove single-use fragment definitions:
+        // self.selection_set = fragments.reduce(&self.selection_set, 2)?;
```
duckki, line 1560: Do you plan to uncomment this line?

duckki (commented):
The code looks simple and effective, albeit conservative.

In the future, for better compression, we might want to explore:
- Grouping idiomatic selections into fragments. This might amount to Huffman encoding of some sort :D
- Choosing the abstract-most type condition applicable, so that two inline spreads with different (concrete) type conditions may be replaced the same abstract-type fragment. But, this is tricky since `FieldsInSetCanMerge` rule is too strict for abstarct types and may prevent valid compressions in some cases.

goto-bus-stop:
Yes, it's very conservative, just the JS implementation with improved behaviour on `@skip`/`@include` :)

Honestly I think we could get very far by just doing the same thing for entire field subselections, even without doing anything for partially overlapping selections.

apollo-federation/src/operation/optimize.rs
goto-bus-stop: I've removed it. This is something that we could do, but it doesn't make a big difference in query size, and I think a great benefit of fragment generation is how simple and obvious it is (including being a single pass, while reducing fragments requires looping until getting to a stable result).

We should consider it if/when we are doing more complicated extractions, where the entire fragment definition is overhead

apollo-federation/src/operation/optimize.rs
```diff
@@ -1579,6 +1589,155 @@ impl Operation {
     }
 }
 
+/// Returns a consistent GraphQL name for the given index.
+fn fragment_name(mut index: usize) -> Name {
```
dariuszkuc, line 1593: Unsure how much we gain by auto generating random fragment names vs something human readable. Guess folks generally shouldn't be looking at those ops... but it would make it easier to debug if needed.

goto-bus-stop: You *always* need something like this to make sure that names don't collide, that decision is just between serializing the number as a string or serializing it to a minimal name like this. Making it human readable is just doing a bit more work ;)
The JS uses a name like `generated_onTypeCondition`, but the type condition is already in the definition, so it doesn't really help in my opinion.

duckki (approved):

goto-bus-stop:
Enabling auto merge since feedback was positive in calls, and the implementation is fairly simple. Please submit a post-merge review if you have more comments/concerns!

PR #5778, opened 2024-08-06
chore(federation): remove `for_each_element`

goto-bus-stop:
Today's 15-minute cleanup :)

This set of functions is used in one place, and it has some odd behaviours. In JS, an "element" is free, because its data structures are built on element/selection set pairings. But we don't have the "element" type exactly in our data structures, our closest equivalent is the `Selection` type. Therefore the `for_each_element` function had to convert types including some inexpensive, but non-zero clones.

`for_each_element` also translates named fragment spreads to inline fragments. AFAICT, this is a historical artifact ported from JS, as the only place where `for_each_element` was used does not support named fragment spreads to begin with.

I replaced it, for now, with a manual visitor. I think it might make sense to have something like a `for_each_selection` API but it's honestly not that much harder to do it manually...

### DIFF
--- a/apollo-federation/src/operation/mod.rs
+++ b/apollo-federation/src/operation/mod.rs
@@ -942,22 +942,6 @@ impl Selection {
             }
         }
     }
-
-    pub(crate) fn for_each_element(
-        &self,
-        parent_type_position: CompositeTypeDefinitionPosition,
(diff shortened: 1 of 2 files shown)


dariuszkuc (approved):
Nice!

PR #5782, opened 2024-08-06
Ignore non-apollo directives when extracting demand control directives to subgraphs

tninesling:
When users set up a custom directive which conflicts with demand control directive names, we would incorrectly try to extract that to subgraphs as the federation directive. This updates the extraction code to ignore non-apollo directives when propagating demand control directives to subgraphs.

---

**Checklist**

- Tests added and passing[^3]
    - [X] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-federation/tests/extract_subgraphs.rs
+++ b/apollo-federation/tests/extract_subgraphs.rs
@@ -511,3 +511,93 @@ fn extracts_renamed_demand_control_directives() {
     }
     insta::assert_snapshot!(snapshot);
 }
+
+#[test]
+fn does_not_extract_demand_control_directive_name_conflicts() {
+    let subgraphs = Supergraph::new(r#"
+        schema
+          @link(url: "https://specs.apollo.dev/link/v1.0")
+          @link(url: "https://specs.apollo.dev/join/v0.5", for: EXECUTION)
+          @link(url: "https://example.com/myCustomDirective/v1.0", import: ["@cost"])
+          @link(url: "https://example.com/myOtherCustomDirective/v1.0", import: ["@listSize"])
+        {
+          query: Query
+        }
+        
+        directive @cost(name: String!) on FIELD_DEFINITION
+        
+        directive @join__directive(graphs: [join__Graph!], name: String!, args: join__DirectiveArguments) repeatable on SCHEMA | OBJECT | INTERFACE | FIELD_DEFINITION
+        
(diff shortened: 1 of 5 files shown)


apollo-federation/tests/extract_subgraphs.rs
```diff
@@ -511,3 +511,93 @@ fn extracts_renamed_demand_control_directives() {
...
+          EXECUTION
+        }
+        
+        type Query
+          @join__type(graph: SUBGRAPH_A)
+          @join__type(graph: SUBGRAPH_B)
+        {
+          a: Int @join__field(graph: SUBGRAPH_A) @cost(name: "cost")
+          b: [Int] @join__field(graph: SUBGRAPH_B) @listSize(name: "listSize")
+        }
```
lrlna, line 585: for _extra_ posterity, can you add a test where a _renamed_ directive is using an apollo directive (and therefore is ignored when extracting0?

lrlna (changes requested):
Nice catch! I had an idea for an extra test with renamed directives, which is why the `request changes`.

apollo-federation/tests/extract_subgraphs.rs
lrlna: ah i wasn't very clear. let me try again. there is a test above, `extracts_renamed_demand_control_directives()` that uses a rename as part of `@link`:
```graphql
 @link(url: "https://specs.apollo.dev/cost/v0.1", import: [{name: "@cost", as: "@renamedCost"}, {name: "@listSize", as: "@renamedListSize"}])
```

What would happen if we use the renamed directives in that import statement, and the `Query` definition you have here,
```graphql
        type Query
          @join__type(graph: SUBGRAPH_A)
          @join__type(graph: SUBGRAPH_B)
        {
          a: Int @join__field(graph: SUBGRAPH_A) @renamedCost(name: "renamedCost")
          b: [Int] @join__field(graph: SUBGRAPH_B) @renamedListSize(name: "renamedListSize")
        }
```

or, alternatively something like this (although given the description of this PR and what you've changed, I don't think this example actually applies):
```graphql
        type Query
          @join__type(graph: SUBGRAPH_A)
          @join__type(graph: SUBGRAPH_B)
        {
          a: Int @join__field(graph: SUBGRAPH_A) @renamedCost(name: "cost")
          b: [Int] @join__field(graph: SUBGRAPH_B) @renamedListSize(name: "listSize")
        }
```

PR #5785, opened 2024-08-07
chore(federation): remove incorrect test

goto-bus-stop:
This tests a functionality that does not make sense and is invalid per GraphQL spec. `@include` cannot be used on a named fragment definition.

### DIFF
--- a/apollo-federation/src/operation/optimize.rs
+++ b/apollo-federation/src/operation/optimize.rs
@@ -2987,66 +2987,6 @@ mod tests {
     ///
 
     #[test]
-    #[should_panic(expected = "directive cannot be used on FRAGMENT_DEFINITION")]
-    // TODO: Investigate this restriction on query document in Rust version.
-    fn reuse_fragments_with_same_directive_on_the_fragment() {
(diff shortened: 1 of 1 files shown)


apollo-federation/src/operation/optimize.rs
```diff
@@ -2986,66 +2986,6 @@ mod tests {
     /// applied directives
     ///
 
-    #[test]
-    #[should_panic(expected = "directive cannot be used on FRAGMENT_DEFINITION")]
-    // TODO: Investigate this restriction on query document in Rust version.
-    fn reuse_fragments_with_same_directive_on_the_fragment() {
```
dariuszkuc, line 2992: instead of deleting this test -> should we update it to use valid (custom) directive?

goto-bus-stop: Yeah on second thought you are right. I'm not sure what we even do if there's a custom directive on a fragment definition. There's a good chance we do something nonsensical and that warrants a test.

lrlna (approved):

PR #5790, opened 2024-08-08
Add `experimental_query_planner_mode: both_best_effort` config

SimonSapin:
Some supergraph schemas supported by current Router versions (notably those using Apollo Federation 1) are not supported by the new Rust query planner. When a user explicitly configures `experimental_query_planner_mode: new` or `experimental_query_planner_mode: both`, we want unsupported schemas to cause an error at Router startup in order to make the issue very visible.

This adds `experimental_query_planner_mode: both_best_effort` which is like `both` whenever possible, and silently falls back to `legacy` otherwise. This behavior makes it suitable to potentially become the new default config.

Fixes ROUTER-601

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [x] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/query_planner/bridge_query_planner.rs
+++ b/apollo-router/src/query_planner/bridge_query_planner.rs
@@ -161,7 +161,13 @@ impl PlannerMode {
             }
             QueryPlannerMode::BothBestEffort => match Self::rust(schema, configuration) {
                 Ok(planner) => Ok(Some(planner)),
-                Err(_) => Ok(None),
+                Err(error) => {
+                    tracing::warn!(
+                        "Failed to initialize the new query planner, \
+                         falling back to legacy: {error}"
+                    );
+                    Ok(None)
+                }
             },
         }
     }
--- a/apollo-router/src/query_planner/bridge_query_planner.rs
+++ b/apollo-router/src/query_planner/bridge_query_planner.rs
@@ -137,6 +137,16 @@ impl PlannerMode {
                     "expected Rust QP instance for `experimental_query_planner_mode: both`",
                 ),
             },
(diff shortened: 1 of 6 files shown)


apollo-router/src/query_planner/bridge_query_planner.rs
```diff
@@ -149,6 +159,10 @@ impl PlannerMode {
             QueryPlannerMode::New | QueryPlannerMode::Both => {
                 Ok(Some(Self::rust(schema, configuration)?))
             }
+            QueryPlannerMode::BothBestEffort => match Self::rust(schema, configuration) {
+                Ok(planner) => Ok(Some(planner)),
+                Err(_) => Ok(None),
+            },
```
SimonSapin, line 165: This makes fallback completely silent. Should we log a warning instead?

goto-bus-stop: it'd be nice to warn for fed 1 (since it's a future compatibility issue), but maybe not for Rust QP bugs or missing features

TylerBloom: It seems unlike that constructing the Rust QP would fail in places that the JS QP doesn't. If we were to log errors here, what would we want to log?

TylerBloom (approved):

apollo-router/src/query_planner/bridge_query_planner.rs
SimonSapin: Added a `tracing::warn!` call and a test for it

PR #5793, opened 2024-08-09
support for clustering in the redis DEL command

Geal:
When using redis in cluster mode, if we are trying to use any comounding commands like MGET or DEL, we must first split the list of keys per cluster hash, then query each redis instance according to its hash.
This was [already implemented for MGET](https://github.com/apollographql/router/commit/2c506fe128ff6acdeab42c38fbdc2049ea4f7ebb), this PR fixes it for the DEL command used in invalidation.

Our integration tests cannot use redis clusters yet, so this was checked with manual tests. I can help in setting it up to reproduce the issue and fix if someone else wants to try

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-589]: https://apollographql.atlassian.net/browse/ROUTER-589?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


garypen (approved):

apollo-router/src/cache/redis.rs
```diff
@@ -562,16 +562,29 @@ impl RedisCacheStorage {
     }
 
     pub(crate) async fn delete<K: KeyType>(&self, keys: Vec<RedisKey<K>>) -> Option<u32> {
-        self.inner
-            .del(keys)
-            .await
-            .map_err(|e| {
-                if !e.is_not_found() {
+        let mut h: HashMap<u16, Vec<String>> = HashMap::new();
```
bnjjj, line 565:
```suggestion
        let mut h: HashMap<u16, Vec<String>> = HashMap::with_capacity(keys.len());
```

bnjjj: Does it make sense ?

bnjjj (approved):

apollo-router/src/cache/redis.rs
Geal: yes

Geal: kinda

Geal: there will be less entries in the hashmap than the number of keys

bnjjj: Feel free to decide whatever is better. It was just a suggestion :)

PR #5796, opened 2024-08-09
chore(federation): fix merged abstract types test

dariuszkuc:
Due to the differences between JS and RS graph structures (JS is always topologically sorted) we occassionally end up with different parallel fetch node  order....

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-federation/tests/query_plan/build_query_plan_tests/merged_abstract_types_handling.rs
+++ b/apollo-federation/tests/query_plan/build_query_plan_tests/merged_abstract_types_handling.rs
@@ -632,8 +632,6 @@ fn handles_spread_unions_correctly() {
 }
 
 #[test]
-#[should_panic(expected = "snapshot assertion")]
-// TODO: investigate this failure (reverse order of parallel fetches)
 fn handles_case_of_key_chains_in_parallel_requires() {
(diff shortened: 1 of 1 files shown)


duckki:
The `it_handles_complex_require_chain` test is unstable and sporadically fails. We need to handle that non-determinism as well.

goto-bus-stop:
A result being unstable is very undesireable. It might be due to using a plain HashMap or HashSet somewhere, as hashing is seeded with OS randomness and affects the order of the map.

lrlna:
> The it_handles_complex_require_chain test is unstable and sporadically fails. We need to handle that non-determinism as well.

Should we then fix the non-determinism first before fixing the test? 😅

dariuszkuc:
> The `it_handles_complex_require_chain` test is unstable and sporadically fails. We need to handle that non-determinism as well.

Indeed. Didn't notice that before. I'll revert the change.

duckki (approved):
With the non-determinism out of the picture, the change looks good!

PR #5797, opened 2024-08-09
fix(federation): fixed a QP panic saying "would create a cycle"

duckki:
The `merge_fetches_to_same_subgraph_and_same_inputs` function must merge decendant nodes into ancestor nodes and not the other way around (to avoid cycles).

This PR fixes it by topologically sorting the fetch dependency graph nodes before collecting nodes to merge. So, their merger is always topologically ordered.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [x] Integration Tests
    - [x] Manual Tests

[ROUTER-546]: https://apollographql.atlassian.net/browse/ROUTER-546?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apollo-federation/src/query_plan/fetch_dependency_graph.rs
+++ b/apollo-federation/src/query_plan/fetch_dependency_graph.rs
@@ -1451,7 +1451,9 @@ impl FetchDependencyGraph {
         // subgraph name, but have no worries for `mergeAt` since it contains either number of
         // field names, and the later is restricted by graphQL so as to not be an issue.
         let mut by_subgraphs = MultiMap::new();
-        for node_index in self.graph.node_indices() {
+        let sorted_nodes = petgraph::algo::toposort(&self.graph, None)
+            .map_err(|_| FederationError::internal("Failed to sort nodes due to cycle(s)"))?;
+        for node_index in sorted_nodes {
             let node = self.node_weight(node_index)?;
             // We exclude nodes without inputs because that's what we look for. In practice, this
             // mostly just exclude root nodes, which we don't really want to bother with anyway.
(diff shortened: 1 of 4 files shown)


apollo-federation/src/query_plan/fetch_dependency_graph.rs
```diff
@@ -1451,7 +1451,9 @@ impl FetchDependencyGraph {
         // subgraph name, but have no worries for `mergeAt` since it contains either number of
         // field names, and the later is restricted by graphQL so as to not be an issue.
         let mut by_subgraphs = MultiMap::new();
-        for node_index in self.graph.node_indices() {
+        let sorted_nodes = petgraph::algo::toposort(&self.graph, None)
+            .map_err(|_| FederationError::internal("Failed to sort nodes due to cycle(s)"))?;
+        for node_index in sorted_nodes {
```
lrlna, line 1456: non-blocking comment: what's the performance cost of doing a topological sort here, do you know?

lrlna (approved):

apollo-federation/src/query_plan/fetch_dependency_graph.rs
goto-bus-stop: I tried a few queries with known-big fetch dependency graphs, and this function doesn't really show up in a flame graph. I think because it only happens once, as the absolute last step, when the graph has already been reduced as much as it can.

lrlna: excellent excellent!!

duckki: Thanks for that analysis. I didn't observe any noticeable slowdown during corpus comparison runs, either.

PR #5798, opened 2024-08-09
Connectors request limit

pubmodmatt:
Add an optional limit on the number of REST API requests per GraphQL operation for a connectors source.

The limit can be specified in the router config YAML:

```yaml
preview_connectors:
  max_requests_per_operation_per_source: 100
```

The limit applies to each connectors source, or to each connector for connectors without a source configured.

Alternatively, it is possible to set the limit for each individual connector source:

```yaml
preview_connectors:
  subgraphs:
    subgraph_name:
      sources:
        source_name:
          max_requests_per_operation: 2
```

The limit can also be set with the `APOLLO_CONNECTORS_MAX_REQUESTS_PER_OPERATION` environment variable.

When the maximum number of requests is exceeded, a `null` will be returned at the corresponding field in the GraphQL response, and a `REQUEST_LIMIT_EXCEEDED` error will be added to the GraphQL response. Data for requests that ran successfully will still be returned.

The router will also log a warning indicating that the maximum number of requests was exceeded:

```
WARN ... Request limit exceeded for connector source jsonPlaceholder: max: 2, attempted: 11
```

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

**Notes**

[CNN-342]: https://apollographql.atlassian.net/browse/CNN-342?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apollo-router/src/plugins/connectors/configuration.rs
+++ b/apollo-router/src/plugins/connectors/configuration.rs
@@ -24,7 +24,7 @@ pub(crate) struct ConnectorsConfig {
 
     /// The maximum number of requests for a connector source
     #[serde(default)]
-    pub(crate) max_requests: Option<usize>,
+    pub(crate) max_requests_per_operation_per_source: Option<usize>,
 }
 
 /// Configuration for a connector subgraph
@@ -47,7 +47,7 @@ pub(crate) struct SourceConfiguration {
     pub(crate) override_url: Option<Url>,
 
     /// The maximum number of requests for this source
-    pub(crate) max_requests: Option<usize>,
+    pub(crate) max_requests_per_operation: Option<usize>,
 }
 
 /// Modifies connectors with values from the configuration
@@ -72,7 +72,7 @@ pub(crate) fn apply_config(config: &Configuration, mut connectors: Connectors) -
             if let Some(url) = source_config.override_url.as_ref() {
                 connector.transport.source_url = Some(url.clone());
             }
-            if let Some(max_requests) = source_config.max_requests {
+            if let Some(max_requests) = source_config.max_requests_per_operation {
                 connector.max_requests = Some(max_requests);
             }
         }
--- a/apollo-router/src/plugins/connectors/configuration.rs
+++ b/apollo-router/src/plugins/connectors/configuration.rs
@@ -21,6 +21,10 @@ pub(crate) struct ConnectorsConfig {
     /// Enables connector debugging information on response extensions if the feature is enabled
     #[serde(default)]
     pub(crate) debug_extensions: bool,
+
+    /// The maximum number of requests for a connector source
+    #[serde(default)]
+    pub(crate) max_requests: Option<usize>,
 }
 
 /// Configuration for a connector subgraph
@@ -41,6 +45,9 @@ pub(crate) struct SubgraphConnectorConfiguration {
 pub(crate) struct SourceConfiguration {
     /// Override the `@source(http: {baseURL:})`
     pub(crate) override_url: Option<Url>,
+
+    /// The maximum number of requests for this source
+    pub(crate) max_requests: Option<usize>,
 }
 
 /// Modifies connectors with values from the configuration
@@ -56,15 +63,18 @@ pub(crate) fn apply_config(config: &Configuration, mut connectors: Connectors) -
         let Some(subgraph_config) = config.subgraphs.get(&connector.id.subgraph_name) else {
             continue;
         };
-        let source_config = connector
+        if let Some(source_config) = connector
             .id
             .source_name
             .as_ref()
-            .and_then(|source_name| subgraph_config.sources.get(source_name));
-        if let Some(url) =
-            source_config.and_then(|source_config| source_config.override_url.as_ref())
+            .and_then(|source_name| subgraph_config.sources.get(source_name))
         {
-            connector.transport.source_url = Some(url.clone());
+            if let Some(url) = source_config.override_url.as_ref() {
+                connector.transport.source_url = Some(url.clone());
+            }
+            if let Some(max_requests) = source_config.max_requests {
+                connector.max_requests = Some(max_requests);
+            }
         }
 
         connector.config = Some(subgraph_config.custom.clone());
--- a/apollo-router/src/plugins/connectors/handle_responses.rs
+++ b/apollo-router/src/plugins/connectors/handle_responses.rs
@@ -4,11 +4,13 @@ use apollo_compiler::validation::Valid;
 use apollo_compiler::Schema;
 use apollo_federation::sources::connect::ApplyTo;
 use apollo_federation::sources::connect::Connector;
+use inflector::cases::screamingsnakecase::to_screaming_snake_case;
 use parking_lot::Mutex;
 use serde_json_bytes::ByteString;
 use serde_json_bytes::Value;
 
 use crate::graphql;
+use crate::plugins::connectors::error::Error as ConnectorError;
 use crate::plugins::connectors::make_requests::ResponseKey;
 use crate::plugins::connectors::make_requests::ResponseTypeName;
 use crate::plugins::connectors::plugin::ConnectorContext;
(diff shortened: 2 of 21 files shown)


apollo-router/src/plugins/connectors/handle_responses.rs
```diff
@@ -48,116 +50,137 @@ pub(crate) async fn handle_responses(
...
-                            errors: apply_to_errors,
-                        }),
-                    );
-                }
-                res.unwrap_or_else(|| Value::Null)
-            };
+        let mut error = None;
+        let response_key = response.1;
+        match response.0 {
+            Err(e) => error = Some((e.to_string(), to_screaming_snake_case(&format!("{:?}", e)))),
```
lennyburdette, line 56: i'm wondering if this should be something like `e.code()` on ConnectorError — we should be deliberate how we emit error information because customers start relying on these values real quick

apollo-router/src/plugins/connectors/configuration.rs
```diff
@@ -21,6 +21,10 @@ pub(crate) struct ConnectorsConfig {
     /// Enables connector debugging information on response extensions if the feature is enabled
     #[serde(default)]
     pub(crate) debug_extensions: bool,
+
+    /// The maximum number of requests for a connector source
+    #[serde(default)]
+    pub(crate) max_requests: Option<usize>,
```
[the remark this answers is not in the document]
pubmodmatt: It seems reasonable to expect that someone setting this kind of configuration would read the documentation, but I'm ok with making the names more clear. These names are on the long side, but maybe that's ok. Though the first one is really `max_requests_per_operation_per_source_unless_it_has_no_source_in_which_case_its_per_connector` 😆

apollo-router/src/plugins/connectors/handle_responses.rs
pubmodmatt: Done

apollo-router/src/plugins/connectors/configuration.rs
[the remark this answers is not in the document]
pubmodmatt: Done

apollo-router/src/plugins/connectors/handle_responses.rs
```diff
@@ -36,7 +35,10 @@ pub(crate) enum HandleResponseError {
 // --- RESPONSES ---------------------------------------------------------------
 
 pub(crate) async fn handle_responses(
-    responses: Vec<http::Response<RouterBody>>,
+    responses: Vec<(
+        Result<http::Response<RouterBody>, ConnectorError>,
+        ResponseKey,
+    )>,
```
nicholascioli, line 41: Small nit, but I think that it might be worth making a struct to contain more complex tuples.

apollo-router/src/plugins/connectors/handle_responses.rs
```diff
@@ -48,116 +50,137 @@ pub(crate) async fn handle_responses(
     let count = responses.len();
 
     for response in responses {
```
nicholascioli, line 52: Small nit: you could deconstruct the tuple (or separate struct) here to make all uses of `response.#` a little more legible. Something like the following:

```rust
for (fetch_result, response_key) in responses {...}
```

apollo-router/src/plugins/connectors/handle_responses.rs
```diff
@@ -48,116 +50,137 @@ pub(crate) async fn handle_responses(
...
+
+                if parts.status.is_success() {
+                    let Ok(json_data) = serde_json::from_slice::<Value>(body) else {
+                        if let Some(debug) = debug {
+                            debug.lock().push_invalid_response(&parts, body);
+                        }
+                        return Err(InvalidResponseBody(
+                            "couldn't deserialize response body".into(),
+                        ));
+                    };
```
nicholascioli, line 71:
```suggestion
                    let json_data = serde_json::from_slice::<Value>(body).map_err(|e| {
                        if let Some(debug) = debug {
                            debug.lock().push_invalid_response(&parts, body);
                        }
                        InvalidResponseBody(
                            format!("couldn't deserialize response body: {e}"),
                        )
                    })?;
```

nicholascioli: This is mostly to preserve the original error, but feel free to ignore.

apollo-router/src/plugins/connectors/handle_responses.rs
nicholascioli, line 1: Small nit, but I would suggest clarifying what the various tuple fields are as either named variables or a struct with named members. It can be hard to follow what `errors_to_push.0` might be down the line.

apollo-router/src/plugins/connectors/plugin.rs
```diff
@@ -44,36 +47,56 @@ impl Plugin for Connectors {
             );
         }
 
-        Ok(Connectors { debug_extensions })
+        let max_requests = init
+            .config
+            .max_requests_per_operation_per_source
+            .or(std::env::var(CONNECTORS_MAX_REQUESTS_ENV)
```
nicholascioli, line 53: Do we document anywhere that the env var is second to the config? I'm not sure if it would be obvious to users that it can't be overwritten with an env var.

apollo-router/src/plugins/connectors/request_limit.rs
```diff
@@ -0,0 +1,114 @@
...
+impl RequestLimit {
+    pub(crate) fn new(max_requests: usize) -> Self {
+        Self {
+            max_requests,
+            total_requests: AtomicUsize::new(0),
+        }
+    }
+
+    pub(crate) fn allow(&self) -> bool {
+        self.total_requests.fetch_add(1, Ordering::Relaxed) < self.max_requests
```
nicholascioli, line 61: Genuine question, is `Relaxed` the right concurrency model for this? From my understanding of the [rust nomicon](https://doc.rust-lang.org/nomicon/atomics.html#relaxed), it seems like the value returned might be reordered, thus (maybe) allowing slightly more requests through than we'd expect. 

Is it OK if the amount of requests that we allow through be slightly higher than the specified max?

(I could also be misreading the nomicon. Feel free to ignore if so)

nicholascioli: Oh, I see that in your log function you warn if it goes over the max. That's probably fine then. I'll leave the comment above for reference, but feel free to mark as resolved.

nicholascioli (commented):
This looks great! I left a few small nits and one question around atomic ordering.

apollo-router/src/plugins/connectors/handle_responses.rs
pubmodmatt: This code isn't really modified in this PR, only the indentation level changed, but I can include this.

apollo-router/src/plugins/connectors/handle_responses.rs
pubmodmatt: There's additional refactoring that should be done here, but I held off on it for now. For instance, connectors shouldn't be tied to http at all, that entire body parameter should be generic. I'm looking at doing that as a follow-on.

apollo-router/src/plugins/connectors/plugin.rs
pubmodmatt: The documentation isn't written yet, but I've added this to the issue for writing the documentation.

apollo-router/src/plugins/connectors/request_limit.rs
pubmodmatt: The counter can go over the max (intentionally, so we can track the total number of attempts), but it will never make more requests than the max. That nomicon entry is confusing - what it really means by "using the counter to synchronize any other accesses" are any places where there are additional memory reads or writes that need to be synchronized. Here the result returned by the atomic operation is compared on the current thread to the immutable max value, which doesn't require additional synchronization.

apollo-router/src/plugins/connectors/handle_responses.rs
pubmodmatt: I went ahead and just refactored this here.

PR #5800, opened 2024-08-11
Upgrade to rustls 0.22.4 and aws-lc-rs crypto backend

pubmodmatt:
Update to `rustls` version 22.4, and enable the `aws-lc-rs` crypto backend instead of the default `ring` backend. The `aws-lc-rs` backend is the default crypto backend in version 23.0 and later, but this upgrades to 22.4 to minimize changes for easier backporting. Additional related dependency updates are required for compatibility.

The `aws-lc-rs` crypto backend supports the `ECDSA_P521_SHA512` signature algorithm, which is used by Cloudflare WARP. **This enables the Apollo Router to work with WARP Zero Trust enabled.** Tests will also pass with WARP enabled. 

This additionally enables the use of HTTPS using any of the additional signature algorithms supported by `aws-lc-rs`, for example when sending requests to Connectors and Coprocessors, or through a security gateway using these signature algorithms.

The full list of supported algorithms [can be found here](https://github.com/rustls/rustls/blob/v/0.22.4/rustls/src/crypto/aws_lc_rs/mod.rs#L110-L125).

The new crypto backend uses [AWS libcrypto (AWS-LC)](https://github.com/aws/aws-lc), which uses a [license](https://github.com/aws/aws-lc/blob/main/LICENSE) based on OpenSSL. This is added to the `cargo deny` configuration. The previous `ring` backend used a similar license, according to comments in `deny.toml`.

Additional required packages are installed for CI [on Windows](https://aws.github.io/aws-lc-rs/requirements/windows.html).

**Moving to Draft since this increases build times and duplicates dependencies.** As more dependencies are updated to newer `rustls` versions (particuarly `aws-smithy-runtime`) and once the router moves up to `hyper` 1.0, this should be revisted.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

**Notes**

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -238,10 +238,12 @@ commands:
               - equal: [ *windows_test_executor, << parameters.platform >> ]
           steps:
             - run:
-                name: Install CMake
+                name: Install CMake and LIBCLANG
                 command: |
                   choco install cmake.install -y
                   echo 'export PATH="/c/Program Files/CMake/bin:$PATH"' >> "$BASH_ENV"
+                  choco install llvm -y
+                  echo 'export LIBCLANG_PATH="/c/Program Files/LLVM/bin"' >> "$BASH_ENV"
                   exit $LASTEXITCODE
       - when:
           condition:
@@ -501,6 +503,8 @@ commands:
             # TODO: remove this workaround once we update to Xcode >= 15.1.0 
             # See: https://github.com/apollographql/router/pull/5462
             RUST_LIB_BACKTRACE: 0
+            # See https://aws.github.io/aws-lc-rs/resources.html?highlight=AWS_LC_SYS_NO_ASM#troubleshooting
+            AWS_LC_SYS_NO_ASM: 1
           command: xtask test --workspace --locked --features ci
       - run:
           name: Delete large files from cache
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -522,7 +522,7 @@ dependencies = [
  "humantime",
  "humantime-serde",
  "hyper",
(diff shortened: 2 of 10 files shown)


dylan-apollo:
I think we need to check out compile times before merging this change—CI already takes a _long_ time to wait for, so we need to make sure that adding dependencies doesn't explode that.

Individual runs are tough to measure, since there are so many caching factors that affect them. So maybe we can just time clean builds locally to see how much the difference is?

We might be able to mitigate some by also upgrading other packages that depend on rustls, but seems like we'll need to update to hyper 1.0 as part of that which could be a big lift.

pubmodmatt:
> Individual runs are tough to measure, since there are so many caching factors that affect them. So maybe we can just time clean builds locally to see how much the difference is?

Build time impact is about 9.6%
Binary size impact is 1.5% for debug and 1.9% for release

## Before
* ` cargo build --all-targets  791.87s user 75.90s system 597% cpu 2:25.29 total`
* Debug: `-rwxr-xr-x  1 matt  staff   261M Aug 12 14:18 target/debug/router`
* Release: `-rwxr-xr-x  1 matt  staff   105M Aug 12 14:31 target/release/router`

## After
* `cargo build --all-targets  813.57s user 91.77s system 567% cpu 2:39.66 total`
* Debug: `-rwxr-xr-x  1 matt  staff   265M Aug 12 14:22 target/debug/router`
* Release: `-rwxr-xr-x  1 matt  staff   107M Aug 12 14:27 target/release/router`

PR #5802, opened 2024-08-12
(bug) Trace propagation via header doesn't work 

BrynCooke:
When propagating trace ID via header:

```yaml
  exporters:
    tracing:
      propagation:
        request:
          header_name: "id_from_header"
```

The trace ID was not being correctly transferred to the root span of the request. This resulted in spans being incorrectly attributed to a new trace.

During propagation the trace ID was injected into a remote span with an invalid span ID.
This is then set as the context so that the subsequent root span can pick this up.
However, the root span creation logic checks that the parent span is valid, which it isn't and then created a brand new trace.

The logic is now modified to also check if the trace ID alone is valid.
 

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [x] Integration Tests
    - [x] Manual Tests

**Exceptions**

This is covered by integration tests rather than unit tests as it requires a tracing functionality to be fully set up.
No docs updates required
*Note any exceptions here*

**Notes**


garypen (approved):

.changesets/fix_bryn_remote_spans.md
```diff
@@ -0,0 +1,15 @@
...
+
+The router now correctly propagates trace IDs when using the `propagation.request.header_name` configuration option.
+
+```yaml
+  exporters:
+    tracing:
+      propagation:
+        request:
+          header_name: "id_from_header"
+```
```
bnjjj, line 11: https://github.com/apollographql/router/pull/5845

klippx:
Just for clarification, when not using `propagation` then it works correctly?

PR #5805, opened 2024-08-13
test(federation): added `FetchDependencyGraph::to_dot` method

duckki:
- Generates a string representing the graph in GraphViz dot format.
- Useful for debugging.

It generates diagram like this:

image

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [x] Manual Tests

### DIFF
--- a/apollo-federation/src/query_plan/fetch_dependency_graph.rs
+++ b/apollo-federation/src/query_plan/fetch_dependency_graph.rs
@@ -2270,6 +2270,46 @@ impl std::fmt::Display for FetchDependencyGraph {
     }
 }
 
+// Necessary for `petgraph::dot::Dot` to compile, but executed.
+impl std::fmt::Display for FetchDependencyGraphNode {
+    fn fmt(&self, _f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        unimplemented!()
+    }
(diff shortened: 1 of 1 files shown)


apollo-federation/src/query_plan/fetch_dependency_graph.rs
```diff
@@ -2270,6 +2270,46 @@ impl std::fmt::Display for FetchDependencyGraph {
...
+        unimplemented!()
+    }
+}
+
+// Necessary for `petgraph::dot::Dot::with_attr_getters` calls to compile, but not executed.
+impl std::fmt::Display for FetchDependencyGraphEdge {
+    fn fmt(&self, _f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        unimplemented!()
+    }
+}
```
TylerBloom, line 2285: I understand why this is needed, but I still dislike it. I think rather than panic here, it's probably a better idea to return an `Err`. Unfortunately, `std::fmt::Error` does not contain any data, but I think this is better than an immediate panic.

TylerBloom (approved):
Besides the `unimplemented!()` comment, everything looks good to me.

PR #5810, opened 2024-08-13
validate empty selections (selection: and http: { body: })

lennyburdette:
most of the changes are from making selection valid and prettier changing formatting and therefore line numbers in snapshots

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-federation/src/sources/connect/validation/extended_type.rs
```diff
@@ -214,6 +216,15 @@ fn validate_field(
         )
     });
 
+    if let Some((_, body)) = http_arg
+        .iter()
+        .find(|(name, _)| name == &CONNECT_BODY_ARGUMENT_NAME)
```
dylan-apollo, line 221: Should this be using the `get_http_body_arg` function?

dylan-apollo (approved):
Just one question about an unused function

apollo-federation/src/sources/connect/validation/selection.rs
```diff
@@ -56,6 +57,33 @@ pub(super) fn validate_selection(
...
+    field: &Component<FieldDefinition>,
+    schema: &Schema,
+    selection_node: &Node<Value>,
+) -> Result<(), Message> {
+    let coordinate =
+        connect_directive_http_body_coordinate(&connect_directive.name, parent_type, &field.name);
+
+    let selection_str = require_value_is_str(&selection_node, &coordinate, &schema.sources)?;
+
+    if selection_str.is_empty() {
```
pubmodmatt, line 72: Is something before this removing any whitespace so something like `selection: "   "` will be considered empty?

dylan-apollo: Oh good call, maybe here and in body we should parse JSONSelection and then check if it's empty? Probably means putting in helper method in selection which returns true for _any_ expression which is semantically empty? Since a selection containing only a comment is still empty.

apollo-federation/src/sources/connect/validation/extended_type.rs
lennyburdette: removed, good catch!

apollo-federation/src/sources/connect/validation/selection.rs
lennyburdette: great idea, i added a is_empty() method on JSONSelection instead

pubmodmatt (approved):

PR #5811, opened 2024-08-13
Move QP compatibility checks into constructor and add metric

SimonSapin:
This aligns better with `experimental_query_planner_mode: both_best_effort` falling back to `legacy` when the `QueryPlanner::new` constructor returns an error.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [x] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


SimonSapin:
Draft while we consider adding a new metric for constructor errors

apollo-federation/src/query_plan/query_planner.rs
```diff
@@ -534,6 +540,87 @@ impl QueryPlanner {
...
+                } else {
+                    None
+                }
+            })
+            .flat_map(|object_type| &object_type.fields)
+            .flat_map(|(_, field)| {
+                field
+                    .directives
+                    .iter()
+                    .filter(|d| d.name.as_str() == JOIN_FIELD)
```
goto-bus-stop, line 569: Can we look up the correct name of the directive on the supergraph's `.metadata()`?

SimonSapin: On one hand, probably yes. On the other hand, we have enough occurrences of `"join__` in the code base that I doubt anything works if a supregraph uses import renames.

Maybe composition supports renames when reading subgraphs be never emits them in a supergraph?

apollo-router/src/query_planner/bridge_query_planner.rs
```diff
@@ -975,6 +1004,25 @@ pub(crate) fn metric_query_planning_plan_duration(planner: &'static str, start:
     );
 }
 
+pub(crate) fn metric_rust_qp_init(init_error_kind: Option<&'static str>) {
+    if let Some(init_error_kind) = init_error_kind {
+        u64_counter!(
+            "apollo.router.lifecycle.query_planner.init",
+            "Rust query planner initialisation",
```
goto-bus-stop, line 1011: Nit bc we generally use american
```suggestion
            "Rust query planner initialization",
```

apollo-router/src/query_planner/bridge_query_planner.rs
```diff
@@ -975,6 +1004,25 @@ pub(crate) fn metric_query_planning_plan_duration(planner: &'static str, start:
...
+            1,
+            "init.error_kind" = init_error_kind,
+            "init.is_success" = false
+        );
+    } else {
+        u64_counter!(
+            "apollo.router.lifecycle.query_planner.init",
+            "Rust query planner initialisation",
```
goto-bus-stop, line 1019:
```suggestion
            "Rust query planner initialization",
```

apollo-router/src/query_planner/bridge_query_planner.rs
```diff
@@ -189,10 +194,34 @@ impl PlannerMode {
...
+                    }
+                    _ => metric_rust_qp_init(Some(INTERNAL_INIT_ERROR)),
+                },
+                _ => {
+                    metric_rust_qp_init(Some(INTERNAL_INIT_ERROR));
+                }
+            }
+        }
+
+        metric_rust_qp_init(None);
```
goto-bus-stop, line 223: this should be in an else or we get duplicate metics

goto-bus-stop (changes requested):
I like the single attribute!

One thing we should keep in mind: does this give users a scary warning on startup if they're using default configuration (`both_best_effort`) and new federation features (like prog. override)? It should be clear that they don't need to worry about anything except fed 1 issues, I think.

apollo-router/src/query_planner/bridge_query_planner.rs
lrlna: ooops british spell checker was on

apollo-router/src/query_planner/bridge_query_planner.rs
lrlna: this made me realise we are missing the other two kinds of errors in this match - AggregateFederationError and MultipleFederationErrors. They would have been counted as a "success" 🙈

lrlna:
@goto-bus-stop at the moment we get a `WARN`:
```bash
2024-08-16T13:45:45.605895Z WARN  Failed to initialize the new query planner, falling back to legacy: The supergraph schema failed to produce a valid API schema:  `experimental_query_planner_mode: new` or `both` cannot yet be used with progressive overrides. Remove uses of progressive overrides to try the experimental query planner, otherwise switch back to `legacy` or `both_best_effort`.",
```
Followed by a log saying the reload is complete:
```bash
2024-08-16T13:45:45.777923Z INFO  reload complete
```

Should we soften the wording a bit more in the `falling back to legacy`?

apollo-router/tests/integration/query_planner.rs
```diff
@@ -0,0 +1,443 @@
...
+            "could not create router: \
+             The supergraph schema failed to produce a valid API schema: \
+             Supergraphs composed with federation version 1 are not supported.",
+        )
+        .await;
+    router.assert_shutdown().await;
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn fed1_schema_with_both_best_effort_qp() {
```
lrlna, line 62: added reload to both_best_effort integration tests for all 3 features (fed1, context, progressive overrides) so we can also check that the log differs between setting the config to new and both_best_effort

apollo-router/tests/integration/query_planner.rs
```diff
@@ -0,0 +1,443 @@
+use std::path::PathBuf;
+
+use crate::integration::common::graph_os_enabled;
+use crate::integration::IntegrationTest;
+
+const PROMETHEUS_METRICS_CONFIG: &str = include_str!("telemetry/fixtures/prometheus.router.yaml");
```
lrlna, line 6: added prometheus to the config so we can also check that the metrics are emitted for the reload integration tests.

lrlna:
This is ready for a re-review on Monday @goto-bus-stop. The failing CI is due to a transitive dependency that requires a rustc upgrade.

apollo-router/src/query_planner/bridge_query_planner.rs
```diff
@@ -975,6 +1004,25 @@ pub(crate) fn metric_query_planning_plan_duration(planner: &'static str, start:
...
+            "init.is_success" = false
+        );
+    } else {
+        u64_counter!(
+            "apollo.router.lifecycle.query_planner.init",
+            "Rust query planner initialization",
+            1,
+            "init.is_success" = true
+        );
+    }
```
SimonSapin, line 1023: Is it ok for a metric attribute to be sometimes missing?

goto-bus-stop: I think so. but it could maybe be collapsed to a single attribute (`init.error_kind = "none"`). Though it's the same cardinality either way

SimonSapin:
We can make the warning less scary by downgrading it to `info` level and rewording the message to start with "Falling back to legacy QP" instead of having that part be in the middle of a long message

apollo-router/tests/integration/query_planner.rs
```diff
@@ -0,0 +1,466 @@
...
+    let mut router = IntegrationTest::builder()
+        .config(BOTH_QP)
+        .supergraph("../examples/graphql/local.graphql")
+        .build()
+        .await;
+    router.start().await;
+    router
+        .assert_log_contains(
+            "could not create router: \
+             The supergraph schema failed to produce a valid API schema: \
```
SimonSapin, line 54: The mention of API schema here is not quite correct. It comes from `apollo-router/src/error.rs`:

```
impl From<FederationError> for ServiceBuildError {
    fn from(err: FederationError) -> Self {
        ServiceBuildError::ApiSchemaError(err)
    }
}
```

QP initialization is the only place where this enum variant is used (since API schema generation has moved to `apollo_router::spec::Schema` creation), so it should be renamed to `QueryPlannerInit` or something, and the message reworded

lrlna (approved):

PR #5812, opened 2024-08-13
Upgrade to new visitor traits in validation

dylan-apollo:
Also now detects when a scalar field is selected which isn't defined.

[CNN-387]: https://apollographql.atlassian.net/browse/CNN-387?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apollo-federation/src/sources/connect/validation/selection.rs
```diff
@@ -111,89 +122,175 @@ struct SelectionValidator<'schema> {
...
+
+        match (field_type, is_group) {
+            (ExtendedType::Object(object), true) => {
+                self.check_for_circular_reference(field, object)
+            },
+            (_, true) => {
+                Err(Message {
+                    code: Code::GroupSelectionIsNotObject,
+                    message: format!(
+                        "{coordinate} selects a group `{field_name}{{}}`, but `{parent_type}.{field_name}` is of type `{type_name}` which is not an object.",
```
nicholascioli, line 275: Is the extra `{{}}` needed here?

apollo-federation/src/sources/connect/expand/visitors/input.rs
```diff
@@ -58,17 +58,6 @@ impl FieldVisitor<InputObjectFieldDefinitionPosition>
...
-        &self,
-        group: InputObjectTypeDefinitionPosition,
-    ) -> Result<
-        Vec<InputObjectFieldDefinitionPosition>,
-        <Self as FieldVisitor<InputObjectFieldDefinitionPosition>>::Error,
-    > {
-        let def = group.get(self.original_schema.schema())?;
-        Ok(def.fields.keys().cloned().map(|f| group.field(f)).collect())
-    }
-
```
nicholascioli, line 71: Good call removing this. I originally had it to make it slightly more modular but then only used it in one spot that needed to be implemented anyway.

nicholascioli (approved):
This is great! I'm glad that the visitor allowed for better validation handling.

apollo-federation/src/sources/connect/validation/selection.rs
dylan-apollo: I want to have the braces to better indicate to users what we mean by "group", since that's not a term they'll necessarily be familiar with. So the message would be like "selections a group `user {}`", evoking the selection syntax.

PR #5814, opened 2024-08-13
Normalize subgraph names when generating expanded enum values

dylan-apollo:
This _almost_ completely implements [the JS equivalent](https://github.com/apollographql/federation/blob/c4744da360235d8bb8270ea048f0e0fa5d03be1e/internals-js/src/specs/joinSpec.ts#L18), but I intentionally left out adding the `_` suffix, since it'll update _all_ our snapshots and be messy. It's not obvious to me _why_ that rule was in place initially, and clearly this pattern has been working for us so far, so 🤷.

Implemented with a new type, just so I could easily see all the things this impacts. I reduced/moved a couple clones while updating type signatures, since I was looking.

[CNN-399]: https://apollographql.atlassian.net/browse/CNN-399?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apollo-federation/src/merge.rs
+++ b/apollo-federation/src/merge.rs
@@ -135,16 +135,13 @@ impl Merger {
             .map(|(_, subgraph)| subgraph)
             .collect_vec();
         subgraphs.sort_by(|s1, s2| s1.name.cmp(&s2.name));
-        let mut subgraphs_and_enum_values: Vec<(&ValidFederationSubgraph, Name)> = Vec::new();
+        let mut subgraphs_and_enum_values = Vec::new();
         for subgraph in &subgraphs {
             // TODO: Implement JS codebase's name transform (which always generates a valid GraphQL
             // name and avoids collisions).
-            if let Ok(subgraph_name) = Name::new(&subgraph.name.to_uppercase()) {
-                subgraphs_and_enum_values.push((subgraph, subgraph_name));
-            } else {
-                self.errors.push(String::from(
-                    "Subgraph name couldn't be transformed into valid GraphQL name",
-                ));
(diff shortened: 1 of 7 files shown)


apollo-federation/src/merge.rs
```diff
@@ -135,16 +135,11 @@ impl Merger {
...
-            // TODO: Implement JS codebase's name transform (which always generates a valid GraphQL
-            // name and avoids collisions).
-            if let Ok(subgraph_name) = Name::new(&subgraph.name.to_uppercase()) {
-                subgraphs_and_enum_values.push((subgraph, subgraph_name));
-            } else {
-                self.errors.push(String::from(
-                    "Subgraph name couldn't be transformed into valid GraphQL name",
-                ));
+            match EnumValue::new(&subgraph.name) {
+                Ok(enum_value) => subgraphs_and_enum_values.push((subgraph, enum_value)),
```
pubmodmatt, line 141: Does this need to worry about collisions resulting from the normalization? For example, if two names were the same except for characters that were all converted to "_"?

dylan-apollo: This is only being done for our auto-transformed names, so _hopefully_ that's really unlikely. It'd need to be two subgraphs which have nearly identical names and also have a `@connect` on the same path.

Long term, though, for sure we should also port [this](https://github.com/apollographql/federation/blob/c4744da360235d8bb8270ea048f0e0fa5d03be1e/internals-js/src/specs/joinSpec.ts#L212)

lennyburdette (approved):

PR #5819, opened 2024-08-14
chore(federation): coerce hardcoded GraphQL values to match JS

goto-bus-stop:
This PR does a small amount of GraphQL value coercion to better support mismatch testing with the JS query planner.

Now if you have a schema like this:
```graphql
type Query {
  getUsers(ids: [Int]): [User]
}
```
and you send a query that uses a single literal for the list argument:
```graphql
{ getUsers(ids: 1) }
```
The query planner outputs a subgraph query that does:
```graphql
{ getUsers(ids: [ 1 ]) }
```

It's valid to provide a single value where a list is expected. QP will just translate it to an actual list, similarly to what graphql-js does.

We already have a graphql-js compatibility layer like this for the API schema implementation. This PR reuses some of that code.


apollo-federation/src/compat.rs
```diff
@@ -281,10 +284,133 @@ pub fn coerce_schema_default_values(schema: &mut Schema) {
     }
 }
 
+fn coerce_directive_application_values(
```
lrlna, line 287: fly by thought: this looks slightly different from input coercion we do in apollo-compiler/spec, which is making me think why JS QP ever wanted to do a different a different form of input coercion. i feel like the router and the js qp ended up doing it twice _and_ did it differently 🤔 🤔 ... anyways

lrlna (approved):

dariuszkuc (approved):

duckki:
This does reduce differences in some graphs! 🎉

PR #5820, opened 2024-08-14
Fixed off-by-one bugs in subgraph jump minimizing optimization

TylerBloom:
There were a few off-by-one bugs in the logic of the optimization that determines if we are taking an unnecessary jump between subgraphs. These bugs largely nullified this optimization. For very particular operations, this creates a runaway effect, making the operation effectively infinite.

Fixes ROUTER-550

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-federation/src/query_graph/graph_path.rs
+++ b/apollo-federation/src/query_graph/graph_path.rs
@@ -1642,7 +1642,7 @@ where
                                     .get(root_kind)
                                     .copied()
                             } else {
-                                Some(last_subgraph_entering_edge_tail)
+                                Some(last_subgraph_entering_edge_head)
                             };
 
                             // If the previous subgraph is a federated root, as noted above we take
--- a/apollo-federation/src/query_graph/graph_path.rs
+++ b/apollo-federation/src/query_graph/graph_path.rs
@@ -1051,13 +1051,6 @@ where
                 }
             }
         }
-        let cross = self.graph.is_cross_subgraph_edge(new_edge)?;
-        if cross {
-            last_subgraph_entering_edge_info = Some(SubgraphEnteringEdgeInfo {
-                index: self.edges.len().saturating_sub(1),
-                conditions_cost: condition_cost,
-            });
-        }
 
         if matches!(
             edge_weight.transition,
@@ -1082,6 +1075,12 @@ where
                 edges.push(edge);
                 edge_triggers.push(Arc::new(trigger));
                 edge_conditions.push(condition_path_tree);
+                if self.graph.is_cross_subgraph_edge(new_edge)? {
+                    last_subgraph_entering_edge_info = Some(SubgraphEnteringEdgeInfo {
+                        index: self.edges.len() - 1,
+                        conditions_cost: condition_cost,
+                    });
+                }
                 return Ok(GraphPath {
                     graph: self.graph.clone(),
                     head: self.head,
@@ -1111,12 +1110,12 @@ where
         edges.push(edge);
         edge_triggers.push(Arc::new(trigger));
         edge_conditions.push(condition_path_tree);
-        /*
-        last_subgraph_entering_edge_info = Some(SubgraphEnteringEdgeInfo {
-            index: self.edges.len(),
-            conditions_cost: condition_cost,
-        });
-        */
+        if self.graph.is_cross_subgraph_edge(new_edge)? {
+            last_subgraph_entering_edge_info = Some(SubgraphEnteringEdgeInfo {
+                index: self.edges.len(),
+                conditions_cost: condition_cost,
+            });
+        }
         Ok(GraphPath {
             graph: self.graph.clone(),
             head: self.head,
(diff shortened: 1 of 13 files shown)


apollo-federation/src/query_graph/graph_path.rs
```diff
@@ -3094,7 +3055,7 @@ impl OpGraphPath {
         // account (it may very well be that whatever comes after `u` is not in `A`, for instance).
         let self_tail_weight = self.graph.node_weight(self.tail)?;
         let other_tail_weight = self.graph.node_weight(other.tail)?;
-        if self_tail_weight.source == other_tail_weight.source {
+        if self_tail_weight.source != other_tail_weight.source {
```
TylerBloom, line 3058: This change does not affect the linked ticket. This differs from the JS code and was found while trying to locate the bug that I was looking for.

apollo-federation/src/query_graph/graph_path.rs
```diff
@@ -1077,6 +1078,12 @@ where
                 edges.push(edge);
                 edge_triggers.push(Arc::new(trigger));
                 edge_conditions.push(condition_path_tree);
+                if self.graph.is_cross_subgraph_edge(new_edge)? {
+                    last_subgraph_entering_edge_info = Some(SubgraphEnteringEdgeInfo {
+                        index: self.edges.len() - 1,
+                        conditions_cost: condition_cost,
+                    });
+                }
```
TylerBloom, line 1086: This logic (and the similar statement below) more accurately reflects the JS code. The old logic would have `last_subgraph_entering_edge_info` be `None` if `is_cross_subgraph_edge` returned `false`, but the JS code would not.

apollo-federation/src/query_graph/graph_path.rs
```diff
@@ -1644,7 +1642,7 @@ where
                                     .get(root_kind)
                                     .copied()
                             } else {
-                                Some(last_subgraph_entering_edge_head)
+                                Some(last_subgraph_entering_edge_tail)
```
TylerBloom, line 1645: This is an off-by-one error. `last_subgraph_entering_edge_head` corresponds to a key edge in the `GraphPath`. Key edges in the path's edge triggers are always `OpPathElement::Context`. If a context is iterated over by `check_direct_path_from_node`, that method always returns `None`, which nullifies this optimization.

dariuszkuc (approved):

apollo-federation/src/query_graph/graph_path.rs
dariuszkuc: just wondering should we also set it to `None` otherwise?

TylerBloom: No, it should be `Some`. We are checking to see if the new edge crosses subgraphs. If so, the index needs to be updated. If the new edge doesn't cross a subgraph, we still need to track the info of the last entering edge.

apollo-federation/src/query_graph/graph_path.rs
```diff
@@ -1077,6 +1078,12 @@ where
                 edges.push(edge);
                 edge_triggers.push(Arc::new(trigger));
                 edge_conditions.push(condition_path_tree);
+                if self.graph.is_cross_subgraph_edge(new_edge)? {
```
sachindshinde, line 1081: Note that the JS code (and previous Rust code) checks `defer.is_none()` here.

sachindshinde (changes requested):

apollo-federation/src/query_graph/graph_path.rs
sachindshinde: @TylerBloom
Good catch! We're indeed setting `last_subgraph_entering_edge_info` to `None` when we shouldn't be.

apollo-federation/src/query_graph/graph_path.rs
```diff
@@ -1115,6 +1113,12 @@ where
         edges.push(edge);
         edge_triggers.push(Arc::new(trigger));
         edge_conditions.push(condition_path_tree);
+        if self.graph.is_cross_subgraph_edge(new_edge)? {
```
sachindshinde, line 1116: Similarly, you'll want to check `defer.is_none()` here.

apollo-federation/src/query_graph/graph_path.rs
sachindshinde: @TylerBloom Good catch! It's indeed flipped from the JS code.

apollo-federation/src/query_graph/graph_path.rs
```diff
@@ -1671,8 +1685,8 @@ where
                                             "Edge tail is unexpectedly a federated root",
                                         ));
                                     };
-                                    self.check_direct_path_from_node(
-                                        last_subgraph_entering_edge_info.index + 1,
+                                    to_advance.check_direct_path_from_node(
```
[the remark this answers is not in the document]
sachindshinde: It should indeed be `to_advance` here instead of `self`.

sachindshinde (approved):

PR #5822, opened 2024-08-14
Add ability to reference $config in header values

pubmodmatt:
Add the ability to reference `$config` variables in headers sent to connectors. This allows, for example, specifying a static token in an authorization header.

Config YAML:

```yaml
preview_connectors:
  subgraphs:
    subgraph_name:
      $config:
        api_token: ${env.API_TOKEN}
```

GraphQL Schema:

```graphql
@source(
    name: "jsonPlaceholder"
    http: {
      baseURL: "https://jsonPlaceholder.typicode.com/"
      headers: [
        { name: "Authorization", value: "Bearer {$config.api_token}" }
      ]
    }
  )
```

Note that curly brace characters `{` and `}` are technically allowed within header values. Currently, since the variable reference grammar implemented here only accepts names starting with the `$confg` namespace, this should not cause a problem unless the customer actually wants to send a literal value containing the string `{$config`, which is unlikely. However, if the grammar is expanded in the future, this may become a concern - one possible solution would be to allow escaping the curly brace characters with something like `{{` and `}}`.

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

**Notes**

[CNN-361]: https://apollographql.atlassian.net/browse/CNN-361?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apollo-federation/src/sources/connect/header.rs
+++ b/apollo-federation/src/sources/connect/header.rs
@@ -1,4 +1,4 @@
-//! Headers defined in connectors `@source` directives.
+//! Headers defined in connectors `@source` and `@connect` directives.
 use std::str::FromStr;
 
 use nom::branch::alt;
@@ -48,11 +48,12 @@ impl HeaderValue {
                     let value = vars
                         .get(&var_path_bytes)
                         .ok_or_else(|| format!("Missing variable: {}", var.path))?;
-                    result.push_str(
-                        value
-                            .as_str()
-                            .ok_or_else(|| format!("Variable {} is not a string", var.path))?,
-                    );
+                    let value = if let JSON::String(string) = value {
+                        string.as_str().to_string()
+                    } else {
+                        value.to_string()
+                    };
+                    result.push_str(value.as_str());
                 }
             }
         }
@@ -126,6 +127,8 @@ fn variable_reference(input: &str) -> IResult<&str, &str> {
 
 #[cfg(test)]
 mod tests {
+    use rstest::rstest;
+
     use super::*;
 
     #[test]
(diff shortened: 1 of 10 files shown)


apollo-federation/src/sources/connect/header.rs
```diff
@@ -0,0 +1,279 @@
...
+    fn parse(input: &str) -> IResult<&str, Self> {
+        map(many1(HeaderValuePart::parse), Self::new)(input)
+    }
+
+    /// Replace variable references in the header value with the given variable definitions.
+    ///
+    /// # Errors
+    /// Returns an error if a variable used in the header value is not defined or if a variable
+    /// value is not a string.
+    pub fn interpolate(&self, vars: &Map<ByteString, JSON>) -> Result<String, String> {
```
lennyburdette, line 41: what happens if you try to interpolate an object? like
```yaml
$config:
  foo:
    bar: 1
```
with
```graphql
headers: [{ name: "x-foo" value: "$config.foo" }] # not $config.foo.bar
```

lennyburdette: we can't validate against that because composition isn't aware of router config, so i'm curious what the error message is. 

maybe some unit tests for interpolation here to help cover that?

lennyburdette (approved):
one comment, otherwise looks pretty awesome!

apollo-federation/src/sources/connect/header.rs
pubmodmatt: The error is: `Unable to interpolate header value: "Variable $config.foo is not a string"`, and the header won't be included in the request. I'll add a unit test.

lennyburdette: i dunno why my eyes completely glazed over that a few lines down! what if the value is a JSON number? i can't remember if it's stringified

lennyburdette: or bool, i guess

pubmodmatt: I added code to handle all the types, and unit tests to test.

PR #5823, opened 2024-08-14
fix(federation): use supergraph when parsing requires fieldset

dariuszkuc:
When parsing `@requires` field set selection we need to use supergraph schema instead of a target subgraph schema.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-federation/src/query_graph/extract_subgraphs_from_supergraph.rs
+++ b/apollo-federation/src/query_graph/extract_subgraphs_from_supergraph.rs
@@ -2107,7 +2107,6 @@ fn remove_inactive_applications(
                         .into();
                 // @requires needs to be validated against the supergraph schema
                 (fields, parent_type_pos, supergraph_schema.schema())
-                // (fields, parent_type_pos, schema.schema())
             }
         };
         // TODO: The assume_valid_ref() here is non-ideal, in the sense that the error messages we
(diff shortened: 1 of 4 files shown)


apollo-federation/src/query_graph/extract_subgraphs_from_supergraph.rs
```diff
@@ -2080,15 +2087,15 @@ fn remove_inactive_applications(
...
+        let (fields, parent_type_pos, is_requires_field_set) = match directive_kind {
             FieldSetDirectiveKind::Provides => {
                 let fields = federation_spec_definition
                     .provides_directive_arguments(directive)?
                     .fields;
                 let parent_type_pos: CompositeTypeDefinitionPosition = schema
                     .get_type(field.ty.inner_named_type().clone())?
                     .try_into()?;
-                (fields, parent_type_pos)
+                (fields, parent_type_pos, false)
```
goto-bus-stop, line 2098: I would instead write `(fields, parent_type_pos, &schema)` here and `(fields, parent_type_pos, &supergraph_schema)` below. Then you don't need the if/else to populate `fields` later.

lrlna (changes requested):
Can you add a test for this?

duckki (approved):
The code change and test look good.
I confirmed the test reproduces the issue and the PR eliminates the known errors from the corpus.

lrlna (approved):

PR #5825, opened 2024-08-15
chore(federation): introduce sorted `ArgumentList`, `DirectiveList` types

goto-bus-stop:
This PR introduces two new types: `DirectiveList`, for all directive lists in operation types, and `ArgumentList`, for field arguments. On some queries this PR saves about 3% in planning time, but I wouldn't focus too much on the performance here vs enforcing the sorting in the type system and the ergonomics of being able to use `==` vs. comparison helper functions.

DirectiveList and ArgumentList provide sorting and order-independent equality guarantees not provided by `executable::DirectiveList` and `Vec<Argument>`. We are already using those types like that in the operation module, but we have no compile-time way to enforce those guarantees, and repeatedly re-sort lists when creating selection keys. So this change is meant to enforce guarantees and improve performance by reducing repeated work.

ArgumentList is immutable. When constructed, it recursively sorts arguments and any nested input objects. You can then use `==` to compare two ArgumentLists, and `a: true, b: false` will compare equal to `b: false, a: true`. The original order is not stored: I expect that maintaining argument order or input field order doesn't matter for subgraph operations.

DirectiveList maintains two sort orders for directive applications. The original sort order is stored so it can be maintained in subgraph queries. Additionally a consistent sort order is stored for hashing and comparisons. DirectiveList also recursively sorts all arguments and input values, providing the same guarantees as ArgumentList. This means `==` and hashing are order-independent, but when serializing a directive list, it will use the original order. Note it's possible that the original order is "lost" if two DirectiveLists representing the same applications in different order are stored in a hash map: one will replace the other because they are considered equal.

DirectiveList stores a precomputed hash value because DirectiveLists are repeatedly hashed in selection keys. At least twice (to create the input operation and to create subgraph operations), more than that on average. A nice side benefit of this is that comparing directive lists becomes very cheap on average because we can short-circuit if the cached hash values are not equal.

DirectiveList supports one mutating operation: removing a directive. This is to support removing `@defer`.

Both new types are cheap to clone (`Arc`) and do not allocate when the list is empty. Most selections have no directives and no arguments so this prevents having many small unused allocations in the common case.


apollo-federation/src/operation/mod.rs
```diff
@@ -1082,8 +1172,6 @@ mod field_selection {
     pub(crate) struct Field {
         data: FieldData,
         key: SelectionKey,
-        #[serde(serialize_with = "crate::display_helpers::serialize_as_debug_string")]
-        sorted_arguments: Arc<Vec<Node<executable::Argument>>>,
```
goto-bus-stop, line 1086: This is a key motivator for the change: not requiring this additional field in the `Field` struct means we can generalise/simplify selection key structures in the future.

apollo-federation/src/operation/mod.rs
```diff
@@ -2596,7 +2666,9 @@ impl SelectionSet {
     /// Removes the @defer directive from all selections without removing that selection.
     fn without_defer(&mut self) {
         for (_key, mut selection) in Arc::make_mut(&mut self.selections).iter_mut() {
-            Arc::make_mut(selection.get_directives_mut()).retain(|dir| dir.name != name!("defer"));
+            // TODO(@goto-bus-stop): doing this changes the key of the selection!
+            // We have to rebuild the selection map.
```
goto-bus-stop, line 2670: I didn't fix this here but this is actually a bug in `without_defer`: it creates invalid outputs. `get_directives_mut()` is not a method that should exist on selection types because directives affect the key, and you're not meant to be able to change things about the selection that affect the key.

apollo-federation/src/operation/directive_list.rs
```diff
@@ -0,0 +1,410 @@
...
+use apollo_compiler::Node;
+
+use super::sort_arguments;
+
+/// Compare sorted input values, which means specifically establishing an order between the variants
+/// of input values, and comparing values for the same variants accordingly.
+///
+/// Note that Floats and Ints are compared textually and not parsed numerically. This is fine for
+/// the purposes of hashing.
+fn compare_sorted_value(left: &executable::Value, right: &executable::Value) -> std::cmp::Ordering {
```
goto-bus-stop, line 21: The `compare_*` functions are unchanged, just moved into this module.

apollo-federation/src/operation/contains.rs
```diff
@@ -11,202 +7,6 @@ use super::InlineFragmentSelection;
 use super::Selection;
 use super::SelectionSet;
 
-/// Compare two input values, with two special cases for objects: assuming no duplicate keys,
-/// and order-independence.
-///
-/// This comes from apollo-rs: https://github.com/apollographql/apollo-rs/blob/6825be88fe13cd0d67b83b0e4eb6e03c8ab2555e/crates/apollo-compiler/src/validation/selection.rs#L160-L188
-/// Hopefully we can do this more easily in the future!
-fn same_value(left: &executable::Value, right: &executable::Value) -> bool {
```
goto-bus-stop, line 19: `same_value` and `same_arguments` are removed. `==` on an ArgumentList or DirectiveList does the same thing now.

apollo-federation/src/operation/mod.rs
lrlna: good to know for when we do defer!

lrlna (approved):

apollo-federation/src/operation/mod.rs
goto-bus-stop: Well, this is actually more of a problem while we don't have defer 😛

PR #5834, opened 2024-08-16
Fix session counting and the reporting of file handle shortage

garypen:
Session counting incorrectly included connections to the health check or other non-graphql connections. This is now corrected so that only connections to the main graphql port are counted.

Warnings about file handle shortages are now handled correctly as a global resource.

Note: This was manually tested since integration/unit testing isn't really viable.


BrynCooke (approved):
Approved, but put in the description of the PR that this was manually tested and cannot be unit tested.

PR #5840, opened 2024-08-19
chore(rust): update to rustc 1.80.1

lrlna:
We haven't update rustc in a while, let's see if switching to 1.80.1 will work for us.

Closes https://github.com/apollographql/router/pull/5841

### DIFF
--- a/fuzz/subgraph/src/model.rs
+++ b/fuzz/subgraph/src/model.rs
@@ -127,6 +127,7 @@ impl User {
     }
 }
 
+#[allow(static_mut_refs)]
 fn users() -> &'static [User] {
     static mut USERS: Vec<User> = vec![];
     static INIT: Once = Once::new();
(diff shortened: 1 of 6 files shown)


goto-bus-stop:
Oh, I opened https://github.com/apollographql/router/pull/5841 because I *think* we pinned to 1.76 because of a deno issue. if we *can* safely update, we should definitely do this instead :)

goto-bus-stop:
This is what I was thinking of: https://github.com/apollographql/federation-rs/pull/483

fuzz/subgraph/src/model.rs
```diff
@@ -127,6 +127,7 @@ impl User {
     }
 }
 
+#[allow(static_mut_refs)]
```
SimonSapin, line 130: Why not use `OnceLock` instead? This could also removes the need for unsafe code

lrlna:
I didn't know it was only in `--release` mode :((( thought that showed in with regular cargo build/check
Screenshot 2024-08-19 at 15 28 23

welppp closing this then. @goto-bus-stop your PR with pinning rowan should work for now

PR #5847, opened 2024-08-19
Revert some IndexMap/Set changes that break `--release` builds for x86

dylan-apollo:


lennyburdette (approved):

sachindshinde (commented):
Note that not everything was a `HashMap`/`HashSet` in `apollo-federation/src/query_graph/extract_subgraphs_from_supergraph.rs` prior to https://github.com/apollographql/router/pull/5827 . If I revert just the changes to that file (and `cost_spec_definition.rs`), it's able to compile for me (not sure if it works for y'all).

sachindshinde (approved):

duckki:
I wish I could understand what went wrong with the HashMaps/Sets.

BTW, why don't we push to `dev`? Wouldn't it create more divergence between dev and next?

dylan-apollo:
> BTW, why don't we push to `dev`? Wouldn't it create more divergence between dev and next?

@duckki Backported in https://github.com/apollographql/router/pull/5854

There's another, bigger fix coming in https://github.com/apollographql/router/pull/5853 which I will also backport once we're unblocked

PR #5848, opened 2024-08-19
Remove actix dependencies by switching fuzz subgraph to axum

dylan-apollo:
Hopefully this will speed up CI since there will be fewer crates to download. Not sure if it works as intended, since I'm not sure what the _expected_ result of running the fuzz test is.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -25,242 +25,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "actix"
-version = "0.13.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


Geal:
great idea! have you tested that that subgraph still runs? The point of that service is that it can serve the entire graph that we often use in tests (accounts/products/reviews/inventory) but also act as a subgraph with any subset of the types. So it can be used to test the planner, with the invariant that calling it directly, or through the router with the graph cut into any amount of subgraphs, should return the same response

Geal (approved):

bnjjj (approved):

dylan-apollo:
> have you tested that that subgraph still runs?

Yup! I ran some basic queries against it that look good.

PR #5850, opened 2024-08-20
feat: ease of use debugging

### LINKED ISSUE #5789: ease of use: debug mode
Currently, if we wanted to use debug (heaptrack) mode on kubernetes, we have few issues:
- the file it writes to, doesn't have any identifier, so if there are multiple pods, they override the same file
- There's no easy way to specify that a container shouldn't be restarted inside the pod. (so that we get new pod if the current pod crashes)

- writing our own charts and forking dockerfile ourselves, but this is unreasonable

If you guys are okay to change this, I'd like to send a PR that:
- let's user specify [pod lifecycle](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/)
- use `${hostname}` in the heaptrack file path

```[tasklist]
### Tasks
- [ ] add ability to specify pod lifecycle
- [ ] use ${hostname} in the filepath when using debug mode
```

cyberhck:
When we want to use debug mode, we can only get 1 file per node in Kubernetes, this change allows us to do 2 things:
- set `restartPolicy` of pod as desired.
- add hostname in path for collecting healtrack data.

Fixes #5789

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/dockerfiles/Dockerfile.router
+++ b/dockerfiles/Dockerfile.router
@@ -65,7 +65,7 @@ RUN \
 \nset -e \
 \n \
 \nif [ -f "/usr/bin/heaptrack" ]; then \
-\n    exec heaptrack -o /dist/data/router_heaptrack /dist/router "$@" \
+\n    exec heaptrack -o /dist/data/router_heaptrack/${hostname} /dist/router "$@" \
 \nelse \
 \n    exec /dist/router "$@" \
 \nfi \
--- a/dockerfiles/Dockerfile.router
+++ b/dockerfiles/Dockerfile.router
@@ -65,7 +65,7 @@ RUN \
 \nset -e \
 \n \
 \nif [ -f "/usr/bin/heaptrack" ]; then \
-\n    exec heaptrack -o /dist/data/router_heaptrack/${hostname} /dist/router "$@" \
+\n    exec heaptrack -o /dist/data/$(hostname)/router_heaptrack  /dist/router "$@" \
 \nelse \
 \n    exec /dist/router "$@" \
 \nfi \
--- a/.changesets/config_musket_minute_linen_kitchen.md
+++ b/.changesets/config_musket_minute_linen_kitchen.md
@@ -1,0 +1,5 @@
+### Include hostname on heaptrack path, specify pod lifecycle ([Issue #5789](https://github.com/apollographql/router/issues/5789))
+
+Use hostname in the heaptrack path to identify an individual container/instance/machine where the router is running. Additionally, allow the specification of restartPolicy on deployment (defaults to `Always` which is kubernetes default)
+
+By [@cyberhck](https://github.com/cyberhck) in https://github.com/apollographql/router/pull/5850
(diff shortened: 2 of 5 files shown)


cyberhck:
@garypen let me know if there's any more change you would like.

cyberhck:
@garypen @BrynCooke please let me know anything else that I need to do to get this merged.

cyberhck:
let me know if you'd like me to rebase this branch with latest dev branch.

dockerfiles/Dockerfile.router
```diff
@@ -63,7 +63,7 @@ RUN \
 \nset -e \
 \n \
 \nif [ -f "/usr/bin/heaptrack" ]; then \
-\n    exec heaptrack -o /dist/data/router_heaptrack /dist/router "$@" \
+\n    exec heaptrack -o /dist/data/router_heaptrack/${hostname} /dist/router "$@" \
```
garypen, line 66: This looks to be reading a variable called hostname which I don't think is a standard on linux. You probably mean to use `$(hostname)`. Also, the last element of the path is the actual filename of the file, so I think you have your change in the wrong place in the path. With your change `router_heaptrack` is added as a directory in the path and the filename of the output file changes from a fixed name `router_heaptrack` to a variable name (the hostname).

Has this change been tested in your environment?

I think your change should be:

```
\n exec heaptrack -o /dist/data/$(hostname)/router_heaptrack.$(hostname)  /dist/router "$@" \
```
This tags the hostname as part of the filename. I'm using a `.` as a separator, it could equally well be a `_`.

or:

```
\n exec heaptrack -o /dist/data/$(hostname)/router_heaptrack  /dist/router "$@" \
```
This produces a file with the same name as currently, under a new sub-directory which is the hostname of the executing system. Alternatively, it could be incorporated into the file name as follows:

Note: `heaptrack` appears to provide a %h variable which looks like it would do what you are looking for. However, I've found that to be a bit unreliable, so using `$(hostname)` is safer...

garypen (changes requested):
Thanks for your contribution, it's much appreciated.

If we make this change, it may be a little surprising to users, since their output heaptrack data files will now have a hostname in them. I don't think that's a big deal, since we aren't offering any guarantees about the stability of output debug file names. However, I would like this to be called out explicitly in the changelog.

Once you decide exactly which format of output file name to use, please update the changelog so that it has an example that illustrates the impact of the change.

dockerfiles/Dockerfile.router
cyberhck: Thank you so much for the review, much appreciated, will update as soon as I get some time.

cyberhck:
@garypen does this look better? Sorry about the text in changeset, English is my second language. But if you suggest something, I'll use it.

garypen:
> @garypen does this look better? Sorry about the text in changeset, English is my second language. But if you suggest something, I'll use it.

Looks great! Thanks for pushing this forward.

garypen (approved):
Thanks very much for the contribution!

cyberhck:
@garypen not sure why CI is failing, anything I can do to help?

garypen:
> @garypen not sure why CI is failing, anything I can do to help?

It's an issue relating to a test which won't work for external contributions. I'm going to merge your PR without waiting for the tests to pass. It's a change that won't impact any of our tests anyway...

PR #5854, opened 2024-08-20
Revert some IndexMap/Set changes that break `--release` builds for x86

dylan-apollo:
Backport of https://github.com/apollographql/router/pull/5847


dylan-apollo:
Actually we don't need this anymore, since we made a bigger fix

PR #5858, opened 2024-08-21
Add a subgraph request id

Geal:
This is a unique string identifying a subgraph request and response, allowing plugins and coprocessors to keep some state per subgraph request by matching on this id. It is available in coprocessors as `subgraphRequestId` and rhai scripts as `request.subgraph.id` and `response.subgraph.id`.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-658]: https://apollographql.atlassian.net/browse/ROUTER-658?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bnjjj:
As every request is executed in a new span, can we just try to share the span id instead of creating a new id ? I'm just wondering, I don't know if it's something we can do, we also have to make sure we still have an id if the span is not sampled (I think yes).

apollo-router/tests/fixtures/request_response_test.rhai
```diff
@@ -90,6 +90,10 @@ fn process_subgraph_request(request) {
     process_common_request(true, request);
     // subgraph doesn't have a context member
     process_common_request(false, request.subgraph);
+
+    if request.subgraph_request_id = () {
```
Geal, line 94:
```suggestion
    if request.subgraph_request_id == () {
```

apollo-router/tests/fixtures/request_response_test.rhai
```diff
@@ -185,6 +189,9 @@ fn process_subgraph_response(response) {
     process_common_response(response);
     test_response_body(response);
     test_response_status_code(response);
+     if response.subgraph_request_id = () {
```
Geal, line 192:
```suggestion
     if response.subgraph_request_id == () {
```

apollo-router/src/services/subgraph.rs
```diff
@@ -167,6 +175,8 @@ pub struct Response {
     /// Name of the subgraph, it's an Option to not introduce breaking change
     pub(crate) subgraph_name: Option<String>,
     pub context: Context,
+    /// unique id matching the corresponding field in the request
+    pub(crate) id: String,
```
Geal, line 179: make it a new type (Id, etc)

apollo-router/src/services/subgraph_service.rs
```diff
@@ -807,7 +809,7 @@ fn http_response_to_graphql_response(
 pub(crate) async fn process_batch(
     client_factory: HttpClientServiceFactory,
     service: String,
-    mut contexts: Vec<Context>,
+    mut contexts: Vec<(Context, String)>,
```
Geal, line 812: maybe a newtype here too

apollo-router/src/services/subgraph_service.rs
```diff
@@ -1014,9 +1017,10 @@ pub(crate) async fn process_batch(
                 .map(|mut http_res| {
                     *http_res.headers_mut() = parts.headers.clone();
                     // Use the original context for the request to create the response
-                    let context = contexts.pop().expect("we have a context for each response");
+                    let (context, id) =
+                        contexts.pop().expect("we have a context for each response");
```
Geal, line 1021: needs a test around batching

ashkan25:
@Geal Hi! I've been playing around with the Router with your changes included from this PR but I can't seem to get the `subgraphRequestId` to show up in the coprocessor `subgraphresponse` stage. I've enabled it for both subgraph requests and responses in the router yaml file using: `subgraph_request_id: true`. I can see the new id in the `SubgraphRequest` stage, but not `SubgraphResponse`. Is it possible I'm missing something on my side or is there still work to be done for this PR?

apollo-router/src/services/subgraph.rs
Geal: fixed in 71598607f3e302dae6aef3ddc798488f71e070ef

Geal:
@ashkan25 you're right, it was missing, this is fixed in https://github.com/apollographql/router/pull/5858/commits/d3f1ee6695b8a45852a9eedc606c429c0f66f6bf

apollo-router/src/services/subgraph_service.rs
Geal: batching test added in [98a4ba3](https://github.com/apollographql/router/pull/5858/commits/98a4ba3004a31808a18df7d55d2a46431d5473e8)

abernix:
@Geal At your leisure, what would it take to give that last commit an accompanying test for coprocessors?

apollo-router/src/services/subgraph.rs
```diff
@@ -220,10 +267,16 @@ impl Response {
 
         *response.headers_mut() = headers.unwrap_or_default();
 
+        // Warning: the id argument forthis builder is an Option to make that a non breaking change
```
bnjjj, line 270:
```suggestion
        // Warning: the id argument for this builder is an Option to make that a non breaking change
```

docs/source/customizations/coprocessor.mdx
```diff
@@ -504,6 +504,7 @@ Properties of the JSON body are divided into two high-level categories:
   "stage": "SubgraphRequest",
   "control": "continue",
   "id": "666d677225c1bc6d7c54a52b409dbd4e",
+  "subgraphRequestId": "b5964998b2394b64a864ef802fb5a4b3",
```
bnjjj, line 507: Could we show an example of yaml configuration to include that key ?

bnjjj (approved):

docs/source/customizations/coprocessor.mdx
Geal: [a07cafb](https://github.com/apollographql/router/pull/5858/commits/a07cafb7b1aa916236be15ed962b5e0873674e8f)

PR #5859, opened 2024-08-21
fix(subgraph_service): when the subgraph connection is closed or in error, return a proper subgraph response

bnjjj:
When the subgraph connection is closed or in error, return a proper subgraph response containing an error. This was preventing subgraph response service to be triggered in coprocessor and rhai.

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [x] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


Geal (approved):
this looks ok, but I wonder if we should instead have a wrapper in Service::call to convert Err to a subgraph response every time, to make sure we won't see that issue again. Or maybe make sure that `call_single_http` and others do not return a `Result` but a graphql response

bnjjj:
@Geal IMHO there is still bad things that can happen and should still appear in `map_err`, here the situation is different because it's directly related to the subgraph call. But we could have a critical internal error that should be treated as a real error and so IMHO we should keep Result anywhere else.

Geal:
yup. as long as everything that gets out of the subgraph service is a graphql response, it will be fine

garypen (commented):
Looks good.

PR #5861, opened 2024-08-21
chore(tests): use fed2 supergraphs in tests

lrlna:
There are a lot of tests using Federation1 instead of Federation2. This PR updates almost all of them (except for the ones that are actually checking for core schema functionality) to use Fed2. 

There are a few instances where we had files like:
* supergraph.graphql
* supergraph_fed2.graphql

This implied that `fed1` is the default, where as `fed2` is an addition.

I swapped these around with the changes in this PR:
* supergraph.graphql will always mean it's a fed2 supergraph
* supergraph_fed1.graphql is an additional fed1 supergraph we might need for certain test cases. 

This PR leaves `@core` tests in `apollo-federation` for the time being as the functionality there sometimes exists and needs to be tested.


goto-bus-stop (dismissed):
2 small things:
- <strike>`dump.rdb` file in the root is probably not intended to be commited</strike>
- `examples/graphql/supergraph-fed2.graphql` is moved to `supergraph-schema.graphql` in the root; probably by accident?

bnjjj (approved):

apollo-router/tests/integration/redis.rs
```diff
@@ -23,10 +23,11 @@ use crate::integration::IntegrationTest;
...
     // If this test fails and the cache key format changed you'll need to update the key here.
     // 1. Force this test to run locally by removing the cfg() line at the top of this file.
-    // 2. run `docker compose up -d` and connect to the redis container by running `docker-compose exec redis /bin/bash`.
+    // 2. run `docker-compose up -d` and connect to the redis container by running `docker-compose exec redis /bin/bash`.
     // 3. Run the `redis-cli` command from the shell and start the redis `monitor` command.
     // 4. Run this test and yank the updated cache key from the redis logs.
-    let known_cache_key = "plan:0:v2.8.3:16385ebef77959fcdc520ad507eb1f7f7df28f1d54a0569e3adabcb4cd00d7ce:3973e022e93220f9212c18d0d0c543ae7c309e46640da93a4a0314de999f5112:3106dfc3339d8c3f3020434024bff0f566a8be5995199954db5a7525a7d7e67a";
+    // If you have redis running locally, you can skip step 1 and 2, and proceed with steps 3-4.
+    let known_cache_key = "plan:0:v2.8.3:70f115ebba5991355c17f4f56ba25bb093c519c4db49a30f3b10de279a4e3fa4:3973e022e93220f9212c18d0d0c543ae7c309e46640da93a4a0314de999f5112:4f9f0183101b2f249a364b98adadfda6e5e2001d1f2465c988428cf1ac0b545f";
```
Geal, line 30: the cache key changes because:
- first hash is the schema hash
- schema id is part of the last hash

lrlna: but it does make sense that it'd change if the schema changed, right? or it shouldn't have happened?

Geal: it makes sense yes. Just pointing out for context why the key changes (we should check that everything is ok every time that key changes)

lrlna:
> examples/graphql/supergraph-fed2.graphql is moved to supergraph-schema.graphql in the root; probably by accident?

@goto-bus-stop part of it was accident (adding it to root), but the rename was intentional - supergraph-schema.graphql is the default fed2 schema. I added an explanation to the description as well.

lrlna:
The snapshot changes are a bit strange, and I didn't yet want to accept them. I'll take a look a bit more tomorrow, but hopefully that's the last of the failing tests.

SimonSapin (approved):

PR #5862, opened 2024-08-21
refactor: apply_selection once per connector

lennyburdette:
this is a minor optimization, but for entity requests the selection is always the same so we shouldn't transform the selection mapping for each request

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/plugins/connectors/handle_responses.rs
+++ b/apollo-router/src/plugins/connectors/handle_responses.rs
@@ -68,12 +68,7 @@ pub(crate) async fn handle_responses<T: HttpBody>(
                     })?;
 
                     let mut res_data = {
-                        // TODO: caching of the transformed JSONSelection with the selection set applied?
-                        let transformed_selection = connector
-                            .selection
-                            .apply_selection_set(response_key.selection_set());
-
-                        let (res, apply_to_errors) = transformed_selection.apply_with_vars(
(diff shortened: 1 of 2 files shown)


apollo-router/src/plugins/connectors/handle_responses.rs
```diff
@@ -68,12 +68,7 @@ pub(crate) async fn handle_responses<T: HttpBody>(
                     })?;
 
                     let mut res_data = {
-                        // TODO: caching of the transformed JSONSelection with the selection set applied?
```
pubmodmatt, line 71: I was thinking more broadly with this comment, that there's opportunity here to cache the transformed selection across GraphQL queries. Queries with the same selection set (which already implements Hash) could reuse the cached JSONSelection. So there might be more that can be done here, but that's outside the scope of this PR.

pubmodmatt (approved):
Nice improvement!

apollo-router/src/plugins/connectors/handle_responses.rs
lennyburdette: agreed, i'll leave a comment on the existing issue about how this improvement is related but not the same

PR #5864, opened 2024-08-22
fix(federation): remove redundant typename for inline fragments

dariuszkuc:
When adding `__typename` field for abstract types, we need to check for inline fragment type condition and not the parent selection set. If we have an inline fragment without type condition and it is pointing to and abstract type, then the required `__typename` field should have already been added to its parent field.

We continue adding `__typename` for fragments with a type condition as their type condition _may_ be different than the parent type.

```graphql
foo {  
  # if this is abstract type we will add __typename
  a
  ... @skip(if: false) {  
    # if this fragment is on abstract type we don't need to add anything as
    # we already added the __typename to the parent selection
    b
  }
}
```

 
---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-federation/src/operation/mod.rs
+++ b/apollo-federation/src/operation/mod.rs
@@ -895,7 +895,19 @@ impl Selection {
     }
 
     fn sub_selection_type_position(&self) -> Option<CompositeTypeDefinitionPosition> {
-        Some(self.selection_set()?.type_position.clone())
+        match self {
+            Selection::Field(field_selection) => field_selection
+                .selection_set
+                .as_ref()
+                .map(|s| s.type_position.clone()),
+            Selection::FragmentSpread(fragment_selection) => {
+                Some(fragment_selection.spread.type_condition_position.clone())
+            }
+            Selection::InlineFragment(inline_fragment_selection) => inline_fragment_selection
+                .inline_fragment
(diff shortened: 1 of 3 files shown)


apollo-federation/src/operation/mod.rs
```diff
@@ -895,7 +895,19 @@ impl Selection {
     }
 
     fn sub_selection_type_position(&self) -> Option<CompositeTypeDefinitionPosition> {
```
goto-bus-stop, line 897: This function now doesn't return the subselection's type position. I think we should rename it or inline the function where it's used (as it's only used in one place).

goto-bus-stop (changes requested):
Sweet! 🥳

goto-bus-stop (approved):
approved assuming comment addressed : )

apollo-federation/src/operation/mod.rs
dariuszkuc: Decided to inline it to contain to just that single function call. Otherwise it would have to be something like `field_sub_selection_type_position_or_fragment_type_condition` or something along those lines.

PR #5865, opened 2024-08-22
add a JSON manipulation plugin stage

Geal:
***Note**: I am asking for an early review of this before going through all the relevant API, docs, test work to make this production ready, so we can discuss first if the pipeline changes described below make sense, or if some of them should be moved elsewhere*

this stage sits between the router stage and supergraph stage, and allows request and response manipulation in JSON format, without needing explicit buffering and deserialization at the router stage, and working with data that does not serialize cleanly to GraphQL requests and response objects.

This change will only add the new stage with a private plugin part. The public plugin, rhai and coprocessor usage will come as a follow up.

**Notable changes:**
- batching is now handled at the JSON stage. This is a bit easier to handle since it is directly working on JSON values. Batching has been updated to use `serde_json_bytes::Value` natively instead of the similar type from `serde_json`
- operations transmitted in the query string are transformed to a JSON body at the router service level, to make sure that JSON service plugins are always working with JSON bodies (in the future, headers would not be relevant anymore at that stage)
- the multipart protocol needs to work on JSON values now instead of GraphQL response objects. This makes it a bit more brittle
- APQ and persisted queries are now executed in the JSON service. They could be moved to a JSON service plugin
- query analysis now runs in the JSON service. That means that the query will be modifiable by users (once we add a plugin/rhai/coprocessor step) before going through validation, without having to deserialize and reserialize a JSON request
- this will probably have some latency overhead, because instead of going directly between bytes and GraphQL request/response objects, we go through a JSON value. Probably low overhead on the request side, and we should check on the response side if we can reduce it, because most of the data will already be in the form of a JSON value

**Questions that we should ask:**
- are the different tasks performed at the right point? (especially query string parsing and batching)
- which plugins should be moved from supergraph service to JSON service?
- what do we need in the request and response API at that level?
- the fact that this passes existing tests is a good sign, but are there more tests that we culd add to validate the interaction between router service, JSON service and supergraph service?
- is `JsonServerService` the right name for it?

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-629]: https://apollographql.atlassian.net/browse/ROUTER-629?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apollo-router/src/protocols/multipart.rs
```diff
@@ -104,9 +104,17 @@ impl Stream for Multipart {
...
+                    let is_still_open = response
+                        .as_object()
+                        .and_then(|o| o.get("hasNext").and_then(|v| v.as_bool()))
+                        .unwrap_or(false)
+                        || response
+                            .as_object()
+                            .and_then(|o| o.get("payload"))
+                            .and_then(|v| v.as_object())
+                            .map(|o| !o.is_empty())
+                            .unwrap_or(false);
```
bnjjj, line 117: I think it won't work, it's more complicated than this IIRC

apollo-router/src/protocols/multipart.rs
```diff
@@ -116,26 +124,24 @@ impl Stream for Multipart {
...
+                                    .as_object()
+                                    .and_then(|o| o.get("payload"))
+                                    .and_then(|v| v.as_object())
+                                    .is_none()
+                                && response
+                                    .as_object()
+                                    .and_then(|o| o.get("errors"))
+                                    .and_then(|v| v.as_array())
+                                    .map(|v| v.is_empty())
+                                    .unwrap_or(true)
```
bnjjj, line 139: same here. I think we make a difference between `Some(Null)` and `None`

apollo-router/src/protocols/multipart.rs
```diff
@@ -216,7 +222,16 @@ mod tests {
...
-        let gql_responses = stream::iter(responses);
+        let gql_responses = stream::iter(responses).map(|mut response| {
+            let r = SubscriptionPayload {
+                errors: response.errors.drain(..).collect(),
+                payload: match response.data {
+                    None | Some(Value::Null) if response.extensions.is_empty() => None,
+                    _ => response.into(),
+                },
+            };
+            serde_json_bytes::to_value(r).unwrap()
```
bnjjj, line 233: Maybe use expect

bnjjj (changes requested):

apollo-router/src/protocols/multipart.rs
Geal: do you see cases where it would fail, that I could add to unit tests?

PR #5869, opened 2024-08-22
support application/x-www-form-urlencoded request bodies

lennyburdette:
if a connector is configured to send the content-type: application/x-www-form-urlencoded header, the JSON request body (after JSON selection mapping) is converted to a form-encoded string.

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [x] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/plugins/connectors/http_json_transport.rs
```diff
@@ -77,43 +78,72 @@ pub(crate) fn make_request(
...
+        == Some("application/x-www-form-urlencoded");
+
+    let (json_body, form_body, body, apply_to_errors) = if let Some(ref selection) = transport.body
+    {
         let (json_body, apply_to_errors) = selection.apply_with_vars(&json!({}), &inputs);
+        let mut form_body = None;
         let body = if let Some(json_body) = json_body.as_ref() {
-            hyper::Body::from(serde_json::to_vec(json_body)?)
+            if is_form_urlencoded {
+                let encoded = encode_json_as_form(json_body.as_object().unwrap());
```
dylan-apollo, line 104: Is this guaranteed by some other piece of code to never not be an object? I feel like we'd want to return an error, not panic.

apollo-router/src/plugins/connectors/http_json_transport.rs
```diff
@@ -77,43 +78,72 @@ pub(crate) fn make_request(
...
+        request,
+        original_request.supergraph_request.headers(),
+        &transport.headers,
+        &flat_inputs,
+    );
+
+    let is_form_urlencoded = content_type
+        .as_ref()
+        .map(|v| v.to_str().unwrap_or_default())
+        == Some("application/x-www-form-urlencoded");
```
dylan-apollo, line 96: Does `http` give us a way to compare the semantic value here? Like, if they wanted to pass some extra data after a `;`, I believe we should ignore that.

apollo-router/src/plugins/connectors/form_encoding.rs
```diff
@@ -0,0 +1,155 @@
+use itertools::Itertools;
+use serde_json_bytes::ByteString;
+use serde_json_bytes::Map;
+use serde_json_bytes::Value;
+
+pub(super) fn encode_json_as_form(value: &Map<ByteString, Value>) -> String {
+    let mut encoded: form_urlencoded::Serializer<String> =
+        form_urlencoded::Serializer::new(String::new());
+
+    fn print_json_value_literal(value: &Value) -> String {
```
dylan-apollo, line 10: I feel like there's a way to avoid all the `to_string` and `format!` by embedding this in `recurse`. I'm gonna play with it and send it to you if it works.

apollo-router/src/plugins/connectors/http_json_transport.rs
lennyburdette: correct! good catch

apollo-router/src/plugins/connectors/http_json_transport.rs
lennyburdette: good question, i'll look around

apollo-router/src/plugins/connectors/http_json_transport.rs
```diff
@@ -77,43 +78,72 @@ pub(crate) fn make_request(
...
+            debug.lock().push_request(
+                &request,
+                json_body.as_ref(),
+                transport.body.as_ref().map(|body| SelectionData {
+                    source: body.to_string(),
+                    transformed: body.to_string(), // no transformation so this is the same
+                    result: json_body.clone(),
+                    errors: apply_to_errors,
+                }),
+            );
```
dylan-apollo, line 145: I think the only differences between these two, down in the new functions, is that the JSON body is converted to a string later than the form body and that "kind" is set. So can we consolidate the functions if we pass in `kind` here and do the json to_string here?

lennyburdette: yeah probably! i'll look into it

apollo-router/src/plugins/connectors/http_json_transport.rs
lennyburdette: turns out the mime crate handles this well

apollo-router/src/plugins/connectors/http_json_transport.rs
lennyburdette: done!

apollo-router/src/plugins/connectors/http_json_transport.rs
```diff
@@ -77,43 +78,74 @@ pub(crate) fn make_request(
...
+
+    // add the headers and if content-type is specified, we'll check that when constructing the body
+    let (mut request, content_type) = add_headers(
+        request,
+        original_request.supergraph_request.headers(),
+        &transport.headers,
+        &flat_inputs,
+    );
+
+    let is_form_urlencoded = content_type.as_ref() == Some(&mime::APPLICATION_WWW_FORM_URLENCODED);
```
dylan-apollo, line 93: `CONTENT_TYPE` is still in `RESERVED_HEADERS` up above. What impact does that have elsewhere?

apollo-router/src/plugins/connectors/http_json_transport.rs
```diff
@@ -77,43 +78,74 @@ pub(crate) fn make_request(
...
+        let mut form_body = None;
         let body = if let Some(json_body) = json_body.as_ref() {
-            hyper::Body::from(serde_json::to_vec(json_body)?)
+            if is_form_urlencoded {
+                let encoded = encode_json_as_form(json_body)
+                    .map_err(HttpJsonTransportError::FormBodySerialization)?;
+                form_body = Some(encoded.clone());
+                hyper::Body::from(encoded)
+            } else {
+                request = request.header(CONTENT_TYPE, mime::APPLICATION_JSON.essence_str());
```
dylan-apollo, line 106: So, if they provided an explicit header which resolves to urlencoded form, we use that exact header. Otherwise, we replace whatever they gave us with `application/json`, right?

Definitely need a validation to make it clear that this will happen.

apollo-router/src/plugins/connectors/http_json_transport.rs
lennyburdette: that's used for `{ name: "x" from: "y" }` — it won't forward the content type from the graphql client. that does actually mean i can remove a tiny bit of code — good catch!

apollo-router/src/plugins/connectors/http_json_transport.rs
lennyburdette: yep that's exactly right. validating values is a good idea, though it'll be possible to bypass static validation with `value: "{$config.some_content_type_we_cant_see}"`

apollo-router/src/plugins/connectors/http_json_transport.rs
lennyburdette: right now it'll emit a warning log, so disallowing might be the right idea. let's file a ticket

lennyburdette: https://apollographql.atlassian.net/browse/CNN-413

dylan-apollo (approved):
:shipit:

PR #5871, opened 2024-08-23
feat(federation): respect the `subgraph_graphql_validation` option

duckki:
- validate subgraph queries only when `subgraph_graphql_validation` configuration option is true.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-667]: https://apollographql.atlassian.net/browse/ROUTER-667?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apollo-federation/src/operation/mod.rs
+++ b/apollo-federation/src/operation/mod.rs
@@ -3955,7 +3955,7 @@ impl TryFrom<Operation> for Valid<executable::ExecutableDocument> {
         document.fragments = fragments;
         document.operations.insert(operation);
         coerce_executable_values(value.schema.schema(), &mut document);
-        Ok(document.validate(value.schema.schema())?)
+        Ok(Valid::assume_valid(document))
     }
 }
 
--- a/apollo-federation/src/operation/mod.rs
+++ b/apollo-federation/src/operation/mod.rs
@@ -3934,7 +3934,7 @@ impl From<&FragmentSpreadSelection> for executable::FragmentSpread {
     }
 }
 
-impl TryFrom<Operation> for Valid<executable::ExecutableDocument> {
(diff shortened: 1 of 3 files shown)


apollo-federation/src/operation/mod.rs
```diff
@@ -3955,7 +3955,7 @@ impl TryFrom<Operation> for Valid<executable::ExecutableDocument> {
         document.fragments = fragments;
         document.operations.insert(operation);
         coerce_executable_values(value.schema.schema(), &mut document);
-        Ok(document.validate(value.schema.schema())?)
+        Ok(Valid::assume_valid(document))
```
goto-bus-stop, line 3958: We should not do `assume_valid` inside the `TryFrom` implementation. If we do this we should change the impl to `TryFrom<Operation> for executable::ExecutableDocument`.

goto-bus-stop (commented):
Tbh, I have a slight preference for always validating, and returning an internal error if we produced an invalid query.

duckki:
JS QP doesn't seem to validate subgraph queries. So, Rust QP is taking a performance penalty (didn't measure how much). Also, this validation has never failed so far over the corpus.

duckki:
This is not urgent. We can revisit this when we work on performance.

abernix:
Closing this to clean-up some open and aging PRs, but don't hesitate to just click the re-open button it if it's still relevant.

PR #5875, opened 2024-08-23
Add a histogram metric tracking evaluated query plans

Geal:
the `max_evaluated_plans` option can be used to limit the number of query plans evaluated for a query, but there was no way to know what is a reasonable value for that option. This adds a histogram metric tracking the number of evaluated query plans, which should give more context to configure this.

~This cannot be merged until https://github.com/apollographql/federation-rs/pull/566 is merged and router-bridge published~ (This is done as `v0.6.1+v2.9.0`)

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-668]: https://apollographql.atlassian.net/browse/ROUTER-668?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apollo-router/src/query_planner/bridge_query_planner.rs
```diff
@@ -554,13 +556,20 @@ impl BridgeQueryPlanner {
...
+                        evaluated_plan_count,
                     },
                 mut usage_reporting,
             } => {
                 if let Some(sig) = operation_signature {
                     usage_reporting.stats_report_key = sig;
                 }
 
+                u64_histogram!(
+                    "apollo.router.query_planning.evaluated_plans",
```
lrlna, line 568: would it make sense to put `evaluated_plans` under `plan` separator as well? in the same we have `apollo.router.query_planning.plan.duration`? so `apollo.router.query_planning.plan.evaluated_plans`

lrlna (commented):
could this have some tests too? 😇

apollo-router/src/query_planner/bridge_query_planner.rs
Geal: sure!

lrlna (approved):

PR #5879, opened 2024-08-23
Refactor subscriptions to go through FetchService

pubmodmatt:
The FetchService wraps the SubgraphService and ConnectorService. A request to the FetchService determines which subservice is appropriate to call, and makes this decision transparent to callers.

However, subscriptions, which are not currently supported with connectors, were bypassing the FetchService and getting a handle directly to the SubgraphService. This breaks the abstraction of the FetchService, and creates multiple paths to the SubgraphService.

Instead, this refactors the subscription code to go through the FetchService. Execution of a SubscriptionNode will send a request to the FetchService. To accomplish this, the `Request` type is refactored into an enum with `Fetch` and `Subscription` variants. Since subscriptions aren't supported with connectors, the current implementation in FetchService always uses SubgraphService for the `Subscription` variant.

Some error handling code was also refactored into a common trait implementation to reduce duplicated boilerplate.

In addition to unit and integration tests, this was manually tested with [this subscription demo](https://www.apollographql.com/blog/federated-subscriptions-with-routers-http-callback-protocol).

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

**Notes**

[CNN-372]: https://apollographql.atlassian.net/browse/CNN-372?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bnjjj (commented):
LGTM, could you just make sure when you're closing the subscription from server side and from client side it's also closing everything properly ? It's also unit tested but that's great to also test it manually. For example open a client, open a subscription, and cut it from the client side. On the router you have to enable the prometheus metrics, go to the metrics endpoint and check that `apollo_router_opened_subscriptions` metric is set to 0 when everything is closed. And same scenario but instead of closing the client just kill the subgraph and double check that the client connection is properly closed and the metric is set to 0.

pubmodmatt:
> LGTM, could you just make sure when you're closing the subscription from server side and from client side it's also closing everything properly ? It's also unit tested but that's great to also test it manually. For example open a client, open a subscription, and cut it from the client side. On the router you have to enable the prometheus metrics, go to the metrics endpoint and check that `apollo_router_opened_subscriptions` metric is set to 0 when everything is closed. And same scenario but instead of closing the client just kill the subgraph and double check that the client connection is properly closed and the metric is set to 0.

Thanks, I've confirmed the metric goes to 0 in both cases.

lennyburdette (approved):

PR #5886, opened 2024-08-27
Allow sending the `subgraph.graphql.document` attribute to Studio via OTLP

timbotnik:
There is valuable information in OTel traces that is not being passed to Studio, specifically the GraphQL document being sent to a subgraph.  Today, we filter this attribute out from the Studio path.  

After this change, if a user is using the “recommended” attribute config or enables this with a specific attribute configuration, Router would allow the subgraph document to flow through to traces sent to Studio.  e.g.
```yaml
telemetry:
  instrumentation:
    spans:
```

or
```yaml
telemetry:
  instrumentation:
    spans:
      subgraph:
        attributes:
          subgraph.graphql.document: true
```
We should decide if this is something that would make sense to enable.  There are potential downsides:

- More data will be sent on the wire to Studio to capture these GraphQL document payloads.
- Potentially PII could be included in any inline variables  

image

I'd feel less concerned about this if it was:
- Fully opt-in rather than being included with our "recommended" options
- Using the normalized operation string instead of the raw document

Fixes #**issue_number**

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/plugins/telemetry/tracing/apollo_telemetry.rs
+++ b/apollo-router/src/plugins/telemetry/tracing/apollo_telemetry.rs
@@ -62,6 +62,7 @@ use crate::plugins::telemetry::apollo_otlp_exporter::ApolloOtlpExporter;
 use crate::plugins::telemetry::config::ApolloMetricsReferenceMode;
 use crate::plugins::telemetry::config::Sampler;
 use crate::plugins::telemetry::config::SamplerOption;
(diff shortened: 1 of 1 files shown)


abernix:
Totally reasonable to re-open this, but I'll close this for now!

PR #5890, opened 2024-08-27
Fix invalidated entries counter metric

Geal:
---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-682]: https://apollographql.atlassian.net/browse/ROUTER-682?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apollo-router/src/plugins/cache/invalidation.rs
+++ b/apollo-router/src/plugins/cache/invalidation.rs
@@ -190,7 +190,7 @@ async fn handle_request(
                         u64_counter!(
                             "apollo.router.operations.entity.invalidation.entry",
                             "Entity cache counter for invalidated entries",
-                            len,
+                            count,
                             "origin" = origin,
                             "subgraph.name" = subgraph.clone()
                         );
--- a/apollo-router/src/plugins/cache/invalidation.rs
+++ b/apollo-router/src/plugins/cache/invalidation.rs
@@ -173,7 +173,6 @@ async fn handle_request(
                         .map(|k| RedisKey(k.to_string()))
                         .collect::<Vec<_>>();
                     if !keys.is_empty() {
-                        let len = keys.len() as u64;
                         count += storage.delete(keys).await.unwrap_or(0) as u64;
 
                         u64_counter!(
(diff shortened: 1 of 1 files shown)


apollo-router/src/plugins/cache/invalidation.rs
```diff
@@ -184,13 +184,13 @@ async fn handle_request(
...
-                        count += keys.len() as u64;
-                        storage.delete(keys).await;
+                        let len = keys.len() as u64;
+                        count += storage.delete(keys).await.unwrap_or(0) as u64;
 
                         u64_counter!(
                             "apollo.router.operations.entity.invalidation.entry",
                             "Entity cache counter for invalidated entries",
-                            1u64,
+                            len,
```
garypen, line 193: What is this counter measuring? Scanned or deleted?

I think it should be "deleted", so:

```
let len = storage.delete(keys).await.unwrap_or(0) as u64;
count += len
```

bnjjj (approved):

apollo-router/src/plugins/cache/invalidation.rs
Geal: what do you mean? Updating the description?

garypen: I mean `keys.len()` isn't the same as `storage.delete(keys)...`. That's all. If the motivation is to have a counter that displays how many entries were invalidated, should it be "what we deleted" or "what we scanned". Honestly, I'm not sure, maybe it is better to have it as scanned, since something else might have deleted the records, but ...

Geal: it's more useful to have the number of actually deleted entries yes

abernix:
Note: This had to be retargeted from `main` to `dev` and the PR looked a bit funny so I re-based this and force-pushed.

apollo-router/src/plugins/cache/invalidation.rs
Geal: fixed in [37fb0f0](https://github.com/apollographql/router/pull/5890/commits/37fb0f0ab4ac9bd5abcf5ae994a3c9c441d4fa3c)

Geal:
I think this should be good to merge now (after 1.54 is cut though). It probably does not need a changeset

PR #5891, opened 2024-08-27
Add ability to reference request context values in headers

pubmodmatt:
Add the ability to define headers for connectors that reference request context values. For example, this allows REST API calls made by connectors to include a header with the JWT claims added by a Coprocessor.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[CNN-417]: https://apollographql.atlassian.net/browse/CNN-417?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apollo-router/src/plugins/connectors/handle_responses.rs
+++ b/apollo-router/src/plugins/connectors/handle_responses.rs
@@ -70,7 +70,7 @@ pub(crate) async fn handle_responses<T: HttpBody>(
                     let mut res_data = {
                         let (res, apply_to_errors) = response_key.selection().apply_with_vars(
                             &json_data,
-                            &response_key.inputs().merge(connector.config.as_ref()),
+                            &response_key.inputs().merge(connector.config.as_ref(), None),
                         );
 
                         if let Some(ref debug) = debug {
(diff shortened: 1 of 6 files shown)


pubmodmatt:
@lennyburdette and I had some discussion around whether the name should be `$context` or something like  `$request.context` or `$requestContext`. We settled on `$context`, as there is no conflict with [Federation context](https://www.apollographql.com/docs/federation/entities/use-contexts/). The latter sets the value returned by `@fromContext` onto a named argument, so it can be referenced through `$args`.

lennyburdette (approved):

apollo-router/src/plugins/connectors/handle_responses.rs
```diff
@@ -70,7 +70,7 @@ pub(crate) async fn handle_responses<T: HttpBody>(
                     let mut res_data = {
                         let (res, apply_to_errors) = response_key.selection().apply_with_vars(
                             &json_data,
-                            &response_key.inputs().merge(connector.config.as_ref()),
+                            &response_key.inputs().merge(connector.config.as_ref(), None),
```
lennyburdette, line 73: there's a ticket for making context available in responses right?

pubmodmatt: Yes, I believe so

PR #5896, opened 2024-08-27
fix: use default values on arguments

lennyburdette:
---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [x] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/plugins/connectors/make_requests.rs
```diff
@@ -1709,6 +1709,13 @@ mod graphql_utils {
         variables: &Map<ByteString, JSONValue>,
     ) -> Result<Map<ByteString, JSONValue>, BoxError> {
         let mut arguments = Map::new();
+
+        for argument_def in field.definition.arguments.iter() {
+            if let Some(value) = argument_def.default_value.as_ref() {
+                arguments.insert(argument_def.name.as_str(), argument_value_to_json(value)?);
```
dylan-apollo, line 1715: Instead of always jsonifying up here, we could do `entry().or_insert_with()` after inserting the actual values. Also, this file length makes me itchy 😓

apollo-router/src/plugins/connectors/testdata/nullability.yaml
```diff
@@ -56,4 +56,18 @@ subgraphs:
               """
               entity: true
             )
+
+          defaultArgs(
+            str: String = "default"
+            int: Int = 42
+            float: Float = 1.23
+            bool: Boolean = true
+            # TODO: enums after 2.10.0-alpha.1
```
dylan-apollo, line 65:
```suggestion
            # TODO: enums after 2.10.0-alpha.2
```

(alpha 1 exists, that was the fbp fix)

apollo-router/src/plugins/connectors/tests.rs
```diff
@@ -874,6 +874,49 @@ async fn test_nullability() {
...
+        vec![Matcher::new()
+            .method("POST")
+            .path("/default-args")
+            .body(serde_json::json!({
+              "str": "default",
+              "int": 42,
+              "float": 1.23,
+              "bool": true,
+              "arr": ["default"],
+            }))
```
dylan-apollo, line 915: We should make a second test that verifies that providing values overrides the defaults

apollo-router/src/plugins/connectors/make_requests.rs
lennyburdette: move the module to its own file and took your suggestion. let me know if you have ideas on handling the fallible conversion fn!

apollo-router/src/plugins/connectors/tests.rs
lennyburdette: done!

apollo-router/src/plugins/connectors/make_requests/graphql_utils.rs
```diff
@@ -0,0 +1,132 @@
...
+        }
+    }
+
+    for argument_def in field.definition.arguments.iter() {
+        if let Some(value) = argument_def.default_value.as_ref() {
+            arguments
+                .entry(argument_def.name.as_str())
+                .or_insert_with(|| {
+                    argument_value_to_json(value).unwrap_or_else(|err| {
+                        tracing::warn!("failed to convert default value to json: {}", err);
```
pubmodmatt, line 43: Nitpick: this would result in a very vague warning like `failed to convert default value to json: invalid int`. It'd be helpful to have the name of the field and argument.

pubmodmatt (approved):

apollo-router/src/plugins/connectors/make_requests/graphql_utils.rs
lennyburdette: added!

PR #5902, opened 2024-08-27
Update to router-bridge@0.6.0+v2.9.0

tninesling:


abernix (approved):

abernix (approved):
Thank you!  LGTM.  Would [suggest adding a changeset](https://github.com/apollographql/router/blob/dev/.changesets/README.md#how-to-create-a-changelog-entry) for federation version bumps in the future, just to ensure we have the right changelog entries for it and do the `federation-version-support.mdx` docs update, but we're good this time. (I got it, we're about to finalize.)

PR #5915, opened 2024-08-29
validate fields selected

lennyburdette:
---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-federation/src/sources/connect/validation/test_data/validation/all_fields_selected.graphql
```diff
@@ -0,0 +1,47 @@
...
+      child {
+        id
+      }
+      """
+    )
+}
+
+type T @key(fields: "id") {
+  id: ID!
+  unselected: String!
```
pubmodmatt, line 23: Nitpick: there may be some additional cases that would be good to add tests for. For example, if the selection aliases a field, or unwraps a child field, or wraps a field.

pubmodmatt (approved):

apollo-federation/src/sources/connect/validation/test_data/validation/all_fields_selected.graphql
lennyburdette: i'm not too concerned about this because we're using the same visitor used in expansion code, but i added a few more cases

Issue #3264, opened 2023-06-15
Support passing Apollo Key/GraphRef via Config File


lleadbet:
**Is your feature request related to a problem? Please describe.**

The router currently requires that I pass an APOLLO_KEY and APOLLO_GRAPH_REF environment variable to start, however in some situations I would like to pass it as a static value within a configuration or another environment variable name, or as passing from a file's contents. Currently this isn't supported, however. 

The YAML implementation supports both passing values by environment variable and passing file contents directly, which would address my concerns. 

**Describe the solution you'd like**

A new configuration option to pass the APOLLO_KEY/GRAPH_REF, such as: 

```yml
supergraph:
  apollo_key: 1234
  apollo_graph_ref: my-supergraph@prod
```

Which would default to: 
```yml
supergraph:
  apollo_key: ${env.APOLLO_KEY}
  apollo_graph_ref: ${env.APOLLO_GRAPH_REF}
```

To make it a non-breaking change. 

**Describe alternatives you've considered**

* Using the existing environment variables and populating them via a prelaunch script (e.g. sourcing a .env file) 

**Additional context**

While this feature would be relatively uncommon, it would be helpful for some organizations.

lleadbet:
Supporting a file could be nice as a stopgap, however agree re: inline secrets.

smyrick:
If we could support loading a `.env` file that would also help: https://crates.io/crates/dotenv

smyrick:
Updating here as another customer is asking for this support. They have an environment that has strict requirements and nothing can be saved as a env var so everything must be read from file. They can manage all the other secrets but right now the `APOLLO_KEY` is one secret that does not even have the ability to be read from file.

If we added support to set the key in the config we could use variaible expansion `${file.apollo-key.txt}`, but I agree to Jesse's point, to maybe defer anyone from even accidentally saving the raw value into the file we bypass that and just add another way to load the Apollo key is from a file and the Router does that for you

lleadbet:
> I'm very sympathetic to supporting reading from a file. Also very sympathetic to reading from another environment variable, though the "defaulting" pattern ${var-:default} (or whatever it is, I can't remember at the moment 😭 ), makes it still possible that someone could leverage that to put a secret in-line.

I would also like to add that we have plenty of other places this can happen- OTLP setup is a great example. 

I agree we should probably try to avoid footguns, but there's a point where we can only prevent so much for the user without affecting the UX.

gbarsky:
Hello. 
[...]

PR #5917, opened 2024-08-29
feat: allow users to load apollo key from file

lleadbet:
Customers sometimes would rather not pass sensitive keys to the router through environment variables out of an abundance of caution related to concerns about leaking them.

This PR allows users to pass a new argument `--apollo-key-path` that takes a file location as an argument, and uses it as the APOLLO_KEY value for use with Uplink and usage reporting.

Addresses a portion of #3264, specifically the APOLLO_KEY.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

This does not fully address #3264 as it doesn't support APOLLO_GRAPH_REF, but addresses the most significant concern of that issue.

**Notes**

I'm unsure where both tests and docs should be updated here; if there's a good place, let me know.


.changesets/feat_feat_key_from_file.md
```diff
@@ -0,0 +1,7 @@
+### feat: allow users to load apollo key from file ([PR #5917](https://github.com/apollographql/router/pull/5917))
+
+Users sometimes would rather not pass sensitive keys to the router through environment variables out of an abundance of caution. To help address this, you can now pass an argument `--apollo-key-path` that takes a file location as an argument which is read and then used as the Apollo key for use with Uplink and usage reporting.
```
smyrick, line 3:
```suggestion
Users sometimes would rather not pass sensitive keys to the router through environment variables out of an abundance of caution. To help address this, you can now pass an argument `--apollo-key-path` or env var `APOLLO_KEY_PATH`, that takes a file location as an argument which is read and then used as the Apollo key for use with Uplink and usage reporting.
```

smyrick:
Do we also need to make doc changes in the same PR?

lleadbet:
@smyrick - yes, but unsure where in the docs it should live; probably a good question for @apollographql/docs.

smyrick:
@lleadbet What about updating this page for docs? https://github.com/apollographql/router/blob/main/docs/source/configuration/overview.mdx

lleadbet:
> @lleadbet What about updating this page for docs? https://github.com/apollographql/router/blob/main/docs/source/configuration/overview.mdx
> 
> Which is this page on the UI: https://www.apollographql.com/docs/router/configuration/overview/#environment-variables

good shout, updated

abernix:
I'm going to push an empty commit to your PR.

abernix:
Closing this in favor of #5982.

PR #5919, opened 2024-08-29
Refactor SelectionSet::merge_selections_into and remove unnecessary directives in Selection::from_element

TylerBloom:
This PR introduces two optimizations to `SelectionSet::merge_selections_into`.

The simpler of the two removes the intermediate buffers used in the method. Now, selections whose keys already exist in the selection map are immediately merged together. This means there is more recursion but no allocating buffers. It remains unclear if this will have any performance impact (positive or negative), but it simplifies the code by a fair amount. Because of this, all of the `*SelectionValue::merge_into` methods have also be refactored to take a single selection rather than an iterator. This removes more intermediate buffering.

The second change, which actually relates to the ticket, adds logic to `Selection::from_element` to remove optionally unnecessary directives from the element. The intention of this is to remove `@include`/`@skip` directives. There is already logic to remove these directives from the selection set beforehand, but this is incomplete without also removing those directives from the element that it is paired with to form a `Selection`.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-federation/src/operation/merging.rs
+++ b/apollo-federation/src/operation/merging.rs
@@ -48,7 +48,7 @@ impl<'a> FieldSelectionValue<'a> {
         if self.get().selection_set.is_some() {
             let Some(other_selection_set) = &other.selection_set else {
                 return Err(FederationError::internal(format!(
-                    "Field \"{}\" has non-composite type but also has a selection set",
+                    "Field \"{}\" has composite type but not a selection set",
                     other_field.field_position,
                 )));
             };
--- a/apollo-federation/src/operation/merging.rs
+++ b/apollo-federation/src/operation/merging.rs
@@ -155,12 +155,17 @@ impl SelectionSet {
             }
             selections_to_merge.extend(other.selections.values());
         }
-        self.merge_selections_into(selections_to_merge.into_iter())
+        self.merge_selections_into(selections_to_merge.into_iter(), false)
     }
 
     /// NOTE: This is a private API and should be used with care, use `add_selection` instead.
     ///
     /// A helper function for merging the given selections into this one.
+    ///
+    /// The `do_fragment_inlining` flag enables a check to see if any inline fragments yielded from
+    /// `others` can be recursively merged into the selection set instead of just merging in the
+    /// fragment. This requires that the fragment has no directives and either has no type
+    /// condition or the type condition matches this selection set's type position.
     ///
     /// # Errors
     /// Returns an error if the parent type or schema of any selection does not match `self`'s.
@@ -169,7 +174,8 @@ impl SelectionSet {
     #[allow(unreachable_code)]
     pub(super) fn merge_selections_into<'op>(
         &mut self,
-        others: impl Iterator<Item = &'op Selection>,
+        mut others: impl Iterator<Item = &'op Selection>,
+        do_fragment_inlining: bool,
     ) -> Result<(), FederationError> {
         fn insert_selection(
             target: &mut SelectionMap,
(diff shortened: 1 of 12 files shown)


apollo-federation/src/operation/merging.rs
```diff
@@ -173,116 +175,222 @@ impl SelectionSet {
...
+        // never change.
+        if try_flattening_fragments {
+            fn recurse_on_inline_fragment<'a>(
+                buffer: &mut SelectionBuffer<'a>,
+                type_pos: &CompositeTypeDefinitionPosition,
+                others: impl Iterator<Item = &'a Selection>,
+            ) -> Result<(), FederationError> {
+                for other_selection in others {
+                    if let Selection::InlineFragment(inline) = other_selection {
+                        if inline.inline_fragment.data().directives.len() != 0
```
dariuszkuc, line 279: I think this is problematic -> we can only flatten the fragment IFF parent type == type condition AND (either we don't have any directives OR we have same directives as the parent)

If there is no explicit type condition we still need to compare directives.

apollo-federation/src/operation/merging.rs
```diff
@@ -173,119 +156,118 @@ impl SelectionSet {
...
+                        let inline_directives = &inline.inline_fragment.data().directives;
+                        if (inline_directives.len() == 0 || inline_directives == directives)
+                            && inline
+                                .inline_fragment
+                                .data()
+                                .type_condition_position
+                                .as_ref()
+                                .map(|pos| pos == type_pos)
+                                .unwrap_or(true)
+                        {
```
dariuszkuc, line 253:
```suggestion
                        if inline.is_unnecessary(&type_pos)
```

let's just keep it simple and do what JS does (no need to pass the parent directives as we don't have access to those when adding op path elements)

TylerBloom: Hmm... I see what you're saying. I was under the impression that `add_at_path` (or rather its call sites) did checks to see if the directives matched. I could easily be wrong, though.

apollo-federation/src/operation/mod.rs
```diff
@@ -2651,7 +2651,7 @@ impl SelectionSet {
                     } else {
                         // add leaf
                         let selection = Selection::from_element(element, None)?;
-                        self.add_local_selection(&selection)?
+                        self.add_local_selection(&selection, Some(selection.directives()))?
```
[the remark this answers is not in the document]
TylerBloom: Hmm... I see what you're saying. I was under the impression that `add_at_path` (or rather its call sites) did checks to see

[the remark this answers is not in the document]
dariuszkuc: on JS side of things we don't -> 
1. when adding paths we either selections OR [`PathBasedUpdate`](https://github.com/apollographql/federation/blob/main/internals-js/src/operations.ts#L2288)
2. when we construct the final the selection sets we just check whether its a regular selection or path based update -> [link](https://github.com/apollographql/federation/blob/main/internals-js/src/operations.ts#L2409-L2415)
3. if it is a single path update we then we [remove unnecessary fragments](https://github.com/apollographql/federation/blob/main/internals-js/src/operations.ts#L2433) from the passed in selection by checking [if fragments are unnecessary](https://github.com/apollographql/federation/blob/main/internals-js/src/operations.ts#L2366)

We don't check for parent selection directives -> we shouldn't be doing this here either (and we don't have access to them either).

apollo-federation/src/query_plan/conditions.rs
```diff
@@ -262,14 +262,12 @@ pub(crate) fn remove_unneeded_top_level_fragment_directives(
...
-                                needed_directives.push(directive.clone());
-                            }
-                        }
-                    }
+                    let needed_directives: Vec<Node<Directive>> = fragment
+                        .directives
+                        .iter()
+                        .filter(|directive| !unneded_directives.contains(directive))
+                        .cloned()
+                        .collect();
```
TylerBloom, line 270: This should only be a stylistic change.

apollo-federation/src/query_plan/conditions.rs
```diff
@@ -282,12 +280,13 @@ pub(crate) fn remove_unneeded_top_level_fragment_directives(
                         let final_selection =
                             inline_fragment.with_updated_selection_set(updated_selections);
                         selection_map.insert(Selection::InlineFragment(Arc::new(final_selection)));
+                    } else {
+                        // We can skip some of the fragment directives directive.
+                        let final_selection = inline_fragment
+                            .with_updated_directives(DirectiveList::from_iter(needed_directives))
+                            .with_updated_selection_set(updated_selections);
+                        selection_map.insert(Selection::InlineFragment(Arc::new(final_selection)));
```
TylerBloom, line 288: This is unrelated to the problem that is being solved, but it is a bug. The JS code returned inside the `if` statement. Also, the selection set was not being updated like it should have been.

apollo-federation/src/query_plan/conditions.rs
```diff
@@ -282,12 +280,13 @@ pub(crate) fn remove_unneeded_top_level_fragment_directives(
                         let final_selection =
                             inline_fragment.with_updated_selection_set(updated_selections);
                         selection_map.insert(Selection::InlineFragment(Arc::new(final_selection)));
+                    } else {
+                        // We can skip some of the fragment directives directive.
+                        let final_selection = inline_fragment
+                            .with_updated_directives(DirectiveList::from_iter(needed_directives))
+                            .with_updated_selection_set(updated_selections);
```
dariuszkuc, line 287: nit: should we expose a single method that updates both at the same time? otherwise we are doing extra allocation for no reason

apollo-federation/src/operation/mod.rs
```diff
@@ -2618,6 +2649,7 @@ impl SelectionSet {
                             // We immediately add a selection afterward to make this selection set
                             // valid.
                             Some(SelectionSet::empty(self.schema.clone(), sub_selection_type)),
+                            None,
```
dariuszkuc, line 2652: should we also pass those `unneeded directives` here?

apollo-federation/src/operation/merging.rs
```diff
@@ -28,43 +29,38 @@ impl<'a> FieldSelectionValue<'a> {
...
                     other_field.field_position,
                 )));
-            }
+            };
+            selection_set = Some(other_selection_set);
+        } else if other.selection_set.is_some() {
+            return Err(FederationError::internal(format!(
+                "Field \"{}\" has non-composite type but also has a selection set",
+                other_field.field_position,
+            )));
```
dariuszkuc, line 60: looks like we flipped error messages on those conditions?

TylerBloom: Ah, ya. I made that mistake during clean up...

dariuszkuc (approved):

PR #5921, opened 2024-08-29
Implement REST API response snapshots for connectors

pubmodmatt:
Add a debugging feature to store REST API responses as local JSON snapshot files. This allows:

* Development and debugging without having to hit the backend REST endpoint repeatedly. This is helpful, for example, if there is a rate limit or short-lived auth token required, or if you just need to work offline.
* Easy development of tests that use real-world REST API responses. There is an `offline` option, which can be used once test development is complete, that will force the test to use the snapshot and prevent ever trying to hit the real backend API. There is also an `update` option to force an update, which can be used to update test data after changes to the REST service.
* Mock a REST API that doesn't exist. By turning on `offline` mode, and observing the warning messages for missing snapshots, it's possible to manually create the JSON snapshot files and mock up a completely fictional REST API. The `@source` URL can be any arbitrary URL in this case, since it will never be used. This provides a quick way to test with connectors without having to code up and run a server.
* Mock error responses. You can snapshot REST API responses, then change the status in the snapshot to test error handling.

As an example, the quickstart tests are rewritten using the snapshot feature. This demonstrates that no mock server or hand-coding of mocks is required to write such a test. Note that I've redacted all headers except `Content-Type` in these snapshots, and reduced the number of posts returned from 100 to 2 (for parity with the previous implementation of the test).

See the `snapshot` module rustdoc for more details.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/plugins/connectors/configuration.rs
+++ b/apollo-router/src/plugins/connectors/configuration.rs
@@ -28,7 +28,7 @@ pub(crate) struct ConnectorsConfig {
 
     /// Configuration for HTTP response snapshots
     #[serde(default)]
-    pub(crate) snapshot: SnapshotConfiguration,
+    pub(crate) experimental_snapshots: SnapshotConfiguration,
 }
 
 /// Configuration for a connector subgraph
--- a/apollo-router/src/plugins/connectors/configuration.rs
+++ b/apollo-router/src/plugins/connectors/configuration.rs
@@ -25,6 +25,10 @@ pub(crate) struct ConnectorsConfig {
     /// The maximum number of requests for a connector source
     #[serde(default)]
     pub(crate) max_requests_per_operation_per_source: Option<usize>,
+
+    /// Configuration for HTTP response snapshots
+    #[serde(default)]
+    pub(crate) snapshot: SnapshotConfiguration,
 }
 
 /// Configuration for a connector subgraph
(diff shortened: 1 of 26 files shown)


apollo-router/src/services/connector_service/snapshot.rs
```diff
@@ -0,0 +1,122 @@
...
+#[derive(Serialize, Deserialize)]
+pub(crate) struct Snapshot<'a> {
+    url: Cow<'a, str>,
+    headers: IndexMap<String, String>,
+    body: Cow<'a, Value>,
+}
+
+impl<'a> Snapshot<'a> {
+    pub(crate) fn new(url: &'a str, body: &'a Value, headers: &'a HeaderMap<HeaderValue>) -> Self {
+        let headers = headers
```
lennyburdette, line 57: should this be a `Vec<(String, String)>` to support repeated header keys?

apollo-router/src/services/connector_service/snapshot.rs
```diff
@@ -0,0 +1,122 @@
...
+    }
+}
+
+fn snapshot_path(base: PathBuf, url: &str) -> PathBuf {
+    let url_hash = {
+        let mut hasher = DefaultHasher::new();
+        url.hash(&mut hasher);
+        hasher.finish()
+    };
+    base.join(url_hash.to_string()).with_extension("json")
```
lennyburdette, line 121: we might need to hash the body too. i've seen apis like `POST /search` where all the unique parameters are in the body

pubmodmatt: Interesting - that could be a good improvement as long as the body doesn't contain any data that might change, like a timestamp or something. I also realize I should probably include HTTP method in case somebody uses the same URL for PUT/POST or something.

apollo-router/src/services/connector_service/snapshot.rs
pubmodmatt: Good call. The `HeaderMap` doesn't support serialization so I had to switch map types, but I overlooked that `HeaderMap` is a multi-map.

apollo-router/src/plugins/connectors/configuration.rs
```diff
@@ -25,6 +25,10 @@ pub(crate) struct ConnectorsConfig {
     /// The maximum number of requests for a connector source
     #[serde(default)]
     pub(crate) max_requests_per_operation_per_source: Option<usize>,
+
+    /// Configuration for HTTP response snapshots
+    #[serde(default)]
+    pub(crate) snapshot: SnapshotConfiguration,
```
lennyburdette, line 31: as the only user-facing API change, i'm wondering if we should be conservative here and call it something like `experimental_snapshot` or `preview_snapshot`. i'd like to think through the product implications of this feature a bit more before committing to it as public api in perpetuity.

lennyburdette (approved):
this is very cool! i'm wondering if there's an opportunity to build upon this as a product feature with sandbox integration, or language server integration (powering autocomplete?), or something else. it'll be interesting to see how this interacts (or doesn't!) with future source specification awareness.

pubmodmatt:
Closing in favor of #6032

PR #5922, opened 2024-08-29
Remove `test_updated` from release workflows

dylan-apollo:
We've had a couple of releases blocked at the final step by this job, because some dependency updates their MSRV. We use `Cargo.lock` to build our binaries, so it shouldn't matter if we can't pass tests on the _latest_ versions of those dependencies.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1101,20 +1101,6 @@ workflows:
               ignore: /.*/
             tags:
               only: /v.*/
-      - test_updated:
-          requires:
-            - lint
-            - check_helm
-            - check_compliance
-          matrix:
-            parameters:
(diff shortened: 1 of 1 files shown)


Geal (changes requested):
this test is used to guarantee that users building rust plugins with the [published crate](https://crates.io/crates/apollo-router) will still have code that builds and passes tests. That test has already caught a lot of issues with dependencies publishing breaking changes on minor versions.
If you see a dependency that fails to build due to that test, you should instead fix its version in Cargo.toml

Geal:
CI is working again with https://github.com/apollographql/router/pull/5924 and https://github.com/apollographql/router/pull/5926

PR #5925, opened 2024-08-30
Paginate redis scan pages properly and don't use Notify anymore for entity caching

bnjjj:
Calling https://docs.rs/fred/latest/fred/types/trait.Scanner.html#tymethod.next everytime.

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [x] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/plugins/cache/invalidation.rs
```diff
@@ -73,27 +66,24 @@ pub(crate) enum InvalidationOrigin {
 
 impl Invalidation {
     pub(crate) async fn new(storage: Arc<EntityStorage>) -> Result<Self, BoxError> {
-        let mut notify = Notify::new(None, None, None);
-        let (handle, _b) = notify.create_or_subscribe(InvalidationTopic, false).await?;
-
-        let h = handle.clone();
+        let (tx, rx) = tokio::sync::mpsc::channel(128);
```
garypen, line 69: Why 128?

I think this now acts as an informal concurrency limit on the number of in-flight invalidation requests. That's probably desirable, but we should think carefully about the exact number we want and maybe also consider documenting it.

We should make it a global `const` and use the const in the tests and give it a really good name that explains what it's doing.

Geal (approved):

apollo-router/src/plugins/cache/invalidation.rs
bnjjj: I just wanted to keep the same configuration than in Notify https://github.com/apollographql/router/blob/bnjjj/fix_redis_scan/apollo-router/src/notification.rs#L33:L33 the previous impl we used. But I think it should be 1024 that's the other const I have to take. I'll create a new const for this

PR #5931, opened 2024-08-30
enabling connectors debugging via embed

tayrrible:
*Description here*

Fixes #**issue_number**

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/templates/sandbox_index.html
+++ b/apollo-router/templates/sandbox_index.html
@@ -64,6 +64,7 @@
             },
             hideCookieToggle: false,
             endpointIsEditable: false,
+            enableConnectorsDebugging: true,
             runtime: "apollo-router@{{ apollo_router_version }}"
         });
         </script>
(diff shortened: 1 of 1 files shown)


lennyburdette (approved):

apollo-router/templates/sandbox_index.html
```diff
@@ -64,6 +64,7 @@ <h1>Welcome to the Apollo Router</h1>
             },
             hideCookieToggle: false,
             endpointIsEditable: false,
+            enableConnectorsDebugging: true,
```
lennyburdette, line 67: should the name match the other similar flag `initialRequestQueryPlan`?

tayrrible: @lennyburdette can do but would require refactoring & releasing studio code first

tayrrible: Would `initialRequestConnectorsDebugging` make sense?

lennyburdette: yeah that's what i was thinking. it's certainly not a requirement, but once this is baked into router's sandbox html, we'll need to support it in perpetuity.

PR #5935, opened 2024-08-30
Remove `test_updated` from release workflows

dylan-apollo:
Originally pointed at `dev`, but this test is designed to catch issues for consumers of the `apollo-router` crate, so we'd need to replace it with something else for that use-case.

Since `next` doesn't _have_ any crate consumers yet, it should be safe to remove it to speed along alphas. We can replace it once we don't have to be pinned to an older version of Rust anymore (soon on next 🤞) which will resolve most of the failures.

Since I'm thinking about it, when we replace it we should also move crate publishing to CI, since this is designed to protect the crates, not the binaries.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1101,20 +1101,6 @@ workflows:
               ignore: /.*/
             tags:
               only: /v.*/
-      - test_updated:
-          requires:
-            - lint
-            - check_helm
-            - check_compliance
-          matrix:
-            parameters:
-              platform:
-                [ amd_linux_test ]
-          filters:
-            branches:
-              ignore: /.*/
-            tags:
-              only: /v.*/
       - test:
           requires:
             - lint
@@ -1133,7 +1119,6 @@ workflows:
           requires:
             - pre_verify_release
             - test
-            - test_updated
           matrix:
             parameters:
               platform:


pubmodmatt (approved):
Sounds good to me, as long as we have the replacement tracked in an open issue

Geal:
still not a good idea to do that, even just on next, because you will not see potential issues until the moment we try to realign versions

abernix (changes requested):
Marking as requests changes to make sure this is followed up with an issue that the core team agrees with.

Let's probably chat with the architectural group about what would make that whole group comfortable with this as I believe this was for a bit more than _just_ protecting published crates.  We did intentionally keep `test_updated` out of the (`main`) merge required branch protections, specifically to avoid "surprise you can't release" moments in the releasing pipeline which I believe have contributed to the "Motivation" for this PR? 

At the very least, to feel good about landing this I would personally want to make sure this is ticketed as a high priority thing to fix on our project planning boards by a specific date because this needs to get re-introduced at a not-too-distant point and the divergence could easily be meaningful when we turn more attention to `next` soon.

abernix:
#6312 now runs `test_updated` as a separate job, on a crontab, and connected to a Slack channel that is monitored  Thanks the inspiration.

PR #5807, opened 2024-08-13
Remove experimental label from apollo usage reporting configs

bonnici:
Removes the experimental label and renames some config options related to Apollo usage reporting:
* `experimental_apollo_metrics_generation_mode` has been removed since the Rust implementation has been the default since v1.49.0 and it is generating reports identical to the router-bridge implementation
* `telemetry.apollo.experimental_apollo_signature_normalization_algorithm` is now `telemetry.apollo.signature_normalization_algorithm`
* `telemetry.apollo.experimental_apollo_metrics_reference_mode` is now `telemetry.apollo.metrics_reference_mode`

The defaults have not been changed.

---

**Checklist**

Complete the checklist (and note appropriate exceptions) before the PR is marked ready-for-review.

- [x] Changes are compatible[^1]
- [x] Documentation[^2] completed
- [ ] Performance impact assessed and acceptable
- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[^1]: It may be appropriate to bring upcoming changes to the attention of other (impacted) groups. Please endeavour to do this before seeking PR approval. The mechanism for doing this will vary considerably, so use your judgement as to how and when to do this.
[^2]: Configuration is an important part of many changes. Where applicable please try to document configuration examples.
[^3]: Tick whichever testing boxes are applicable. If you are adding Manual Tests, please document the manual testing (extensively) in the Exceptions.

[ROUTER-696]: https://apollographql.atlassian.net/browse/ROUTER-696?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #5938, opened 2024-09-03
docs: update/annotate formerly experimental configs

Meschreiber:
This PR removes the `<ExperimentalFeature>` annotation for two configs that are no longer experimental and adds a note for when they became generally available. I considered adding the following text to the note, but found it unnecessary. 

> Router configurations with `telemetry.apollo.experimental_apollo_metrics_reference_mode` will continue to work, but Apollo recommends updating the configuration to `telemetry.apollo.metrics_reference_mode`.

Happy to add it back if others deem it's helpful.

See https://github.com/apollographql/router/pull/5807 for more info.

### DIFF
--- a/docs/source/configuration/overview.mdx
+++ b/docs/source/configuration/overview.mdx
@@ -598,9 +598,13 @@ supergraph:
 
 </MinVersion>
 
-<ExperimentalFeature />
+<Note>
 
-Beginning in v1.49.0, the router supports enhanced operation signature normalization.
+Beginning with [v1.49.0](https://github.com/apollographql/router/releases/tag/v1.49.0), the router supported enhanced operation signature normalization [experimentally](/resources/product-launch-stages/#experimental-features).
+In [v.1.53.0](https://github.com/apollographql/router/releases/tag/v1.53.0), this configuration became [generally available](/resources/product-launch-stages/#general-availability).
+
+</Note>
+
 Apollo's legacy operation signature algorithm removes information about certain fields, such as input objects and aliases.
 This removal means some operations may have the same normalized signature though they are distinct operations.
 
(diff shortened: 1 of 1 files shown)


docs/source/configuration/overview.mdx
```diff
@@ -598,9 +598,13 @@ supergraph:
 
 </MinVersion>
 
-<ExperimentalFeature />
+<Note>
+
+Beginning with [v1.49.0](https://github.com/apollographql/router/releases/tag/v1.49.0), the router supported enhanced operation signature normalization [experimentally](/resources/product-launch-stages/#experimental-features).
+In [v.1.53.0](https://github.com/apollographql/router/releases/tag/v1.53.0), this configuration became [generally available](/resources/product-launch-stages/#general-availability).
```
shorgi, line 604: Suggest listing GA and Experimental versions to be more scannable. Also applies to other Note in PR.

```suggestion

The router supports enhanced operation signature normalization in the following versions:

* [General availability](/resources/product-launch-stages/#general-availability) in v1.53.0 and later
* [Experimental](/resources/product-launch-stages/#experimental-features) in v1.49.0 to v1.52

```

shorgi (approved):

docs/source/configuration/overview.mdx
bonnici, line 604: The config rename will happen in v1.54.0, so we should use that as the version it became GA.

docs/source/configuration/overview.mdx
```diff
@@ -761,11 +765,17 @@ query AliasedQuery {
 
 </MinVersion>
 
-<ExperimentalFeature />
+<Note>
+
+Beginning with [v1.50.0](https://github.com/apollographql/router/releases/tag/v1.50.0), the router supported extended reference reporting [experimentally](/resources/product-launch-stages/#experimental-features).
+In [v.1.53.0](https://github.com/apollographql/router/releases/tag/v1.53.0), this configuration became [generally available](/resources/product-launch-stages/#general-availability).
```
bonnici, line 771: Same comment - this should be 1.54.0

docs/source/configuration/overview.mdx
bonnici, line 604: The config rename will happen in v1.54.0, so we should use that as the version it became GA (and not merge this until then).

Meschreiber: This is merging into dev, so I think we can/should merge before the release, right?

abernix: Yes! We'll start 1.54.0 (as a release candidate) today and this will not land in our docs until next Tuesday when we finalize 1.54.0. 😄   I'll merge this to make sure it makes the cut.

PR #5941, opened 2024-09-03
remove the invalidation task

Geal:
this task was executing invalidation requests one by one, introducing a lot of latency in the process. What we actually want is executing those requests as soon as possible, but limiting the number of in flight invalidation requests at any point

TODO:
- [ ] rewrite the removed tests in another way (maybe from integration tests?)
- [ ] add rate limiting

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/plugins/cache/invalidation.rs
```diff
@@ -128,117 +88,130 @@ async fn start(
...
+                }
+                Ok(scan_res) => {
+                    if let Some(keys) = scan_res.results() {
+                        let keys = keys
+                            .iter()
+                            .filter_map(|k| k.as_str())
+                            .map(|k| RedisKey(k.to_string()))
+                            .collect::<Vec<_>>();
+                        if !keys.is_empty() {
+                            let deleted = redis_storage.delete(keys).await.unwrap_or(0) as u64;
```
bnjjj, line 136: In my branch when I tested to improve perfs I put all the keys to delete in an array to not block the scan and then delete everything once. What do you think ?

Geal: In what way would this block the scan?

PR #5577, opened 2024-07-01
chore(tests): Apply nextest retries rules to known-flaky tests.

abernix:
### Motivation

We have some known flaky tests that are in our test suite.  These flaky tests are valuable, but often fail because of network-related environment things or approaches that are based on timing that don't always work out.

### Approach

This attempts to apply retries — using `nextest` [retries](https://nexte.st/docs/features/retries/) — to known flaky tests to provide some air-cover in CI while we action the specific reasons.  This choses to only retry known failures to avoid introducing net-new flaky tests inadvertently.

This intends to apply the retries to **both CI and local development.**

### Review Instructions

Please validate that this applies correctly to CI and local environments and please verify that tests are running appropriately.  Feel free to re-run CI jobs on this PR, or run multiple times locally, in order to validate!

---

**Checklist**

Complete the checklist (and note appropriate exceptions) before the PR is marked ready-for-review.

- [x] Changes are compatible[^1]
- [ ] Documentation[^2] completed
- [ ] Performance impact assessed and acceptable
- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

⚠️ : ___Several details of the checklist are not relevant for this type of PR and a changelog is not necessary.___

**Notes**

[^1]: It may be appropriate to bring upcoming changes to the attention of other (impacted) groups. Please endeavour to do this before seeking PR approval. The mechanism for doing this will vary considerably, so use your judgement as to how and when to do this.
[^2]: Configuration is an important part of many changes. Where applicable please try to document configuration examples.
[^3]: Tick whichever testing boxes are applicable. If you are adding Manual Tests, please document the manual testing (extensively) in the Exceptions.

PR #5944, opened 2024-09-03
chore(tests): Update flaky tests retry list from CircleCI API

abernix:
Repeats the https://github.com/apollographql/router/pull/5577 motion again, which was successful in reducing time spent in CI overall.

Overall, the API only shows tests that are repeatedly failing BEFORE retriese are applied, so this is careful to not change the existing sorting in generating the diff so we can see the new tests in the context of this PR but still maintain the previously flaky list.  This was done with `sort | uniq` with the existing list to interpolate the new tests.


abernix (commented):
Overall, none of these new entries surprise me.  They just weren't on the top 100 list last time.  They all look like the same class of flaky integration tests we've had previously.

tninesling (approved):

PR #5945, opened 2024-09-03
docs: Run everything in parallel when releasing

dylan-apollo:
When we're kicking off a release, we _expect_ everything to pass, since presumably the base branch was passing before we started releasing. This PR should shorten the amount of time we have to wait on CI to finish a release, by running everything in parallel.

The publish does still wait for _all_ jobs to complete, so if one _does_ fail, we won't publish a bad release.


abernix (dismissed):
I'm almost certain this was all intentionally designed this way to prevent actual problems in releases.  Could we bring the problems to be solved to an issue first or the architectural meeting?  I would not be at all comfortable with the risk right now at face value and this doesn't seem to mitigate those or consider how we got here.

dylan-apollo:
Ready for review after the architectural meeting

abernix (approved):

PR #5947, opened 2024-09-03
Replace internal connectors subgraph name in errors

pubmodmatt:
Replace the internal connectors subgraph name with the connector in error responses. Previously, errors at the HTTP transport level were returning the internal subgraph name.

The error response now looks like:

```json
{
  "data": null,
  "errors": [
    {
      "message": "HTTP fetch failed from 'characters.simpsons http: GET /characters': error trying to connect: tcp connect error: Connection refused (os error 61)",
      "path": [],
      "extensions": {
        "service": "characters.simpsons http: GET /characters",
        "reason": "error trying to connect: tcp connect error: Connection refused (os error 61)"
      }
    }
  ],
  "extensions": {
    "valueCompletion": [
      {
        "message": "Cannot return null for non-nullable field Query.characters",
        "path": []
      }
    ]
  }
}
```

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[CNN-433]: https://apollographql.atlassian.net/browse/CNN-433?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apollo-router/src/services/connector_service.rs
+++ b/apollo-router/src/services/connector_service.rs
@@ -18,6 +18,7 @@ use super::connect::BoxService;
 use super::http::HttpClientServiceFactory;
 use super::http::HttpRequest;
 use super::new_service::ServiceFactory;
+use crate::error::FetchError;
 use crate::plugins::connectors::error::Error as ConnectorError;
 use crate::plugins::connectors::handle_responses::handle_responses;
 use crate::plugins::connectors::http::Response as ConnectorResponse;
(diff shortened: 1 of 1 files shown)


apollo-router/src/services/connector_service.rs
```diff
@@ -167,7 +168,24 @@ async fn execute(
                 http_request: req,
                 context,
             };
-            let res = client.oneshot(req).await?;
+            let res = client.oneshot(req).await.map_err(|e| {
+                match e.downcast::<FetchError>() {
+                    // Replace the internal subgraph name with the connector label
+                    Ok(inner) => match *inner {
+                        FetchError::SubrequestHttpError {
```
lennyburdette, line 175: a few different FetchError variants include service name: https://github.com/apollographql/router/blob/dev/apollo-router/src/error.rs#L66,75 ... patch response, ws, and batching are unlikely to happen but should we check for them anyway?

lennyburdette (approved):

apollo-router/src/services/connector_service.rs
pubmodmatt: The fetch service [forces everything into `SubrequestHttpError`](https://github.com/apollographql/router/blob/next/apollo-router/src/services/fetch.rs#L107-L128). However, I missed that it's setting the `reason` using the `Display` implementation of the original error, and since that's derived, it will include the `service` field. I'll need to fix.

pubmodmatt: Actually, looking at it more, those errors are never returned except through the subgraph service, so they would never happen for connector service. The only error that can come out of the connector service is `SubrequestHttpError`, either from the HTTP client service (which this PR handles) or directly [emitted by the fetch service](https://github.com/apollographql/router/blob/0b67964fccb367f20b89ee64c66cd62a9b9e9509/apollo-router/src/plugins/connectors/handle_responses.rs#L159), which already returned the connector label.

PR #5949, opened 2024-09-04
composition fixes

lennyburdette:
---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [x] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-federation/src/sources/connect/expand/mod.rs
```diff
@@ -614,6 +614,32 @@ mod helpers {
                         ))
                     }
                 }?;
+            } else {
```
lennyburdette, line 617: @nicholascioli let's chat about this when you get a second — this is a bit hacky and there might be another way

apollo-router/src/plugins/connectors/testdata/interface-object.yaml
```diff
@@ -0,0 +1,48 @@
+# rover supergraph compose --config apollo-router/src/plugins/connectors/testdata/interface-object.yaml > apollo-router/src/plugins/connectors/testdata/interface-object.graphql
+# the result is manually edited because 2.10 disallows interface object
+federation_version: =2.10.0-alpha.1
```
dylan-apollo, line 3:
```suggestion
federation_version: =2.10.0-alpha.2
```

Set to the next release version, though if we don't get this change in before tomorrow, we'll want to bump to `.3` probably just to set future testers' expectations

apollo-router/src/plugins/connectors/testdata/interface-object.yaml
```diff
@@ -0,0 +1,48 @@
...
+subgraphs:
+  connectors:
+    routing_url: none
+    schema:
+      sdl: |
+        extend schema
+          @link(url: "https://specs.apollo.dev/federation/v2.10", import: ["@key"])
```
dylan-apollo, line 10:
```suggestion
          @link(url: "https://specs.apollo.dev/federation/v2.10", import: ["@key", "@interfaceObject"])
```

apollo-router/src/plugins/connectors/testdata/interface-object.yaml
```diff
@@ -0,0 +1,48 @@
...
+        type Query {
+          x: [Itf2]
+            @connect(
+              source: "json"
+              http: { GET: "/itf" }
+              selection: "id"
+            )
+        }
+
+        type Itf2 @key(fields: "id") {
```
dylan-apollo, line 26:
```suggestion
        type Itf2 @key(fields: "id") @interfaceObject {
```

Is this how this works? I haven't used it since it first came out 😓

apollo-federation/src/sources/connect/expand/mod.rs
```diff
@@ -614,6 +614,32 @@ mod helpers {
...
+                // If we don't have a key, but the type is an interfaceObject
+                // we'll try just copying over the defined keys.
+                let output_ty = output_type.get(self.original_schema.schema())?;
+                let has_interface_object = output_ty
+                    .directives()
+                    .iter()
+                    .any(|d| d.name == name!("federation__interfaceObject"));
```
dylan-apollo, line 624: I'm not positive how this works, is this name guaranteed, or can it be renamed? Should we instead be looking to see that the output type is an `interface` and has `join__type(isInterfaceObject: true)`?

apollo-federation/src/sources/connect/expand/mod.rs
```diff
@@ -614,6 +614,32 @@ mod helpers {
...
+                if has_interface_object {
+                    let keys = output_ty
+                        .directives()
+                        .iter()
+                        .filter(|d| d.name == name!("federation__key"));
```
dylan-apollo, line 629: Same sort of questions here. Also, I don't think this branch is tested yet since the test has `@key`. Can you write a second test that is missing the `@key`?

apollo-router/src/plugins/connectors/testdata/interface-object.yaml
lennyburdette: yeah — what i did was compose without the interfaceObject (and using `Itf` vs `Itf2`) and then manually editing the supergraph. i can add this behind a comment.

apollo-federation/src/sources/connect/expand/mod.rs
lennyburdette: good question, i'll confirm.

lennyburdette: so expansion always happens on subgraphs extracted from the supergraph, which right now defines directives using [`new_empty_fed_2_subgraph_schema()` ](https://github.com/apollographql/router/blob/dev/apollo-federation/src/supergraph/schema.rs#L48) — there are no user-defined imports at this point.

but i should use constants regardless

apollo-federation/src/sources/connect/expand/mod.rs
```diff
@@ -614,6 +614,32 @@ mod helpers {
...
+                    for key in keys {
+                        match &output_type {
+                            TypeDefinitionPosition::Object(o) => {
+                                o.insert_directive(to_schema, key.clone())
+                            }
+                            TypeDefinitionPosition::Interface(i) => {
+                                i.insert_directive(to_schema, key.clone())
+                            }
+
+                            _ => Ok(()),
```
nicholascioli, line 639: Do we need to worry about types marked with `@interfaceObject` that aren't actually an interface nor object? I would assume that validation catches that, but maybe it would still be a good idea to return an error here.

nicholascioli (commented):
Nice! I left a small comment, but other than that I agree with what @dylan-apollo brought up.

lennyburdette:
moving this to draft. i added more fields to the interface object and now extract_subgraphs_from_supergraph is failing. i guess we need to teach it about interface object

apollo-federation/src/sources/connect/expand/mod.rs
lennyburdette: @interfaceObject's only valid schema location is OBJECT so this cannot happen

lennyburdette:
i fixed all the issues with suboptimal code (too many clones and repetition in merge.rs). still keeping this as draft and we can revisit in october

lennyburdette:
i renamed this branch now that it includes a few different merge fixes

apollo-federation/src/supergraph/mod.rs
```diff
@@ -2226,14 +2226,56 @@ fn extract_join_directives(
         }
     }
 
+    for intf_field_pos in &join_directives.interface_fields {
```
dylan-apollo, line 2229: This is the _exact_ file which caused the `rustc` bug before, so please move this new content into a new file 😁

dylan-apollo (changes requested):

apollo-federation/src/supergraph/mod.rs
lennyburdette: good point!

dylan-apollo (approved):
I've only skimmed the changes, but if they fix the bugs we hit, we should get this merged!

My only other thought is that maybe folks will want this on `dev` if they're doing any other work on merge code? I think I remember you saying that they were going to start from scratch, though

lennyburdette:
i'll confirm but my understanding is that merge.rs will be thrown away

PR #5950, opened 2024-09-04
fix(federation): always update conditions on FetchSelectionSet updates

dariuszkuc:
In JS implementation, conditions were recalculated at the very end when we were creating final selection set. We were missing condition update in one of the update paths.

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-federation/src/query_plan/fetch_dependency_graph.rs
+++ b/apollo-federation/src/query_plan/fetch_dependency_graph.rs
@@ -2479,7 +2479,8 @@ impl FetchDependencyGraphNode {
         })?;
 
         // this function removes unnecessary pieces of the query plan requires selection set.
-        fn trim_requires_selection_node(
+        // PORT NOTE: this function was called trimSelectioNodes in the JS implementation
+        fn trim_requires_selection_set(
             selection_set: &executable::SelectionSet,
         ) -> Vec<executable::Selection> {
             selection_set
@@ -2488,7 +2489,7 @@ impl FetchDependencyGraphNode {
                 .filter_map(|s| match s {
                     executable::Selection::Field(field) => Some(executable::Selection::from(
                         executable::Field::new(field.name.clone(), field.definition.clone())
-                            .with_selections(trim_requires_selection_node(&field.selection_set)),
+                            .with_selections(trim_requires_selection_set(&field.selection_set)),
                     )),
                     executable::Selection::InlineFragment(inline_fragment) => {
                         let new_fragment = if inline_fragment.type_condition.is_some() {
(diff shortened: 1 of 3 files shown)


apollo-federation/src/query_plan/fetch_dependency_graph.rs
```diff
@@ -2478,6 +2478,38 @@ impl FetchDependencyGraphNode {
...
+                        } else {
+                            executable::InlineFragment::without_type_condition(
+                                inline_fragment.selection_set.ty.clone(),
+                            )
+                        }
+                        .with_selections(trim_requires_selection_set(
+                            &inline_fragment.selection_set,
+                        ));
+                        Some(executable::Selection::from(new_fragment))
+                    }
```
TylerBloom, line 2508: nit: this can be more cleanly expressed with `Option` methods:
```suggestion
                    executable::Selection::InlineFragment(inline_fragment) => {
                        let new_fragment = inline_fragment
                            .type_condition
                            .clone()
                            .map(executable::InlineFragment::with_type_condition)
                            .unwrap_or_else(|| {
                                executable::InlineFragment::without_type_condition(
                                    inline_fragment.selection_set.ty.clone(),
                                )
                            })
                            .with_selections(trim_requires_selection_set(
                                &inline_fragment.selection_set,
                            ));
                        Some(executable::Selection::from(new_fragment))
                    }
```

apollo-federation/src/query_plan/fetch_dependency_graph.rs
```diff
@@ -2881,6 +2913,9 @@ impl FetchSelectionSet {
 
     fn add_selections(&mut self, selection_set: &Arc<SelectionSet>) -> Result<(), FederationError> {
         Arc::make_mut(&mut self.selection_set).add_selection_set(selection_set)?;
+        // TODO: when calling this multiple times, maybe only re-compute conditions at the end?
+        // Or make it lazily-initialized and computed on demand?
```
TylerBloom, line 2917: Is there a ticket for this? I know there is a similar note in `add_at_path`. If so, maybe we should make a note on that ticket.

TylerBloom (approved):
LGTM! Only one small nit. Good catch.

apollo-federation/src/query_plan/fetch_dependency_graph.rs
dariuszkuc: Unsure if there is one, I just copied over same comment so when folks look at the TODOs they can tackle both at the same time. The only place we are reading it is in `process_node` so I guess there is not much benefit of storing it on the `FetchSelectionSet` as we can always just compute it directly from its selection set (this is what we do in this assignment). Which in turn brings a question do we even need a `FetchSelectionSet` wrapper in the first place (as it would only store selection set).

Issue #5648, opened 2024-07-12
GRAPHQL_VALIDATION_FAILED returned when a field with variables is filtered out due to authN/Z directives


kschelonka:
## Describe the bug

When 

* an unauthorized portion of a query is removed during filtering (e.g. due to the `@requiresScope` directive)
* and the removed portion references a variable
* and some of the query still remains after the filter step (i.e. the query is not fully unauthorized)

then an error with GRAPHQL_VALIDATION_FAILED is returned and the entire query is rejected instead.

## To Reproduce

1. Set up a router + graph with an authorization directive such as @requiresScopes, which accepts an input

Example: 
```
Query {
   returnMe(var: String!): String @requiresScopes(scopes: [["role:admin"]])
   sayHi: String
}
```

2. Submit a request which requests an unauthorized field with a term, but some of the query should still succeed

Example:

variables = `{"var": "Send this back"}`

```
query returnAndSayHi($var: String!) {
  returnMe(var: $var)
  sayHi
}
```

3. GRAPHQL_VALIDATION_FAILED is returned instead of partial data with an array of UNAUTHORIZED_FIELD_OR_TYPE error(s)

```
{
  "errors": [
    {
      "message": "Variable \"$var\" is never used.",
      "extensions": {
        "code": "GRAPHQL_VALIDATION_FAILED"
      }
    }
  ]
}
```

## Expected behavior
For the authorized components of the query to execute and return data, and the rejected components to be returned as an array of UNAUTHORIZED_FIELD errors.

Example: 

```
{
  "data": {
    returnMe: null,
    sayHi: "Hi!"
  },
  "errors": [
    {
      "message": "Unauthorized field or type",
      "path": [
        "returnMe",
        "@"
      ],
      "extensions": {
        "code": "UNAUTHORIZED_FIELD_OR_TYPE"
      }
    }
  ]
}
```

## Output
Studio examples of this bug on a dev deployment of Pocket's (Mozilla) graph:

[...]

PR #5952, opened 2024-09-04
Create valid documents with authorization filtering

Geal:
Closes https://github.com/apollographql/router/issues/5648

## Problem

To support the router's authorization feature, we have a "transform" query visitor that can be used to modify the query. In authorization, it is used to remove the parts of the query that are not authorized, resulting in null values being inserted in their place in the client response.
That visitor was generating invalid queries for 2 reasons:
- some fragments may not be used anymore because the selection wrapping the fragment spread was removed. Example:
```graphql
query {
  a {
    ... F
  }
  b
}

fragment F on A {
  c
}
```

would be tranformed to this if we removed the `a` field, resulting in an unused fragment:
```graphql
query {
  b
}

fragment F on A {
  c
}
```
while we would expect this instead:
```graphql
query {
  b
}
```
- variables used in removed selections have to be removed from the query's arguments. Example:
```graphql
query($arg1:Int, $arg2:String, $arg3:Int) {
  a(arg: $arg1) {
    c @directive(arg: $arg2)
  }
  b(arg: $arg3)
}
```
would be transformed to this if we removed the `a` field, resulting in `$arg1` and `$arg2` being unused in the query:
```graphql
query($arg1:Int, $arg2:String, $arg1:Int) {
  b(arg: $arg3)
}
```
while we would expect this instead:
```graphql
query($arg1:Int) {
  b(arg: $arg3)
}
```

## Notes
To fix this, we have to take special care in how it is implemented, for the following reasons:
- the visitor is supposed to run in one pass, but this runs at odds with managing dependencies between operations (there can be multiple operations per query document) and fragments, and dependencies between fragments (a fragment can require an other through a spread in its selections)
- the visitor is first going through the list of fragments and producing their modified version in the end document, then the list of operations, but maybe after processing operations we would realize that we don't need some of the fragments
- the transform visitor should take care of cleaning up unused fragments, unused variables, but also removing fragment spreads for removed fragments, and do all of that without requiring special code from the `Visitor` trait implementers
- some of the `Visitor` implementers, like authorization filters, need to have all the information on fragments at the point of application. As an example, in `... F`, if we defined `fragment F on Obj` with `Obj` being unauthorized for this query, or the fragment being removed because all of its selections were removed, the visitor must have processed the fragment first. And that also applies when one fragment inserts another fragment

## Proposed solution
The visitor now follows this process:
- make a first light pass through the query's fragments, to get the dependencies between fragments, and reorder them, so that the first fragments do not use any other fragments, the next ones only use the previous ones in the list, and so on
- go through the ordered list of fragments:
  - modify their selections if needed
  - remove the fragment entirely if needed
  - record the list of variables it uses in field arguments and directives
  - record the list of fragments it uses
  - store the fragment but do not output it to the final document yet
- go through the list of operations:
  - modify their selections if needed
  - record the list of variables it uses in field arguments and directives
  - record the list of fragments it uses
  - for each fragment it uses:
    - get the list of variables it uses, add to the operation's list
    - get the list of fragments it uses, add to the operation's list, and go through their list of fragments and variables as well
  - filter the operation's variables, only keeping the ones we use
  - output the modified operation to the final document
- output to the final document all of the fragments that are used in at least one query

TODO:
- [x] one test is not passing yet: introspection queries should not be filtered. But maybe that should be handled directly in `bridge_query_planner.rs` ? Maybe in a follow up PR?

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-211]: https://apollographql.atlassian.net/browse/ROUTER-211?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ROUTER-690]: https://apollographql.atlassian.net/browse/ROUTER-690?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[ROUTER-437]: https://apollographql.atlassian.net/browse/ROUTER-437?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Geal:
while working on this I uncovered some issues linked to fragment ordering: when encountering a fragment spread, authorization plugins need the referenced fragment to have been already processed, but this is not always the case if we process them in the order they are defined in. As an example:

```graphql
query {
 ... F
}

fragment F on Query {
  ... G
}

fragment G on Query {
  op
}
```

In this query, when we look at the `... G` fragment spread, `G` has not been processed yet, so we don't know which type it applies on, or if we keep that fragment: maybe all of its selection were removed, so we need to remove the fragment and all related spreads.

This is fixed in this PR by first going through the fragments and the spreads in their selections, then ordering the fragments by their dependencies (first processing fragments that do not reference other fragments) and porocessing them, then processing operations.

BrynCooke (approved):
I ran a few more tests locally and everything worked a treat.

apollo-router/src/spec/query/transform.rs
```diff
@@ -158,6 +292,27 @@ pub(crate) fn field(
...
     };
+
+    for argument in def.arguments.iter() {
+        if let Some(var) = argument.value.as_variable() {
+            visitor
+                .state()
+                .used_variables
+                .insert(var.as_str().to_string());
+        }
+    }
```
IvanGoncharov, line 303: Values can have variables deeply nested, and there can be multiple variables per argument.
```
query Test($a: String, $b: String) {
  foo(arg: { someField: [[$a, "bar", $b]] })
}
```

It's works for arguments in both fields and directives

PR #5956, opened 2024-09-04
Add support for include_subgraph_errors to connectors

pubmodmatt:
If `include_subgraph_errors` if false for a subgraph containing connectors, any errors returned by connectors HTTP requests will be redacted from the response. Note that if the connectors debugging feature is enabled, information about the failed request will still be returned in the debug data.

Note that because the `include_subgraph_errors` plugin doesn't know about connectors, and connectors don't have access to the `include_subgraph_errors` plugin, there is no direct way to do the redaction. Instead, the connectors feature stores the subgraph name on the error as an extension, and the plugin checks for this extension to see if the error should be redacted.

With `include_subgraph_errors` set to `true`:

```json
{
  "data": null,
  "errors": [
    {
      "message": "HTTP fetch failed from 'posts.jsonPlaceholder http: GET /users/foo': 404: Not Found",
      "path": [],
      "extensions": {
        "service": "posts.jsonPlaceholder http: GET /users/foo",
        "reason": "404: Not Found",
        "http": {
          "status": 404
        }
      }
    }
  ],
  "extensions": {
    "valueCompletion": [
      {
        "message": "Cannot return null for non-nullable field Query.users",
        "path": []
      }
    ]
  }
}
```

With `include_subgraph_errors` set to `false`:

```json
{
  "data": null,
  "errors": [
    {
      "message": "Subgraph errors redacted",
      "path": []
    }
  ],
  "extensions": {
    "valueCompletion": [
      {
        "message": "Cannot return null for non-nullable field Query.users",
        "path": []
      }
    ]
  }
}
```

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[CNN-431]: https://apollographql.atlassian.net/browse/CNN-431?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apollo-router/src/plugins/include_subgraph_errors.rs
+++ b/apollo-router/src/plugins/include_subgraph_errors.rs
@@ -1,14 +1,20 @@
 use std::collections::HashMap;
+use std::future;
 
+use futures::stream;
+use futures::StreamExt;
 use schemars::JsonSchema;
 use serde::Deserialize;
 use tower::BoxError;
+use tower::ServiceBuilder;
 use tower::ServiceExt;
 
 use crate::json_ext::Object;
 use crate::plugin::Plugin;
 use crate::plugin::PluginInit;
 use crate::register_plugin;
+use crate::services::execution;
+use crate::services::fetch::SubgraphNameExt;
 use crate::services::subgraph;
 use crate::services::SubgraphResponse;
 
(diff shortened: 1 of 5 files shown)


apollo-router/src/plugins/include_subgraph_errors.rs
```diff
@@ -73,6 +79,33 @@ impl Plugin for IncludeSubgraphErrors {
         }
         service
     }
+
+    fn execution_service(&self, service: execution::BoxService) -> execution::BoxService {
```
lennyburdette, line 83: it sure would be nice to do this in a fetch_service hook, huh! maybe add a todo comment suggesting that?

lennyburdette (approved):

PR #5958, opened 2024-09-04
[docs] Update kubernetes docs about logging into helm

smyrick:
Logging into helm is not required to pull public images and charts


shorgi (approved):

abernix (approved):
If @shorgi agrees, I agree.

Issue #2687, opened 2023-02-24
Router treats subgraph 4xx responses as errors


Meemaw:
This is more of a question, than a bug/issue. Currently router treats subgraph/upstream 4xx responses as errors. Those are logged, and will also mark traces as errors.

This might be fine, as according to spec all responses should be 200. However, a lot of GraphQL servers do not 100% comply to the spec.

A common example of a 400 error from a subgraph is a custom scalar that router does not have enough info about to validate, but the subgraph will reject it in the validation phase.

This will result in a high error rate in router, which makes observability/monitoring/alerting a challenge.

What is your stance on this?

Geal:
We have to follow the specification on this. It would make more sense for the subgraph to return a graphql error (with a 200 status) for this validation, because that can then be integrated with the rest of the response.
With https://github.com/apollographql/router/pull/2524 though, it should be possible to hook at the subgraph service stage. Maybe a `map_error` on the rhai side could make sense too, but we have not looked at that yet

BrynCooke:
Related to #3328

abernix:
I'm going to close this for now.  Thanks for opening it originally, @Meemaw.  We've made a number of changes on this behavior since Feb (see linked issues), but we're happy to continue the discussion if you have new thoughts. (Probably open a GH Discussion to discuss! 😄 )

PR #5963, opened 2024-09-05
Corrected insertion logic in `add_at_path`

TylerBloom:
In a recent change that added logic in `add_at_path` to remove unnecessary directives from selections, the insertion logic into the selection set had a bug. If the set did not have the key of the `OpPathElement`, we would generate a selection from the element after removing unnecessary directives. However, this was using the `entry` API, which requires that the key of the `Selection` passed to `.or_insert` matches the original. Because we removed the unnecessary directives, these keys were prone to mismatching.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


lrlna (dismissed):
could this have a test? 😇

apollo-federation/src/operation/mod.rs
```diff
@@ -2636,18 +2636,23 @@ impl SelectionSet {
...
                             element,
                             // We immediately add a selection afterward to make this selection set
                             // valid.
                             Some(SelectionSet::empty(self.schema.clone(), sub_selection_type)),
                             Some(&unnecessary_directives),
-                        )
-                    })?;
+                        )?;
+                        let key = selection.key();
+                        let _ = target.insert(selection);
```
dariuszkuc, line 2652: do we need to check if this entry already exists?

TylerBloom: You're right. We do need to check for this. If one exists, do we merge them together?

dariuszkuc: since we would attempt to insert empty selection set I think we should just get the existing entry

apollo-federation/src/operation/mod.rs
```diff
@@ -2632,34 +2642,49 @@ impl SelectionSet {
...
-                    SelectionValue::InlineFragment(fragment) => fragment
-                        .get_selection_set_mut()
-                        .add_at_path(path, selection_set)?,
-                    SelectionValue::FragmentSpread(_fragment) => {
-                        return Err(FederationError::internal("add_at_path encountered a named fragment spread which should never happen".to_string()));
+                        )?;
+                        target
+                            .entry(selection.key())
+                            .try_and_modify(|selection| {
+                                add_at_path_inner(selection, path, selection_set)
```
dariuszkuc, line 2682: wouldn't this effectively try to add at path twice? I think we just want to get a reference to the element here (or insert it) and then just do single call to `add_at_path_inner` at line #2687 (IMHO there is no need to have that extracted to a method)

TylerBloom: Ah, you're right. What should we do if we discover an existing entry? Do nothing and return the selection?

> IMHO there is no need to have that extracted to a method
Agreed, I only extracted it because I was using it in two locations.

dariuszkuc: Yeah I think we just return the existing selection - then it should follow the same logic regardless whether we remove unneccessady directives or not (ie get an entry or try inserting new one)

apollo-federation/src/operation/mod.rs
```diff
@@ -789,7 +799,7 @@ impl Selection {
     pub(crate) fn from_element(
         element: OpPathElement,
         sub_selections: Option<SelectionSet>,
-        unnecessary_directives: Option<&DirectiveList>,
+        unnecessary_directives: Option<&HashSet<Node<apollo_compiler::ast::Directive>>>,
```
dariuszkuc, line 802: why `HashSet` and not just `DirectiveList`?

apollo-federation/src/operation/mod.rs
```diff
@@ -632,6 +632,16 @@ mod selection_map {
                 Self::Vacant(entry) => entry.insert(produce()?),
             }
         }
+
+        pub fn try_and_modify(
```
dariuszkuc, line 636: it doesn't look like this is used anymore?

lrlna (approved):

apollo-federation/src/operation/mod.rs
TylerBloom: Yes, this avoids duplicates, but more importantly, `DirectiveList`s don't provide a way to be mutated. You must allocate a new list if you want to "mutate" it.

PR #5966, opened 2024-09-05
connectors: add debug cli for expansion

nicholascioli:
This PR adds a new CLI option to apollo-federation to allow for just expanding supergraphs to either stdout or a specified folder.

```bash
cargo run -p apollo-federation-cli -- expand <SUPERGRAPH>
```

The command also supports specifying a prefix to limit which expanded subgraphs to show using the flag `--filter-prefix <PREFIX>`.

__Note: This is mainly for debugging expansion while we wait for a source-aware future. Do NOT rely on the output of this command for anything other than debugging or reference.__

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-federation/cli/src/main.rs
+++ b/apollo-federation/cli/src/main.rs
@@ -11,7 +11,10 @@ use apollo_federation::error::SingleFederationError;
 use apollo_federation::query_graph;
 use apollo_federation::query_plan::query_planner::QueryPlanner;
 use apollo_federation::query_plan::query_planner::QueryPlannerConfig;
+use apollo_federation::sources::connect::expand::expand_connectors;
+use apollo_federation::sources::connect::expand::ExpansionResult;
 use apollo_federation::subgraph;
+use apollo_federation::Supergraph;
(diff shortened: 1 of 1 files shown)


apollo-federation/cli/src/main.rs
```diff
@@ -295,6 +320,64 @@ fn cmd_extract(file_path: &Path, dest: Option<&PathBuf>) -> Result<(), Federatio
...
+            })?;
+        }
+    } else {
+        for (name, subgraph) in subgraphs {
+            // Skip any files not matching the prefix, if specified
+            if should_skip(&name) {
+                continue;
+            }
+
+            println!("[Subgraph `{}`]", name);
```
lennyburdette, line 372: could you print this in YAML format so we can immediately pipe it into `rover supergraph compose`?

lennyburdette (approved):

PR #5969, opened 2024-09-06
rest: *

lennyburdette:
---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [x] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-federation/src/sources/connect/validation/selection.rs
```diff
@@ -392,3 +394,105 @@ impl SelectionValidator<'_, '_> {
...
+                return Some(Message {
+                    code: Code::InvalidStarSelection,
+                    message: format!(
+                        "{coordinate} contains `{field_name}: *` but the field `{parent_type}.{field_name}: {ty}` returns a list. It must be a non-list custom scalar.",
+                        coordinate = selection_coordinate,
+                        field_name = field_name,
+                        parent_type = group.ty.name,
+                        ty = definition.ty
+                    ),
+                    locations: selection_location.iter().cloned().collect(),
```
dylan-apollo, line 443: I usually also include the location of the field definition, to point users at where they should modify the type.

apollo-federation/src/sources/connect/validation/selection.rs
```diff
@@ -392,3 +394,105 @@ impl SelectionValidator<'_, '_> {
...
+                        parent_type = group.ty.name,
+                        ty = definition.ty,
+                        ty_kind = match ty {
+                            ExtendedType::Object(_) => "an object",
+                            ExtendedType::Interface(_) => "an interface",
+                            ExtendedType::Union(_) => "a union",
+                            ExtendedType::Enum(_) => "an enum",
+                            ExtendedType::InputObject(_) => "an input object",
+                            ExtendedType::Scalar(_) => "a scalar",
+                        }
```
dylan-apollo, line 475: We should probably make this a helper so we can also use this in other validations. I think right now we just say "is not an object" or whatever for leaf agreement

apollo-federation/src/sources/connect/validation/selection.rs
```diff
@@ -392,3 +394,105 @@ impl SelectionValidator<'_, '_> {
...
+                    ),
+                    locations: selection_location.iter().cloned().collect(),
+                });
+            }
+
+            if ty.is_built_in() {
```
dylan-apollo, line 481: Is `String` not okay? We are stringifying whatever is put in there, right?

apollo-router/src/plugins/connectors/tests.rs
```diff
@@ -452,13 +456,23 @@ async fn test_root_field_plus_entity_plus_requires() {
             "id": 1,
             "name": "Leanne Graham",
             "username": "Bret",
-            "d": "1-770-736-8031 x56442"
+            "d": "1-770-736-8031 x56442",
+            "rest": {
+              "phone": "1-770-736-8031 x56442",
+              "email": "Sincere@april.biz",
+              "website": "hildegard.org"
+            }
```
dylan-apollo, line 464: Shouldn't this return a stringified version? Does GraphQL let you return objects for custom scalars?

lennyburdette: Custom scalars can be literally anything, as long as it can be serializable by the transport encoding!

I hadn’t noticed till now that the guild provides both a JSON scalar (stringified) and a JSONObject scalar (raw json values). 

I’m not sure how we’d support stringifying. `rest: *->stringify` maybe?

lennyburdette: oh nope i'm wrong. they have json for any json value, and jsonobject for specifically json objects: https://github.com/taion/graphql-type-json/blob/master/src/index.js. no JSON.parse() anywhere in the file

apollo-federation/src/sources/connect/expand/tests/schemas/steelthread.graphql
```diff
@@ -1,6 +1,7 @@
 schema
   @link(url: "https://specs.apollo.dev/link/v1.0")
-  @link(url: "https://specs.apollo.dev/join/v0.4", for: EXECUTION)
+  @link(url: "https://specs.apollo.dev/join/v0.5", for: EXECUTION)
```
nicholascioli, line 3: Not really a comment on the impl, but is there a reason that our docs for join stop after `v0.3`?

[This link](https://specs.apollo.dev/join/v0.5/) results in a 404 :(

apollo-federation/src/sources/connect/validation/selection.rs
nicholascioli: I thought that the idea behind star selections is that it should really only be there for debugging / collection purposes. If we allow the type to be `String` then I worry that people might start depending on the stringification to follow a specific layout vs. just being an opaque type. Either way, we should document what sort of expectations users should have about star selections.

apollo-federation/src/sources/connect/validation/selection.rs
nicholascioli, line 1: For all of the errors around expected types for star selections, should we also include a suggestion? Something simple like `e.g. scalar RestFields` or similar might at least lead users to a quick solution.

nicholascioli (approved):
This looks great! Just a few nits, but can be ignored.

apollo-federation/src/sources/connect/expand/tests/schemas/steelthread.graphql
lennyburdette: yeah we need to keep these up to date better. they're very much internal tho

PR #5982, opened 2024-09-10
feat: allow users to load apollo key from file

abernix:
> [!NOTE]
> This opens up @lleadbet's https://github.com/apollographql/router/pull/5917 again.

Customers sometimes would rather not pass sensitive keys to the router through environment variables out of an abundance of caution related to concerns about leaking them.

This PR allows users to pass a new argument `--apollo-key-path` that takes a file location as an argument, and uses it as the APOLLO_KEY value for use with Uplink and usage reporting.

Addresses a portion of #3264, specifically the APOLLO_KEY.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

This does not fully address #3264 as it doesn't support APOLLO_GRAPH_REF, but addresses the most significant concern of that issue.

**Notes**

I'm unsure where both tests and docs should be updated here; if there's a good place, let me know.

### DIFF
--- a/.changesets/feat_feat_key_from_file.md
+++ b/.changesets/feat_feat_key_from_file.md
@@ -4,4 +4,6 @@ Users sometimes would rather not pass sensitive keys to the router through envir
 
 This addresses a portion of #3264, specifically the APOLLO_KEY.
 
+Note: This feature is not available on Windows.
+
 By [@lleadbet](https://github.com/lleadbet) in https://github.com/apollographql/router/pull/5917
--- a/docs/source/configuration/overview.mdx
+++ b/docs/source/configuration/overview.mdx
@@ -61,6 +61,8 @@ The [graph API key](/graphos/api-keys/#graph-api-keys) that the router should us
 </td>
 <td>
 
+⚠️ **This is not available on Windows.**
+
 A path to a file containing the [graph API key](/graphos/api-keys/#graph-api-keys) that the router should use to authenticate with GraphOS when fetching its supergraph schema.
 
 **Required** when using [managed federation](/federation/managed-federation/overview/), except when using an [offline license](#--license) to run the router or when using `APOLLO_KEY`.
@@ -139,6 +141,8 @@ The absolute or relative path to the router's optional [YAML configuration file]
 
 </td>
 <td>
+
+⚠️ **This is not available on Windows.**
 
 The absolute or relative path to a file containing the Apollo graph API key for use with managed federation.
 


.circleci/config.yml
```diff
@@ -1,5 +1,4 @@
 version: 2.1
-
 # These "CircleCI Orbs" are reusable bits of configuration that can be shared
```
abernix, line 2:
```suggestion

# These "CircleCI Orbs" are reusable bits of configuration that can be shared
```

garypen (changes requested):
Let's follow good security practice, e.g.: `ssh`, and check if the permissions of the key file allow only the owner to read the file.

Not sure what that means on Windows, but let's figure it out.

lleadbet:
> Not sure what that means on Windows, but let's figure it out.

It's a good question since none of the native libraries from what I can tell offer Windows-specific APIs for this, and the *nix libraries only apply to the respective OS. 

The `std` lib, for example, doesn't provide bindings like it does for Unix: 

* Windows: https://doc.rust-lang.org/std/os/windows/fs/index.html
* Unix: https://doc.rust-lang.org/std/os/unix/fs/index.html

Third-party wise, there's only a couple of libraries from what I can find in varying states of support: 

* https://crates.io/crates/faccess
* https://crates.io/crates/windows-permissions

Both of which haven't had much in terms of releases or updates in the past few years (nor star count, if that's a helpful health indicator). 

Doing so for *nix seems sane- just checking that the last two in the file mode is `00` (e.g. `400` or `600`), but Windows differences would be something we should consider if it's going to create a weird experience for devs locally vs. hosted.

garypen:
@lleadbet I'll assign myself to look into this further. I'll update this PR when I have some ideas. Thanks for the suggestions around Windows.

garypen:
I did some brief digging yesterday and concluded that it's not straightforward to add support for file mode and effective user id checking on Windows. We could definitely do this, but I don't think it's worth the effort unless a users specifically requests it.

I propose that we either modify this feature to be "Unixen" specific (i.e.: no flag for Windows) or note in documentation that extended security checks are performed on "Unixen" (file mode and effective uid) which are not performed on Windows. My preference is for the first option, since it is least confusing, unless we are sure that there is demand for this flag on Windows.

BrynCooke (changes requested):
The integration tests now spin up a router as a separate process. It should be possible to add support for passing args and therefore a test.

Here's the relevant section of code: https://github.com/apollographql/router/blob/dev/apollo-router/tests/common.rs#L381-L401

garypen:
I've added an integration test for the "loose file permissions" part of this work. I haven't got one for the "not the correct user id", because that's difficult to do in an automated test. I have tested it manually though and the code is "correct by visual inspection".

garypen (approved):

abernix:
@BrynCooke can you take another look

BrynCooke (approved):

PR #5991, opened 2024-09-10
Validate URL templates

dylan-apollo:
Based on https://github.com/apollographql/router/pull/5989, so review & merge that first.

The user-facing changes are:

1. Composition will now detect when a URL template is completely invalid syntax (e.g., forgot a closing `}`)
2. Composition will detect when a URL template contains an unsupported variable name—so doesn't start with `$config`, `$args`, or `$this`
3. Router will now fail at _startup_ if those problems occur, rather than when it tries to hit the rest endpoint. Because composition catches the problems, this should also never happen.
4. Allows two variables to be next to each other, since we didn't end up needing that limitation
5. Stops including the undocumented `!` syntax in query plans

Internally, this PR also:

1. Adds location data to some URLTemplate errors and all URLTemplate variables (for use in future validations)
2. Refactors some coordinates into `Copy` structs instead of pre-rendering and passing around strings
3. Deletes all the URL template syntax which is undocumented or unsupported, including:
    1. Batch syntax
    2. Required `!` 
    3. Non-global variables (i.e., `{blah}` instead of `{$args.blah}`)

Left to do (but this PR is already huge):

1. Verify that `$args` & `$this` match the GraphQL schema
2. Enforce (or hint?) that path parameters should come from non-nullable GraphQL types


apollo-federation/src/sources/connect/validation/http/url.rs
```diff
@@ -0,0 +1,85 @@
...
+    value: &Node<Value>,
+    coordinate: HttpMethodCoordinate,
+    sources: &SourceMap,
+) -> Result<URLTemplate, Message> {
+    let str_value = require_value_is_str(value, coordinate, sources)?;
+    let template =
+        URLTemplate::from_str(str_value).map_err(|url_template::Error { message, location }| {
+            Message {
+                code: Code::InvalidUrl,
+                message: format!("{coordinate} must be a valid URL template. {message}",),
```
lennyburdette, line 28: i've noticed a few trailing commas, is this a jetbrains formatting thing?

apollo-federation/src/sources/connect/validation/http/url.rs
```diff
@@ -0,0 +1,85 @@
...
+fn select_substring_location(
+    line_column: Option<Range<LineColumn>>,
+    full_url: &str,
+    substring_location: Option<Range<usize>>,
+) -> Vec<Range<LineColumn>> {
+    line_column
+        .map(|mut template_location| {
+            // The default location includes the parameter name, we just want the value,
+            // so we need to calculate that.
+            template_location.end.column -= 1; // Get rid of the end quote
```
lennyburdette, line 72: what happens with `"""triple quotes"""`?

lennyburdette (approved):
looks great!

apollo-federation/src/sources/connect/validation/http/url.rs
dylan-apollo: ick. No way for me to fix that one, I think, without contributing something to `apollo-compiler`. Have any ideas for a workaround? Or is it okay for that to be ugly for now?

apollo-federation/src/sources/connect/validation/http/url.rs
dylan-apollo: No, it's a `rustfmt` doesn't get rid of them if they're left over from elsewhere thing 🙃. I'll try to hunt them down by hand.

apollo-federation/src/sources/connect/validation/http/url.rs
lennyburdette: fine to be ugly for now! hopefully we can pick up the conversation about semantic locations in apollo-rs after summit

apollo-federation/src/sources/connect/validation/http/url.rs
dylan-apollo: Well, searching globally for `,)` in our repo certain illustrates how common that limitation is. Seems like it only happens in macro invocations. I _think_ I removed all the ones I added here.

apollo-federation/src/sources/connect/url_template.rs
```diff
@@ -429,96 +291,76 @@ impl ParameterValue {
...
-impl Display for ParameterValue {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        for part in &self.parts {
-            part.fmt(f)?;
+/// The supported types of variables for URLs, a subset of `KnownVariable` in JSONSelection
+#[derive(Clone, Copy, Debug, EnumIter, Eq, Hash, Ord, PartialEq, PartialOrd)]
+pub enum VariableType {
+    Args,
+    This,
+    Config,
```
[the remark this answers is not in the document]
dylan-apollo: I think `$context` is only supported in headers right now. Adding support in url template & JSONSelection is probably just a matter of updating those two enums?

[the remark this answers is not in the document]
lennyburdette: gotcha, makes sense!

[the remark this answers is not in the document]
pubmodmatt: Right, its only in headers so far. You need both support in the grammar and adding it to the variables input when doing the interpolation at runtime.

PR #5994, opened 2024-09-11
[JSONSelection] Support parenthesized `ExprPath` syntax to simplify expression injection

benjamn:
You can find a more detailed explanation in the `README.md` changes, and examples in the tests, but the basic idea here is that we'd like to be able to write selections like
```graphql
__typename: $("Product")
```
instead of the previously recommended
```graphql
__typename: $->echo("Product")
```
While `->echo` is still useful, it no longer needs to be used for this particular use case (where the input value `$` is ignored), since wrapping a JSON-like expression in parentheses ensures it will be parsed as a `LitExpr` without ambiguity.

In case you're wondering, the parentheses are necessary because the following selections parse as field selections, not constant-valued properties:
```graphql
# Attempts to select a field called "Product", which probably does not exist
__typename: "Product"

# Attempts to select a field called `true`, which probably does not exist
condition: true
```

You can also continue to append `.key` and `->method` elements after the `$(<LitExpr>)`, as in
```graphql
alphabetSlice: $("abcdefghijklmnopqrstuvwxyz")->slice($args.start, $args.end)
```

Thanks to @lennyburdette for sparking this conversation, as this new syntax feels much better than abusing `$->echo`.


apollo-federation/src/sources/connect/json_selection/apply_to.rs
```diff
@@ -1405,6 +1413,234 @@ mod tests {
...
+                one: (1)
+                two: (2)
+                negativeThree: (-  3)
+                true: (true  )
+                false: (  false)
+                null: (null)
+                string: ("string")
+                array: ( [ 1 , 2 , 3 ] )
+                object: ( { "key" : "value" } )
+                path: (nested.path)
```
lennyburdette, line 1447: the parentheses on this line basically do nothing, right?

benjamn: Yep, just wanted to test they are not harmful when added redundantly.

nicholascioli:
Could you rebase this? I think that GitHub is confused by the changes up (down?) stream.

nicholascioli:
Also, is there a design document around JSONSelection anywhere that I can refer to? I want to make sure that I understand how the various features of the language are supposed to mesh with each other while reviewing this.

nicholascioli:
Have we considered doing something like bash here and making this syntax look more like `$(...)`? I worry that it might not be obvious to users why certain things are wrapped in what appears to be redundant parenthesis at a quick glance.

benjamn:
@nicholascioli Rebased, though I'm seeing one strange snapshot failure locally that might also fail here. Specifically, `sources::connect::expand::tests::it_expand_supergraph` passes locally when I run it by itself, but fails when I run `cargo test`, and then `cargo insta review` says `done: no snapshots to review`. Weird.

The best / most current design document is the [`README.md`](https://github.com/apollographql/router/blob/benjamn/JSONSelection-parenthesized-ExprPath/apollo-federation/src/sources/connect/json_selection/README.md), which includes a [section](https://github.com/apollographql/router/blob/benjamn/JSONSelection-parenthesized-ExprPath/apollo-federation/src/sources/connect/json_selection/README.md#exprpath-) about the new `ExprPath` rule introduced here.

apollo-federation/src/sources/connect/json_selection/README.md
```diff
@@ -89,10 +89,11 @@ NamedPathSelection   ::= Alias PathSelection
 NamedFieldSelection  ::= Alias? Key SubSelection?
 NamedGroupSelection  ::= Alias SubSelection
 Alias                ::= Key ":"
-PathSelection        ::= (VarPath | KeyPath | AtPath) SubSelection?
+PathSelection        ::= (VarPath | KeyPath | AtPath | ExprPath) SubSelection?
 VarPath              ::= "$" (NO_SPACE Identifier)? PathStep*
 KeyPath              ::= Key PathStep+
 AtPath               ::= "@" PathStep*
+ExprPath             ::= "(" LitExpr ")" PathStep*
```
lennyburdette, line 96: oh good, parsing will enforce a single expression here. that's actually a nice improvement over echo()!

benjamn: indeed!

benjamn:
@lennyburdette @nicholascioli @dylan-apollo Switching to `$(...)` went smoothly, though I did have to rearrange the cases in `PathList::parse_with_depth` to avoid parsing `$(...)` as a `$` variable followed by syntax that isn't allowed after a `$` variable (hence a parse error). Clever lookahead logic might have prevented that, but would have been an even more significant change.

I also spiked out the idea of only requiring `(...)` (rather than `$(...)`) when already parsing a `LitExpr` (see https://github.com/apollographql/router/commit/7e458e800492b3f057427d8141f16cad56063a05 on [this branch](https://github.com/apollographql/router/commits/benjamn/JSONSelection-ExprPathInExpr/)), but it significantly complicated the grammar (requiring the `PathSelectionInExpr` and `ExprInExpr` variants) and the parser (adding a `PathParsingMode` to `PathList::Expr` and `PathList::parse_with_depth`). Given that you need this syntax within a `LitExpr` only when you want to apply a `.key` or `->method` to the expression, `$(...)` should be relatively uncommon inside `LitExpr` expressions, so the ergonomic benefits of `(...)` over `$(...)` in that context seem not worth the complications.

apollo-federation/src/sources/connect/json_selection/grammar/Alias.svg
benjamn, line 1: This `grammar/Alias.svg` change snuck in from a previous PR (#5987), since I forgot to update the SVG diagrams.

PR #5995, opened 2024-09-11
Include request variables in demand control scoring

tninesling:
We're missing variable support in demand control scoring, so they're currently not counted. Additionally, they don't get passed through to the list size directive, which prevents users from using variables as slicing arguments. This change includes the request variables in context extensions so they can be accessed from the demand control plugin pipeline.

This includes a small refactor to wrap some of the common scoring arguments (schema, request document, etc) into a scoring context to reduce the number of arguments to each function.

---

**Checklist**

- Tests added and passing[^3]
    - [X] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/json_ext.rs
+++ b/apollo-router/src/json_ext.rs
@@ -479,12 +479,11 @@ impl ValueExt for Value {
         match self {
             Value::Null => apollo_compiler::ast::Value::Null,
             Value::Bool(b) => apollo_compiler::ast::Value::Boolean(*b),
+            Value::Number(n) if n.is_f64() => {
+                apollo_compiler::ast::Value::Float(n.as_f64().expect("is float").into())
+            }
             Value::Number(n) => {
-                if n.is_f64() {
-                    apollo_compiler::ast::Value::Float(n.as_f64().expect("is float").into())
-                } else {
-                    apollo_compiler::ast::Value::Int((n.as_i64().expect("is int") as i32).into())
-                }
+                apollo_compiler::ast::Value::Int((n.as_i64().expect("is int") as i32).into())
             }
             Value::String(s) => apollo_compiler::ast::Value::String(s.as_str().to_string()),
             Value::Array(inner_vars) => apollo_compiler::ast::Value::List(
@@ -508,7 +507,7 @@ impl ValueExt for Value {
     }
 
     fn as_i32(&self) -> Option<i32> {
-        self.as_i64().and_then(|i| i.to_i32())
+        self.as_i64()?.to_i32()
     }
 }
 
--- a/apollo-router/src/json_ext.rs
+++ b/apollo-router/src/json_ext.rs
@@ -5,6 +5,7 @@
 use std::cmp::min;
 use std::fmt;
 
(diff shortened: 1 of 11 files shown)


apollo-router/src/json_ext.rs
```diff
@@ -468,6 +474,42 @@ impl ValueExt for Value {
                     typename == maybe_type || schema.is_subtype(maybe_type, typename)
                 })
     }
+
+    fn to_ast(&self) -> apollo_compiler::ast::Value {
+        match self {
+            Value::Null => apollo_compiler::ast::Value::Null,
+            Value::Bool(b) => apollo_compiler::ast::Value::Boolean(*b),
+            Value::Number(n) => {
+                if n.is_f64() {
```
tninesling, line 483: Is there a better way to do this conversion? This inner `n` is private so it's not so easy to cleanly convert it to the underlying numeric type.

apollo-router/src/plugins/demand_control/mod.rs
```diff
@@ -306,9 +306,10 @@ impl Plugin for DemandControl {
             let strategy = self.strategy_factory.create();
             ServiceBuilder::new()
                 .checkpoint(move |req: execution::Request| {
-                    req.context
-                        .extensions()
-                        .with_lock(|mut lock| lock.insert(strategy.clone()));
+                    req.context.extensions().with_lock(|mut lock| {
+                        lock.insert(strategy.clone());
+                        lock.insert(req.supergraph_request.body().variables.clone());
```
tninesling, line 311: This is going to insert an entry of type `Map<ByteString, Value>` into the context. This type is used both for graphql request variables and graphql response extensions. Should we create some wrapper type to make sure it's disambiguated in context?

apollo-router/src/json_ext.rs
BrynCooke: An alternative is:
```
            Value::Number(n) if n.is_f64() => {
                apollo_compiler::ast::Value::Float(n.as_f64().expect("is float").into())
            }

            Value::Number(n) => {
                apollo_compiler::ast::Value::Int((n.as_i64().expect("is int") as i32).into())
            }
```

apollo-router/src/plugins/demand_control/mod.rs
BrynCooke: Yeah, I agree that a wrapper type needs to be used. Maybe also include cost strategy in it so there is just one entry?

I spent a bit of time trying to figure out why we need variables for the response. This is because input fields can have a cost directive right?

BrynCooke (changes requested):
Needs a wrapper for variables in context.

apollo-router/src/plugins/demand_control/mod.rs
tninesling: Yes, we need to include argument cost for a field when scoring responses, and we can't score arguments purely from the schema because we may over/under count optional fields.

BrynCooke (approved):

PR #5996, opened 2024-09-12
Fix query plan cache gauges that stopped working after hot reload

BrynCooke:
When a hot reload happens the opentelemetry meter provider is refreshed.

Gauges in particular MUST be created after this refresh has happened.

The query planner pipeline is currently created outside of the plugin lifecycle, and therefore does not have the opportunity to do this.

We can consider reintroducing the `query_planner` and `activate` hooks in plugins in future it would probably help, or have some equivalent. But for now this PR just adds an `activate` call that is threaded right down to the `CacheStorage`. This can be called on the query planner pipeline after telemetry has been activated.

This is tested in integration test by starting the router without prometheus and then adding it after a reload.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [x] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*
* Bug fix, no docs needed
* No perf impact

**Notes**

### DIFF
--- a/.changesets/fix_bryn_fix_gauges.md
+++ b/.changesets/fix_bryn_fix_gauges.md
@@ -1,0 +1,9 @@
+### Query plan cache gauges stop working after hot reload ([PR #5996](https://github.com/apollographql/router/pull/5996))
+
+When the router reloads the schema or config, the query plan cache gauges stop working. These are:
+* `apollo.router.cache.storage.estimated_size`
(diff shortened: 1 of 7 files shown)


Geal (approved):

Geal:
I think we can reintroduce query planner hook points, and leave them in the private plugin trait for now

PR #5998, opened 2024-09-12
Run new query planner off of Tokio threads

SimonSapin:
Planning code is synchronous (it never `await`s) and can take non-trivial time, so it is effectively "blocking" from Tokio’s point of view. If too many Tokio threads are busy planning queries, few threads remain available to run async code.

This runs the planner inside `tokio::task::spawn_blocking`, following the precedent of graphql parsing in `query_analysis.rs`. This uses a thread pool of maximum size 512 threads, which may not be what we want. From its documentation, `spawn_blocking` appears to be made for code that often blocks in synchronous syscalls.

Instead we may prefer to have a thread pool for "compute tasks" that defaults to $NUM_CPUS threads, and wrap it with something that uses `tokio::sync::oneshot` to send results back to async code.

### DIFF
--- a/apollo-router/src/query_planner/bridge_query_planner.rs
+++ b/apollo-router/src/query_planner/bridge_query_planner.rs
@@ -287,31 +287,40 @@ impl PlannerMode {
                 }
                 Ok(success)
             }
-            PlannerMode::Rust(rust) => {
-                let start = Instant::now();
+            PlannerMode::Rust(rust_planner) => {
+                let doc = doc.clone();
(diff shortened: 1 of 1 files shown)


bnjjj (approved):

Geal (approved):
as you said in the PR description, this should use a separate pool with a reduced number of threads. Probably lower than the number of CPUs, because we saw issues with the JS planner pool when using its `auto` setting.
Let's note that the query analysis layer has the same issue: it runs under a `spawn_blocking` task, without a specific thread pool

PR #5762, opened 2024-08-01
[`JSONSelection`] Support `->` method syntax for inline data transforms

benjamn:
This PR extends [`JSONSelection` syntax](https://github.com/apollographql/router/blob/benjamn/JSONSelection-method-syntax/apollo-federation/src/sources/connect/json_selection/README.md) to allow any field value (including fields selected by `nested.path.selections`) to be transformed in-line using a "method" syntax that looks like either `field->method` or `field->method(arg1, arg2, ...)`, where `method` is an identifier that specifies which method implementation to invoke against the preceding `field`'s value.

When a `->` method takes arguments, those arguments can be any JSON literal expression, with an additional allowance for `PathSelection` values, so you can refer to `$variables` like `$this` or `$args` or `$` (see #5142) within the JSON literal. This additional capability is why I introduced the `JSLiteral` enum and related `parse*`/`ApplyTo` methods, rather than just reusing `serde_json_bytes::Value`. In some sense, this extension brings the `JSLiteral` syntax closer to JavaScript literal syntax (which can refer to variables) than to JSON syntax, which is why I called it `JSLiteral` rather than `JSONLiteral`.

The implementation of a method has access not only to its arguments, but also to the input value preceding the `->`. Sometimes, you may need to refer to that input value within the method arguments, which is possible using the newly introduced `@` symbol. This syntax was directly inspired by [`JSONPath` syntax](https://www.ietf.org/archive/id/draft-goessner-dispatch-jsonpath-00.html#name-overview-of-jsonpath-expres), which uses the `@` character for analogous purposes (representing the current value within filter expressions).

This syntax was first inspired by the need to operate on arrays, using methods like `array->first`, `array->last`, and `array->slice(0, 3)`, but the set of potential use cases for the syntax has grown rapidly since then, and is very much still growing. You can find examples of all the currently supported methods in [the `README.md`](https://github.com/apollographql/router/blob/benjamn/JSONSelection-method-syntax/apollo-federation/src/sources/connect/json_selection/README.md#pathstep-), and their actual implementations can be found in the [`json_selection/methods.rs`](https://github.com/apollographql/router/blob/benjamn/JSONSelection-method-syntax/apollo-federation/src/sources/connect/json_selection/methods.rs) file.

For now, the set of available methods is fixed by the Rust implementation, and registered in a `lazy_static!` map. In the future, we might make the method list configurable, with a flexible type system to validate inputs and outputs, but for now it seems simpler to restrict this functionality to methods we know about and have carefully reviewed.

[...]

PR #6003, opened 2024-09-13
[JSONSelection] Remove `&mut errors: IndexSet<ApplyToError>` from `apply_to_path` method signatures

benjamn:
Following up on this suggestion from @nicholascioli: https://github.com/apollographql/router/pull/5762#discussion_r1727811782

The `apply_to_path` method used to have more than one mutable parameter (`input_path` as well as `errors`), but now that `input_path` is immutable, I agree it's cleaner to go all the way and avoid any mutable parameters, so `apply_to_path` can be a pure function.

One big benefit of this refactoring is that each `apply_to_path` method now returns just the errors resulting from the subtree that it handled, whereas the previous `&mut errors: IndexSet<ApplyToError>` strategy made it difficult to know which errors came from which `apply_to_path` method call. This will help if we ever decide to memoize the results of `apply_to_path` methods, or if we want to expose `ApplyToError`s for processing by `->` methods (imagine some sort of `->catch` handler that captures/handles errors from further down its `PathList` chain).

As discussed in https://github.com/apollographql/router/pull/5987#discussion_r1755547444, there might not be much point in deduplicating errors now, since their `path` and `range` information effectively makes them all unique, but if we decide to stop the deduplication that's now an easy change within `JSONSelection::apply_with_vars`.

These changes should be entirely an implementation detail. As evidence for this claim, I did not have to change any tests.

### DIFF
--- a/apollo-federation/src/sources/connect/json_selection/apply_to.rs
+++ b/apollo-federation/src/sources/connect/json_selection/apply_to.rs
@@ -58,7 +58,14 @@ impl JSONSelection {
         // selection set was applied to.
         vars_with_paths.insert(KnownVariable::Dollar, (data, InputPath::empty()));
 
-        let value = self.apply_to_path(data, &vars_with_paths, &InputPath::empty(), &mut errors);
+        let (value, apply_errors) = self.apply_to_path(data, &vars_with_paths, &InputPath::empty());
+
+        // Since errors is an IndexSet, this line effectively deduplicates the
+        // errors, in an attempt to make them less verbose. However, now that we
+        // include both path and range information in the errors, there's an
+        // argument to be made that errors can no longer be meaningfully
+        // deduplicated, so we might consider sticking with a Vec<ApplyToError>.
+        errors.extend(apply_errors);
 
         (value, errors.into_iter().collect())
     }
@@ -72,8 +79,7 @@ pub(super) trait ApplyToInternal {
         data: &JSON,
         vars: &VarsWithPathsMap,
         input_path: &InputPath<JSON>,
-        errors: &mut IndexSet<ApplyToError>,
-    ) -> Option<JSON>;
+    ) -> (Option<JSON>, Vec<ApplyToError>);
 
     // When array is encountered, the Self selection will be applied to each
     // element of the array, producing a new array.
@@ -82,20 +88,21 @@ pub(super) trait ApplyToInternal {
         data_array: &[JSON],
         vars: &VarsWithPathsMap,
         input_path: &InputPath<JSON>,
-        errors: &mut IndexSet<ApplyToError>,
-    ) -> Option<JSON> {
+    ) -> (Option<JSON>, Vec<ApplyToError>) {
         let mut output = Vec::with_capacity(data_array.len());
+        let mut errors = Vec::new();
 
         for (i, element) in data_array.iter().enumerate() {
             let input_path_with_index = input_path.append(json!(i));
-            let applied = self.apply_to_path(element, vars, &input_path_with_index, errors);
+            let (applied, apply_errors) = self.apply_to_path(element, vars, &input_path_with_index);
+            errors.extend(apply_errors);
             // When building an Object, we can simply omit missing properties
             // and report an error, but when building an Array, we need to
             // insert null values to preserve the original array indices/length.
             output.push(applied.unwrap_or(JSON::Null));
         }
 
-        Some(JSON::Array(output))
+        (Some(JSON::Array(output)), errors)
     }
 }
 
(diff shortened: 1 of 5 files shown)


apollo-federation/src/sources/connect/json_selection/apply_to.rs
```diff
@@ -148,27 +155,60 @@ impl ApplyToError {
     }
 }
 
+// Rust doesn't allow implementing methods directly on tuples like
+// (Option<JSON>, Vec<ApplyToError>), so we define a trait to provide the
+// methods we need, and implement the trait for the tuple in question.
+pub(super) trait ApplyToResultMethods {
```
dylan-apollo, line 161: Nice trick, I always forget that I can hack in methods like this

apollo-federation/src/sources/connect/json_selection/apply_to.rs
```diff
@@ -148,27 +155,60 @@ impl ApplyToError {
...
+impl ApplyToResultMethods for (Option<JSON>, Vec<ApplyToError>) {
+    // Intentionally taking ownership of self to avoid cloning, since we pretty
+    // much always use this method to replace the previous (value, errors) tuple
+    // before returning.
+    fn prepend_errors(self, errors: &[ApplyToError]) -> Self {
+        if errors.is_empty() {
+            self
+        } else {
+            let (value_opt, apply_errors) = self;
+            (value_opt, [errors, apply_errors.as_slice()].concat())
```
dylan-apollo, line 179: I believe this is reallocating the error vec every single time even if we haven't accumulated any new errors. The good news is, if there are _no_ errors at all, I believe it's basically a no-op. I think, though, if we pass an owned `Vec<ApplyToError>` here and extend it we can save some cycles in the error case.

This should also make sure we aren't forgetting to pass around errors somewhere and doing extra cloning/duplicating, since when we pass ownership, the value is gone and we have to get a new one (the return value of this function)

apollo-federation/src/sources/connect/json_selection/apply_to.rs
```diff
@@ -148,27 +155,60 @@ impl ApplyToError {
...
+        }
+    }
+
+    // A substitute for Option<_>::and_then that accumulates errors behind the
+    // scenes. I'm no Haskell programmer, but this feels monadic? ¯\_(ツ)_/¯
+    fn and_then_collecting_errors(
+        &self,
```
dylan-apollo, line 186: This would also take `self`, I think, if we switch to passing owned values. Which means that `f` _could_ take an owned value if that's useful (would prevent cloning?)

apollo-federation/src/sources/connect/json_selection/helpers.rs
```diff
@@ -78,6 +78,16 @@ pub fn json_type_name(v: &JSON) -> &str {
...
 }
 
+pub fn immutable_vec_push<T>(vec: &[T], item: T) -> Vec<T>
+where
+    T: Clone,
+{
+    let mut new_vec = Vec::with_capacity(vec.len() + 1);
+    new_vec.extend_from_slice(vec);
+    new_vec.push(item);
+    new_vec
```
dylan-apollo, line 88: I'd prefer we just extend the vec directly, if we can. I _believe_ we never want duplicates of these, so it should be a problem to hand around ownership.

Passing `mut Vec<blah>` doesn't carry the same risks as `&mut Vec<blah>`, since you're forced to still return the new value for consumers to get the updates—so the functions are still "pure"

dylan-apollo (approved):
The more functional style feels much nicer to me, thanks! I _think_ it will be a more efficient in the error case to switch to passing owned values instead of references which we reallocate on. Might shake up too many other things, though, so if it ends up being a mess, we can stick with this.

apollo-federation/src/sources/connect/json_selection/apply_to.rs
benjamn: Improved this in https://github.com/apollographql/router/pull/6003/commits/37ab98d8159b9732223f88691d1703a1403bf762 and https://github.com/apollographql/router/pull/6003/commits/b04f54241bf4760b92a3c197d15b580f16ce149f, thanks!

apollo-federation/src/sources/connect/json_selection/apply_to.rs
benjamn: Yes that works: https://github.com/apollographql/router/commit/b04f54241bf4760b92a3c197d15b580f16ce149f

apollo-federation/src/sources/connect/json_selection/helpers.rs
benjamn: It doesn't seem like there's already a `Vec` method to push an item and return the same (owned) vector, so this helper function still exists but has now become very short:
```rust
pub fn vec_push<T>(mut vec: Vec<T>, item: T) -> Vec<T> {
    vec.push(item);
    vec
}
```

apollo-federation/src/sources/connect/json_selection/apply_to.rs
```diff
@@ -148,27 +155,61 @@ impl ApplyToError {
...
+        self,
+        f: impl FnOnce(&JSON) -> (Option<JSON>, Vec<ApplyToError>),
+    ) -> (Option<JSON>, Vec<ApplyToError>);
+}
+
+impl ApplyToResultMethods for (Option<JSON>, Vec<ApplyToError>) {
+    // Intentionally taking ownership of self to avoid cloning, since we pretty
+    // much always use this method to replace the previous (value, errors) tuple
+    // before returning.
+    fn prepend_errors(self, mut errors: Vec<ApplyToError>) -> Self {
```
benjamn, line 174: It's interesting that this implementation of `ApplyToResultMethods::prepend_errors` can take the owned `errors` vector as `mut`, even though the trait declaration doesn't say anything about that.

nicholascioli: This is because the method takes ownership of the vector, so it's only mutable in the context of the function!

PR #6007, opened 2024-09-16
defer integration tests

Geal:
This adds support for testing `@defer` in integration tests

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/tests/common.rs
```diff
@@ -582,7 +582,6 @@ impl IntegrationTest {
 
                 let mut request = builder.json(&query).build().unwrap();
                 telemetry.inject_context(&mut request);
-                request.headers_mut().remove(ACCEPT);
```
Geal, line 585: @BrynCooke do you remember why the integration tests were removing the `Accept` header from the request?

Geal:
interestingly, multer is unable to parse the multipart body we send with defer and subs:
```
--graphql

{"data":null,"errors":[{"message":"HTTP fetch failed from 'accounts': 404: Not Found","path":[],"extensions":{"code":"SUBREQUEST_HTTP_ERROR","service":"accounts","reason":"404: Not Found","http":{"status":404}}},{"message":"HTTP fetch failed from 'accounts': subgraph response does not contain 'content-type' header; expected content-type: application/json or content-type: application/graphql-response+json","path":[],"extensions":{"code":"SUBREQUEST_HTTP_ERROR","service":"accounts","reason":"subgraph response does not contain 'content-type' header; expected content-type: application/json or content-type: application/graphql-response+json","http":{"status":404}}}],"hasNext":true}
--graphql

{"hasNext":false}
--graphql--
```

it fails with the error `incomplete multipart stream`. Maybe some different assumptions in how we chunk the response?

Geal:
the multer failure was due to a boundary reported as a quoted string

Geal:
Supporting subscriptions in those tests would be challenging, as they are not designed to execute tasks asynchronously, so we cannot(for now) use the HTTP callback while we are already waiting for a response. We could have subgraph emulation support websockets though

bnjjj (approved):

PR #6009, opened 2024-09-17
Fix __typename inside router response

IvanGoncharov:
This PR fixes the following edge cases:

# 1. The initial response of the query containing `@defer` should be an empty object

For a query like this:
```graphql
{
  ... @defer { me { name } }
}
```

The initial response should have `data` set to an empty object, as in this graphql-js test case:
https://github.com/graphql/graphql-js/blob/2b42a70191243d0ca0e0e4f1d580d6794718fbd5/src/execution/__tests__/defer-test.ts#L371-L384

# 2. `data: null` semantic should be maintained

According to GraphQL spec, null can be propagated to `data`: https://spec.graphql.org/draft/#sel-FANTNRCAACGBrwa
But the fix:
https://github.com/apollographql/router/blob/b443ad0014bd20ce81d56161848074607371671d/apollo-router/src/spec/query.rs#L241-L257

Added in #2253, resulted `data: null` being replaced with `data: { __typename: "<some typename>" }`, for queries like this:
```graphql
{
  nonNullFieldThatErrors
  __typename
}
```

# 3. Handle cases where query with `@defer` also have `__typename` inside fragment

Fix added in #2253, here:
https://github.com/apollographql/router/blob/b443ad0014bd20ce81d56161848074607371671d/apollo-router/src/spec/query.rs#L241-L257

And here:
https://github.com/apollographql/router/blob/b443ad0014bd20ce81d56161848074607371671d/apollo-router/src/spec/query.rs#L149-L160

Only worked for cases where __typename is specified directly in the topmost selection set, but didn't work for __typenames wrapped in fragments, e.g.:
```graphql
{
  ... { __typename }
  ... defer {
    me { name }
  }
}
```

**Because of `#1`, `#2` and `#3` fix added in #2253 was removed and but fix for `#4` also fixes issues with @defer**

# 4. Remaping subgraph's root type names to supergraph names didn't work for __typename wrapped in fragments

Subgraphs could have their root types named differently than supergraphs.
Moreover, different subgraphs could have different names for root types.

QP removes __typename from the topmost selection set, so in combination with:
https://github.com/apollographql/router/blob/b443ad0014bd20ce81d56161848074607371671d/apollo-router/src/spec/query.rs#L840-L843

This issue is fixed for queries like this:
```
{
  __typename
  me { name }
}
```

But QP doesn't delete __typename wrapped in fragments like so (same true for named fragment):
```
{
  ... { __typename }
  me { name }
}
```

That means 
https://github.com/apollographql/router/blob/b443ad0014bd20ce81d56161848074607371671d/apollo-router/src/spec/query.rs#L840-L843

Can't be reached because subgraph will return __typename value for root, and it will trigger this "if" instead:
https://github.com/apollographql/router/blob/b443ad0014bd20ce81d56161848074607371671d/apollo-router/src/spec/query.rs#L814

That's why I changed the order of these if branches, so the subgraph's value of __typename for the root type is always ignored, and the supergraph's root type names are always used. Also, the exact same order of if branches are already used inside `apply_selection_set.`

However, this code is not reachable for __typenames wrapped in fragments because `apply_root_selection_set` incorrectly calls `apply_selection_set` for selection sets wrapped in fragments:
https://github.com/apollographql/router/blob/b443ad0014bd20ce81d56161848074607371671d/apollo-router/src/spec/query.rs#L873-L882
https://github.com/apollographql/router/blob/b443ad0014bd20ce81d56161848074607371671d/apollo-router/src/spec/query.rs#L909-L918

If root fields are wrapped in fragments, they are still applied on root type, e.g.:
```graphql
{
   ... { 
     a: __typename # applied on query root
       ... {
         b: __typename # still query root
         
         me {
           __typename # not a root
         }
       }
     }
   }
}
```

So, I switch those calls to be `apply_root_selection_set`.
All of the above steps in combination fixed `#4` 

# 5. Incorrect errors on inline fragments using root's interfaces

Root type can implement interfaces and it's legal to use those interfaces in queries, like so:
```graphql
interface Foo {
   foo: String
}

type Query implements {
   foo: String
}
```

with query:

```graphql
{
  ... on Foo {
    foo
  }
}
```

But before my change, this query would error here:
https://github.com/apollographql/router/blob/b443ad0014bd20ce81d56161848074607371671d/apollo-router/src/spec/query.rs#L865-L867

# 6. __typename with alias returned by subgraph is not validated

__typename is validated here:
https://github.com/apollographql/router/blob/b443ad0014bd20ce81d56161848074607371671d/apollo-router/src/spec/query.rs#L547-L561

And is used to assign `current_type` here:
https://github.com/apollographql/router/blob/b443ad0014bd20ce81d56161848074607371671d/apollo-router/src/spec/query.rs#L574-L586

But aliased __typename handled here (same code for non-aliased, but they already validated by code above):
https://github.com/apollographql/router/blob/b443ad0014bd20ce81d56161848074607371671d/apollo-router/src/spec/query.rs#L574-L586
 
This code guarantees that the aliased __typename contains the name of the object type from the supergraph schema.
But it doesn't guarantee that all __typenames (aliased and not) have the same value that is compatible with `current_type`.

As `#4` proves, we can't trust the subgraph to return the correct names, and QP doesn't guarantee that either.
So, since we already use `current_type` to track the type of the current selection set, it makes sense to use it as a source of truth for all __typenames but fallback to subgraph's __typename if `current_type` is not an object (shouldn't happen unless QP has bugs).

Fixes #**issue_number**

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [x] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/tests/integration/subgraph_response.rs
```diff
@@ -9,6 +9,35 @@ include_subgraph_errors:
...
+                "data": { "__typename": "Query" }
+            })
+        );
+        Ok(())
+    }
+
+    check_query(&router, "{ __typename }").await?;
+    check_query(&router, "{ ...{ __typename } }").await
+}
+
```
IvanGoncharov, line 40: I'm working on adding more tests here

apollo-router/tests/integration/typename.rs
```diff
@@ -71,6 +71,72 @@ async fn aliased() {
     "###);
 }
 
+/* FIXME: should be fixed in query planner, failing with:
+   > value retrieval failed: empty query plan. This behavior is unexpected and we suggest opening an issue to apollographql/router with a reproduction.
```
IvanGoncharov, line 75: Can't be fixed in scope of this PR since they fail on query planning

apollo-router/src/spec/query.rs
```diff
@@ -239,31 +229,31 @@ impl Query {
                 }
             }
             Some(Value::Null) => {
-                // Detect if root __typename is asked in the original query (the qp doesn't put root __typename in subselections)
-                // cf https://github.com/apollographql/router/issues/1677
-                let operation_kind_if_root_typename = original_operation.and_then(|op| {
-                    op.selection_set
-                        .iter()
-                        .any(|f| f.is_typename_field())
-                        .then(|| *op.kind())
```
IvanGoncharov, line 248: Issue with old code:
1. It doesn't handle __typename inside fragments
2. If subgraphql returns `{ data: null, errors: [...] }` due to null propagation, this code still injects the type name, which breaks spec

apollo-router/src/spec/query.rs
```diff
@@ -146,18 +148,6 @@ impl Query {
...
-                            let operation_kind_if_root_typename =
-                                original_operation.and_then(|op| {
-                                    op.selection_set
-                                        .iter()
-                                        .any(|f| f.is_typename_field())
-                                        .then(|| *op.kind())
-                                });
-                            if let Some(operation_kind) = operation_kind_if_root_typename {
-                                output.insert(TYPENAME, operation_kind.default_type_name().into());
-                            }
```
IvanGoncharov, line 160: We don't need it anymore since I changed code here:
https://github.com/apollographql/router/pull/6009/files?diff=split&w=1#diff-ca21c6c2d8f5eacbc8fd08524c17571552b1ec649c2911b908b03b94b2c7c99fR793

apollo-router/src/spec/query.rs
```diff
@@ -239,31 +229,31 @@ impl Query {
...
 
+                response.data = Some(Value::Null);
                 return vec![];
             }
             _ => {
                 failfast_debug!("invalid type for data in response. data: {:#?}", data);
             }
         }
 
-        response.data = Some(Value::default());
```
IvanGoncharov, line 266: I changed `Value::default` to explicit null since it easier to compare with GraphQL spec algorithms

apollo-router/src/spec/query.rs
```diff
@@ -571,19 +561,13 @@ impl Query {
...
                             .map(apollo_compiler::ast::Type::Named);
 
-                        let current_type = if parameters
-                            .schema
-                            .get_interface(field_type.inner_named_type())
-                            .is_some()
-                            || parameters
-                                .schema
-                                .get_union(field_type.inner_named_type())
-                                .is_some()
```
IvanGoncharov, line 581: I replace 2 sepate searches in a map with one search + patter matching

apollo-router/tests/integration/subgraph_response.rs
```diff
@@ -9,6 +9,35 @@ include_subgraph_errors:
   all: true
 "#;
 
+#[tokio::test(flavor = "multi_thread")]
+async fn test_different_typename_on_query_root() -> Result<(), BoxError> {
```
IvanGoncharov, line 13: Test in progress, not yet ready for review

apollo-router/src/spec/query.rs
```diff
@@ -927,6 +906,36 @@ impl Query {
         Ok(())
     }
 
+    fn has_only_typename_field(&self, selection_set: &Vec<Selection>, variables: &Object) -> bool {
```
IvanGoncharov, line 909: I need this function because of: https://github.com/apollographql/router/pull/6009/files?diff=split&w=1#r1762118849

Note in order to address point 2 from above comment I need to skip differed fields

apollo-router/src/spec/query.rs
```diff
@@ -892,30 +873,28 @@ impl Query {
...
+                                root_type_name,
+                                &selection_set,
+                                parameters,
+                                input,
+                                output,
+                                path,
+                            )?;
                         }
-
-                        self.apply_selection_set(
```
IvanGoncharov, line 909: it should be `apply_root_selection_set` here since we are still inside root selection even if we go into a fragment

apollo-router/src/spec/query.rs
```diff
@@ -861,25 +840,27 @@ impl Query {
...
-                    // top level objects will not provide a __typename field
-                    if type_condition.as_str() != root_type_name {
-                        return Err(InvalidValue);
-                    }
-
                     if include_skip.should_skip(parameters.variables) {
                         continue;
                     }
 
-                    self.apply_selection_set(
```
IvanGoncharov, line 873: We are still inside root selection, even if we go inside an inline fragment.
So it should be `apply_root_selection_set`

apollo-router/src/spec/query.rs
```diff
@@ -861,25 +840,27 @@ impl Query {
                     include_skip,
                     ..
                 } => {
-                    // top level objects will not provide a __typename field
-                    if type_condition.as_str() != root_type_name {
-                        return Err(InvalidValue);
-                    }
```
IvanGoncharov, line 867: It doesn't handle subtypes, but root types can implement interfaces

apollo-router/src/spec/query.rs
```diff
@@ -837,10 +820,6 @@ impl Query {
                         );
                         path.pop();
                         res?
-                    } else if name.as_str() == TYPENAME {
-                        if !output.contains_key(field_name_str) {
-                            output.insert(field_name.clone(), Value::String(root_type_name.into()));
-                        }
```
IvanGoncharov, line 843: Move it in front as the first option since the subgraph can have different names of root type, but the root type is always object type, so we ignore value returned from the subgraph by moving this code above.

apollo-router/src/spec/query.rs
```diff
@@ -751,11 +725,15 @@ impl Query {
                         continue;
                     }
 
-                    if let Some(fragment) = self.fragments.get(name) {
+                    if let Some(Fragment {
+                        type_condition,
+                        selection_set,
+                    }) = self.fragments.get(name)
```
IvanGoncharov, line 731: This is a nonessential change, but I like that we use all fields from the Fragment type, so I made explicit restructuring here.

apollo-router/src/services/supergraph/tests.rs
```diff
@@ -1154,13 +1154,31 @@ async fn root_typename_with_defer_and_empty_first_response() {
...
+                            name
+                        }
+                    }
+                }
+            }
+        }
+
+        fragment OnlyTypename on Query {
+          __typename
+        }
```
IvanGoncharov, line 1175: I wrapped __typename into a fragment here to test my change together with differ

apollo-router/src/spec/query.rs
```diff
@@ -640,21 +624,18 @@ impl Query {
...
 
                     if name.as_str() == TYPENAME {
-                        let input_value = input
-                            .get(field_name.as_str())
-                            .cloned()
-                            .filter(|v| v.is_string())
-                            .unwrap_or_else(|| {
-                                Value::String(ByteString::from(
-                                    current_type.inner_named_type().as_str().to_owned(),
-                                ))
```
IvanGoncharov, line 650: here we tried to use __typename values returned by subgraph
but we don't validate this value at all it can even be a name of interface if subgraph uses `@interfaceObject`

By GraphQL spec `__typename` should always use object name (interfaces and unions are forbidden), so I change code to first use `current_type` (always valid) and fallback to using subgraph's value only if `current_type` is not an object type.

apollo-router/src/spec/query.rs
```diff
@@ -640,21 +624,18 @@ impl Query {
                     }
 
                     if name.as_str() == TYPENAME {
-                        let input_value = input
-                            .get(field_name.as_str())
-                            .cloned()
```
IvanGoncharov, line 645: also as bonus, I removed this clone

apollo-router/src/spec/query.rs
IvanGoncharov: I found correct fix for it:
https://github.com/apollographql/router/blob/cad52650f68ce260e7b1e829fae2ef8959d0a8ea/apollo-router/src/query_planner/execution.rs#L315

apollo-router/src/query_planner/execution.rs
```diff
@@ -311,6 +311,8 @@ impl PlanNode {
                             let _ = primary_sender.send((value.clone(), errors.clone()));
                         } else {
                             let _ = primary_sender.send((value.clone(), errors.clone()));
+                            // primary response should be an empty object
+                            value.deep_merge(Value::Object(Default::default()));
```
IvanGoncharov, line 315: here is a test case from graphql-js:
https://github.com/graphql/graphql-js/blob/2b42a70191243d0ca0e0e4f1d580d6794718fbd5/src/execution/__tests__/defer-test.ts#L371-L384

apollo-router/tests/integration/subgraph_response.rs
```diff
@@ -9,6 +9,78 @@ include_subgraph_errors:
...
+        .await;
+
+    router.start().await;
+    router.assert_started().await;
+
+    let query = "{ __typename topProducts { name } }";
+    let (_trace_id, response) = router.execute_query(&json!({ "query":  query })).await;
+    assert_eq!(response.status(), 200);
+    assert_eq!(
+        serde_json::from_str::<serde_json::Value>(&response.text().await?)?,
```
Geal, line 27:
```suggestion
        &response.json::<serde_json::Value>().await?,
```
Nit

apollo-router/tests/integration/subgraph_response.rs
```diff
@@ -9,6 +9,78 @@ include_subgraph_errors:
...
+        json!({ "data": null })
```
Geal, line 28: why should `{ __typename topProducts { name } }` return a null `data` if the subgraph returns null? `topProducts` is a nullable field, right?
Is it for the case where the subgraph would nullify because there was a non nullable field?

apollo-router/tests/integration/subgraph_response.rs
```diff
@@ -9,6 +9,78 @@ include_subgraph_errors:
...
+    );
+    Ok(())
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn test_subgraph_returning_different_typename_on_query_root() -> Result<(), BoxError> {
```
Geal, line 34: are we sure in this test that the `__typename` selections got all the way to the subgraph? If that's the case, we are supposed to have some field rewriting driven by the query plan fetch nodes (`input_rewrites` and `output_rewrites`). Is that test exercising that rewriting, or what is happening in query formatting?

apollo-router/src/spec/query.rs
Geal: good, that part was really fragile

apollo-router/src/spec/query.rs
Geal: this should be checked with what the output rewriting code in the QP is doing

apollo-router/src/spec/query.rs
Geal: this sounds reasonable but looks like a dangerous change. Do you have a test that would show what happens before and after that change?

apollo-router/src/query_planner/execution.rs
IvanGoncharov: You right, it should be correct.
I hesitated to do it since I'm less familiar with how QP executes and also I don't have capacity to test all possible scenarios in context of this PR.
So I choose to do it just inside this specific branch that I'm 100% sure is correct.

@Geal Do you think it safe to make this change for entire QP execution?
Or I can create issue for that I do it in separate PR?

apollo-router/src/services/supergraph/tests.rs
IvanGoncharov: Done ✅

apollo-router/tests/integration/subgraph_response.rs
IvanGoncharov: fixed

apollo-router/tests/integration/subgraph_response.rs
IvanGoncharov: You are right, it was mainly to test null propagation that goes up to `data.`
However, graphql servers can return `data: null` if execution is started by no resolvers executed:
https://spec.graphql.org/October2021/#sel-FAPHPHCAACEBr6C

For example, here is the test from graphql-js:
https://github.com/graphql/graphql-js/blob/993d7ce2ee6db2a13973b037817f3eac60607b8a/src/execution/__tests__/executor-test.ts#L971-L981

These cases are but could happen, so we should keep semantics.

apollo-router/tests/integration/subgraph_response.rs
IvanGoncharov: Happening in query formatting, that's test that answer your comment here: https://github.com/apollographql/router/pull/6009#discussion_r1764961754

By switching to `apply_root_selection_set` we always execute this code (even if parts of root selection is wrapped in fragments): 
https://github.com/apollographql/router/blob/ddeed8d51df2c9c0edf19eb8fbb5530d31152295/apollo-router/src/spec/query.rs#L783-L786

So now we 100% sure __typename on root type is valid independently from anything that QP will return.

apollo-router/src/query_planner/execution.rs
Geal: maybe in a separate PR, considering the timing

apollo-router/tests/integration/subgraph_response.rs
Geal: I guess the "right" way would be to detect from errors if we got a null because a non null field was nullified by the subgraph, but that may be too complex

apollo-router/tests/integration/subgraph_response.rs
Geal: that's good to make sure of that but probably we need a test that exercise the entire pipeline, not only the formatting

IvanGoncharov: yes, it's done in this test: https://github.com/apollographql/router/blob/e871c9331b9d1e7a66d646574f012f116f48b0c1/apollo-router/tests/integration/subgraph_response.rs#L34-L82

IvanGoncharov: it testing all possible scenario for how __typename can be nested and tests that __typename returned by subgraph are ignored.

apollo-router/src/query_planner/execution.rs
IvanGoncharov: I created a router issue for it.

apollo-router/tests/integration/typename.rs
IvanGoncharov: I will open an issue right now.
Probably open it for the router and see if it should be fixed in router or in federation.

IvanGoncharov: I open an issue

apollo-router/src/spec/query.rs
IvanGoncharov: yes it tested here: https://github.com/apollographql/router/blob/e871c9331b9d1e7a66d646574f012f116f48b0c1/apollo-router/tests/integration/subgraph_response.rs#L34-L82

apollo-router/src/spec/query.rs
IvanGoncharov: Not sure what to do here.
Realistically, I can't inspect QP in the scope of this PR.
I can revert this particular change, because it separated from other fixes.
@Geal Do you think it worth to revert it?

.changesets/fix_fix_typename.md
```diff
@@ -0,0 +1,22 @@
+### Fix variout edge cases for `__typename` field ([PR #6009](https://github.com/apollographql/router/pull/6009))
```
abernix, line 1:
```suggestion
### Fix various edge cases for `__typename` field ([PR #6009](https://github.com/apollographql/router/pull/6009))
```

apollo-router/src/query_planner/execution.rs
SimonSapin: I agree that changing the linked line of `execution.rs` is a better fix, but it doesn’t have to be in this PR

apollo-router/src/spec/query.rs
SimonSapin: The change looks correct to me: when a field is an object type in the supergraph we know what the correct `__typename` is and could even skip asking the subgraph

apollo-router/tests/integration/typename.rs
SimonSapin: I’d prefer not to have entire functions or tests commented out in the repo. Please use `#[ignore]` or `#[should_panic]` instead. Also please link to the issue in the FIXME comment

SimonSapin (approved):

SimonSapin:
`test_updated-amd_linux_test` is failing because of the new apollo-parser version with https://github.com/apollographql/apollo-rs/pull/918. https://github.com/apollographql/router/pull/6128 update the router test expectation accordingly, but either way this job is not required for merging.

goto-bus-stop (approved):
Nothing to add, nice work

PR #6010, opened 2024-09-17
Allow resolvable key fields to have no connect directive

pubmodmatt:
The validation to check that entity fields are resolvable through `@connect` directives was incorrectly marking resolvable key fields as errors. The key fields are now added to the list of visited fields to prevent this. This requires iterating down a tree of key fields if there are input types in use. That is implemented using a `Vec`-based stack instead of recursion, and it has checks to avoid cycles. Also added a unit test, and refactored some code to handle `@key` directives to avoid code duplication.

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

**Notes**

[CNN-456]: https://apollographql.atlassian.net/browse/CNN-456?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apollo-federation/src/sources/connect/validation/entity.rs
+++ b/apollo-federation/src/sources/connect/validation/entity.rs
@@ -4,14 +4,12 @@ use apollo_compiler::ast::InputValueDefinition;
 use apollo_compiler::ast::Value;
 use apollo_compiler::executable::FieldSet;
 use apollo_compiler::executable::Selection;
-use apollo_compiler::parser::Parser;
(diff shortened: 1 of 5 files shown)


lennyburdette (approved):
very nice!

PR #6012, opened 2024-09-17
Fix APQ and introspection gauges

BrynCooke:
Similar to the query planner gauges, these must be initialized after telemetry has been activated.

Fixes:
* `apollo_router_cache_size` for `APQ` and `introspection`

As a result of this change every query planner in the query planner pool will share the same cache storage for introspection queries. Otherwise many identical gauges are created, and they are not summed. It doesn't really make sense that each query planner should have a separate cache anyway.

Another issue was noticed, and commented on. APQLayer uses deduplicating cache which is fallible. This is initialized after telemetry has been activated, so if it fails it is likely that telemetry will stop working. This is not fixable without a larger refactor to bring APQ init forward, which is out of scope for this PR. We'll tackle router lifecycle as a larger holistic refactor later.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [x] Integration Tests
    - [x] Manual Tests

**Exceptions**

* docs not needed because this is a bug fix.
* Added to existing integration tests because this is telemetry.

**Notes**

### DIFF
--- a/apollo-router/src/introspection.rs
+++ b/apollo-router/src/introspection.rs
@@ -14,6 +14,13 @@ use crate::query_planner::QueryPlanResult;
 const DEFAULT_INTROSPECTION_CACHE_CAPACITY: NonZeroUsize =
     unsafe { NonZeroUsize::new_unchecked(5) };
 
+pub(crate) async fn default_cache_storage() -> CacheStorage<String, Response> {
+    // This cannot fail as redis is not used.
+    CacheStorage::new(DEFAULT_INTROSPECTION_CACHE_CAPACITY, None, "introspection")
+        .await
+        .expect("failed to create cache storage")
+}
+
 /// A cache containing our well known introspection queries.
 pub(crate) struct Introspection {
     cache: CacheStorage<String, Response>,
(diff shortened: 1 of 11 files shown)


bnjjj (approved):

apollo-router/src/introspection.rs
```diff
@@ -14,33 +14,37 @@ use crate::query_planner::QueryPlanResult;
...
     cache: CacheStorage<String, Response>,
     pub(crate) planner: Arc<Planner<QueryPlanResult>>,
 }
 
 impl Introspection {
-    pub(crate) async fn with_capacity(
+    pub(crate) async fn with_cache(
         planner: Arc<Planner<QueryPlanResult>>,
-        capacity: NonZeroUsize,
+        cache: CacheStorage<String, Response>,
```
Geal, line 33: I see why we now need to pass the cache as an argument, but I feel it points to something wrong in the `activate` call. Instead of calling it directly everywhere, could the plugins register a closure to handles activation when they are created?

BrynCooke: I agree with the sentiment of this, but none of the query planner or APQ stuff are plugins, so they don't get to participate lifecycle and these are the point that need to register a closure. We need to discuss the lifecycle hooks that we need and work towards refactoring both non-plugin and plugin code to use an agreed upon model.

As part of this we should also examine what we want in terms of hot reload functionality, as actually having any hot reload significantly complicates our codebase.

PR #6013, opened 2024-09-17
Enable `generate_query_fragments` by default

lrlna:
The router previously had `experimental_reuse_query_fragments` enabled by default when trying to optimize fragments before sending operations to subgraphs. While on occasion this algorithm can be more performant, we found that in vast majority of cases the query planner can be just as performant using `generate_query_fragments` query plan option, which also significantly reduces query size being sent to subgraphs. While the two options will produce correct responses, the queries produced internally by the query planner will differ.

This change enables `generate_query_fragments` by default, while disabling `experimental_reuse_query_fragments`. You can change this behavior with the following options:

```yaml
supergraph:
  generate_query_fragments: false
  experimental_reuse_query_fragments: true
```


docs/source/configuration/overview.mdx
```diff
@@ -1175,18 +1175,27 @@ example:
...
 ### Fragment reuse and generation
 
-By default, the router will attempt to reuse fragments from the original query while forming subgraph requests. This behavior can be disabled by setting the option to `false`:
+By default, the router will _generate_ fragments for subgraph requests,
+extracting _inline fragments only_ into fragment definitions before sending
+queries to subgraphs. In very many cases this significantly reduces the size of
+the query sent to subgraphs, but may increase the time it takes for planning.
+Note that this option and `experimental_reuse_query_fragments` are mutually
+exclusive; if both are explicitly set to `true`, `generate_query_fragments` will
+take precedence. This behaviour can be disabled by setting the option to `false`:
```
goto-bus-stop, line 1184: I wouldn't emphasise too much what exactly it does (wrt inline fragment extraction), as we may want to improve the algorithm in the future, for example also generating fragments for repeating patterns in field selections.

We should not say it increases the time it takes for planning. Generating query fragments takes a trivial amount of time compared to fragment reuse.

Perhaps something along these lines:

```suggestion
By default, the router will compress subgraph requests by generating fragment
definitions based on the shape of the subgraph operation. In many cases this
significantly reduces the size of the query sent to subgraphs.
Note that this option and `experimental_reuse_query_fragments` are mutually
exclusive; if both are explicitly set to `true`, `generate_query_fragments` will
take precedence. This behaviour can be disabled by setting the option to `false`:
```

apollo-router/src/configuration/mod.rs
```diff
@@ -690,7 +690,8 @@ pub(crate) struct Supergraph {
     pub(crate) reuse_query_fragments: Option<bool>,
 
     /// Enable QP generation of fragments for subgraph requests
-    /// Default: false
+    /// Default: true
+    #[serde(default = "default_generate_query_fragments")]
```
lrlna, line 694: this fails our own `default` tests:

Screenshot 2024-09-17 at 14 59 07

The only way for this config to have its own `Default` implementation is to be defined as an enum/struct, which feels wayyy too much for something like this. Any other ideas on setting up a default in a single place for this config?

SimonSapin: Remove `#[serde(default = "default_generate_query_fragments")]` and replace two instances of `generate_query_fragments.unwrap_or_default()` in this file with `generate_query_fragments.unwrap_or_else(default_generate_query_fragments)`

Since the struct already has `#[serde(default)]` serde will rely on the existing `impl Default for Supergraph`

lrlna: i felt that using `unwrap_or_else` hid away the default definition. but i guess since we've got a doc comment with `Default: true` it's maybe a bit more clear.

chandrikas (approved):

goto-bus-stop:
Oh there's been some new tests introduced in the mean time that are failing now 😭

apollo-router/src/plugins/expose_query_plan.rs
```diff
@@ -218,6 +218,10 @@ mod tests {
             build_mock_supergraph(serde_json::json! {{
                 "plugins": {
                     "experimental.expose_query_plan": true
+                },
+                "supergraph": {
+                    // TODO(@goto-bus-stop): need to update the mocks and remove this, #6013
```
dariuszkuc, line 223: should this be linking to a new issue (6013 is the existing PR#)?

goto-bus-stop: The work would likely be done as part of removing `reuse_query_fragments`, I linked the PR as a place where you can find context

apollo-router/tests/integration/fixtures/query_planner_redis_config_update_query_fragments.router.yaml
```diff
@@ -7,5 +7,5 @@ supergraph:
         urls:
           - redis://localhost:6379
         ttl: 10s
-  generate_query_fragments: true
+  generate_query_fragments: false
```
dariuszkuc, line 10: was this intentional switch to false? guess the test is to verify non-default?

dariuszkuc (changes requested):
Shouldn't we also flip the defaults in the query planner itself? i.e. here -> https://github.com/apollographql/router/blob/dev/apollo-federation/src/query_plan/query_planner.rs#L110

apollo-router/tests/integration/fixtures/query_planner_redis_config_update_query_fragments.router.yaml
goto-bus-stop: Yes, it's to make sure the cache key updates when the config is flipped

goto-bus-stop:
That also requires updating many tests. I've instead disabled *both* optimizations by default in the planner. We should change the default inside apollo-federation after some improvements to the generation algorithm I think.

apollo-federation/src/query_plan/query_planner.rs
```diff
@@ -103,15 +103,16 @@ pub struct QueryPlannerConfig {
...
     fn default() -> Self {
         Self {
-            reuse_query_fragments: true,
+            reuse_query_fragments: false,
             generate_query_fragments: false,
             subgraph_graphql_validation: false,
             incremental_delivery: Default::default(),
             debug: Default::default(),
-            type_conditioned_fetching: Default::default(),
+            type_conditioned_fetching: false,
```
dariuszkuc, line 115: nit: I believe `Default::default()` for boolean is false. Should we update other defaults to explicit false as well (for readability)?

dariuszkuc (approved):

apollo-federation/src/query_plan/query_planner.rs
goto-bus-stop: Indeed. I wanted to be explicit about this. `incremental_delivery` and `debug` are not booleans but structs, so i kept them as `Default::default`. We could consider manually writing out Default impls for those types.

PR #6014, opened 2024-09-17
Fix introspection bug, run off Tokio threads, enable "both" mode

SimonSapin:
* Update apollo-compiler to fix ROUTER-703 introspection incorrectly listing unused built-in scalars: https://github.com/apollographql/apollo-rs/pull/911
* Run Rust introspection off of Tokio threads, with the same reasoning as for the query planner in https://github.com/apollographql/router/pull/5998
* Enable introspection "both" mode by default. The legacy implementation is still used in the response but both are run and their results compared. The result of comparing both responses is logged at `DEBUG` level. At `TRACE` level, mismatches print the full query and the response diff. Example usage: `router --log error,apollo_router=info,apollo_router::query_planner::dual_introspection=trace`

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [x] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


lrlna (approved):
The only thing missing is a changelog. Thanks for getting this in!

PR #6019, opened 2024-09-18
Fix bugs identified by interface object tests

sachindshinde:
This PR:
- Fixes bugs identified by failing tests in `interface_object.rs`.
- Fixes a test bug in `can_use_a_key_on_an_interface_object_type_even_for_a_concrete_implementation()`, which was asserting the incorrect result for plan rewrites ([JS for comparison](https://github.com/apollographql/federation/blob/%40apollo/query-planner%402.7.8/query-planner-js/src/__tests__/buildPlan.interfaceObject.test.ts#L334)).
- Adjusts a snapshot assertion in the test `it_handles_interface_object_input_rewrites_when_cloning_dependency_graph()` to account for traversal ordering differences with JS. (The parallel node order is different, and `S3` is replaced with `S4` due to the subgraphs being equivalent/plan costs being equal.)

After this PR, all `interface_object.rs` tests are passing.

### DIFF
--- a/apollo-federation/src/query_graph/graph_path.rs
+++ b/apollo-federation/src/query_graph/graph_path.rs
@@ -2169,7 +2169,7 @@ impl OpGraphPath {
         let Some(self_edge) = self.edges[diff_pos] else {
             return Ok(false);
         };
-        let Some(other_edge) = self.edges[diff_pos] else {
+        let Some(other_edge) = other.edges[diff_pos] else {
             return Ok(false);
         };
         let other_edge_weight = other.graph.edge_weight(other_edge)?;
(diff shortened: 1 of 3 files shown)


dariuszkuc (approved):

apollo-federation/tests/query_plan/build_query_plan_tests/interface_object.rs
```diff
@@ -797,33 +789,33 @@ fn it_handles_interface_object_input_rewrites_when_cloning_dependency_graph() {
               }
             },
             Parallel {
-              Flatten(path: "i") {
-                Fetch(service: "S2") {
+              Flatten(path: "i.i2") {
+                Fetch(service: "S4") {
```
lrlna, line 793: We are okay that this snapshot changed and we are picking a different service than before? I am asking because I am fairly certain I've seen quite a few mismatches that resembled this sort of change.

apollo-federation/src/query_graph/graph_path.rs
```diff
@@ -2169,7 +2169,7 @@ impl OpGraphPath {
         let Some(self_edge) = self.edges[diff_pos] else {
             return Ok(false);
         };
-        let Some(other_edge) = self.edges[diff_pos] else {
+        let Some(other_edge) = other.edges[diff_pos] else {
```
goto-bus-stop, line 2172: 😱 💀 🙈

PR #6020, opened 2024-09-18
fix: group mapping errors

lennyburdette:
Before
Screenshot 2024-09-17 at 11 43 12 PM

After
Screenshot 2024-09-17 at 11 42 50 PM

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/plugins/connectors/plugin.rs
+++ b/apollo-router/src/plugins/connectors/plugin.rs
@@ -1,3 +1,4 @@
+use std::collections::HashMap;
 use std::sync::Arc;
 
 use apollo_federation::sources::connect::ApplyToError;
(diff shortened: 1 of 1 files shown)


apollo-router/src/plugins/connectors/plugin.rs
```diff
@@ -327,16 +328,31 @@ fn serialize_response(
...
+            acc.entry((e.message(), path))
+                .and_modify(|c| *c += 1)
+                .or_insert(1);
+            acc
+        },
+    ) {
         aggregated.push(json!({
             "message": key.0,
             "path": key.1,
-            "range": key.2,
```
nicholascioli, line 338: Is it OK to lose the range information here? I know that the studio UI doesn't present this information in the resulting table, but are there any other downstream consumers of these mapping errors?

nicholascioli (changes requested):
Forgot to actually comment on the downstream issue.

apollo-router/src/plugins/connectors/plugin.rs
lennyburdette: there are no other consumers besides the sandbox debugging panel. maybe someday we'll send down the raw errors with range information so we can present that information to users, but that would be separate from this aggregated view.

apollo-router/src/plugins/connectors/plugin.rs
```diff
@@ -327,16 +328,31 @@ fn serialize_response(
 fn aggregate_apply_to_errors(errors: &[ApplyToError]) -> Vec<serde_json_bytes::Value> {
     let mut aggregated = vec![];
 
-    for (key, group) in &errors
-        .iter()
-        .chunk_by(|e| (e.message(), e.path(), e.range()))
-    {
-        let group = group.collect_vec();
+    for (key, count) in &errors.iter().fold(
```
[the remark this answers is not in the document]
lennyburdette: yep that's better!

nicholascioli (approved):

apollo-router/src/plugins/connectors/plugin.rs
benjamn: For what it's worth, I believe including the `range` would still be reasonable here, if/when we display that information in the UI.

First, I was probably mistaken to think the `range` information would destroy the value of the deduplication (as discussed [here](https://github.com/apollographql/router/pull/5987#discussion_r1755547444)), since two errors with the same `message` whose `path`s differ only by numeric indices are likely to come from the same location in the JSONSelection string, so their `range`s would likely be the same in many cases. This is a consequence of how arrays are handled in the mapping language: typically the same AST structure handles every element/index of a given array, so the `range` of the selector stays the same.

Second, there might be some cases where a difference in `range` is useful to know about, like when you're accessing a similar `path` using two distinct `PathSelection`s:
```graphql
# Imagine .path does not exist within .some, generating two errors with
# the same message and path ["some", "path"], but different ranges:
lower: some.path->toLowerCase
upper: some.path->toUpperCase
```

PR #6025, opened 2024-09-18
Follow up fixes for authorization filtering

Geal:
Follow up for missing cases in #5952

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-736]: https://apollographql.atlassian.net/browse/ROUTER-736?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apollo-router/src/spec/query/transform.rs
+++ b/apollo-router/src/spec/query/transform.rs
@@ -251,6 +251,12 @@ pub(crate) fn operation(
         return Ok(None);
     };
 
+    for directive in def.directives.iter() {
+        for argument in directive.arguments.iter() {
+            used_variables_from_value(visitor, &argument.value);
+        }
+    }
+
     Ok(Some(ast::OperationDefinition {
         name: def.name.clone(),
         operation_type: def.operation_type,
@@ -271,6 +277,13 @@ pub(crate) fn fragment_definition(
     else {
         return Ok(None);
     };
+
+    for directive in def.directives.iter() {
+        for argument in directive.arguments.iter() {
+            used_variables_from_value(visitor, &argument.value);
+        }
+    }
+
     Ok(Some(ast::FragmentDefinition {
         name: def.name.clone(),
         type_condition: def.type_condition.clone(),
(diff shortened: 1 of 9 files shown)


apollo-router/src/spec/query/transform.rs
```diff
@@ -424,6 +417,35 @@ pub(crate) fn selection_set(
...
 }
 
+fn used_variables_from_value<V: Visitor>(
+    visitor: &mut V,
+    argument_value: &apollo_compiler::ast::Value,
+) {
+    println!(
+        "used_variables_from_value: argument value: {:?}",
+        argument_value
+    );
```
Geal, line 427:
```suggestion
```

dariuszkuc (approved):

duckki (approved):
Looks good!

The other bug I worked on was related to operation-level directives and their arguments that reference variables. That case is handled by this PR.

BTW, I got curious if query variable definition's default values could refernce another query variable. But, that turned out to be prevented by the spec. So, we don't need to visit (operation's) variable definitions here.

apollo-router/src/spec/query/transform.rs
```diff
@@ -827,5 +853,85 @@ fragment F on Query {
         let doc = ast::Document::parse(query, "query.graphql").unwrap();
         let result = document(&mut visitor, &doc).unwrap();
         insta::assert_snapshot!(TestResult { query, result });
+
+        // test removed field with variable in list argument
+        let query = r#"
+            query($a: String, $b: String) {
+                c
+                d(arg: ["a", $a, "b"]) @remove
```
lrlna, line 861: i am going to be very pedantic 😅😅😅. 

can these tests also test for a variable in list in list, and a variable in object in a list. example (and not a semantically correct one, so just to illustrate a point):
* d(arg: ["a", [$a], "b"]) @remove
* d(arg: ["a", {c: $a}, "b"]) @remove

Geal: sure thing: [ce12831](https://github.com/apollographql/router/pull/6025/commits/ce128316ff95394bb1a0017bd42b6a89990ee844)

Geal:
(rebasing the PR to retargetet it to the 1.55.0 branch)

IvanGoncharov (approved):

PR #6030, opened 2024-09-19
check that we actually check for the cache key with updated config

Geal:
In https://github.com/apollographql/router/pull/6029/files#diff-233bab14edc417866e44f00208db77c61ec8ee2d08d86bea7bbbde0a90d537a4R947 the test was not failing with this key change, because instead of testing the presence of the cache key for the updated config, it was testing the cache key for the initial config, that would still be present in cache. This makes sure that the initial and updated config keys are different

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


lrlna (approved):

Geal:
this should pass now that main was merged

PR #6031, opened 2024-09-19
Better locations for URL strings

dylan-apollo:
This fixes error locations within multiline/triple-quote URL strings, and sets up utilities so we can do the same within JSONSelection!
Screenshot 2024-09-19 at 5 36 52 PM

There's also a significant refactor to reduce the amount of things we pass between functions, since I just added one more thing ™️


apollo-federation/src/sources/connect/validation/graphql/strings.rs
```diff
@@ -0,0 +1,158 @@
...
+    raw_string: &'schema str,
+    /// Where `raw_string` _starts_ in the source text
+    offset: usize,
+}
+
+impl<'schema> GraphQLString<'schema> {
+    /// Get the raw string value of this GraphQL string literal
+    ///
+    /// Returns `None` if the value was not a string literal or if location data is messed up
+    pub fn new(value: &'schema Node<Value>, sources: &'schema SourceMap) -> Option<Self> {
```
pubmodmatt, line 29: Nitpick: idiomatically, this seems more like a `Result` than an `Option`, since the "messed up" cases would be an error. All callers also use `ok_or...` to convert to a `Result`.

apollo-federation/src/sources/connect/validation/graphql.rs
```diff
@@ -0,0 +1,25 @@
...
+use apollo_compiler::Schema;
+use line_col::LineColLookup;
+
+mod strings;
+
+pub(super) use strings::GraphQLString;
+
+pub(super) struct SchemaInfo<'schema> {
+    pub(crate) schema: &'schema Schema,
+    pub(crate) lookup: LineColLookup<'schema>,
```
pubmodmatt, line 14: I'm wary of including dependencies in public API (even `pub(crate)`) as it makes it more difficult to remove or update the dependency. Especially in this case, where there is really only a single method on the dependency type, this would be very easy to just wrap with a method on this struct rather than exposing the `LineColLookup` directly. Then pass around `SchemaInfo` instead of the `lookup` field.

apollo-federation/src/sources/connect/validation/graphql/strings.rs
```diff
@@ -0,0 +1,158 @@
...
+
+    pub fn line_col_for_subslice(
+        &self,
+        substring_location: Range<usize>,
+        lookup: &LineColLookup,
+    ) -> Option<Range<LineColumn>> {
+        let start_offset = self.offset + substring_location.start;
+        let end_offset = self.offset + substring_location.end;
+
+        let (line, column) = lookup.get(start_offset);
```
pubmodmatt, line 65: This call panics if the offset is incorrect, which is very worrisome. You need to either document this in a `# Panics` rustdoc on this method, or you could validate the offsets within this method with a fallible `Result`. But really this is making me question this dependency even more.

apollo-federation/src/sources/connect/validation/graphql/strings.rs
```diff
@@ -0,0 +1,158 @@
...
+
+    pub fn as_str(&self) -> &str {
+        self.raw_string
+    }
+
+    pub fn line_col_for_subslice(
+        &self,
+        substring_location: Range<usize>,
+        lookup: &LineColLookup,
+    ) -> Option<Range<LineColumn>> {
```
pubmodmatt, line 61: This appears to be infallible, why the `Option`?

Cargo.lock
```diff
@@ -3984,6 +3985,12 @@ dependencies = [
  "vcpkg",
 ]
 
+[[package]]
+name = "line-col"
```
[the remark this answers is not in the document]
dylan-apollo: This is the crate that the language server is using for line/col lookup, so I just followed their lead. Ultimately, we want `apollo-compiler` to provide what we need instead, so hopefully it's temporary.

apollo-federation/src/sources/connect/validation/graphql.rs
dylan-apollo: So you're suggesting a `SchemaInfo::get_line_col` which calls, internally to `lookup.get`?;

apollo-federation/src/sources/connect/validation/graphql/strings.rs
dylan-apollo: If we lookup beyond the length of the input, that's a bug in our implementation. We could make a `Code::Internal`, maybe, which indicates as much?

apollo-federation/src/sources/connect/validation/graphql.rs
pubmodmatt: yes.

apollo-federation/src/sources/connect/validation/graphql/strings.rs
pubmodmatt: Right, it's just a matter of how we handle such a bug. An internal error code would be fine. If it's really a bug that should never happen, it could arguably be a panic, and be documented with `Panics` - there are a few other cases of that around the router, but there are also places where lints are ensuring no panics, so I'm not sure what the most consistent handling would be here.

Cargo.lock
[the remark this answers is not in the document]
pubmodmatt: I'm ok with that.

pubmodmatt (approved):

benjamn:
@dylan-apollo @lennyburdette @pubmodmatt Could I get your eyes on an easy fix for a build error that was introduced by merging this PR (no blame, just a merge conflict that Git could not have detected)? #6066

PR #6032, opened 2024-09-20
HTTP snapshot server

pubmodmatt:
This version of HTTP snapshots differs from #5921 by implementing a snapshot server. This can be used similarly to a `wiremock` server. It listens on a local port, and proxies requests to an upstream HTTP API. Responses are stored in a local file, and subsequent matching requests return the stored response.

This has several advantages:

* The implementation is completely separate from the router and does not require changes to the router. The existing `override_subgraph_url` and `override_url` config can be used to point the router to the running snapshot server.
* The standalone server can be used to snapshot any HTTP requests, not just those made by Connectors. For example, it can be used with regular subgraph requests.
* This implementation can be integrated with the existing `samples_tests` integration tests, used in unit tests, or used with Apollo Sandbox.

An existing integration test and unit test are migrated to the snapshot server to demonstrate how it can be used.

The implementation is behind a `snapshot` feature flag.

Below is an example of running a standalone server and connecting to it from Sandbox:

```yaml
preview_connectors:
  subgraphs:
    posts:
      sources:
        jsonPlaceholder:
          override_url: http://localhost:5280
```

```
> cargo run --bin snapshot --features="snapshot" -- --snapshot-path /tmp/snapshots.json --url https://jsonPlaceholder.typicode.com/ --offline --port 5280
...
2024-09-20T01:54:43.377471Z  INFO apollo_router::test_harness::http_snapshot: Running in offline mode ⛔️
2024-09-20T01:54:43.380146Z  INFO apollo_router::test_harness::http_snapshot: Loaded 11 snapshots
2024-09-20T01:54:43.502781Z  INFO apollo_router::test_harness::http_snapshot: Snapshot server listening on port 5280
2024-09-20T01:55:16.314233Z  INFO apollo_router::test_harness::http_snapshot: Found existing snapshot url=https://jsonPlaceholder.typicode.com/users method=GET
2024-09-20T01:55:16.323124Z  INFO apollo_router::test_harness::http_snapshot: Found existing snapshot url=https://jsonPlaceholder.typicode.com/users/1/posts method=GET
2024-09-20T01:55:16.323451Z  INFO apollo_router::test_harness::http_snapshot: Found existing snapshot url=https://jsonPlaceholder.typicode.com/users/2/posts method=GET
2024-09-20T01:55:16.323723Z  INFO apollo_router::test_harness::http_snapshot: Found existing snapshot url=https://jsonPlaceholder.typicode.com/users/8/posts method=GET
2024-09-20T01:55:16.323957Z  INFO apollo_router::test_harness::http_snapshot: Found existing snapshot url=https://jsonPlaceholder.typicode.com/users/4/posts method=GET
2024-09-20T01:55:16.324189Z  INFO apollo_router::test_harness::http_snapshot: Found existing snapshot url=https://jsonPlaceholder.typicode.com/users/3/posts method=GET
2024-09-20T01:55:16.324437Z  INFO apollo_router::test_harness::http_snapshot: Found existing snapshot url=https://jsonPlaceholder.typicode.com/users/5/posts method=GET
2024-09-20T01:55:16.324641Z  INFO apollo_router::test_harness::http_snapshot: Found existing snapshot url=https://jsonPlaceholder.typicode.com/users/10/posts method=GET
2024-09-20T01:55:16.324788Z  INFO apollo_router::test_harness::http_snapshot: Found existing snapshot url=https://jsonPlaceholder.typicode.com/users/6/posts method=GET
2024-09-20T01:55:16.324930Z  INFO apollo_router::test_harness::http_snapshot: Found existing snapshot url=https://jsonPlaceholder.typicode.com/users/7/posts method=GET
2024-09-20T01:55:16.325033Z  INFO apollo_router::test_harness::http_snapshot: Found existing snapshot url=https://jsonPlaceholder.typicode.com/users/9/posts method=GET
```
 

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

**Notes**


apollo-router/src/test_harness/http_snapshot.rs
```diff
@@ -0,0 +1,520 @@
...
+    }
+
+    async fn inner_start<P: AsRef<Path>>(
+        snapshot_path: P,
+        base_url: Uri,
+        spawn: bool,
+        offline: bool,
+        update: bool,
+        include_headers: Option<Vec<String>>,
+        listener: Option<TcpListener>,
```
nicholascioli, line 351: Might be useful to encode `spawn` and `listener` into an enum, since they (seem) to always need to be the two following values:

- `spawn: true` and `listener: None`
- `spawn: false` and `listener: Some(...)`

apollo-router/src/test_harness/http_snapshot.rs
```diff
@@ -0,0 +1,520 @@
...
+        )
+    });
+
+    std::fs::write(path, serde_json::to_string_pretty(&snapshots)?).map_err(Into::into)
+}
+
+fn load<P: AsRef<Path>>(path: P) -> Result<Vec<Snapshot>, SnapshotError> {
+    let str = std::fs::read_to_string(path)?;
+    let result: Result<Vec<Snapshot>, SnapshotError> =
+        serde_json::from_str(&str).map_err(Into::into);
```
nicholascioli, line 287: I think that you can omit the `Result::map_err` here since you mark the `SnapshotError` with [`thiserror::from`](https://docs.rs/thiserror/latest/thiserror/#details).

```suggestion
        serde_json::from_str(&str)?;
```

nicholascioli: Oh, I think I see that you wanted to preserve the error vs. early returning since you optionally log later. Maybe have the early return with `?` and then you don't need the conditional afterwards? Just end with `Ok(result)`

apollo-router/src/test_harness/http_snapshot.rs
```diff
@@ -0,0 +1,520 @@
...
+        listener: Option<TcpListener>,
+    ) -> Self {
+        if update {
+            info!("Running in update mode ⬆️");
+        } else if offline {
+            info!("Running in offline mode ⛔️");
+        }
+
+        let snapshot_file = snapshot_path.as_ref();
+        let snapshots: Vec<Snapshot> = load(snapshot_file).unwrap_or(vec![]);
```
nicholascioli, line 360: Small nit: it might be nice to comment here why you throw away the error from `load`, since this is the only place it is used.

apollo-router/src/test_harness/http_snapshot.rs
```diff
@@ -0,0 +1,520 @@
...
+            rustls::ClientConfig::builder()
+                .with_safe_defaults()
+                .with_native_roots()
+                .with_no_client_auth(),
+            Client::builder().build(),
+        )
+        .expect("can create a HttpService");
+        let app = Router::new()
+            .route("/", any(root_handler))
+            .route("/*path", any(handler))
```
nicholascioli, line 373: After reading axum's docs on [wildcard paths](https://docs.rs/axum/latest/axum/routing/struct.Router.html#wildcards), I see why you needed separate handlers (`root_handler` and `handler`) that defer to the actual handler (`handle`).

It might be nice to add a comment here as to the limitations of axum's path matching for reference.

apollo-router/src/test_harness/http_snapshot.rs
```diff
@@ -0,0 +1,520 @@
...
+    let router_body: RouterBody = RouterBody::from(hyper_body.clone());
+    let request_json_body = serde_json::from_slice(&hyper_body).unwrap_or(Value::Null);
+
+    {
+        // check if we have an existing snapshot for this request
+        let mut snapshots = state.snapshots.lock().unwrap();
+        let existing = snapshots.iter().enumerate().find(|(_, snapshot)| {
+            snapshot.request.path == Some(path.clone())
+                && snapshot.request.method == Some(method.to_string())
+                && snapshot.request.body.clone() == request_json_body
```
nicholascioli, line 138: Cloning the path and method in this loop means that performance might tank if there are sufficiently large snapshots, right? Could this be updated to check against references?

apollo-router/src/test_harness/http_snapshot.rs
nicholascioli, line 1: Could this be changed to load snapshots into a `HashMap`? It seems like `O(1)` access / update would be preferable here, especially in cases where there are multiple snapshots stored in one file.

apollo-router/src/test_harness/http_snapshot.rs
nicholascioli: Switching to a `HashMap` instead of a flat list might also help.

apollo-router/src/test_harness/http_snapshot.rs
```diff
@@ -0,0 +1,520 @@
...
+        if let Some((i, snapshot)) = existing {
+            if state.update {
+                snapshots.remove(i);
+            } else {
+                info!(
+                    url = %uri,
+                    method = %method,
+                    "Found existing snapshot"
+                );
+                return Ok(snapshot.clone().try_into().unwrap());
```
nicholascioli, line 149: Since snapshots can be edited by hand is it OK to unwrap here?

apollo-router/src/test_harness/http_snapshot.rs
```diff
@@ -0,0 +1,520 @@
...
+}
+
+impl TryFrom<Snapshot> for http::Response<RouterBody> {
+    type Error = ();
+
+    fn try_from(snapshot: Snapshot) -> Result<Self, Self::Error> {
+        let mut response = http::Response::builder().status(snapshot.response.status);
+        if let Some(headers) = response.headers_mut() {
+            for (name, values) in snapshot.response.headers.into_iter() {
+                if let Ok(name) = HeaderName::from_str(&name.clone()) {
```
nicholascioli, line 428: Is it OK to throw away unsupported header names? I would assume probably, but maybe have it warn / error / something if configured incorrectly since it would probably cause confusion when testing if things are just missing from the snapshots.

nicholascioli (changes requested):
Just a bunch of small comments here and there about results and needed comments! Otherwise, looks great!

apollo-router/src/test_harness/http_snapshot.rs
pubmodmatt: Added an error here.

apollo-router/src/test_harness/http_snapshot.rs
pubmodmatt: I moved to using a HashMap and use existing data as much as possible when generating the key. It's *way* faster now.

apollo-router/src/test_harness/http_snapshot.rs
pubmodmatt: They're unrelated concerns, it's just used that way today.

apollo-router/src/test_harness/http_snapshot.rs
```diff
@@ -0,0 +1,563 @@
...
+    snapshots: &mut HashMap<String, Snapshot>,
+) -> Result<(), SnapshotError> {
+    let path = path.as_ref();
+    if let Some(parent) = path.parent() {
+        std::fs::create_dir_all(parent)?;
+    }
+
+    // Keep the snapshots sorted to make source control merges easier when updating snapshots
+    let mut snapshots = snapshots.values().cloned().collect::<Vec<_>>();
+    snapshots.sort_by_key(Snapshot::key);
```
nicholascioli, line 281: If we need to always have the entries sorted by key, then maybe a [`BTreeMap`](https://doc.rust-lang.org/std/collections/struct.BTreeMap.html) would work better here since it will always keep its elements sorted by key.

apollo-router/src/test_harness/http_snapshot.rs
```diff
@@ -0,0 +1,563 @@
...
+            }
+        } else {
+            let message = "Unable to read response body".to_string();
+            error_message = message.clone();
+            error!(
+                url = %uri,
+                method = %method,
+                message
+            )
+        }
```
nicholascioli, line 244: Small nit: this is a bit gnarly. It might be nice to break this up into fallible sections that use the `?` to return early, but up to you.

nicholascioli (approved):
Looks great! Left a small nit about using a `BTreeMap` and about legibility, but feel free to ignore!

apollo-router/src/test_harness/http_snapshot.rs
pubmodmatt: Excellent point, no idea why this did not occur to me. I'll update this.

apollo-router/src/test_harness/http_snapshot.rs
pubmodmatt: More than fair. I'll take a look at cleaning this up.

PR #6036, opened 2024-09-21
WIP: arrow method validation

pubmodmatt:
Validation changes:
* Implement a JSON Selection visitor that can visit each part of a selection, including arrow methods
* Adds validation of arrow method names against the known list of public arrow methods
* Adds validation of arrow method arity
* Add a unit test that valid uses of arrow methods do not cause any validation errors
* Add a unit test with use of a non-existent arrow method name
* Add a unit test with use of an arrow method with incorrect arity

method

JSON Selection changes:
* Makes some more of the JSON selection code `pub(crate)` so the validation code is able to visit all the places where an arrow method name can appear

Known Issues:
* There is currently an issue with another validation that reports on fields that do not have an associated connector. This keeps track of an `already_seen` list as the selection is validated. However, if any other validation fails and returns early, this validation will not see any remaining fields and incorrectly reports errors. This will need to be addressed separately, but the result can be seen in the test snapshots for this PR.
* The validation code requires significant knowledge about arrow methods, such as their names and arity (and in the future, type information and how it applies to both the HTTP response and GraphQL schema). This is currently just hard coded. There will need to be some mechanism to provide this information, ideally one that is extensible as new arrow methods are added.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[CNN-415]: https://apollographql.atlassian.net/browse/CNN-415?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apollo-federation/src/sources/connect/json_selection/parser.rs
+++ b/apollo-federation/src/sources/connect/json_selection/parser.rs
@@ -1079,8 +1079,8 @@ pub(crate) fn parse_string_literal(input: Span) -> IResult<Span, WithRange<Strin
 }
 
 #[derive(Debug, PartialEq, Eq, Clone, Default)]
-pub(super) struct MethodArgs {
-    pub(super) args: Vec<WithRange<LitExpr>>,
+pub(crate) struct MethodArgs {
+    pub(crate) args: Vec<WithRange<LitExpr>>,
     pub(super) range: OffsetRange,
 }
 
--- a/apollo-federation/src/sources/connect/json_selection/parser.rs
+++ b/apollo-federation/src/sources/connect/json_selection/parser.rs
@@ -297,7 +297,7 @@ impl ExternalVarPaths for NamedSelection {
 
 #[derive(Debug, PartialEq, Eq, Clone)]
 pub struct PathSelection {
-    pub(super) path: WithRange<PathList>,
+    pub(crate) path: WithRange<PathList>,
 }
 
 // Like NamedSelection, PathSelection is an AST structure that takes its range
@@ -378,7 +378,7 @@ impl From<PathList> for PathSelection {
 }
 
 #[derive(Debug, PartialEq, Eq, Clone)]
-pub(super) enum PathList {
+pub(crate) enum PathList {
     // A VarPath must start with a variable (either $identifier, $, or @),
     // followed by any number of PathStep items (the WithRange<PathList>).
     // Because we represent the @ quasi-variable using PathList::Var, this
--- a/apollo-federation/src/sources/connect/validation/selection/visitor.rs
+++ b/apollo-federation/src/sources/connect/validation/selection/visitor.rs
@@ -59,19 +59,6 @@ enum SelectionGroup<'schema> {
         parent: SelectionPart<'schema>,
         children: Vec<SelectionPart<'schema>>,
     },
-    Empty {
-        parent: SelectionPart<'schema>,
-    },
-}
-
-impl<'schema> SelectionGroup<'schema> {
-    fn children(&self) -> Vec<SelectionPart<'schema>> {
-        match self {
-            SelectionGroup::Root { children } => children.clone(),
-            SelectionGroup::Child { children, .. } => children.clone(),
-            SelectionGroup::Empty { .. } => vec![],
-        }
-    }
 }
 
 impl<'schema> SelectionGroup<'schema> {
@@ -80,7 +67,17 @@ impl<'schema> SelectionGroup<'schema> {
     }
 
     fn empty(parent: SelectionPart<'schema>) -> Self {
-        SelectionGroup::Empty { parent }
+        SelectionGroup::Child {
+            parent,
+            children: vec![],
+        }
+    }
+
+    fn children(&self) -> Vec<SelectionPart<'schema>> {
+        match self {
+            SelectionGroup::Root { children } => children.clone(),
+            SelectionGroup::Child { children, .. } => children.clone(),
+        }
     }
 }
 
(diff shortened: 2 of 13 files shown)


apollo-federation/src/sources/connect/validation/selection/visitor.rs
```diff
@@ -0,0 +1,288 @@
...
+        children: Vec<SelectionPart<'schema>>,
+    },
+    Child {
+        parent: SelectionPart<'schema>,
+        children: Vec<SelectionPart<'schema>>,
+    },
+    Empty {
+        parent: SelectionPart<'schema>,
+    },
+}
```
benjamn, line 65: Could this be simplified to avoid the `Root`/`Child` distinction by having a single `parent: Option<SelectionPart<'schema>>` that would be `None` in the `Root` case?

It also seems like `Empty` is just the case where `children` is empty. If that's true, I don't think this type benefits much from being an enum with three different cases, when it could be a single struct.

apollo-federation/src/sources/connect/json_selection/parser.rs
```diff
@@ -263,7 +263,7 @@ impl ExternalVarPaths for NamedSelection {
 
 #[derive(Debug, PartialEq, Eq, Clone)]
 pub struct PathSelection {
-    pub(super) path: WithRange<PathList>,
+    pub(crate) path: WithRange<PathList>,
```
benjamn, line 266: As far as I know, the only way to keep the `PathList` private in Rust is to avoid opening up its visibility like this. Please reconsider this change.

apollo-federation/src/sources/connect/json_selection/parser.rs
```diff
@@ -1020,7 +1020,7 @@ pub fn parse_string_literal(input: Span) -> IResult<Span, WithRange<String>> {
 
 #[derive(Debug, PartialEq, Eq, Clone, Default)]
 pub struct MethodArgs {
-    pub(super) args: Vec<WithRange<LitExpr>>,
+    pub(crate) args: Vec<WithRange<LitExpr>>,
```
benjamn, line 1023: I don't see an immediate problem here (compared to `PathList`), but I would note that the `WithRange` representation was relatively recently added and probably should not be considered stable yet.

benjamn (changes requested):
I know this PR is still a draft, but I wanted to give some feedback early enough for you to handle it easily.

Broadly speaking, I'm finding it's difficult to defend abstraction boundaries in a Rust codebase, to prevent coupling and allow for implementation details to evolve in (relative) isolation. Tools like keeping `PathList` as `pub(super)` are sometimes all we have, so please don't change visibility without discussion.

apollo-federation/src/sources/connect/validation/selection/visitor.rs
```diff
@@ -0,0 +1,288 @@
...
+/// A part of a JSON Selection to be visited
+#[derive(Clone, Debug)]
+pub(super) enum SelectionPart<'schema> {
+    JSONSelection(&'schema JSONSelection),
+    LitExpr(&'schema LitExpr),
+    MethodArgs(&'schema MethodArgs),
+    NamedSelection(&'schema NamedSelection),
+    PathList(&'schema PathList),
+    SubSelection(&'schema SubSelection),
+}
```
benjamn, line 35: What qualifies a given AST structure as a part of the selection that should be visited, since this list is not exhaustive? I'd suggest adding more detail to the comment.

pubmodmatt: Just that I needed it to locate arrow methods. This is really just the beginning of a proper AST visitor to allow validating arrow methods. I think @nicholascioli is working on designing something more extensive.

pubmodmatt:
> I know this PR is still a draft, but I wanted to give some feedback early enough for you to handle it easily.
> 
> Broadly speaking, I'm finding it's difficult to defend abstraction boundaries in a Rust codebase, to prevent coupling and allow for implementation details to evolve in (relative) isolation. Tools like keeping `PathList` as `pub(super)` are sometimes all we have, so please don't change visibility without discussion.

@benjamn - this PR is specifically created to have that discussion. I was waiting to ask for your input until I have a few more changes in, but thanks for the early feedback! I'll add individual responses on the comments, and we can also meet to discuss. Implementing JSON Selection validation definitely brings up some important topics.

apollo-federation/src/sources/connect/validation/selection/visitor.rs
pubmodmatt: The `Empty` one was just vestigial and related to something I abandoned - I'll remove that. The `Root` vs `Child` is a way to use Rust type system to ensure correctness of data. In other languages, you might have a parent that is `Optional` or can be `null`. This allows creation of instances that are incorrect. For example, you could have a node return a child that has its parent set to `Empty` or `null`, which would be invalid. Every bit of code that accesses the parent field then needs to check for this and somehow handle it to avoid hitting a `NullPointerExdeption` or the like - but there isn't really anything sensible they can do if they encounter it. In Rust, we can avoid this situation entirely by easily creating type variants that enforce correctness of data. A `Root` is the only node that has no parent, and everything else requires a parent.

apollo-federation/src/sources/connect/validation/selection/visitor.rs
```diff
@@ -0,0 +1,288 @@
...
+            SelectionPart::JSONSelection(json_selection) => match json_selection {
+                JSONSelection::Named(sub_selection) => Some(SelectionGroup::new(
+                    field,
+                    vec![SelectionPart::SubSelection(sub_selection)],
+                )),
+                JSONSelection::Path(path_selection) => Some(SelectionGroup::new(
+                    field,
+                    vec![SelectionPart::PathList(&path_selection.path)],
+                )),
+            },
```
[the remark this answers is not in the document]
pubmodmatt: The plan was to move this visitor into the JSON Selection module once we align on what this visitor looks like. I just put it here temporarily. I also think @nicholascioli also has some work going on in this area that might supersede this.

apollo-federation/src/sources/connect/json_selection/parser.rs
pubmodmatt: I'm fine with it being unstable and the validation code needing to change. There is significant value here for customers if we can get something in relatively soon - being able to point customers to a specific range associated with a validation error (such as an arrow method) is a way better experience that just highlighting the entire selection.

pubmodmatt:
There's a lot of work going on with JSON Selection validation that replaces this.

PR #6038, opened 2024-09-23
Finalize defer support in Rust QP

TylerBloom:
Implemented support for `@defer` in the QP.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [x] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-federation/src/operation/mod.rs
+++ b/apollo-federation/src/operation/mod.rs
@@ -38,6 +38,7 @@ use crate::error::FederationError;
 use crate::error::SingleFederationError;
 use crate::error::SingleFederationError::Internal;
 use crate::link::graphql_definition::BooleanOrVariable;
+use crate::link::graphql_definition::DeferDirectiveArguments;
 use crate::query_graph::graph_path::OpPathElement;
 use crate::query_plan::conditions::Conditions;
 use crate::query_plan::FetchDataKeyRenamer;
@@ -207,6 +208,70 @@ pub(crate) struct NormalizedDefer {
     pub defer_conditions: IndexMap<String, IndexSet<String>>,
 }
 
+struct DeferNormalizer {
+    used_labels: HashSet<String>,
+    problems: HashSet<SelectionKey>,
+    label_offset: usize,
+}
+
+impl DeferNormalizer {
+    fn new(selection_set: &SelectionSet) -> Self {
+        fn collect_labels(normalizer: &mut DeferNormalizer, selection: &Selection) {
+            // TODO: Does this need to be checking FragmentSpread + InlineFragment or just
+            // spreads? The JS code checks "FragmentSelection"
+            if let Selection::InlineFragment(inline) = selection {
+                if let Some(args) = inline
+                    .inline_fragment
+                    .data()
+                    .defer_directive_arguments()
+                    // TODO: Ideally, a version of this method exists that doesn't return a
+                    // Result...
+                    .unwrap()
+                {
+                    let DeferDirectiveArguments { label, if_ } = args;
+                    if let Some(label) = label {
+                        normalizer.used_labels.insert(label);
+                    } else {
+                        normalizer.problems.insert(inline.key());
+                    }
+                    if matches!(if_, Some(BooleanOrVariable::Boolean(_))) {
+                        normalizer.problems.insert(inline.key());
+                    }
+                }
+            }
+            selection
+                .selection_set()
+                .into_iter()
+                .flatten()
+                .for_each(|(_, op)| collect_labels(normalizer, op))
+        }
+        let mut digest = Self {
+            used_labels: HashSet::default(),
+            problems: HashSet::default(),
+            label_offset: 0,
+        };
+        selection_set
+            .iter()
+            .for_each(|sel| collect_labels(&mut digest, sel));
+        digest
+    }
+
+    fn is_problem(&self, key: &SelectionKey) -> bool {
+        self.problems.contains(key)
+    }
+
+    fn get_label(&mut self) -> String {
+        loop {
+            let digest = format!("qp__{}", self.label_offset);
+            if self.used_labels.contains(&digest) {
+                self.label_offset += 1;
+            } else {
+                return digest;
+            }
+        }
+    }
+}
+
 impl Operation {
     /// Parse an operation from a source string.
     #[cfg(any(test, doc))]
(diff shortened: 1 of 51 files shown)


apollo-federation/src/operation/mod.rs
```diff
@@ -406,6 +505,40 @@ mod selection_map {
...
+        /// Iterates over the inner map, passing each selection through the mapper. If the mapper
+        /// returns `Some`, the current selection is removed from the map and returned selection is
+        /// inserted in its place. This method is intented from sparce updates to selections which
+        /// update their keys (such as defer normalization). As such, update selections will hold
+        /// the same position as the former selection.
+        ///
+        /// NOTE: The given `mapper` function is given a mutable reference so that the `Selection`
+        /// can recurse if needed. Should any other change be needed, the mapper should clone the
+        /// `Selection` and return the updated copy.
+        pub(crate) fn update_and_reinsert<F>(&mut self, mut mapper: F)
```
TylerBloom, line 517: This method is likely not going to stay. While it is safely used currently, if the returned selection has a key that already exists in the map, that selection is removed. This is fine if the key of the original selection equals the new key, but this causes problems because the new selection could remove a separate selection.

apollo-federation/src/operation/mod.rs
```diff
@@ -3464,6 +3660,100 @@ impl InlineFragmentSelection {
...
+        }
+
+        if args_copy.label.is_none() {
+            args_copy.label = Some(normalizer.get_label());
+        }
+
+        if remove_defer {
+            todo!();
+            // return;
+        }
```
TylerBloom, line 3692: I still need to write a test for this case.

apollo-federation/src/query_plan/fetch_dependency_graph.rs
```diff
@@ -1686,13 +1688,21 @@ impl FetchDependencyGraph {
...
+                    "{} == {}",
+                    DisplayOption(node.defer_ref.as_ref()),
+                    DisplayOption(child.defer_ref.as_ref())
+                );
                 let Some(child_defer_ref) = &child.defer_ref else {
+                    // FIXME: We should not be unwrapping here. Does a `DisplayOption` make sense?
+                    let parent_defer_ref = node.defer_ref.as_ref().unwrap();
                     panic!("{} has defer_ref `{parent_defer_ref}`, so its child {} cannot have a top-level defer_ref.",
                            node.display(node_index),
                            child.display(child_index),
```
TylerBloom, line 1708: Instead of `unwrap`-ing here (`parent_defer_ref` can be `None`), we should use `DisplayOption` instead.

TylerBloom (commented):
Things still todo because these comments:
- [ ] Remove print statements
- [x] Cleanup generated schema
- [ ] Update snapshots that are now outdated because of changed to query plan display
- [ ] Make `normalize_defer` consume an operation and return an operation

apollo-federation/src/operation/mod.rs
TylerBloom: Good point. It is not safe 🙈

apollo-federation/src/operation/mod.rs
```diff
@@ -201,7 +205,79 @@ pub(crate) struct NormalizedDefer {
...
+            label_offset: 0,
+            assigned_labels: IndexSet::default(),
+            conditions: IndexMap::default(),
+        };
+        let mut stack = selection_set
+            .into_iter()
+            .map(|(_, sel)| sel)
+            .collect::<Vec<_>>();
+        while let Some(selection) = stack.pop() {
+            // TODO: Does this need to be checking FragmentSpread + InlineFragment or just
```
dariuszkuc, line 233: when this is initialized all fragments are already inlined/expanded so we shouldn't have any `FragmentSpread`s

apollo-federation/src/query_plan/query_planning_traversal.rs
```diff
@@ -84,8 +87,9 @@ pub(crate) struct QueryPlanningTraversal<'a, 'b> {
     /// True if query planner `@defer` support is enabled and the operation contains some `@defer`
     /// application.
     has_defers: bool,
-    /// The initial fetch ID generation (used when handling `@defer`).
-    starting_id_generation: u64,
+    /// A handle to the sole generator of fetch IDs. While planning an operation, only one of
+    /// generator can be used.
+    id_generator: Arc<FetchIdGenerator>,
```
TylerBloom, line 92: This was causing a defer test to fail because some fetch nodes has overlapping IDs. Now, there is a single "global" ID generator used for the entire planning process. While looking at some of the test, I noticed a side effect of this. Sometime, an id will be generated but never needed/used. This can cause holes in the IDs used in the first QP.

I point this out because the only invariant around fetch IDs mentioned is that they must be unique. However, I'm unsure if the router expects fetch IDs to not have holes. If this is the case, I can rework this solution.

apollo-federation/src/operation/mod.rs
```diff
@@ -200,7 +205,62 @@ pub(crate) struct NormalizedDefer {
...
+            self.label_offset += 1;
+            if !self.used_labels.contains(&digest) {
+                self.assigned_labels.insert(digest.clone());
+                return digest;
+            }
+        }
+    }
+
+    fn register_condition(&mut self, label: String, cond: Name) {
+        self.conditions.entry(cond).or_default().insert(label);
```
dariuszkuc, line 262: what if a condition already has an associated entry with different label? it is possible?

apollo-federation/src/operation/mod.rs
```diff
@@ -200,7 +205,62 @@ pub(crate) struct NormalizedDefer {
...
+            used_labels: HashSet::default(),
+            label_offset: 0,
+            assigned_labels: IndexSet::default(),
+            conditions: IndexMap::default(),
+        };
+        let mut stack = selection_set
+            .into_iter()
+            .map(|(_, sel)| sel)
+            .collect::<Vec<_>>();
+        while let Some(selection) = stack.pop() {
```
dariuszkuc, line 230: should this be recursive instead? we wouldn't need a mutating selection stack

apollo-federation/src/operation/mod.rs
TylerBloom: That is very much possible. That is way `self.conditions` is a `HashMap<String, Vec<Name>>`.

apollo-federation/src/operation/mod.rs
TylerBloom: I had the exact same thought and the original impl was recursive, but that was what the causing the stack overflows. It is entirely possible that in release mode, those function calls would be optimized away (and there would be tail recursion), but we don't run our tests in release mode.

apollo-federation/src/operation/mod.rs
```diff
@@ -3648,6 +3616,397 @@ impl NamedFragments {
...
+impl DeferFilter<'_> {
+    fn remove_defer(&self, directive_list: &mut DirectiveList, schema: &apollo_compiler::Schema) {
+        match self {
+            Self::All => {
+                directive_list.remove_one(DEFER_DIRECTIVE_NAME);
+            }
+            Self::Labels(set) => {
+                let label = directive_list
+                    .get(DEFER_DIRECTIVE_NAME)
+                    .and_then(|directive| directive.argument_by_name("label", schema).ok())
```
dariuszkuc, line 3638: nit: since we have `DEFER_DIRECTIVE_NAME` as a const -> should `label` also be a const?

apollo-federation/src/query_plan/fetch_dependency_graph.rs
```diff
@@ -3993,32 +3985,32 @@ fn compute_input_rewrites_on_key_fetch(
...
     };
 
-    let updated_defer_ref = defer_args.label.as_ref().ok_or_else(||
-        // PORT_NOTE: The original TypeScript code has an assertion here.
-        FederationError::internal(
-                    "All defers should have a label at this point",
-                ))?;
-    let updated_operation = operation.without_defer();
-    let updated_path_to_defer_parent = match updated_operation {
+    // PORT_NOTE: The original TypeScript code has an assertion here.
```
dariuszkuc, line 4007: nit: we are throwing `FederationError` below which mimics the assertion so I think this comment should be removed

goto-bus-stop (approved):

lrlna (approved):

PR #6040, opened 2024-09-23
fix(federation): fix selection key integrity in without_defer

goto-bus-stop:
`without_defer` and `reduce_defer` now produce new selection sets, to make sure that modified selections are merged properly.

In the future, this should be ported to a selection set mutation API, but we don't have one today. This is far less efficient, but it happens relatively infrequently.

Added a few unit tests to check that defer is actually removed from nested selection sets and fragment definitions.


apollo-federation/src/operation/mod.rs
```diff
@@ -3922,6 +3648,400 @@ impl NamedFragments {
...
+    /// operation will:
+    ///  - have a (unique) label. Which implies that this method generates a label for any `@defer`
+    ///    not having a label.
+    ///  - have a non-trivial `if` condition, if any. By non-trivial, we mean that the condition
+    ///    will be a variable and not an hard-coded `true` or `false`. To do this, this method will
+    ///    remove the condition of any `@defer` that has `if: true`, and will completely remove any
+    ///    `@defer` application that has `if: false`.
+    pub(crate) fn with_normalized_defer(mut self) -> Result<NormalizedDefer, FederationError> {
+        if self.has_defer() {
+            // TODO(@goto-bus-stop): does this need named fragment handling?
```
goto-bus-stop, line 4023: Should solve this and add tests before merge

goto-bus-stop:
Actually, since we run without_defer after expansion, we may not need to handle named fragment definitions. We would need to remove defer in `.reuse_fragments()` (or just ignore fragments that contain it) but not in `.generate_fragments()`.

I'm going to keep the `@defer` in named fragment handling for now to support `.reuse_fragments()`. It's also the more safe/conservative thing to do given that we do use the `NamedFragments` structure. In the future, if we are fully committed to `.generate_fragments()`, we can remove *all* named fragment tracking during planning. (We can also maintain fragments throughout planning, but that's not the direction we are currently leaning, I guess)

apollo-federation/src/operation/mod.rs
```diff
@@ -3922,6 +3648,400 @@ impl NamedFragments {
...
+        filter: DeferFilter<'_>,
+        named_fragments: &NamedFragments,
+    ) -> Result<Self, FederationError> {
+        let mut without_defer =
+            SelectionSet::empty(self.schema.clone(), self.type_position.clone());
+        for selection in self.selections.values() {
+            without_defer
+                .add_local_selection(&selection.without_defer(filter, named_fragments)?, true)?;
+        }
+        debug_assert!(!without_defer.has_defer());
```
TylerBloom, line 3945: This is only true if `filter` is `All`.

TylerBloom (approved):

apollo-federation/src/operation/mod.rs
goto-bus-stop: A! good spot, this check should only happen in `Operation::without_defer`

apollo-federation/src/operation/mod.rs
goto-bus-stop: Okay, I'll explicitly not handle this, as it requires quite a bit of work just for potentially helping `.reuse_fragments()`. It should be fine since we don't track which named fragment an expanded fragment comes from

PR #6043, opened 2024-09-24
Fix bugs identified by `@skip`/`@include` tests

sachindshinde:
This PR:
- Reverts most of #5919:
  - The changes to selection set merge logic have been reversed, as it turns out the performance impact can be noticed in some rare cases.
  - The changes to `Selection::from_element()` have been reversed, as it deviated from JS QP. The bugs causing the failure of test `it_handles_a_simple_at_requires_triggered_within_a_conditional()` have been identified, and fixed in this PR.
    - The followup PR #5963 with fixes to `Selection::from_element()`-related changes has also been reverted.
  - The only changes kept are the ones to `remove_unneeded_top_level_fragment_directives()` (and the added function `with_updated_directives_and_selection_set()`), as those fixed deviations from JS QP.
  - The tests from #5963 are also kept (they still pass).
- Fixes a bug in `OpGraphPathContext::with_context_of()` where the conditionals were pushed in the wrong order (this can result in `@skip`/`@include`s being applied in inverse order, causing spurious differences in generated QPs).
  - This caused a change in the snapshot for test `handles_multiple_conditions_on_abstract_types()`.
- Fixes bugs in `SelectionSet::without_unnecessary_fragments()` where:
  - It did not recurse into the selections of removed inline fragments.
  - It checked whether the parent type and fragment's type conditions were strictly equal and forgot about subtyping.
- Fixes bugs in  `SelectionSet::is_useless_followup_element()` where:
  - It flipped a boolean for checking whether directives were useless.
  - It flipped the order of arguments passed to `Schema::is_subtype()`.

Test-wise, this PR causes the following tests to pass:
  - `it_handles_a_simple_at_requires_triggered_within_a_conditional()`
    - As mentioned above, the previous PRs had caused this test to pass but deviated from JS QP behavior to do so. This PR fixes the bugs in the ported JS QP logic that caused the test failure, and reverses the introduced deviations.
  - `it_handles_an_at_requires_where_multiple_conditional_are_involved()`


dariuszkuc (approved):

apollo-federation/tests/query_plan/build_query_plan_tests/requires/include_skip.rs
```diff
@@ -230,167 +228,11 @@ fn it_handles_an_at_requires_where_multiple_conditional_are_involved() {
...
                 },
-              }
-            },
-          }
-        "###
-    );
-}
-
-#[test]
-fn unnecessary_include_is_stripped_from_fragments() {
```
lrlna, line 243: Should these two tests be removed?

apollo-federation/tests/query_plan/build_query_plan_tests/requires/include_skip.rs
```diff
@@ -230,167 +228,11 @@ fn it_handles_an_at_requires_where_multiple_conditional_are_involved() {
...
+              },
             },
-          },
-        }
-        "###
-    );
-}
-
-#[test]
-fn selections_are_not_overwritten_after_removing_directives() {
```
lrlna, line 333: this is the second test that's removed and I am not sure why.

lrlna (changes requested):
We should probably keep the two tests that were removed in `include_skip.rs`.

goto-bus-stop (approved):
Looks great. I also wonder about the tests, they are part of the reverted commits, were they testing something incorrectly, useless, or just caught in the crossfire?

apollo-federation/tests/query_plan/build_query_plan_tests/requires/include_skip.rs
sachindshinde: Good catch, added the tests back and they still pass, it should be fine to keep them.

sachindshinde:
@goto-bus-stop 
The tests being dropped were caught in crossfire, they still pass and should be fine to keep.

lrlna (approved):

PR #6045, opened 2024-09-24
router next documentation

garypen:
Router next documentation.

[ROUTER-776]: https://apollographql.atlassian.net/browse/ROUTER-776?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docs/source/executing-operations/native-query-planner.mdx
```diff
@@ -0,0 +1,55 @@
...
+
+- `new`- enables only the new Rust-native query planner
+- `legacy` - enables only the legacy JavaScript query planner 
+- `both_best_effort` - enables both new and legacy query planners for comparison. The legacy query planner is used for execution. 
+
+<MinVersion version="2.0.0">
+
+## Optimize native query planner
+
+</MinVersion>
```
shorgi, line 33: @lrlna is this correct that 2.0.0 is the minimum router version that supports all-native query planning?

docs/source/executing-operations/native-query-planner.mdx
```diff
@@ -0,0 +1,55 @@
...
+- `generate_query_fragments: true` enables faster throughput by reducing the size of queries sent to subgraphs.
+- `experimental_introspection_mode: new` enables Rust-native schema introspection.
+
+## Metrics for native query planner
+
+TBD
+
+## Limitations of native query planner
+
+TBD
```
shorgi, line 55: @lrlna is there any telemetry, limitations, are other topics that need to be called out on this page?

docs/source/release-notes.mdx
```diff
@@ -0,0 +1,15 @@
+---
+title: GraphOS Router Release Notes
+---
+
+## GraphOS Router v2.0 (Alpha)
```
shorgi, line 5: Will remove this for this PR, if the upgrading-from-version-1.x guide is approved to be sufficient for this release. Will be adding for subsequent releases

docs/source/upgrading-from-version-1.x.mdx
```diff
@@ -0,0 +1,69 @@
...
+
+The router supports reporting operation usage metrics to GraphOS via OpenTelemetry Protocol (OTLP).
+
+Prior to version 1.49.0 of the router, all GraphOS reporting was performed using a [private tracing format](graphos/metrics/sending-operation-metrics#reporting-format). In v1.49.0, we introduced support for using OTEL to perform this reporting. In v1.x, this is controlled using the `experimental_otlp_tracing_sampler` flag, and it's disabled by default. 
+
+Now in v2.x, this flag is renamed to `otlp_tracing_sampler`, and it's enabled by default.
+
+Learn more about configuring [usage reporting via OTLP](/router/configuration/telemetry/apollo-telemetry#usage-reporting-via-opentelemetry-protocol-otlp).
+
+#### Metrics reporting defaults
```
garypen, line 27: I think this should be `###` not `####`.

garypen (commented):
Looks good from my perspective. I think I saw one formatting issue which I commented on.

docs/source/executing-operations/native-query-planner.mdx
chandrikas: No, will be available from 1.56.0

docs/source/upgrading-from-version-1.x.mdx
```diff
@@ -0,0 +1,69 @@
...
+
+Learn how to upgrade from your GraphOS Router deployment from version 1.x to 2.x.
+
+<Note>
+
+The latest version of GraphOS Router v2.x is **v2.0.0-preview.0**.
+
+</Note>
+
+## New features in GraphOS Router v2.x
```
chandrikas, line 15:
```suggestion
## New features in GraphOS Router v2.0.0-preview.x
```

docs/source/configuration/overview.mdx
```diff
@@ -579,6 +579,31 @@ You can configure certain caching behaviors for generated query plans and APQ (b
...
+
+### Native query planner
+
+</MinVersion>
+
+Starting with v1.49.0, the router can run a Rust-native query planner. This native query planner can be run by itself to plan all queries, replacing the legacy Javascript implementation.
+
+<Tip>
+
+Starting with v2.0.0, to run the most performant and resource-efficient native query planner, set the following options in your `router.yaml`:
```
lrlna, line 592: So the configuration below not only enables native query planning, but also does not start up the v8, the javascript runtime. We should note that in this paragraph. But do we need to mention the javascript runtime a bit earlier in this section? will it be too confusing it's brought up here? or should the native query planner section just take care of that?
```suggestion
Starting with v1.56.0, to run the most performant and resource-efficient native query planner and to disable the JavaScript runtime in the router, set the following options in your `router.yaml`:
```

docs/source/executing-operations/native-query-planner.mdx
```diff
@@ -0,0 +1,55 @@
...
+minVersion: 1.49.0
+---
+
+<PreviewFeature />
+
+Learn to run the GraphOS Router with the Rust-native query planner and improve your query planning performance and scalability.
+
+## Background about query planner implementations
+
+In v1.49.0 the router introduced a [query planner](/graphos/routing/about-router#query-planning) implemented natively in Rust. This native query planner improves the overall performance and resource utilization of query planning. It exists alongside the legacy Javascript implementation. and it will eventually replace the legacy implementation.
```
lrlna, line 13:
```suggestion
In v1.49.0 the router introduced a [query planner](/graphos/routing/about-router#query-planning) implemented natively in Rust. This native query planner improves the overall performance and resource utilization of query planning. It exists alongside the legacy Javascript implementation, and it will eventually replace the legacy implementation.
```

docs/source/executing-operations/native-query-planner.mdx
```diff
@@ -0,0 +1,55 @@
...
+
+As part of the effort to ensure correctness and stability of the new query planner, starting in v1.53.0 the router can enable both the new and legacy planners and run them in parallel to compare their results. After their comparison, the router discards the native query planner's results and uses only the legacy planner to execute requests.
```
lrlna, line 15: So this happens by _default_ since 1.53.0. We are not switching the default of running both side by side for this release.
```suggestion
As part of the effort to ensure correctness and stability of the new query planner, starting in v1.53.0 the router enables both the new and legacy planners and runs them in parallel to compare their results by default. After their comparison, the router discards the native query planner's results and uses only the legacy planner to execute requests.
```

docs/source/executing-operations/native-query-planner.mdx
```diff
@@ -0,0 +1,55 @@
...
+
+The native query planner uses a single thread in the cold path of the router. It has a bounded queue of ten queries. If the queue is full, the router simply does not run the comparison to avoid excessive resource consumption.
```
lrlna, line 17: I am afraid this is not making it clear that this only happens in the comparison - `both` - mode 🤔 do you think we should change the title of this section to mention that this is only in comparison (which does run by default at the moment)

docs/source/configuration/overview.mdx
```diff
@@ -579,6 +579,31 @@ You can configure certain caching behaviors for generated query plans and APQ (b
...
+
+```yaml title="router.yaml"
+supergraph:
+  generate_query_fragments: true
+
```
lrlna, line 597: let's drop this for this configuration, but mention it as an additional configuration that makes native query planning more optimal?
```suggestion
```

docs/source/executing-operations/native-query-planner.mdx
```diff
@@ -0,0 +1,55 @@
...
+- `legacy` - enables only the legacy JavaScript query planner 
+- `both_best_effort` - enables both new and legacy query planners for comparison. The legacy query planner is used for execution. 
+
+<MinVersion version="2.0.0">
+
+## Optimize native query planner
+
+</MinVersion>
+
+To run the native query planner with the best performance and resource utilization, configure your router with the following options:
```
lrlna, line 35: This config not only enable native query planning, but also _does not_ startup v8, the javascript runtime, which is what we have to run in order to run the legacy query planner.

v8 eats through a lot of CPU and memory in a router process when it's running, and now that we have confidence in our native implementation, that we want users to disable it when running the router. This is the config that gets them the native query planner, and does not start v8.

docs/source/executing-operations/native-query-planner.mdx
```diff
@@ -0,0 +1,55 @@
...
+
+```yaml title="router.yaml"
+supergraph:
+  generate_query_fragments: true
+
```
lrlna, line 40: I've been thinking about this some more, and now I think this should be a separate point from the call to action of enabling the rust planner and disabling v8. Perhaps we can say something like:
> Additionally, to enable faster throughput by reducing the size of queries sent to subgraphs, and more optimal native query planning, enable query fragment generation:
>```yaml title="router.yaml"
> supergraph:
>   generate_query_fragments: true
>```
What do you think?

chandrikas: Agree - changing too many things at once might feel scary. Let's have clear outcomes 
* Test new QP
* Disable v8

docs/source/configuration/overview.mdx
shorgi: Will describe the JS runtime earlier

chandrikas (approved):

docs/source/executing-operations/native-query-planner.mdx
shorgi: edited

docs/source/executing-operations/native-query-planner.mdx
```diff
@@ -0,0 +1,65 @@
...
+Learn to run the GraphOS Router with the Rust-native query planner and improve your query planning performance and scalability.
+
+## Background about query planner implementations
+
+In v1.49.0 the router introduced a [query planner](/graphos/routing/about-router#query-planning) implemented natively in Rust. This native query planner improves the overall performance and resource utilization of query planning. It exists alongside the legacy Javascript implementation that uses the V8 JavaScript engine, and it will eventually replace the legacy implementation.
+
+### Comparing query planner implementations
+
+As part of the effort to ensure correctness and stability of the new query planner, starting in v1.53.0 the router can enable both the new and legacy planners and run them in parallel to compare their results by default. After their comparison, the router discards the native query planner's results and uses only the legacy planner to execute requests. The native query planner uses a single thread in the cold path of the router. It has a bounded queue of ten queries. If the queue is full, the router simply does not run the comparison to avoid excessive resource consumption.
```
lrlna, line 17: It runs `both_best_effort` by default, rather than it being a possibility to run both by default. Do you think this would be more clear?

```suggestion
As part of the effort to ensure correctness and stability of the new query planner, starting in v1.53.0 the router enables both the new and legacy planners and runs them in parallel to compare their results by default. After their comparison, the router discards the native query planner's results and uses only the legacy planner to execute requests. The native query planner uses a single thread in the cold path of the router. It has a bounded queue of ten queries. If the queue is full, the router simply does not run the comparison to avoid excessive resource consumption.
```

docs/source/executing-operations/native-query-planner.mdx
```diff
@@ -0,0 +1,65 @@
...
+minVersion: 1.49.0
+---
+
+<PreviewFeature />
+
+Learn to run the GraphOS Router with the Rust-native query planner and improve your query planning performance and scalability.
+
+## Background about query planner implementations
+
+In v1.49.0 the router introduced a [query planner](/graphos/routing/about-router#query-planning) implemented natively in Rust. This native query planner improves the overall performance and resource utilization of query planning. It exists alongside the legacy Javascript implementation that uses the V8 JavaScript engine, and it will eventually replace the legacy implementation.
```
lrlna, line 13:
```suggestion
In v1.49.0 the router introduced a [query planner](/graphos/routing/about-router#query-planning) implemented natively in Rust. This native query planner improves the overall performance and resource utilization of query planning. It exists alongside the legacy JavaScript implementation that uses the V8 JavaScript engine, and it will eventually replace the legacy implementation.
```

docs/source/executing-operations/native-query-planner.mdx
```diff
@@ -0,0 +1,65 @@
...
+To run the native query planner with the best performance and resource utilization, configure your router with the following options:
+
+```yaml title="router.yaml"
+experimental_query_planner_mode: new
+experimental_introspection_mode: new
+```
+
+Setting `experimental_query_planner_mode: new` and `experimental_introspection_mode: new` not only enables native query planning and schema introspection, it also disables the V8 JavaScript runtime used by the legacy query planner. Disabling V8 frees up CPU and memory and improves native query planning performance.
+
+Additionally, to enable more optimal native query planning and faster throughput by reducing the size of queries sent to subgraphs, you can enable query fragment generation with the following option:
```
lrlna, line 44: @chandrikas do you think should we mention here that this option will be the only option for handling fragments in the future?

also debating on whether we should like to our docs on [fragment reuse and generation](https://www.apollographql.com/docs/router/configuration/overview/#fragment-reuse-and-generation) and whether or not we should also update those docs under `generate_query_fragments` saying this will be the only way to deal with fragments in the future? 🤔 🤔 🤔

docs/source/configuration/overview.mdx
```diff
@@ -579,6 +579,35 @@ You can configure certain caching behaviors for generated query plans and APQ (b
 - You can configure a Redis-backed _distributed_ cache that enables multiple router instances to share cached values. For details, see [Distributed caching in the GraphOS Router](./distributed-caching/).
 - You can configure a Redis-backed _entity_ cache that enables a client query to retrieve cached entity data split between subgraph reponses. For details, see [Subgraph entity caching in the GraphOS Router](./entity-caching/).
 
+<MinVersion version ="1.49.0">
+
+### Native query planner
+
+</MinVersion>
+
+Starting with v1.49.0, the router can run a Rust-native query planner. This native query planner can be run by itself to plan all queries, replacing the legacy Javascript implementation.
```
lrlna, line 588:
```suggestion
Starting with v1.49.0, the router can run a Rust-native query planner. This native query planner can be run by itself to plan all queries, replacing the legacy JavaScript implementation.
```

docs/source/executing-operations/native-query-planner.mdx
shorgi: updated

PR #6053, opened 2024-09-25
Remove pruning when computing best plan

sachindshinde:
This PR removes option pruning in `compute_best_plan_from_closed_branches()`, as it has been removed from JS QP. (The reasoning was that it was observed pruning could save time in some cases, but could be substantially worse in a few others.) Note that we should also be able to get rid of overriding IDs, although that requires changing more code (and the JS QP hasn't done it yet either), so I'm leaving it to the future/later PRs.


lrlna (approved):

goto-bus-stop (approved):
✅

PR #6059, opened 2024-09-25
Updating redis and subgraphs configs to be children of preview_entity…

apollonauttrey:
…_cache

Customer shared that when they were configuring (through trial and error) they discovered that the `redis` config should be a child of `preview_entity_cache` - upon making this change, I believe the same is true for the `subgraphs` config.

*Description here*

Fixes #**issue_number**

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/docs/source/configuration/entity-caching.mdx
+++ b/docs/source/configuration/entity-caching.mdx
@@ -94,19 +94,19 @@ preview_entity_cache:
   subgraph:
     all:
       enabled: true
-      # Configure Redis
-      redis:
-        urls: ["redis://..."]
-        timeout: 2s # Optional, by default: 500ms
-        ttl: 24h # Optional, by default no expiration
-    # Configure entity caching per subgraph, overrides options from the "all" section
-    subgraphs:
-      products:
-        ttl: 120s # overrides the global TTL
-      inventory:
-        enabled: false # disable for a specific subgraph
-      accounts:
-        private_id: "user_id"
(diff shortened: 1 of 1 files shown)


Geal:
which version of the router are they using?

apollonauttrey:
1.50

Meschreiber (commented):
From this [migration](https://github.com/apollographql/router/blob/35d83a44ae713d584b41e7ffa22069ff0dbe8a2d/apollo-router/src/configuration/migrations/0026-entity_cache_invalidation_preview.yaml#L4) it looks as though this is the legacy format and the currently documented format is correct from `1.52` onward. (@Geal , could you confirm please?)

We can add a note that the format changed and that the `redis` and `subgraphs` configurations used to be directly on  `preview_entity_cache`, i.e. `preview_entity_cache.redis`. WDYT @apollonauttrey ?

apollonauttrey:
i think that's a reasonable compromise @Meschreiber - 1.50 was only late june/early july so it's possible someone else could run into this.

Geal:
note that this is still a preview feature, so things like configuration could change a bit before GA

Meschreiber:
Closing in favor of https://github.com/apollographql/router/pull/6083

PR #6061, opened 2024-09-25
Add deny lints to QP

TylerBloom:
This PR changes the crate-wide `#[allow(dead_code)]` lint directive to `deny` and `deny`s a few other lints.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-571]: https://apollographql.atlassian.net/browse/ROUTER-571?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apollo-federation/src/link/cost_spec_definition.rs
+++ b/apollo-federation/src/link/cost_spec_definition.rs
@@ -147,11 +147,6 @@ impl CostSpecDefinition {
         apollo_compiler::ast::DirectiveList,
         Node::new
     );
-    propagate_demand_control_directives!(
-        propagate_demand_control_schema_directives,
-        apollo_compiler::schema::DirectiveList,
-        Component::from
-    );
 
     propagate_demand_control_directives_to_position!(
         propagate_demand_control_directives_for_enum,
--- a/apollo-federation/src/query_graph/graph_path.rs
+++ b/apollo-federation/src/query_graph/graph_path.rs
@@ -769,14 +769,7 @@ impl Display for Unadvanceable {
 }
 
 #[derive(Debug, Clone, strum_macros::Display, serde::Serialize)]
-enum UnadvanceableReason {
-    UnsatisfiableKeyCondition,
-    UnsatisfiableRequiresCondition,
-    UnresolvableInterfaceObject,
-    NoMatchingTransition,
-    UnreachableType,
-    IgnoredIndirectPath,
-}
+enum UnadvanceableReason {}
 
 /// One of the options for a `ClosedBranch` (see the documentation of that struct for details). Note
 /// there is an optimization here, in that if some ending section of the path within the GraphQL
@@ -1983,12 +1976,6 @@ impl OpGraphPath {
             })
             .collect();
         (new_self, new_others)
-    }
-
-    pub(crate) fn is_overridden_by(&self, other: &Self) -> bool {
-        self.overriding_path_ids
-            .iter()
-            .any(|overriding_id| other.own_path_ids.contains(overriding_id))
     }
 
     pub(crate) fn subgraph_jumps(&self) -> Result<u32, FederationError> {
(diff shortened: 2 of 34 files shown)


apollo-federation/src/subgraph/database.rs
```diff
@@ -13,29 +13,21 @@
```
goto-bus-stop, line 13: I think we can remove this whole file. It's only used in a test

goto-bus-stop (commented):
I don't think we should `deny` unused code as it would prevent compilation if you're eg. commenting something out while working on a bug. It should be a warning.

apollo-federation/src/subgraph/database.rs
TylerBloom: Are you saying to move the content to where the test is or to also rewrite that test?

TylerBloom:
> I don't think we should `deny` unused code as it would prevent compilation if you're eg. commenting something out while working on a bug. It should be a warning.

My thought about `deny`ing these lints is that changing `deny` to `allow`/`warn` is easy to do while debugging or writing a feature. When the PR is ready, that can be changed back. However, we shouldn't merge a PR that has dead code, and`deny` is the easiest way to prevent that.

apollo-federation/src/subgraph/database.rs
goto-bus-stop: AFAICT it's only used in a test *that tests this file*, so we can remove this file and the test.

goto-bus-stop:
We shouldn't merge a PR with warnings 😛 

In apollo-rs we run the linter with `-D warnings` in CI, I guess we don't do that here. Anyways, it's just my personal preference, we don't have to do it 🙂

apollo-federation/src/query_graph/graph_path.rs
```diff
@@ -769,14 +769,7 @@ impl Display for Unadvanceable {
...
 #[derive(Debug, Clone, strum_macros::Display, serde::Serialize)]
-enum UnadvanceableReason {
-    UnsatisfiableKeyCondition,
-    UnsatisfiableRequiresCondition,
-    UnresolvableInterfaceObject,
-    NoMatchingTransition,
-    UnreachableType,
-    IgnoredIndirectPath,
-}
+enum UnadvanceableReason {}
```
goto-bus-stop, line 772:
```suggestion
// PORT_NOTE: This is only used by composition, which is not ported to Rust yet.
enum UnadvanceableReason {}
```

apollo-federation/src/schema/type_and_directive_specification.rs
```diff
@@ -778,6 +787,9 @@ mod tests {
     use crate::schema::type_and_directive_specification::DirectiveSpecification;
     use crate::schema::FederationSchema;
 
+    #[test]
+    fn dead_code_filter() {}
```
goto-bus-stop, line 791: What does this do?

apollo-federation/src/link/cost_spec_definition.rs
```diff
@@ -147,11 +147,6 @@ impl CostSpecDefinition {
         apollo_compiler::ast::DirectiveList,
         Node::new
     );
-    propagate_demand_control_directives!(
```
goto-bus-stop, line 150: @tninesling The `propagate_demand_control_schema_directives` function generated by this macro is dead code. Is it meant to be used?

goto-bus-stop (approved):
Some questions and notes in comments but I'm happy to see it merged either way.

apollo-federation/src/schema/type_and_directive_specification.rs
TylerBloom: Ah, missed this during my clean up. Rather than mark that module with `allow(dead_code)`, I was going to construct each of the types in this dummy test.

apollo-federation/src/link/cost_spec_definition.rs
tninesling: That's fine to be deleted. Looks like I forgot to delete it when I replaced it with the `to_position` version underneath

PR #6062, opened 2024-09-26
fix: emit connectors debugging log only once

lennyburdette:
with rover dev or hot reloading, plugins are initialized each supergraph update, leading to redundant logs

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/plugins/connectors/plugin.rs
+++ b/apollo-router/src/plugins/connectors/plugin.rs
@@ -8,7 +8,6 @@ use futures::stream::once;
 use futures::StreamExt;
 use http::HeaderValue;
 use itertools::Itertools;
-use lazy_static::lazy_static;
 use parking_lot::Mutex;
 use serde::Deserialize;
 use serde::Serialize;
@@ -29,9 +28,7 @@ const CONNECTORS_DEBUG_HEADER_NAME: &str = "Apollo-Connectors-Debugging";
 const CONNECTORS_DEBUG_ENV: &str = "APOLLO_CONNECTORS_DEBUGGING";
 const CONNECTORS_MAX_REQUESTS_ENV: &str = "APOLLO_CONNECTORS_MAX_REQUESTS_PER_OPERATION";
 
-lazy_static! {
-    static ref LAST_DEBUG_ENABLED_VALUE: Arc<Mutex<bool>> = Arc::new(Mutex::new(false));
-}
+static LAST_DEBUG_ENABLED_VALUE: AtomicBool = AtomicBool::new(false);
 
 #[derive(Debug, Clone)]
 struct Connectors {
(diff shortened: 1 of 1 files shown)


pubmodmatt:
If the `debug_extensions` value changes from `false` to `true`, and the config is reloaded, showing the warning again would be desirable because it has just been turned on again. That doesn't matter for `dev` since it defaults to `true` always there, but if there is a case outside of `dev` where the config would be reloaded, it seems preferable to emit the warning every time.

lennyburdette:
I tested it this way:

1. run router with `--hot-reload` and a config without connectors debugging: no warning message
2. change the router to enable connectors debugging: the router reloads and emits the warning

if you were to switch it off and back on, you wouldn't see the warning again, but i think that's fine. 

maybe in the future there could be a way to distinguish between supergraph updates and configuration updates, but until then i would prefer not to see this warning every time the supergraph updates.

pubmodmatt (approved):
Since we can't differentiate, it seems ok to optimize for the supergraph reloading case rather than the config reloading.

apollo-router/src/plugins/connectors/plugin.rs
```diff
@@ -45,11 +47,14 @@ impl Plugin for Connectors {
...
-            LOG_ONE_TIME.get_or_init(|| ());
+        {
+            if debug_extensions && *LAST_DEBUG_ENABLED_VALUE.lock() != debug_extensions {
+                tracing::warn!(
+                    "Connector debugging is enabled, this may expose sensitive information."
+                );
+            }
+
+            *LAST_DEBUG_ENABLED_VALUE.lock() = debug_extensions;
         }
```
pubmodmatt, line 58:
```suggestion
        let last_value = LAST_DEBUG_ENABLED_VALUE.load(Ordering::Relaxed);
        if LAST_DEBUG_ENABLED_VALUE
            .compare_exchange(
                last_value,
                debug_extensions,
                Ordering::Relaxed,
                Ordering::Relaxed,
            )
            .is_ok()
            && debug_extensions
        {
            tracing::warn!(
                "Connector debugging is enabled, this may expose sensitive information."
            );
        }
```

apollo-router/src/plugins/connectors/plugin.rs
```diff
@@ -29,7 +29,9 @@ const CONNECTORS_DEBUG_HEADER_NAME: &str = "Apollo-Connectors-Debugging";
 const CONNECTORS_DEBUG_ENV: &str = "APOLLO_CONNECTORS_DEBUGGING";
 const CONNECTORS_MAX_REQUESTS_ENV: &str = "APOLLO_CONNECTORS_MAX_REQUESTS_PER_OPERATION";
 
-static LOG_ONE_TIME: OnceLock<()> = OnceLock::new();
+lazy_static! {
+    static ref LAST_DEBUG_ENABLED_VALUE: Arc<Mutex<bool>> = Arc::new(Mutex::new(false));
+}
```
pubmodmatt, line 34:
```suggestion
static LAST_DEBUG_ENABLED_VALUE: AtomicBool = AtomicBool::new(false);
```

apollo-router/src/plugins/connectors/plugin.rs
```diff
@@ -9,6 +8,7 @@ use futures::stream::once;
 use futures::StreamExt;
 use http::HeaderValue;
 use itertools::Itertools;
+use lazy_static::lazy_static;
```
pubmodmatt, line 11:
```suggestion
```

apollo-router/src/plugins/connectors/plugin.rs
```diff
@@ -1,6 +1,5 @@
 use std::collections::HashMap;
 use std::sync::Arc;
```
pubmodmatt, line 2:
```suggestion
use std::sync::atomic::AtomicBool;
```

apollo-router/src/plugins/connectors/plugin.rs
```diff
@@ -42,7 +44,17 @@ impl Plugin for Connectors {
         let debug_extensions = init.config.debug_extensions
             || std::env::var(CONNECTORS_DEBUG_ENV).as_deref() == Ok("true");
 
-        if debug_extensions {
+        let last_value = LAST_DEBUG_ENABLED_VALUE.load(Ordering::Relaxed);
```
pubmodmatt, line 47: I missed a `use` for `Ordering` here in my suggestions, so your build failed 😬

PR #6067, opened 2024-09-26
Implement HTTP metrics for Connectors

pubmodmatt:
Add standard metrics for connectors HTTP requests and responses:

* `http.client.request.duration`
* `http.client.request.body.size`
* `http.client.response.body.size`

Selectors are defined for:

* `subgraph_name`
* `connector_source`
* `connector_http_request_header`
* `connector_http_response_header`
* `connector_http_response_status`
* `connector_http_method`
* `connector_url_template`
* `static`

Attributes:

* `subgraph.name`
* `connector.source.name`
* `connector.http.method`
* `connector.url.template`

The implementation uses the HTTP client service, as a single call to the connector service can result in multiple HTTP calls. The metrics make more sense at the HTTP client level anyway. However, since the instrumentation types are generic over a single request/response type, this means we'll be unable to have any other instruments at the connector service level that use the same selectors. However, that should be fine as the interesting data is at the HTTP level for connectors. 

Since the HTTP client service is not public, this required making the telemetry plugin private. This in turn required some small changes to the plugin test harness, which was previously incompatible with private plugins.

# Examples

Add a metric for the number of 404 response codes from a particular connector source API:

```yaml
telemetry:
  instrumentation:
    instruments:
      connector:
        not.found.count:
          value: unit
          type: counter
          unit: count
          description: "Count of 404 responses from the user API"
          condition:
            all:
              - eq:
                - 404
                - connector_http_response_status: code
              - eq:
                - "user_api"
                - connector_source: name
```

Add connector attributes to the `http.client.request.duration` instrument:

```yaml
http.client.request.duration:
  attributes:
    subgraph.name: true
    connector.source:
      connector_source: name
    connector.http.method: true
    connector.url.template: true
```

Create a histogram of the remaining rate limit from an API based on a response header value:

```yaml
rate.limit:
  value:
    connector_http_response_header: "x-ratelimit-remaining"
  unit: count
  type: histogram
  description: "Rate limit remaining"
  condition:
    eq:
      - "user_api"
      - connector_source: name
```

It's also now possible to configure events for connectors:

```yaml
telemetry:
  instrumentation:
    events:
      connector:
        request: off
        response: off
        error: error
        connector.request.event:
          message: "Connector request"
          level: info
          on: request
          attributes:
            http_method:
              connector_http_method: true
            url_template:
              connector_url_template: true
        connector.response.event:
          message: "Connector response"
          level: info
          on: response
          attributes:
            http_method:
              connector_http_method: true
            url_template:
              connector_url_template: true
            response_status:
              connector_http_response_status: code
```

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

**Notes**

[CNN-238]: https://apollographql.atlassian.net/browse/CNN-238?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apollo-federation/src/sources/connect/validation/graphql/strings.rs
```diff
@@ -109,7 +109,7 @@ mod tests {
         };
         let field = query.fields.get("field").unwrap();
         let directive = field.directives.get("connect").unwrap();
-        directive.argument_by_name(name).unwrap()
+        directive.argument_by_name(name, schema).unwrap()
```
lennyburdette, line 112: this was fixed on dev with a different method

apollo-router/src/plugins/telemetry/config_new/attributes.rs
```diff
@@ -264,6 +272,78 @@ impl DefaultForLevel for SubgraphAttributes {
     }
 }
 
+#[derive(Deserialize, JsonSchema, Clone, Default, Debug)]
+#[serde(deny_unknown_fields, default)]
+pub(crate) struct ConnectorHttpAttributes {
```
bnjjj, line 277: I think we should put this in subgraph attributes. Let's only keep subgraph attributes for now, and as it's namespaced by connector. It will be enough, I also think we shouldn't have both `connector.http.method` and `subgraph.http.method` we should just keep the subgraph one for now. If people wants to know if it's a connector or a subgraph I would suggest adding in `SubgraphSelelector` something like `service_kind` which could be `connector` or `subgraph`. (I don't think `service_kind` is the right naming but you got the idea).

apollo-router/src/plugins/telemetry/config_new/instruments.rs
```diff
@@ -81,6 +86,11 @@ pub(crate) struct InstrumentsConfig {
         SubgraphInstrumentsConfig,
         Instrument<SubgraphAttributes, SubgraphSelector, SubgraphValue>,
     >,
+    /// Connector service instruments. For more information see documentation on Router lifecycle.
+    pub(crate) connector: Extendable<
```
bnjjj, line 90: Same here, I think everything should be in subgraph for now. To keep only subgraph. Because I think usually people will want to have the same attributes even if it's a connector or subgraph (if not they would still be able to know using the selector and a condition to add specific attributes or metrics if it's a connector for example).

apollo-router/src/plugins/telemetry/config_new/instruments.rs
```diff
@@ -685,6 +695,227 @@ impl InstrumentsConfig {
...
+                });
+        ConnectorHttpInstruments {
+            http_client_request_duration,
+            http_client_request_body_size,
+            http_client_response_body_size,
+            custom: CustomInstruments::new(&self.connector.custom, static_instruments),
+        }
+    }
+
+    pub(crate) fn new_builtin_connector_instruments(&self) -> HashMap<String, StaticInstrument> {
```
bnjjj, line 829: I think this one is basically the same than subgraph instruments so probably one more reason to merge everything in subgraph

apollo-router/src/plugins/telemetry/config_new/selectors.rs
```diff
@@ -3569,4 +3852,203 @@ mod test {
             Some("default".into())
         );
     }
+
```
bnjjj, line 3855: We need tests including both subgraph calls and connectors calls to make sure it works properly

apollo-router/src/plugins/telemetry/mod.rs
```diff
@@ -851,6 +858,58 @@ impl Plugin for Telemetry {
             .boxed()
     }
 
+    fn http_client_service(
```
bnjjj, line 861: I might be wrong but I'm wondering here what happens if I try to count for `http.client.request.duration` and it's a subgraph and it also contains connector attributes. For a subgraph we will pass both in subgraph service and http client service. Which means we could measure the value twice and so increment twice instead of just one. I think with the idea of having everything in subgraph service we should just move the subgraph logics here. During our discussion I thought we have a connector_service or so that would allow us to not move the current logics of subgraph into http_client_service. So I think we have 2 options, either we put this in a connector service or we migrate everything into the http_client_service (the latter is a more long term strategy I think). But that means we will have to move some stuffs into the context to make sure we have access to the same data we need for current subgraph selectors

bnjjj (changes requested):
Example of config for instruments:

```yaml
telemetry:
  instrumentation:
    instruments:
      subgraph:
        http.client.request.duration:
          attributes:
            subgraph.name: true
            connector.source:
              connector_source: name
            subgraph.http.method: true
            subgraph.graphql.operation.type: string
            connector.url.template: true
        specific.metrics.for.a.connector:
          condition:
            eq:
            - service_kind: true
            - connector
          value:
            connector_http_response_header: "x-ratelimit-remaining"
          unit: count
          type: histogram
          description: "my desc"
```

We also have to think about testing other instrumentation elements like events and spans using the new selectors and attributes

bnjjj:
cc @BrynCooke I need your review here too

apollo-router/src/plugins/telemetry/config_new/attributes.rs
pubmodmatt: The issue here is that the relationship between subgraph and connector is not 1-to-1. There can be a number of connectors on a given subgraph, and there can be multiple calls to each connector per GraphQL request. So a connector isn't really a "kind" of subgraph - the subgraph service lifecycle stage is not invoked for connectors. The lifecycle is really now like this:

```
                 ------ subgraph service ------ http client service
                |
fetch service --|
                |                                ------- http client service
                 ------ connector service ------|
                                                 ------- http client service
```

The fetch service and connector service aren't currently exposed as plugin lifecycle hooks, but likely will be.

apollo-router/src/plugins/telemetry/config_new/instruments.rs
pubmodmatt: Another issue here is that HTTP connectors are only *one kind* of connector - there will be more. So unlike subgraphs, where every request is HTTP, we may have connector requests that are SQL, gRPC, etc. As we go forward with these, sharing attributes between connectors and subgraph would make less and less sense, since the attributes would never apply to a subgraph.

apollo-router/src/plugins/telemetry/config_new/instruments.rs
pubmodmatt: The request and response types limit things here. You can't have attributes or selectors that are for *both* subgraph request/response *and* http request/response - they can only be generic over one type of request/response. That limits how things can be arranged.

apollo-router/src/plugins/telemetry/mod.rs
pubmodmatt: This is the reason for the line:

```rust
.contains_key(CONNECTOR_INFO_CONTEXT_KEY)
```

It will currently only invoke the connector http `on_request` and `on_response` for connectors, not subgraphs, so it will only be called once.

apollo-router/src/plugins/telemetry/config_new/attributes.rs
bnjjj: Hmm indeed, I didn't know we could have several http calls for a connector. Any idea @BrynCooke ?

BrynCooke: I think what's here is the way forward, and then we add `http_client` service support to the telemetry plugin.

I know it's a big job, but it feels like the right thing to do.

BrynCooke (changes requested):
In general this is looking good. Can we pull the connectors stuff out into a separate module though rather than adding to the already large `attributes.rs` `instruments.rs` and `selectors.rs`

We should eventually be aiming for a separate module for each router pipeline stage.

apollo-router/src/plugins/telemetry/config_new/selectors.rs
```diff
@@ -3569,4 +3852,203 @@ mod test {
...
+        context
+    }
+
+    #[fixture]
+    fn http_request(context: RouterContext) -> HttpRequest {
+        HttpRequest {
+            http_request: http::Request::builder().body("".into()).unwrap(),
+            context,
+        }
+    }
```
bnjjj, line 3888: I don't feel confident introducing this technique to debug. We try as much as we can to lower the compilation time and if we introduce macros in our tests it will increase the time. Could you try to use the same kind of tests we did before. If it's something you would like to introduce in the router codebase then I would suggest to talk about this during one of our router arch meeting because we need opinion from the team. Also we have yaml based tests in `apollo-router/src/plugins/telemetry/config_new/fixtures` Could you add more tests directly there please ?

apollo-router/src/plugins/telemetry/mod.rs
bnjjj: Oh ok I missed that,in the future we will be able to get rid of that if we move everything in the http_client_service

bnjjj (changes requested):
It just needs more tests for instruments but also for events and spans, feel free to ask if you need help or if you don't find these existing tests

apollo-router/src/plugins/telemetry/config_new/selectors.rs
pubmodmatt: I added a test with both connectors and subgraph, let me know if it's sufficient.

apollo-router/src/plugins/telemetry/config_new/selectors.rs
pubmodmatt: > I don't feel confident introducing this technique to debug. We try as much as we can to lower the compilation time and if we introduce macros in our tests it will increase the time. Could you try to use the same kind of tests we did before. If it's something you would like to introduce in the router codebase then I would suggest to talk about this during one of our router arch meeting because we need opinion from the team. Also we have yaml based tests in `apollo-router/src/plugins/telemetry/config_new/fixtures` Could you add more tests directly there please ?

I replaced the use of `rstest` with just plain unit tests, and added more tests in `fixtures`.

pubmodmatt:
> It just needs more tests for instruments but also for events and spans, feel free to ask if you need help or if you don't find these existing tests

I've added support for connectors HTTP events, and added tests for that in the places I found existing tests for other events. Please let me know if I didn't find the tests you were looking for.

Spans are another matter - I haven't added support for spans in this PR. The HTTP client service isn't really set up for this - it creates the `http_request` span internally, so there's no way to add attributes to it at the plugin hook level. This is something we can look at adding in the future, but changes here could impact other uses of that service besides connectors.

pubmodmatt:
> In general this is looking good. Can we pull the connectors stuff out into a separate module though rather than adding to the already large `attributes.rs` `instruments.rs` and `selectors.rs`
> 
> We should eventually be aiming for a separate module for each router pipeline stage.

I moved connectors related things out to a `connectors` module, and there's an `http` module under that for these HTTP specific things.

apollo-router/src/services/connector_service.rs
```diff
@@ -155,6 +179,12 @@ async fn execute(
             // inner result fails just that one task, but an `Err` on the outer result cancels all the
             // tasks and fails the whole operation.
             let context = context.clone();
+            if context
+                .insert(CONNECTOR_INFO_CONTEXT_KEY, ConnectorInfo::from(connector))
+                .is_err()
+            {
+                warn!("Failed to store connector info in context - instruments may be inaccurate");
```
bnjjj, line 186: I would log it as an error

docs/source/configuration/telemetry/instrumentation/events.mdx
```diff
@@ -17,7 +17,7 @@ You can configure events for each service in `router.yaml`. Events can be standa
 
 <RouterServices />
 
-The `router`, `supergraph` and `subgraph` sections are used to define custom event configuration for each service:
+The `router`, `supergraph`, `subgraph` and `conenctor` sections are used to define custom event configuration for each service:
```
bnjjj, line 20:
```suggestion
The `router`, `supergraph`, `subgraph` and `connector` sections are used to define custom event configuration for each service:
```

docs/source/configuration/telemetry/instrumentation/selectors.mdx
```diff
@@ -96,6 +96,24 @@ The subgraph service executes multiple times during query execution, with each e
...
+
+#### HTTP
+
+Apollo Connectors for REST APIs make HTTP calls to the upstream HTTP API. These selectors let you extract metrics from these HTTP requests and responses.
+
+| Selector                         | Defaultable | Values           | Description                                                       |
+|----------------------------------|-------------|------------------|-------------------------------------------------------------------|
+| `subgraph_name`                  | No          | `true`\|`false`  | The name of the subgraph containing the connector                 |
+| `connector_source `              | No          | `name`           | The name of the `@source` associated with this connector, if any  |
+| `connector_http_request_header`  | Yes         | `true`\|`false`  | The name of a connector request header                            |
```
bnjjj, line 109: Are you sure about the values column here ? Is it really true or false and not the header name ? A string ?

docs/source/configuration/telemetry/instrumentation/selectors.mdx
```diff
@@ -96,6 +96,24 @@ The subgraph service executes multiple times during query execution, with each e
...
+| `connector_http_response_header` | Yes         | `true`\|`false`  | The name of a connector response header                           |
```
bnjjj, line 110: same

bnjjj (changes requested):
This looks great, could you also open an issue explaining the connector selectors are not already supported for spans link it directly in this PR please ? Also could be great to document that limitation in selectors docs and spans docs probably.

pubmodmatt:
> This looks great, could you also open an issue explaining the connector selectors are not already supported for spans link it directly in this PR please ? Also could be great to document that limitation in selectors docs and spans docs probably.

https://apollographql.atlassian.net/browse/CNN-476

bnjjj:
Hey @pubmodmatt I opened a PR https://github.com/apollographql/router/pull/6125 to also add the ability to configure span attributes for connectors. It's wip and needs tests. I'll probably need your help to locally set up a supergraph working with connectors and also add integration tests. 

When writing this PR I noticed couple of things I didn't see when reviewing. 
+ In order to be consistent with our current codebase I think the way you're handling rust modules should switch to the `mod.rs` model. You'll see the changes in my PR. If you disagree then I think that's something we should discuss as a team and keep this mechanism everywhere in order to be consistent and not create confusions.
+ Also for consistency I think we should not have `http` subsection in the connector instruments and connector events configuration as the attributes and metrics are already prefixed by `http.*` and that's how we did for other instruments and events.

pubmodmatt:
> * In order to be consistent with our current codebase I think the way you're handling rust modules should switch to the `mod.rs` model. You'll see the changes in my PR. If you disagree then I think that's something we should discuss as a team 
and keep this mechanism everywhere in order to be consistent and not create confusions.

Use of `mod.rs` has been discouraged [since Rust 1.3](https://doc.rust-lang.org/reference/items/modules.html#module-source-filenames). That said, since the codebase is still using it, I can see the a consistency argument. However, note that one of the reasons this is discouraged is for consistency. For example, the existing `telemetry` module has submodules `config` in `telemetry/config.rs` and `config_new` in `telemetry/config_new/mod.rs`. With the newer style, they'd be more consistently `telemetry/config.rs` and `telemetry/config_new.rs`.

> * Also for consistency I think we should not have `http` subsection in the connector instruments and connector events configuration as the attributes and metrics are already prefixed by `http.*` and that's how we did for other instruments and events.

I'm open to suggestions on this. My concern is when we add a new connector for SQL, and presumably a new SQL client service with its own request and response types, for example. Say we have this;

```yaml
telemetry:
  instrumentation:
    events:
      connector:
        request: off
        response: off
```

That would turn off events for both HTTP and SQL. There would be no way to turn them on separately. Alternatively, this would give the ability to turn HTTP or SQL events on or off separately:

```yaml
telemetry:
  instrumentation:
    events:
      connector:
        http:
          request: off
          response: info
        sql:
          request: off
          response: off
```

bnjjj:
Both of your arguments are correct. For modules I would suggest adding this topic for the next router arch meeting, by doing this we could decide together what's the best strategy to adopt. 

For the http section I was thinking about something a little bit different for your use case. That's already what we can do for specific GraphQL operation for example. But basically as you can set conditions on events I was thinking about using conditions to say something like if connector.kind is sql then display events. That's already what we do. What do you think ?

pubmodmatt:
> if connector.kind is sql then display events. That's already what we do. What do you think ?

That makes sense. I'll need to add connector type as an attribute, and for now it will just always be `http`.

pubmodmatt:
> For modules I would suggest adding this topic for the next router arch meeting

Done.

bnjjj:
>Actually, this won't work for the same reasons I mentioned earlier - the StandardEventConfig and Condition types are generic on the Selector type, and the Selector type has associated types for the request and response types. So you need separate config for each kind of connector, as they have different request and response types.

I agree about the implementation but I don't think our implementation issues should be reflected in the configuration so I think we should use `flatten` from serde like we did [here](https://github.com/apollographql/router/blob/next/apollo-router/src/plugins/telemetry/config_new/attributes.rs#L114). I think it would solve the implementation. What do you think ?

pubmodmatt:
> our implementation issues should be reflected in the configuration so I think we should use `flatten` from serde like we did [here](https://github.com/apollographql/router/blob/next/apollo-router/src/plugins/telemetry/config_new/attributes.rs#L114). I think it would solve the implementation. What do you think ?

Agreed that the configuration shouldn't expose these implementation issues. I think we can solve for the request and response types using enums - working on coding that up now. I'll look at `flatten` as well.

bnjjj:
@pubmodmatt One last thing, could you add metrics for connectors like we have for subgraph here https://github.com/apollographql/router/blob/pubmodmatt/connectors/instruments/apollo-router/src/configuration/metrics.rs#L350 please ?

BrynCooke (approved):

BrynCooke:
> @pubmodmatt One last thing, could you add metrics for connectors like we have for subgraph here https://github.com/apollographql/router/blob/pubmodmatt/connectors/instruments/apollo-router/src/configuration/metrics.rs#L350 please ?

I think this could be a separate PR.

pubmodmatt:
> One last thing, could you add metrics for connectors like we have for subgraph here

Thanks @bnjjj - I didn't know about that metric. I'll create a new PR to add that.

PR #6069, opened 2024-09-27
chore: update router-bridge@0.6.3+v2.9.2

dariuszkuc:
Update to latest router-bridge and federation version. This federation version fixes sibling `__typename` optimization (that will reduce JS vs RS QP mismatch) and `@context/@fromContext` fix for handling argument name collisions.

### DIFF
--- a/.changesets/chore_update_router_bridge_292.md
+++ b/.changesets/chore_update_router_bridge_292.md
@@ -1,0 +1,5 @@
+### Update to Federation v2.9.2 ([PR #6069](https://github.com/apollographql/router/pull/6069))
+
+This release updates to Federation v2.2.2, which fixes sibling `__typename` optimization (that will reduce JS vs RS QP mismatch) and `@context/@fromContext` fix for handling argument name collisions.
+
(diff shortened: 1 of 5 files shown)


lrlna (approved):

.changesets/chore_update_router_bridge_292.md
```diff
@@ -0,0 +1,5 @@
+### Update to Federation v2.9.2 ([PR #6069](https://github.com/apollographql/router/pull/6069))
+
+This release updates to Federation v2.9.2, which fixes sibling `__typename` optimization (that will reduce JS vs RS QP mismatch) and `@context/@fromContext` fix for handling argument name collisions.
```
goto-bus-stop, line 3: User-facing text so let's not mention the mismatch ratio, which they should not care about:
```suggestion
This release updates to Federation v2.9.2, with a small fix to the internal `__typename` optimization and a fix to prevent argument name collisions in the `@context`/`@fromContext` directives.
```

goto-bus-stop (approved):

PR #6073, opened 2024-09-27
add more tests for otlp with datadog

bnjjj:
More tests for otlp using datadog agent

### DIFF
--- a/apollo-router/tests/integration/telemetry/otlp.rs
+++ b/apollo-router/tests/integration/telemetry/otlp.rs
@@ -1,5 +1,6 @@
 extern crate core;
 
+use std::collections::HashMap;
 use std::collections::HashSet;
 use std::time::Duration;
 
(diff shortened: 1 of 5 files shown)


apollo-router/tests/integration/telemetry/otlp.rs
```diff
@@ -208,7 +347,7 @@ async fn test_untraced_request_sample_datadog_agent_unsampled() -> Result<(), Bo
 #[tokio::test(flavor = "multi_thread")]
 async fn test_priority_sampling_propagated() -> Result<(), BoxError> {
     if !graph_os_enabled() {
-        return Ok(());
+        panic!("Error: test skipped because GraphOS is not enabled");
```
BrynCooke, line 350: This will mean users that don't have a graphOS key will not be able to have passing tests

BrynCooke (changes requested):
Looks great but don't think we want to panic if graphOS isn't enabled. I am not sure any of these tests need graphOS anyway, so maybe we just remove the check.

PR #6074, opened 2024-09-27
chore(federation): borrow selection keys

goto-bus-stop:
This PR changes the implementation of `SelectionMap` so it does not have owned key data.

I'm opening this already so I can share it but I would like to improve the PR text a bit after our standup 😛 

Previously, to use `SelectionMap`, each selection type provided a `key() -> SelectionKey` method, returning some owned data, in practice two `Arc`s. Keys are copied frequently. However, key data must always already exist in the selection itself, so this is quite wasteful.

Selection keys now *borrow* from their Selection. They are `Copy`, so there is no overhead[^1] in creating or dropping them. They are meant to be created whenever they're necessary, not stored.

Because keys are now "free" to create, there is no need to cache them in `Field`, `FragmentSpread`, and `InlineFragment` structures. As a result, there is now also no need for the `Field`/`FieldData` split. The `Field`/`FieldData` data structures existed to disallow arbitrary mutable access to `FieldData` that may affect the cached key in the `Field` wrapper. This limitation is lifted, as keys are now only used by `SelectionMap`: you can freely mutate a `Field` or a `FieldSelection`, as long as it is not in a `SelectionMap`. `SelectionMap` only hands out mutable access through `FieldSelectionValue`, which enforces the mutability constraints.

I found this to improve performance across the board by 1-3%. That's not super impressive, but not bad either for something that affects *every* plan. I think this is also a stepping stone towards having mutable selections/selection sets so we won't need to copy so much.

## Review guidance
There's a lot of code changes here. I suggest reviewing the different groups of changes one by one. I'd group them like this:
- Change the SelectionMap implementation to not store `SelectionKey`s
  - This uses `hashbrown`'s `RawTable` and a Vec instead of IndexMap. The implementation is similar to IndexMap, except that keys are borrowed as necessary. Some unsafe code is required to work with RawTable.
  - The main work is in https://github.com/apollographql/router/commit/738395ecf51698b03781cb6a63313feb770dcac4
  - And a follow-up to optimize removing individual selections: https://github.com/apollographql/router/commit/f5a732891f7c2dd83a873df36b4192430b3deb64
  - And making RawTable equality function easier to read: https://github.com/apollographql/router/commit/99a6c531b9ab0cc145dbfffda809a92f1a36679a
- Change `SelectionKey` to be borrowed
  - Move `SelectionKey` to the `selection_map` module: https://github.com/apollographql/router/commit/7f44b3298ebbb1f5f809b51740f5ec23bf555f85 
  - Remove the `iter()` and `iter_mut()` iterators from SelectionMap: https://github.com/apollographql/router/commit/7d7671cfe4f92250e3c16ba54863388253cbed70
    - iter_mut is not possible to implement with borrowed keys because it would have to yield `(SelectionKey<'a>, SelectionValue<'a>)`, i.e. an immutable and a mutable reference to the same selection. The methods are unnecessary as you can trivially derive the key from the value.
  - The actual migration: https://github.com/apollographql/router/commit/bc4ac9a403099a442273f0548545d01e66f07bc0
  - There is an `OwnedSelectionKey` type that's used in selection set merging, the way that's written requires an owned key structure, that can be adjusted in the future but I decided to keep it like this also because we recently had to revert a change to this code
- Combining the selection "Data" and wrapper structures:
  - Adjust `InlineFragment::can_rebase_on` and `InlineFragmentData::can_rebase_on` so the names won't collide after the merge https://github.com/apollographql/router/commit/06554bae9da9cd36f682d729165dce3eab9cf88e
  - Combine `Field` and `FieldData` into a single type `Field`: https://github.com/apollographql/router/commit/00a95ecbed9ad33b113b1d8e1b0137da0ffe0442
  - `InlineFragment`: https://github.com/apollographql/router/commit/381459fdecb3267dd18a22c745e7343d3d07f6ce
  - `FragmentSpread`: https://github.com/apollographql/router/commit/56003462d6ef378955d33b520299a100ff0fb9e8
- And some cleanup:
  - Move a method from `SelectionKey` to `Selection`: https://github.com/apollographql/router/commit/d3b2e3c50357cedd452bd89649b4a971faf63fca (its only usage site was `selection.key().is_typename_field()`)

[^1]: I mean, it's still some memory being moved around, but it's not very substantial.


apollo-federation/src/operation/mod.rs
```diff
@@ -1443,33 +906,46 @@ mod fragment_spread_selection {
...
+
+    impl FragmentSpread {
+        pub(super) fn directives_mut(&mut self) -> &mut DirectiveList {
+            &mut self.directives
+        }
+    }
+
+    impl HasSelectionKey for FragmentSpread {
+        fn key(&self) -> SelectionKey<'_> {
             if is_deferred_selection(&self.directives) {
```
goto-bus-stop, line 925: This is a tradeoff right now: we call `key()` much more frequently, and for inline fragments and fragment spreads it has to search through the directive list. In many cases it'll be free (because the directive list is `None`), but in some cases it will have to walk the list and do a few string comparisons. It could potentially be optimized by storing a `.has_defer` boolean in `DirectiveList`, but I didn't see it on flame graphs yet, so that could be overengineering it.

apollo-federation/src/operation/merging.rs
```diff
@@ -206,7 +206,7 @@ impl SelectionSet {
                             ));
                         };
                         fields
-                            .entry(other_key)
+                            .entry(other_key.to_owned_key())
```
goto-bus-stop, line 209: This is not easy to rewrite to borrowed keys, because it creates intermediate maps grouping selections that have the same key. We can probably rewrite merging without the intermediate maps entirely later.

apollo-federation/src/operation/selection_map.rs
```diff
@@ -0,0 +1,642 @@
...
+                self.table.insert_no_grow(hash, index);
+            }
+        }
+    }
+
+    /// Decrements all the indices in the table starting at `pivot`.
+    fn decrement_table(&mut self, pivot: usize) {
+        // SAFETY: `self.table` must outlive `iter`
+        // None of the buckets can exit the function
+        unsafe {
```
TylerBloom, line 315: Since this is the first intro of `unsafe` into the crate, should we add the [`clippy::missing_safety_doc`](https://rust-lang.github.io/rust-clippy/master/index.html#/missing_safety_doc) safety lint to the crate-wide `warn` lints?

apollo-federation/src/operation/selection_map.rs
```diff
@@ -0,0 +1,642 @@
...
+        }
+    }
+
+    /// Add selections from another selection map to this one. If there are key collisions, the
+    /// selections are *overwritten*.
+    pub(crate) fn extend_ref(&mut self, other: &SelectionMap) {
+        for selection in other.values() {
+            self.insert(selection.clone());
+        }
+    }
```
TylerBloom, line 404: Is there a reason to have this over implementing the `extend` trait directly?

goto-bus-stop:
I'll update this to use `hashbrown::HashTable` over `hashbrown::raw::RawTable`. Drafting until then

apollo-federation/src/operation/selection_map.rs
goto-bus-stop: Resolved as `hashbrown::HashTable`-based implementation does not require unsafe.

goto-bus-stop:
Rebased, tried to rewrite commit history to make a little more sense, will write a new PR description and fix linting before undrafting for review.

apollo-federation/src/operation/selection_map.rs
```diff
@@ -0,0 +1,647 @@
...
+    /// Remove a selection from the map. Returns the selection and its numeric index.
+    pub(crate) fn remove(&mut self, key: SelectionKey<'_>) -> Option<(usize, Selection)> {
+        let hash = self.hash(key);
+        let entry = self
+            .table
+            .find_entry(hash, key_eq(&self.selections, key))
+            .ok()?;
+        let (bucket, _) = entry.remove();
+        let selection = self.selections.remove(bucket.index);
+        self.decrement_table(bucket.index);
```
TylerBloom, line 334: I think we can be slight more efficient here. Rather than iterating through the entire table, iterate through `self.selections` starting at `bucket.index`, get that key's bucket, and decrement. Roughly this:
```rust
let selection = self.selections.remove(bucket.index);
for selection in &self.selection[bucket.index..] {
  self.table.find_mut(hash, key_eq(&self.selections, key)).unwrap().index -= 1;
}
```

goto-bus-stop: It's a good thought, but I think it's not a slam-dunk improvement... The table is an array internally, so going through it linearly from beginning to end is going to be very fast, while starting from the selections will require jumping around in the memory. I think it's sometimes going to be better and sometimes worse. I like the simplicity of iterating the whole thing.

dariuszkuc (approved):

TylerBloom (approved):

abernix (approved):
license change

PR #6076, opened 2024-09-27
[JSONSelection] Allow `NamedSelection ::= Path SubSelection | ...` without an `Alias` for flattening multiple deeply nested properties

benjamn:
It was pointed out to me recently (by @dylan-apollo and @lennyburdette) that JSONSelection strings sometimes become unnecessarily verbose when you want to include one or more fields from a higher level of an input object, along with multiple fields extracted from some deeply nested path within that object.

For example, the response shape returned by the [OpenAI `/chat/completions` endpoint](https://platform.openai.com/docs/api-reference/chat) buries the actual response text from the LLM within an array of `choices`, containing objects with a `message` property that provides the `role` and `content` of the response, where the `content` is the desired response from the LLM (irrelevant properties omitted for clarity):
```json
{
  "id": "chatcmpl-123",
  "created": 1677652288,
  "model": "gpt-4o-mini",
  "choices": [{
    "message": {
      "role": "assistant",
      "content": "\n\nHello there, how may I assist you today?",
    },
  }]
}
```

With the current JSONSelection syntax, if you wanted to select the `id`, `created`, and `model` fields while also flattening the `role` and `content` fields, you'd have to use a selection like
```graphql
id
created
model
role: choices->first.message.role
content: choices->first.message.content
```
In this example, the repetition of the `choices->first.message.` syntax feels somehow unnecessary, and it gets even worse if those properties are more deeply nested, and/or there are more than two of them.

To build an intuition for the solution I'm about to propose, remember that you don't need this repetition if you're only selecting the deeply nested properties (not `id` or `created` or `model`), because you can use a single top-level path selection that selects the `role` and `content` properties (as an object) from the `choices->first.message` path:
```graphql
choices->first.message { role content }
```

Wouldn't it be nice if you could keep using that syntax, and just add other fields to it as necessary?
```graphql
id
created
model
choices->first.message { role content }
```

In terms of the JSONSelection grammar, this new syntax is available and unambiguous because the grammar currently requires adding an alias to the path selection if you want to use it together with other named selections:
```graphql
id
created
model
roleAndContent: choices->first.message { role content }
```
This is not quite what we want in this case, because `roleAndContent` becomes the only sibling key to `id`, `created` and `model`, and there's no syntax for merging `role` and `content` as siblings at that level. However, the absence of a syntax for that merging behavior gives us the freedom to choose the syntax we want to use.

Of course, the desired syntax only works when `choices->first.message` evaluates to an object, but we can enforce that expectation by requiring the trailing `{ role content }` subselection, which produces a runtime `ApplyToError` if `choices->first.message` is not an object or does not have those properties (which is the best we can do, since the actual runtime value of `choices->first.message` depends on dynamic input data).

In other words, the following syntax should continue to be forbidden, even if we happen to know `choices->first.message` is an object with only the `role` and `content` fields:
```graphql
id
created
model
 # STILL ILLEGAL if there are any other named properties!
choices->first.message
```

This new syntax respects [guiding principle #3 (safe subsetting)](https://github.com/apollographql/router/blob/next/apollo-federation/src/sources/connect/json_selection/README.md#guiding-principles), because a selection consisting of only
```graphql
choices->first.message { role content }
```
is functionally equivalent to the selection
```graphql
role: choices->first.message.role
content: choices->first.message.content
```
and continues to contribute the `role` and `content` properties in the same way even when you add/remove other named sibling properties to/from the selection (such as `id`, `created`, and `model`), which can happen any time your GraphQL query happens not to select those properties.

We considered requiring some sort of syntax to indicate the merging behavior, such as `... choices->first.message { role content }`, but having to add or remove the `...` syntax according to the presence or absence of other named selections seemed burdensome and inconsistent. Instead, I think this PR gives a natural meaning to path selections that do not have an alias but do have a trailing subselection, which were previously allowed only at the top-level, by themselves, and forbidden within any nested `SubSelection`.

The first commit in the PR refactors the singular `NamedSelection::name` method to a plural `NamedSelection::names` method that can now return multiple names. I would appreciate @nicholascioli and @pubmodmatt's eyes on that commit especially, since it affects code outside the `json_selection` directory that they're familiar with. Those changes look more drastic than they are because I needed to reindent some existing code.

### DIFF
--- a/apollo-federation/src/sources/connect/expand/tests/snapshots/it_expand_supergraph@keys.graphql-2.snap
+++ b/apollo-federation/src/sources/connect/expand/tests/snapshots/it_expand_supergraph@keys.graphql-2.snap
@@ -501,19 +501,21 @@ input_file: apollo-federation/src/sources/connect/expand/tests/schemas/expand/ke
                         None,
                     ),
                     Path(
-                        Alias {
-                            name: WithRange {
-                                node: Field(
-                                    "id2",
-                                ),
+                        Some(
+                            Alias {
+                                name: WithRange {
+                                    node: Field(
+                                        "id2",
+                                    ),
+                                    range: Some(
+                                        3..6,
+                                    ),
+                                },
                                 range: Some(
-                                    3..6,
+                                    3..7,
                                 ),
                             },
-                            range: Some(
-                                3..7,
-                            ),
-                        },
+                        ),
                         PathSelection {
                             path: WithRange {
                                 node: Var(
@@ -613,19 +615,21 @@ input_file: apollo-federation/src/sources/connect/expand/tests/schemas/expand/ke
                     SubSelection {
                         selections: [
                             Path(
-                                Alias {
-                                    name: WithRange {
-                                        node: Field(
-                                            "id",
-                                        ),
+                                Some(
+                                    Alias {
+                                        name: WithRange {
+                                            node: Field(
+                                                "id",
+                                            ),
+                                            range: Some(
+                                                0..2,
+                                            ),
+                                        },
                                         range: Some(
-                                            0..2,
+                                            0..3,
                                         ),
                                     },
-                                    range: Some(
-                                        0..3,
-                                    ),
-                                },
+                                ),
                                 PathSelection {
                                     path: WithRange {
                                         node: Var(
@@ -761,19 +765,21 @@ input_file: apollo-federation/src/sources/connect/expand/tests/schemas/expand/ke
                     SubSelection {
                         selections: [
                             Path(
-                                Alias {
-                                    name: WithRange {
-                                        node: Field(
-                                            "id",
-                                        ),
+                                Some(
+                                    Alias {
+                                        name: WithRange {
+                                            node: Field(
+                                                "id",
+                                            ),
+                                            range: Some(
+                                                0..2,
+                                            ),
+                                        },
                                         range: Some(
-                                            0..2,
+                                            0..3,
                                         ),
                                     },
-                                    range: Some(
-                                        0..3,
-                                    ),
-                                },
+                                ),
                                 PathSelection {
                                     path: WithRange {
                                         node: Var(
@@ -836,19 +842,21 @@ input_file: apollo-federation/src/sources/connect/expand/tests/schemas/expand/ke
                         None,
                     ),
                     Path(
-                        Alias {
-                            name: WithRange {
-                                node: Field(
-                                    "id2",
-                                ),
+                        Some(
+                            Alias {
+                                name: WithRange {
+                                    node: Field(
+                                        "id2",
+                                    ),
+                                    range: Some(
+                                        3..6,
+                                    ),
+                                },
                                 range: Some(
-                                    3..6,
+                                    3..7,
                                 ),
                             },
-                            range: Some(
-                                3..7,
-                            ),
-                        },
+                        ),
                         PathSelection {
                             path: WithRange {
                                 node: Var(
--- a/apollo-federation/src/sources/connect/expand/visitors/selection.rs
+++ b/apollo-federation/src/sources/connect/expand/visitors/selection.rs
@@ -39,85 +39,93 @@ impl FieldVisitor<NamedSelection> for SchemaVisitor<'_, ObjectTypeDefinitionPosi
         ))?;
 
         // Get the type of the field so we know how to visit it
-        let field_name = Name::new(field.name())?;
-        let field = definition
-            .field(field_name.clone())
-            .get(self.original_schema.schema())?;
-        let field_type = self
-            .original_schema
-            .get_type(field.ty.inner_named_type().clone())?;
-        let extended_field_type = field_type.get(self.original_schema.schema())?;
+        for field_name in field.names() {
+            let field_name = Name::new(field_name)?;
+            let field = definition
+                .field(field_name.clone())
+                .get(self.original_schema.schema())?;
+            let field_type = self
+                .original_schema
+                .get_type(field.ty.inner_named_type().clone())?;
+            let extended_field_type = field_type.get(self.original_schema.schema())?;
 
-        // We only need to care about the type of the field if it isn't built-in
-        if !extended_field_type.is_built_in() {
-            match field_type {
-                TypeDefinitionPosition::Scalar(scalar) => {
-                    let def = scalar.get(self.original_schema.schema())?;
-                    let def = ScalarType {
-                        description: def.description.clone(),
-                        name: def.name.clone(),
-                        directives: filter_directives(self.directive_deny_list, &def.directives),
-                    };
+            // We only need to care about the type of the field if it isn't built-in
+            if !extended_field_type.is_built_in() {
+                match field_type {
+                    TypeDefinitionPosition::Scalar(scalar) => {
+                        let def = scalar.get(self.original_schema.schema())?;
+                        let def = ScalarType {
+                            description: def.description.clone(),
+                            name: def.name.clone(),
+                            directives: filter_directives(
+                                self.directive_deny_list,
+                                &def.directives,
+                            ),
+                        };
 
-                    try_pre_insert!(self.to_schema, scalar)?;
-                    try_insert!(self.to_schema, scalar, Node::new(def))?;
-                }
-                TypeDefinitionPosition::Enum(r#enum) => {
-                    let def = r#enum.get(self.original_schema.schema())?;
-                    let def = EnumType {
-                        description: def.description.clone(),
-                        name: def.name.clone(),
-                        directives: filter_directives(self.directive_deny_list, &def.directives),
-                        values: def.values.clone(),
-                    };
+                        try_pre_insert!(self.to_schema, scalar)?;
+                        try_insert!(self.to_schema, scalar, Node::new(def))?;
+                    }
+                    TypeDefinitionPosition::Enum(r#enum) => {
+                        let def = r#enum.get(self.original_schema.schema())?;
+                        let def = EnumType {
+                            description: def.description.clone(),
+                            name: def.name.clone(),
+                            directives: filter_directives(
+                                self.directive_deny_list,
+                                &def.directives,
+                            ),
+                            values: def.values.clone(),
+                        };
 
-                    try_pre_insert!(self.to_schema, r#enum)?;
-                    try_insert!(self.to_schema, r#enum, Node::new(def))?;
-                }
+                        try_pre_insert!(self.to_schema, r#enum)?;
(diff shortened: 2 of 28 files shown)


apollo-federation/src/sources/connect/json_selection/README.md
```diff
@@ -81,15 +81,17 @@ worry if it doesn't seem helpful yet, as every rule will be explained in detail
...
 NamedFieldSelection  ::= Alias? Key SubSelection?
 NamedGroupSelection  ::= Alias SubSelection
 Alias                ::= Key ":"
-PathSelection        ::= (VarPath | KeyPath | AtPath | ExprPath) SubSelection?
+Path                 ::= VarPath | KeyPath | AtPath | ExprPath
+PathSelection        ::= Path SubSelection?
+PathWithSubSelection ::= Path SubSelection
 VarPath              ::= "$" (NO_SPACE Identifier)? PathStep*
 KeyPath              ::= Key PathStep+
 AtPath               ::= "@" PathStep*
```
benjamn, line 97: A good place to start reviewing this PR is the `README.md`, especially these changes to the formal grammar.

apollo-federation/src/sources/connect/json_selection/parser.rs
```diff
@@ -712,36 +755,6 @@ impl SubSelection {
...
-    pub fn get_at_index_mut(&mut self, index: &NamedSelectionIndex) -> &mut NamedSelection {
-        self.selections
-            .get_mut(index.pos)
-            .expect("NamedSelectionIndex out of bounds")
-    }
-}
-
-pub struct NamedSelectionIndex {
-    // Intentionally private so NamedSelectionIndex cannot be forged.
-    pos: usize,
```
benjamn, line 744: I removed this code because it's currently unused, and I didn't see a good way to support selections with multiple names under the single-`name` assumption of `SubSelection::index_of_named_selection`.

apollo-federation/src/sources/connect/json_selection/parser.rs
```diff
@@ -191,28 +198,51 @@ impl NamedSelection {
...
-        tuple((Alias::parse, PathSelection::parse))(input)
-            .map(|(input, (alias, path))| (input, Self::Path(alias, path)))
+        let (remainder, (alias_opt, path)) =
+            tuple((opt(Alias::parse), PathSelection::parse))(input)?;
+
+        if alias_opt.is_none() && !path.has_subselection() {
+            Err(nom::Err::Error(nom::error::Error::new(
+                input,
+                nom::error::ErrorKind::IsNot,
+            )))
```
benjamn, line 210: This is where we enforce that a `PathSelection` used alongside other `NamedSelection` items has either an `Alias` or a `SubSelection`. This error leaves a lot to be desired, but I have a separate plan for improving parsing error messages (coming soon, assuming it works).

lennyburdette (approved):
looks great!

apollo-federation/src/sources/connect/json_selection/parser.rs
lennyburdette: do we have a ticket tracking your error plans?

apollo-federation/src/sources/connect/expand/tests/snapshots/it_expand_supergraph@keys.graphql-2.snap
```diff
@@ -501,19 +501,21 @@ input_file: apollo-federation/src/sources/connect/expand/tests/schemas/expand/ke
...
+                                    node: Field(
+                                        "id2",
+                                    ),
+                                    range: Some(
+                                        3..6,
+                                    ),
+                                },
                                 range: Some(
-                                    3..6,
+                                    3..7,
```
dylan-apollo, line 515: Why did the location of these change?

apollo-federation/src/sources/connect/expand/visitors/selection.rs
```diff
@@ -134,21 +142,26 @@ impl GroupVisitor<JSONSelectionGroup, NamedSelection>
...
-
-        let TypeDefinitionPosition::Object(field_type) =
-            self.original_schema.get_type(field_type_name.clone())?
-        else {
-            return Ok(None);
-        };
 
-        Ok(field.next_subselection().cloned().map(|s| (field_type, s)))
+        match field.names().as_slice() {
+            &[field_name] => {
```
dylan-apollo, line 147: Could maybe be `field.names().first()`?

apollo-federation/src/sources/connect/json_selection/apply_to.rs
```diff
@@ -1662,6 +1675,247 @@ mod tests {
         );
     }
 
+    #[test]
+    fn test_inline_paths_with_subselections() {
```
lennyburdette, line 1679: worth testing this with variables? i've had use cases something like 
```
body: """
id: $this.id
$args.input {
  title
  body
}
"""

File: apollo-federation/src/sources/connect/json_selection/apply_to.rs [hunk 7]
```diff
@@ -256,12 +256,25 @@ impl ApplyToInternal for NamedSelection {
...
+                    // Handle the NamedPathSelection case.
+                    if let Some(value) = value_opt {
+                        output.insert(alias.name(), value);
+                    }
+                } else if let Some(JSON::Object(value)) = value_opt {
+                    // Handle the PathWithSubSelection case.
+                    output.extend(value);
+                } else {
+                    errors.push(ApplyToError::new(
+                        "Unnamed path selection did not return an object".to_string(),
```
dylan-apollo, line 274: I wonder if this error message is too technical for users—they're unlikely to know what a "path selection" is. Maybe something more generic like "Expected an object, got a <whatever type>" would raise fewer questions?

I believe composition will prevent them from writing an unnamed path selection which doesn't have a subselection, so really the problem is that the JSON didn't match the expected type `{ }`, right?

I guess it should match whatever the error is when we attempt to apply a subselection and the input isn't an object.

dylan-apollo (approved):
LGTM, left a couple of comments, but no blockers

apollo-federation/src/sources/connect/expand/tests/snapshots/it_expand_supergraph@keys.graphql-2.snap
benjamn: Good question! My heart skipped a beat when I saw this, but it's a harmless diff artifact.

In this example, the output before looked like
```
                    Path(
                        Alias {
                            name: WithRange {
                                node: Field(
                                    "id2",
                                ),
                                range: Some(
                                    3..6,
                                ),
                            },
                            range: Some(
                                3..7,
                            ),
                        },
                        PathSelection {
                          # omitting this for brevity
                        },
                    ),
```

Since we changed the `NamedSelection::Path` enum to have an `Option<Alias>` as its first argument, there's a new layer of wrapping, so the new output looks like this:
```
                    Path(
                        Some(
                            Alias {
                                name: WithRange {
                                    node: Field(
                                        "id2",
                                    ),
                                    range: Some(
                                        3..6,
                                    ),
                                },
                                range: Some(
                                    3..7,
                                ),
                            },
                        ),
                        PathSelection {
                          # omitting this for brevity
                        },
                    ),
```

All that's really happening here is the `Alias` from before is getting wrapped with `Some(...)`, and the ranges are staying the same.

If you drop these two snippets into different files and diff them, however, the diff logic somehow decides the inner range from the first corresponds to the outer range from the second, so the `3..6` appears to be changing in place (to `3..7`):
```diff
diff --git a/before.snap b/after.snap
index 6d12da5fe..9142140d8 100644
--- a/before.snap
+++ b/after.snap
@@ -1,17 +1,19 @@
                     Path(
-                        Alias {
-                            name: WithRange {
-                                node: Field(
-                                    "id2",
-                                ),
+                        Some(
+                            Alias {
+                                name: WithRange {
+                                    node: Field(
+                                        "id2",
+                                    ),
+                                    range: Some(
+                                        3..6,
+                                    ),
+                                },
                                 range: Some(
-                                    3..6,
+                                    3..7,
                                 ),
                             },
-                            range: Some(
-                                3..7,
-                            ),
-                        },
+                        ),
                         PathSelection {
                           # omitting this for brevity
                         },
```

If you run the diff with `-w` to ignore whitespace, the amount of change more closely matches what's logically happening:
```diff
diff --git a/before.snap b/after.snap
index 6d12da5fe..9142140d8 100644
--- a/before.snap
+++ b/after.snap
@@ -1,18 +1,20 @@
                     Path(
+                        Some(
                             Alias {
                                 name: WithRange {
                                     node: Field(
                                         "id2",
                                     ),
                                     range: Some(
                                         3..6,
                                     ),
                                 },
                                 range: Some(
                                     3..7,
                                 ),
                             },
+                        ),
                         PathSelection {
                           # omitting this for brevity
                         },
                     ),
```

pubmodmatt (approved):
Looks great! Likely requires similar changes in that other draft I started, but should be straightforward.

apollo-federation/src/sources/connect/json_selection/apply_to.rs
benjamn: This was interesting! https://github.com/apollographql/router/pull/6076/commits/dede80cb4ee87ca67f2cfa7250cbbc4078d5cd7c?diff=unified&w=0

apollo-federation/src/sources/connect/expand/visitors/selection.rs
benjamn: https://github.com/apollographql/router/pull/6076/commits/560d27d21fd303e4b00db563539e54a171f92ee4

apollo-federation/src/sources/connect/json_selection/apply_to.rs
benjamn: I think these errors make sense, but there's usually more than one error reported now, so have a look and let me know what you think: https://github.com/apollographql/router/pull/6076/commits/99e8dc9eaeaf14cd744aa4f9a7e0d289e5fad581

apollo-federation/src/sources/connect/json_selection/apply_to.rs
benjamn: As an example of something I learned while writing these tests, you can flatten more than one level of nesting using this syntax:
```rust
            assert_eq!(
                selection!(
                    r#"
                    id: $this.id
                    $args { .input { title body } extra }
                    from
                "#
                )
                .apply_with_vars(&data, &vars),
                (
                    Some(json!({
                        "id": 1234,
                        "title": "The capital of Australia",
                        "body": "Canberra",
                        "extra": "extra",
                        "from": "data",
                    })),
                    vec![],
                ),
            );
```

PR #6079, opened 2024-09-30
Replace askama by manual string replace

goto-bus-stop:
We aren't exactly doing rocket science here

askama takes about 3-4 seconds to compile on my machine.

Closes #5878

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [x] Manual Tests

**Exceptions**

We do not appear to have tests for the static pages. I could add some but I think the best option is snapshotting, which is only to prevent accidental changes here 🤷🏻‍♀️ 

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -237,7 +237,6 @@ dependencies = [
  "apollo-compiler",
  "apollo-federation",
  "arc-swap",
- "askama",
  "async-channel 1.9.0",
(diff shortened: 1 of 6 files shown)


garypen (approved):
Works for me.

Geal (approved):

PR #6084, opened 2024-09-30
fix: make test supergraphs valid

lennyburdette:
this becomes a problem when switching to the native QP by default

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


goto-bus-stop:
I think it's a little suspicious that Rust doesn't accept those. Could there be customers with such "incomplete" supergraphs? will discuss with the QP team if we shouldn't change the native QP instead

lennyburdette:
the contract we have with customers is that they should use only federation composition to generate supergraphs — i don't think we *need* to relax our supergraph validation. whether we *should*, maybe to adhere to the robustness principle, might be a discussion worth having, but i'm not sure.

goto-bus-stop (approved):

PR #6086, opened 2024-09-30
[JSONSelection] Forbid single-`.key` paths, requiring `$.key` instead

benjamn:
Now that PR #6076 allows `PathWithSubSelection` syntax to appear within a `SubSelection` alongside other named selections, the shorthand of using `.field` to refer to a path consisting of a single `field` key has become potentially ambiguous:

```graphql
__typename: $("Product")
.field { nested child }
```

Because parsing is insensitive to spaces/newlines/comments between tokens, this code parses the same as

```graphql
__typename: $("Product").field { nested child }
```

which produces a runtime `ApplyToError` because the `"Product"` string does not have a `field` property.

Fortunately, using the `$` variable (i.e. `$.field`, introduced in #5142) has identical behavior to the shorthand, and does not suffer from any parsing ambiguity:

```graphql
__typename: $("Product")
$.field { nested child }
```

As a reminder, thanks to #6076, this selection produces an output object with keys `__typename`, `nested`, and `child`, all at the same level.

This commit statically forbids using the `.field` shorthand, so `$.field` is effectively required, which is a breaking change, but it seems better to implement this restriction while we're still in preview than after we've publicly released this code.

As another reminder, the leading `.` has been optional for some time when the path consists of more than one key, so if you were previously using `.some.nested.field` or `.object->method(...)`, you have the option of simply removing the leading `.` (i.e. `some.nested.field` and `object->method(...)`) rather than prepending the `$` variable, though either option will produce the same behavior.

Subjectively, I find `$.field` more explicit and harder to misread than `.field`, so I believe this restriction is an improvement.

### DIFF
--- a/apollo-federation/src/sources/connect/json_selection/apply_to.rs
+++ b/apollo-federation/src/sources/connect/json_selection/apply_to.rs
@@ -662,10 +662,10 @@ mod tests {
             }),
         );
 
-        check_ok(selection!(".nested.hello"), json!("world"));
+        check_ok(selection!("nested.hello"), json!("world"));
         check_ok(selection!("$.nested.hello"), json!("world"));
 
-        check_ok(selection!(".nested.world"), json!("hello"));
+        check_ok(selection!("nested.world"), json!("hello"));
         check_ok(selection!("$.nested.world"), json!("hello"));
 
         check_ok(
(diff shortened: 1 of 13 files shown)


benjamn:
Oops, need to fix tests over in the `apollo-router` crate… stay tuned!

lennyburdette (approved):

PR #6089, opened 2024-09-30
[JSONSelection] Report better parsing error messages using `extra: X` field of `LocatedSpan<T, X = ()>`

benjamn:
The `LocatedSpan<T, X = ()>` type provided by the [`nom_locate`](https://docs.rs/nom_locate/latest/nom_locate/) crate (first introduced in PR #5987) supports an optional user-defined `extra: X` field, which we can use to smuggle meaningful custom error messages out of the parser.

It would be ideal if we could report multiple parsing errors and continue best-effort parsing after an error occurs, but that will have to wait for future PRs. Also, there are some cases where parsing fails because various `alt(...)` combinators ran out of options, and we can't (yet) determine which underlying error was "most" responsible for the failure, so the error message remains an opaque `nom::error::ErrorKind` code.

As a side benefit, I've attempted to stabilize the method signature of the public `JSONSelection::parse` method, so consumers don't depend on `nom`- and `nom_locate`-specific types, so we have the freedom to swap out implementation details in the future without breaking consumers. This primarily means continuing to accept a `&str` as input (rather than a `LocatedSpan`) and returning a custom `JSONSelectionParseError` type in case of failure.

While improving parsing error messages is still a work in progress, I think this PR meaningfully improves the situation, and is worth considering as a first step.


apollo-federation/src/sources/connect/json_selection/snapshots/apollo_federation__sources__connect__json_selection__parser__tests__error_snapshots.snap
```diff
@@ -0,0 +1,13 @@
...
+---
+Err(
+    Failure(
+        JSONSelectionParseError {
+            message: "Key paths may no longer start with just .key (use $.key instead)",
+            fragment: ".data",
+            offset: 0,
+        },
+    ),
+)
```
benjamn, line 13: This error message should help explain/mitigate the breaking change introduced by #6086.

lennyburdette: "no longer" isn't going to make much sense to folks not involved with the preview (so: everyone 😁 )

benjamn: Fair enough: https://github.com/apollographql/router/pull/6089/commits/39da23563041cd2c857dda808aaae8c9d733f7d2

apollo-federation/src/sources/connect/json_selection/parser.rs
```diff
@@ -58,9 +134,43 @@ impl JSONSelection {
...
+    // if we drastically change implementation details. That's why we use &str
+    // as the input type and a custom JSONSelectionParseError type as the error
+    // type, rather than using Span or nom::error::Error directly.
+    pub fn parse(input: &str) -> IResult<&str, Self, JSONSelectionParseError> {
+        match JSONSelection::parse_span(new_span(input)) {
+            Ok((remainder, selection)) => {
+                // To avoid exposing the implementation details of nom_locate's
+                // LocatedSpan, we report the remainder as a &str instead.
+                Ok((remainder.fragment(), selection))
+            }
```
dylan-apollo, line 148: Can we make this `Result<Self, JSONSelection>`, using std result, and make remainder an error? Right now all the consumers have some half-baked version of returning an error anyway when there's a remainder, and if we use std Result, we can do `impl From <JSONSelectionParseError> for Message` (and every other error type) and use `?`. So something like this:

```suggestion
    pub fn parse(input: &str) -> Result<Self, JSONSelectionParseError> {
        match JSONSelection::parse_span(new_span(input)) {
            Ok((remainder, selection)) => {
                let fragment = remainder.fragment();
                if fragment.is_empty() {
                    Ok(selection)
                } else {
                    Err(JSONSelectionParseError {
                        message: "Unexpected trailing characters".to_string(),
                        fragment: fragment.to_string(),
                        offset: remainder.location_offset(),
                    })
                }
            }
```

This also keeps noms `Err` out of our public API, so we can more easily directly use `JSONSelectionParserError`.

dylan-apollo (commented):
I finally tried this out and it works great! One suggestion on the API that will make using these improved errors a bit easier elsewhere (but requires updating consumers _now_ since the API changes). I'm happy to make a PR into this which updates validation & the like if that'd be helpful.

apollo-federation/src/sources/connect/json_selection/parser.rs
benjamn: Good idea! https://github.com/apollographql/router/pull/6089/commits/76949bd53a152f620bebfcef13c76f9e52e44dfe

benjamn: I think I updated all the consumers (including both `apollo-federation` and `apollo-router`), unless there are any outside of this repository?

apollo-federation/src/sources/connect/json_selection/parser.rs
```diff
@@ -21,12 +22,53 @@ use super::helpers::spaces_or_comments;
...
 use super::location::WithRange;
 
+// ParseResult is the internal type returned by most ::parse methods, as it is
+// convenient to use with nom's combinators. The top-level JSONSelection::parse
+// method returns a slightly different IResult type that hides implementation
+// details of the nom-specific types.
+//
+// TODO Consider switching the third IResult type parameter to VerboseError
+// here, if error messages can be improved with additional context.
+pub(super) type ParseResult<'a, T> = IResult<Span<'a>, T>;
```
dylan-apollo, line 39: In the future (not suggesting you need to go and change the ones in this PR), if you use triple-slash quotes right before an item, it sets that comment as the docs for the item. Docs also support Markdown syntax, so you can get nicely-rendered docs for things in your editor. Here's in RustRover, where that link to `JSONSelection::parse` actually brings you to the right code!

Screenshot 2024-11-01 at 10 59 47 AM
Screenshot 2024-11-01 at 11 00 41 AM

dylan-apollo (approved):
Looks great, thanks! I'll get it wired up to show the offsets in the LSP in a follow-up.

PR #6090, opened 2024-09-30
fix(dual-query-planner): fixed the remaining of coercion mismatches between Rust QP and JS QP

duckki:
Summary:
- updated the coercion check to recurse into List and Object values.
- loosened Object value check to be a inclusion check, instead of empty check

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [x] Manual Tests

[ROUTER-753]: https://apollographql.atlassian.net/browse/ROUTER-753?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apollo-router/src/query_planner/dual_query_planner.rs
+++ b/apollo-router/src/query_planner/dual_query_planner.rs
@@ -734,22 +734,24 @@ fn ast_value_maybe_coerced_to(x: &ast::Value, y: &ast::Value) -> bool {
             }
         }
 
-        // Special case 2: Rust QP expands a list/object value by filling in its
-        // default item/field values.
-        // - If the Rust QP value subsumes the JS QP value, consider it a match.
-        // - Assuming the Rust QP object value has only default item/field values.
+        // Special case 2: Rust QP expands a object value by filling in its
+        // default field values.
+        // - If the Rust QP object value subsumes the JS QP object value, consider it a match.
+        // - Assuming the Rust QP object value has only default field values.
         // - Warning: This is an unsound heuristic.
-        (ast::Value::List(ref x), ast::Value::List(ref y)) => {
-            if vec_includes_as_set(y, x, |yy, xx| {
-                xx == yy || ast_value_maybe_coerced_to(xx, yy)
-            }) {
-                return true;
-            }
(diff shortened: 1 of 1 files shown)


apollo-router/src/query_planner/dual_query_planner.rs
```diff
@@ -879,6 +900,48 @@ mod ast_comparison_tests {
...
+        let ast_x = ast::Document::parse(op_x, "op_x").unwrap();
+        let ast_y = ast::Document::parse(op_y, "op_y").unwrap();
+        assert!(super::same_ast_document(&ast_x, &ast_y).is_ok());
+    }
+
+    #[test]
+    fn test_query_variable_decl_list_value_coercion_empty_case() {
+        // Note: Rust QP expands empty list default values by filling in its default items.
+        let op_x = r#"query($qv1: [Int!]! = []) { x(arg1: $qv1) }"#;
+        let op_y = r#"query($qv1: [Int!]! = [1, 42]) { x(arg1: $qv1) }"#;
```
goto-bus-stop, line 918: This seems strange? these are not equivalent queries.

duckki: `y` is expected to have additional items over `x` by filling in the (assumed-to-be) default value.

goto-bus-stop: I'm confused about that. The default value is declared to be `[]` in one and `[1, 42]` in the other. This isn't a list that would pull in default values--it's a declaration of a default value.

goto-bus-stop: For objects, default property values come from the input object definition, but that doesn't exist for lists

duckki: Ah, you are right. We don't have to loosen lists.

goto-bus-stop (approved):
thanks, this all looks right to me 🥳

PR #6092, opened 2024-10-01
Add datadog span processor.

BrynCooke:
This PR makes the datadog agent sampling feature use Record rather than RecordAndSample for spans that would otherwise have been dropped. 
This allows us to stay closer to the original telemetry logic for setting the sampled trace flag.

This complexity arrises because BatchSPanProcessors usually drop spans that are not sampled.
To get around this we introduce a wrapper that is active for otlp and datadog exporters if agent sampling is enabled that updates the spandata to say that it was sampled.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/tests/integration/telemetry/otlp.rs
```diff
@@ -239,7 +239,7 @@ async fn test_otlp_request_with_zipkin_trace_context_propagator_with_datadog(
         .await?;
     // ----------------------
     // Testing for a trace containing a tracestate m and psr with psr set to 1 for DD and an unsampled trace, so it should be sent to the otlp exporter with sampling priority set to 1
-    // But it should send the trace to subgraph as the tracestate contains a psr set to 1
+    // It should not send the trace to the subgraph as we didn't use the datadog propagator and therefore the trace will remain unsampled.
```
BrynCooke, line 242: @bnjjj This is a key test change.

apollo-router/src/plugins/telemetry/tracing/datadog_agent_span_processor.rs
```diff
@@ -0,0 +1,131 @@
...
+    }
+}
+
+impl<T: SpanProcessor> SpanProcessor for DatadogBatchSpanProcessor<T> {
+    fn on_start(&self, span: &mut Span, cx: &Context) {
+        self.delegate.on_start(span, cx)
+    }
+
+    fn on_end(&self, mut span: SpanData) {
+        let new_trace_flags = span.span_context.trace_flags().with_sampled(true);
```
bnjjj, line 28: Should we check for either sampling priority to be set or measured to be set to set it to true ? I know for now we don't have that case but should we already try to handle this ?

apollo-router/src/plugins/telemetry/tracing/datadog_agent_sampler.rs
```diff
@@ -0,0 +1,265 @@
...
+            span_kind,
+            attributes,
+            links,
+        );
+        // Override the sampling decision to record and make sure that the trace state is set correctly
+        // if either parent sampling is disabled or it has not been populated by a propagator.
+        // The propagator gets first dibs on setting the trace state, so if it sets it, we don't override it unless we are not parent based.
+        match result.decision {
+            SamplingDecision::Drop | SamplingDecision::RecordOnly => {
+                result.decision = SamplingDecision::RecordOnly;
```
BrynCooke, line 71: @bnjjj This is the only change to the logic

apollo-router/tests/integration/telemetry/otlp.rs
bnjjj: Are we sure we should not take care of this tracestate ? Do we have any specs or docs to help us to choose ?

bnjjj: https://github.com/DataDog/dd-trace-js/blob/fcd3ab8244800e64b01bdd746933812dada90ad5/packages/dd-trace/test/opentelemetry/span_context.spec.js#L58

apollo-router/src/plugins/telemetry/tracing/datadog_agent_span_processor.rs
BrynCooke: I had a long think about this. I think it's OK for the sole purpose is to change the sampled flag for exporting. I've refactored and renamed to reflect this: [a5efd62](https://github.com/apollographql/router/pull/6092/commits/a5efd6233589c7a2b8f23d25cd3661dc8ac2ac6a)

BrynCooke:
> Are we sure we should not take care of this tracestate ? Do we have any specs or docs to help us to choose ?

I think the way to think about this is that without datadog propagation trace state doesn't mean anything. They are just key, values. But let's discuss tomorrow.

PR #6097, opened 2024-10-01
[JSONSelection] Fix validations to accommodate inline `NamedSelection::Path` items with multiple output keys

benjamn:
PR #6076 introduced a kind of `NamedSelection::Path` that can contribute multiple keys to the current output object, rather than just a single key.

Because the `SubSelection::selections_iter` method was left unchanged, these new multi-name `NamedSelection`s began showing up as single items in the iterator consumed by [`GroupVisitor::enter_group`](https://github.com/apollographql/router/blob/21f7c7ae76a50f47f8a94b5aae17562599242649/apollo-federation/src/sources/connect/validation/selection.rs#L287), which interfered with the assumptions of that code, causing it to misjudge whether the nested properties should have object values or not.

This PR adjusts `SubSelection::selections_iter` to continue emitting only `NamedSelection` items that have a singular output key, which involves recursively emitting the individual keys produced by the new `NamedSelection::Path` elements, without emitting the `NamedSelection::Path` directly. Other kinds of `NamedSelection` are emitted as before.

### DIFF
--- a/apollo-federation/src/sources/connect/expand/visitors/mod.rs
+++ b/apollo-federation/src/sources/connect/expand/visitors/mod.rs
@@ -373,6 +373,37 @@ mod tests {
     }
 
     #[test]
+    fn it_iterates_over_paths() {
+        let mut visited = Vec::new();
+        let visitor = TestVisitor::new(&mut visited);
+        let (unmatched, selection) = JSONSelection::parse(
+            "a
+            $.b {
+                c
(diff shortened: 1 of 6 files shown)


dylan-apollo (approved):
Looks great, can you just replace the `unreachable!` with a `debug_assert!`?

lennyburdette (approved):

PR #6099, opened 2024-10-01
chore(federation): clarify TODO comment that we will keep for query planning GA

goto-bus-stop:
We discussed what to do with this in the triage call,
recording the outcome of the discussion so we don't have to talk about it again later :)


lrlna (approved):
tyyy

PR #6113, opened 2024-10-03
chore: Retire Projects Tagger

abernix:
We no longer necessitate triggering an action that adds issues to Project boards, and we're sunsetting the endpoint that we were previously calling.


goto-bus-stop (approved):

abernix:
Thanks!

PR #6116, opened 2024-10-03
Corrected overwritten fields in `remove_conditions_from_selection_set`

TylerBloom:
In SelectionSet::remove_conditions_from_selection_set, a `SelectionMap` was being constructed directly from an existing selection set. Each selection had its conditions removed and inserted into the new map; however, it is possible that removing the directives from the selection would now cause a key collision. An example of this can be seen in the tests that were added:
```graphql
query Test($b: Boolean!) {
  foo @include(if: $b) {
    bar {
      name
    }
    bar @include(if: $b) {
      thing
    }
  }
}
```
When the selection set was finalized, the selection `bar { name }` was getting overwritten because, after the next selection had its `@include` directive removed, the two selections' keys collided. In the JS code, they used `SelectionSet::lazyMap`, so now the Rust code does too.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [x] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-771]: https://apollographql.atlassian.net/browse/ROUTER-771?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apollo-federation/src/query_plan/conditions.rs
+++ b/apollo-federation/src/query_plan/conditions.rs
@@ -10,8 +10,10 @@ use serde::Serialize;
 
 use crate::error::FederationError;
 use crate::operation::DirectiveList;
+use crate::operation::NamedFragments;
 use crate::operation::Selection;
 use crate::operation::SelectionMap;
+use crate::operation::SelectionMapperReturn;
 use crate::operation::SelectionSet;
 use crate::query_graph::graph_path::OpPathElement;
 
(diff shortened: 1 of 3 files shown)


dariuszkuc (approved):
`lazy_map` change seems fine but it looks like the issue could have been solved by simply changing `SelectionMap::insert` logic to a merge selection instead  (as it previously assumed that there would be collisions)

goto-bus-stop (approved):

PR #6117, opened 2024-10-03
Add dotenv to allow router to read environment variables from .env file

andrewmcgivery:
Add dotenv to allow router to read environment variables from .env file

Possibly worth discussing:

- This is currently implemented to be "on by default". Reason I didn't put this behind a config or option flag is because at that point, it is too late since the options parse environment values. Possible workaround is to do a more "manual" read of the argument separate from the normal parsing code specifically for this flag.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


goto-bus-stop (commented):
If we do this we should use the `dotenvy` crate, as `dotenv` is unmaintained.

andrewmcgivery:
> If we do this we should use the `dotenvy` crate, as `dotenv` is unmaintained.

Good catch. Switched to that package!

bnjjj (commented):
The overall PR LGTM, I would really have this feature in the router because I already struggled with that. I'm just wondering what could be the impact on our existing users ? Could it break something or have weird behavior ? Like for example what if they already had that .env file with wrong values but they're still providing environment variables like they always did before with the right values ? What's the priority order in that case ? Also does it keep the same logics if they use the argument in command line (like `--config` for example) but they're also providing the `APOLLO_ROUTER_CONFIG_PATH` ? And what's the behavior now if additionally we provide the same env var in .env file ? I think we need to test it manually (at least) and document it somewhere

abernix:
Also plenty of folks use `direnv` or _many other_ tools (including Docker itself) that handles this in their local environment and additionally automatically uses different `.env` files for different environments.

I'm really not sure that the Router should really take on this concern when there are off the shelf tools that handle this generically.  I'd like to know why those other tools can't just be used without introducing this to the Router directly, before we entertain a PR that adds it.

PR #6118, opened 2024-10-04
fix(federation): updated PathTree::from_path to use the last trigger value

duckki:
- When paths have equivalent (but with minor difference) trigger values, the last path's value is used in the returned tree.
- This change is to match JS QP implementation's behavior.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [x] Integration Tests
    - [x] Manual Tests

[ROUTER-772]: https://apollographql.atlassian.net/browse/ROUTER-772?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apollo-federation/src/query_graph/path_tree.rs
+++ b/apollo-federation/src/query_graph/path_tree.rs
@@ -226,11 +226,20 @@ where
 
         struct ByUniqueEdge<'inputs, TTrigger, GraphPathIter> {
             target_node: NodeIndex,
-            by_unique_trigger:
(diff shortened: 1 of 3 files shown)


dariuszkuc (approved):
approved so we can more easily compare the mismatches.... but I think we probably should remove this for GA as this seems pretty unnecessary

PR #6122, opened 2024-10-04
Move computation-heavy tasks to a dedicated thread pool

SimonSapin:
These components can take non-trivial amounts of CPU time:
* GraphQL parsing
* GraphQL validation
* Query planning
* Schema introspection

In order to avoid blocking threads that execute asynchronous code, they are now run (in their respective Rust implementations) in a new pool of as many threads as CPU cores are available. Previously we used Tokio’s [`spawn_blocking`] for this purpose, but it is appears to be intended for blocking I/O and uses up to 512 threads so it isn’t a great fit for computation tasks.

[`spawn_blocking`]: https://docs.rs/tokio/latest/tokio/task/fn.spawn_blocking.html

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


SimonSapin:
Still using `spawn_blocking`:

* Things apparently related to shutdown in `apollo-router/src/executable.rs`, `apollo-router/src/metrics/mod.rs`, and `apollo-router/src/plugins/telemetry/mod.rs`
* `tokio::sync::broadcast::Sender::send` in `apollo-router/src/plugins/traffic_shaping/deduplication.rs`: is this one really blocking?

apollo-router/src/compute_task.rs
```diff
@@ -0,0 +1,48 @@
...
+/// let output = compute_task::execute(|| some_heavy_computation()).await;
+/// ```
+pub(crate) fn execute<T, F>(f: F) -> oneshot::Receiver<T>
+where
+    F: FnOnce() -> T + Send + 'static,
+    T: Send + 'static,
+{
+    // https://docs.rs/threadpool/latest/threadpool/struct.Builder.html#method.num_threads
+    // > defaults the number of threads to the number of CPUs
+    let pool = POOL.get_or_init(|| threadpool::Builder::new().build());
```
dylan-apollo, line 41: We don't need to / want to reserve some CPUs for Tokio tasks? Obviously we have no control over how threads get scheduled, I just wonder if we should pull back on the default number of threads since we know we have another task scheduler.

duckki: Why don't we use tokio's `spawn_blocking` method? It may work better together with I/O tasks.

SimonSapin: Maybe! If we have more threads that want CPU time than available CPU cores things don’t break down immediately, the kernel scheduler preempts threads and tries to distribute time fairly. So at 1.5x or even 2x utilization (over compute threads + tokio threads) we may have some context switching overhead but it shouldn’t be as bad as with 512 threads. We can tweak the number but it’s hard to guess what a good value is: 50% of `num_cpus()`? 80%? We could make it configurable but most users may not know any more about what a good value for this obscure number is.

apollo-router/src/compute_task.rs
```diff
@@ -0,0 +1,48 @@
+use std::sync::OnceLock;
+
+use tokio::sync::oneshot;
+
+static POOL: OnceLock<threadpool::ThreadPool> = OnceLock::new();
+
+/// Drop-in replacement for Tokio’s [`spawn_blocking`]
+/// but intended for for tasks that keep a CPU core active: it uses a thread pool
```
garypen, line 8:
```suggestion
/// but intended for tasks that keep a CPU core active: it uses a thread pool
```

apollo-router/src/compute_task.rs
garypen: Let's merge this as is because it's already an improvement on `spawn_blocking()` and we can think about the best way to slice up CPU resource once this change is in place.

PR #6123, opened 2024-10-04
Refactored `SelectionSet::split_top_level_fields`

TylerBloom:
This PR refactors `SelectionSet::split_top_level_fields` away from using `Box<dyn Iterator>`s. While not a performance issue (the function is called from exactly one place), it was a very smelly implementation, which required every item to be to be allocated in its own allocation before being yielded by the iterator.

Since the iterator is recursive, I had to just implement `Iterator` on an internal type rather than use the opaque type (which can't be recursive at all) returned by the function.

Lastly, I saw a couple of other `Box<dyn Iterator>`s that that could be turned into RPITs by using the `Either` type from the `either` crate. `either` is already a transitive dependency, so it adds no extra dependencies to the `apollo-federation` or `apollo-router`.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-491]: https://apollographql.atlassian.net/browse/ROUTER-491?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -182,6 +182,7 @@ version = "1.56.0"
 dependencies = [
  "apollo-compiler",
  "derive_more",
+ "either",
  "hex",
  "indexmap 2.2.6",
  "insta",
--- a/apollo-federation/Cargo.toml
+++ b/apollo-federation/Cargo.toml
@@ -36,6 +36,7 @@ thiserror = "1.0"
 url = "2"
 tracing = "0.1.40"
 ron = { version = "0.8.1", optional = true }
+either = "1.13.0"
 
 [dev-dependencies]
 hex.workspace = true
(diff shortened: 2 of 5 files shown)


goto-bus-stop (approved):
LGTM. Some of this implementation reminds me of [`SelectionSet::field_selections`](https://github.com/apollographql/router/blob/5a1249c50ee624aa4c1f3fe7146c72fcc518d16f/apollo-federation/src/operation/mod.rs#L2663-L2670) (different ownership aside, and field_selections also doesn't track the `parent_type`)

PR #6125, opened 2024-10-07
feat(telemetry): add ability to set span attributes on connector span

bnjjj:
WIP, needs tests

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/services/connector_service.rs
```diff
@@ -124,6 +124,9 @@ impl tower::Service<ConnectRequest> for ConnectorService {
                 "apollo.connector.source.detail" = tracing::field::Empty,
                 "apollo_private.sent_time_offset" = fetch_time_offset,
             );
+            // TODO: I think we should get rid of these attributes by default and only add it from custom telemetry. We just need to double check it's not required for Studio.
```
bnjjj, line 127: cc @timbotnik I need your knowledge here

apollo-router/src/plugins/telemetry/mod.rs
```diff
@@ -882,37 +884,72 @@ impl PluginPrivate for Telemetry {
                             .events
                             .new_connector_http_events();
                         custom_events.on_request(http_request);
+
+                        let custom_span_attributes = req_fn_config
+                            .instrumentation
+                            .spans
+                            .connector
+                            .attributes
+                            .on_request(http_request);
```
pubmodmatt, line 893: I'm not seeing how this would work when we add the next connector type. If there's a gRPC connector, for example, we'd need to call `.on_request` with a gRPC request type - but here, the `attributes` on the span is only for the HTTP request type. I think there needs to be `spans.connector.http.attributes` and `spans.connector.grpc.attributes` for this to work.

apollo-router/src/plugins/telemetry/config_new/connector/events.rs
```diff
@@ -10,6 +10,7 @@ use crate::plugins::telemetry::config_new::extendable::Extendable;
 #[derive(Clone, Deserialize, JsonSchema, Debug, Default)]
 #[serde(deny_unknown_fields, default)]
 pub(crate) struct ConnectorEventsKind {
+    // TODO: I don't think we need this, we should be consistent with what we have today and everything is namespaced like http.*
```
pubmodmatt, line 13: Mentioned this in #6067 , but the `StandardEventConfig` and `Condition` types are generic on the `Selector` type, and the `Selector` type has associated types for the request and response types. So in order to support future connectors with request and response types other than `HttpRequest` and `HttpResponse`, there needs to be another nesting level here.

bnjjj: Answered [here](https://github.com/apollographql/router/pull/6067#issuecomment-2400027367)

apollo-router/src/plugins/telemetry/mod.rs
bnjjj: I think right now we have issues because we have 
 `subgraph` and `connector` but our end goal will be to get rid of both subgraph and connector and only have `http` section. When we will have this in place then we will probably end up with `http`, `router`, `supergraph`, `grpc`, `sql`. That's how we see the future of telemetry with @BrynCooke. But I think we should definitely not reflect the configuration shape based on our implementation. If we plan to keep `subgraph` and `connector` there is still a solution which is using a custom deserializer for serde and under the hood we would have different types and so we will be able to implement different selector on this.

pubmodmatt (approved):
This looks really good! Just a couple minor comments. It enables some nice things:

```yaml
telemetry:
  instrumentation:
    spans:
      connector:
        attributes:
          connector.url.template: true
          connector.http.method: true
          "rate_limit":
            connector_http_response_header: "x-ratelimit-remaining"
```

apollo-router/src/plugins/telemetry/mod.rs
```diff
@@ -881,38 +883,69 @@ impl PluginPrivate for Telemetry {
...
-                                Ok(http_response) => {
-                                    custom_instruments.on_response(http_response);
-                                    custom_events.on_response(http_response);
-                                }
-                                Err(err) => {
-                                    custom_instruments.on_error(err, &context);
-                                    custom_events.on_error(err, &context);
+                        match custom_telemetry {
+                            Some((custom_instruments, custom_events, custom_span_attributes)) => {
+                                let span = Span::current();
```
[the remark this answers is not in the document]
Geal: so maybe we need another hook point in the future

Geal (approved):

apollo-router/src/plugins/telemetry/mod.rs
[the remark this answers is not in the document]
bnjjj: I think we're in a temporary situation here. The goal is either to provide another hook point or to move everything (including subgraph) in the http_client hook. Once we're there we won't have this issue anymore.

pubmodmatt (approved):

PR #6126, opened 2024-10-07
Add `apollo_router_deduplicated_queries_total` metric

andrewmcgivery:
Add `apollo_router_deduplicated_queries_total` metric so that a platform team can understand the impact of enabling query deduplication in traffic shaping.

Note: I couldn't find any examples of how to do a unit or integration test for this so I did a manual test. I setup a local router with a K6 script that hit a subgraph with an artificially slow response time. I was then able to see this metric appear in my Prometheus metrics endpoint:

Screenshot 2024-10-07 at 12 08 10 PM

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/plugins/traffic_shaping/deduplication.rs
+++ b/apollo-router/src/plugins/traffic_shaping/deduplication.rs
@@ -99,6 +99,9 @@ where
 
                     match receiver.recv().await {
                         Ok(value) => {
+                            tracing::info!(
+                                monotonic_counter.apollo_router_deduplicated_queries_total = 1u64,
+                            );
                             return value
                                 .map(|response| {
                                     SubgraphResponse::new_from_response(
@@ -107,7 +110,7 @@ where
                                         request.subgraph_name.unwrap_or_default(),
                                     )
                                 })
-                                .map_err(|e| e.into())
+                                .map_err(|e| e.into());
                         }
                         // there was an issue with the broadcast channel, retry fetching
                         Err(_) => continue,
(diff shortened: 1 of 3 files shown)


apollo-router/src/plugins/traffic_shaping/deduplication.rs
```diff
@@ -99,6 +99,9 @@ where
 
                     match receiver.recv().await {
                         Ok(value) => {
+                            tracing::info!(
+                                monotonic_counter.apollo_router_deduplicated_queries_total = 1u64,
```
bnjjj, line 103: Could you use the new macro instead ? Search for u64_counter in the codebase. And use the new conventions with dots. Example apollo.router...

bnjjj (changes requested):
The hardest part will be to test it now :p

apollo-router/src/plugins/traffic_shaping/deduplication.rs
andrewmcgivery: @bnjjj done!

bnjjj:
cc @BrynCooke I need you to have ideas about the name of this metric please

apollo-router/src/plugins/traffic_shaping/mod.rs
```diff
@@ -928,4 +930,54 @@ mod test {
...
+                        .await
+                        .unwrap()
+                        .call(request)
+                        .await
+                        .unwrap();
+                });
+            }
+            join_all(tasks).await;
+
+            assert_counter!("apollo.router.deduplicated.queries", 4);
```
BrynCooke, line 978: This should be `apollo.router.deduplicated_queries` as the dot indicates nesting.
I do wonder if this should just be an attribute on an existing metric though. For instance:

```

telemetry:

  instrumentation:
    spans:
      mode: spec_compliant
    instruments:
      subgraph:
        http.client.request.duration: 
          attributes: 
            deduplicated: true
```

BrynCooke (changes requested):

apollo-router/src/plugins/traffic_shaping/mod.rs
andrewmcgivery: Does making it an attribute not make it inconsistent with the existing subscriptions metric?

PR #6128, opened 2024-10-08
Move some GraphQL analysis earlier in the pipeline

SimonSapin:
The following is now done in `QueryAnalysisLayer` instead of in the `BridgeQueryPlanner` service:

* Finding the relevant operation in a GraphQL document based on `operationName`. Now `ParsedDocument` in context contains a reference to the parsed `apollo_compiler::executable::Operation`.
* Scan that operation’s root fields to find which are present: root `__typename`, schema intropsection, and "normal" non-meta fields. This is a preliminary step to moving introspection out of `BridgeQueryPlanner`.

As a side-effect observable to clients, the exact response for some error cases like an empty document changes slightly as the error no longer comes from JavaScript code.

Also optional drive-by refactor in the second commit: change `spec::Query` to contain a single `Operation` instead of a `Vec<Operation>`. This removes conditionals and the need to pass `operationName` around in a lot of places, notably in execution code.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/query_planner/bridge_query_planner.rs
+++ b/apollo-router/src/query_planner/bridge_query_planner.rs
@@ -543,11 +543,9 @@ impl BridgeQueryPlanner {
             IntrospectionMode::Rust => {
                 let schema = self.schema.clone();
                 let response = Box::new(
-                    tokio::task::spawn_blocking(move || {
-                        Self::rust_introspection(&schema, &key, &doc)
-                    })
-                    .await
-                    .expect("Introspection panicked")?,
+                    tokio::task::spawn_blocking(move || Self::rust_introspection(&schema, &doc))
+                        .await
+                        .expect("Introspection panicked")?,
                 );
                 return Ok(QueryPlannerContent::Response { response });
             }
@@ -582,7 +580,7 @@ impl BridgeQueryPlanner {
                 let schema = self.schema.clone();
                 let js_result_clone = js_result.clone();
                 tokio::task::spawn_blocking(move || {
-                    let rust_result = match Self::rust_introspection(&schema, &key, &doc) {
+                    let rust_result = match Self::rust_introspection(&schema, &doc) {
                         Ok(response) => {
                             if response.errors.is_empty() {
                                 Ok(response)
@@ -619,11 +617,10 @@ impl BridgeQueryPlanner {
 
     fn rust_introspection(
         schema: &Schema,
-        key: &QueryKey,
         doc: &ParsedDocument,
     ) -> Result<graphql::Response, QueryPlannerError> {
         let schema = schema.api_schema();
-        let operation = doc.get_operation(key.operation_name.as_deref())?;
+        let operation = &doc.operation;
         let variable_values = Default::default();
         let variable_values =
             apollo_compiler::execution::coerce_variable_values(schema, operation, &variable_values)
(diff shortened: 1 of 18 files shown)


apollo-router/src/services/layers/query_analysis.rs
```diff
@@ -257,25 +255,63 @@ pub(crate) struct ParsedDocumentInner {
...
+        Ok(operation.clone())
+    } else if let Some(name) = operation_name {
+        Err(SpecError::UnknownOperation(name.to_owned()))
+    } else if executable.operations.is_empty() {
+        // Maybe not reachable?
+        // A valid document is non-empty and has no unused fragments
+        Err(SpecError::NoOperation)
+    } else {
+        debug_assert!(executable.operations.len() > 1);
+        Err(SpecError::MultipleOperationWithoutOperationName)
```
goto-bus-stop, line 314: This should also be unreachable for a valid document

apollo-router/src/query_planner/bridge_query_planner.rs
```diff
@@ -895,70 +890,48 @@ impl BridgeQueryPlanner {
...
-                    }
-                }
+        if doc.has_root_typename && !doc.has_schema_introspection && !doc.has_explicit_root_fields {
+            // Fast path for __typename alone
+            if doc
+                .operation
+                .selection_set
+                .selections
+                .iter()
+                .all(|sel| sel.as_field().is_some_and(|f| f.name == "__typename"))
```
goto-bus-stop, line 900: this could/should also use `operation.root_fields()`, i guess, but doesn't need to be addressed here

goto-bus-stop (approved):
LGTM, removing Vec<Operation> is very nice 🥳

apollo-router/src/services/layers/query_analysis.rs
SimonSapin: Is it really? `operationName` is not an input of validation, only of [`GetOperation`](https://spec.graphql.org/draft/#GetOperation()) which is exactly this function

apollo-router/src/query_planner/bridge_query_planner.rs
SimonSapin: `root_fields()` is used to compute `doc.has_root_typename` and implicitly descends into inline and named fragments regardless of any directive they might have. This specifically only looks at field, in order to not deal with potential `@skip` or `@include` that fragments might have.

… but now I realize that fields can also have `@skip` or `@include`, so I’ll add this restriction to the condition here.

SimonSapin: I’ll simplify this further in the next PR, having a fast path only for literally `{__typename}` and letting introspection code deal with any other combination of fragments, aliases, directives, etc.

apollo-router/src/services/layers/query_analysis.rs
goto-bus-stop: Ohh yes I see, I misunderstood the condition for this error. There are multiple operations (in the document) without an operation name (in the graphql request).

apollo-router/src/query_planner/bridge_query_planner.rs
goto-bus-stop: I figured that if we have a fast path for root typename, we might as well have a fast path for something like this?
```graphql
fragment tyname on Query { __typename }
{
  ... { ...tyname }
}
```

Though I wonder how useful a `__typename` fast path is in general with Rust introspection.

PR #6129, opened 2024-10-08
Add deduplicate_query_ignored_headers option for traffic shaping

andrewmcgivery:
Add deduplicate_query_ignored_headers option for traffic shaping

TODO: Unit Tests

Manually tested with a K6 script which sent through requests where each request had a unique header value (guid). Ran with and without that flag as "ignored".

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


.changesets/feat_feature_deduplicate_query_ignored_headers.md
```diff
@@ -0,0 +1,14 @@
...
+When the Router is figuring out if a query can be deduplicated, it considers the entire request, including headers. This means that if you have two requests that have a different header value but are otherwise identical, they will not be deduplicated.
+
+You can use the new `deduplicate_query_ignored_headers` option to provide a list of headers to remove from this deduplication calculation.
+
+```yaml title="router.yaml"
+traffic_shaping:
+  all:
+    deduplicate_query: true
+    deduplicate_query_ignored_headers: ["x-my-header"]
+```
```
bnjjj, line 12:
```suggestion
```yaml title="router.yaml"
traffic_shaping:
  all:
    deduplicate_query:
      enabled: true
      ignored_headers:
      - "x-my-header"
```
```

Comment (bnjjj) [reply .changesets/feat_feature_deduplicate_query_ignored_headers.md:12]: I was thinking about that kind of configuration instead

File: apollo-router/src/plugins/traffic_shaping/deduplication.rs [hunk 2]
```diff
@@ -57,23 +68,47 @@ impl Clone for CloneSubgraphResponse {
...
+                for header in headers {
+                    cloned_request.subgraph_request.headers_mut().remove(header);
+                }
+                (
+                    (cloned_request.subgraph_request).into(),
+                    authorization_cache_key,
+                )
+            } else {
+                ((&request.subgraph_request).into(), authorization_cache_key)
+            };
```
bnjjj, line 103: I think to make something more performant we will have to change the cache_key itself and hash the whole request. A little bit like we did for subscription. And talking about subscription that would be great if we could replicate the same logics for subscription too. The main issue I can see here is we're cloning the http request and also ignored headers.

apollo-router/src/plugins/traffic_shaping/deduplication.rs
```diff
@@ -88,8 +123,8 @@ where
         }
         loop {
             let mut locked_wait_map = wait_map.lock().await;
-            let authorization_cache_key = request.authorization.clone();
-            let cache_key = ((&request.subgraph_request).into(), authorization_cache_key);
+            let cache_key =
+                Self::get_cache_key(request.clone(), deduplicate_query_ignored_headers.clone());
```
bnjjj, line 127: Double clone happening here and inside. It's bad for perfs

apollo-router/src/plugins/traffic_shaping/deduplication.rs
```diff
@@ -120,8 +155,11 @@ where
                     drop(locked_wait_map);
 
                     let context = request.context.clone();
-                    let authorization_cache_key = request.authorization.clone();
-                    let cache_key = ((&request.subgraph_request).into(), authorization_cache_key);
+                    let cache_key = Self::get_cache_key(
+                        request.clone(),
+                        deduplicate_query_ignored_headers.clone(),
```
bnjjj, line 160: Avoid the clone

apollo-router/src/plugins/traffic_shaping/deduplication.rs
```diff
@@ -191,8 +229,17 @@ where
 
         if request.operation_kind == OperationKind::Query {
             let wait_map = self.wait_map.clone();
-
-            Box::pin(async move { Self::dedup(service, wait_map, request).await })
+            let deduplicate_query_ignored_headers = self.deduplicate_query_ignored_headers.clone();
```
bnjjj, line 232: Let's try to not clone it

bnjjj (changes requested):

.changesets/feat_feature_deduplicate_query_ignored_headers.md
andrewmcgivery: I like that too but wouldn't this be a breaking change? 🤔

apollo-router/src/plugins/traffic_shaping/deduplication.rs
andrewmcgivery: I am cloning here because...

1. The other items are being cloned to pass in due to the `move` below (I believe)
2. If I try to pass a borrowed value, I get into lifetime errors (yuck)

apollo-router/src/plugins/traffic_shaping/deduplication.rs
andrewmcgivery: I managed to get rid most of the cloning by passing borrowed values.

I'm not sure if I can avoid the borrow on line 93 here because I need to create a cache key that doesn't include that header without mutating the original request.

Open to suggestions though!

.changesets/feat_feature_deduplicate_query_ignored_headers.md
Geal: not if a config migration is written. See https://github.com/apollographql/router/tree/dev/apollo-router/src/configuration/migrations for how it is done

apollo-router/src/plugins/traffic_shaping/deduplication.rs
Geal: I agree with @bnjjj here. We don't need to keep the entire request as key, it can be a hash. That will be shorter, fastrr to compare, use less memory etc

.changesets/feat_feature_deduplicate_query_ignored_headers.md
andrewmcgivery: Oh, awesome. Thanks @Geal / @bnjjj ! Updated with this config and a migration!

apollo-router/src/plugins/traffic_shaping/deduplication.rs
andrewmcgivery: Okay, awesome. I rewrote this to use a hash of the request + a hash of the authorization context!

abernix:
Totally reasonable to re-open this, but I'll close this for now for focused review reasons.

PR #6127, opened 2024-10-08
fix(federation): use `out_edges_with_federation_self_edges` in query builder

duckki:
Instead of using petgraph's `edges_directed` method.
This PR matches JS QP's ordering for `provides` edges.

The change in the `copy_for_provides_internal` method was strictly necessary. The other changes in `add_provides_edges` was made for consistency.

---

**Checklist**

Complete the checklist (and note appropriate exceptions) before the PR is marked ready-for-review.

- [x] Changes are compatible[^1]
- [ ] Documentation[^2] completed
- [x] Performance impact assessed and acceptable
- Tests added and passing[^3]
    - [ ] Unit Tests
    - [x] Integration Tests
    - [x] Manual Tests

[ROUTER-802]: https://apollographql.atlassian.net/browse/ROUTER-802?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #6132, opened 2024-10-09
fix(federation): updated GraphPath to use a custom MaxHeap, instead of BinaryHeap

duckki:
This PR is a follow-up of #6127, which fixed an ordering issue.
The custom heap behaves like the JS QP's [`popMin` method](https://github.com/apollographql/federation/blob/cc4573471696ef78d04fa00c4cf8e5c50314ba9f/query-graphs-js/src/graphPath.ts#L1375).
- JS `popMin` method returns the minimum item in the heap. Also, it returns the first inserted one if there are multiple equally minimum items.
- On the other hand, Rust's BinaryHeap returns an arbitrary item in that case, which caused mismatches.
- It appears that `@interfaceObject` is necessary to create the original issue.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [x] Integration Tests
    - [x] Manual Tests

[ROUTER-802]: https://apollographql.atlassian.net/browse/ROUTER-802?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apollo-federation/src/query_graph/graph_path.rs
+++ b/apollo-federation/src/query_graph/graph_path.rs
@@ -1,5 +1,4 @@
 use std::cmp::Ordering;
-use std::collections::BinaryHeap;
 use std::fmt::Display;
 use std::fmt::Formatter;
 use std::fmt::Write;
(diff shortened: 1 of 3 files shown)


duckki:
We have discussed about using a more efficient data structure than a simple array. However, I'd like to get this merged for now, since this PR may be simply reverted. This heap data structor won't hold a lot of items anyways. If we want to keep this behavior and it shows up on flamegraph, we can reimplement a better struct at that time.

goto-bus-stop (approved):
I agree with your assessment @duckki, performance seems to be either good enough or irrelevant here.

PR #6133, opened 2024-10-09
Set OTEL_STATUS_CODE for connector spans

pubmodmatt:
Set the `otel.status_code` for connector spans to `ERROR` if any of the HTTP requests fail. This enables visualizing errors in traces.

Screenshot 2024-10-11 at 5 12 33 PM

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

**Notes**

[CNN-459]: https://apollographql.atlassian.net/browse/CNN-459?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apollo-router/src/services/http/service.rs
+++ b/apollo-router/src/services/http/service.rs
@@ -18,6 +18,7 @@ use hyper_rustls::HttpsConnector;
 #[cfg(unix)]
 use hyperlocal::UnixConnector;
 use opentelemetry::global;
+use opentelemetry_semantic_conventions::trace::HTTP_RESPONSE_STATUS_CODE;
 use pin_project_lite::pin_project;
 use rustls::ClientConfig;
 use rustls::RootCertStore;
@@ -30,6 +31,7 @@ use tower_http::decompression::Decompression;
 use tower_http::decompression::DecompressionBody;
 use tower_http::decompression::DecompressionLayer;
 use tracing::Instrument;
+use tracing::Span;
 
 use super::HttpRequest;
 use super::HttpResponse;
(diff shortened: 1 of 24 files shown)


apollo-router/src/services/http/service.rs
```diff
@@ -267,6 +272,8 @@ impl tower::Service<HttpRequest> for HttpClientService {
             "net.transport" = "ip_tcp",
             //"apollo.subgraph.name" = %service_name,
             //"graphql.operation.name" = %operation_name,
+            "otel.status_code" = tracing::field::Empty,
+            "http.response.status_code" = tracing::field::Empty,
```
bnjjj, line 276: This shouldn't be enabled by default it should follow the custom telemetry configuration like started [here](https://github.com/apollographql/router/pull/6125)

pubmodmatt: Agreed, but there isn't a way to do that today, as I understand it. This span is created internally here inside the HTTP client service, so there wouldn't be a way for a plugin to add to it. Just above this, it similarly adds other attributes like `http.route` and `http.url`. If feel strongly that `http.response.status_code` shouldn't be here, I can remove it, but it's extremely helpful to be able to see this on the trace.

pubmodmatt: In the interest of getting this landed, I've removed the changes to the HTTP client span. This will just set the `OTEL_STATUS_CODE` on the connector trace, making it visible in traces.

bnjjj (approved):

apollo-router/src/services/http/service.rs
bnjjj: I agree it's valuable but everything we add by default cannot be removed afterward because it would create a breaking change so I prefer to not add it right now. Thanks

PR #6135, opened 2024-10-10
fix(dual_query_planner): updated semantic diff to set-comparison on `requires`

duckki:
- In rare cases, the `requires` field can have multiple items with the same SelectionKey.
- The previous comparison method could fail to match them if they are in different order.

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [x] Manual Tests

[ROUTER-803]: https://apollographql.atlassian.net/browse/ROUTER-803?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apollo-router/src/query_planner/dual_query_planner.rs
+++ b/apollo-router/src/query_planner/dual_query_planner.rs
@@ -268,7 +268,7 @@ fn fetch_node_matches(this: &FetchNode, other: &FetchNode) -> Result<(), MatchFa
     check_match_eq!(*operation_kind, other.operation_kind);
     check_match_eq!(*id, other.id);
     check_match_eq!(*authorization, other.authorization);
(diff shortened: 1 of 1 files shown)


goto-bus-stop (commented):
The implementation looks good to me but I do think that this would be better addressed by merging the conditions in the Rust query planner already.

duckki:
At this time, we have unmerged items in both JS QP and Rust QP and those items are identical (just came in different order). This PR just addresses the false mismatches.

I'll create a follow-up ticket to merge/drop redundancies in Rust QP.

PR #6136, opened 2024-10-10
Add Environment Variable To Helm Chart To Check Usage

jonathanrainer:
Add an Environment variable to the container that's deployed so that we can easily scan if customers are deploying using the official Helm Chart or some other way.

Ran this through the templater and it generates correctly so there's little to test beyond that.

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [x] Integration Tests
    - [x] Manual Tests

**Notes**

### DIFF
--- a/helm/chart/router/templates/deployment.yaml
+++ b/helm/chart/router/templates/deployment.yaml
@@ -73,7 +73,7 @@ spec:
             {{- end }}
           {{- end }}
           env:
-            - name: APOLLO_OFFICIAL_ROUTER_HELM_CHART
+            - name: APOLLO_ROUTER_OFFICIAL_HELM_CHART
               value: "true"
           {{- if or .Values.managedFederation.apiKey .Values.managedFederation.existingSecret .Values.managedFederation.graphRef .Values.extraEnvVars }}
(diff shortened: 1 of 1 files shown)


nmoutschen (approved):

nmoutschen (approved):

helm/chart/router/templates/deployment.yaml
```diff
@@ -72,8 +72,10 @@ spec:
             - /app/configuration.yaml
             {{- end }}
           {{- end }}
-          {{- if or .Values.managedFederation.apiKey .Values.managedFederation.existingSecret .Values.managedFederation.graphRef .Values.extraEnvVars }}
           env:
+            - name: APOLLO_OFFICIAL_ROUTER_HELM_CHART
```
garypen, line 76: The naming convention is:
```
APOLLO_ROUTER_<blah>
```

Is this deliberately not following the convention for some reason?

jonathanrainer: No that's just me not knowing the convention, I'll get that fixed up :)

garypen (approved):

PR #6137, opened 2024-10-10
fix(federation): correct order of @skip and @include conditions on the same selection

goto-bus-stop:
PR https://github.com/apollographql/router/pull/6120 was the wrong solution. JS doesn't actually maintain the order
of the conditions, like I thought. Instead, it always puts `@skip` first
and `@include` second, the opposite of what Rust was doing.

https://github.com/apollographql/federation/blob/cc4573471696ef78d04fa00c4cf8e5c50314ba9f/query-graphs-js/src/pathContext.ts#L13-L33

The queries I was testing with just happened to pass in https://github.com/apollographql/router/pull/6120.

This changes the implementation of `Conditions::from_directives` to
pick the directives out manually instead of iterating. Technically, this
does two iterations of the directive list...but, the code is a bit
simpler, I think.

In the second commit I do a small refactor to replace the `is_negated`
boolean by a `ConditionKind` enum. I found it quite tricky to
constantly translate the skip/include concepts to negation or not
while reading the code, so I hope this simplifies it a bit.

[ROUTER-799]: https://apollographql.atlassian.net/browse/ROUTER-799?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apollo-federation/src/query_plan/conditions.rs
+++ b/apollo-federation/src/query_plan/conditions.rs
@@ -9,6 +9,7 @@ use indexmap::map::Entry;
 use serde::Serialize;
 
 use crate::error::FederationError;
+use crate::internal_error;
 use crate::operation::DirectiveList;
 use crate::operation::NamedFragments;
 use crate::operation::Selection;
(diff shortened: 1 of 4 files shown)


apollo-federation/src/query_plan/conditions.rs
```diff
@@ -70,42 +71,52 @@ impl Conditions {
...
+    /// Parse @skip and @include conditions from a directive list.
     pub(crate) fn from_directives(directives: &DirectiveList) -> Result<Self, FederationError> {
         let mut variables = IndexMap::default();
-        for directive in directives.iter() {
-            let negated = match directive.name.as_str() {
-                "include" => false,
-                "skip" => true,
-                _ => continue,
+
+        if let Some(skip) = directives.get("skip") {
```
duckki, line 123: It appears that the `directives` may have multiple skip/include directive applications. This will handle only one of each type.

duckki (changes requested):
Just one question/request. Otherwise, I like the refactoring!

duckki (approved):

apollo-federation/src/query_plan/conditions.rs
```diff
@@ -32,99 +50,140 @@ pub(crate) enum Conditions {
...
-                    )))
+                    internal_error!("expected boolean or variable `if` argument, got {value}");
                 }
             }
         }
+
         Ok(Self::from_variables(variables))
     }
 
+    // TODO(@goto-bus-stop): what exactly is the difference between this and `Self::merge`?
```
duckki, line 170: Indeed, `merge` makes more sense. It's unclear why `update_with` is so loose in the `Boolean(false)` cases.

PR #6139, opened 2024-10-10
Remove legacy schema introspection

SimonSapin:
Schema introspection now runs natively without involving JavaScript code. We have high confidence that the two implementations return matching responses based on differential testing fuzzing arbitrary queries against a large schema, and testing a corpus of customer schemas against a comprehensive query.

In Router YAML configuration:

* The `experimental_introspection_mode` key is removed, `new` is now the only behavior
* The `supergraph.query_planning.legacy_introspection_caching` key is removed, the behavior is now similar to what was `false`: introspection responses are not part of the query plan cache but instead in a separate, small, in-memory only cache.

Migrations ensure that existing configuration files will keep working.

Fixes ROUTER-338

---

**Checklist**

- Existing relevant tests passing[^3]
    - [x] Unit Tests
    - [x] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


lrlna:
If introspection is now completely in rust, does that mean v8 won't start up with just `experimental_query_planner_mode: new`? If so, we need to update a few more docs [here](https://github.com/apollographql/router/blob/9e3219c603657eb17ed351c79349c01decd06686/docs/source/configuration/overview.mdx?plain=1#L592-L599) and [here](https://github.com/apollographql/router/blob/9e3219c603657eb17ed351c79349c01decd06686/docs/source/executing-operations/native-query-planner.mdx?plain=1#L35-L40). It seems the docs we serve are now just on `next` branch 🤔 🤔 🤔 so I guess doc changes have to be PR'ed separately to `next`?

apollo-router/src/query_planner/bridge_query_planner.rs
```diff
@@ -236,27 +224,6 @@ impl PlannerMode {
...
-    ) -> Result<Arc<Introspection>, ServiceBuildError> {
-        let js_planner = match self {
-            Self::Js(js) => js.clone(),
-            Self::Both { js, .. } => js.clone(),
-            Self::Rust(_) => {
-                // JS "planner" (actually runtime) was not created for planning
-                // but is still needed for introspection, so create it now
-                Self::js_planner(sdl, configuration, old_js_planner).await?
-            }
-        };
```
SimonSapin, line 254: > If introspection is now completely in rust, does that mean v8 won't start up with just `experimental_query_planner_mode: new`? 

Correct. This is the relevant code change. Here the badly-named `js_planner` represents a V8 runtime.

I didn’t realize we publish docs from `next`, I’ll make a separate PR

apollo-router/tests/integration/redis.rs
```diff
@@ -959,26 +978,6 @@ async fn query_planner_redis_update_planner_mode() {
     .await;
 }
 
-#[tokio::test(flavor = "multi_thread")]
-async fn query_planner_redis_update_introspection() {
```
SimonSapin, line 963: This test is removed because it specifically tests that enabling introspection in config changes Redis cache keys. This PR changes `caching_query_planner.rs` for this to no longer be the case, since introspection responses (or "introspection is disabled" errors) are now never in this cache.

apollo-router/src/introspection.rs
```diff
@@ -1,142 +1,163 @@
...
-                    incremental_delivery: Some(IncrementalDeliverySupport {
-                        enable_defer: Some(true),
-                    }),
-                    graphql_validation: true,
-                    reuse_query_fragments: Some(false),
-                    generate_query_fragments: None,
-                    debug: None,
-                    type_conditioned_fetching: false,
-                },
+    fn execute_intropsection(schema: &spec::Schema, doc: &ParsedDocument) -> graphql::Response {
```
Geal, line 141:
```suggestion
    fn execute_introspection(schema: &spec::Schema, doc: &ParsedDocument) -> graphql::Response {
```

Geal (approved):
thank you, it's great to see that code finally removed!

docs/source/configuration/overview.mdx
```diff
@@ -610,20 +610,6 @@ The default value of `experimental_parallelism` is `1`.
 
 In practice, you should tune `experimental_parallelism` based on metrics and benchmarks gathered from your router.
 
-<MinVersion version="1.51.0">
-
-### Introspection response caching
-
-</MinVersion>
```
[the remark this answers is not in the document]
SimonSapin: Not as far as I can see

apollo-router/src/introspection.rs
SimonSapin: I seem to like that particular typo, I made it in two places

PR #6140, opened 2024-10-10
feat: experimental circuit breaker plugin

### LINKED ISSUE #2251: Support for circuit breaking in subgraph requests
Some customers are blocked on migrating from Gateway to Router because they've implemented client-side circuit breaking in their RemoteGraphQLDataSource subclasses.

A circuit breaking section of the [traffic shaping](https://www.apollographql.com/docs/router/configuration/traffic-shaping) configuration. I'm not entirely sure what that configuration would look like, but [envoy has an example.](https://www.envoyproxy.io/docs/envoy/latest/configuration/upstream/cluster_manager/cluster_circuit_breakers#config-cluster-manager-cluster-circuit-breakers)

Some customers can use their service mesh, but not all.

**Additional context**
I wonder if we can make this more GraphQL-aware. Subgraphs are usually stateless/horizontally-scalable services that depend on underlying data sources (REST APIs, databases, etc.) We could provide more benefit by protecting those data services without making the subgraphs implement circuit breakers themselves.

We could count and identify problems with:

* Specific subgraph fetches
* Errors on a particular `path` (`{ "errors": [{ "message": "no db connection", "path": ["products", 0, "reviews"] }] }`)

and trip the circuit for isolated sections of the graph. This lessens the blast radius of the circuit breaker (otherwise errors in a single resolver could "take down" an entire subgraph.)

samuelAndalon:
Fixes #2251

Circuit breaker plugin that will return a `GraphQLError` with `CIRCUIT_BREAKER_OPEN` error extension code, if the `StateMachine` is in `Open` state, or if the call is not permitted, like being in `HalfOpen` state.

The plugin is configurable in different ways.

### Target all subgraphs with default configuration
```yaml
experimental.subgraph_circuit_breaker:
  enabled: true
 subgraphs: All
```

### Target all subgraphs with spefific configuration
```yaml
 experimental.subgraph_circuit_breaker:
  enabled: true
  configuration:
    success_rate: 0.5
    minimum_requests: 2
    success_rate_window_seconds: 10
    constant_backoff_seconds: 5
  subgraphs: All
```

### Target all subgraphs with with default circuit_breaker_configuration and overrides
```yaml
experimental.subgraph_circuit_breaker:
  enabled: true
  circuit_breaker_configuration:  # default values
    success_rate: 0.5
     minimum_requests: 2
     success_rate_window_seconds: 10
     constant_backoff_seconds: 5
subgraphs:
  AllWithOverrides:
    subgraph-a:
      success_rate: 0.8
      minimum_requests: 20
      success_rate_window_seconds: 1
      constant_backoff_seconds: 5
   # all other subgraphs will use provided configuration
```

### Target only some subgraphs with default configuration
```yaml
experimental.subgraph_circuit_breaker:
  enabled: true
  subgraphs: 
    Only:
      - subgraph-a
      - subgraph-b
```

### Target only some subgraphs with overrides configuration
```yaml
experimental.subgraph_circuit_breaker:
  enabled: true
subgraphs:
  OnlyWithOverrides:
    subgraph-a:
      success_rate: 0.8
      minimum_requests: 20
      success_rate_window_seconds: 1
      constant_backoff_seconds: 5
   # no more circuit breakers
```

### Target all subgraphs except
```yaml
experimental.subgraph_circuit_breaker:
  enabled: true
  subgraphs: 
    Except:
      - subgraph-a
      - subgraph-b
```

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -268,6 +268,7 @@ dependencies = [
  "diff",
  "displaydoc",
  "ecdsa",
+ "failsafe",
  "flate2",
  "fred",
  "futures",
@@ -2487,6 +2488,18 @@ dependencies = [
  "tokio",
  "tower",
  "tracing",
+]
+
+[[package]]
+name = "failsafe"
+version = "1.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a6b1dc7c28e66a8f5dc32b7043b735e93d9eea8ea6e7be5507dae5850ab7ac70"
+dependencies = [
+ "futures-core",
+ "parking_lot",
+ "pin-project",
+ "rand 0.8.5",
 ]
 
 [[package]]
--- a/apollo-router/Cargo.toml
+++ b/apollo-router/Cargo.toml
@@ -95,6 +95,7 @@ derive_more = { version = "0.99.17", default-features = false, features = [
 dhat = { version = "0.3.3", optional = true }
 diff = "0.1.13"
 displaydoc = "0.2"
+failsafe = "1.3.0"
 flate2 = "1.0.30"
 fred = { version = "7.1.2", features = ["enable-rustls"] }
 futures = { version = "0.3.30", features = ["thread-pool"] }
--- a/apollo-router/src/plugins/mod.rs
+++ b/apollo-router/src/plugins/mod.rs
@@ -41,3 +41,4 @@ pub(crate) mod telemetry;
 #[cfg(test)]
 pub(crate) mod test;
 pub(crate) mod traffic_shaping;
+mod subgraph_circuit_breaker;
--- a/apollo-router/src/plugins/subgraph_circuit_breaker.rs
+++ b/apollo-router/src/plugins/subgraph_circuit_breaker.rs
@@ -1,0 +1,231 @@
+use std::collections::HashMap;
+use std::ops::ControlFlow;
+use std::sync::Arc;
+use std::time::Duration;
+
+use dashmap::DashMap;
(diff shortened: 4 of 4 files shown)


samuelAndalon:
The only pending thing is a review, it might take a while, I advice you to implement this in your own space if you have a custom build of the router.

tninesling:
Hi @samuelAndalon. I'm looking into circuit breaking on the Apollo side, and I'm hoping you could help me better understand your use case. I don't see much GQL-specific code here other than the creation of the `GraphQLError`, so I'm wondering if the Router is the right place for this. Is there something blocking you from using an existing proxy or service mesh to do this?

samuelAndalon:
>so I'm wondering if the Router is the right place for this. Is there something blocking you from using an existing proxy or service mesh to do this?

you could yes, but that means that if the breaker is open @ the mesh layer, the router would still be preparing the graphql request, execute all the plugins logic, etc, just to have the mesh to block the request, that's kind of a waste of resources, IMHO the best place to have a circuit breaker is @ the application layer,

>I don't see much GQL-specific code here other than the creation of the GraphQLError

that's the point of a circuit breaker, to wrap your fallible logic into an state machine

aaronArinder:
this looks pretty sweet, but I'm closing for staleness; we have something cooking internally that might use this as as inspiration, though, so thank you!

PR #6141, opened 2024-10-11
fix(federation): fixed sibling typename optimization handling multiple __typename selections

duckki:
### Problem
The sibling typename optimization treated __typename and __typename @skip(if:$X) interchangeably. Thus, one could become a sibling of the other, depending on the order they appear in query. That is problematic, since sibling typename should not have directives or aliases.

The same issue found in JS QP. The matching PR for that is https://github.com/apollographql/federation/pull/3164.
 
Rust QP didn't have the other problem dropping some __typename selections in fetch operations like JS QP. But, still ordering could be wrong, picking __typename with directives ahead of the plain `__typename`.

### Fix Summary    
- updated to only fold a plain __typename without alias/directives.
- also, fixed __typename ordering in fetch operations to put the plain __typename ahead of others.

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [x] Manual Tests

[ROUTER-801]: https://apollographql.atlassian.net/browse/ROUTER-801?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apollo-federation/src/operation/mod.rs
+++ b/apollo-federation/src/operation/mod.rs
@@ -995,6 +995,7 @@ mod field_selection {
     use apollo_compiler::Name;
     use serde::Serialize;
 
+    use super::TYPENAME_FIELD;
     use crate::error::FederationError;
     use crate::operation::ArgumentList;
     use crate::operation::DirectiveList;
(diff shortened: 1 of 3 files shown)


goto-bus-stop (approved):
This LGTM, I think we should audit all our __typename checks soon to see if we are doing the right thing, IIRC there are several places where we just `== TYPENAME_FIELD`, sometimes checking for directives and sometimes not. I'd also like to know if we do the right thing when there are multiple typename selections with different aliases, and other edge cases

duckki:
> This LGTM!
> 
> e; filed a ticket for the below.
> 
> I think we should audit all our __typename checks soon to see if we are doing the right thing, IIRC there are several places where we just `== TYPENAME_FIELD`, sometimes checking for directives and sometimes not. I'd also like to know if we do the right thing when there are multiple typename selections with different aliases, and other edge cases

I've casually looked around and didn't find any other places that need to be fixed. But, it wasn't exhaustive.

sachindshinde (approved):

PR #6142, opened 2024-10-11
Make BridgeQueryPlannerPool pass-through for new QP

SimonSapin:
Because each task in the pool (by default just one) takes queries to plan from a queue in sequence, it is a parallelism bottleneck even in "new" mode

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/query_planner/bridge_query_planner_pool.rs
+++ b/apollo-router/src/query_planner/bridge_query_planner_pool.rs
@@ -22,6 +22,7 @@ use tower::ServiceExt;
 use super::bridge_query_planner::BridgeQueryPlanner;
 use super::QueryPlanResult;
 use crate::cache::storage::CacheStorage;
+use crate::configuration::QueryPlannerMode;
 use crate::error::QueryPlannerError;
 use crate::error::ServiceBuildError;
 use crate::graphql::Response;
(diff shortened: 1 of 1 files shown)


apollo-router/src/query_planner/bridge_query_planner_pool.rs
```diff
@@ -250,18 +288,16 @@ impl tower::Service<QueryPlannerRequest> for BridgeQueryPlannerPool {
...
-            std::task::Poll::Ready(Err(QueryPlannerError::PoolProcessing(
-                "query plan queue is full".into(),
-            )))
-        } else {
-            std::task::Poll::Ready(Ok(()))
+        match &self.pool_mode {
+            PoolMode::Pool { sender, .. } if sender.is_full() => std::task::Poll::Ready(Err(
+                QueryPlannerError::PoolProcessing("query plan queue is full".into()),
+            )),
+            _ => std::task::Poll::Ready(Ok(())),
```
garypen, line 295: As discussed, we need a way to set a bound on this. I'd be happy with a hard-coded limit (semaphore controlled maybe, perhaps not configurable) until we have a better mechanism.

SimonSapin: If we also merge https://github.com/apollographql/router/pull/6122 then in pass-through mode we could generate backpressure if [`thread_poll.queued_count()`](https://docs.rs/threadpool/latest/threadpool/struct.ThreadPool.html#method.queued_count) is above some threshold

By the way, backpressure be generated by returning `Poll::Pending` rather than `Poll::Ready(Err(…))`?

https://docs.rs/tower/latest/tower/trait.Service.html#tymethod.poll_ready

garypen: Yep. I think the reason it was done as a ready error is to fit in with other problems in our existing pipeline. I agree that we should fix all that stuff and make it `Poll::Pending`.

Geal: the `queued_count()` value can also be used to generate the `apollo.router.query_planning.queued` metric

Geal (approved):
as mentioned by @garypen we should keep a way to provide backpressure on the queue of queries to plan

PR #6143, opened 2024-10-11
Return a specific error when attempting to format an Int and it turns out to be bigger than I32

andrewmcgivery:
Return a specific error when attempting to format an Int and it turns out to be bigger than I32

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


BrynCooke:
This looks great. I think we should have this behaviour.

goto-bus-stop:
Thanks for the PR! It looks like we are actually not raising field errors for *any* values that are of the wrong type. Also, apparently the `parameters.errors` array used here does not actually get returned to clients as a field error... definitely not what I would've expected either 😭

#5787 is implementing result coercion errors more broadly, and it looks like it also addresses the i32 size issue that this PR fixes. In addition to making sure errors are actually reported as errors. My preference is to land #5787. I'll add the test you built here to that PR.

abernix:
👋  Thanks for opening this.  I'm closing this in lieu of #5787 for the reasons above.

PR #6144, opened 2024-10-11
AWS SigV4 signing for Apollo Connectors

pubmodmatt:
Allow AWS SigV4 signing for Apollo Connectors requests. This allows connectors to invoke any AWS service through its HTTP API. Signing options can be configured per Connectors source.

Example configuration:

```yaml
preview_connectors:
  subgraphs:
    aws:
      sources:
        lambda:
          authentication:
            aws_sig_v4:
              default_chain:
                profile_name: "default"
                region: "us-east-1"
                service_name: "lambda"
                assume_role:
                  role_arn: "arn:aws:iam::XXXXXXXXXXX:role/LambdaExecute"
                  session_name: "graphql"
```

Other changes:

* Fix a bug where JSON Selection does not accept the `$context` variable in `apply_with_vars`

Note that tests are currently failing due to an unrelated dependency issue that will be fixed separately.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

**Notes**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -224,9 +224,9 @@ dependencies = [
 
 [[package]]
 name = "apollo-parser"
-version = "0.8.2"
+version = "0.8.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9692c1bfa7e0628e5c46bd0538571dd469138a0f062290fd4bf90e20e46f05da"
+checksum = "b64257011a999f2e22275cf7a118f651e58dc9170e11b775d435de768fad0387"
 dependencies = [
  "memchr",
  "rowan",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -224,9 +224,9 @@ dependencies = [
 
 [[package]]
 name = "apollo-parser"
-version = "0.8.3"
+version = "0.8.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b64257011a999f2e22275cf7a118f651e58dc9170e11b775d435de768fad0387"
+checksum = "9692c1bfa7e0628e5c46bd0538571dd469138a0f062290fd4bf90e20e46f05da"
 dependencies = [
  "memchr",
  "rowan",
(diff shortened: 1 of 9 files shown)


Geal:
why is this done under the connectors configuration and not under the authentication configuration? We can't duplicate the entire configuration under connectors, that will make a confusing UX.

IIRC one of the issues here is that the subgraph name is generated from the source name and other parts from the directive, and that won't make sense for users. So here's a thing that could be done here: extend the `SubgraphConfiguration` type, that's used to handle configuration overrides per subgraph (there's a lot of unintuitive edge cases to do that properly, so that should not be rewritten from scratch for connectors). A good way to do it would be to use a wrapper type that adds the `sources` field and adds a `#[serde(flatten)]` on the `SubgraphConfiguration` field
https://github.com/apollographql/router/blob/871c3c759d9beb9a06c251c148c83ba48ad85203/apollo-router/src/configuration/subgraph.rs#L16-L88

That type could then be used to gradually enable plugins to support connectors

pubmodmatt:
> why is this done under the connectors configuration and not under the authentication configuration?

All of the connectors settings so far are [under `preview_connectors` currently](https://www.apollographql.com/docs/graphos/schema-design/connectors/router), including per-source settings. Since we're now at public preview and moving to GA, it makes sense to move these settings into their integration points. There are some open PRs for things like telemetry that already do this. I'll look at moving this into `authorization`.

The internal subgraph name is a temporary artifact, and not exposed to customers. The current per-source settings, including the ones in this PR, go under the actual name of the subgraph containing the Connector. So rather than a wrapper type, I think `sources` would just go directly under `SubgraphConfiguration`.

However, I'm not sure `sources` alone is enough to convey that these settings are related to connectors, if it moves out from under `preview_connectors`. It may make more sense under a `connector` subheading.

Presumably any Connectors settings under `subgraph` -> `all` would apply to any sources with the same name in multiple subgraphs. That might be useful if the same REST API were used in multiple subgraphs, but the sources would need to have the same name across all subgraphs.

pubmodmatt:
> why is this done under the connectors configuration and not under the authentication configuration?

We had some discussion around this, and have a design where the `connector` settings are at the same level as `subgraph`, and do not implement an `all` mechanism. So settings must be provided for an individual connector source, as is coded up in the latest commits here.

lennyburdette (approved):

PR #6145, opened 2024-10-11
Set up a place for body validations & add the first new check

dylan-apollo:
A bunch of refactoring to set us up for body validations—and adds a _single_ check, if you use empty `$args` with no path but have no arguments defined on the field.

This doesn't use the visitor stuff because we're interested in the opposite info—where's the data coming from, not where is it going to. Left a bunch of TODOs where we should add more validations.

### DIFF
--- a/apollo-federation/src/sources/connect/json_selection/parser.rs
+++ b/apollo-federation/src/sources/connect/json_selection/parser.rs
@@ -297,7 +297,7 @@ impl ExternalVarPaths for NamedSelection {
 
 #[derive(Debug, PartialEq, Eq, Clone)]
 pub struct PathSelection {
-    pub(super) path: WithRange<PathList>,
+    pub(crate) path: WithRange<PathList>,
 }
 
 // Like NamedSelection, PathSelection is an AST structure that takes its range
@@ -378,7 +378,7 @@ impl From<PathList> for PathSelection {
 }
 
 #[derive(Debug, PartialEq, Eq, Clone)]
-pub(super) enum PathList {
+pub(crate) enum PathList {
     // A VarPath must start with a variable (either $identifier, $, or @),
     // followed by any number of PathStep items (the WithRange<PathList>).
     // Because we represent the @ quasi-variable using PathList::Var, this
@@ -1079,7 +1079,7 @@ pub(crate) fn parse_string_literal(input: Span) -> IResult<Span, WithRange<Strin
 }
 
 #[derive(Debug, PartialEq, Eq, Clone, Default)]
-pub(super) struct MethodArgs {
+pub(crate) struct MethodArgs {
     pub(super) args: Vec<WithRange<LitExpr>>,
     pub(super) range: OffsetRange,
 }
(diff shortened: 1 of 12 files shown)


apollo-federation/src/sources/connect/validation/http/body.rs
```diff
@@ -0,0 +1,156 @@
...
+    match json_selection {
+        JSONSelection::Named(_) => Ok(()), // TODO: validate_sub_selection(sub_selection, arg),
+        JSONSelection::Path(path_selection) => validate_path_selection(path_selection, arg),
+    }
+}
+
+fn validate_path_selection(path_selection: PathSelection, arg: Arg) -> Result<(), Message> {
+    let PathSelection { path } = path_selection;
+    match path.as_ref() {
+        PathList::Var(variable, trailing) => {
```
benjamn, line 51: Probably need to process `trailing` in more than just the `$args` case, since other `$args` paths could appear elsewhere along the path, within method arguments for example.

apollo-federation/src/sources/connect/validation/http/body.rs
```diff
@@ -0,0 +1,156 @@
...
+            match variable.as_ref() {
+                KnownVariable::This => {
+                    // TODO: Verify that the name & shape matches arg.coordinate.field_coordinate.object
+                    Ok(())
+                }
+                KnownVariable::Args => validate_args_path(trailing, arg)
+                    .map_err(|err| err.with_fallback_locations(arg.get_selection_location(&path))),
+                KnownVariable::Config => Ok(()), // We have no way of knowing is this is valid, yet
+                KnownVariable::Dollar => {
+                    // TODO: Validate that this is followed only by a method
```
benjamn, line 61: We're now recommending/enforcing the `$.data` syntax for single-key paths, so `$` can definitely be followed by a `.key` as well as a `->` method.

apollo-federation/src/sources/connect/json_selection/parser.rs
```diff
@@ -378,7 +378,7 @@ impl From<PathList> for PathSelection {
 }
 
 #[derive(Debug, PartialEq, Eq, Clone)]
-pub(super) enum PathList {
+pub(crate) enum PathList {
```
benjamn, line 381: I wish the `PathList` enum could remain `pub(super)`, because its internal structure needs to change fairly often, and adding more pattern matching against those internal details makes changing them harder.

Have you considered using the `ExternalVarPaths` trait?
https://github.com/apollographql/router/blob/619b794360d81489e75f8af6c37db99f922406ad/apollo-federation/src/sources/connect/json_selection/parser.rs#L30-L32

This should give you all the variable paths referring to external inputs (like `$this`, `$args`, and `$config`), fully flattened, so you can easily loop over them and validate the `$args` paths.

benjamn (changes requested):
I think the `JSONSelection::external_var_paths` method (provided by `pub(crate) trait ExternalVarPaths`) will help this validation code be a little less coupled to the internal `PathList` structure, and also catch more `$args` paths that are currently skipped.

dylan-apollo:
We're working on a more general interface for iterating over mapping inputs/outputs. We'll use that for validating bodies instead of the AST directly

PR #6149, opened 2024-10-14
Remove experimental_introspection_mode from docs

SimonSapin:
Remove mentions in doc of a config key that is being removed in https://github.com/apollographql/router/pull/6139

This PR goes into the `next` branch to modify docs that only exist there. It should be merged the next time `dev` is merged into `next`, after https://github.com/apollographql/router/pull/6139 is merged into `dev`.

The required migration is in #6139.


garypen (commented):
I wonder if this should be additive... i.e.: keep the docs for v1.56.0 and add the docs for v1.57.0.

I'm not sure what message we want to convey to users. Keeping the 1.56.0 docs is accurate, so...

PR #6151, opened 2024-10-15
Add Fleet Detection Plugin

jonathanrainer:
Adds an initial plugin, that loads at startup and emits metrics for three simple cases: cpus, cpu_freq and total_memory. Not sure this is the correct approach for this, especially as this plugin will expand over time so willing to take any pointers in that regard.

Tested via using OpenTelemetry Collector with the following router config and OTEL config

```yaml
telemetry:
  apollo:
    experimental_otlp_endpoint: "http://0.0.0.0:4317"
  instrumentation:
    spans:
      mode: "spec_compliant"
```

```yaml
receivers:
  otlp:
    protocols:
      grpc:
      http:

exporters:
  debug:
    verbosity: detailed

processors:
  batch:
  filter:
    metrics:
      include:
        match_type: regexp
        metric_names:
          - apollo.router.instance.*

service:
  pipelines:
    metrics:
      receivers: [otlp]
      processors: [filter]
      exporters: [debug]
```

And produced the following 

Screenshot 2024-10-15 at 14 09 42

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [x] Manual Tests

**Notes**


apollo-router/src/plugins/fleet_detector.rs
```diff
@@ -0,0 +1,107 @@
...
+        debug!("beginning environment detection");
+        let sys = &System::new_all();
+        let detector = FleetDetector {};
+        detector.detect_cpu_values(sys);
+        detector.detect_memory_values(sys);
+        Ok(detector)
+    }
+}
+
+impl FleetDetector {
```
nmoutschen, line 31: Since we never use `&self` and there are no field, we might want to move the methods here as functions

apollo-router/src/plugins/fleet_detector.rs
```diff
@@ -0,0 +1,107 @@
...
+        info!(counter.apollo.router.instance.cpu_count = cpu_count);
+    }
+
+    #[cfg(not(target_os = "linux"))]
+    fn detect_cpu_count(&self, system: &System) -> u64 {
+        system.cpus().len() as u64
+    }
+
+    #[cfg(target_os = "linux")]
+    fn detect_cpu_count(&self, system: &System) -> u64 {
```
nmoutschen, line 46: Would be good to explain why we need this additional logic for linux

apollo-router/src/plugins/fleet_detector.rs
```diff
@@ -0,0 +1,107 @@
...
+        detector.detect_cpu_values(sys);
+        detector.detect_memory_values(sys);
+        Ok(detector)
+    }
+}
+
+impl FleetDetector {
+    fn detect_cpu_values(&self, system: &System) {
+        let cpus = system.cpus();
+        let cpu_freq = cpus.iter().map(|cpu| cpu.frequency()).sum::<u64>() / cpus.len() as u64;
```
nmoutschen, line 34: CPU frequency can change over time, is this something we want to detect at regular intervals? Or does this method always return the **max** CPU frequency?

apollo-router/src/configuration/snapshots/apollo_router__configuration__tests__schema_generation.snap
```diff
@@ -3562,6 +3562,32 @@ expression: "&schema"
...
+        },
+        {
+          "description": "Use Rust-based and Javascript-based implementations side by side, logging warnings if the implementations disagree.",
+          "enum": [
+            "both"
+          ],
+          "type": "string"
+        }
+      ]
+    },
```
bnjjj, line 3590: Why is it part of this PR ?

apollo-router/src/configuration/snapshots/apollo_router__configuration__tests__schema_generation.snap
```diff
@@ -4453,6 +4482,11 @@ expression: "&schema"
           "description": "If cache warm up is configured, this will allow the router to keep a query plan created with the old schema, if it determines that the schema update does not affect the corresponding query",
           "type": "boolean"
         },
+        "legacy_introspection_caching": {
+          "default": true,
+          "description": "Activates introspection response caching Historically, the Router has executed introspection queries in the query planner, and cached their response in its cache because they were expensive. This will change soon as introspection will be removed from the query planner. In the meantime, since storing introspection responses can fill up the cache, this option can be used to deactivate it. Default: true",
+          "type": "boolean"
+        },
```
bnjjj, line 4489: same

apollo-router/src/configuration/snapshots/apollo_router__configuration__tests__schema_generation.snap
```diff
@@ -8236,6 +8270,10 @@ expression: "&schema"
       "$ref": "#/definitions/Chaos",
       "description": "#/definitions/Chaos"
     },
+    "experimental_introspection_mode": {
+      "$ref": "#/definitions/IntrospectionMode",
+      "description": "#/definitions/IntrospectionMode"
+    },
     "experimental_query_planner_mode": {
```
bnjjj, line 8277: ??

apollo-router/src/plugins/fleet_detector.rs
```diff
@@ -0,0 +1,126 @@
...
+#[derive(Debug, Default, Deserialize, JsonSchema)]
+struct Conf {}
+
+#[async_trait::async_trait]
+impl Plugin for FleetDetector {
+    type Config = Conf;
+
+    async fn new(_: PluginInit<Self::Config>) -> Result<Self, BoxError> {
+        debug!("beginning environment detection");
+        debug!("spawning continuous detector task");
```
bnjjj, line 26: Do we need these 2 debugs knowing that nothing happens between these 2 logs ? Maybe we can put everything into one log ?

apollo-router/src/plugins/fleet_detector.rs
```diff
@@ -0,0 +1,126 @@
...
+        self.handle.abort();
+    }
+}
+
+fn detect_cpu_values(system: &System) {
+    let cpus = system.cpus();
+    let cpu_count = detect_cpu_count(system);
+    let cpu_freq = cpus.iter().map(|cpu| cpu.frequency()).sum::<u64>() / cpus.len() as u64;
+    info!(value.apollo.router.instance.cpu_freq = cpu_freq);
+    info!(counter.apollo.router.instance.cpu_count = cpu_count);
```
bnjjj, line 54: We're not using this mechanism for metrics anymore. What you're looking for is `u64_counter!` macro (search in the codebase). And a gauge like [here](https://github.com/apollographql/router/blob/dev/apollo-router/src/query_planner/bridge_query_planner_pool.rs#L45:L45) for the gauge please make sure to always keep a reference somewhere it should never be dropped

apollo-router/src/configuration/snapshots/apollo_router__configuration__tests__schema_generation.snap
jonathanrainer: So we kept having CI failures because of changes to the JSON schema for the config I think? And I thought this would fix it but it doesn't! Any ideas much appreciated!

apollo-router/src/plugins/fleet_detector.rs
jonathanrainer: Quite right, that was a hangover from an older version, have shortened it to one :)

jonathanrainer:
Ok @bnjjj I've run the perf tests, building the router from this branch and further enabling metrics so that we can see these now being emitted. I'll upload the `top` file for the router because that's probably the most instructive, I see the memory usage goes up by about 200Mi over the entire course of the test but I don't have any baselines so its hard to compare. I'll also attach the `router` logs. That doesn't seem like a terrible thing to me but it's hard to compare without a baseline.

I have a few further questions as well to push us forward on this:
1. We want to ensure that customers can turn this off, and we were thinking of re-using the APOLLO_TELEMETRY_DISABLED env var, are you folks happy with that?
2. We'll need to add documentation around what's being collected and how to turn it off, should we do that on this branch too for documentation or is it better to do that separately?
3. The CI on this branch seems to be failing because of generating the config schema, and try as I might I can't see how that might be fixed, any pointers?

[otel.log](https://github.com/user-attachments/files/17555381/otel.log)
[router.log](https://github.com/user-attachments
[top.router.txt](https://github.com/user-attachments/files/17555384/top.router.txt)
/files/17555382/router.log)

apollo-router/src/plugins/fleet_detector.rs
```diff
@@ -0,0 +1,213 @@
...
+    }
+}
+
+#[derive(Debug)]
+struct FleetDetector {
+    handle: JoinHandle<()>,
+    #[allow(dead_code)]
+    // We have to store a reference to the gauge otherwise it will be dropped once the plugin is
+    // initialised, even though it still has data to emit
+    freq_gauge: ObservableGauge<u64>,
```
bnjjj, line 67: @BrynCooke Could you confirm it will handle the hot reload properly ?

bnjjj (commented):
@jonathanrainer to make a comparison could you just run the same benchmark on dev and you'll get something to compare with this branch

apollo-router/src/plugins/fleet_detector.rs
```diff
@@ -0,0 +1,213 @@
...
+        if self.start.elapsed() < refresh_at {
+            &self.system
+        } else {
+            self.system.refresh_cpu_all();
+            self.system.refresh_memory();
+            self.refresh_at = ((refresh_at + REFRESH_INTERVAL).as_millis() as u64).into();
+            &self.system
+        }
+    }
+}
```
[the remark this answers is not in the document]
jonathanrainer: Yer that does make it simpler, thanks for the suggestion! :)

jonathanrainer:
@bnjjj Ah yes, apologies should have thought of that, have done that below and redid the tests I posted above just so the comparison is easier. From looking at the memory figures it looks like the plugin does increase memory but it's not the constant increase we were seeing before, also the baseline it starts from in the test appears higher in the branched case which presumably isn't plugin related because it won't be running in the early part of the test (I imagine). Have attached the logs again, let me know if you think there's anything we need to worry about

**Dev Files**
[otel_dev.log](https://gith
[router_dev.log](https://github.com/user-attachments/files/17645650/router_dev.log)
ub.com/user-attachments/files/1764564
[top.router_dev.txt](https://github.com/user-attachments/files/17645652/top.router_dev.txt)
7/otel_dev.log)

**Branch Files**
[otel_branch.log](https://github.com/user-attachments/files/17645656/otel_branch.log)
[router_branch.log](https://github.com/user-attachments/files/17645661/router_branch.log)
[top.router_branch.txt](https://github.com/user-attachments/files/17645668/top.router_branch.txt)

apollo-router/src/plugins/fleet_detector.rs
BrynCooke: Yes I believe it will, but do test

bnjjj (approved):
Thanks @jonathanrainer LGTM

BrynCooke (changes requested):
CPU and memory should be gauges. We also talked about activate.

abernix:
@BrynCooke I think this might be ready for another review?

apollo-router/src/plugins/fleet_detector.rs
```diff
@@ -0,0 +1,248 @@
...
+            .map(|s| String::from(s.trim()))
+            .ok();
+        match (quota, period) {
+            (Some(quota), Some(period)) => {
+                // In v1 quota being -1 indicates no restrictions so return the maximum (all
+                // system CPUs) otherwise divide the two.
+                if quota == "-1" {
+                    system_cpus
+                } else {
+                    quota.parse::<u64>().unwrap() / period.parse::<u64>().unwrap()
```
bnjjj, line 227: Just to make sure, are we feeling confident about these unwrap ?

apollo-router/src/plugins/fleet_detector.rs
```diff
@@ -0,0 +1,248 @@
...
+        match fs::read_to_string("/sys/fs/cgroup/cpu.max") {
+            Ok(readings) => {
+                // The format of the file lists the quota first, followed by the period,
+                // but the quota could also be max which would mean there are no restrictions.
+                if readings.starts_with("max") {
+                    system_cpus
+                } else {
+                    // If it's not max then divide the two to get an integer answer
+                    let (a, b) = readings.split_once(' ').unwrap();
+                    a.parse::<u64>().unwrap() / b.parse::<u64>().unwrap()
```
bnjjj, line 207: Just to make sure, are we feeling confident about these unwrap ?

BrynCooke:
@jonathanrainer I've pushed a commit that makes things simpler. In particular `activate` is now handled uniformly and the signature is changed to ensure that activate will complete.

Please check that you are happy with the change and also do some manual testing to ensure that things still work.

BrynCooke (approved):

nmoutschen (approved):

BrynCooke (approved):

bnjjj (approved):

PR #6158, opened 2024-10-16
fix(dual-query-planner): updated semantic diff to handle fragment name differences

duckki:
Fragment definitions and their names could be slightly reordered/switched. This PR updates semantic diff to handle those differences.

- `same_ast_document` creates a fragment mapping from the JS fragment name to matching Rust fragment name based on their definition body, not their places or names.
- Fragment spreads in selection sets are compared after mapping JS fragment names to Rust fragment names.

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [x] Manual Tests

### DIFF
--- a/apollo-router/src/query_planner/dual_query_planner.rs
+++ b/apollo-router/src/query_planner/dual_query_planner.rs
@@ -719,7 +719,6 @@ fn same_ast_document(x: &ast::Document, y: &ast::Document) -> Result<(), MatchFa
                 _ => others.push(def),
             }
         }
-        fragments.sort_by_key(|frag| frag.name.clone());
         (operations, fragments, others)
     }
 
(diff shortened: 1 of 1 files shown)


duckki:
This PR depends on https://github.com/apollographql/router/pull/6173 to avoid false mismatches.

apollo-router/src/query_planner/dual_query_planner.rs
```diff
@@ -726,28 +733,45 @@ fn same_ast_document(x: &ast::Document, y: &ast::Document) -> Result<(), MatchFa
         "Different number of operation definitions"
     );
 
+    check_match_eq!(x_frags.len(), y_frags.len());
+    let mut fragment_map: HashMap<Name, Name> = HashMap::new();
+    // Assumption: x_frags and y_frags are topologically sorted.
+    //             Thus, we can build the fragment name mapping in a single pass and compare
+    //             fragment definitions using the mapping at the same time, since earlier fragments
+    //             will never reference later fragments.
+    x_frags.iter().try_fold((), |_, x_frag| {
```
TylerBloom, line 742: Not important, but I think you're looking for `.try_for_each` rather than `.try_fold` using a unit.

TylerBloom (approved):

apollo-router/src/query_planner/dual_query_planner.rs
duckki: Well, I wanted to something like `try_all` (https://github.com/rust-lang/rfcs/pull/3233) or `FallibleIterator` (but it's private). We should make `FallibleIterator` more accessible :)

PR #6160, opened 2024-10-16
feat(telemetry): add support of router request_context selector and add schema id in context

bnjjj:
Added a new selector `request_context` for the router service. Now the supergraph schema id is also available through the context so if you want to access or display this schema id at the router service level you can now configure it like this:

```yaml
telemetry:
  instrumentation:
    events:
      router:
        my.request_event:
          message: "my request event message"
          level: info
          on: request
          attributes:
            schema.id:
              request_context: "apollo::supergraph_schema_id" # The key containing the supergraph schema id
```

It can be used in any services at any stages. This example is using selectors on event's attributes but it can also be done on spans and instruments.

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [x] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-806]: https://apollographql.atlassian.net/browse/ROUTER-806?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BrynCooke (approved):
Approved. It doesn't preclude us from adding a dedicated selector later.

PR #6162, opened 2024-10-16
fix(demand_control): use correct parent type in fragment definitions

goto-bus-stop:
Prerequisite for https://github.com/apollographql/router/pull/6013.

Fixes errors like:
```
Attempted to look up a field on type MyInterface, but the field does not exist
```

Fragment definitions have a type condition, and all selections inside
the fragment are executed against that type condition. Previously, the
demand control cost calculator executed them against the parent type
*where the fragment is used*, which is often different.

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

[ROUTER-815]: https://apollographql.atlassian.net/browse/ROUTER-815?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apollo-router/src/plugins/demand_control/cost_calculator/fixtures/basic_supergraph_schema.graphql
+++ b/apollo-router/src/plugins/demand_control/cost_calculator/fixtures/basic_supergraph_schema.graphql
@@ -1,0 +1,102 @@
+schema
+  @link(url: "https://specs.apollo.dev/link/v1.0")
+  @link(url: "https://specs.apollo.dev/join/v0.3", for: EXECUTION)
+{
+  query: Query
+  mutation: Mutation
+}
+
+directive @join__enumValue(graph: join__Graph!) repeatable on ENUM_VALUE
+
+directive @join__field(graph: join__Graph, requires: join__FieldSet, provides: join__FieldSet, type: String, external: Boolean, override: String, usedOverridden: Boolean) repeatable on FIELD_DEFINITION | INPUT_FIELD_DEFINITION
+
+directive @join__graph(name: String!, url: String!) on ENUM_VALUE
+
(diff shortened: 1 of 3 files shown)


apollo-router/src/plugins/demand_control/cost_calculator/fixtures/basic_supergraph_schema.graphql
```diff
@@ -0,0 +1,102 @@
+schema
```
goto-bus-stop, line 1: This is just the existing `basic_schema.graphql` run through rover compose to get a supergraph with a single subgraph.

apollo-router/src/plugins/demand_control/cost_calculator/static_cost.rs
```diff
@@ -321,12 +320,12 @@ impl StaticCostCalculator {
...
             ),
             Selection::FragmentSpread(s) => {
-                self.score_fragment_spread(ctx, s, parent_type, list_size_directive)
+                self.score_fragment_spread(ctx, s, list_size_directive)
             }
             Selection::InlineFragment(i) => self.score_inline_fragment(
                 ctx,
                 i,
-                i.type_condition.as_ref().unwrap_or(parent_type),
+                parent_type,
```
goto-bus-stop, line 328: Moved this into `score_inline_fragment`, I think that makes more sense. It's less brittle if we add more places that call `score_inline_fragment` in the future, as every call site doesn't need to account for the type condition.

goto-bus-stop:
oops, i had enabled auto-merge. I'll follow up with a changeset

PR #6181, opened 2024-10-22
Adding code for extract subgraphs for set context.

clenfest:
A couple of things currently broken:
- the context argument on the join spec definition
- context_spec definition propagation
- some error condition todos


apollo-federation/src/supergraph/mod.rs
```diff
@@ -709,6 +713,21 @@ fn extract_object_type_content(
...
+        .and_then(|metadata| metadata.for_identity(&Identity::context_identity()))
+        .and_then(|context_link| CONTEXT_VERSIONS.find(&context_link.url.version))
+        .and_then(|context_spec_def| {
+            if let Some(name_in_schema) = context_spec_def
+                .context_directive_name_in_schema(supergraph_schema)
+                .ok()
+            {
+                return Some((context_spec_def, name_in_schema));
+            }
+            None
```
TylerBloom, line 727: Two thoughts here. If `context_directive_name_in_schema` fails, we are silencing that error. What could cause that to error? Second, IMO, mapping the output from `context_directive_name_in_schema` would be more readable:
```suggestion
            context_spec_def
                .context_directive_name_in_schema(supergraph_schema)
                .ok()
                .map(|name_in_schema| (context_spec_def, name_in_schema))
```
If we want to bubble up the error, just remove the `.ok()` then call `.transpose()?` after the last `.and_then` to invert the `Option<Result>` to a `Result<Option>` then bubbling up the error with the `?` operator.

apollo-federation/src/supergraph/mod.rs
```diff
@@ -747,6 +766,45 @@ fn extract_object_type_content(
             )?;
         }
 
+        if let Some((context_spec_def, name_in_supergraph)) = context {
+            for directive in type_.directives.get_all(name_in_supergraph.as_str()) {
+                if let Some(context_name) =
+                    FederationSpecDefinition::context_directive_arguments(directive).ok()
```
TylerBloom, line 772: Similar question as before. What kind of errors are being suppressed here?

apollo-federation/src/supergraph/mod.rs
```diff
@@ -747,6 +766,45 @@ fn extract_object_type_content(
...
+                                    context_in_subgraph.to_string(),
+                                )?)
+                                .into(),
+                            );
+                        } else {
+                            todo!();
+                        }
+                    } else {
+                        todo!();
+                    }
```
TylerBloom, line 803: Could you give an example of what it would look like to reach these else branches? It feels like these should be errors (like you mention in your comment), but I'm not sure what the error would be.

apollo-federation/src/supergraph/mod.rs
```diff
@@ -1474,6 +1533,35 @@ fn add_subgraph_field(
...
+                name,
+                type_,
+                context,
+                selection,
+            } = args;
+            let mut split = context.split("__");
+            split.next();
+            let Some(context_name_in_subgraph) = split.last() else {
+                todo!();
+            };
```
TylerBloom, line 1548: This is my mistake. There is an nicer way of expressing this via [`str::rsplit_once`](https://doc.rust-lang.org/std/primitive.str.html#method.rsplit_once), which returns `None` if the string doesn't contain the substring you're looking for.
```suggestion
            let Some((_, context_name_in_subgraph)) = context.rsplit_once() else {
                todo!();
            };
```

PR #6185, opened 2024-10-22
[JSONSelection] Remove `rest: * { a b }`  star selection syntax

benjamn:
This PR removes the under-documented JSONSelection `StarSelection` syntax.

The `StarSelection` syntax provided a way to capture all unselected properties of an input object, which could then be given a name by providing an alias (e.g. `rest: *`) or further sub-selected by appending a `SubSelection` (e.g. `rest: * { id }` to select just the `id`s).

Another use case for the `*` syntax was to make it easier to handle dynamic dictionaries in GraphQL, which does not provide a way to select dynamic object keys:
```graphql
booksByISBN: books { * { title author }}
```
Supposing `books` is a dictionary whose keys are ISBN strings and whose values are objects, this selection grabs the `title` and `author` properties from each of those objects while preserving the ISBN keys and dictionary structure.

Unfortunately, this syntax has interacted poorly with other JSONSelection functionality, including features we want to build. In particular, capturing only unselected fields with the `*` selection makes its behavior sensitive to the removal of other `NamedSelection` fields, which violates [guiding principle #3](https://github.com/apollographql/router/blob/64ed77c7cfb91959e794d930c5cc7dcef22ac562/apollo-federation/src/sources/connect/json_selection/README.md#guiding-principles).

Fortunately, we have new tools to address the use cases `StarSelection` syntax was originally designed to solve. In particular, we can use `->` methods to inspect available data and perform arbitrary processing of JSON dictionaries:
```graphql
# If you need a complete list of keys/values available from
# the current input object, the ->entries method can help
keys: $->entries.key
values: $->entries.value

# Imagine ->mapValues preserves object keys while binding
# each value to @. Note: this method is still hypothetical.
booksByISBN: books->mapValues(@ { title author })
```
Given these alternatives, I'm confident we can remove the special `*` syntax without eliminating any use cases, though this is definitely a breaking change for developers who were previously using the `*` syntax. Please comment below if this removal is disruptive to you.


lennyburdette (approved):
thanks for doing this! i'm a fan of you deleting my code :grin:

PR #6187, opened 2024-10-23
Limit the amount of GraphQL validation errors returned in the response

goto-bus-stop:
A small proposal.

It's possible to craft relatively small queries (within a few dozen KB) that produce tens of thousands of validations errors. The router then needs to spend a bunch of time formatting several megabytes of JSON in response. That's a waste of time.

By limiting the amount of errors we limit the response size and the time spent formatting the response to those queries.

I only applied it to GraphQL parsing and validation errors here. Perhaps it should be a general limit in the router. But for parsing and validation errors I think it's reasonable to silently drop further errors--I'm not sure it's appropriate for every other kind of error.

[ROUTER-822]: https://apollographql.atlassian.net/browse/ROUTER-822?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changesets/fix_renee_limit_errors.md
+++ b/.changesets/fix_renee_limit_errors.md
@@ -1,0 +1,6 @@
+### Limit the amount of GraphQL validation errors returned in the response ([PR #6187](https://github.com/apollographql/router/pull/6187))
+
+When an invalid query is submitted, the router now returns at most 100 GraphQL parsing and validation errors in the response.
+This prevents generating a very large response for nonsense documents.
+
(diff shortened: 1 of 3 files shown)


Geal (approved):

yanns:
This is a good idea. We did that in the sangria library as well for example:
- https://github.com/sangria-graphql/sangria/pull/1017
- https://github.com/sangria-graphql/sangria/pull/1034

PR #6188, opened 2024-10-23
[JSONSelection] Support conditional `...` fragment selections

benjamn:
In order to handle REST endpoints that return a union of possible result types, it's often necessary to consider various discriminator fields in the response, select different fields depending on those discriminator fields, and (when working with GraphQL) specify concrete `__typename` strings to resolve abstract types. This PR introduces a new JSONSelection syntax to address these use cases.

To keep the new syntax unambiguous, a conditional fragment is always introduced by a `...` token, followed by `if (expression) { included fields }`. Unconditional fragment spreads are not currently supported with this syntax. Optionally, additional `else if` and `else` clauses may follow.

In this example, we're using the runtime value of the `kind` field to determine which conditional fragment to apply:
```graphql
id
title
... if (kind->eq("book")) {
  isbn
  author { name }
} else if (kind->eq("movie")) {
  director { name }
} else {
  kind
}
year
```

When working with GraphQL, this type discrimination syntax provides an easy place to specify concrete `__typename` strings based on the `kind` values:
```graphql
id
title
... if (kind->eq("book")) {
  __typename: $("Book")
  isbn
  author { name }
} else if (kind->eq("movie")) {
  __typename: $("Movie")
  director { name }
} else {
  __typename: $("Unknown")
  kind
}
year
```

In the future, we may want to extend this syntax to allow named fragments (to reduce repetition across `selection` strings), as in GraphQL, but that seemed too ambitious for this initial PR.

### DIFF
--- a/apollo-federation/src/sources/connect/json_selection/apply_to.rs
+++ b/apollo-federation/src/sources/connect/json_selection/apply_to.rs
@@ -256,6 +256,7 @@ impl ApplyToInternal for NamedSelection {
                     ));
                 }
             }
+
             Self::Path(alias_opt, path_selection) => {
                 let (value_opt, apply_errors) =
                     path_selection.apply_to_path(data, vars, input_path);
@@ -308,6 +309,7 @@ impl ApplyToInternal for NamedSelection {
                     ));
                 }
             }
+
             Self::Group(alias, sub_selection) => {
                 let (value_opt, apply_errors) = sub_selection.apply_to_path(data, vars, input_path);
                 errors.extend(apply_errors);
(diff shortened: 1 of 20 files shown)


apollo-federation/src/sources/connect/json_selection/apply_to.rs
```diff
@@ -308,19 +309,67 @@ impl ApplyToInternal for NamedSelection {
...
             }
+
+            Self::Spread(spread) => {
+                let (spread_opt, spread_errors) = spread.apply_to_path(data, vars, input_path);
+                errors.extend(spread_errors);
+                if let Some(JSON::Object(spread)) = spread_opt {
+                    // TODO Better merge strategy for conflicting fields
+                    output.extend(spread);
+                }
+            }
```
benjamn, line 328: Calling out this `TODO` as a topic we've previously discussed, but that becomes more urgent with conditional fragments: how should we merge fields (potentially coming from different fragments, with different subselections) when they collide?

abernix:
Feel free to re-open this, but I'll close this for now for focusing the number of active PRs downward some.  Thanks!

PR #6190, opened 2024-10-24
Snapshot improvements for `FetchDependencyGraph` and open branches

duckki:
- `FetchDependencyGraph` snapshots are formatted in GraphViz dot format.
- Open branch snapshots are pretty printed in plain text format.

Those changes are in separate commits. It may be easier to review commit by commit.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [x] Manual Tests

### DIFF
--- a/apollo-federation/src/query_plan/fetch_dependency_graph.rs
+++ b/apollo-federation/src/query_plan/fetch_dependency_graph.rs
@@ -3443,7 +3443,11 @@ pub(crate) fn compute_nodes_for_tree(
             }
         }
     }
-    snapshot!(dependency_graph, "updated_dependency_graph");
+    snapshot!(
+        "FetchDependencyGraph",
+        dependency_graph.to_dot(),
+        "Fetch dependency graph updated by compute_nodes_for_tree"
+    );
     Ok(created_nodes)
 }
 
--- a/apollo-federation/src/query_plan/fetch_dependency_graph.rs
+++ b/apollo-federation/src/query_plan/fetch_dependency_graph.rs
@@ -3352,11 +3352,7 @@ pub(crate) fn compute_nodes_for_tree(
     initial_defer_context: DeferContext,
     initial_conditions: &OpGraphPathContext,
 ) -> Result<IndexSet<NodeIndex>, FederationError> {
-    snapshot!(
-        "OpPathTree",
-        serde_json_bytes::json!(initial_tree.to_string()).to_string(),
-        "path_tree"
-    );
+    snapshot!("OpPathTree", initial_tree.to_string(), "path_tree");
     let mut stack = vec![ComputeNodesStackItem {
         tree: initial_tree,
         node_id: initial_node_id,
--- a/apollo-federation/src/query_plan/query_planning_traversal.rs
+++ b/apollo-federation/src/query_plan/query_planning_traversal.rs
@@ -46,6 +46,7 @@ use crate::schema::position::CompositeTypeDefinitionPosition;
 use crate::schema::position::ObjectTypeDefinitionPosition;
 use crate::schema::position::SchemaRootDefinitionKind;
 use crate::schema::ValidFederationSchema;
+use crate::utils::logging::make_string;
 use crate::utils::logging::snapshot;
 
 // PORT_NOTE: Named `PlanningParameters` in the JS codebase, but there was no particular reason to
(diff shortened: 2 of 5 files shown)


apollo-federation/src/utils/logging.rs
```diff
@@ -50,3 +50,23 @@ macro_rules! snapshot {
 }
 
 pub(crate) use snapshot;
+
+#[allow(unreachable_pub)] // suppress warning: tracing utility
+pub fn make_string<T>(
```
duckki, line 55: I couldn't find a utility function like this on the web. But, I could've missed something better.

apollo-federation/src/query_plan/fetch_dependency_graph.rs
```diff
@@ -3443,7 +3443,11 @@ pub(crate) fn compute_nodes_for_tree(
             }
         }
     }
-    snapshot!(dependency_graph, "updated_dependency_graph");
+    snapshot!(
+        "FetchDependencyGraph",
+        dependency_graph.to_dot(),
+        "Fetch dependency graph updated by compute_nodes_for_tree"
+    );
```
TylerBloom, line 3450: I'm mostly leaving this as a note about what I've described before. In the viz tool, we pair the `FetchDependencyGraph` variant of the snapshot kind with visualizing the data with graph viz. I don't love the leaning into the already string-ness of this API. Perhaps this requires rethinking how the snapshot macro works.

Perhaps, we should have a series of macros that will emit logs with the correct data? For example, this could be replaced with a `graph_snapshot!`, which would add the correct `data` field to the log (which would note how to visualize it rather than using the type) and then call `.to_dot()` on the `dependency_graph`.

I'm not asking for this to change in the PR. More, I'd like to know your thoughts.

apollo-federation/src/query_plan/query_planning_traversal.rs
```diff
@@ -158,6 +159,29 @@ impl BestQueryPlanInfo {
     }
 }
 
+// PORT_NOTE: This is a (partial) port of `QueryPlanningTraversal.debugStack` JS method.
+fn format_open_branch(
+    f: &mut std::fmt::Formatter<'_>,
+    data: &(&Selection, &Vec<SimultaneousPathsWithLazyIndirectPaths>),
+) -> std::fmt::Result {
+    let selection = data.0;
+    let options = data.1;
```
TylerBloom, line 168: Nit:
```suggestion
    (selection, options): &(&Selection, &Vec<SimultaneousPathsWithLazyIndirectPaths>),
) -> std::fmt::Result {
```

apollo-federation/src/query_plan/query_planning_traversal.rs
```diff
@@ -158,6 +159,29 @@ impl BestQueryPlanInfo {
...
+    writeln!(f, "{selection}")?;
+    writeln!(f, " * Options:")?;
+    for option in options {
+        writeln!(f, "   - {option}")?;
+    }
+    Ok(())
+}
+
+#[allow(dead_code)] // suppress warning: tracing utility
+fn open_branch_to_string(
```
TylerBloom, line 178: This should probably be moved to `src/utils/logging.rs` and that whole module should probably be marked with `#![allow(dead_code)]`.

apollo-federation/src/query_plan/query_planning_traversal.rs
```diff
@@ -275,6 +299,35 @@ impl<'a: 'b, 'b> QueryPlanningTraversal<'a, 'b> {
         Ok(self.best_plan)
     }
 
+    // PORT_NOTE: This is a port of `QueryPlanningTraversal.debugStack` JS method.
+    #[allow(dead_code)] // suppress warning: tracing utility
+    fn branch_stack_to_string(&self) -> String {
```
TylerBloom, line 304: Similar idea here. Move to `logging.rs`.

TylerBloom (approved):
Looks good overall. Some minor thoughts and recommend moving potentially dead code/code for debugging into a common location.

apollo-federation/src/query_plan/fetch_dependency_graph.rs
duckki: Good point. String literals like `"FetchDependencyGraph "` are brittle and we need a better system. Maybe some structs (and a trait) to drive this snapshot generation and type tagging.

I've added a future direction section in our hackathon doc.

PR #6191, opened 2024-10-24
Add `extensions.service` for all subgraph errors

IvanGoncharov:
If `include_subgraph_errors` is `true` for a particular subgraph, all errors originating in this subgraph will have the subgraph's name exposed as a `service` extension.

For example, if subgraph errors are enabled, like so:
```yaml title="router.yaml"
include_subgraph_errors:
  all: true # Propagate errors from all subgraphs
```
Note: This option is enabled by default in the [dev mode](./configuration/overview#dev-mode-defaults).

And this `products` subgraph returns an error, it will have a `service` extension:

```json
{
  "data": null,
  "errors": [
    {
      "message": "Invalid product ID",
      "path": [],
      "extensions": {
        "service": "products"
      }
    }
  ]
}
```

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [x] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/plugins/include_subgraph_errors.rs
```diff
@@ -42,36 +42,47 @@ impl Plugin for IncludeSubgraphErrors {
     }
 
     fn subgraph_service(&self, name: &str, service: subgraph::BoxService) -> subgraph::BoxService {
```
IvanGoncharov, line 44: `include_subgraph_errors.rs` is the best place I could find to add this functionality.

Samjin:
It maybe a vulnerable to expose backend service name to public clients for some users. Is it possible to have error extensions allowlist to control what properties should expose to clients? 

nit: I noticed `service` is used instead of `subgraph_name` or `subgraph`. Is this consistent naming from Apollo? If so we'll use service as the name across telemetry and also in resp.

IvanGoncharov:
> It maybe vulnerable to expose backend service name to public clients for some users. Is it possible to have error extensions allowlist/blacklist to control what properties should expose to clients?

@Samjin Errors exposed only if `include_subgraph_errors` is set.
Nothing is leaked by default, see docs: https://www.apollographql.com/docs/graphos/routing/observability/subgraph-error-inclusion


> nit: I noticed service is used instead of subgraph_name or subgraph. Is this consistent naming from Apollo? If so we'll use service as the name across telemetry and also in resp.

Yes, it is consistent with existing subgraphs errors, for example:
https://github.com/apollographql/router/blob/382fbe0a3009294374ebd62d82d8b61569ec8e55/apollo-router/src/error.rs#L147-L149
This PR just make this field consistently present for all errors related to particular subgraph.

yanns:
When subgraphs return a user-friendly errors that we want to propagate, then we need to set `include_subgraph_errors` to `true`. But now we're exposing the service name that is an internal information.
Is there a way to use the user-friendly errors from subgraphs without exposing internal information?

GwlCtovic:
> But now we're exposing the service name that is an internal information.

@yanns The service name propagation through subgraph errors was the default behavior in`@apollo/gateway` if I'm not mistaken.

As a matter of fact, I'm watching this issue because migrating from `@apollo/gateway` to the `router` broke some of our monitoring tools, as they rely on subgraph automatically reporting their `service`

Samjin:
Relate to proposal PR https://github.com/apollographql/router/pull/6344

PR #6192, opened 2024-10-24
Removes two unnecessary warn logs from the ResponseVisitor

tninesling:
When the ResponseVisitor attempts to match a JSON response to its corresponding query, it is possible that the subgraph did not return a requested field. In this case, the ResponseVisitor was creating a noisy warning log which can be run into quite frequently.

This removes those two logs since we will ignore the mismatch anyway, and demand control should not be responsible for policing the format of subgraph responses.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


goto-bus-stop (approved):
LGTM, maybe simplify the changeset a bit? It's perfect as a PR description but users don't/shouldn't know what a `ResponseVisitor` is, so I'd just mention something along the lines that demand control no longer produces warnings when a subgraph did not return a value for a requested field.

PR #6194, opened 2024-10-24
 Add `supergraph.experimental_http1_max_headers` option

IvanGoncharov:
You can use it like this:
```yaml
supergraph:
  experimental_http1_max_headers: 200
```

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -38,18 +38,18 @@ dependencies = [
 
 [[package]]
 name = "addr2line"
-version = "0.22.0"
+version = "0.24.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6e4503c46a5c0c7844e948c9a4d6acd9f50cccb4de1c48eb9e291ea17470c678"
+checksum = "dfbe277e56a376000877090da837660b4427aad530e3028d44e0bffe4f89a1c1"
 dependencies = [
  "gimli",
 ]
 
 [[package]]
-name = "adler"
-version = "1.0.2"
+name = "adler2"
+version = "2.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
+checksum = "512761e0bb2578dd7380c6baaa0f4ce03e84f95e960231d1dec8bf4d7d6e2627"
 
 [[package]]
 name = "ahash"
@@ -104,9 +104,9 @@ checksum = "4b46cbb362ab8752921c97e041f5e366ee6297bd428a31275b9fcf1e380f7299"
 
 [[package]]
 name = "anstream"
-version = "0.6.14"
+version = "0.6.17"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "418c75fa768af9c03be99d17643f93f79bbba589895012a80e3452a19ddda15b"
+checksum = "23a1e53f0f5d86382dafe1cf314783b2044280f406e7e1506368220ad11b1338"
 dependencies = [
  "anstyle",
  "anstyle-parse",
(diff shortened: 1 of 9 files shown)


Cargo.toml
```diff
@@ -76,3 +76,6 @@ sha1 = "0.10.6"
 tempfile = "3.10.1"
 tokio = { version = "1.36.0", features = ["full"] }
 tower = { version = "0.4.13", features = ["full"] }
+
+[patch.crates-io]
+"hyper" = { git = "https://github.com/IvanGoncharov/hyper.git", branch = "backpor_header_size" }
```
abernix, line 81:
```suggestion
"hyper" = { git = "https://github.com/apollographql/hyper.git", tag = "header-customizations-20241108" }
```

apollo-router/src/axum_factory/axum_http_server_factory.rs
```diff
@@ -298,12 +300,22 @@ impl HttpServerFactory for AxumHttpServerFactory {
             let actual_main_listen_address = main_listener
                 .local_addr()
                 .map_err(ApolloRouterError::ServerCreationError)?;
+            let mut http_config = Http::new();
+            http_config.http1_keep_alive(true);
```
Geal, line 304: this already defaults to true

Geal (approved):
LGTM (also looked at https://github.com/apollographql/hyper/commit/c42aec785394b40645a283384838b856beace011 and it looks ok too

Cargo.lock
```diff
@@ -38,18 +38,18 @@ dependencies = [
 
 [[package]]
 name = "addr2line"
-version = "0.22.0"
+version = "0.24.2"
```
SimonSapin, line 41: There's 1000 lines of diff in this file. Did you run `cargo update` without `-p` or is this all a consequence of bumping the hyper requirement to `0.14.31`?

Cargo.toml
```diff
@@ -76,3 +76,6 @@ sha1 = "0.10.6"
 tempfile = "3.10.1"
 tokio = { version = "1.36.0", features = ["full"] }
 tower = { version = "0.4.13", features = ["full"] }
+
+[patch.crates-io]
+"hyper" = { git = "https://github.com/apollographql/hyper.git", tag = "header-customizations-20241108" }
```
SimonSapin, line 81: We do this to unblock in the short term, but is there a path forward for having this functionality upstream?

apollo-router/Cargo.toml
```diff
@@ -52,6 +52,8 @@ docs_rs = ["router-bridge/docs_rs"]
 # and not yet ready for production use.
 telemetry_next = []
 
+experimental_hyper_fork = []
```
SimonSapin, line 55: Could this have a more descriptive name like `experimental_hyper_header_limits`?

docs/source/reference/router/configuration.mdx
```diff
@@ -1145,6 +1147,17 @@ Before increasing this limit significantly consider testing performance
...
+
+Limit the maximum number of headers of incoming HTTP1 requests.
+The default value is 100 headers.
+
+If router receives more headers than the buffer size, it responds to the client with `431 Request Header Fields Too Large`.
+
+##### `experimental_http1_max_request_buf_size`
+
+Limit the maximum buffer size for the HTTP1 connection. Default is ~400kb.
+
```
SimonSapin, line 1160: If we document this feature at all, I think we should also document that custom binary users need to copy the `[patch.crates-io]` section in their own `Cargo.toml`

docs/source/routing/security/request-limits.mdx
```diff
@@ -273,6 +275,17 @@ Before increasing this limit significantly consider testing performance
...
+
+Limit the maximum number of headers of incoming HTTP1 requests.
+The default value is 100 headers.
+
+If router receives more headers than the buffer size, it responds to the client with `431 Request Header Fields Too Large`.
+
+### `experimental_http1_max_request_buf_size`
+
+Limit the maximum buffer size for the HTTP1 connection. Default is ~400kb.
+
```
SimonSapin, line 288: Should one of these two files link to the other instead of duplicate content?

Cargo.toml
IvanGoncharov: yes, PR to upstream https://github.com/hyperium/hyper/pull/3773

docs/source/routing/security/request-limits.mdx
IvanGoncharov: I followed the lead of `http_max_request_bytes` and it's duplicated in those to places.
So I just keep it sync with this PR, but definitely something that need to be improved.

SimonSapin (approved):

Cargo.lock
IvanGoncharov: good catch, I probably did it without `-p`

.changesets/feat_max_headers.md
```diff
@@ -0,0 +1,12 @@
+### Experimental Configuration Options for HTTP/1 Max Headers and Buffer Limits ([PR #6194](https://github.com/apollographql/router/pull/6194))
```
abernix, line 1: My feedback on this block:

* This changelog can be renamed to be prefixed with `exp_` (instead of `feat_`; see [docs](https://github.com/apollographql/router-private/blob/17e88fac8f0184f90d8757183841920b85203673/.changesets/README.md?plain=1#L33)) and it will automatically show up in an "Experimental" section.  I think that's what it's intended for.
* This changelog is excellent in what it says, but I would _also_ update it to specify **WHY** the feature is marked as experimental (if my understanding is correct, it's not because it's unstable, but because we are waiting on the upstream specificity of whether option ultimately applies to HTTP v1 or HTTP/2, is that, correct?)
* I think in the docs we should say, "If you are building a custom binary, you will need to use a custom `cargo patch` entry."

Thoughts?

IvanGoncharov: All points are correct 👍

docs/source/reference/router/configuration.mdx
IvanGoncharov: Added note to users of crate

abernix (approved):

PR #6203, opened 2024-10-28
Add fields referenced only in entity arguments to split subgraphs

dylan-apollo:
Technically also allows some interface stuff that won't work, but it's caught in validation! Based on #6201 which should be merged first.

[CNN-485]: https://apollographql.atlassian.net/browse/CNN-485?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apollo-federation/src/sources/connect/expand/tests/schemas/expand/keys.yaml
+++ b/apollo-federation/src/sources/connect/expand/tests/schemas/expand/keys.yaml
@@ -4,25 +4,34 @@ subgraphs:
     schema:
       sdl: |
         extend schema
-          @link(url: "https://specs.apollo.dev/federation/v2.8", import: ["@key"])
+          @link(url: "https://specs.apollo.dev/federation/v2.10", import: ["@key"])
           @link(url: "https://specs.apollo.dev/connect/v0.1", import: ["@connect"])
         type Query {
           t(id: ID!): T
             @connect(                                                                                                   # expect `key: "id"`
               http: { GET: "http://localhost/ts/{$$args.id}" }
-              selection: "id id2"
+              selection: "id id2 unselected"
               entity: true
             )
-          t2(id: ID! id2: ID!): T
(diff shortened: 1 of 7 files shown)


apollo-federation/src/sources/connect/expand/tests/schemas/expand/keys.yaml
```diff
@@ -1,30 +1,37 @@
...
-          t2(id: ID! id2: ID!): T
+          t2(id: ID!, id2: ID!): T
             @connect(                                                                                                   # expect `key: "id id2"`
               http: { GET: "http://localhost/ts/{$$args.id}?id2={$$args.id2}" }
-              selection: "id id2"
+              selection: "id id2 unselected"
+              entity: true
+            )
+          """ Uses the `unselected` field as a key, but doesn't select it """
+          unselected(unselected: ID!): T
```
lennyburdette, line 23: this is interesting! it makes sense for simple entity resolvers, but i think we'll have to have different rules for batch entity resolvers because we'll need to match the keys in the response to the entity references

lennyburdette (approved):
nice!

apollo-federation/src/sources/connect/expand/tests/schemas/expand/keys.yaml
dylan-apollo: There might also be some REST APIs where the order is guaranteed to be the same, so we'll probably want a mode for that as well

lennyburdette: totally. i don't think that's something we can default to (i've yet to hear about one), but maybe we allow opting into that. i've been thinking that the `batch` argument is an object to support multiple styles of configuration, so a `batch: { assumeOrdered: true }` or something makes sense.

PR #6205, opened 2024-10-29
Fix the query hashing algorithm

Geal:
This splits part of the work from https://github.com/apollographql/router/pull/5255 to make it easier to merge. This PR only covers the fixes for the query hashing algorithm, which is currently used in entity caching, without integrating the changes to the query planner cache key.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-853]: https://apollographql.atlassian.net/browse/ROUTER-853?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Geal:
@IvanGoncharov here's the PR that contains only the query hashing algorithm fixes

apollo-router/src/spec/query/change.rs
```diff
@@ -1,16 +1,58 @@
...
+//! ### Prefixes and suffixes
+//! Across the entire visitor, we add prefixes and suffixes like this:
+//!
+//! ```rust
+//! "^SCHEMA".hash(self);
+//! ```
+//!
+//! This prevents possible collision while hashing multiple things in a sequence. The
+//! `^` character cannot be present in a GraphQL query or schema outside of comments
+//! or strings, so this is a good separator.
```
IvanGoncharov, line 42: Even if `^` were to be used in some new syntax added to the GraphQL, it would still be safe since we hash only names.
```suggestion
//! This prevents possible collision while hashing multiple things in a sequence. The
//! `^` character cannot be present in GraphQL names, so this is a good separator.
```

apollo-router/src/spec/query/change.rs
```diff
@@ -73,7 +112,23 @@ impl<'a> QueryHashVisitor<'a> {
...
                 JOIN_TYPE_DIRECTIVE_NAME,
             ),
+        };
+
+        visitor.hash_schema()?;
+
+        Ok(visitor)
+    }
+
+    pub(crate) fn hash_schema(&mut self) -> Result<(), BoxError> {
```
IvanGoncharov, line 122: As an idea for follow up PR: What if use struct destructuring (without `...`) for accessing fields on `schema`? 
In that way, we can ensure that all fields are considered for hashing, so if new fields are added to `apollo-rs`, they will result in compiler errors.
That way, we protect against stuff that would be added to the GraphQL spec in the future, e.g., support for directives in a new location.

apollo-router/src/spec/query/change.rs
```diff
@@ -132,158 +214,268 @@ impl<'a> QueryHashVisitor<'a> {
                 b.hash(self);
```
IvanGoncharov, line 214: I think it worth to add `^` to prefixes here, even just for consistency.
`^enum`
`^variable`
`^string`
etc

IvanGoncharov: It will also make debugging easier.
The last time I debugged this code, I added `println` to print all hashed values on separate lines.
So having `^` would make it easier to navigate

apollo-router/src/spec/query/change.rs
```diff
@@ -132,158 +214,268 @@ impl<'a> QueryHashVisitor<'a> {
...
+                    "^IMPLEMENTER_OBJECT_LIST_END".hash(self);
+
+                    "^IMPLEMENTER_INTERFACE_LIST".hash(self);
+                    for interface in &implementers.interfaces {
+                        self.hash_type_by_name(interface)?;
+                    }
+                    "^IMPLEMENTER_INTERFACE_LIST_END".hash(self);
+                }
+
+                "^INTERFACE_END".hash(self);
```
IvanGoncharov, line 323: Interface has end, but other types doesn't
For example there are no `^OBJECT_END` or `^UNION_END`
I don't think it's exploitable but I think we need consistency here, so better to add end "keywords" everywhere

apollo-router/src/spec/query/change.rs
```diff
@@ -132,158 +214,268 @@ impl<'a> QueryHashVisitor<'a> {
...
                     self.hash_directive(&directive.node);
                 }
+                "^DIRECTIVE_LIST_END".hash(self);
 
+                "^ENUM_VALUE_LIST".hash(self);
                 for (value, def) in &e.values {
+                    "^VALUE".hash(self);
+
                     value.hash(self);
                     for directive in &def.directives {
```
IvanGoncharov, line 354: All other directives lists are wrapped in `^DIRECTIVE_LIST`.
No exploitable, just consistency.
Maybe it worth to create a function to hash directive lists?

apollo-router/src/spec/query/change.rs
```diff
@@ -132,158 +214,268 @@ impl<'a> QueryHashVisitor<'a> {
...
         &mut self,
         parent_type: String,
-        type_name: String,
         field_def: &FieldDefinition,
-        arguments: &[Node<Argument>],
+        node: &executable::Field,
     ) -> Result<(), BoxError> {
-        if self.hashed_fields.insert((parent_type.clone(), type_name)) {
-            self.hash_type_by_name(&parent_type)?;
+        "^FIELD".hash(self);
```
IvanGoncharov, line 414: I think it makes sense to hash field (parent type + field name) only once, similar to
https://github.com/apollographql/router/blob/0f8d90dace721e749fa3224579eee9ceda0e5aa4/apollo-router/src/spec/query/change.rs#L243-L248

Without it maybe be possible to send some expensive fields (e.g., a lot of arguments) bunch of times (intentionally or not).

apollo-router/src/spec/query/change.rs
```diff
@@ -369,30 +618,47 @@ impl<'a> Visitor for QueryHashVisitor<'a> {
         field_def: &ast::FieldDefinition,
         node: &executable::Field,
     ) -> Result<(), BoxError> {
+        "^VISIT_FIELD".hash(self);
```
IvanGoncharov, line 621: there is no matching end, like `^VISIT_OPERATION-END`

apollo-router/src/spec/query/change.rs
Geal: it could be useful for the other structures we use elsewhere in the visitor. Here, the schema also holds a list of types, but the point of this algorithm would be to ignore irrelevant types

apollo-router/src/spec/query/change.rs
Geal: fixed in 350fde0503c14f91eebf89f67f95f8f3b308c09a

apollo-router/src/spec/query/change.rs
```diff
@@ -401,18 +667,35 @@ impl<'a> Visitor for QueryHashVisitor<'a> {
...
     ) -> Result<(), BoxError> {
+        "^VISIT_INLINE_FRAGMENT".hash(self);
+
         if let Some(type_condition) = &node.type_condition {
             self.hash_type_by_name(type_condition)?;
         }
-        traverse::inline_fragment(self, parent_type, node)
+        for directive in &node.directives {
+            self.hash_directive(directive);
+        }
```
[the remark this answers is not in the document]
Geal: fixed in 350fde0503c14f91eebf89f67f95f8f3b308c09a

apollo-router/src/spec/query/change.rs
Geal: this was there initially but was part of the issue, because if a field was already seen once, it would not hash again both the field definition and the field usage in query, so if a field was present twice with different arguments, it would be incorrect. I think we can reintroduce it safely now

Geal: done in 1598fbfa5, WDYT?

IvanGoncharov: @Geal 👍

apollo-router/src/spec/query/change.rs
```diff
@@ -92,198 +149,335 @@ impl<'a> QueryHashVisitor<'a> {
...
+
+        // for every field, we also need to look at fields defined in `@requires` because
+        // they will affect the query plan
+        self.hash_join_field(parent_type, &field_def.directives)?;
+
+        self.hash_directive_list_ast(&field_def.directives);
+
+        "^ARGUMENT_DEF_LIST".hash(self);
+        for argument in &field_def.arguments {
+            self.hash_input_value_definition(argument)?;
```
IvanGoncharov, line 451: I'm not 100% sure, but it looks like it also should hash `@fromContext`:
https://www.apollographql.com/docs/graphos/reference/federation/directives#fromcontext
I looked into `apollo-federation,` and it seems that Rust QP doesn't support it but it is implemented only in JS QP.
I don't think this directive will affect entity cache because you also hashing selection set of subgraphql query (I'm just assuming here)

However, this directive should affect QP (at least for the JS planner), breaking QP caching.

IvanGoncharov (changes requested):
@Geal I triple checked this PR and it looks great 👍 
The only thing I found is `@fromContext`, otherwise it is ready to be merged.

If `@fromContext` is not an issue, then just ping me on Slack and I will approve this one ASAP.

Geal:
@IvanGoncharov I added the initial support for the context directives. I will take another look tomorrow, but I think the approach is sound:
- as we go through the query, whenever we encounter a `@context` directive, we record the type in a map `context name -> [Type}`
- when we encounter a field with the `join__field` directive and its `contextArguments` argument, then for each argument, we take the context name and selection
- we look up the types linked to that context name. Considering the way we go through the query and relevant types, it might return types that would not appear in this query, so it might be a bit stricter than necessary
- for each of those type, we then use the selection and hash that in the same way we handle `@key` and `@requires`

There are some assumptions in there that we might want to check, like what happens when a `contextArgument` refers to a context that we did not find in the query (maybe that would have been prevented at composition?)

IvanGoncharov:
Converting to draft due to internal discussion regarding the risks introduced by this PR.
We want to add "Query hash stable to incremental schema changes" as the Router's feature.
However, in its current form, it is implemented outside of Query Planner and without any mechanism to determine what parts of the schema are used by Query Planner.

For example, adding new directives that influence Query planning without updating this code can result in a hash collision where a schema update changes the query plan, but the hash remains the same.

IvanGoncharov (approved):

.changesets/fix_fix_query_hashing.md
bnjjj, line 1: Should we add the famous `caution` saying it could remove all the cache entries ?

.changesets/fix_fix_query_hashing.md
```diff
@@ -0,0 +1,5 @@
+### Fix the query hashing algorithm ([PR #6205](https://github.com/apollographql/router/pull/6205))
+
+The Router includes a schema-aware query hashing algorithm designed to return the same hash across schema updates if the query remains unaffected. This update enhances the algorithm by addressing various corner cases, improving its reliability and consistency.
+
+By [@Geal](https://github.com/Geal) in https://github.com/apollographql/router/pull/6205kkk
```
abernix, line 5: There is some `vim`-isms here at the end.

```suggestion
By [@Geal](https://github.com/Geal) in https://github.com/apollographql/router/pull/6205
```

IvanGoncharov: Vim :)

.changesets/fix_fix_query_hashing.md
IvanGoncharov: Nice catch!

PR #6108, opened 2024-10-02
Progressive override: fix query planner cache warmup

Geal:
The query planner needs various metadata along with the query to influence planning, and due to historical reasons that data was passed through the request context, not the planner request. The progressive override labels list was communicated like that, but missed the case of query plan cache warmup, where upon schema or configuration updates, the router takes the list of most used queries from the query planner's in memory cache, and plans them again before activating the new schema or configuration. Due to the mistake, the labels were not transmitted during warmup, which resulted in queries correctly using the overridden fields, but after an update, would revert to non overridden fields, and could not recover (unless the plan was evicted from the cache).

This fices the cache warmup issue, and there is a follow up PR to prevent further issues of that kind: https://github.com/apollographql/router/pull/6208

---

**Checklist**

Complete the checklist (and note appropriate exceptions) before the PR is marked ready-for-review.

- [ ] Changes are compatible[^1]
- [ ] Documentation[^2] completed
- [ ] Performance impact assessed and acceptable
- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[^1]: It may be appropriate to bring upcoming changes to the attention of other (impacted) groups. Please endeavour to do this before seeking PR approval. The mechanism for doing this will vary considerably, so use your judgement as to how and when to do this.
[^2]: Configuration is an important part of many changes. Where applicable please try to document configuration examples.
[^3]: Tick whichever testing boxes are applicable. If you are adding Manual Tests, please document the manual testing (extensively) in the Exceptions.

PR #3298, opened 2023-06-21
add labels to deferred fragments when they are missing

Geal:
We currently use a path and subselection to identify deferred fragments and format individual responses. Unfortunately, with query filtering, matching those fragments between the original and filtered query is hard because they can have a different shape, and thus a different subselection.
As an alternative solution, we can add a label to every deferred fragment. The label argument in the `@defer` directive is optional, but unique across the entire query. Some clients use it for that exact purpose, identifying each deferred response.
So the idea here is to add labels when they are not present, at the query plan cache level,use all the labels to identify the fragments, and then in the execution service, remove the extraneous labels from the responses sent to the client

**Checklist**

Complete the checklist (and note appropriate exceptions) before a final PR is raised.

- [ ] Changes are compatible[^1]
- [ ] Documentation[^2] completed
- [ ] Performance impact assessed and acceptable
- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[^1]. It may be appropriate to bring upcoming changes to the attention of other (impacted) groups. Please endeavour to do this before seeking PR approval. The mechanism for doing this will vary considerably, so use your judgement as to how and when to do this.
[^2]. Configuration is an important part of many changes. Where applicable please try to document configuration examples.
[^3]. Tick whichever testing boxes are applicable. If you are adding Manual Tests:
    - please document the manual testing (extensively) in the Exceptions.
    - please raise a separate issue to automate the test and label it (or ask for it to be labeled) as `manual test`

PR #6208, opened 2024-10-29
Pass parameters explicitely in QueryPlannerRequest instead of context

Geal:
Follow up to https://github.com/apollographql/router/pull/6108/ to pass query planner arguments directly in the request struct instead of the context

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/query_planner/caching_query_planner.rs
```diff
@@ -326,20 +322,12 @@ where
                     }
                 };
 
-                let schema = self.schema.api_schema();
-                if let Ok(modified_query) = add_defer_labels(schema, &doc.ast) {
```
Geal, line 330: this is already called inside BridgeQueryPlanner

apollo-router/src/query_planner/bridge_query_planner.rs
```diff
@@ -635,19 +627,9 @@ impl Service<QueryPlannerRequest> for BridgeQueryPlanner {
                         operation_name.as_deref(),
                         Arc::new(QueryHash(hash)),
                     )?;
-                    context
```
Geal, line 638: I don't think that one was correct (and removing this is probably what triggered the snapshot changes). The defer labels modification was added in https://github.com/apollographql/router/pull/3298 to be able to track which deferred response was linked to which fragment in the query, but this is a modification that should not be apparent anywhere else than in the planner and execution service. If we end up exposing the `ApolloCompiler` instance to user plugins, it should not contain those labels

Geal:
it will conflict with the fix in https://github.com/apollographql/router/pull/6108 because it proposes another way to fix it, but that will be easy to resolve

goto-bus-stop (approved):
lovely

apollo-router/src/query_planner/caching_query_planner.rs
Geal: we could definitely rely on that. This is used when generating the HTTP response, to know which graphql error has to go in which deferred response (especially when there are multiple deferred fragments at the same level) https://github.com/apollographql/router/blob/ffec3f902bda97912929d8df9e0a447c936ac2ae/apollo-router/src/services/execution/service.rs#L376-L380

goto-bus-stop: I've filed a followup to do that as ROUTER-850

PR #6211, opened 2024-10-30
reorganize docs for new docs site

shorgi:
Currently, a router SME cannot at a glance determine which source file to update for a published docs page, because the router content of the new docs site is sourced from both the router repo and the new docs repo.

This PR reorganizes router docs into directories that mirror the paths of the new docs site. It also moves router-specific pages from the new docs repo into the router repo, so all router-specific pages are sourced from here.

Merging these changes into `main`, `dev`, and `next` will be in phases:
1. Merge into `main-docs-replatform`, the branch that the new docs currently sources its router content.
2. Merge `main-docs-replatform` to `main` and update the new docs to source its router content from `main`.
3. Merge back `main` to `dev` and `next`


docs/source/config.json
```diff
@@ -2,130 +2,16 @@
   "title": "Self-Hosted Router",
```
Meschreiber, line 2: I'm a little confused about the presence of this file since it powered the sidebar on the old docs--do we need it at all?

Meschreiber (approved):

docs/source/config.json
shorgi: Summary of offline our conversation: CI currently uses config.json to check for the existence of its listed files, so needed to remove the outdated files from it. Still keeping the file around as a starting point to discuss how we can/should create a similar "manifest" of all router docs files.

PR #6215, opened 2024-10-31
add list of subgraphs/sources to context

lennyburdette:
this allows quickly understanding which connector sources are used in a query plan. you can then use a coprocessor to fetch auth tokens, add them to the context, and use them with `headers: [{ name: "Authorization", value: "Bearer {$context.token}" }]`

 

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/plugins/connectors/plugin.rs
+++ b/apollo-router/src/plugins/connectors/plugin.rs
@@ -25,6 +25,7 @@ use crate::plugin::PluginInit;
 use crate::plugins::connectors::configuration::ConnectorsConfig;
 use crate::plugins::connectors::request_limit::RequestLimits;
 use crate::register_plugin;
+use crate::services::connector_service::ConnectorSourceRef;
 use crate::services::execution;
 use crate::services::router::body::RouterBody;
 use crate::services::supergraph;
@@ -164,23 +165,11 @@ impl Plugin for Connectors {
                     .service_usage()
                     .unique()
                     .flat_map(|service_name| {
-                        let Some(connector) = connectors.get(service_name) else {
-                            return None;
-                        };
-
-                        let Some(ref source_name) = connector.id.source_name else {
-                            return None;
-                        };
-
-                        Some((connector.id.subgraph_name.clone(), source_name.clone()))
+                        connectors
+                            .get(service_name)
+                            .map(|connector| ConnectorSourceRef::try_from(connector).ok())
                     })
                     .unique()
-                    .map(|(subgraph_name, source_name)| {
-                        json!({
-                            "subgraph_name": subgraph_name,
-                            "source_name": source_name,
-                        })
-                    })
                     .collect_vec();
 
                 req.context
--- a/apollo-router/src/plugins/connectors/plugin.rs
+++ b/apollo-router/src/plugins/connectors/plugin.rs
@@ -15,14 +15,18 @@ use serde::Deserialize;
 use serde::Serialize;
 use serde_json_bytes::json;
 use tower::BoxError;
+use tower::ServiceBuilder;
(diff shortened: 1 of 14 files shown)


apollo-router/src/plugins/connectors/plugin.rs
```diff
@@ -143,6 +148,49 @@ impl Plugin for Connectors {
...
+
+                        Some((connector.id.subgraph_name.clone(), source_name.clone()))
+                    })
+                    .unique()
+                    .map(|(subgraph_name, source_name)| {
+                        json!({
+                            "subgraph_name": subgraph_name,
+                            "source_name": source_name,
+                        })
+                    })
```
pubmodmatt, line 183: I previously added a `ConnectorSourceRef` type that is exactly this tuple. This should let you simplify this to:

```rust
                   .flat_map(|service_name| {
                        connectors
                            .get(service_name)
                            .map(|connector| ConnectorSourceRef::try_from(connector).ok())
                    })
                   .unique()
```

You'd need to derive `Serialize` and `Deserialize` for `ConnectorSourceRef`, and add something like the following:

```rust
impl TryFrom<&Connector> for ConnectorSourceRef {
    type Error = ();

    fn try_from(value: &Connector) -> Result<Self, Self::Error> {
        Ok(Self {
            subgraph_name: value.id.subgraph_name.to_string(),
            source_name: value.id.source_name.clone().ok_or(())?,
        })
    }
}
```

apollo-router/src/query_planner/plan.rs
```diff
@@ -468,7 +468,6 @@ impl PlanNode {
         Ok(())
     }
 
-    #[cfg(test)]
     /// Retrieves all the services used across all plan nodes.
```
pubmodmatt, line 471: This was just test code before - are we comfortable making that production code? It does a lot of recursion and heap allocation, which could be slow on very large query plans.

pubmodmatt (changes requested):

apollo-router/src/query_planner/plan.rs
lennyburdette: Thanks for the review! I’m realizing now that we should probably do this differently:

* having it on the context means it’s always created even if unused, and will be impossible to remove
* it should probably be a field on the coprocessor request body so that it’s opt-in and lazily evaluated
* It should probably be generic for subgraphs and subgraph/sources, so that you can fetch auth tokens for any fetch target 

I’ll refactor, and I’ll take a closer look at this function to see if there’s a more efficient way to get the service names from the plan

apollo-router/src/query_planner/execution.rs
```diff
@@ -101,6 +101,10 @@ impl QueryPlan {
     pub fn subgraph_fetches(&self) -> usize {
         self.root.subgraph_fetches()
     }
+
+    pub fn service_usage<'a>(&'a self) -> Box<dyn Iterator<Item = &'a str> + 'a> {
```
lennyburdette, line 105:
```suggestion
    pub(crate) fn service_usage<'a>(&'a self) -> Box<dyn Iterator<Item = &'a str> + 'a> {
```

apollo-router/src/query_planner/plan.rs
lennyburdette: after discussion in slack, i went back to the context-based approach but added a config option to opt into this behavior. (i also implemented a non-recursive version of this method that allocates only a set and a vec.)

pubmodmatt (approved):

PR #6220, opened 2024-11-01
Allow mixing implicit and explicit connectors keys

dylan-apollo:
Using `entity: true` to automatically generate a key should always be valid, even if you _also_ have some explicit resolvable keys.

In order of the commits, this:

1. Reorganizes the entity tests, since the flat list of snapshots was getting hard to navigate
2. Simplifies the existing entity arg validation code
3. Removes the validation which was enforcing that if any `@key` is explicitly set, all `entity: true` connectors match one of those `@key`s

We did lose the ability to match automatic keys to manual ones, which we'll need to put back in a follow-up. At that same time, we'll want to enforce a _different_ rule: all resolvable keys need to have a resolving connector. I'm hoping we can use the expansion code for this so we don't have competing definitions of automatic keys.


apollo-federation/src/sources/connect/validation/entity.rs
```diff
@@ -15,125 +13,141 @@ use apollo_compiler::Node;
...
         .arguments
         .iter()
         .find(|arg| arg.name == CONNECT_ENTITY_ARGUMENT_NAME)
+    else {
+        return Ok(());
+    };
+    let entity_arg_value = &entity_arg.value;
+    if !entity_arg_value
+        .to_bool()
+        .is_some_and(|entity_arg_value| entity_arg_value)
```
lennyburdette, line 46:
```suggestion
        .unwrap_or_default()
```
optional nit

apollo-federation/src/sources/connect/validation/entity.rs
```diff
@@ -15,125 +13,141 @@ use apollo_compiler::Node;
...
+                message: format!(
+                    "{coordinate} is invalid. Entity resolvers can only be declared on root `Query` fields.",
+                    coordinate = connect_directive_entity_argument_coordinate(connect_directive_name, entity_arg_value.as_ref(), object, &field.name)
+                ),
+                locations: entity_arg.line_column_range(&schema.sources)
+                    .into_iter()
+                    .collect(),
             }
+        );
+        // TODO: Allow interfaces
```
lennyburdette, line 65: is this comment misplaced?

apollo-federation/src/sources/connect/validation/entity.rs
```diff
@@ -15,125 +13,141 @@ use apollo_compiler::Node;
...
-                    if let Some(message) = (ArgumentVisitor {
-                        schema,
-                        entity_arg,
-                        entity_arg_value,
+    if field.ty.is_list() || field.ty.is_non_null() {
+        return Err(
+            Message {
+                code: Code::EntityTypeInvalid,
+                message: format!(
+                    "{coordinate} is invalid. Entity connectors must return non-list, nullable, object types.",
```
lennyburdette, line 92:
```suggestion
                    "{coordinate} is invalid. Entity connectors must return non-list, nullable, object types. See https://go.apollo.dev/connectors/directives/#rules-for-entity-true",
```

lennyburdette (approved):

apollo-federation/src/sources/connect/validation/entity.rs
dylan-apollo: The comment was there originally, I just left it. TBH I'm not sure what interfaces have to do with being on the root query type, though...

```suggestion
```

PR #6223, opened 2024-11-04
Added logic to parse `@context` in `FederatedQueryGraphBuilder::build`

TylerBloom:
There are still a few TODOs. The regular expression that is used will fail to build because looking around is not supported. The logic inside the `simpleTraversal` for the context has also not been ported. Lastly, the definition position types might be off and need to be tested.

### DIFF
--- a/apollo-federation/src/query_graph/build_query_graph.rs
+++ b/apollo-federation/src/query_graph/build_query_graph.rs
@@ -1026,7 +1026,7 @@ impl FederatedQueryGraphBuilder {
             .root_kinds_to_nodes_by_source
             .iter()
             .filter(|(source, _)| **source != self.base.query_graph.current_source)
-            .flat_map(|(_, root_kind_to_notes)| root_kind_to_notes.keys().copied())
+            .flat_map(|(_, root_kind_to_nodes)| root_kind_to_nodes.keys().copied())
             .collect::<IndexSet<_>>();
         for root_kind in root_kinds {
             self.base.create_root_node(root_kind.into(), root_kind)?;
--- a/apollo-federation/src/query_graph/build_query_graph.rs
+++ b/apollo-federation/src/query_graph/build_query_graph.rs
@@ -1536,9 +1536,7 @@ impl FederatedQueryGraphBuilder {
             for object_def_pos in &context_refs.object_types {
                 let object = object_def_pos.get(subgraph.schema())?;
                 for dir in object.directives.get_all(&CONTEXT_DIRECTIVE_NAME) {
-                    let application = subgraph_data
-                        .federation_spec_definition
-                        .context_directive_arguments(dir)?;
+                    let application = FederationSpecDefinition::context_directive_arguments(dir)?;
                     context_name_to_types
                         .entry(application.name)
                         .or_default()
--- a/apollo-federation/src/query_graph/build_query_graph.rs
+++ b/apollo-federation/src/query_graph/build_query_graph.rs
@@ -1510,7 +1510,7 @@ impl FederatedQueryGraphBuilder {
             };
 
             // Collect data for @context
-            let mut context_name_to_types: HashMap<&str, HashSet<&ObjectTypeDefinitionPosition>> =
+            let mut context_name_to_types: HashMap<&str, HashSet<CompositeTypeDefinitionPosition>> =
                 HashMap::default();
             for object_def_pos in &context_refs.object_types {
                 let object = object_def_pos.get(subgraph.schema())?;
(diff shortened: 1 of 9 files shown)


apollo-federation/src/query_graph/build_query_graph.rs
```diff
@@ -1474,6 +1480,142 @@ impl FederatedQueryGraphBuilder {
...
+
+            let Some((_, from_context_refs)) = &subgraph
+                .referencers()
+                .directives
+                .iter()
+                // TODO: I think I need to get the subgraph data which contains the name of
+                // fromContext in that subgraph, ya?
+                .find(|(dir, _)| **dir == FROM_CONTEXT_DIRECTIVE_NAME)
+            else {
+                continue;
```
TylerBloom, line 1513: One not about this: This is silencing potential errors. There could be `@context` in the subgraph but not `@fromContext` (or worse, visa versa). This simply ignore that. I'm comfortable with this because composition *should* catch that, but it would also be reasonable track this better and raise an error.

apollo-federation/src/query_graph/mod.rs
```diff
@@ -356,6 +399,11 @@ pub struct QueryGraph {
     /// lowered composition validation on a big composition (100+ subgraphs) from ~4 minutes to
     /// ~10 seconds.
     non_trivial_followup_edges: IndexMap<EdgeIndex, Vec<EdgeIndex>>,
+    // TODO: Are these duplicate fields??
+    // TODO: If not, write docs
+    subgraph_to_args: IndexMap<Arc<str>, Vec<ObjectFieldArgumentDefinitionPosition>>,
+    subgraph_to_arg_indices:
+        IndexMap<Arc<str>, IndexMap<ObjectFieldArgumentDefinitionPosition, String>>,
```
TylerBloom: I want to revisit this was we near the end of porting this feature. These fields contain the same data, in the same order. The only difference is one also contains identifier strings. We should be able to remove `subgraph_to_args`. For now, keeping it will make porting things easier.

apollo-federation/src/query_graph/build_query_graph.rs
```diff
@@ -1474,6 +1482,131 @@ impl FederatedQueryGraphBuilder {
...
+            let mut context_name_to_types: HashMap<&str, HashSet<&ObjectTypeDefinitionPosition>> =
+                HashMap::default();
+            for object_def_pos in &context_refs.object_types {
+                let object = object_def_pos.get(subgraph.schema())?;
+                for dir in object.directives.get_all(&CONTEXT_DIRECTIVE_NAME) {
+                    let application = FederationSpecDefinition::context_directive_arguments(dir)?;
+                    context_name_to_types
+                        .entry(application.name)
+                        .or_default()
+                        .insert(object_def_pos);
```
clenfest, line 1522: I like how elegant this is in Rust.

apollo-federation/src/query_graph/build_query_graph.rs
```diff
@@ -1474,6 +1482,131 @@ impl FederatedQueryGraphBuilder {
...
+                }
+            }
+
+            // Collect data for @fromContext
+            let coordinate_map = coordinate_map.entry(subgraph_name.clone()).or_default();
+            for object_field_arg in &from_context_refs.object_field_arguments {
```
clenfest, line 1528: It's nice that the way we've architected this, we don't have to do some of the checking that's done in the JS code.

clenfest (changes requested):

apollo-federation/src/query_graph/build_query_graph.rs
```diff
@@ -1474,6 +1480,152 @@ impl FederatedQueryGraphBuilder {
...
+        }
+
+        let mut subgraph_to_args: MultiMap<Arc<str>, ObjectFieldArgumentDefinitionPosition> =
+            MultiMap::new();
+        for (source, subgraph) in self.base.query_graph.subgraphs() {
+            let Some((_, context_refs)) = &subgraph
+                .referencers()
+                .directives
+                .iter()
+                // TODO: I think I need to get the subgraph data which contains the name of the
```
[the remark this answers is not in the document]
TylerBloom: Ya, I did need to change this.

apollo-federation/src/query_graph/build_query_graph.rs
```diff
@@ -2151,6 +2316,45 @@ fn resolvable_key_applications<'doc>(
     Ok(applications)
 }
 
+fn parse_context(field: &str) -> Result<(String, String), FederationError> {
+    let pattern = Regex::new(
+        r#"^(?:[\n\r\t ,]|#[^\n\r]*)*\$(?:[\n\r\t ,]|#[^\n\r]*)*([A-Za-z_]\w*)([\s\S]*)$"#,
```
TylerBloom, line 2321: It is worth noting that the regex from JS was this:
```
/^(?:[\n\r\t ,]|#[^\n\r]*(?![^\n\r]))*\$(?:[\n\r\t ,]|#[^\n\r]*(?![^\n\r]))*([A-Za-z_]\w*(?!\w))([\s\S]*)$/
```
The ported regex is similar but does not include any look-arounds because the `regex` crate does not support them.

clenfest (approved):

PR #6226, opened 2024-11-04
add content-length header to requests as appropriate

lennyburdette:
---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/plugins/connectors/http_json_transport.rs
+++ b/apollo-router/src/plugins/connectors/http_json_transport.rs
@@ -94,32 +94,39 @@ pub(crate) fn make_request(
 
     let is_form_urlencoded = content_type.as_ref() == Some(&mime::APPLICATION_WWW_FORM_URLENCODED);
 
-    let (json_body, form_body, body, apply_to_errors) = if let Some(ref selection) = transport.body
-    {
-        // The URL and headers use the $context above, but JSON Selection errors if it is present
-        let inputs = inputs
-            .into_iter()
-            .filter(|(k, _)| *k != "$context")
-            .collect();
-        let (json_body, apply_to_errors) = selection.apply_with_vars(&json!({}), &inputs);
-        let mut form_body = None;
-        let body = if let Some(json_body) = json_body.as_ref() {
-            if is_form_urlencoded {
-                let encoded = encode_json_as_form(json_body)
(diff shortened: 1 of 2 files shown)


apollo-router/src/plugins/connectors/http_json_transport.rs
```diff
@@ -94,32 +94,39 @@ pub(crate) fn make_request(
...
+            (None, None, hyper::Body::empty(), 0, vec![])
         };
-        (json_body, form_body, body, apply_to_errors)
-    } else {
-        (None, None, hyper::Body::empty(), vec![])
-    };
+
+    if content_length > 0 {
+        request = request.header(CONTENT_LENGTH, content_length);
+    }
```
pubmodmatt, line 130: [RFC 9110](https://www.rfc-editor.org/rfc/rfc9110.html#name-content-length) has *"a user agent normally sends Content-Length in a POST request even when the value is 0 (indicating empty content).  A user agent SHOULD NOT send a Content-Length header field when the request message does not contain content and the method semantics do not anticipate such data."*

apollo-router/src/plugins/connectors/make_requests.rs
```diff
@@ -2638,7 +2638,9 @@ mod tests {
                     method: GET,
                     uri: http://localhost/api/path,
                     version: HTTP/1.1,
-                    headers: {},
+                    headers: {
+                        "content-length": "0",
```
pubmodmatt, line 2642: From the CI build, it looks like a few more insta snapshots need updating.

apollo-router/src/plugins/connectors/http_json_transport.rs
lennyburdette: do you think it should be this then?
```
if method != GET {
  request = request.header(CONTENT_LENGTH, content_length);
}
```

pubmodmatt: HTTP semantics for many of the other methods also don't contain content. We currently only support Put, Post, Patch, Get, Delete, but in case others are added, maybe it's better to be explicit, so maybe:

```rust
if method == PUT || method == POST || method == PATCH {
```

pubmodmatt (approved):

PR #6231, opened 2024-11-05
feat(federation): sketch a query completeness correctness check

goto-bus-stop:
This PR sketches out one way to do a correctness check from when I was working on it, asserting that all the paths requested by the supergraph operation are executed by the query plan. It uses this check on all operations planned in tests.

This is not a complete completeness check and has several limitations that are mentioned in the code comments (and probably some that aren't), but it might hopefully serve as a basis for discussion about what else to do!

It also needs eg. a way to handle `@skip`/`@include` conditionals. I think those could be encoded in path elements in this approach.

### DIFF
--- a/apollo-federation/src/correctness.rs
+++ b/apollo-federation/src/correctness.rs
@@ -211,7 +211,7 @@ pub fn supergraph_query_paths(
     supergraph: &ValidFederationSchema,
     operation_document: &Valid<ExecutableDocument>,
     operation_name: Option<&str>,
-) -> Result<Vec<Vec<SupergraphQueryElement>>, FederationError> {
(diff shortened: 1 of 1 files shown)


dariuszkuc:
Wondering whether `correctness.rs` should be under `test` instead of `src`?

goto-bus-stop:
I expect it will be useful to have the `correctness` module accessible from outside the crate (eg. for use in fuzzing), which is not possible if it's part of the tests

goto-bus-stop:
This may still be a useful method to explore in the future, but for now I'll close the PR in favour of Duckki's work in https://github.com/apollographql/router/pull/6498, which is further along.

PR #6232, opened 2024-11-05
Fix fake selection for nested fields in expansion

dylan-apollo:
We generate a fake selection for path-like variables (e.g., `$this.a.b`) so that we can use the same visitor code as elsewhere to collect fields to carry over to intermediate subgraphs. That selection was broken for any nested fields, since `{b}` is not valid selection on its own. This corrects that problem by omitting the surrounding braces at the first level of nesting, and only putting them back when creating the `@key`.

### DIFF
--- a/apollo-federation/src/sources/connect/expand/mod.rs
+++ b/apollo-federation/src/sources/connect/expand/mod.rs
@@ -525,7 +525,7 @@ mod helpers {
                         let (_, parsed) = JSONSelection::parse(&field_and_selection.sub_selection)
                             .map_err(|e| {
                                 FederationError::internal(format!(
-                                    "could not parse fake selection for sibling field: {e}"
+                                    "error creating key from `$this` variable: {e}"
                                 ))
                             })?;
 
--- a/apollo-federation/src/sources/connect/expand/mod.rs
+++ b/apollo-federation/src/sources/connect/expand/mod.rs
@@ -192,11 +192,11 @@ mod helpers {
     use itertools::Itertools;
 
     use super::filter_directives;
-    use super::path_to_selection;
     use super::visitors::try_insert;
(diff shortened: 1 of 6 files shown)


apollo-federation/src/sources/connect/expand/mod.rs
```diff
@@ -522,11 +522,12 @@ mod helpers {
...
                         );
-                        let (_, parsed) = JSONSelection::parse(&selection).map_err(|e| {
-                            FederationError::internal(format!(
-                                "could not parse fake selection for sibling field: {e}"
-                            ))
-                        })?;
+                        let (_, parsed) = JSONSelection::parse(&field_and_selection.sub_selection)
+                            .map_err(|e| {
+                                FederationError::internal(format!(
+                                    "could not parse fake selection for sibling field: {e}"
```
lennyburdette, line 528: I wonder if we should obscure this from customers — maybe?
```suggestion
                                    "error creating key from `$this` variable: {e}"
```

apollo-federation/src/sources/connect/expand/mod.rs
```diff
@@ -522,11 +522,12 @@ mod helpers {
                             to_schema,
                             &self.directive_deny_list,
                         );
-                        let (_, parsed) = JSONSelection::parse(&selection).map_err(|e| {
-                            FederationError::internal(format!(
-                                "could not parse fake selection for sibling field: {e}"
-                            ))
-                        })?;
+                        let (_, parsed) = JSONSelection::parse(&field_and_selection.sub_selection)
```
lennyburdette, line 525: you'll need a rebase to pick up the API change here — `parse` now returns a result.

lennyburdette (approved):

PR #6235, opened 2024-11-06
Do not report usage to Studio when variable values are invalid

goto-bus-stop:
Prevents polluting Studio with invalid enum values in particular, but also with data from queries that were never executed in general. Still needs a test.

This is kind of a workaround, and doesn't fully fix the issue. I think ideally, we would validate variables earlier (maybe using `coerce_variable_values` from apollo-rs), but that has other effects. For example, when I tried to do it in the query analysis layer, file uploads no longer worked. File uploads may require accepting`null` as a value for `Upload!` (as the actual file is provided separately)?

(With this patch, if an error occurs somewhere in between initial GraphQL validation and the call to `validate_variables`, invalid variables can still be reported to Studio.)

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/services/supergraph/service.rs
```diff
@@ -304,6 +304,15 @@ async fn service_call(
                 *response.response.status_mut() = StatusCode::NOT_ACCEPTABLE;
                 Ok(response)
             } else if let Some(err) = plan.query.validate_variables(body, &schema).err() {
+                // Replace the existing usage report so we do not report invalid variable values to
+                // Studio.
+                context.extensions().with_lock(|mut lock| {
+                    lock.insert(Arc::new(UsageReporting {
+                        stats_report_key: "GRAPHQL_VALIDATION_FAILED".to_string(),
```
goto-bus-stop, line 311:
```suggestion
                        stats_report_key: "## GraphQLValidationFailed".to_string(),
```

apollo-bot2:
## Detected Secrets

## :red_circle: Secrets Found

If you are seeing this message, it means that [the secret scanning tool](https://gitleaks.io/) that Apollo uses to prevent secrets from entering our repositories has identified a secret in your branch. Information about what was detected and steps to move forward are below.


If the secret scanner detected a valid secret, please take immediate action to remove it from your branch. Apollo also recommends that you invalidate the secret in the system it grants access to. While there are additional steps to prepare the branch for merging, protecting the secret is the top priority.


If the scanner detected a value that is not actually secret or it detected a value that _was_ a secret but has been invalidated, follow the steps in the "Resolution Process" section below to mark it as a false/benign positive. One common reason for benign positive detections is test values. For example, if you are committing a value (like a private key, JWT, etc.) that would usually be secret but is being used in a test case to validate functionality, a detection will be generated. This is safe to allowlist.

###  Value Detected 

| Signature | File | Commit | Start Line | Start Column | Link To Secret |
|-----------|------|--------|------------|--------------|----------------|
| apollo-graphos-api-key | .config/.env.staging.extended_refs | `9e3f3f9ab5c60a5b139d7819907b012fa60bfddb` | 1 | 12 | <a href="https://github.com/apollographql/router/blob/9e3f3f9ab5c60a5b139d7819907b012fa60bfddb/.config/.env.staging.extended_refs?plain=1#L1" target="_blank">Link</a> |

<hr />
<details>
    <summary><h3>Resolution Process</h3></summary>

> [!NOTE]  
> Apollo does not use all of the native options provided by Gitleaks to perform allowlisting. We use Gitleaks' standard line comments but rely on a custom convention written in a `.gitleaks.toml` file at the root of each repository for allowlisting detections in Git history.

To resolve the issue, you will need to:

1. Prevent new detections of the "secret":
    - Add a comment containing `gitleaks:allow` to the line containing the detected value.
    - Skip this step if it's not feasible. For example, the detection is in file types like JSON or Markdown, you've completely removed the value, etc.
2. Configure the scanner to ignore the "secret" in git history.
    - Follow the instructions below for _either_ "Allowlist by Commit Hash" _or_ "Allowlist by File Path"
    - You will need to ensure that a `.gitleaks.toml` file exists at the root of this repository in your branch. If not, create one!

<details>
    <summary><h4>Allowlist by Commit Hash</h4></summary>

This is the preferred option. Update the `.gitleaks.toml` file with the following contents:

```toml
[[ rules ]]
    id = "signature-from-table-above"
    [ rules.allowlist ]
        commits = [
            "full-commit-hash-1",
            "full-commit-hash-2"
        ]
```

If a `[[rules]]` block already exists for the signature involved in your detection, simply append your list of commits.

If a matching `[[rules]]` block is already present but has a `paths` attribute rather than `commits`, proceed to the next option.

This configuration tells the scanner to ignore the specified signature in the list of commits provided.

</details>
<details>
    <summary><h4>Allowlist by File Path</h4></summary>

This is a backup option to allowlist specific detections. Add/update the `.gitleaks.toml` file with the following contents:

```toml
[[ rules ]]
    id = "signature-from-table-above"
    [ rules.allowlist ]
        paths = [
            '''path/to/detected/file1.json$''',
            '''path/to/other/detected/file.md$'''
        ]
```

This configuration tells the scanner to ignore any detections with the specified signature generated for the list of files.

</details>
</details>
<hr />
<details> 



The Apollo Security team has been notified and is available to assist in resolving this issue. Please tag us on this PR using  `@apollographql/security` if you need assistance! 


</details>
<hr />
<details>
</details>

goto-bus-stop:
It looks like we don't really have the capacity/priority to finish this up any time soon in its current form. I think we should revisit after we reorder when variables are validated, so that we can implement this more easily without accruing more technical debt.

PR #6236, opened 2024-11-06
chore(federation): add format-only error macro

goto-bus-stop:
After DM discussion with @TylerBloom. They brought up that
`internal_error!()` as introduced in https://github.com/apollographql/router/pull/6080 cannot be used in
`.ok_or_else()` and some other common situations in the federation
codebase.

This splits the error creation and error return apart, following the
naming convention from `anyhow`:

- `internal_error!()` creates an internal error value from a format string.
- `bail!()` returns out of the parent function with an internal error.
- `ensure!()` returns out of the parent function with an internal error if its condition fails.

`internal_error!()` doesn't look like a massive win compared to `FederationError::internal(format!())` at first glance, but it does feel a lot more lightweight to read in my opinion.

### DIFF
--- a/apollo-federation/src/error/mod.rs
+++ b/apollo-federation/src/error/mod.rs
@@ -13,11 +13,19 @@ use lazy_static::lazy_static;
 
 use crate::subgraph::spec::FederationSpecError;
 
-/// Break out of the current function, returning an internal error.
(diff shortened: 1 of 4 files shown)


duckki (approved):
Lovely!

PR #6241, opened 2024-11-06
Provide single-character locations for `selection` parse errors

dylan-apollo:
Currently, we highlight the entire string when `selection` can't be parsed. This limits the highlight to just the offset identified by the parser (as of #6089).

An alternative could be to highlight the entire string after the offset, which might make the error more visible in-editor.

Before:
Screenshot 2024-11-06 at 12 25 07 PM

After:
Screenshot 2024-11-06 at 12 25 23 PM


lennyburdette (approved):

benjamn (approved):
Highlighting the suffix is an interesting alternative, but I like this single-character approach.

PR #6243, opened 2024-11-07
fix(federation): recompose test supergraph

goto-bus-stop:
Due to the merge order of #6237 and #6242, this supergraph was using the wrong federation version, causing tests to fail on CI.


lrlna (approved):
thank you for spotting this!

PR #6247, opened 2024-11-07
Move heavy computation to a thread pool with a priority queue

SimonSapin:
These components can take non-trivial amounts of CPU time:

* GraphQL parsing
* GraphQL validation
* Query planning
* Schema introspection

In order to avoid blocking threads that execute asynchronous code, they are now run (in their respective Rust implementations) in a new pool of as many threads as CPU cores are available. Previously we used Tokio’s [`spawn_blocking`] for this purpose, but it is appears to be intended for blocking I/O and uses up to 512 threads so it isn’t a great fit for computation tasks.

Additionally, `QueryPlannerPool` is bypassed for the new planner since the new thread pool and queue fulfill the same role. This makes the new planner parallel by default (whereas the old pool defaults to size 1, making planning effectively sequential).

[`spawn_blocking`]: https://docs.rs/tokio/latest/tokio/task/fn.spawn_blocking.html

The first commit supersedes and closes https://github.com/apollographql/router/pull/6122. The second (non-merge) commit supersedes and closes https://github.com/apollographql/router/pull/6142.

The ageing priority algorithm is based on @garypen’s work in https://github.com/apollographql/ageing

Fixes ROUTER-827

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/compute_job.rs
```diff
@@ -0,0 +1,78 @@
...
+use crate::metrics::meter_provider;
+
+/// We generate backpressure in tower `poll_ready` when reaching this many queued items
+// TODO: what’s a good number? should it be configurable?
+const QUEUE_SOFT_CAPACITY: usize = 100;
+
+// TODO: should this be configurable?
+fn thread_pool_size() -> NonZeroUsize {
+    std::thread::available_parallelism().expect("available_parallelism() failed")
+}
```
SimonSapin, line 21: Some open questions here

My opinion is that if we make something configurable just because we don’t know what a good value would be, most users won’t know either.

apollo-router/src/compute_job.rs
```diff
@@ -0,0 +1,78 @@
...
+use opentelemetry::metrics::ObservableGauge;
+use tokio::sync::oneshot;
+
+use crate::ageing_priority_queue::AgeingPriorityQueue;
+pub(crate) use crate::ageing_priority_queue::Priority;
+use crate::metrics::meter_provider;
+
+/// We generate backpressure in tower `poll_ready` when reaching this many queued items
+// TODO: what’s a good number? should it be configurable?
+const QUEUE_SOFT_CAPACITY: usize = 100;
```
garypen, line 16: I don't think it should be configurable initially. I think this queue represents the tradeoff between memory and losing time here vs being re-scheduled onto a different router, if one is available. If we are rejected from the queue here, then we know at least we have to spend the time/work to move the job elsewhere.

That's hard to quantify, but it's likely in the order of milliseconds. Perhaps we can workshop up a rough calculation based on this thinking?

apollo-router/src/compute_job.rs
```diff
@@ -0,0 +1,78 @@
...
+
+pub(crate) fn create_queue_size_gauge() -> ObservableGauge<u64> {
+    meter_provider()
+        .meter("apollo/router")
+        .u64_observable_gauge("apollo.router.compute_jobs.queued")
+        .with_description(
+            "Number of computation jobs (parsing, planning, …) waiting to be scheduled",
+        )
+        .with_callback(move |m| m.observe(queue().queued_count() as u64, &[]))
+        .init()
```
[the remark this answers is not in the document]
garypen: This new metric should be documented.

apollo-router/src/compute_job.rs
```diff
@@ -0,0 +1,78 @@
+use std::future::Future;
```
garypen, line 1: It would be great to see some unit tests for this new module.

apollo-router/src/ageing_priority_queue.rs
```diff
@@ -0,0 +1,124 @@
+use std::sync::atomic::AtomicUsize;
```
garypen, line 1: It would be great to see some unit tests for this new module.

garypen (commented):
This looks great in general. Maybe we can find some time over the next couple of days to discuss how we figure out good default values for queue/pool sizes.

apollo-router/src/compute_job.rs
goto-bus-stop: Should it be `available - 1` so we keep 1 core free to handle traffic? Or is it fine to rely on the OS scheduler to still let traffic go through the router?

apollo-router/src/ageing_priority_queue.rs
SimonSapin: Added some

apollo-router/src/compute_job.rs
SimonSapin: Added some

apollo-router/src/compute_job.rs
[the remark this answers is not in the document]
SimonSapin: I’ve added a short description in `docs/source/reference/router/telemetry/instrumentation/standard-instruments.mdx`. Are there other places to add it to?

apollo-router/src/compute_job.rs
SimonSapin: My thinking with the initial PR was to rely on the OS scheduler, but minus one might be ok too. The downside is that for example minus one out of 2 available cores has a much bigger impact that minus one out of 32 cores

garypen: Here's my proposal:

size: max(1, available - (ceiling(available/8)))

WORKINGS:
AVAILABLE: 1 POOL SIZE: 1
AVAILABLE: 2 POOL SIZE: 1
AVAILABLE: 3 POOL SIZE: 2
AVAILABLE: 4 POOL SIZE: 3
AVAILABLE: 5 POOL SIZE: 4
...
AVAILABLE: 8 POOL SIZE: 7
AVAILABLE: 9 POOL SIZE: 7
...
AVAILABLE: 16 POOL SIZE: 14
AVAILABLE: 17 POOL SIZE: 14
...
AVAILABLE: 32 POOL SIZE: 28

Tweaks on the basic approach are welcome, but it seems to offer reasonable scaling for query planning. We can always refine it later.

apollo-router/src/compute_job.rs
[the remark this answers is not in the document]
garypen: I think that's the correct location.

apollo-router/src/compute_job.rs
```diff
@@ -0,0 +1,116 @@
...
+        let test_thread = std::thread::current().id();
+        let job_thread = execute(Priority::P4, || std::thread::current().id())
+            .await
+            .unwrap();
+        assert_ne!(job_thread, test_thread)
+    }
+
+    #[tokio::test]
+    async fn test_parallelism() {
+        if thread_pool_size().get() < 2 {
```
garypen, line 98: :)

garypen (approved):
The remaining decision about QUEUE_SOFT_CAPACITY is problematic. It might be best understood as a function of the `thread_pool_size()` for now.

Proposal: 20 * `thread_pool_size()` as a starting point.

I'm approving this as is, but I'd like to know what you think of my sizing suggestions.

SimonSapin:
I’ve updated the changelog and enabled auto-merge

PR #6157, opened 2024-10-15
new docs site updates

shorgi:
Reorganizes router docs to match the IA and paths of Apollo's new docs site. Improves the authoring experience for the new docs site and makes it easier to publish docs for multiple router versions in the future.

- Changes the docs directories in router to mirror the paths and IA of the new docs site.
   - For example, the source of https://www.apollographql.com/docs/graphos/routing/performance/query-batching will now be `docs/source/routing/performance/query-batching.mdx` instead of `docs/source/executing-operations/query-batching.mdx`

- Moves all router-specific .mdx files into the router repo

PR #6248, opened 2024-11-08
reorg dev docs for new docs IA

shorgi:
Reorganize docs in `dev` in the same way as `main` (see https://github.com/apollographql/router/pull/6157) to align with new docs site's IA


Meschreiber (approved):
LGTM!

PR #6250, opened 2024-11-08
Do not override concrete type names with interface names when merging responses

tninesling:
When using `@interfaceObject`, the planner can request an object's concrete `__typename` in parallel with its data. If the data comes back with an interface name, and the router receives it after receiving the concrete name, the existing implementation will overwrite the concrete name with the interface name. The downstream result is that the router will nullify any data paths with a `__typename` that doesn't correspond to a concrete type in the schema.

This fix makes the response merging type-aware. For any field which isn't a type name, we follow the existing deep merge logic for JSON values. If we are merging two instances of `__typename`, this now checks the schema to see if the existing type is an implementer or subtype of the incoming type. If it is, we do not overwrite it with the more generic type. If the existing type is not a descendant of the incoming one, such as when they are unrelated, we overwrite it with the incoming one as usual.

---

**Checklist**

- Tests added and passing[^3]
    - [X] Unit Tests
    - [ ] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/query_planner/execution.rs
```diff
@@ -393,6 +400,58 @@ impl PlanNode {
             (value, errors)
         })
     }
+
+    /// This is a reimplementation of `ValueExt::deep_merge` which takes type hierarchies into account
+    /// when merging response objects. If we are merging two instances of `__typename`, we check the
+    /// schema to make sure we do not overwrite a subtype with an interface or parent type. Otherwise,
+    /// the data associated with the interface/parent type is removed from the response later on.
+    fn type_aware_deep_merge(schema: &Schema, target: &mut Value, source: Value) {
```
lrlna, line 408: Should this be added directly in `ValueExt` and its `Value` implementation instead? 

Also, there is another usage of `deep_merge` happening in [`execute_selection_set.rs`](https://github.com/apollographql/router/blob/6c32fd2b8fb19ecc4b3ade2bfd09cd9ac8508992/apollo-router/src/query_planner/selection.rs#L178). Does that need to be typed aware as well?

lrlna (changes requested):
I think this potentially needs to live in the same place as `.deep_merge` implementation currently lives. 

Do you know whether or not there are any performance implication for this change?

Geal:
when using interfaceObject, the planner is supposed to provide the input_rewrites and output_rewrites fields in fetch nodes to indicate how the typename field has to change. Before merging this, could it be that you found a case where the planner did not provide it?

apollo-router/src/query_planner/execution.rs
```diff
@@ -460,7 +519,7 @@ impl DeferredNode {
                     // or because it is lagging, but here we only send one message so it
                     // will not happen
                     if let Some(Ok((deferred_value, err))) = v {
-                        value.deep_merge(deferred_value);
+                        PlanNode::type_aware_deep_merge(&sc, &mut value, deferred_value);
```
sachindshinde, line 522: You missed replacing a `deep_merge()` with a type-aware merge after this usage on line 502/in `DeferredNode::execute()`.

sachindshinde (changes requested):

apollo-router/src/query_planner/execution.rs
```diff
@@ -393,6 +400,58 @@ impl PlanNode {
...
+                            e.insert(value);
+                        }
+                        Entry::Occupied(e) => match (e.into_mut(), value) {
+                            (Value::String(type1), Value::String(type2))
+                                if k.as_str() == TYPENAME =>
+                            {
+                                // If type1 implements type2, or if it is a subtype of type2, we skip this overwrite
+                                // to preserve the more specific `__typename` in the response
+                                if !schema.is_implementation(type2.as_str(), type1.as_str())
+                                    && !schema.is_subtype(type2.as_str(), type1.as_str())
```
sachindshinde, line 424: You don't need the `is_subtype()` here, it's sufficient to just check `!schema.is_implementation(type2.as_str(), type1.as_str())`. That's because `__typename` can only ever be an object type (the common case), or an interface type (the `@interfaceObject` case). It can't be a union. It's also better for performance to skip the `is_subtype()` check.

sachindshinde (changes requested):

apollo-router/src/query_planner/execution.rs
tninesling: I thought that would be the case, but my simple test fails if I remove the `is_subtype` check.

apollo-router/src/query_planner/execution.rs
sachindshinde: My understanding is that we don't need a type-aware merge in `execute_selection_set()`. Type-aware merging needs to happen whenever we're merging fetched data with other fetched data. In the case of `execute_selection_set()`, it's being called to execute selection sets against already-merged fetched data, for the purposes of computing variable values. That selection set execution may need to merge to handle inline fragments, but there won't ever be disagreements on the value of `__typename` in that merging process.

sachindshinde: > Should this be added directly in `ValueExt` and its `Value` implementation instead?

Yeah, I think that would be a better place for it.

apollo-router/src/query_planner/execution.rs
sachindshinde: Ah, looking at the implementation of `Schema::is_implementation()`, it's just bugged. It's:
1. Fetching the given interface
2. Checking whether the interface implements the implementor (should be the other way around).

There's a `FIXME` there added by @SimonSapin , not sure why we didn't fix it.

For now at least, we can just use `!schema.is_subtype(type2.as_str(), type1.as_str())` and avoid the call to `Schema::is_implementation()`.

tninesling:
> when using interfaceObject, the planner is supposed to provide the input_rewrites and output_rewrites fields in fetch nodes to indicate how the typename field has to change. Before merging this, could it be that you found a case where the planner did not provide it?

I'm not familiar with the rewriting, would you be able to explain a bit more how this works? I'm not sure it would work in the situation this PR addresses because the type name and the data are being fetched separately in parallel nodes, but the problem only seems to surface when the client explicitly requests the typename for an entity.

apollo-router/src/query_planner/execution.rs
tninesling: Moved this to `ValueExt` and updated the `execute_selection_set` implementation as well.

> Do you know whether or not there are any performance implication for this change?

The latest perf test results look normal to me, but let me know if there's a specific part you're worried about.

sachindshinde:
@Geal 
> when using interfaceObject, the planner is supposed to provide the input_rewrites and output_rewrites fields in fetch nodes to indicate how the typename field has to change. Before merging this, could it be that you found a case where the planner did not provide it?

Those query plan fields won't help in this case. To elaborate:
1. When querying an `@interfaceObject` type from a subgraph, we'll query and get back a `__typename` which is the interface type name instead of an object one.
    - We can't use output rewrites to overwrite it, since QP doesn't know what the actual type name is. (That's determined by a later query that fetches `__typename` from an interface-`@key` subgraph.)
    - We can't just not query `__typename` either, since query plan execution needs to track the type information somehow.
2. This means the response data will have a `__typename` that is incorrect, until the query to the interface-`@key` subgraph is made (which should overwrite it).

The problem is that we've encountered query plans where, even though the interface-`@key` subgraph query is ordered afterwards, the incorrect `__typename` isn't getting overwritten in execution. The real problem here has to do with how we track this kind of type information. I intend to fix this more generally when we get to source-awareness as part of the `@interfaceObject` changes, but in the meanwhile, we're looking to get a quick fix out to folks. 

The fix in this PR is to avoid clobbering the correct `__typename` with the incorrect one when merging, so it'll work regardless of merge order.

apollo-router/src/json_ext.rs
```diff
@@ -192,6 +198,53 @@ impl ValueExt for Value {
...
+            (Value::Object(a), Value::Object(b)) => {
+                for (key, value) in b.into_iter() {
+                    let k = key.clone();
+                    match a.entry(key) {
+                        Entry::Vacant(e) => {
+                            e.insert(value);
+                        }
+                        Entry::Occupied(e) => match (e.into_mut(), value) {
+                            (Value::String(type1), Value::String(type2))
+                                if k.as_str() == TYPENAME =>
```
sachindshinde, line 212: If you change this line to
```suggestion
                                if e.key().as_str() == TYPENAME =>
```
you can avoid the `key.clone()` above, which avoids some unnecessary metadata cloning in the common case.

sachindshinde (changes requested):

sachindshinde (approved):

lrlna (approved):

PR #6252, opened 2024-11-11
[docs] Update Router telemetry docs

smyrick:
We commonly get questions and confusion on `always_on` vs `always_off`

### DIFF
--- a/docs/source/reference/router/telemetry/trace-exporters/overview.mdx
+++ b/docs/source/reference/router/telemetry/trace-exporters/overview.mdx
@@ -107,8 +107,10 @@ telemetry:
          parent_based_sampler: true # (default) If an incoming span has OpenTelemetry headers then the request will always be sampled. 
 ```
 
-- `sampler` sets the sampling rate as a decimal percentage, `always_on`, or `always_off`. For example, setting `sampler: 0.1` samples 10% of your requests.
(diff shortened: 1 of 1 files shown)


docs/source/reference/router/telemetry/trace-exporters/overview.mdx
```diff
@@ -108,6 +108,7 @@ telemetry:
 ```
 
 - `sampler` sets the sampling rate as a decimal percentage, `always_on`, or `always_off`. For example, setting `sampler: 0.1` samples 10% of your requests.
+  - [`always_off`](https://opentelemetry.io/docs/specs/otel/trace/tracestate-probability-sampling/#always-off-sampler) means that no spans will make it to your APM. It doesn't mean that sampling is disabled
```
Meschreiber, line 111: I might be missing something, but there doesn't seem to be a `#always-off-sampler` anchor on https://opentelemetry.io/docs/specs/otel/trace/tracestate-probability-sampling/#always-off-sampler.
I also wonder if adding a link to Otel's docs adds more to the explanation?

In any case, I'd recommend splitting this out into the three examples, will add a suggestion.

Meschreiber, line 111:
```suggestion
- `sampler` sets the sampling rate as a decimal percentage, `always_on`, or `always_off`.
  - For example, setting `sampler: 0.1` samples 10% of your requests.
  - `always_on` (the default) sends all spans to your APM.
  - `always_off` turns off sampling. No spans reach your APM.
```

Meschreiber (approved):
Thanks for adding @smyrick ! I suggested a formatting tweak for better consistency/comprehension.

PR #6254, opened 2024-11-12
Port can satisfy context condition logic

TylerBloom:


clenfest (approved):

apollo-federation/src/query_graph/graph_path.rs
```diff
@@ -1554,11 +1629,124 @@ where
...
+        let mut context_map: IndexMap<String, ContextMapEntry> = IndexMap::default();
+
+        if !edge_weight.required_contexts.is_empty() {
+            let schema = self.graph.schema()?;
+            let mut was_unsatisfied = false;
+            for ctx in &edge_weight.required_contexts {
+                let mut levels_in_query_path = 0;
+                let mut levels_in_data_path = 0;
+                // TODO: Verify that `self.edges` and `self.edge_triggers` are the same length.
+                // Otherwise, this loop will be truncated.
```
TylerBloom, line 1648: I assume this is the case, but I want to double-check.
```suggestion
```

apollo-federation/src/query_graph/graph_path.rs
```diff
@@ -1554,11 +1629,124 @@ where
...
+                                };
+                                let Some(id) = arg_indices.get(&ctx.argument_coordinate).cloned()
+                                else {
+                                    internal_error!("TODO")
+                                };
+
+                                // TODO: This logic was in the JS code. Does a -1 cost me the
+                                // resolution was unsatisfied?
+                                // if (resolution.cost === -1 || totalCost === -1) {
+                                //  totalCost = -1;
```
TylerBloom, line 1713: This check is not needed, correct? We can't have `-1` costs nor total costs
```suggestion
```

apollo-federation/src/query_graph/graph_path.rs
```diff
@@ -1554,11 +1629,124 @@ where
...
+                                            levels_in_data_path,
+                                            levels_in_query_path,
+                                            path_tree: path_tree.clone(),
+                                            selection_set,
+                                            inbound_edge: (*e).into().unwrap(),
+                                            param_name: ctx.named_parameter.clone(),
+                                            arg_type: ctx.arg_type.clone(),
+                                            id,
+                                        };
+                                        // TODO: Do we need to check for map collisions here?
```
TylerBloom, line 1729: The JS code didn't check for this, but I want to double check. Is `named_parameter` unique?
```suggestion
```

TylerBloom (commented):
I have a few lingering questions that I would like reviewed.

apollo-federation/src/query_graph/graph_path.rs
clenfest: Yes, they will be the same length. See the comment in my PR about generating a `GraphPathItem` from these vectors.

apollo-federation/src/query_graph/graph_path.rs
```diff
@@ -1554,11 +1629,124 @@ where
...
+                                )?;
+                                // TODO: What happens if we can't convert `e` into an EdgeIndex?
+                                let resolution = condition_resolver.resolve(
+                                    (*e).into().unwrap(),
+                                    context,
+                                    excluded_destinations,
+                                    excluded_conditions,
+                                )?;
+                                // TODO: Is this still needed?
+                                // assert(edge.transition.kind === 'FieldCollection', () => `Expected edge to be a FieldCollection edge, got ${edge.transition.kind}`);
```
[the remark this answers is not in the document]
clenfest: Don't think so. I'm not sure if it's actually necessary in the original code either.

apollo-federation/src/query_graph/graph_path.rs
clenfest: We were using `resolution.cost` as a proxy for the `ConditionResolution`. Since you're checking that another way, it's fine to remove this.

apollo-federation/src/query_graph/graph_path.rs
TylerBloom: If that's the case, we should be using the `ContextCondition`'s `argument_coordinate` not the `named_parameter`. The named parameter is just a string.

sachindshinde: The named parameter represents an argument name, and in this function, we're (among other things) resolving conditions imposed by the `@fromContext`s on the arguments in the field edge handed to the function. This means the named parameters here are from the same field and unique, so they won't collide.

apollo-federation/src/query_graph/graph_path.rs
```diff
@@ -1221,12 +1315,33 @@ where
...
         let last_parameter_to_context = new_parameter_to_context.last();
 
-        let trigger = {
-            let mut trigger_rc = Arc::new(trigger);
-            if let Some(Some(last_parameter_to_context)) = last_parameter_to_context {
-                if let GraphPathTrigger::Op(op) = trigger_rc.clone().into() {
-                    if let OpGraphPathTrigger::OpPathElement(OpPathElement::Field(field)) =
-                        op.as_ref()
+        if let Some(Some(last_parameter_to_context)) = last_parameter_to_context {
+            // TODO: Perhaps it is better to explicitly cast this to `GraphPathTriggerRefMut` and
```
dariuszkuc, line 1319: do we have a followup ticket to tackle/cleanup this TODO?

PR #6255, opened 2024-11-12
Update path_tree to support set context

clenfest:


apollo-federation/src/query_graph/graph_path.rs
```diff
@@ -179,10 +189,10 @@ where
     #[serde(skip)]
     defer_on_tail: Option<DeferDirectiveArguments>,
     /// At the point where a `@context` is set, we will have fields to select
-    context_to_selection: Vec<Option<IndexSet<String>>>,
+    context_to_selection: Vec<Option<ContextToSelection>>,
```
goto-bus-stop, line 192: Does this being `None` mean something else than it being an empty set inside `Some()`? If not, we can do without the option which will make it easier to use everywhere

apollo-federation/src/query_graph/path_tree.rs
```diff
@@ -334,6 +368,16 @@ where
                         (Some(cond_a), Some(cond_b)) => cond_a.equals_same_root(cond_b),
                         _ => false,
                     }
+                    && match (&a.context_to_selection, &b.context_to_selection) {
+                        (None, None) => true,
+                        (Some(set_a), Some(set_b)) => set_a == set_b,
+                        _ => false,
+                    }
```
TylerBloom, line 375: This should just be `a.context_to_selection == b.context_to_selection`, no?
```suggestion
                    && a.context_to_selection** == b.context_to_selection
```

apollo-federation/src/query_graph/graph_path.rs
```diff
@@ -71,14 +72,23 @@ use crate::schema::ValidFederationSchema;
...
 }
 
+impl PartialEq for ContextAtUsageEntry {
+    fn eq(&self, other: &Self) -> bool {
+        self.context_id == other.context_id
+            && self.relative_path == other.relative_path
+            && self.selection_set == other.selection_set
+            && self.subgraph_arg_type == other.subgraph_arg_type
+    }
+}
```
TylerBloom, line 90: `PartialEq` can be derived.
```suggestion
#[derive(Clone, serde::Serialize, Debug, Eq)]
pub(crate) struct ContextAtUsageEntry {
    pub(crate) context_id: String,
    pub(crate) relative_path: Vec<String>,
    pub(crate) selection_set: SelectionSet,
    pub(crate) subgraph_arg_type: Type,
}
```

apollo-federation/src/query_graph/path_tree.rs
```diff
@@ -334,6 +368,16 @@ where
...
+                    && match (&a.parameter_to_context, &b.parameter_to_context) {
+                        (None, None) => true,
+                        (Some(map_a), Some(map_b)) => map_a == map_b,
+                        _ => false,
+                    }
```
TylerBloom, line 380: Same thing
```suggestion
                    && a.parameter_to_context == b.parameter_to_context
```

apollo-federation/src/query_graph/path_tree.rs
```diff
@@ -415,6 +459,14 @@ where
                     trigger: child.trigger.clone(),
                     conditions: merge_conditions(&child.conditions, &other_child.conditions),
                     tree: child.tree.merge(&other_child.tree),
+                    context_to_selection: merge_context_to_selection(
+                        &child.context_to_selection,
+                        &other_child.context_to_selection,
+                    ),
```
TylerBloom, line 465: This is somewhat inline with Renee's comment. If `context_to_selection` were not an `Option`, this could be replaced with the following. If `None` does have meaning here or is meant to communicate that these might not be used, ignore this.
```
                    context_to_selection: child
                      .context_to_selection
                      .iter()
                      .chain(&other_child.context_to_selection)
                      .collect()
```

apollo-federation/src/query_graph/path_tree.rs
clenfest: good call

apollo-federation/src/query_graph/path_tree.rs
clenfest: @TylerBloom It's a little confusing to have separate vectors for `edge_triggers`, `edge_conditions`, `context_to_selection`, and `parameter_to_context`, but the idea is that they are all of the same length and each index will translate to a single `GraphPathItem`. However a single `GraphPathItem` might have some of these values be None, hence the Option inside of the vector.

apollo-federation/src/query_graph/graph_path.rs
clenfest: See my comment to Tyler below. Yes, None does have special meaning because all of the vectors need to be the same length. Each index translates to a `GraphPathItem`, but a `GraphPathItem` may or may not have values set for each of these things.

```rust
pub(crate) type GraphPathItem<'path, TTrigger, TEdge> = (
    TEdge,
    &'path Arc<TTrigger>,
    &'path Option<Arc<OpPathTree>>,
    Option<ContextToSelection>,
    Option<ParameterToContext>,
);```

PR #6258, opened 2024-11-13
Reusable variable validation

pubmodmatt:
Reusable types and validation for variables in Apollo Connectors directives.

* Provides common parsing and validation for variable references, with location awareness for the variable reference itself as well as the parts within a reference so validation can provide specific problem locations
* Separates out the logic around what variables are available in a given context, and how those variables are resolved to types, from the expression language. For example, `$status` is only available in an HTTP response context.
* Designed to be generically useable outside of connectors if variable expressions or JSON Selection syntax becomes used more widely
* Updates both URL template and HTTP header expressions to use the common variable types and validation, so these will consistently use the same validations
* JSON Selection has *not* been updated to use the variable types directly, but a `TryFrom` implementation is added on `KnownVariable` to allow easily adapting types

Screenshot 2024-11-13 at 10 22 25 AM

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-federation/src/sources/connect/expand/mod.rs
+++ b/apollo-federation/src/sources/connect/expand/mod.rs
@@ -211,9 +211,8 @@ mod helpers {
     use crate::schema::FederationSchema;
     use crate::schema::ValidFederationSchema;
     use crate::sources::connect::json_selection::ExternalVarPaths;
-    use crate::sources::connect::json_selection::KnownVariable;
-    use crate::sources::connect::url_template::Variable;
-    use crate::sources::connect::url_template::VariableType;
+    use crate::sources::connect::variable::Namespace;
+    use crate::sources::connect::variable::VariableReference;
     use crate::sources::connect::ConnectSpecDefinition;
     use crate::sources::connect::Connector;
     use crate::sources::connect::EntityResolver;
@@ -456,28 +455,28 @@ mod helpers {
             // The actual selection might make use of the $this variable, so we grab them too
             let selection_parameters = extract_params_from_selection(&connector.selection);
 
-            let (key_for_type, var_filter) =
+            let (key_for_type, namespace_filter) =
                 if matches!(connector.entity_resolver, Some(EntityResolver::Explicit)) {
                     // `entity: true` connectors only exist on Query.
                     // We don't generate keys for `Query`, these are keys for the output type of the field.
                     // Therefore, we're also only considering the `$args` fields as keys, which should
                     // map 1-1 with output type fields.
-                    (output_type, VariableType::Args)
+                    (output_type, Namespace::Args)
                 } else {
                     // We're extending an entity by adding a new field dependent on some other fields
                     // (identified by `$this`).
-                    (parent_type, VariableType::This)
+                    (parent_type, Namespace::This)
                 };
 
             // We'll need to collect all synthesized keys for the output type, adding a federation
             // `@key` directive once completed.
             let mut keys = Vec::new();
-            for Variable { path, .. } in body_parameters
+            for VariableReference { path, .. } in body_parameters
                 .into_iter()
                 .chain(url_parameters)
                 .chain(selection_parameters)
                 .unique()
-                .filter(|var| var.var_type == var_filter)
+                .filter(|var| var.namespace.namespace == namespace_filter)
             {
                 // Arguments should be added to the synthesized key, since they are mandatory
                 // to resolving the output type. The synthesized key should only include the portions
@@ -489,7 +488,8 @@ mod helpers {
                 //
                 // All sibling fields marked by $this in a transport must be carried over to the output type
                 // regardless of its use in the output selection.
-                let field_and_selection = FieldAndSelection::from_path(&path);
+                let joined = path.iter().map(|part| part.to_string()).join(".");
+                let field_and_selection = FieldAndSelection::from_path(&joined);
                 let field_name = Name::new(field_and_selection.field_name)?;
                 let field: Box<dyn Field> = match &key_for_type {
                     TypeDefinitionPosition::Object(o)  => Box::new(o.field(field_name)),
@@ -756,33 +756,11 @@ mod helpers {
     /// Extract all seen parameters from a JSONSelection
     fn extract_params_from_selection(
         selection: &JSONSelection,
-    ) -> impl Iterator<Item = Variable> + '_ {
+    ) -> impl Iterator<Item = VariableReference<Namespace>> + '_ {
         selection
             .external_var_paths()
             .into_iter()
-            .flat_map(|var_path| {
-                let (known_type, keys) = var_path.var_name_and_nested_keys()?;
-                let var_type = match known_type {
-                    KnownVariable::Args => VariableType::Args,
-                    KnownVariable::This => VariableType::This,
-                    KnownVariable::Config => VariableType::Config,
-                    // To get the safety benefits of the KnownVariable enum, we need
-                    // to enumerate all the cases explicitly, without wildcard
-                    // matches. However, body.external_var_paths() only returns free
-                    // (externally-provided) variables like $this, $args, and
-                    // $config. The $ and @ variables, by contrast, are always bound
-                    // to something within the input data.
-                    KnownVariable::Dollar | KnownVariable::AtSign | KnownVariable::Status => {
-                        return None;
-                    }
-                };
-
-                Some(Variable {
-                    var_type,
-                    path: keys.join("."),
-                    location: 0..0, // Doesn't matter for this case, we won't report errors here
-                })
-            })
+            .flat_map(|var_path| var_path.var_name_and_nested_keys()?.try_into().ok())
     }
 
     // TODO: contribute some code to `position.rs` to make those types more flexible rather than adding it here
--- a/apollo-federation/src/sources/connect/expand/mod.rs
+++ b/apollo-federation/src/sources/connect/expand/mod.rs
@@ -488,7 +488,7 @@ mod helpers {
                 //
                 // All sibling fields marked by $this in a transport must be carried over to the output type
                 // regardless of its use in the output selection.
-                let parts = path.iter().map(|part| part.part.as_str());
+                let parts = path.iter().map(|part| part.part.as_ref());
                 let field_and_selection = FieldAndSelection::from_path(parts);
                 let field_name = Name::new(field_and_selection.field_name)?;
                 let field: Box<dyn Field> = match &key_for_type {
--- a/apollo-federation/src/sources/connect/expand/mod.rs
+++ b/apollo-federation/src/sources/connect/expand/mod.rs
@@ -488,8 +488,8 @@ mod helpers {
                 //
                 // All sibling fields marked by $this in a transport must be carried over to the output type
                 // regardless of its use in the output selection.
-                let joined = path.iter().map(|part| part.to_string()).join(".");
-                let field_and_selection = FieldAndSelection::from_path(&joined);
+                let parts = path.iter().map(|part| part.part.as_str());
+                let field_and_selection = FieldAndSelection::from_path(parts);
                 let field_name = Name::new(field_and_selection.field_name)?;
                 let field: Box<dyn Field> = match &key_for_type {
                     TypeDefinitionPosition::Object(o)  => Box::new(o.field(field_name)),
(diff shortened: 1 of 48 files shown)


apollo-federation/src/sources/connect/expand/mod.rs
```diff
@@ -489,7 +488,8 @@ mod helpers {
                 //
                 // All sibling fields marked by $this in a transport must be carried over to the output type
                 // regardless of its use in the output selection.
-                let field_and_selection = FieldAndSelection::from_path(&path);
+                let joined = path.iter().map(|part| part.to_string()).join(".");
+                let field_and_selection = FieldAndSelection::from_path(&joined);
```
dylan-apollo, line 492: Probably better to update `FieldAndSelection::from_path`, since it just splits this again. It's also only used right here.

apollo-federation/src/sources/connect/header.rs
```diff
@@ -302,20 +281,73 @@ mod tests {
         );
     }
 
-    #[rstest]
-    #[case(JSON::Array(vec!["one".into(), "two".into()]), Ok("[\"one\",\"two\"]".into()))]
-    #[case(JSON::Bool(true), Ok("true".into()))]
-    #[case(JSON::Null, Ok("null".into()))]
-    #[case(JSON::Number(1.into()), Ok("1".into()))]
-    #[case(JSON::Object(Map::new()), Ok("{}".into()))]
-    #[case(JSON::String("string".into()), Ok("string".into()))]
```
dylan-apollo, line 311: I have... mixed feeling about `rstest`. Curious about what lead to converting to standard tests in this particular case?

apollo-federation/src/sources/connect/snapshots/apollo_federation__sources__connect__url_template__test_parse__url_path_template_parse-3.snap
```diff
@@ -35,9 +43,17 @@ Ok(
...
-                            path: "b",
+                        VariableReference {
+                            namespace: VariableNamespace {
+                                namespace: $config,
+                                location: 21..28,
+                            },
+                            path: [
+                                VariablePathPart {
+                                    part: "b",
+                                    location: 29..30,
```
dylan-apollo, line 54: I'm not scrutinizing all these locations closely—but just to double check, did you verify that the highlight locations via LSP look correct for paths after the namespace? Easy to get off by ones

apollo-federation/src/sources/connect/url_template.rs
```diff
@@ -200,7 +201,13 @@ impl Component {
             remaining = suffix;
 
             if let Some((var, suffix)) = remaining.split_once('}') {
-                parts.push(ValuePart::Var(Variable::parse(var, url_template)?));
+                let start_offset = var.as_ptr() as usize - url_template.as_ptr() as usize;
+                parts.push(ValuePart::Var(
+                    VariableReference::parse(var, start_offset).map_err(|e| Error {
+                        message: e.message,
+                        location: e.location.clone(),
```
dylan-apollo, line 208:
```suggestion
                        location: e.location,
```

apollo-federation/src/sources/connect/validation/http/headers.rs
```diff
@@ -119,13 +129,82 @@ fn validate_value(
...
+                        return Some(Message {
+                            code: message.code,
+                            message: format!("{coordinate} contains an invalid variable reference {{{reference}}} - {message}", message = message.message),
+                            locations: message.locations,
+                        })
+                    },
+                    Ok(Some(ty)) => {
+                        if !ty.is_non_null() {
+                            return Some(Message {
+                                code: Code::NullablePathVariable,
```
dylan-apollo, line 189: Maybe we should rename this code

dylan-apollo (commented):
Haven't quite finished reviewing, but submitting the feedback so far

apollo-federation/src/sources/connect/validation/coordinates.rs
```diff
@@ -47,6 +48,15 @@ pub(super) struct ConnectDirectiveCoordinate<'a> {
...
 
+impl<'a> From<ConnectDirectiveCoordinate<'a>> for variable::Directive<'a> {
+    fn from(value: ConnectDirectiveCoordinate<'a>) -> Self {
+        variable::Directive::Connect {
+            field: value.field_coordinate.field,
+            object: value.field_coordinate.object,
+        }
+    }
+}
+
```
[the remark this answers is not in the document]
pubmodmatt: Agreed. This is just one step toward creating strong type representation of some of these concepts, but we can go further. We can make this more general and consolidate the coordinate into it. We'd have benefitted from creating more types like this sooner, I think, and that's led us to having a lot of redundancy and verbosity in the code.

apollo-federation/src/sources/connect/variable.rs
```diff
@@ -0,0 +1,512 @@
...
+    pub(crate) fn new(directive: Directive<'schema>, target: Target) -> Self {
+        Self { directive, target }
+    }
+}
+
+impl<'schema> ExpressionContext<Namespace> for ConnectorsContext<'schema> {
+    fn available_namespaces(&self) -> impl Iterator<Item = Namespace> {
+        match (&self.directive, &self.target) {
+            (Directive::Source, Target::ResponseBody) => {
+                vec![Namespace::Config, Namespace::Context, Namespace::Status]
```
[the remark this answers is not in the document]
pubmodmatt: Agreed, some of the combinations don't make sense, and this is part of the logic I'm trying to keep the language parsers from having to worry about. I'll think about this case - an empty set might be more appropriate here.

apollo-federation/src/sources/connect/variable.rs
```diff
@@ -0,0 +1,512 @@
...
+            .map_err(|e| VariableError {
+                message: e.message,
+                location: e
+                    .location
+                    .map(|range| range.start + start_offset..range.end + start_offset),
+            })
+    }
+}
+
+impl<N: FromStr<Err = VariableError> + ToString> FromStr for VariableReference<N> {
```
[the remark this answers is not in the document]
pubmodmatt: We need `ToString`, it just happens that the blanket implementation uses `Display`. But it could be either here, really.

apollo-federation/src/sources/connect/variable.rs
```diff
@@ -0,0 +1,512 @@
...
+    type Err = VariableError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
```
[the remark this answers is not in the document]
pubmodmatt: This was just added as a back-door for the URL template code, which (for some reason!) isn't using `nom` 😂 

The nom parsers using `nom_locate` will always get the locations.

apollo-federation/src/sources/connect/expand/mod.rs
pubmodmatt: Nice, I'll make that change.

apollo-federation/src/sources/connect/header.rs
pubmodmatt: I had feedback from the router team on a different PR that they were opposed to using macro expansion during tests as it would drive up compile time. A nice trick is that AI an convert `rstest` into regular tests if you just highlight the `rstest` and say "remove rstest", allowing you to get the declarative benefit of defining test cases with `rstest`, with the compile time efficiency of not using it.

apollo-federation/src/sources/connect/snapshots/apollo_federation__sources__connect__url_template__test_parse__url_path_template_parse-3.snap
pubmodmatt: I've done a bunch of testing in VS Code manually, and added cases to these snapshot tests, exactly for this reason (and did find and fix a few bugs as a result).

apollo-federation/src/sources/connect/validation/http/headers.rs
pubmodmatt: What do you mean here?

apollo-federation/src/sources/connect/variable.rs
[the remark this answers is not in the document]
lennyburdette: it'll be useful for defining `isSuccess: "$status->in([200, 201])` or `errors: { message: "$.message { localized_string }" }`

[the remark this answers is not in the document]
pubmodmatt: I ended up separating out a concept of `Phase`, which is either `Request` or `Response`. This aligns better with things like `$status`, which is only available in a response, regardless of what the target is. Target is now just `Url`, `Header`, or `Body`. For the case Lenny mentioned, we might add something like `IsSuccess`.

apollo-federation/src/sources/connect/validation/http/headers.rs
```diff
@@ -1,22 +1,32 @@
...
 use crate::sources::connect::validation::Message;
+use crate::sources::connect::variable::ConnectorsContext;
+use crate::sources::connect::variable::Directive;
+use crate::sources::connect::variable::Phase;
+use crate::sources::connect::variable::Target;
 
 pub(crate) fn validate_arg<'a>(
     http_arg: &'a [(Name, Node<Value>)],
+    schema: &'a SchemaInfo,
     source_map: &'a SourceMap,
```
dylan-apollo, line 30:
```suggestion
```

We can use `&schema.sources` now, rather than passing both around.

apollo-federation/src/sources/connect/validation/http/headers.rs
```diff
@@ -79,7 +89,7 @@ pub(crate) fn validate_arg<'a>(
                     validate_name(&FROM_ARG, from_value, coordinate, source_map).err()
                 },
                 (None, Some(value_arg)) => {
-                    validate_value(value_arg, coordinate, source_map)
+                    validate_value(value_arg, schema, coordinate, source_map)
```
dylan-apollo, line 92:
```suggestion
                    validate_value(value_arg, schema, coordinate)
```

apollo-federation/src/sources/connect/validation/http/headers.rs
```diff
@@ -111,6 +121,7 @@ pub(crate) fn validate_arg<'a>(
 
 fn validate_value(
     value: &Node<Value>,
+    schema: &SchemaInfo,
     coordinate: HttpHeadersCoordinate,
     source_map: &SourceMap,
```
dylan-apollo, line 126:
```suggestion
```

apollo-federation/src/sources/connect/validation/http/headers.rs
dylan-apollo: The error code `NullablePathVariable` is not just about URL "path" anymore. So maybe like `NullabilityWarning` or something more generic that we can use?

Also, nullable URI paths might be okay in the future, depending on which proposal we go with.

apollo-federation/src/sources/connect/validation/snapshots/validation_tests@invalid-path-parameter.graphql.snap
```diff
@@ -6,7 +6,7 @@ input_file: apollo-federation/src/sources/connect/validation/test_data/invalid-p
 [
     Message {
         code: InvalidUrl,
-        message: "`GET` in `@connect(http:)` on `Query.resources` must be a valid URL template. Variable type must be one of $args, $this, $config, got `$blah`",
+        message: "`GET` in `@connect(http:)` on `Query.resources` must be a valid URL template. Unknown variable namespace `$blah`",
```
dylan-apollo, line 9: Can we still list the allowed namespaces as a hint? I guess maybe we should also add a link to variables docs to some errors.

apollo-federation/src/sources/connect/validation/http/headers.rs
```diff
@@ -119,13 +130,82 @@ fn validate_value(
...
+            .collect(),
+    }) {
+        Err(message) => return Some(message),
+        Ok(header_value) => {
+            for reference in header_value.variable_references() {
+                match variable_resolver.resolve(reference, expression) {
+                    Err(message) => {
+                        return Some(Message {
+                            code: message.code,
+                            message: format!("{coordinate} contains an invalid variable reference {{{reference}}} - {message}", message = message.message),
```
dylan-apollo, line 183:
```suggestion
                            message: format!("{coordinate} contains an invalid variable reference `{{{reference}}}` - {message}", message = message.message),
```

dylan-apollo (approved):
Left a couple more suggestions, but looks good overall!

apollo-federation/src/sources/connect/variable.rs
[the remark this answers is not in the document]
dylan-apollo: I think phase makes a lot of sense, thanks!

apollo-federation/src/sources/connect/validation/snapshots/validation_tests@invalid-path-parameter.graphql.snap
pubmodmatt: This required some significant changes to propagate the invalid namespace error up to a point in the stack that has the expression context, so it can enumerate the available namespaces, but I got it working.

PR #6259, opened 2024-11-13
Make all header validations consistent

dylan-apollo:
WIP

Before this PR, there were four separate validations for headers:

1. At composition, header names and values are parsed to check for validity. This also checks whether `to` and `from` are both set or just one of them, and disallows duplicate headers.
2. When the router loads a schema, it has a separate, less thorough check for argument validity on `@source`
3. When the router loads a schema, it also checks `@connect` headers in a similar code path, but one that didn't allow single header objects yet.
4. At runtime, certain header names are disallowed in some contexts

This PR unifies the base definition of what's considered a valid `headers` argument, catching disallowed header names at composition time and at router schema load (so there's no need to validate again at runtime).

Improves the locations of a number of validations, pointing at the specific header that errored rather than the whole `headers` block. Also makes the locations of a couple pieces worse, when there are multiple locations within a single header value (i.e., pointing at the whole header instead of the `name` and `value` arguments separately).

TODO: 

- [ ] Check for variables after #6258 merges. None allowed in `@source` (I think?) and only valid types allowed in `@connect`

### DIFF
--- a/apollo-federation/src/sources/connect/header.rs
+++ b/apollo-federation/src/sources/connect/header.rs
@@ -37,11 +37,11 @@ impl HeaderValue {
     ///
     /// # Errors
     /// Returns an error if a variable used in the header value is not defined.
-    pub fn interpolate(&self, vars: &Map<ByteString, JSON>) -> Result<String, String> {
-        let mut result = String::new();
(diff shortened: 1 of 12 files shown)


dylan-apollo:
Remaking post-merge

PR #6263, opened 2024-11-15
Refactor `handle_requires()` to support `@context` use cases

sachindshinde:
In order to support `@context` during fetch dependency graph processing, this PR refactors `handle_requires()` into two functions: `handle_conditions_tree()` and `create_post_requires_node()` (the first of which gets reused by `@context` logic).

### DIFF
--- a/apollo-federation/src/query_plan/fetch_dependency_graph.rs
+++ b/apollo-federation/src/query_plan/fetch_dependency_graph.rs
@@ -29,6 +29,7 @@ use serde::Serialize;
 
 use super::query_planner::SubgraphOperationCompression;
 use super::FetchDataKeyRenamer;
+use crate::bail;
 use crate::display_helpers::DisplayOption;
 use crate::error::FederationError;
 use crate::error::SingleFederationError;
@@ -4170,10 +4171,10 @@ fn handle_conditions_tree(
                 let fetch_node = dependency_graph.node_weight(fetch_node_id)?;
                 let subgraph_name = fetch_node.subgraph_name.clone();
                 let Some(merge_at) = fetch_node.merge_at.clone() else {
-                    return Err(FederationError::internal(format!(
+                    bail!(
                         "Fetch node {} merge_at_path is required but was missing",
                         fetch_node_id.index()
-                    )));
+                    );
                 };
                 let defer_ref = fetch_node.defer_ref.clone();
                 let copied_node_id =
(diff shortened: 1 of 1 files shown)


apollo-federation/src/query_plan/fetch_dependency_graph.rs
```diff
@@ -4064,134 +4071,183 @@ fn extract_defer_from_operation(
...
+                // Since we may want the condition fields in this case to be added into an earlier
+                // node, we create a copy of the current node (with only the inputs), and the
+                // condition fields will be added to this node instead of the current one.
+                let fetch_node = dependency_graph.node_weight(fetch_node_id)?;
+                let subgraph_name = fetch_node.subgraph_name.clone();
+                let Some(merge_at) = fetch_node.merge_at.clone() else {
+                    return Err(FederationError::internal(format!(
+                        "Fetch node {} merge_at_path is required but was missing",
+                        fetch_node_id.index()
+                    )));
```
TylerBloom, line 4128: You can use `bail!` rather than `Err(FederationError::internal(format!(..)))`.

TylerBloom (approved):
Looks good to me. There are a few places were you can use the `bail!` macro rather than `return Err(FederationError::internal(format!(..)))`. Other than that, it looks good to me.

apollo-federation/src/query_plan/fetch_dependency_graph.rs
```diff
@@ -4269,11 +4325,86 @@ fn handle_requires(
...
+        .federated_query_graph
+        .schema_by_source(&head.source)?
+        .clone();
+    let QueryGraphNodeType::SchemaType(OutputTypeDefinitionPosition::Object(entity_type_position)) =
+        head.type_.clone()
+    else {
+        return Err(FederationError::internal(
+            "@requires applied on non-entity object type",
+        ));
+    };
```
dariuszkuc, line 4383: this check was now moved after we've done all the `handle_conditions_tree` vs very first check that we did in the past - should we move it up? I'd assume we should never hit this error as this should be handled by composition so probably doesn't matter.

apollo-federation/src/query_plan/fetch_dependency_graph.rs
sachindshinde: Makes sense, I've replaced those patterns in the refactored functions with `bail!`.

PR #6264, opened 2024-11-15
Add context to FetchDependencyGraphNode and FetchInputs

clenfest:


TylerBloom (approved):

apollo-federation/src/query_plan/fetch_dependency_graph.rs
```diff
@@ -167,6 +168,8 @@ pub(crate) struct FetchDependencyGraphNode {
     inputs: Option<Arc<FetchInputs>>,
     /// Input rewrites for query plan execution to perform prior to executing the fetch.
     input_rewrites: Arc<Vec<Arc<FetchDataRewrite>>>,
+    /// Rewrites that will need to occur to store contextual data for future use
+    context_inputs: Option<Vec<FetchDataKeyRenamer>>,
```
dariuszkuc, line 172: would it make it simpler to just use `Vec` with empty vector implying None? Or is there a situation where we need to distinguish between `None` and empty vec?

PR #6268, opened 2024-11-18
[next] Remove --apollo-uplink-poll-interval

goto-bus-stop:
This is targeting Router 2.0.

`--apollo-uplink-poll-interval` does not do a lot today. It's supposed to configure the interval between polls of Apollo Uplink to retrieve new configuration or a new supergraph schema. But after one poll, Uplink responds with a new delay, which is used going forward. So in effect this only configures the *initial poll*, which is certainly not expected.

**Open questions**
- [ ] While removing this I saw that the poll interval value is used if you provide URLs for the supergraph schema source with hot_reload. Should we continue to support that?
- [x] Should I document this change somewhere? Do we have a migration document? Or should I just use a changeset?

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

There were no tests for this feature

[ROUTER-825]: https://apollographql.atlassian.net/browse/ROUTER-825?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bnjjj (approved):
The migration guide https://github.com/apollographql/router/blob/next/docs/source/reference/migration/from-router-v1.mdx

garypen (dismissed):
As discussed: Please add details of the removal to the migration guide so we don't lose visibility of this change.

goto-bus-stop:
I added a new section for "Removed features" in the migration guide.

garypen (commented):
I don't understand this question:
```
While removing this I saw that the poll interval value is used if you provide URLs for the supergraph schema source with hot_reload. Should we continue to support that?
```
But if you supply an URL (I'm assuming you mean using the -s flag) how would the interval be used? The files provided via -s interact with a File Watcher which has its own default watching duration.

goto-bus-stop:
I'm not entirely sure what it is or if or how anyone uses it. But this is what I'm referring to:
https://github.com/apollographql/router/blob/ba697567927f0b0fe52fec56663ac51219ed039f/apollo-router/src/executable.rs#L202-L204

If you provide `--supergraph-urls`, you get a `SchemaSource::URLs` that re-fetches every `--apollo-uplink-poll-interval` seconds: (snippet is from this branch, so already modified to the hardcoded INITIAL_UPLINK_POLL_INTERVAL)
https://github.com/apollographql/router/blob/ba697567927f0b0fe52fec56663ac51219ed039f/apollo-router/src/executable.rs#L545-L554

PR #6269, opened 2024-11-18
Merge dev to next

garypen:
Take the appropriate changes from dev to next.

dev is at 0bdea7eaa40dc21cec727e9cf7d3d5c84e4619b7


goto-bus-stop:
is this a squash merge? we should use a "normal" merge commit i imagine?

garypen:
> is this a squash merge? we should use a "normal" merge commit i imagine so the whole history is visible on next?

It's a `git merge dev`. Are you implying we should rebase?

garypen:
Closing, since merge wasn't performed correctly.

PR #6270, opened 2024-11-18
fix(federation): remove special-casing of adding reserved or built-in items to a schema

goto-bus-stop:
This is my proposed "solution" for handling the shadowing of builtin types and directives in federated schemas.

That is, what happens if a subgraph (*or supergraph*) SDL contains a scalar declaration like:
```graphql
scalar Boolean
```
where the type name collides with a name provided by the GraphQL spec.

The PR contains some cleanup commits. The first two commits are cleanup, the rest is implementing the below.

## Background

graphql-js completely ignores SDL definitions of built-in scalars, instead only using the built-in ones:

https://github.com/graphql/graphql-js/blob/c079ae3cf8bf5bc0f255bde1ce2fbd424436cffb/src/utilities/extendSchema.ts#L185-L188 

For directives, graphql-js uses the built-in definitions if no custom definition was provided:

https://github.com/graphql/graphql-js/blob/c079ae3cf8bf5bc0f255bde1ce2fbd424436cffb/src/utilities/buildASTSchema.ts#L81-L89

As for federation, JS composition does not propagate a shadowed builtin scalar definition *or* a redefined builtin directive in a subgraph to the supergraph. If you write `scalar Boolean` in your subgraph, the supergraph does not have it. The same goes for a directive like `@skip`. I was surprised by this, but if I declare a subgraph with:
```graphql
directive @skip(if: Boolean!) on FIELD
directive @myskip(if: Boolean!) on FIELD
```
then only `@myskip` is propagated to the supergraph.

In Rust, apollo-rs validation rejects shadowing of built-in types and directives. This could be seen as a bug, but I don’t think it is relevant to us today, as supergraphs do not contain shadowed builtin declarations. It may be a problem for the composition work, as subgraphs that write `scalar Boolean` will be rejected by apollo-rs schema validation. It may be appropriate for apollo-rs to accept shadowed builtin types, like graphql-js does.

I conclude that there are no supergraphs today that shadow builtin scalars or builtin directives. The router will not start up with a supergraph like that.

## Solution

As supergraphs produced by JS composition do not contain shadowing declarations, I suggest we should not treat them specially in the position.rs code. I think the caller of `.pre_insert()` and `.insert()` should ensure that they do not insert a preexisting builtin scalar or directive. Since the query planner and the router both make many assumptions about the meaning of `@skip` and `@include`, and of builtin types, I think we should explicitly not support customising them.

This PR implements that by removing the handling of shadowed types in position.rs.

[ROUTER-488]: https://apollographql.atlassian.net/browse/ROUTER-488?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apollo-federation/src/schema/position.rs
+++ b/apollo-federation/src/schema/position.rs
@@ -4,7 +4,6 @@ use std::fmt::Formatter;
 use std::ops::Deref;
 
 use apollo_compiler::ast;
-use apollo_compiler::collections::IndexSet;
 use apollo_compiler::name;
 use apollo_compiler::schema::Component;
 use apollo_compiler::schema::ComponentName;
@@ -25,7 +24,6 @@ use apollo_compiler::Name;
 use apollo_compiler::Node;
 use apollo_compiler::Schema;
 use either::Either;
-use lazy_static::lazy_static;
 use serde::Serialize;
 use strum::IntoEnumIterator;
 
@@ -1051,11 +1049,6 @@ impl ScalarTypeDefinitionPosition {
 
     pub(crate) fn pre_insert(&self, schema: &mut FederationSchema) -> Result<(), FederationError> {
         if schema.referencers.contains_type_name(&self.type_name) {
-            if is_graphql_reserved_name(&self.type_name)
-                || GRAPHQL_BUILTIN_SCALAR_NAMES.contains(&self.type_name)
-            {
-                return Ok(());
-            }
             bail!(r#"Type "{self}" has already been pre-inserted"#);
         }
         schema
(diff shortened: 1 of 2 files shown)


apollo-federation/src/schema/position.rs
```diff
@@ -1087,22 +1077,10 @@ impl ScalarTypeDefinitionPosition {
...
         {
-            return Err(SingleFederationError::Internal {
-                message: format!("Type \"{}\" has not been pre-inserted", self),
-            }
-            .into());
+            bail!(r#"Type "{self}" has not been pre-inserted"#);
         }
         if schema.schema.types.contains_key(&self.type_name) {
-            // TODO: Allow built-in shadowing instead of ignoring them
-            if is_graphql_reserved_name(&self.type_name)
```
goto-bus-stop, line 1097: This implies that adding a duplicate reserved name is now also an error. It makes sense in my opinion.

dariuszkuc (approved):
As mentioned in the PR description - we should not accept those in the supergraph. We do have to handle it in the subgraphs (as pretty much all code-first solutions will dump SDL with all directives) but that should be handled in `apollo-rs`.

lrlna (approved):
Your reasoning makes a lot of sense to me. And I think once we have to deal with composition, we can think of how and if we need to do something differently. I'd say merge this in before it gets too stale.

PR #6274, opened 2024-11-18
Remove support for scaffold

garypen:
In router 2.0, we are going to be re-thinking how best to support customization.

Scaffold isn't going to be part of the solution, so let's remove it.


deny.toml
```diff
@@ -93,7 +93,7 @@ highlight = "all"
 
 # List of crates to deny
 deny = [
-  # `cargo-scaffold` uses `git2` which uses `libssh2-sys` and `libgit2-sys`.
+  # `<deleted>` uses `git2` which uses `libssh2-sys` and `libgit2-sys`.
```
garypen, line 96: I need to figure out how to update this comment. I think it might be that we can do a lot of simplification here, but I'll need to discuss this with someone more knowledgeable about deny.toml first.

SimonSapin (approved):

BrynCooke (approved):

PR #6277, opened 2024-11-18
Fix YAML reference URL link in values.yaml

lambertjosh:
Fixes the link to the router YAML configuration definition. Presently it was referring to a valid page, but that page has been modified so that the anchor no longer is present and the YAML definition has actually moved.


Meschreiber (approved):
Thank you @lambertjosh !

PR #6278, opened 2024-11-18
Consolidate & expand connectors headers validations

dylan-apollo:
Before this PR, there were four separate validations for headers:

1. At composition, header names and values are parsed to check for validity. This also checks whether `to` and `from` are both set or just one of them, and disallows duplicate headers.
2. When the router loads a schema, it has a separate, less thorough check for argument validity on `@source`
3. When the router loads a schema, it also checks `@connect` headers in a similar code path, but one that didn't allow single header objects yet.
4. At runtime, certain header names are disallowed in some contexts

This PR unifies the base definition of what's considered a valid `headers` argument, catching disallowed header names at composition time and at router schema load (so there's no need to validate again at runtime).

Improves the locations of a number of validations, pointing at the specific header that errored rather than the whole `headers` block.


apollo-federation/src/sources/connect/variable.rs
```diff
@@ -190,6 +190,10 @@ pub(crate) enum VariableParseError<I> {
         location: Range<usize>,
     },
     Nom(I, ErrorKind),
+    InvalidHeaderValue {
```
pubmodmatt, line 193: This doesn't belong here - variables are agnostic to where they are used, so adding a header-specific variant here doesn't make sense. I think you added this because you need to propagate this error our of the HeaderValue parser and its using the same error type. That was just laziness on my part - at the time, the HeaderValue parser didn't *have* any errors that weren't variable errors, so I just passed back the variable error type. But since you're now adding an error that is specfiic to headers, you should add a new header error type and add this variant there instead.

apollo-federation/src/sources/connect/header.rs
```diff
@@ -145,28 +149,42 @@ impl From<VariableParseError<Span<'_>>> for HeaderValueError {
...
+            map(parse_header_value, Self::Constant),
         ))(input)
     }
 }
 
 type Span<'a> = LocatedSpan<&'a str>;
 
-fn text(input: Span) -> IResult<Span, Span, VariableParseError<Span>> {
-    recognize(many1(none_of("{}")))(input)
+fn parse_header_value(input: Span) -> IResult<Span, http::HeaderValue, VariableParseError<Span>> {
```
pubmodmatt, line 178: This is where you should use a new `HeaderParseError` type, or something like that. One variant of that might be `VariableError`, and another might be `ValueError` or something. This was just returning `VariableParseError` because variable errors were the only thing that could go wrong previously.

pubmodmatt (changes requested):
Looks good overall! Just one change requested on a new header error type.

apollo-federation/src/sources/connect/models.rs
```diff
@@ -225,10 +237,194 @@ impl HTTPMethod {
...
+            .iter()
+            .find(|(name, value)| *name == HTTP_HEADER_MAPPING_FROM_ARGUMENT_NAME);
+        let value = mappings
+            .iter()
+            .find(|(name, value)| *name == HTTP_HEADER_MAPPING_VALUE_ARGUMENT_NAME);
+
+        match (from, value) {
+            (Some(_), None) if Self::is_static(&name) => {
+                Err(HeaderParseError::Other{ message: format!(
+                    "header '{name}' can't be set dynamically, only via `{HTTP_HEADER_MAPPING_VALUE_ARGUMENT_NAME}`"
```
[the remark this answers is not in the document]
dylan-apollo: Decided to just specifically say it can be set with `from`, so hopefully the fix is obvious.

Issue #6154, opened 2024-10-15
Router errors on queries that contains only __typename on root level that wrapped in fragment


IvanGoncharov:
## Describe the bug
Queries like { __typename } works on router but if you wrap it in fragment like this `{ ... { __typename  } }` it produces an error:

> value retrieval failed: empty query plan. This behavior is unexpected and we suggest opening an issue to apollographql/router with a reproduction.

## To Reproduce

See test cases here:
https://github.com/apollographql/router/blob/7402a1b62fb7a08acf1f03149986771af0baf7f9/apollo-router/tests/integration/typename.rs#L74-L138

[ROUTER-749]: https://apollographql.atlassian.net/browse/ROUTER-749?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

DimaPF:
Hello. It appears that I've found a way to reproduce this issue or a similar one (at least in version 1.59.1). 
Router returns INTERNAL_SERVER_ERROR with “empty query plan” when all planned fields are excluded via @include(false) and only __typename remains.

When I send any requests like:
`mutation UpdateRoom($input: RoomUpdateDto!) {
  updateRoom(input: $input) @include(if: false){
    id
    name
  }
  __typename
}`
I got the next response:
`{
    "errors": [
        {
            "message": "value retrieval failed: empty query plan. This behavior is unexpected and we suggest opening an issue to apollographql/router with a reproduction.",
            "extensions": {
                "code": "INTERNAL_SERVER_ERROR"
            }
        }
    ]
}`

**Environment**:
Router version: v1.59.1
Federation/supergraph

**Additional context**:
Tried Apollo-Expose-Query-Plan: true, but environment doesn’t return extensions.queryPlan even for successful requests (likely plugin disabled)

SimonSapin:
Hi

* Consider opening a new issue since it’s likely that no-one else will watch a resolved one. (I am not working on this anymore and will unsubscribe from this issue now.)
* I think what you’re running into is: there’s a "root `__typename` only" code path that may currently only apply to queries and maybe should also apply to mutations.

PR #6282, opened 2024-11-19
Enable typename tests that now pass

SimonSapin:
Fixes https://github.com/apollographql/router/issues/6154

These tests were using `#[should_panic]` without an `expected` message. At some point the underlying bug was fixed but these tests still panicked for a different reason, so we didn’t notice at the time.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/tests/integration/typename.rs
+++ b/apollo-router/tests/integration/typename.rs
@@ -106,11 +106,7 @@ async fn aliased() {
     "###);
 }
 
-// FIXME: bellow test panic because of bug in query planner, failing with:
-// "value retrieval failed: empty query plan. This behavior is unexpected and we suggest opening an issue to apollographql/router with a reproduction."
-// See: https://github.com/apollographql/router/issues/6154
 #[tokio::test]
-#[should_panic]
 async fn inside_inline_fragment() {
     let request = Request::fake_builder()
         .query("{ ... { __typename } }")
@@ -120,14 +116,13 @@ async fn inside_inline_fragment() {
     insta::assert_json_snapshot!(response, @r###"
     {
       "data": {
-        "n": "MyQuery"
+        "__typename": "MyQuery"
       }
     }
     "###);
 }
 
 #[tokio::test]
-#[should_panic] // See above FIXME
 async fn inside_fragment() {
     let query = r#"
        { ...SomeFragment }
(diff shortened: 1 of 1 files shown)


IvanGoncharov (approved):

apollo-router/tests/integration/typename.rs
```diff
@@ -215,6 +209,9 @@ async fn two_named_operations() {
 async fn make_request(request: Request) -> apollo_router::graphql::Response {
     apollo_router::TestHarness::builder()
         .configuration_json(json!({
+            "supergraph": {
+                "introspection": true,
+            },
```
IvanGoncharov, line 214: To make sure. 
These tests don't work without this flag, right?
If the query only has `__typename`, is it treated as an introspection query?

SimonSapin: Yes without this these tests return `INTROSPECTION_DISABLED`

… but I didn’t realize opening this PR: that’s also a bug! Even if we internally use the introspection execution engine to handle `__typename` in fragment, it probably shouldn’t be rejected when (schema-)introspection is disabled

I’ll tweak the rejection logic in this PR

SimonSapin:
I’ve tweak the execution logic and rephrased the PR title and description, please take another look

apollo-router/src/introspection.rs
```diff
@@ -54,10 +54,16 @@ impl IntrospectionCache {
...
+                if doc.has_explicit_root_fields {
+                    ControlFlow::Break(Self::mixed_fields_error())?;
+                } else {
+                    ControlFlow::Break(self.cached_introspection(schema, key, doc).await)?
+                }
             } else if !doc.has_explicit_root_fields {
-                ControlFlow::Break(self.cached_introspection(schema, key, doc).await)?
+                // root __typename only, probably a small query
+                // Execute it without caching:
+                ControlFlow::Break(Self::execute_introspection(schema, doc))?
```
goto-bus-stop, line 66: Shouldn't this be handled by the `maybe_lone_root_typename` call earlier in the function? Or maybe we can remove that function (which doesn't handle fragments), and only rely on this branch now?

goto-bus-stop (commented):
LGTM but i'd like to have the question clarified before merging

apollo-router/src/introspection.rs
SimonSapin: `maybe_lone_root_typename` handles exactly `{__typename}` (which is very common) and ~nothing else. This handles cases where `__typename` is the only field, but fragments or directives may be used.

It’s true that the `maybe_lone_root_typename` fast path becomes less impactful with this PR since the cache is bypassed either way. I wouldn’t oppose remove it, though I don’t know if this fast path is still faster than going through the executor machinery.

PR #6287, opened 2024-11-19
QP comparison mode: skip comparing PlanNode.authorization

lrlna:
We are getting mismatches when comparing PlanNodes between JS & Rust with ~authorisation~ authorization directives. This mismatch comes from `authorization` field in PlanNode. The executing query plan (and in comparison mode, it will be the JS QP) receives the appropriate `CacheKeyMetadata`,

```rust
CacheKeyMetadata { 
  is_authenticated: true,
  scopes: [],
  policies: []
}
```
but the Rust QP does not, since it will not be used to execute, authenticate, etc any of the requests and gets discarded in this comparison mode. 

Because we do not attach this metadata to the Rust PlanNode **in comparison mode**, we see a ton of false mismatches for everyone using auth directives. 

This PR skips comparing `authorization` field in PlanNode, as this field gets populated _after_ the query plan is already present. In addition, this data is not based on the query plan itself, so it's not necessary to establish equivalence between the two implementations.

When **just the Rust QP is used**, the cache key metadata is present and accounted for correctly, which is why none of the mocks fail when running with only the Rust planner.

It's incredibly difficult to write a test for this little nit in semantic equivalence, so this PR does not come with any, apologies.

### DIFF
--- a/apollo-router/src/plugins/authorization/tests.rs
+++ b/apollo-router/src/plugins/authorization/tests.rs
@@ -51,6 +51,7 @@ async fn authenticated_request() {
         "include_subgraph_errors": {
             "all": true
         },
+        "experimental_query_planner_mode": "new",
         "authorization": {
             "require_authentication": true
         }}))
@@ -123,6 +124,7 @@ async fn unauthenticated_request() {
         "include_subgraph_errors": {
             "all": true
         },
+        "experimental_query_planner_mode": "new",
         "authorization": {
             "require_authentication": true
         }}))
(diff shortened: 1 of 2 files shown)


duckki:
The `FetchNode::authentication` field is ultimately computed in the `extract_authorization_metadata` method. And it's called after the query plan is generated. So, it's not really a part of query planning. Even though we could pass the `CacheKeyMetadata` to the comparison thread and populate the same info on the Rust query plan, it seems unnecessary to me.

Thus, I agree that we remove the authentication field comparison in semantic diff. And the code change makes sense.

duckki (approved):

PR #6288, opened 2024-11-19
Ensure that demand control scoring handles custom scalars that represent non-GraphQL-compliant JSON

tninesling:
When scoring arguments for demand control, we were handling variables by converting them to the equivalent AST values to reuse the argument scoring code. This works when the entire JSON object passed as a variable value is GraphQL-compliant. However, there is an edge case where the schema contains a custom scalar representing arbitrary JSON which is not GraphQL-compliant. The case we ran into was a JSON object with a key containing a dot ".", which caused the AST conversion to panic. Instead of converting the entire structure to the AST equivalent, this change traverses the JSON and uses the Schema types to properly stop when it encounters a scalar.

---

**Checklist**

- Tests added and passing[^3]
    - [X] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


lrlna (commented):
Do we _want_ to support arbitrary JSON that's not GraphQL-compliant? It's interesting that this type surfaces only in demand control? Do we fail anywhere else with this kind of custom scalar? 🤔

apollo-router/src/plugins/demand_control/cost_calculator/static_cost.rs
```diff
@@ -1073,4 +1129,25 @@ mod tests {
...
+                    json: $myJsonValue
+                })
+            }
+        "#;
+        let variables = r#"
+            {
+                "myJsonValue": {
+                    "field.with.dots": 1
+                }
+            }
```
lrlna, line 1148: the idea is not to let demand control scoring happen via this arbitrary json, right? or is that also the idea?

.changesets/fix_tninesling_demand_control_variable_scoring.md
```diff
@@ -0,0 +1,3 @@
+### Fix demand control panic for custom scalars that represent non-GraphQL-compliant JSON ([PR #6288](https://github.com/apollographql/router/pull/6288))
+
+By [@tninesling](https://github.com/tninesling) in https://github.com/apollographql/router/pull/6288
```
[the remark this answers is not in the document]
tninesling: Yes! Added

apollo-router/src/plugins/demand_control/cost_calculator/static_cost.rs
tninesling: The idea is that even though it is a JSON structure, it should be scored as a scalar in the same way we treat built-in scalars like `Int` or `String` (free, unless annotated with `@cost`). Previously, I was naively assuming that any JSON object structure passed in as variables could be converted into the equivalent AST structure. Unfortunately the custom scalar pattern means that the otherwise compliant JSON structure could contain a subtree of non-compliant JSON.

tninesling:
> Do we _want_ to support arbitrary JSON that's not GraphQL-compliant? It's interesting that this type surfaces only in demand control? Do we fail anywhere else with this kind of custom scalar? 🤔

@lrlna I think we have to support it at this point. @dariuszkuc mentioned this is considered an antipattern, but some customers use it to get around the lack of an arbitrary map type, as is the case for the customer who reported this.

PR #6291, opened 2024-11-20
Removed TODOs in context snapshot tests

TylerBloom:
This PR adds a macro to perform the assertions necessary for checking the `FetchDataRewrite::Rename` nodes in the `@context` snapshot tests.


clenfest:
One global comment is that I think we do need to validate the path of the rewrite.

PR #6293, opened 2024-11-20
fix(file_upload): remove placeholders from query variables

IvanGoncharov:
*Description here*

Fixes #**issue_number**

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/plugins/file_uploads/mod.rs
+++ b/apollo-router/src/plugins/file_uploads/mod.rs
@@ -254,7 +254,7 @@ fn replace_value_at_path<'a>(
 
 // Removes value at path.
 fn remove_value_at_path<'a>(variables: &'a mut json_ext::Object, path: &'a [String]) {
-    let _ = get_value_at_path(variables, path).take();
+    let _ = get_value_at_path(variables, path).map(|v| v.take());
 }
 
 fn get_value_at_path<'a>(
--- a/apollo-router/tests/integration/file_upload.rs
+++ b/apollo-router/tests/integration/file_upload.rs
@@ -23,6 +23,125 @@ macro_rules! make_handler {
 }
 
 #[tokio::test(flavor = "multi_thread")]
+async fn it_uploads_file_to_subgraph() -> Result<(), BoxError> {
+    use reqwest::multipart::Form;
+    use reqwest::multipart::Part;
+
+    const FILE: &str = "Hello, world!";
+    const FILE_NAME: &str = "example.txt";
+
+    let request = Form::new()
+        .part(
+            "operations",
+            Part::text(
+                serde_json::json!({
+                    "query": "mutation SomeMutation($file: Upload) {
+                        file: singleUpload(file: $file) { filename body }
+                    }",
+                    "variables": { "file": null },
+                })
+                .to_string(),
+            ),
+        )
+        .part(
+            "map",
+            Part::text(serde_json::json!({ "0": ["variables.file"] }).to_string()),
+        )
+        .part("0", Part::text(FILE).file_name(FILE_NAME));
+
+    async fn subgraph_handler(
+        mut request: http::Request<hyper::Body>,
+    ) -> impl axum::response::IntoResponse {
+        let boundary = request
+            .headers()
+            .get(CONTENT_TYPE)
+            .and_then(|v| multer::parse_boundary(v.to_str().ok()?).ok())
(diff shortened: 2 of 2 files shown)


apollo-router/tests/integration/file_upload.rs
```diff
@@ -22,6 +22,125 @@ macro_rules! make_handler {
     }
 }
 
+#[tokio::test(flavor = "multi_thread")]
+async fn it_uploads_file_to_subgraph() -> Result<(), BoxError> {
```
IvanGoncharov, line 26: My best effort is writing tests that test stuff that is sent to the subgraph.
I chose not to use the existing test harness and made this test as simple as possible.
My initial plan was to update the existing test to snapshot stuff sent to the subgraph, but I failed to finish it in the time allocated for this fix, so I just wrote a super basic test that tests the subgraph request in the most basic scenario.

apollo-router/tests/integration/file_upload.rs
```diff
@@ -1043,7 +1162,7 @@ mod helper {
         pub body: Option<String>,
     }
 
-    #[derive(Serialize, Deserialize)]
+    #[derive(Debug, Serialize, Deserialize)]
```
IvanGoncharov, line 1165: helped me during debugging

apollo-router/tests/integration/file_upload.rs
```diff
@@ -1168,10 +1287,8 @@ mod helper {
             return Err(FileUploadError::UnexpectedFile);
         }
 
-        let field_name: String = map
-            .into_keys()
-            .take(1)
-            .next()
+        let (field_name, _) = map
+            .last_key_value()
```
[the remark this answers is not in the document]
goto-bus-stop: the equivalent of the existing code is `.first_key_value()`, i think? doesn't make a difference if it's just 1 of course...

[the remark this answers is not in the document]
IvanGoncharov: Nice 👍 
I was looking into the BTreeMap source but missed that method.
Need to learn to read docs instead of source code :(

apollo-router/src/plugins/file_uploads/mod.rs
```diff
@@ -254,7 +254,7 @@ fn replace_value_at_path<'a>(
 
 // Removes value at path.
 fn remove_value_at_path<'a>(variables: &'a mut json_ext::Object, path: &'a [String]) {
-    let _ = get_value_at_path(variables, path).take();
+    let _ = get_value_at_path(variables, path).map(|v| v.take());
```
IvanGoncharov, line 257: Previously, it just replaced the reference to the value inside the option returned by `get_value_at_path`, so no mutation of `variables` occurred.
Now, if `variables` contain a value, it is correctly replaced with `null`.

goto-bus-stop: Maybe we can make this a bit more explicit by doing `*v = Value::Null`? it's a bit confusing to read it like this

goto-bus-stop (approved):
Nits aside, this makes sense to me 👍🏻

apollo-router/src/plugins/file_uploads/mod.rs
```diff
@@ -254,7 +254,7 @@ fn replace_value_at_path<'a>(
 
 // Removes value at path.
 fn remove_value_at_path<'a>(variables: &'a mut json_ext::Object, path: &'a [String]) {
-    let _ = get_value_at_path(variables, path).take();
+    let _ = get_value_at_path(variables, path).map(|v| *v = serde_json_bytes::Value::Null);
```
BrynCooke, line 257: I think this would better communicate intent
```suggestion
    if let Some(v) = get_value_at_path(variables, path) {
        *v = serde_json_bytes::Value::Null;
    }
```

BrynCooke (approved):
Approved, but please take in my suggestion

apollo-router/src/plugins/file_uploads/mod.rs
IvanGoncharov: Love it 😍 
Both you and @goto-bus-stop normalized writing procedural code in Rust in your comments.
I was pushing myself to use a functional style in Rust, which I'm not fond of.
Thanks a lot 👍

PR #6294, opened 2024-11-20
extract context onto interface and union

clenfest:
Refactoring to avoid copypasta as much as possible, though there's still room to improve. Also some changes in can_satisfy_conditions to accomodate union.

### DIFF
--- a/apollo-federation/src/supergraph/mod.rs
+++ b/apollo-federation/src/supergraph/mod.rs
@@ -1000,7 +1000,7 @@ fn extract_interface_type_content(
                 }
             }
         }
-                
+        
         let context = supergraph_schema
             .metadata()
             .and_then(|metadata| metadata.for_identity(&Identity::context_identity()))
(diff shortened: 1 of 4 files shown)


apollo-federation/src/supergraph/mod.rs
```diff
@@ -1696,6 +1698,140 @@ lazy_static! {
     };
 }
 
+trait CompositeType {
```
TylerBloom, line 1701: I like this, and I think we (as in the whole team) should look into applying this pattern across the crate. Just a note for the future.

apollo-federation/src/supergraph/mod.rs
```diff
@@ -1696,6 +1698,140 @@ lazy_static! {
...
+        directive: Component<Directive>,
+    ) -> Result<(), FederationError> {
+        match pos {
+            CompositeTypeDefinitionPosition::Union(pos) => {
+                pos.insert_directive(schema, directive.into())?;
+                return Ok(());
+            }
+            _ => Err(internal_error!("an error")),
+        }
+    }
```
TylerBloom, line 1728: Is there error necessary here? You don't actually use `&self` in the method. Honestly, this looks like something that could be a method on `CompositeTypeDefinitionPosition`.

TylerBloom (approved):
I think the new trait can be trimmed down, but, other than that, this looks good.

PR #6297, opened 2024-11-20
Validate variables in JSON selection

pubmodmatt:
* Return the common `VariableReference` type from JSON selections
* Use the common variable code to validate variable references in JSON selection
* Refactor the variable parsing code into a `parser` submodule

Screenshot 2024-11-20 at 7 07 23 AM

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-federation/src/sources/connect/json_selection/known_var.rs
+++ b/apollo-federation/src/sources/connect/json_selection/known_var.rs
@@ -1,18 +1,12 @@
-use std::borrow::Cow;
+use std::str::FromStr;
 
 #[cfg(test)]
 use super::location::WithRange;
 use crate::sources::connect::variable::Namespace;
-use crate::sources::connect::variable::VariableNamespace;
-use crate::sources::connect::variable::VariablePathPart;
-use crate::sources::connect::variable::VariableReference;
 
 #[derive(PartialEq, Eq, Clone, Hash)]
 pub(crate) enum KnownVariable {
-    This,
-    Args,
-    Config,
-    Status,
+    Identifier(Namespace),
     Dollar,
     AtSign,
 }
@@ -20,22 +14,15 @@ pub(crate) enum KnownVariable {
 impl KnownVariable {
     pub(crate) fn from_str(var_name: &str) -> Option<Self> {
         match var_name {
-            "$this" => Some(Self::This),
-            "$args" => Some(Self::Args),
-            "$config" => Some(Self::Config),
-            "$status" => Some(Self::Status),
             "$" => Some(Self::Dollar),
             "@" => Some(Self::AtSign),
-            _ => None,
+            s => Namespace::from_str(s).ok().map(Self::Identifier),
         }
     }
 
     pub(crate) fn as_str(&self) -> &'static str {
         match self {
-            Self::This => "$this",
-            Self::Args => "$args",
-            Self::Config => "$config",
-            Self::Status => "$status",
+            Self::Identifier(namespace) => namespace.as_str(),
             Self::Dollar => "$",
             Self::AtSign => "@",
         }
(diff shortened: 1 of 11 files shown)


apollo-federation/src/sources/connect/validation/snapshots/validation_tests@invalid_namespace_in_body_selection.graphql.snap
```diff
@@ -0,0 +1,14 @@
...
+source: apollo-federation/src/sources/connect/validation/mod.rs
+expression: "format!(\"{:#?}\", errors)"
+input_file: apollo-federation/src/sources/connect/validation/test_data/invalid_namespace_in_body_selection.graphql
+---
+[
+    Message {
+        code: InvalidJsonSelection,
+        message: "`@connect(http: {body:})` on `Mutation.createUser` contains an invalid variable namespace `$status`, must be one of $args, $config, $context, $this",
+        locations: [
+            14:9..14:16,
```
pubmodmatt, line 11: Oof, looks like this test failed on windows since the range is 14:8..14:15 there. Will need to investigate this.

apollo-federation/src/sources/connect/validation/selection.rs
```diff
@@ -96,13 +114,59 @@ pub(super) fn validate_body_selection(
...
+    for reference in selection
+        .external_var_paths()
+        .into_iter()
+        .flat_map(|var_path| var_path.variable_reference())
+    {
+        if !namespaces.contains(&reference.namespace.namespace) {
+            return Some(Err(Message {
+                code: Code::InvalidJsonSelection,
+                message: format!(
+                    "{coordinate} contains an invalid variable namespace `{namespace}`, must be one of {available}",
```
lennyburdette, line 149: is "namespace" a term we want to expose to users? at the moment, they're documented just as ["Variables"](https://www.apollographql.com/docs/graphos/schema-design/connectors/selection#variables)

apollo-federation/src/sources/connect/validation/selection.rs
```diff
@@ -35,7 +41,18 @@ pub(super) fn validate_selection(
     schema: &SchemaInfo,
     seen_fields: &mut IndexSet<(Name, Name)>,
 ) -> Result<(), Message> {
-    let (selection_arg, json_selection) = get_json_selection(coordinate, schema)?;
+    let (selection_arg, json_selection, selection_node) = get_json_selection(coordinate, schema)?;
+
+    if let Some(value) = validate_selection_variables(
```
lennyburdette, line 46: just curious: why this pattern and not `if let Err(error) = validate_selection_variables(...) { return Err(error) }`?

apollo-federation/src/sources/connect/validation/selection.rs
```diff
@@ -96,13 +114,59 @@ pub(super) fn validate_body_selection(
...
+    coordinate: String,
+    selection: &JSONSelection,
+) -> Option<Result<(), Message>> {
+    let namespaces: HashSet<Namespace> = expression_context.available_namespaces().collect();
+    for reference in selection
+        .external_var_paths()
+        .into_iter()
+        .flat_map(|var_path| var_path.variable_reference())
+    {
+        if !namespaces.contains(&reference.namespace.namespace) {
```
lennyburdette, line 145: this doesn't yet use `VariableResolver` to validate the variable's path, right?

lennyburdette (approved):

apollo-federation/src/sources/connect/validation/selection.rs
pubmodmatt: Good call, I'll add that.

apollo-federation/src/sources/connect/validation/selection.rs
pubmodmatt: Ugh, I refactored this code out to a method using the IDE refactoring, and it did... whatever this is. I missed that weird `Option<Result<...>>`. Will fix.

apollo-federation/src/sources/connect/validation/snapshots/validation_tests@invalid_namespace_in_body_selection.graphql.snap
pubmodmatt: The problem here was that JSON `JSONSelection::parse` takes `&str` as the input parameter. I had mistakenly passed in the result of calling `as_str()` on `Node<Value>`, instead of a the result of calling `as_str()` on a `GraphQLString` created from the `Node<Value>`. The difference is that the `Node<Value>` is from the parsed GraphQL AST, which has the [GraphQL block string value rules](https://spec.graphql.org/October2021/#BlockStringValue()) applied - so it has `\r` removed, and has the common indent whitespace removed from the start of the lines. The `GraphQLString`, on the other hand, points directly to the raw string data from the schema file, which has the `\r` and all leading whitespace in tact.

When doing validation, or anything else that will use the location offsets returned by the parser, you must parse the raw string data from a `GraphQLString`, rather than the GraphQL AST data. If you do the latter, any location offsets returned by the parser will be completely invalid with respect to the original schema file.

I've fixed the mistake, but there may be something better we can do here. Rust has the ability to make bugs like this impossible. But we aren't taking advantage of the language type system, and are instead just working with bare `&str`, which has no type information to enforce what that reference is actually pointing to. So the compiler can't help prevent this kind of mistake. This isn't particular to JSON selection - the URL template and header parsers also take `&str`.

apollo-federation/src/sources/connect/validation/selection.rs
```diff
@@ -96,6 +120,56 @@ pub(super) fn validate_body_selection(
...
+                    namespace = reference.namespace.namespace,
+                    available = expression_context.namespaces_joined(),
+                ),
+                locations: selection_str.line_col_for_subslice(reference.location, schema)
+                    .into_iter()
+                    .collect(),
+            });
+        }
+
+        variable_resolver
```
lennyburdette, line 162: 💯

apollo-federation/src/sources/connect/json_selection/known_var.rs
```diff
@@ -1,41 +1,28 @@
...
 
 #[derive(PartialEq, Eq, Clone, Hash)]
 pub(crate) enum KnownVariable {
-    This,
-    Args,
-    Config,
-    Status,
+    Identifier(Namespace),
     Dollar,
     AtSign,
```
benjamn, line 11:
```suggestion
    External(Namespace),
    Dollar,
    AtSign,
```
Could we call this variant something like `::External`, to reflect that these namespace variables are provided from the outside, in contrast to `$` and `@` which always refer internally within the data?

My gripe with `Identifier` is it's representing a variable like `$this` which is intentionally not a valid `Identifier` according to the grammar.

benjamn (approved):

lennyburdette (approved):

PR #6299, opened 2024-11-20
Fix list test and adds rewrite validation to context tests

clenfest:


apollo-federation/tests/query_plan/build_query_plan_tests/context.rs
```diff
@@ -860,7 +962,7 @@ fn set_context_test_accesses_a_different_top_level_query() {
 }
 
 #[test]
-fn set_context_variable_is_a_list() {
+fn set_context_one_subgraph() {
```
clenfest, line 965: I kept this test and renamed it because I actually was seeing a problem with it previously when in the other branch, so I thought it might be exercising something I don't quite understand. We can delete it later if necessary.

TylerBloom (approved):

PR #6305, opened 2024-11-21
Change naming convention for contextual arguments

clenfest:


clenfest:
Note that the subgraph indices are different than in JS. In JS it was the subgraph index, but here it's the subgraph index of those that use contextual arguments. As long as there are no collisions between subgraphs, we're good.

TylerBloom (approved):

apollo-bot2:

PR #6307, opened 2024-11-21
fix(fed): filter out invalid fragment directives instead of error

dariuszkuc:
Operation was already validated so the passed in document should be `Valid`. Instead of throwing an explicit error if fragment spread dectives are not available in the api schema OR if they are not applicable on an inline fragment, we should be filtering those instead.

This specifically addresses an issue when users disable `incremental_delivery` support (i.e. `@defer`) but pass in a query with `@defer`. Instead of throwing an exception we should be gracefully handling it as all `@defer` usages will be normalized away (i.e. removed) if user opted out of `incremental_delivery` feature.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-federation/src/operation/mod.rs
+++ b/apollo-federation/src/operation/mod.rs
@@ -2700,24 +2700,25 @@ impl InlineFragmentSelection {
         fragment_spread_selection: &Arc<FragmentSpreadSelection>,
     ) -> Result<InlineFragmentSelection, FederationError> {
         let schema = fragment_spread_selection.spread.schema.schema();
-        for directive in fragment_spread_selection.spread.directives.iter() {
-            let Some(definition) = schema.directive_definitions.get(&directive.name) else {
-                return Err(FederationError::internal(format!(
-                    "Undefined directive {}",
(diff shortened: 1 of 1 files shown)


dariuszkuc:
This is an alternative to https://github.com/apollographql/router/pull/6308 which assumes that passed in operation was already validated so instead of throwing an error, we can just filter out those invalid directive entries.

dariuszkuc:
Since we are now filtering directives with invalid locations - we now fail [one of the tests](https://github.com/apollographql/router/blob/dev/apollo-federation/src/operation/tests/mod.rs#L1693-L1705). 

What should be the "ideal" behavior?
- filter directives that are not defined in the schema BUT still validate location
VS
- filter out directives not defined in the schema OR the ones with invalid locations

dariuszkuc:
We should rethink how we handle this logic. Closing it for now.

PR #6311, opened 2024-11-21
Fix bug with `@context` state tracking in `GraphPath::add()`

sachindshinde:
This PR fixes a Rust-only bug in `GraphPath::add()` where it didn't properly push to `context_to_selection` and `parameter_to_context` in the case of a `None` edge, causing the `GraphPath` iterator to prematurely end due to `zip` semantics.


goto-bus-stop (approved):

goto-bus-stop:
Test failures are separate, I'll fix them in #6310.

PR #6313, opened 2024-11-21
Response formatting follow-up

garypen:
I've done about 1/2 tests. So far that has taken me about 2 hours.

Follows up https://github.com/apollographql/router/pull/5787


abernix:
@garypen Thoughts on this PR?  Keep or close?

garypen:
If we ever unrevert #5787 , then we will need to do this work. Just tagging so we don't lose it.

PR #6314, opened 2024-11-21
Revert "add errors for response validation (#5787)"

goto-bus-stop:
This reverts commit 39c6c03ed9f15c3872bcad083e1560ddbb1b6d3e.

#5787 contained a bug that reported a GraphQL field error when a subgraph replied with `null` for an enum field, *even if that field is declared as nullable*. While that is easy to address, we found in #6313 that many of our tests were not properly checking for error cases, so the work became a bit more sprawling and not able to be completed for 1.58.0.

After finding this bug, we are also not as confident on the change as a whole, and if it breaks, the way it breaks is very bad. Hence the choice to revert it out of 1.58.0.

To land this again, I believe we need stronger testing, including all possible permutations of nullable/non-nullable, list/non-list fields, and all permutations of GraphQL and JSON types. Additionally, we may need to put the new error handling in router 2.0 only. I think we should land the tests in both 1.x and 2.0, to nail down the current behaviour.

#5787 *also* contained minor fixes to existing error messages. We can reintroduce those fixes in 1.x along with the expanded tests.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


bnjjj (dismissed):

goto-bus-stop:
One new test is failing because the revert reintroduces a bug in the error message that we report. I'll update that test to expect the wrong thing, as it's not critical

BrynCooke (dismissed):

BrynCooke (approved):

abernix:
> if we ever unrevert this, we should look at #6313

PR #6316, opened 2024-11-21
fix: propagate evaluated plans limit and paths limit to the native query planner

goto-bus-stop:
The experimental `max_evaluated_plans` and `paths_limit` options were only propagated to the legacy JS planner, not the native query planner.

Now they work with both planners.

Adds integration tests for the `apollo.router.query_planning.plan.evaluated_plans` metric, with and without `max_evaluated_plans` set, with the legacy and the native query planners.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [x] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-875]: https://apollographql.atlassian.net/browse/ROUTER-875?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apollo-router/src/configuration/mod.rs
+++ b/apollo-router/src/configuration/mod.rs
@@ -5,6 +5,7 @@ use std::io::BufReader;
 use std::iter;
 use std::net::IpAddr;
 use std::net::SocketAddr;
+use std::num::NonZeroU32;
 use std::num::NonZeroUsize;
 use std::str::FromStr;
 use std::sync::Arc;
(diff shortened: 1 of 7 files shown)


apollo-router/src/configuration/mod.rs
```diff
@@ -416,16 +417,31 @@ impl Configuration {
...
-                apollo_federation::query_plan::query_planner::QueryPlanIncrementalDeliveryConfig {
-                    enable_defer: self.supergraph.defer_support,
-                },
+            incremental_delivery: QueryPlanIncrementalDeliveryConfig {
+                enable_defer: self.supergraph.defer_support,
+            },
             type_conditioned_fetching: self.experimental_type_conditioned_fetching,
-            debug: Default::default(),
+            debug: QueryPlannerDebugConfig {
+                bypass_planner_for_single_subgraph: false,
```
dariuszkuc, line 441: we probably also pass this down ([link](https://github.com/apollographql/router/blob/dev/apollo-federation/src/query_plan/query_planner.rs#L370))?

goto-bus-stop: there is no router config to enable this option as far as i know. my general feeling is that the option could be removed entirely from the query planner, but 🤷🏻‍♀️

goto-bus-stop:
Wait. Why are there two QP configurations?

https://github.com/apollographql/router/blob/dev/apollo-router/src/query_planner/bridge_query_planner.rs#L169-L182

apollo-router/src/configuration/mod.rs
dariuszkuc: Hmm it is something that we do support in the gateway (which I believe was done as an experimental feature to skip the expensive QP computation) - I'll open a ticket to figure out what we want to do on RS side of things

lrlna:
@goto-bus-stop are you continuing this work still? is it just missing an integration test you wanted to do?

>Wait. Why are there two QP configurations?

Looks like we are not using the `rust_query_planner_config` at all, or at least not in the same way we are using the neighbouring `js_query_planner_config` function. Probably something got lost along the way.

goto-bus-stop:
`rust_query_planner_config` is used for a cache key. so it's important that we have only one source of truth for actual QP & the cache key.

the main thing this needs is a test indeed.

apollo-router/src/configuration/mod.rs
goto-bus-stop: I think it's risky to support this if in the future we move support for authentication directives etc. into the query planner

lrlna (approved):

apollo-router/src/configuration/mod.rs
```diff
@@ -416,16 +417,31 @@ impl Configuration {
...
+        use apollo_federation::query_plan::query_planner::QueryPlannerConfig;
+        use apollo_federation::query_plan::query_planner::QueryPlannerDebugConfig;
+
+        let max_evaluated_plans = self
+            .supergraph
+            .query_planning
+            .experimental_plans_limit
+            // Fails if experimental_plans_limit is zero; use our default.
+            .and_then(NonZeroU32::new)
+            .unwrap_or(NonZeroU32::new(10_000).expect("it is not zero"));
```
lrlna, line 430: doesn't QueryPlannerDebugConfig already set max_evaluated_plans to `10_000`? 🤔

goto-bus-stop: Only if you use `QueryPlannerDebugConfig::default()`. The property is required so we would have to do something like this to propagate it correctly:

```rust
let mut debug = QueryPlannerDebugConfig::default();
if let Some(plans_limit) = configuration.supergraph.query_planning.experimental_plans_limit {
    debug.max_evaluated_plans = NonZeroU32::new(plans_limit).unwrap_or(debug.max_evaluated_plans);
}
```
That's even more ceremony and if we use the `Default` impl, and a new option is added to `QueryPlannerDebugConfig`, we could easily forget to pass it on from the router side.

By doing it like this we duplicate the default value but we are at least forced by the compiler to provide all the options intentionally.

lrlna: makes sense!

lrlna (approved):

PR #6323, opened 2024-11-22
fix(federation): fixed `Selection::conditions` to merge field's subselection's conditions when necessary

duckki:
In `Selection::conditions` method, which is the Rust port of JS function `conditionsOfSelection`, has following comment:

```rust
                Selection::Field(_) => {
                    // The sub-selections of this field don't affect whether we should query this
                    // field, so we explicitly do not merge them in.
                    //
                    // PORT_NOTE: The JS codebase merges the sub-selections' conditions in with the
                    // field's conditions when field's selections are non-boolean. This is arguably
                    // a bug, so we've fixed it here.
                    Ok(self_conditions)
                }
                Selection::InlineFragment(inline) => {
                    Ok(self_conditions.merge(inline.selection_set.conditions()?))
                }
```

The comment suggested that the JS code was buggy and it was fixed in Rust port. But, it turned out that there are cases where the `Selection::Field(...)` case should merge with its subselection-set's conditions.

Example:

```graphql
#!subgraph
type Query {
    test: T!
}

type T {
    id: ID!
    name: String!
    x: Int!
}

#!operation
query($v1: Boolean!) {
    test @include(if: $v1) {
        id @include(if: false)
    }
}
```

In the example above, the `test` selection's condition may become a `Boolean(false)`, influenced by its sub-selection-set.

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [x] Manual Tests


duckki:
I made this PR to match JS QP. But, now I doubt this is the right way to handle the query.

lrlna:
@duckki as a heads up, i changed the base of this commit since we just created the 1.58.1 branch and we plan on landing this in 1.58.1.

goto-bus-stop:
I'm not sure that we know what to do here in order to get it into 1.58.1?

I prefer the current Rust approach because it's doing less work 😇 .

duckki:
I think we don't have to do this. Initially, I tried to restore JS QP's behavior with this PR, but I found there's a bigger bug around condition handling in both JS QP and Rust QP. I'm going to propose a overall fix.

PR #6325, opened 2024-11-22
chore(federation): add unit tests and documentation for conditions handling

goto-bus-stop:
The `Conditions` structure represents the combined skip/include conditions of a GraphQL query path (and of a FetchDependencyGraphNode).

The storage of variable conditions looks inefficient. I tried some quick refactors to make it more efficient, which didn't hugely affect overall planning performance, even on plans that spend a lot of time computing conditions. Perhaps the representation isn't the biggest issue. I think we can still simplify it, though. I might tinker with that if I find some spare cycles in December.

This PR does not change the representation. I only added tests and documentation.

I also flipped the order of the `update_with()` argument. The order of the arguments disagreed with the name of the method. We were not updating the argument at all: we were updating `self`, removing conditions already handled by the argument.

### DIFF
--- a/apollo-federation/src/query_plan/conditions.rs
+++ b/apollo-federation/src/query_plan/conditions.rs
@@ -130,12 +130,6 @@ impl VariableConditions {
     }
 }
 
-#[derive(Debug, Clone, PartialEq)]
-pub(crate) struct VariableCondition {
-    variable: Name,
-    kind: ConditionKind,
-}
-
 impl Conditions {
     /// Create conditions from a map of variable conditions.
     ///
--- a/apollo-federation/src/query_plan/conditions.rs
+++ b/apollo-federation/src/query_plan/conditions.rs
@@ -42,11 +42,16 @@ impl Display for ConditionKind {
     }
 }
 
-/// This struct is meant for tracking whether a selection set in a `FetchDependencyGraphNode` needs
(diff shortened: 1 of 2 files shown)


apollo-federation/src/query_plan/conditions.rs
```diff
@@ -46,6 +53,28 @@ pub(crate) enum Conditions {
     Boolean(bool),
 }
 
+impl Display for Conditions {
```
goto-bus-stop, line 56: This is mostly for use in tests, though it might help with debugging as well

apollo-federation/src/query_plan/fetch_dependency_graph.rs
```diff
@@ -1779,7 +1779,7 @@ impl FetchDependencyGraph {
             .graph
             .node_weight_mut(node_index)
             .ok_or_else(|| FederationError::internal("Node unexpectedly missing"))?;
-        let conditions = handled_conditions.update_with(&node.selection_set.conditions);
+        let conditions = node.selection_set.conditions.update_with(&handled_conditions);
```
goto-bus-stop, line 1782: Only usage site of `update_with()`.

lrlna (approved):
I like `always` and `never` as an api to conditions!

PR #6326, opened 2024-11-22
fix(federation): avoid constantly recomputing conditions

goto-bus-stop:
While I was working on #6325, I figured I might as well tackle this related issue quickly.

Certain graph and query shapes cause the `FetchSelectionSet::conditions` to be recomputed very, very frequently. Now, they are only recomputed when required.

On most plans there is no noteworthy difference, but on plans with many large fetch nodes and many possible plans, the difference can be high-single-digit % in *total planning time*, which is very worthwhile.

When `OnceLock::get_or_try_init` is stabilised, this will be easier :)

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

[ROUTER-484]: https://apollographql.atlassian.net/browse/ROUTER-484?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apollo-federation/src/query_plan/fetch_dependency_graph.rs
+++ b/apollo-federation/src/query_plan/fetch_dependency_graph.rs
@@ -213,9 +213,10 @@ impl FetchIdGenerator {
 pub(crate) struct FetchSelectionSet {
     /// The selection set to be fetched from the subgraph.
     pub(crate) selection_set: Arc<SelectionSet>,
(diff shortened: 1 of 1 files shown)


lrlna (approved):
🏎️ 🏁

PR #6328, opened 2024-11-22
Ensure cost directives are picked up when not explicitly imported

tninesling:
With the recent composition changes, importing `@cost` results in a supergraph schema with the cost specification import at the top. The `@cost` directive itself is not explicitly imported, as it's expected to be available based on being the default export from the cost specification. In contrast, uses of `@listSize` to translate to an explicit import in the supergraph.

Old SDL link
```
@link(
    url: "https://specs.apollo.dev/cost/v0.1"
    import: ["@cost", "@listSize"]
)
```

New SDL link
```
@link(url: "https://specs.apollo.dev/cost/v0.1", import: ["@listSize"])
```

This exposed the brittle initial implementation of cost directive handling, which built up a map of directive name aliases from import statements. The core change made to address this issue is to remove the directive name map and replace it with calls to `SpecDefinition::directive_name_in_schema`, which is equivalent to the change we made on the composition side.

To avoid passing these names around during directive parsing, I have moved the directive parsing code into the `CostSpecDefinition`. This means the demand control plugin does not need to be concerned with the directives names in each schema.

---

**Checklist**

- [X] Performance impact assessed and acceptable
- Tests added and passing[^3]
    - [X] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-federation/src/link/cost_spec_definition.rs
```diff
@@ -119,35 +120,36 @@ impl CostSpecDefinition {
...
         schema: &FederationSchema,
         arguments: Vec<Node<Argument>>,
     ) -> Result<Directive, FederationError> {
-        let name = self
-            .directive_name_in_schema(schema, &COST_DIRECTIVE_NAME_IN_SPEC)?
-            .unwrap_or(COST_DIRECTIVE_NAME_DEFAULT);
+        let name =
+            Self::cost_directive_name(schema)?.ok_or_else(|| SingleFederationError::Internal {
+                message: "The \"@cost\" directive is undefined in the target schema".to_string(),
+            })?;
```
lrlna, line 129: we have a convenient macro for this:

```suggestion
            Self::cost_directive_name(schema)?.ok_or_else(|| 
                internal_error!("The \"@cost\" directive is undefined in the target schema")
            )?;
```

apollo-federation/src/link/cost_spec_definition.rs
```diff
@@ -119,35 +120,36 @@ impl CostSpecDefinition {
...
     ) -> Result<Directive, FederationError> {
-        let name = self
-            .directive_name_in_schema(schema, &LIST_SIZE_DIRECTIVE_NAME_IN_SPEC)?
-            .unwrap_or(LIST_SIZE_DIRECTIVE_NAME_DEFAULT);
+        let name = Self::list_size_directive_name(schema)?.ok_or_else(|| {
+            SingleFederationError::Internal {
+                message: "The \"@listSize\" directive is undefined in the target schema"
+                    .to_string(),
+            }
+        })?;
```
lrlna, line 143: same here:
```suggestion
        let name = Self::list_size_directive_name(schema)?.ok_or_else(||
                internal_error!("The \"@listSize\" directive is undefined in the target schema")
        )?;
```

apollo-router/src/plugins/demand_control/mod.rs
```diff
@@ -163,6 +166,10 @@ impl IntoGraphQLErrors for DemandControlError {
                 .extension_code(self.code())
                 .message(self.to_string())
                 .build()]),
+            DemandControlError::FederationError(_) => Ok(vec![graphql::Error::builder()
+                .extension_code(self.code())
+                .message(self.to_string())
+                .build()]),
```
lrlna, line 172: The only two errors coming from federation would be undefined directive definitions, right? do we want to consider that a federation error or a graphql validation error?

lrlna (approved):
I like that we are moving this logic to `apollo-federation` crate. Makes this a lot cleaner! I have a couple of small suggestion and a question, but neither is blocking. 

And overall question though, is this getting an update on the JS side, and do we need to bump router-bridge still? If so, we should get that in today, since the rc goes on wednesday.

tninesling:
> And overall question though, is this getting an update on the JS side, and do we need to bump router-bridge still? If so, we should get that in today, since the rc goes on wednesday.

This change was made on the JS side in https://github.com/apollographql/federation/pull/3136, which was release in federation 2.9.3, so it's already in the current bridge.

apollo-router/src/plugins/demand_control/mod.rs
tninesling: The only case this should arise is when we get an unexpected failure in `spec.directive_name_in_schema` during execution. Realistically, this should never actually arise as an execution error because we call the same thing during subgraph extraction on startup. I don't think the spec definitions guarantees that error to specifically be a validation issue, so we probably need to keep the name here generic.

PR #6286, opened 2024-11-19
add entity caching invalidation configuration metrics

bnjjj:

Issue #6249, opened 2024-11-08
Running simultaneous introspection queries frequently fails with `value retrieval failed: unhandled planner result`


matt-kunze-nutrien:
## Describe the bug

When upgrading to v1.57.0 or newer we're seeing errors from introspection queries if multiple requests are triggered in a short period of time. This is causing errors with our `graphql-codegen` scripts that end up fetching the schema for multiple projects during the build

## To Reproduce

Given a file with the following introspection query:

introspection.json
```json
{
  "query": "query IntrospectionQuery{__schema{queryType{name}mutationType{name}subscriptionType{name}types{...FullType}directives{name description locations args(includeDeprecated:true){...InputValue}}}}fragment FullType on __Type{kind name description fields(includeDeprecated:true){name description args(includeDeprecated:true){...InputValue}type{...TypeRef}isDeprecated deprecationReason}inputFields(includeDeprecated:true){...InputValue}interfaces{...TypeRef}enumValues(includeDeprecated:true){name description isDeprecated deprecationReason}possibleTypes{...TypeRef}}fragment InputValue on __InputValue{name description type{...TypeRef}defaultValue isDeprecated deprecationReason}fragment TypeRef on __Type{kind name ofType{kind name ofType{kind name ofType{kind name ofType{kind name ofType{kind name ofType{kind name ofType{kind name ofType{kind name ofType{kind name}}}}}}}}}}"
}
```

if I run two simultaneous introspection queries at least one will reliably fail:

```sh
❯ curl -s -H 'accept: application/json' -H 'content-type: application/json' 'https://<router address>' -d "@./introspection.json" | jq .errors & \
      curl -s -H 'accept: application/json' -H 'content-type: application/json' 'https://<router address>' -d "@./introspection.json" | jq .errors
[
  {
    "message": "value retrieval failed: unhandled planner result",
    "extensions": {
      "code": "INTERNAL_SERVER_ERROR"
    }
  }
]
null
```

Running a single introspection query in isolation is working as expected

## Expected behavior

Both queries should reliably succeed. This is the behavior we have seen with v1.56.0 and prior

## Desktop (please complete the following information):

I can reproduce this both in a deployed environment (ECS using `--platform=linux/amd64 ghcr.io/apollographql/router:v1.57.1`) and locally running router (M1 mac)

matt-kunze-nutrien:
I would imagine this is related to the switch to a native introspection implementation: https://github.com/apollographql/router/pull/6139

PR #6331, opened 2024-11-26
prep release: v1.58.0

abernix:
> **Note**
>
> When approved, this PR will merge into **the `1.58.0` branch** which will — upon being approved itself — merge into `main`.
>
>  - Version bumps
>  - That it targets the right release branch (`1.58.0` in this case!).
>
---
### Support DNS resolution strategy configuration ([PR #6109](https://github.com/apollographql/router/pull/6109))

The router now supports a configurable DNS resolution strategy for the URLs of coprocessors and subgraphs.
The new option is called `dns_resolution_strategy` and supports the following values:
* `ipv4_only` - Only query for `A` (IPv4) records.
* `ipv6_only` - Only query for `AAAA` (IPv6) records.
* `ipv4_and_ipv6` - Query for both `A` (IPv4) and `AAAA` (IPv6) records in parallel.
* `ipv6_then_ipv4` - Query for `AAAA` (IPv6) records first; if that fails, query for `A` (IPv4) records.
* `ipv4_then_ipv6`(default) - Query for `A` (IPv4) records first; if that fails, query for `AAAA` (IPv6) records.

You can change the DNS resolution strategy applied to a subgraph's URL:

```yaml title="router.yaml"
traffic_shaping:
  all:
    dns_resolution_strategy: ipv4_then_ipv6

```

You can also change the DNS resolution strategy applied to a coprocessor's URL:

```yaml title="router.yaml"
coprocessor:
  url: http://coprocessor.example.com:8081
  client:
    dns_resolution_strategy: ipv4_then_ipv6

```

By [@IvanGoncharov](https://github.com/IvanGoncharov) in https://github.com/apollographql/router/pull/6109

### Configuration options for HTTP/1 max headers and buffer limits ([PR #6194](https://github.com/apollographql/router/pull/6194))

This update introduces configuration options that allow you to adjust the maximum number of HTTP/1 request headers and the maximum buffer size allocated for headers.

By default, the router accepts HTTP/1 requests with up to 100 headers and allocates ~400 KiB of buffer space to store them. If you need to handle requests with more headers or require a different buffer size, you can now configure these limits in the router's configuration file:
```yaml
limits:
  http1_request_max_headers: 200
  http1_request_max_buf_size: 200kib
```

If you are using the router as a Rust crate, the `http1_request_max_buf_size` option requires the `hyper_header_limits` feature and also necessitates using Apollo's fork of the Hyper crate until the [changes are merged upstream](https://github.com/hyperium/hyper/pull/3523).
You can include this fork by adding the following patch to your Cargo.toml file:
```toml
[patch.crates-io]
"hyper" = { git = "https://github.com/apollographql/hyper.git", tag = "header-customizations-20241108" }
```

By [@IvanGoncharov](https://github.com/IvanGoncharov) in https://github.com/apollographql/router/pull/6194

### Compress subgraph operations by generating fragments ([PR #6013](https://github.com/apollographql/router/pull/6013))

The router now compresses operations sent to subgraphs by default by generating fragment
definitions and using them in the operation.

This change enables `generate_query_fragments` by default while disabling `experimental_reuse_query_fragments`. When enabled, `experimental_reuse_query_fragments` attempts to intelligently reuse the fragment definitions
from the original operation. However, fragment generation with `generate_query_fragments` is much faster also produces better outputs in most cases.

If you are relying on the shape of fragments in your subgraph operations or tests, you can opt out of the new algorithm with the configuration below. 

> Note: The subgraph operations generated by the query planner are not guaranteed consistent release over release. We strongly recommend against relying on the shape of planned subgraph operations, as new router features and optimizations will continuously affect it. We plan to remove `experimental_reuse_query_fragments` in a future release.

```yaml
supergraph:
  generate_query_fragments: false
  experimental_reuse_query_fragments: true
```

By [@lrlna](https://github.com/lrlna) in https://github.com/apollographql/router/pull/6013

### Add subgraph request id ([PR #5858](https://github.com/apollographql/router/pull/5858))

The router now supports a subgraph request ID that is a unique string identifying a subgraph request and response. It allows plugins and coprocessors to keep some state per subgraph request by matching on this ID. It's available in coprocessors as `subgraphRequestId` and Rhai scripts as `request.subgraph.id` and `response.subgraph.id`.

By [@Geal](https://github.com/Geal) in https://github.com/apollographql/router/pull/5858

### Add `extensions.service` for all subgraph errors ([PR #6191](https://github.com/apollographql/router/pull/6191))

For improved debuggability, the router now supports adding a subgraph's name as an extension to all errors originating from the subgraph.

If `include_subgraph_errors` is `true` for a particular subgraph, all errors originating in this subgraph will have the subgraph's name exposed as a `service` extension.

You can enable subgraph errors with the following configuration:
```yaml title="router.yaml"
include_subgraph_errors:
  all: true # Propagate errors from all subgraphs
```
> Note: This option is enabled by default by the router's [dev mode](https://www.apollographql.com/docs/graphos/reference/router/configuration#dev-mode-defaults).

Consequently, when a subgraph returns an error, it will have a `service` extension with the subgraph name as its value. The following example shows the extension for a `products` subgraph:

```json
{
  "data": null,
  "errors": [
    {
      "message": "Invalid product ID",
      "path": [],
      "extensions": {
        "service": "products"
      }
    }
  ]
}
```

By [@IvanGoncharov](https://github.com/IvanGoncharov) in https://github.com/apollographql/router/pull/6191

### Add `@context` support in the native query planner ([PR #6310](https://github.com/apollographql/router/pull/6310))

The [`@context`](https://www.apollographql.com/docs/graphos/reference/federation/directives#context) feature is now available in the native query planner.
This brings the native query planner to feature parity with the legacy query planner for all Federation v2 graphs.

By [@clenfest](https://github.com/clenfest), [@TylerBloom](https://github.com/TylerBloom) in https://github.com/apollographql/router/pull/6310

### Remove noisy demand control logs ([PR #6192](https://github.com/apollographql/router/pull/6192))

Demand control no longer logs warnings when a subgraph response is missing a requested field.

By [@tninesling](https://github.com/tninesling) in https://github.com/apollographql/router/pull/6192

### Renamed headers' original values can again be propagated ([PR #6281](https://github.com/apollographql/router/pull/6281))

[PR #4535](https://github.com/apollographql/router/pull/4535) introduced a regression where the following header propagation config would not work:

```yaml
headers:
- propagate:
    named: a
    rename: b
- propagate:
    named: a
    rename: c
```

The goal of the original PR was to prevent multiple headers from being mapped to a single target header. However, it did not consider renames and instead prevented multiple mappings from the same source header. 
The router now propagates headers properly and ensures that a target header is only propagated to once.

By [@BrynCooke](https://github.com/BrynCooke) in https://github.com/apollographql/router/pull/6281

### Introspection response deduplication should always produce results ([Issue #6249](https://github.com/apollographql/router/issues/6249))

To reduce CPU usage, query planning and introspection queries are deduplicated. In some cases, deduplicated introspection queries were not receiving their result. This issue has been fixed, and the router now sends results in all cases.

By [@Geal](https://github.com/Geal) in https://github.com/apollographql/router/pull/6257

### Don't log response data upon notification failure for subgraph batching ([PR #6150](https://github.com/apollographql/router/pull/6150))

For a subgraph batching operation, the router now doesn't log the entire subgraph response when failing to notify a waiting batch participant. This saves the router from logging the large amount of data (PII and/or non-PII data) that a subgraph response may contain.

By [@garypen](https://github.com/garypen) in https://github.com/apollographql/router/pull/6150

### Move heavy computation to a thread pool with a priority queue ([PR #6247](https://github.com/apollographql/router/pull/6247))

The router now avoids blocking threads when executing asynchronous code by using a thread pool with a priority queue.

This improves the performance of the following components can take non-trivial amounts of CPU time:

* GraphQL parsing
* GraphQL validation
* Query planning
* Schema introspection

In order to avoid blocking threads that execute asynchronous code,
they are now run in a new thread pool with a priority queue. The size of the thread pool is based on the number of available CPU cores.

The thread pool replaces the router's prior implementation that used Tokio’s [`spawn_blocking`](https://docs.rs/tokio/latest/tokio/task/fn.spawn_blocking.html).

`apollo.router.compute_jobs.queued` is a new gauge metric for the number of items in the thread pool's priority queue. 

> Note: when the native query planner is enabled, the dedicated queue of the legacy query planner is no longer used, so the `apollo.router.query_planning.queued` metric is no longer emitted.

By [@SimonSapin](https://github.com/SimonSapin) in https://github.com/apollographql/router/pull/6247

### Limit the amount of GraphQL validation errors returned per response ([PR #6187](https://github.com/apollographql/router/pull/6187))

When an invalid query is submitted, the router now returns at most one hundred GraphQL parsing and validation errors in a response. This prevents generating too large of a response for a nonsensical document.

By [@goto-bus-stop](https://github.com/goto-bus-stop) in https://github.com/apollographql/router/pull/6187

### Remove placeholders from file upload query variables ([PR #6293](https://github.com/apollographql/router/pull/6293))

Previously, file upload query variables in subgraph requests incorrectly contained internal placeholders.
According to the [GraphQL Multipart Request Spec](https://github.com/jaydenseric/graphql-multipart-request-spec?tab=readme-ov-file#multipart-form-field-structure), these variables should be set to null.
This issue has been fixed by ensuring that the router complies with the specification and improving compatibility with subgraphs handling file uploads.

By [@IvanGoncharov](https://github.com/IvanGoncharov) in https://github.com/apollographql/router/pull/6293

### Overhead processing metrics should exclude subgraph response time when deduplication is enabled ([PR #6207](https://github.com/apollographql/router/pull/6207))

The router's calculated overhead processing time has been fixed, where the time spent waiting for the subgraph response of a deduplicated request had been incorrectly included.

By [@Geal](https://github.com/Geal) in https://github.com/apollographql/router/pull/6207

### Fix demand control panic for custom scalars that represent non-GraphQL-compliant JSON ([PR #6288](https://github.com/apollographql/router/pull/6288))

Previously, a panic could be triggered in the router's demand control plugin with the following schema:

```
scalar ArbitraryJson

type MyInput {
    json: ArbitraryJson
}

type Query {
    fetch(args: MyInput): Int
}
```

Then, submitting the query

```
query FetchData(: ArbitraryJson) {
    fetch(args: {
        json: 
    })
}
```

and variables

```
{
    "myJsonValue": {
        "field.with.dots": 1
    }
}
```

During scoring, the demand control plugin would attempt to convert the variable structure into a GraphQL-compliant structure requiring valid GraphQL names as keys. The dot characters in the keys however would cause a panic. 

With this fix, only the GraphQL compliant part of the input object is scored, and the arbitrary JSON marked by the custom scalar is scored as an opaque scalar (similar to how built-ins like `Int` or `String` are processed).

By [@tninesling](https://github.com/tninesling) in https://github.com/apollographql/router/pull/6288

### Fix incorrect overriding of concrete type names with interface names when merging responses ([PR #6250](https://github.com/apollographql/router/pull/6250))

When using `@interfaceObject`, differing pieces of data can come back with either concrete types or interface types depending on the source. Previously, receiving the data in a particular order could incorrectly result in the interface name of a type overwriting its concrete name.

To make the response merging order-agnostic, the router now checks the schema to ensure concrete types are not overwritten with interfaces or less specific types.

By [@tninesling](https://github.com/tninesling) in https://github.com/apollographql/router/pull/6250

### Query planner cache key improvements ([Issue #5160](https://github.com/apollographql/router/issues/5160))

> [!IMPORTANT]
> If you have enabled [Distributed query plan caching](https://www.apollographql.com/docs/router/configuration/distributed-caching/#distributed-query-plan-caching), this release changes the hashing algorithm used for the cache keys.  On account of this, you should anticipate additional cache regeneration cost when updating between these versions while the new hashing algorithm comes into service.

Several performance improvements have been implemented for query plan cache key generation. In particular, the distributed cache's key format has changed, which adds prefixes to the different key segments to help in debugging.

By [@Geal](https://github.com/Geal) in https://github.com/apollographql/router/pull/6206

### Add entity caching invalidation configuration metrics ([PR #6286](https://github.com/apollographql/router/pull/6286))

We've added metrics for our analytics to know if entity caching invalidation is enabled.

By [@bnjjj](https://github.com/bnjjj) in https://github.com/apollographql/router/pull/6286

### Avoid creating stub span for supergraph events if current span exists ([PR #6096](https://github.com/apollographql/router/pull/6096))

The router optimized its telemetry implementation by not creating a redundant span when it already has a span available to use the span's extensions for supergraph events.

By [@bnjjj](https://github.com/bnjjj) in https://github.com/apollographql/router/pull/6096

### Clarify docs for authorization directive composition ([PR #6216](https://github.com/apollographql/router/pull/6216))

The docs for [authorization directive composition](https://www.apollographql.com/docs/graphos/routing/security/authorization#composition-and-federation) have been clarified, including corrected code examples.

By [@Meschreiber](https://github.com/Meschreiber) in https://github.com/apollographql/router/pull/6216


CHANGELOG.md
```diff
@@ -4,6 +4,293 @@ All notable changes to Router will be documented in this file.
...
+* `ipv4_only` - Only query for `A` (IPv4) records.
+* `ipv6_only` - Only query for `AAAA` (IPv6) records.
+* `ipv4_and_ipv6` - Query for both `A` (IPv4) and `AAAA` (IPv6) records in parallel.
+* `ipv6_then_ipv4` - Query for `AAAA` (IPv6) records first; if that fails, query for `A` (IPv4) records.
+* `ipv4_then_ipv6`(default) - Query for `A` (IPv4) records first; if that fails, query for `AAAA` (IPv6) records.
+
+You can change the DNS resolution strategy applied to a subgraph's URL:
+
+```yaml title="router.yaml"
+traffic_shaping:
```
BrynCooke, line 24: @garypen For router 2.0 we should rename this to `http`

CHANGELOG.md
```diff
@@ -4,6 +4,293 @@ All notable changes to Router will be documented in this file.
...
+
+By [@IvanGoncharov](https://github.com/IvanGoncharov) in https://github.com/apollographql/router/pull/6194
+
+### Compress subgraph operations by generating fragments ([PR #6013](https://github.com/apollographql/router/pull/6013))
+
+The router now compresses operations sent to subgraphs by default by generating fragment
+definitions and using them in the operation.
+
+This change enables `generate_query_fragments` by default while disabling `experimental_reuse_query_fragments`. When enabled, `experimental_reuse_query_fragments` attempts to intelligently reuse the fragment definitions
+from the original operation. However, fragment generation with `generate_query_fragments` is much faster also produces better outputs in most cases.
```
goto-bus-stop, line 68:
```suggestion
from the original operation. However, fragment generation with `generate_query_fragments` is much faster and produces better outputs in most cases.
```

goto-bus-stop: or "and also"

CHANGELOG.md
```diff
@@ -4,6 +4,293 @@ All notable changes to Router will be documented in this file.
...
+To make the response merging order-agnostic, the router now checks the schema to ensure concrete types are not overwritten with interfaces or less specific types.
+
+By [@tninesling](https://github.com/tninesling) in https://github.com/apollographql/router/pull/6250
+
+## 🛠 Maintenance
+
+### Query planner cache key improvements ([Issue #5160](https://github.com/apollographql/router/issues/5160))
+
+> [!IMPORTANT]
+> If you have enabled [Distributed query plan caching](https://www.apollographql.com/docs/router/configuration/distributed-caching/#distributed-query-plan-caching), this release changes the hashing algorithm used for the cache keys.  On account of this, you should anticipate additional cache regeneration cost when updating between these versions while the new hashing algorithm comes into service.
```
goto-bus-stop, line 265: the same is true for fragment generation. should this only be called out down here or at the top of the changelog, as it affects the whole release?

CHANGELOG.md
```diff
@@ -4,6 +4,293 @@ All notable changes to Router will be documented in this file.
...
+
+For a subgraph batching operation, the router now doesn't log the entire subgraph response when failing to notify a waiting batch participant. This saves the router from logging the large amount of data (PII and/or non-PII data) that a subgraph response may contain.
+
+By [@garypen](https://github.com/garypen) in https://github.com/apollographql/router/pull/6150
+
+### Move heavy computation to a thread pool with a priority queue ([PR #6247](https://github.com/apollographql/router/pull/6247))
+
+The router now avoids blocking threads when executing asynchronous code by using a thread pool with a priority queue.
+
+This improves the performance of the following components can take non-trivial amounts of CPU time:
```
goto-bus-stop, line 172:
```suggestion
This improves the performance of the following components that can take non-trivial amounts of CPU time:
```

CHANGELOG.md
abernix: ebf375a35

BrynCooke (changes requested):
Blocking while perf is checked

BrynCooke (approved):

CHANGELOG.md
```diff
@@ -4,6 +4,293 @@ All notable changes to Router will be documented in this file.
...
+  ]
+}
+```
+
+By [@IvanGoncharov](https://github.com/IvanGoncharov) in https://github.com/apollographql/router/pull/6191
+
+### Add `@context` support in the native query planner ([PR #6310](https://github.com/apollographql/router/pull/6310))
+
+The [`@context`](https://www.apollographql.com/docs/graphos/reference/federation/directives#context) feature is now available in the native query planner.
+This brings the native query planner to feature parity with the legacy query planner for all Federation v2 graphs.
```
lrlna, line 128:
````suggestion
The [`@context`](https://www.apollographql.com/docs/graphos/reference/federation/directives#context) feature is now available in the [native query planner](https://www.apollographql.com/docs/graphos/routing/query-planning/native-query-planner).
This brings the native query planner to feature parity with the legacy query planner for all Federation v2 graphs. The native query planner can be enabled with the following configuration:
```yaml, filename=router.yaml
experimental_query_planner_mode: new
```

````

File: CHANGELOG.md [hunk 6]
```diff
@@ -4,6 +4,293 @@ All notable changes to Router will be documented in this file.
...
+
+This improves the performance of the following components that can take non-trivial amounts of CPU time:
+
+* GraphQL parsing
+* GraphQL validation
+* Query planning
+* Schema introspection
+
+In order to avoid blocking threads that execute asynchronous code,
+they are now run in a new thread pool with a priority queue. The size of the thread pool is based on the number of available CPU cores.
```
lrlna, line 183: This is repeating a sentence from above
```suggestion
The size of the thread pool is based on the number of available CPU cores.
```

PR #6337, opened 2024-11-26
chore(federation): Improved serialization impls

TylerBloom:
This PR changes how several types are serialized in `apollo-federation`. Namely, there are several types from `apollo-compiler` that do not implement `serde::Serialize`. Before, these types were serialized as strings or simply skipped. This has been changes to use the provided `serialize` method on those `apollo-compiler` types. Additionally, this changes the inner type of `SelectionId` and `OverrideId` from a `usize` to a `Uuid`. This adds `uuid` as a dependency to `apollo-federation`, but `uuid` was already a (transitive) dependency of `apollo-router`.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-841]: https://apollographql.atlassian.net/browse/ROUTER-841?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apollo-federation/src/operation/mod.rs
+++ b/apollo-federation/src/operation/mod.rs
@@ -83,7 +83,9 @@ impl SelectionId {
 #[derive(Clone, PartialEq, Eq, Default, serde::Serialize)]
 pub(crate) struct ArgumentList {
     /// The inner list *must* be sorted with `sort_arguments`.
-    #[serde(serialize_with = "crate::utils::serde_bridge::serialize_optional_slice_of_exe_argument_nodes")]
+    #[serde(
+        serialize_with = "crate::utils::serde_bridge::serialize_optional_slice_of_exe_argument_nodes"
+    )]
     inner: Option<Arc<[Node<executable::Argument>]>>,
 }
 
--- a/apollo-federation/src/operation/mod.rs
+++ b/apollo-federation/src/operation/mod.rs
@@ -17,6 +17,7 @@ use std::fmt::Display;
 use std::fmt::Formatter;
 use std::hash::Hash;
 use std::ops::Deref;
+use std::sync::atomic;
 use std::sync::Arc;
 
 use apollo_compiler::collections::IndexMap;
@@ -28,7 +29,6 @@ use apollo_compiler::validation::Valid;
 use apollo_compiler::Name;
 use apollo_compiler::Node;
 use itertools::Itertools;
-use uuid::Uuid;
 
 use crate::compat::coerce_executable_values;
 use crate::error::FederationError;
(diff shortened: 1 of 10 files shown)


apollo-federation/src/operation/mod.rs
```diff
@@ -66,23 +65,13 @@ pub(crate) use rebase::*;
...
-///
-/// Note that we shouldn't add `derive(Serialize, Deserialize)` to this without changing the types
-/// to be something like UUIDs.
-#[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
-// NOTE(@TylerBloom): This feature gate can be removed once the condition in the comment above is
-// met. Note that there are `serde(skip)` statements that should be removed once this is removed.
-#[cfg_attr(feature = "snapshot_tracing", derive(Serialize))]
-pub(crate) struct SelectionId(usize);
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Hash, serde::Serialize)]
+pub(crate) struct SelectionId(Uuid);
```
dariuszkuc, line 70: should this change be in a separate PR?

goto-bus-stop:
I'd prefer not to introduce UUID if we can avoid it, I like the rest though 😁

apollo-federation/src/operation/mod.rs
TylerBloom: This can be moved into a different PR or reverted. My motivation was to make our serialization uniform (i.e. not feature gating anything), and both using `apollo-compiler`'s `serialize` methods and removing feature-gating fall into this. While we aren't serializing to cache, simply implementing `Serialize` would violate what that comment is saying but changing the representation and not cache is still inline.

My goal with making things uniform (and specifically removing the feature-gating) was to minimize what we would need to consider when/if we started to serialize for caching. That said, this is ultimately a small change that can be revisited at that point in time if this change isn't inline with what would be needed.

goto-bus-stop: I think the comment is a bit over-eager and it should maybe just generally warn about it being a fleeting value that is not designed to be unique between planning runs

goto-bus-stop: Oh, the uniform attributes point makes sense. I think it's okay to not feature gate serialization and warn about deserialization in the comment?

TylerBloom: >  I think it's okay to not feature gate serialization and warn about deserialization in the comment?

I think this works. We already do not implement `Deserialize` and there are several outstanding questions around what that would look like (there is even a new one in the PR). I'll revert the `uuid` changes, update that comment, and leave a new comment (either at the top of `lib.rs` or by `QueryPlan`) saying that we don't support deserialization yet and we need to answer a few questions before we can.

dariuszkuc (approved):

PR #6338, opened 2024-11-26
Remove experimental_retry

bnjjj:
Remove `experimental_retry`. Looking at our analytics it looks like it's not used and it's not working properly. As it was an experimental feature in order to check if we needed this feature or not we decided to remove it.

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/plugins/traffic_shaping/retry.rs
```diff
@@ -30,21 +29,57 @@ impl RetryPolicy {
...
+    ) -> Option<Self::Future> {
+        let subgraph_name = req.subgraph_name.clone().unwrap_or_default();
         match result {
-            Ok(_) => {
-                // Treat all `Response`s as success,
-                // so deposit budget and don't retry...
-                self.budget.deposit();
-                None
+            Ok(resp) => {
+                if resp.response.status() >= http::StatusCode::BAD_REQUEST {
```
garypen, line 47: Is this correct? I mean, what if the subgraph does something "weird" like returning a redirect?

i.e.: should this be:
```
if resp.response.status() != http::StatusCode::OK {
```
?

apollo-router/src/plugins/traffic_shaping/retry.rs
```diff
@@ -53,20 +88,27 @@ impl<Res, E> Policy<subgraph::Request, Res, E> for RetryPolicy {
 
                 let withdrew = self.budget.withdraw();
                 if withdrew.is_err() {
-                    tracing::info!(
-                        monotonic_counter.apollo_router_http_request_retry_total = 1u64,
+                    u64_counter!(
+                        "apollo_router_http_request_retry_total",
+                        "Number of retry for an http request to a subgraph",
```
garypen, line 93:
```suggestion
                        "Number of retries for an http request to a subgraph",
```

garypen: Same for other "retry".

apollo-router/src/plugins/traffic_shaping/retry.rs
```diff
@@ -76,3 +118,129 @@ impl<Res, E> Policy<subgraph::Request, Res, E> for RetryPolicy {
...
+            assert!(retry
+                .retry(
+                    &subgraph_req,
+                    Ok::<&subgraph::Response, &BoxError>(
+                        &subgraph::Response::fake_builder()
+                            .status_code(StatusCode::BAD_REQUEST)
+                            .build()
+                    )
+                )
+                .is_some());
```
garypen, line 235: Add a test for a redirect: [MOVED_PERMANENTLY](https://docs.rs/http/latest/http/status/struct.StatusCode.html#associatedconstant.MOVED_PERMANENTLY)

garypen (changes requested):
Some things to consider changing. Definitely change "retry" to "retries" though.

apollo-router/src/plugins/traffic_shaping/retry.rs
bnjjj: I recently saw users using 302 status code as a correct status code and they were making some stuffs in rhai about this. So I prefer to be conservative and use 400 status code as a threshold

garypen: It seems weird to me. If we go with this, then it means that we are encoding "weird" behaviour into the router and that might not be what some customers want.

My preference is to do the expected thing, rather than produce something which fits into a "weird" behaviour of a customer. See what other people think, since I may really be in the minority here.

bnjjj: >It seems weird to me. If we go with this, then it means that we are encoding "weird" behaviour into the router and that might not be what some customers want.

I agree, but I also agree with what @goto-bus-stop why would we retry on 3XX ?

garypen: To be honest: I find the concept of the retry module a bit weird. Maybe the best thing to do is to define **exactly** which codes we will retry on and document that?

@goto-bus-stop Says definitely on a 429. I'd probably also want to retry a 503. Apart for those two, I probably wouldn't want to retry.

goto-bus-stop: I like starting with a small selection of codes.

I'm second-guessing 429 because we don't support `Retry-After` in this implementation and maybe that's something that clients should be doing, rather than us? If we just retry instantly then there is a good chance that we get a 429 again.

BrynCooke: https://www.rfc-editor.org/rfc/rfc9110.html#name-retry-after mentions 3xx is legit for retry.

goto-bus-stop: I don't think it says that. It says that the user agent should wait `Retry-After` seconds before following the redirect, not that it should retry the original request:
> When sent with any [3xx (Redirection)](https://www.rfc-editor.org/rfc/rfc9110.html#status.3xx) response, Retry-After indicates the minimum time that the user agent is asked to wait before issuing the redirected request.

BrynCooke: Would we want to do that? Retry to the redirected request?

goto-bus-stop (commented):
LGTM with a changeset and updated PR title 😇

goto-bus-stop (approved):
LGTM, does it need further discussion w/ anyone in the team or did we already have agreement?

theJC:
I believe retries should be removed from being mentioned in the shaping ordering the doc here as well: https://github.com/apollographql/router/pull/6338/files#diff-f3c78d90b287606d615d6a2932685740f4302889e7e846a8e31940e2a9ca861aR176

PR #6344, opened 2024-11-27
feat: redact subgraph errors

Samjin:
This breaking change is to update `include_subgraph_errors` to `redact_subgraph_errors` with additional blacklist/whitelist capabilities. 
Note that test cases are not fully updated yet. I'd like to get input from Apollo team regarding the config proposal. 

Reasons we need this: 
1. Some subgraphs may want to pass non-vulnerable data to client via error extensions.
2. Some clients depend on certain data in error extensions to render UI. Although they shouldn't do it, migrating away from it would take time, especially from natives. While waiting their migration, it allows the original error message and selected keys to be redacted or propagated per subgraph.

Configs
1. This is a breaking change as long as user defines `include_subgraph_errors` in router.yaml. However, it shouldn't affect default behavior as the new change will redact subgraph error message and error extensions by default. 

Before
```yaml
include_subgraph_errors:
  all: false
```
After
```yaml
redact_subgraph_errors:
  all: true # Will redact error's message and extensions
```

2. Additionally, redact specific error extension keys for all subgraphs
```yaml
redact_subgraph_errors:
    all: true # Will redact message.
    redact_extensions_keys:
      - "service"
      - "exception"
```

3. Redact specific subgraphs only
```yaml
redact_subgraph_errors: 
    all: false # Will not redact error message and extensions.
    redact_extensions_keys:
      - "service"
      - "exception"
    subgraphs:
      product:
        enabled: true # will redact message and keys based on default config. 
      reviews:
        - "service" # Will not redact these keys. Other keys from default config and error message will be redacted.
```

4. Redact for all subgraphs, but with subgraph override.
```yaml
redact_subgraph_errors:
    all: true # Redact all subgraph's error messages. 
    redact_extensions_keys:
      - "service"
      - "exception"
    subgraphs:
      product:
        enabled: false # Will not redact anything.
      reviews:
        - "service" # Will not redact these keys. Other keys from default config and error message will be redacted.
```

There is no issue number yet. 

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


bnjjj:
Hey @Samjin I didn't take a look at the code yet but I already have some questions reading the description. It looks like you're renaming an existing feature. Is it backward compatible with the existing one ? We can't introduce breaking changes so I just would like to know if you introduced a migration or if you thoughts about backward compatibility and non breaking changes. It would also be great to have a comparison in PR's description to know how it was before and how it would be with your changes (config wise)

Samjin:
> Hey @Samjin I didn't take a look at the code yet but I already have some questions reading the description. It looks like you're renaming an existing feature. Is it backward compatible with the existing one ? We can't introduce breaking changes so I just would like to know if you introduced a migration or if you thoughts about backward compatibility and non breaking changes. It would also be great to have a comparison in PR's description to know how it was before and how it would be with your changes (config wise)

I updated the description. In short, it's a breaking change. The intention to rename the plugin is to make subgraph's override clear. 

However we could also update to something like this if it has to be backward compatible. 
```yaml 
include_subgraph_errors: 
  all: false
  redact_extensions_keys:
    - "service"
    - "exception"
  subgraphs: 
    product:
       include: true # redact error
    reviews:
       - "service"
```

I do find `redact_subgraph_errors` easier to use when setting overrides for subgraphs as I often mistakenly set `false` in subgraph to turn off redaction, but it ends up opposite. But this is subjective, not sure who else shares the same experience 😁 .

Samjin:
Sorry, I just realized this **maybe** not a breaking change since it won't allow user to build if they don't update their explicit config. Also, the default behavior is the same as `include_subgraph_errors` if user never defined it. 

If this is correct, I can update the description later with more migration steps.

bnjjj:
@Samjin Not being able to run the router after upgrading the version without changing the configuration is considered as a breaking change for us

Samjin:
> @Samjin Not being able to run the router after upgrading the version without changing the configuration is considered as a breaking change for us

Ok. How about [this approach mentioned above](https://github.com/apollographql/router/pull/6344#issuecomment-2504411378) that doesn't break current config?

Samjin:
Closing this in favor of https://github.com/apollographql/router/pull/6402

PR #6350, opened 2024-11-27
Migrate monotonic counter metrics to `u64_counter!`

goto-bus-stop:
This is part 1 out of... 4 or 5? of a series of work to move over to our new telemetry macros.

This part does the easiest part :), the `tracing::info!(monotonic_counter.)` macros that should now use `u64_counter!()`. I used a lot of commits so I could take notes while doing the work, I'll copy them to PR comments so there's no need to look at each commit individually.

I avoided breaking changes to the metrics for now, so this is targeted at 1.x.

Two uses of `tracing::info!(monotonic_counter.)` remain, these are already being addressed in #6338.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/notification.rs
```diff
@@ -510,7 +510,11 @@ where
 
         match Pin::new(&mut this.msg_receiver).poll_next(cx) {
             Poll::Ready(Some(Err(BroadcastStreamRecvError::Lagged(_)))) => {
-                tracing::info!(monotonic_counter.apollo_router_skipped_event_count = 1u64,);
+                u64_counter!(
+                    "apollo_router_skipped_event_count",
+                    "Amount of times the router missed subscription events",
```
goto-bus-stop, line 515: I'm quite unsure about this description

apollo-router/src/uplink/mod.rs
```diff
@@ -210,7 +210,7 @@ where
     Response: Send + 'static + Debug,
     TransformedResponse: Send + 'static + Debug,
 {
-    let query = query_name::<Query>();
+    let query_name = query_name::<Query>();
```
goto-bus-stop, line 213: Just to clarify that it isn't the full query text.

apollo-router/src/services/subgraph_service.rs
```diff
@@ -507,19 +511,23 @@ async fn call_websocket(
...
     if !created {
         subscription_stream_tx
             .send(Box::pin(handle.into_stream()))
             .await?;
-        tracing::info!(
-            monotonic_counter.apollo_router_deduplicated_subscriptions_total = 1u64,
-            mode = %"passthrough",
+        u64_counter!(
+            "apollo_router_deduplicated_subscriptions_total",
+            "Total deduplicated subscription requests (deprecated)",
```
goto-bus-stop, line 528: I marked a few apparently duplicate, or poorly named metrics as deprecated. This one appears to be a less informative duplicate of `apollo.router.operations.subscriptions`, which has a `subscriptions.deduplicated` attribute.

apollo-router/src/plugins/authorization/mod.rs
```diff
@@ -556,8 +556,10 @@ impl Plugin for AuthorizationPlugin {
                         Ok(ControlFlow::Continue(request))
                     } else {
                         // This is a metric and will not appear in the logs
-                        tracing::info!(
-                            monotonic_counter.apollo_require_authentication_failure_count = 1u64,
+                        u64_counter!(
+                            "apollo_require_authentication_failure_count",
+                            "Number of unauthenticated requests (deprecated)",
```
goto-bus-stop, line 561: I marked this as deprecated because the naming doesn't seem to follow convention, but I don't know if we have a proper alternative for this? We have `apollo.router.operations.authorization` below but I'm not sure it reports the same thing.

bnjjj: I think that's the right thing to do. cc @BrynCooke you probably have more context

apollo-router/src/services/router/service.rs
```diff
@@ -378,8 +378,10 @@ impl RouterService {
 
                     Ok(RouterResponse { response, context })
                 } else {
-                    tracing::info!(
-                        monotonic_counter.apollo.router.graphql_error = 1u64,
+                    u64_counter!(
+                        "apollo.router.graphql_error",
+                        "Number of GraphQL error responses returned by the router",
```
bnjjj, line 383:
```suggestion
                        "Number of GraphQL error responses returned by the router (DEPRECATED)",
```

apollo-router/src/services/router/service.rs
```diff
@@ -799,12 +801,18 @@ impl RouterService {
...
+                        count
+                    );
                 }
                 Some(code) => {
-                    tracing::info!(
-                        monotonic_counter.apollo.router.graphql_error = count,
-                        code = code
+                    u64_counter!(
+                        "apollo.router.graphql_error",
+                        "Number of GraphQL error responses returned by the router",
```
bnjjj, line 813:
```suggestion
                        "Number of GraphQL error responses returned by the router (DEPRECATED)",
```

apollo-router/src/services/subgraph_service.rs
bnjjj: Oh! Good catch. It makes sense.

apollo-router/src/notification.rs
goto-bus-stop: Right. I was trying to make it meaningful to users. I could only find this being used for subscriptions, but the way I wrote it is probably not what actually happens. How would you explain it to someone who is running the router and doesn't know much about the implementation details?

apollo-router/src/services/router/service.rs
```diff
@@ -799,12 +801,18 @@ impl RouterService {
         for (code, count) in map {
             match code {
                 None => {
-                    tracing::info!(monotonic_counter.apollo.router.graphql_error = count,);
+                    u64_counter!(
+                        "apollo.router.graphql_error",
+                        "Number of GraphQL error responses returned by the router",
```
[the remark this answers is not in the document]
goto-bus-stop: Ah really? I thought we quite recently added more of these (though the metric name is definitely not good). Is there a replacement?

apollo-router/src/notification.rs
goto-bus-stop:
```suggestion
                    "Amount of events dropped from the internal message queue",
```
We could just say it's "internal" if it's not something easily explained to a user

apollo-router/src/plugins/authorization/mod.rs
```diff
@@ -588,11 +590,13 @@ impl Plugin for AuthorizationPlugin {
                 let needs_requires_scopes = request.context.contains_key(REQUIRED_SCOPES_KEY);
 
                 if needs_authenticated || needs_requires_scopes {
-                    tracing::info!(
-                        monotonic_counter.apollo.router.operations.authorization = 1u64,
+                    u64_counter!(
+                        "apollo.router.operations.authorization",
+                        "",
```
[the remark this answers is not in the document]
goto-bus-stop: I wasn't sure what to use for the description here :/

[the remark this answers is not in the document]
goto-bus-stop: I came up with something (a little vague, but doesn't seem like a lie at least)

apollo-router/src/notification.rs
bnjjj: Your last description LGTM

apollo-router/src/plugins/traffic_shaping/timeout/future.rs
```diff
@@ -49,7 +49,11 @@ where
         match Pin::new(&mut this.sleep).poll(cx) {
             Poll::Pending => Poll::Pending,
             Poll::Ready(_) => {
-                tracing::info!(monotonic_counter.apollo_router_timeout = 1u64,);
+                u64_counter!(
+                    "apollo_router_timeout",
+                    "Number of timed out client requests",
```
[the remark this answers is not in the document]
bnjjj:
```yaml
telemetry:
  instrumentation:
    instruments:
      router:
        http.server.request.duration:
          # Adding subgraph name, response status code from the router and the operation name
          attributes:
            http.response.status_code: true # If status code == 504 then it's a timeout at the router http request level
            graphql.operation.name:
              operation_name: string
      subgraph:
        # Adding subgraph name, response status code from the subgraph and original operation name from the supergraph
        http.client.request.duration:
          attributes:
            subgraph.name: true
            http.response.status_code: # If status code == 504 then it's a timeout at the subgraph http request level
              subgraph_response_status: code
```

apollo-router/src/services/router/service.rs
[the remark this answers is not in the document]
bnjjj:
```yaml
telemetry:
  instrumentation:
    instruments:
      router:
        http.server.request.duration:
          attributes:
            http.response.status_code: true
            graphql.operation.name:
              operation_name: string
            # This attribute will be set to true if the response contains graphql errors
            graphql.errors: # This will be true if it contains graphql errors
              on_graphql_error: true
```

Here is the alternative

apollo-router/src/plugins/traffic_shaping/timeout/future.rs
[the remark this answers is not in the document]
goto-bus-stop: Custom instrumentation is an enterprise only feature, but we can also get the status code without using enterprise features:
```yaml
telemetry:
  instrumentation:
    instruments:
      default_requirement_level: required
```

docs/source/reference/router/telemetry/instrumentation/standard-instruments.mdx
```diff
@@ -125,4 +122,9 @@ The initial call to Uplink during router startup is not reflected in metrics.
...
+- `apollo_router_span` - **Deprecated**: use `apollo_router_processing_time` instead.
+- `apollo_router_deduplicated_subscriptions_total` - **Deprecated**: use the `apollo.router.operations.subscriptions` metric's `subscriptions.deduplicated` attribute.
+- `apollo_authentication_failure_count` - **Deprecated**: use the `apollo.router.operations.authentication.jwt` metric's `authentication.jwt.failed` attribute.
+- `apollo_authentication_success_count` - **Deprecated**: use the `apollo.router.operations.authentication.jwt` metric instead. If the `authentication.jwt.failed` attribute is *absent* or `false`, the authentication succeeded.
+- `apollo_require_authentication_failure_count` - **Deprecated**: TODO @goto-bus-stop: no replacement?
+- `apollo_router_timeout` - **Deprecated**: this metric conflates timed-out requests from client to the router, and requests from the router to subgraphs. Timed-out requests have HTTP status code 504. Use the `http.response.status_code` attribute on the `http.server.request.duration` metric to identify timed-out router requests, and on the `http.client.request.duration` metric to identify timed-out subgraph requests.
```
goto-bus-stop, line 130: Please verify this. @BrynCooke shared a config for `default_requirement_level: required`, but that is the default, so I think this should work?

goto-bus-stop:
Un-deprecated `apollo.router.graphql_error` as the alternative is an enterprise feature currently.

The rest should be in order. @BrynCooke @bnjjj PTAL at `apollo_require_authentication_failure_count` and `apollo_router_timeout`.

apollo-router/src/plugins/authorization/mod.rs
BrynCooke: For now let's not mark these as deprecated. We will then circle back and fix the names.

BrynCooke: We're going to have to add a new metric with the desired name.... 🤔

BrynCooke: I think this can just go. The existing `http.server.request.duration` metric will have a status code on it already, and people can just look for `StatusCode::UNAUTHORIZED`.

docs/source/reference/router/telemetry/instrumentation/standard-instruments.mdx
```diff
@@ -125,4 +122,9 @@ The initial call to Uplink during router startup is not reflected in metrics.
 
 The following metrics have been deprecated and should not be used.
 
-- `apollo_router_span` - **Deprecated**—use `apollo_router_processing_time` instead.
+- `apollo_router_span` - **Deprecated**: use `apollo_router_processing_time` instead.
+- `apollo_router_deduplicated_subscriptions_total` - **Deprecated**: use the `apollo.router.operations.subscriptions` metric's `subscriptions.deduplicated` attribute.
+- `apollo_authentication_failure_count` - **Deprecated**: use the `apollo.router.operations.authentication.jwt` metric's `authentication.jwt.failed` attribute.
+- `apollo_authentication_success_count` - **Deprecated**: use the `apollo.router.operations.authentication.jwt` metric instead. If the `authentication.jwt.failed` attribute is *absent* or `false`, the authentication succeeded.
+- `apollo_require_authentication_failure_count` - **Deprecated**: TODO @goto-bus-stop: no replacement?
```
BrynCooke, line 129: `http.server.request.duration` has an attribute [http.response.status_code](https://opentelemetry.io/docs/specs/semconv/attributes-registry/http/) that users can use.

goto-bus-stop:
@BrynCooke I see some comments in my email that I can't find on Github. I think the metrics I marked as deprecated here all have a free-to-use migration path in 1.x and have appropriate steps in the docs?

bnjjj (approved):

PR #6354, opened 2024-11-27
Drop experimental reuse fragments optimization from RS QP

dariuszkuc:
Drop support for the experimental reuse fragment query optimization from Rust QP. This implementation was not only very slow but also very buggy due to its complexity. `experimental_reuse_query_fragments` option can be applied for the JS QP only.

Auto generation of fragments is a much simpler (and faster) algorithm that in most cases produces better results. Fragment auto generation is the default optimization since v1.58 release.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


goto-bus-stop:
We should also make sure that people aren't opting back into the old algorithm en masse in 1.58.0. But this is a very exciting diff 😁

lrlna (changes requested):
This PR also needs to update [docs](https://github.com/apollographql/router/blob/dev/docs/source/reference/router/configuration.mdx#fragment-generation-and-reuse) for fragment generation and reuse.

Otherwise really exciting to get rid of all the dead code!!

lrlna (changes requested):
I am so sorry, I read a tracing::warn message you wrote and realised this is going to need a [config migration](https://github.com/apollographql/router/blob/dev/apollo-router/src/configuration/migrations/README.md). For the time being it can be a `log`, and for the next release we can do a `delete`.

apollo-router/src/configuration/snapshots/apollo_router__configuration__tests__schema_generation.snap
```diff
@@ -6644,7 +6644,7 @@ expression: "&schema"
         },
         "experimental_reuse_query_fragments": {
           "default": null,
-          "description": "Enable reuse of query fragments Default: depends on the federation version",
+          "description": "Enable reuse of query fragments\n\nThis feature is deprecated and will be removed in next release. It is only available in JS QP. Use generate_query_fragments instead.",
```
dariuszkuc, line 6647:
```suggestion
          "description": "Enable reuse of query fragments\n\nThis feature is deprecated and will be removed in next release.\nThe config can only be set when the legacy query planner is explicitly enabled.",
```

lrlna (approved):
Thank you!

PR #6359, opened 2024-11-29
Truncating Invalid Error Paths

### LINKED ISSUE #5397: Propagate errors from subgraph without hiding
Apologies if this is not a bug. Our team has invested the better part of a week trying to get this working, and we have been unsuccessful.

**Describe the bug**
We are unable to get the Router to propagate error message from our subgraphs.

In all the following scenarios, we have router config
```
include_subgraph_errors:
  all: true
```

We have a service (Service P) which is a subgraph in our supergraph. If we send a query with a bad input directly to the Service P, it returns a helpful response error:

```
{
    "errors": [
        {
            "message": "UUID cannot parse the given literal of type `StringValueNode`.",
            "path": [
                "input",
                "acceptorId"
            ],
            "extensions": {
                "field": "Foo.acceptorId",
                "fieldType": "UUID"
            }
        }
    ]
}
```
However if we send the same query to the subgraph via the Router, we get less helpful responses, depending on what `Content-Type` and `Status` we return from Service P.

- If Service P returns status `200` and content type `application/graphql-response+json; charset=utf-8` or a content-type of `application/json`, then the response we get from the router is:
```
{
    "data": null
}
```
^ The secondary concern here, is that it doesn't indicate that an error happened at all. 

- If Service P returns status `400` and content type `application/graphql-response+json; charset=utf-8`, then the response we get from the router is:

```
{
    "data": null,
    "errors": [
        {
            "message": "HTTP fetch failed from 'paymentprocessing': 400: Bad Request",
            "extensions": {
                "service": "paymentprocessing",
                "reason": "400: Bad Request",
                "http": {
                    "status": 400
                }
            }
        }
    ]
}

```

In this case, it at least indicates that an error happened, but doesn't have any information as to why, which is a bad experience for our customers who are beginning their integration.

**To Reproduce**
1. Any schema on a subgraph behind the router.
2. Return one or more errors from subgraph on request.

**Expected behavior**
I was expecting the errors returned by the subgraph to be included in the response from the router.

- We run Router in a docker container on a Arm64 Linux container.

**Additional context**
- tested on multiple version of Router, including the latest 1.48.1
- include_subgraph_errors is set for all subgraphs
- We tried different combinations of status codes and content types for the subgraph response.

**Workarounds?**
We are also interested in work arounds if anyone can provide direction for that. We already have a couple small Rhai scripts running in production, and are able to add another if required.

Thank you in advanced.

IvanGoncharov:
This fix addresses an issue where the router was silently dropping subgraph errors that included invalid paths.
 
According to the [GraphQL Specification](https://spec.graphql.org/draft/#sel-GAPHRPHCAACCpC8-T) an error path must point to a **response field**:
> If an error can be associated to a particular field in the GraphQL result, it must contain an entry with the key path that details the path of the response field which experienced the error.

If a subgraph error includes a path that can't be matched to a response field, the router now truncates the path to the nearest valid field path and includes the truncated portion in a `truncatedPath` extension. For example:
```json
{
    "data": { "topProducts": null },
    "errors": [{
        "message": "Some error on subgraph",
        "path": ["topProducts"],
        "extensions": {
           "service": "products",
           "truncatedPath": ["invalid", 42]
        }
    }]
}
```

Fixes #5397

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/services/execution/service.rs
```diff
@@ -344,6 +344,25 @@ impl ExecutionService {
...
+                                .entry("truncatedPath")
+                                .or_insert(truncated_path);
+                        }
+
+                        // if path was invalid that means we can't trust locations either
+                        error.locations.clear();
+                    }
+                }
+            }
+
```
IvanGoncharov, line 365: Previously, we had errors with invalid paths, and they were dropped later in the pipeline.
Because code that filtered errors for primary and deferred responses couldn't match errors with invalid paths to any of the responses so it silently dropped them.

IvanGoncharov: https://github.com/apollographql/router/blob/b42ead68f93eab6eb527304e4c17647dd5dac804/apollo-router/src/services/execution/service.rs#L373-L378
https://github.com/apollographql/router/blob/b42ead68f93eab6eb527304e4c17647dd5dac804/apollo-router/src/services/execution/service.rs#L451-L457

apollo-router/src/spec/selection.rs
```diff
@@ -189,9 +189,13 @@ impl Selection {
         })
     }
 
-    pub(crate) fn contains_error_path(&self, path: &[PathElement], fragments: &Fragments) -> bool {
+    pub(crate) fn matching_error_path_length(
```
bnjjj, line 192: If we don't find anything then we return `0` right ? Why not just returning an `Option` ?

IvanGoncharov: The main reason: If I put `usize` inside `Option` it would mean I need to handle both `None` and `Some(0)` as results. 

Also since this is recursive function it needs to add `+ 1` which would require more code to handle options.

.changesets/fix_fix_invalid_error_path.md
```diff
@@ -0,0 +1,23 @@
...
+If a subgraph error includes a path that can't be matched to a response field, the router now truncates the path to the nearest valid field path and includes the truncated portion in a `truncatedPath` extension. For example:
+```json
+{
+    "data": { "topProducts": null },
+    "errors": [{
+        "message": "Some error on subgraph",
+        "path": ["topProducts"],
+        "extensions": {
+           "service": "products",
+           "truncatedPath": ["invalid", 42]
```
bnjjj, line 17: I'm not sure to understand what does that mean

bnjjj: Is it the field name ?

apollo-router/src/spec/selection.rs
bnjjj: Got it, I didn't understand `0` has a meaning

apollo-router/src/spec/selection.rs
```diff
@@ -224,42 +234,56 @@ impl Selection {
...
+                        .max()
+                        .unwrap_or(0)
                 } else {
-                    false
+                    0
                 }
             }
             (Some(PathElement::Index(_)), _) | (Some(PathElement::Flatten(_)), _) => {
-                self.contains_error_path(&path[1..], fragments)
+                self.matching_error_path_length(&path[1..], fragments) + 1
```
bnjjj, line 277: `&path[1..]` are we sure this call cannot panic ? what if we have a path size of 1 ? Does it panic ? I can't remember

goto-bus-stop: if path has 1 element, then `&path[1..]` is the empty slice.

We could use `path.split_first()` at the top of this match statement instead of `path.first()` and `&path[1..]` to compile time guarantee that there can't be a panic, but it should be fine already

goto-bus-stop:
I think the implementation looks good in principle.

I'm slightly uneasy about exposing the `truncatedPath` in extensions. Even when we added `extensions.service` some customers were surprised and considered it leaky of internal details. I think `truncatedPath` is likely to run into the same issue.

I think it could be better to land this PR *without* the extension, i.e., silently truncating paths. This would be a definite improvement over the current situation while not introducing new fields to extensions that someone could consider (rightly or wrongly) as exposing "private" details. Do you think there's a requirement for users to have access to the truncated part of the path?

.changesets/fix_fix_invalid_error_path.md
IvanGoncharov: @bnjjj name inside `extensions`.
Original issues is caused some GraphQL libraries trying to put path inside query arguments into the path.

For example:
```json5
{
    "errors": [
        {
            "message": "UUID cannot parse the given literal of type `StringValueNode`.",
            "path": [
                "input", 
                "acceptorId"
            ],
            "extensions": {
                "field": "Foo.acceptorId",
                "fieldType": "UUID"
            }
        }
    ]
}
```
This is invalid according to the GraphQL spec, so we should remove `path` (to be spec compliant), but the old `path` value is still useful for debugging.
Plus, if we remove a path, we switch router behavior from "silently dropping errors" to "silently dropping path inside errors".
My idea is to put the truncated part inside extensions as `truncatedPath`, so above error would be change to this:
```
{
    "errors": [
        {
            "message": "UUID cannot parse the given literal of type `StringValueNode`.",
            "extensions": {
                "field": "Foo.acceptorId",
                "fieldType": "UUID"
                "path": [
                    "input", 
                    "acceptorId"
                ],
            }
        }
    ]
}
```

IvanGoncharov:
>  Even when we added extensions.service some customers were surprised and considered it leaky of internal details. I think truncatedPath is likely to run into the same issue.

@goto-bus-stop Yes, you are right.
I just saw #6402, and it explains why this is problematic.
I think `truncatedPath` is helpful, but I agree it should be added after #6402 or something similar is merged.
I will split it into separate draft PRs so as not to lose it.

.changesets/fix_fix_invalid_error_path.md
IvanGoncharov: I will split it up in a separate PR as discussed here: https://github.com/apollographql/router/pull/6359#issuecomment-2535841673

apollo-router/src/spec/selection.rs
bnjjj: @IvanGoncharov I saw you used `split_first` but you're not using the second part of this `split_first`here instead of `&path[1..]`

bnjjj (approved):

PR #6372, opened 2024-12-02
Update usages of tower::util::Either and tower::retry::Policy

goto-bus-stop:
Both turn out to be pretty trivial API changes.

`cargo check` before:
```
error: could not compile `apollo-router` (lib) due to 29 previous errors; 7 warnings emitted
```

`cargo check` after:
```
error: could not compile `apollo-router` (lib) due to 24 previous errors; 7 warnings emitted
```

---

**Checklist**

- [ ] Tests added and passing[^3]

**Exceptions**

We can't actually run the tests right now, so the best we have is a reduction in compile errors.

**Notes**

[ROUTER-907]: https://apollographql.atlassian.net/browse/ROUTER-907?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -289,6 +289,7 @@ dependencies = [
  "hex",
  "hickory-resolver",
  "hmac",
+ "http 0.2.12",
  "http 1.1.0",
  "http-body 1.0.1",
  "http-body-util",
--- a/apollo-router/Cargo.toml
+++ b/apollo-router/Cargo.toml
@@ -104,6 +104,7 @@ graphql_client = "0.14.0"
 hex.workspace = true
 # http.workspace = true
 http = { version = "1.1.0" }
+http-0_2 = { version = "0.2.12", package = "http" }
 # http-body = "0.4.6"
 http-body = "1.0.1"
 http-body-util = {version = "0.1.2" }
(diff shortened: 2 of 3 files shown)


apollo-router/src/plugins/traffic_shaping/retry.rs
```diff
@@ -67,12 +71,12 @@ impl<Res, E> Policy<subgraph::Request, Res, E> for RetryPolicy {
                     subgraph = %self.subgraph_name,
                 );
 
-                Some(future::ready(self.clone()))
+                Some(future::ready(()))
```
goto-bus-stop, line 74: instead of returning a new policy, you can now mutate the policy inside this function, so there is no need to return `self.clone()`.

apollo-router/src/plugins/traffic_shaping/mod.rs
```diff
@@ -273,8 +274,8 @@ impl Plugin for TrafficShaping {
     }
 }
 
-pub(crate) type TrafficShapingSubgraphFuture<S> = Either<
-    Either<
+pub(crate) type TrafficShapingSubgraphFuture<S> = EitherResponseFuture<
+    EitherResponseFuture<
         BoxFuture<'static, Result<subgraph::Response, BoxError>>,
         BoxFuture<'static, Result<subgraph::Response, BoxError>>,
```
goto-bus-stop, line 280: Hm, that's funny, these branches are the same so this probably doesn't need to be an `Either` anymore.

BrynCooke (approved):
Approved, but let's follow up with a removal of Either in favour of `option_layer` like so: https://github.com/apollographql/router-private/pull/349/files#diff-b052a31029b1a2349a626a98a47d3ce54f7a599a370d58ea74ac22d44c2e84fcL268

PR #6374, opened 2024-12-02
plugins/fleet_detector: add instance gauge

nmoutschen:
Adds an additional metric on the `fleet_detector` plugin to track the instance and its environment (OS, distribution, compute platform, etc.) in a non-intrusive way.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/plugins/fleet_detector.rs
+++ b/apollo-router/src/plugins/fleet_detector.rs
@@ -22,6 +22,7 @@ use crate::plugin::PluginInit;
 use crate::plugin::PluginPrivate;
 
 const REFRESH_INTERVAL: Duration = Duration::from_secs(60);
+const COMPUTE_DETECTOR_THRESHOLD: u16 = 24576;
 
 #[derive(Debug, Default, Deserialize, JsonSchema)]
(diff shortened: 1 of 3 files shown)


apollo-router/src/plugins/fleet_detector.rs
```diff
@@ -65,7 +66,37 @@ impl GaugeStore {
...
+            // Operating System
+            attributes.push(KeyValue::new("os.type", get_otel_os()));
+            if OS == "linux" {
+                attributes.push(KeyValue::new(
+                    "linux.distribution",
+                    System::distribution_id(),
+                ));
+            }
+            // Compute Environment
+            if let Some(env) = apollo_environment_detector::detect_one(24576) {
```
jonathanrainer, line 84: Is it worth capturing the magic number in a constant somewhere?

jonathanrainer (approved):
One small thing, otherwise looks good

PR #6376, opened 2024-12-02
Move distributed cache key version from router-bridge to apollo-federation

SimonSapin:
We plan to stop using the router-bridge crate entirely. The apollo-federation crate seems like a good place for this since most query plan changes would come from there.

This PR should *not* change current cache keys.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


SimonSapin:
After some discussion we agreed that it’s too easy to forget to increment this new constant. So closing in favor of a new approach: https://github.com/apollographql/router/pull/6406

PR #6378, opened 2024-12-02
merge main after 1.58

lennyburdette:
---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-federation/src/supergraph/mod.rs
```diff
@@ -1576,6 +1691,103 @@ lazy_static! {
     };
 }
 
+fn insert_directive(
+    schema: &mut FederationSchema,
+    pos: &CompositeTypeDefinitionPosition,
+    directive: Component<Directive>,
```
pubmodmatt, line 1697: Looks like a merge error here - the build is failing due to a missing `use` for `Directive`

lennyburdette: yep, thanks. i had fixed it just forgot to commit it!

pubmodmatt (approved):

garypen (changes requested):
Please don't merge this PR. I'm about to merge `dev` into `next`. See: https://github.com/apollographql/router/pull/6382

I need to understand why we should merge `main` to `next` rather than `dev` as we have been so far.

abernix:
This is my fault.  I suggested to @lennyburdette  that `main` might be mildly less volatile than `dev` to merge in prior to a release on the `next` branch (a 2.0.0-preview).   I didn't think it would cause more problems since they all share a branch strategy, but if we are regularly merging `dev` into `next` already then using `main` literally doesn't make sense anyhow.

Maybe sort this out on the agenda of today's group huddle and I'll keep out of the way next time. 😄

PR #6379, opened 2024-12-02
fix: use the original unescaped string for single-quoted strings

lennyburdette:
If the value is not a block string, it may include escaped characters like newlines. We need to use the "unescaped" value because our sers like JSONSelection will puke on doubly-escaped characters like `\\n`. This is to avoid issues if the SDL content is reformatted by a tool that collapses block strings into single-quoted strings.

Offsets will be wrong for non-block strings with escaped racters but manually writing \n will be uncommon (especially in JSONSelection where all whitespace is insignificant).

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


lennyburdette:
the perf test failure is from the connectors test hitting 4204 rps when the limit is 4200 🤣

lennyburdette:
closed in favor of https://github.com/apollographql/router/pull/6390

PR #6380, opened 2024-12-03
Miscellaneous @context/@fromContext bugfixes

sachindshinde:
In summary, this PR makes changes to the Rust QP port of `@context`/`@fromContext` for the sake of:
- Bugfixes.
- Renames/Refactors to simplify and reduce confusion.
- Adherence to existing conventions.
  - This isn't necessarily an endorsement of those conventions, but primarily for consistency sake.

There's a lot of these, so I'm going to leave comments on the code with relevant information (mainly bugfix notes, but with some occasional notes about refactors).

### DIFF
--- a/apollo-federation/src/lib.rs
+++ b/apollo-federation/src/lib.rs
@@ -42,6 +42,7 @@ pub(crate) mod utils;
 use apollo_compiler::ast::NamedType;
 use apollo_compiler::validation::Valid;
 use apollo_compiler::Schema;
+use itertools::Itertools;
 use link::join_spec_definition::JOIN_VERSIONS;
 use schema::FederationSchema;
 
@@ -102,15 +103,15 @@ pub(crate) fn validate_supergraph(
         }.into());
     };
     let context_spec_definition = metadata.for_identity(&Identity::context_identity()).map(|context_link| {
-        context_versions.find(&context_link.url.version).map_or_else(|| {
-            Err(SingleFederationError::InvalidFederationSupergraph {
+        context_versions.find(&context_link.url.version).ok_or_else(|| {
+            SingleFederationError::InvalidFederationSupergraph {
                 message: format!(
                     "Invalid supergraph: uses unsupported context spec version {} (supported versions: {})",
                     context_link.url.version,
-                    context_versions.versions().map(|v| v.to_string()).collect::<Vec<_>>().join(", "),
+                    context_versions.versions().join(", "),
                 ),
-            })
-        }, Ok)
+            }
+        })
     }).transpose()?;
     Ok((
         link_spec_definition,
--- a/apollo-federation/src/lib.rs
+++ b/apollo-federation/src/lib.rs
@@ -48,6 +48,8 @@ use schema::FederationSchema;
 pub use crate::api_schema::ApiSchemaOptions;
 use crate::error::FederationError;
 use crate::error::SingleFederationError;
+use crate::link::context_spec_definition::ContextSpecDefinition;
+use crate::link::context_spec_definition::CONTEXT_VERSIONS;
 use crate::link::join_spec_definition::JoinSpecDefinition;
 use crate::link::link_spec_definition::LinkSpecDefinition;
 use crate::link::spec::Identity;
@@ -59,18 +61,23 @@ use crate::subgraph::ValidSubgraph;
 pub use crate::supergraph::ValidFederationSubgraph;
 pub use crate::supergraph::ValidFederationSubgraphs;
 
-pub(crate) type SupergraphSpecs = (&'static LinkSpecDefinition, &'static JoinSpecDefinition);
+pub(crate) type SupergraphSpecs = (
+    &'static LinkSpecDefinition,
+    &'static JoinSpecDefinition,
+    Option<&'static ContextSpecDefinition>,
+);
 
 pub(crate) fn validate_supergraph_for_query_planning(
     supergraph_schema: &FederationSchema,
 ) -> Result<SupergraphSpecs, FederationError> {
-    validate_supergraph(supergraph_schema, &JOIN_VERSIONS)
+    validate_supergraph(supergraph_schema, &JOIN_VERSIONS, &CONTEXT_VERSIONS)
 }
 
 /// Checks that required supergraph directives are in the schema, and returns which ones were used.
 pub(crate) fn validate_supergraph(
     supergraph_schema: &FederationSchema,
     join_versions: &'static SpecDefinitions<JoinSpecDefinition>,
+    context_versions: &'static SpecDefinitions<ContextSpecDefinition>,
 ) -> Result<SupergraphSpecs, FederationError> {
     let Some(metadata) = supergraph_schema.metadata() else {
         return Err(SingleFederationError::InvalidFederationSupergraph {
@@ -94,7 +101,22 @@ pub(crate) fn validate_supergraph(
             ),
         }.into());
     };
-    Ok((link_spec_definition, join_spec_definition))
+    let context_spec_definition = metadata.for_identity(&Identity::context_identity()).map(|context_link| {
+        context_versions.find(&context_link.url.version).map_or_else(|| {
+            Err(SingleFederationError::InvalidFederationSupergraph {
+                message: format!(
+                    "Invalid supergraph: uses unsupported context spec version {} (supported versions: {})",
+                    context_link.url.version,
+                    context_versions.versions().map(|v| v.to_string()).collect::<Vec<_>>().join(", "),
+                ),
+            })
+        }, Ok)
+    }).transpose()?;
+    Ok((
+        link_spec_definition,
+        join_spec_definition,
+        context_spec_definition,
+    ))
 }
 
 pub struct Supergraph {
--- a/apollo-federation/src/link/argument.rs
+++ b/apollo-federation/src/link/argument.rs
@@ -144,7 +144,7 @@ pub(crate) fn directive_optional_list_argument<'a>(
             Value::Null => Ok(None),
             Value::List(values) => Ok(Some(values.as_slice())),
             _ => bail!(
-                r#"Argument "{name}" of directive "@{}" must be a boolean."#,
+                r#"Argument "{name}" of directive "@{}" must be a list."#,
                 application.name
             ),
         },
--- a/apollo-federation/src/link/context_spec_definition.rs
+++ b/apollo-federation/src/link/context_spec_definition.rs
@@ -1,8 +1,13 @@
+use apollo_compiler::ast::Directive;
+use apollo_compiler::ast::DirectiveDefinition;
 use apollo_compiler::name;
 use apollo_compiler::Name;
+use apollo_compiler::Node;
 use lazy_static::lazy_static;
 
 use crate::error::FederationError;
+use crate::error::SingleFederationError;
+use crate::link::argument::directive_required_string_argument;
 use crate::link::spec::Identity;
 use crate::link::spec::Url;
 use crate::link::spec::Version;
@@ -11,7 +16,11 @@ use crate::link::spec_definition::SpecDefinitions;
 use crate::schema::FederationSchema;
 
 pub(crate) const CONTEXT_DIRECTIVE_NAME_IN_SPEC: Name = name!("context");
-pub(crate) const CONTEXT_DIRECTIVE_NAME_DEFAULT: Name = name!("federation__context");
+pub(crate) const CONTEXT_NAME_ARGUMENT_NAME: Name = name!("name");
+
+pub(crate) struct ContextDirectiveArguments<'doc> {
+    pub(crate) name: &'doc str,
+}
 
 #[derive(Clone)]
 pub(crate) struct ContextSpecDefinition {
@@ -30,13 +39,26 @@ impl ContextSpecDefinition {
         }
     }
 
-    pub(crate) fn context_directive_name_in_schema(
+    pub(crate) fn context_directive_definition<'schema>(
         &self,
-        schema: &FederationSchema,
-    ) -> Result<Name, FederationError> {
-        Ok(self
-            .directive_name_in_schema(schema, &CONTEXT_DIRECTIVE_NAME_IN_SPEC)?
-            .unwrap_or(CONTEXT_DIRECTIVE_NAME_DEFAULT))
+        schema: &'schema FederationSchema,
+    ) -> Result<&'schema Node<DirectiveDefinition>, FederationError> {
+        self.directive_definition(schema, &CONTEXT_DIRECTIVE_NAME_IN_SPEC)?
+            .ok_or_else(|| {
+                SingleFederationError::Internal {
+                    message: "Unexpectedly could not find context spec in schema".to_owned(),
+                }
+                .into()
+            })
+    }
+
+    pub(crate) fn context_directive_arguments<'doc>(
+        &self,
+        application: &'doc Node<Directive>,
+    ) -> Result<ContextDirectiveArguments<'doc>, FederationError> {
+        Ok(ContextDirectiveArguments {
+            name: directive_required_string_argument(application, &CONTEXT_NAME_ARGUMENT_NAME)?,
+        })
     }
 }
 
--- a/apollo-federation/src/link/context_spec_definition.rs
+++ b/apollo-federation/src/link/context_spec_definition.rs
@@ -6,7 +6,7 @@ use apollo_compiler::Node;
 use lazy_static::lazy_static;
 
 use crate::error::FederationError;
-use crate::error::SingleFederationError;
+use crate::internal_error;
 use crate::link::argument::directive_required_string_argument;
 use crate::link::spec::Identity;
 use crate::link::spec::Url;
@@ -44,12 +44,7 @@ impl ContextSpecDefinition {
         schema: &'schema FederationSchema,
     ) -> Result<&'schema Node<DirectiveDefinition>, FederationError> {
         self.directive_definition(schema, &CONTEXT_DIRECTIVE_NAME_IN_SPEC)?
-            .ok_or_else(|| {
-                SingleFederationError::Internal {
-                    message: "Unexpectedly could not find context spec in schema".to_owned(),
-                }
-                .into()
-            })
+            .ok_or_else(|| internal_error!("Unexpectedly could not find context spec in schema"))
     }
 
     pub(crate) fn context_directive_arguments<'doc>(
--- a/apollo-federation/src/link/federation_spec_definition.rs
+++ b/apollo-federation/src/link/federation_spec_definition.rs
@@ -469,15 +469,16 @@ impl FederationSpecDefinition {
         })
     }
 
-    pub(crate) fn context_directive_arguments(
-        application: &Node<Directive>,
-    ) -> Result<ContextDirectiveArguments, FederationError> {
+    pub(crate) fn context_directive_arguments<'doc>(
+        &self,
+        application: &'doc Node<Directive>,
+    ) -> Result<ContextDirectiveArguments<'doc>, FederationError> {
         Ok(ContextDirectiveArguments {
             name: directive_required_string_argument(application, &FEDERATION_NAME_ARGUMENT_NAME)?,
         })
     }
 
-    // The directive is named `@fromContex`. This is confusing for clippy, as
+    // The directive is named `@fromContext`. This is confusing for clippy, as
     // `from` is a conventional prefix used in conversion methods, which do not
     // take `self` as an argument. This function does **not** perform
     // conversion, but extracts `@fromContext` directive definition.
@@ -495,24 +496,7 @@ impl FederationSpecDefinition {
             })
     }
 
-    // The directive is named `@fromContex`. This is confusing for clippy, as
-    // `from` is a conventional prefix used in conversion methods, which do not
-    // take `self` as an argument. This function does **not** perform
-    // conversion, but extracts `@fromContext` directive arguments.
-    #[allow(clippy::wrong_self_convention)]
-    pub(crate) fn from_context_directive_arguments<'doc>(
-        &self,
-        application: &'doc Node<Directive>,
-    ) -> Result<FromContextDirectiveArguments<'doc>, FederationError> {
-        Ok(FromContextDirectiveArguments {
-            field: directive_required_string_argument(
-                application,
-                &FEDERATION_FIELD_ARGUMENT_NAME,
-            )?,
-        })
-    }
-
-    // The directive is named `@fromContex`. This is confusing for clippy, as
+    // The directive is named `@fromContext`. This is confusing for clippy, as
     // `from` is a conventional prefix used in conversion methods, which do not
     // take `self` as an argument. This function does **not** perform
     // conversion, but extracts `@fromContext` directive.
@@ -536,6 +520,23 @@ impl FederationSpecDefinition {
         Ok(Directive {
             name: name_in_schema,
             arguments,
+        })
+    }
+
+    // The directive is named `@fromContext`. This is confusing for clippy, as
+    // `from` is a conventional prefix used in conversion methods, which do not
+    // take `self` as an argument. This function does **not** perform
+    // conversion, but extracts `@fromContext` directive arguments.
+    #[allow(clippy::wrong_self_convention)]
+    pub(crate) fn from_context_directive_arguments<'doc>(
+        &self,
+        application: &'doc Node<Directive>,
+    ) -> Result<FromContextDirectiveArguments<'doc>, FederationError> {
+        Ok(FromContextDirectiveArguments {
+            field: directive_required_string_argument(
+                application,
+                &FEDERATION_FIELD_ARGUMENT_NAME,
+            )?,
         })
     }
 
--- a/apollo-federation/src/link/join_spec_definition.rs
+++ b/apollo-federation/src/link/join_spec_definition.rs
@@ -49,7 +49,7 @@ pub(crate) const JOIN_OVERRIDE_LABEL_ARGUMENT_NAME: Name = name!("overrideLabel"
 pub(crate) const JOIN_USEROVERRIDDEN_ARGUMENT_NAME: Name = name!("usedOverridden");
 pub(crate) const JOIN_INTERFACE_ARGUMENT_NAME: Name = name!("interface");
 pub(crate) const JOIN_MEMBER_ARGUMENT_NAME: Name = name!("member");
-pub(crate) const JOIN_MEMBER_CONTEXT_ARGUMENTS: Name = name!("contextArguments");
+pub(crate) const JOIN_CONTEXTARGUMENTS_ARGUMENT_NAME: Name = name!("contextArguments");
 
 pub(crate) struct GraphDirectiveArguments<'doc> {
     pub(crate) name: &'doc str,
@@ -81,7 +81,9 @@ impl<'doc> TryFrom<&'doc Value> for ContextArgument<'doc> {
             value: &'a Value,
         ) -> Result<(), FederationError> {
             if let Some(first_value) = field {
-                bail!("Duplicate contextArgument for '{name}' field: {first_value} and {value}")
+                bail!(
+                    r#"Input field "{name}" in contextArguments was repeated with value "{value}" (previous value was "{first_value}")"#
+                )
             }
             let _ = field.insert(value);
             Ok(())
@@ -94,36 +96,36 @@ impl<'doc> TryFrom<&'doc Value> for ContextArgument<'doc> {
             field
                 .ok_or_else(|| {
                     FederationError::internal(format!(
-                        "'{field_name}' field was missing from contextArgument"
+                        r#"Input field "{field_name}" was missing from contextArguments"#
                     ))
                 })?
                 .as_str()
                 .ok_or_else(|| {
                     FederationError::internal(format!(
-                        "'{field_name}' field of contextArgument was not a string"
+                        r#"Input field "{field_name}" in contextArguments was not a string"#
                     ))
                 })
         }
 
-        let Value::Object(names) = value else {
-            bail!("Item in contextArgument list is not an object {value}")
+        let Value::Object(input_object) = value else {
+            bail!(r#"Item "{value}" in contextArguments list was not an object"#)
         };
         let mut name = None;
         let mut type_ = None;
         let mut context = None;
         let mut selection = None;
-        for (arg_name, value) in names.as_slice() {
-            match arg_name.as_str() {
-                "name" => insert_value(arg_name, &mut name, value)?,
-                "type" => insert_value(arg_name, &mut type_, value)?,
-                "context" => insert_value(arg_name, &mut context, value)?,
-                "selection" => insert_value(arg_name, &mut selection, value)?,
-                _ => bail!("Found unknown contextArgument {arg_name}"),
+        for (input_field_name, value) in input_object {
+            match input_field_name.as_str() {
+                "name" => insert_value(input_field_name, &mut name, value)?,
+                "type" => insert_value(input_field_name, &mut type_, value)?,
+                "context" => insert_value(input_field_name, &mut context, value)?,
+                "selection" => insert_value(input_field_name, &mut selection, value)?,
+                _ => bail!(r#"Found unknown contextArguments input field "{input_field_name}""#),
             }
         }
 
         let name = field_or_else("name", name)?;
-        let type_ = field_or_else("type_", type_)?;
+        let type_ = field_or_else("type", type_)?;
         let context = field_or_else("context", context)?;
         let selection = field_or_else("selection", selection)?;
 
@@ -308,7 +310,7 @@ impl JoinSpecDefinition {
             )?,
             context_arguments: directive_optional_list_argument(
                 application,
-                &JOIN_MEMBER_CONTEXT_ARGUMENTS,
+                &JOIN_CONTEXTARGUMENTS_ARGUMENT_NAME,
             )?
             .map(|values| {
                 values
--- a/apollo-federation/src/link/join_spec_definition.rs
+++ b/apollo-federation/src/link/join_spec_definition.rs
@@ -82,7 +82,7 @@ impl<'doc> TryFrom<&'doc Value> for ContextArgument<'doc> {
         ) -> Result<(), FederationError> {
             if let Some(first_value) = field {
                 bail!(
-                    r#"Input field "{name}" in contextArguments was repeated with value "{value}" (previous value was "{first_value}")"#
+                    r#"Input field "{name}" in contextArguments is repeated with value "{value}" (previous value was "{first_value}")"#
                 )
             }
             let _ = field.insert(value);
@@ -96,19 +96,19 @@ impl<'doc> TryFrom<&'doc Value> for ContextArgument<'doc> {
             field
                 .ok_or_else(|| {
                     FederationError::internal(format!(
-                        r#"Input field "{field_name}" was missing from contextArguments"#
+                        r#"Input field "{field_name}" is missing from contextArguments"#
                     ))
                 })?
                 .as_str()
                 .ok_or_else(|| {
                     FederationError::internal(format!(
-                        r#"Input field "{field_name}" in contextArguments was not a string"#
+                        r#"Input field "{field_name}" in contextArguments is not a string"#
                     ))
                 })
         }
 
         let Value::Object(input_object) = value else {
-            bail!(r#"Item "{value}" in contextArguments list was not an object"#)
+            bail!(r#"Item "{value}" in contextArguments list is not an object"#)
         };
         let mut name = None;
         let mut type_ = None;
--- a/apollo-federation/src/operation/rebase.rs
+++ b/apollo-federation/src/operation/rebase.rs
@@ -22,7 +22,6 @@ use super::SelectionSet;
 use super::TYPENAME_FIELD;
 use crate::ensure;
 use crate::error::FederationError;
-use crate::link::federation_spec_definition::get_federation_spec_definition_from_subgraph;
 use crate::schema::position::CompositeTypeDefinitionPosition;
 use crate::schema::position::OutputTypeDefinitionPosition;
 use crate::schema::ValidFederationSchema;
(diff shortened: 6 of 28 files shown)


apollo-federation/src/supergraph/schema.rs
```diff
@@ -99,10 +99,12 @@ pub(crate) fn new_empty_fed_2_subgraph_schema() -> Result<FederationSchema, Fede
 
     directive @federation__listSize(assumedSize: Int, slicingArguments: [String!], sizedFields: [String!], requireOneSlicingArgument: Boolean = true) on FIELD_DEFINITION
 
-    directive @federation__fromContext(field: String) on ARGUMENT_DEFINITION
+    directive @federation__fromContext(field: federation__ContextFieldValue) on ARGUMENT_DEFINITION
```
sachindshinde, line 102: Bugs:
- The type of `@federation__fromContext(field:)` isn't `String`, but should instead be a custom scalar, as it is in the JS.

apollo-federation/src/supergraph/subgraph.rs
```diff
@@ -11,6 +13,7 @@ pub(super) struct FederationSubgraph {
     pub(super) name: String,
     pub(super) url: String,
     pub(super) schema: FederationSchema,
+    pub(super) graph_enum_value: Name,
```
sachindshinde, line 16: Added to provide an easy mapping from subgraph names to enum values.

apollo-federation/src/supergraph/mod.rs
```diff
@@ -713,16 +752,6 @@ fn extract_object_type_content(
             message: "@join__implements should exist for a fed2 supergraph".to_owned(),
         })?;
 
-    let context = supergraph_schema
```
sachindshinde, line 716: Refactors:
- Instead of refetching this supergraph schema spec each time, this has been moved into `validate_supergraph_for_query_planning()` (where other relevant supergraph schema specs are fetched).

apollo-federation/src/supergraph/mod.rs
```diff
@@ -713,16 +752,6 @@ fn extract_object_type_content(
...
-        .metadata()
-        .and_then(|metadata| metadata.for_identity(&Identity::context_identity()))
-        .and_then(|context_link| CONTEXT_VERSIONS.find(&context_link.url.version))
-        .and_then(|context_spec_def| {
```
sachindshinde, line 720: Bugs:
- When the context version is unknown, we should error out early (this is done now in `validate_supergraph_for_query_planning()`).

apollo-federation/src/supergraph/mod.rs
```diff
@@ -1693,103 +1653,6 @@ lazy_static! {
...
-    directive: Component<Directive>,
-) -> Result<(), FederationError> {
-    match pos {
-        CompositeTypeDefinitionPosition::Union(pos) => pos.insert_directive(schema, directive),
-        CompositeTypeDefinitionPosition::Object(pos) => pos.insert_directive(schema, directive),
-        CompositeTypeDefinitionPosition::Interface(pos) => pos.insert_directive(schema, directive),
-    }
-}
-
-trait CompositeType {
```
sachindshinde, line 1708: Refactors: 
- We end up not needing this `directives()` trait after refactoring (we're effectively using `ExtendedType::directives()` instead).

apollo-federation/src/supergraph/mod.rs
```diff
@@ -1693,103 +1653,6 @@ lazy_static! {
     };
 }
 
-fn insert_directive(
```
sachindshinde, line 1696: Refactors:
- This kind of fan-out generally goes in `position.rs`

apollo-federation/src/supergraph/mod.rs
```diff
@@ -1693,103 +1653,6 @@ lazy_static! {
...
-    }
-}
-
-impl CompositeType for InterfaceType {
-    fn directives(&self) -> &DirectiveList {
-        &self.directives
-    }
-}
-
-fn apply_context_to_type<T>(
```
sachindshinde, line 1730: Refactors:
- This was moved into empty type addition, similar to the JS code. This allows the code to be called once.

apollo-federation/src/supergraph/mod.rs
```diff
@@ -1693,103 +1653,6 @@ lazy_static! {
...
-    graph_enum_value_name_to_subgraph_name: &IndexMap<Name, Arc<str>>,
-    federation_spec_definitions: &IndexMap<Name, &'static FederationSpecDefinition>,
-    context_name_in_supergraph: &Name,
-    pos: &CompositeTypeDefinitionPosition,
-) -> Result<(), FederationError>
-where
-    T: CompositeType,
-{
-    for directive in ty.directives().get_all(context_name_in_supergraph.as_str()) {
-        FederationSpecDefinition::context_directive_arguments(directive).and_then(
```
sachindshinde, line 1742: Bugs:
- We should be using the context spec for this, instead of the federation spec (the structure ends up being similar enough that it's fine here, but they're separate specs).

apollo-federation/src/supergraph/mod.rs
```diff
@@ -1693,103 +1653,6 @@ lazy_static! {
...
-            |context_name| {
-                let mut arr = context_name.name.split("__");
```
sachindshinde, line 1744: Bugs:
- This should be an `rsplit` instead of `split`, since context names can't have underscores but subgraph names can.

apollo-federation/src/supergraph/mod.rs
```diff
@@ -1666,16 +1636,6 @@ fn get_subgraph<'subgraph>(
     })
 }
 
-fn get_index_from_subgraph_name<'a>(
```
sachindshinde, line 1669: Refactors:
- To get the enum value more easily for a subgraph name, it's been added to `FederationSubgraph`.

apollo-federation/src/supergraph/mod.rs
```diff
@@ -1693,103 +1653,6 @@ lazy_static! {
...
-                let subgraph_name = arr.next().ok_or_else(|| {
-                    SingleFederationError::InvalidFederationSupergraph {
-                        message: format!(
-                            "Could not parse context name from supergraph '{}'",
-                            context_name_in_supergraph
-                        )
-                        .to_owned(),
-                    }
-                })?;
-                let subgraph_name = Name::new_unchecked(subgraph_name);
```
sachindshinde, line 1754: Bugs:
- There's not a guarantee that subgraph names are valid GraphQL names (unlike the enum values).

apollo-federation/src/supergraph/mod.rs
```diff
@@ -672,25 +684,52 @@ fn add_empty_type(
                     }
                     .into());
                 }
-                TypeDefinitionPosition::Object(pos) => {
```
sachindshinde, line 675: Refactors:
- Moved this fan-out into `position.rs` alongside the other one mentioned above.

apollo-federation/src/subgraph/spec.rs
```diff
@@ -434,7 +444,8 @@ impl FederationSpecDefinitions {
             arguments: vec![InputValueDefinition {
                 description: None,
                 name: name!("field"),
-                ty: ty!(String!).into(),
```
sachindshinde, line 437: Bugs:
- As noted above, the type should be a custom scalar here.

apollo-federation/src/link/join_spec_definition.rs
```diff
@@ -49,7 +49,7 @@ pub(crate) const JOIN_OVERRIDE_LABEL_ARGUMENT_NAME: Name = name!("overrideLabel"
 pub(crate) const JOIN_USEROVERRIDDEN_ARGUMENT_NAME: Name = name!("usedOverridden");
 pub(crate) const JOIN_INTERFACE_ARGUMENT_NAME: Name = name!("interface");
 pub(crate) const JOIN_MEMBER_ARGUMENT_NAME: Name = name!("member");
-pub(crate) const JOIN_MEMBER_CONTEXT_ARGUMENTS: Name = name!("contextArguments");
```
sachindshinde, line 52: Refactors:
- This file's changes are renames for consistency and more descriptive names, and a bit of error message tweaking.

apollo-federation/src/link/federation_spec_definition.rs
```diff
@@ -469,15 +469,16 @@ impl FederationSpecDefinition {
         })
     }
 
-    pub(crate) fn context_directive_arguments(
```
sachindshinde, line 472: Refactors:
- We generally provide `self` here, as a directive's definition, whether it should be present/absent, or its processing/validations may depend on spec version.

apollo-federation/src/link/federation_spec_definition.rs
```diff
@@ -495,24 +496,7 @@ impl FederationSpecDefinition {
             })
     }
 
-    // The directive is named `@fromContex`. This is confusing for clippy, as
```
sachindshinde, line 498: Conventions:
- This is just a small nit, though spec files seem to have a pattern/order of definition getter, application creator, and argument extractor.

apollo-federation/src/link/argument.rs
```diff
@@ -144,7 +144,7 @@ pub(crate) fn directive_optional_list_argument<'a>(
             Value::Null => Ok(None),
             Value::List(values) => Ok(Some(values.as_slice())),
             _ => bail!(
-                r#"Argument "{name}" of directive "@{}" must be a boolean."#,
```
sachindshinde, line 147: Bugs:
- Minor copy/paste issue with error messaging.

apollo-federation/src/link/context_spec_definition.rs
```diff
@@ -30,13 +39,26 @@ impl ContextSpecDefinition {
...
+            })
+    }
+
+    pub(crate) fn context_directive_arguments<'doc>(
         &self,
-        schema: &FederationSchema,
-    ) -> Result<Name, FederationError> {
-        Ok(self
-            .directive_name_in_schema(schema, &CONTEXT_DIRECTIVE_NAME_IN_SPEC)?
-            .unwrap_or(CONTEXT_DIRECTIVE_NAME_DEFAULT))
```
sachindshinde, line 39: Bugs:
- This only returns `None` when there's no `@core`/`@link` in the schema, which should never be the case for supergraph schemas. (Generally we've been using functions like `context_directive_definition()` instead, which gets you the name/definition and also asserts the definition exists.)

apollo-federation/src/query_graph/mod.rs
```diff
@@ -134,18 +134,24 @@ impl TryFrom<QueryGraphNodeType> for ObjectTypeDefinitionPosition {
...
+    // This is purposely left unparsed in query graphs, due to @fromContext selection sets being
+    // duck-typed.
     selection: String,
     types_with_context_set: IndexSet<CompositeTypeDefinitionPosition>,
+    // PORT_NOTE: This field was renamed because the JS name (`namedParameter`) left confusion to
+    // how it was different from the argument name.
+    argument_name: Name,
     // PORT_NOTE: This field was renamed because the JS name (`coordinate`) was too vague.
     argument_coordinate: ObjectFieldArgumentDefinitionPosition,
-    named_parameter: Name,
```
sachindshinde, line 147: Refactors:
- Some renames here for clarity (argument instead of parameter to make it clear it's an actual GraphQL argument).

apollo-federation/src/query_graph/mod.rs
```diff
@@ -389,10 +397,13 @@ pub struct QueryGraph {
     /// lowered composition validation on a big composition (100+ subgraphs) from ~4 minutes to
     /// ~10 seconds.
     non_trivial_followup_edges: IndexMap<EdgeIndex, Vec<EdgeIndex>>,
-    /// Maps each subgraph name to each field argument of `@fromContext`.
```
sachindshinde, line 392: Refactors:
- Similarly to above, there's renames here for clarity/consistency across QP data structures.
- It turns out `subgraph_to_args` isn't really used anywhere (there's a usage in JS QP, but it's replaceable by `arguments_to_context_ids_by_source`)

apollo-federation/src/operation/rebase.rs
```diff
@@ -276,15 +275,17 @@ impl Field {
         else {
             return Ok(None);
         };
-        if let Ok(federation_spec_definition) = get_federation_spec_definition_from_subgraph(schema)
```
sachindshinde, line 279: Refactors:
- I forgot when writing this that we could just use the subgraph metadata in the schema, I've updated it accordingly.

apollo-federation/src/query_graph/build_query_graph.rs
```diff
@@ -1557,36 +1564,41 @@ impl FederatedQueryGraphBuilder {
                     .or_default()
                     .push(object_field_arg.clone());
                 let field_coordinate = object_field_arg.parent();
-                if let Some(dir) = input_value.directives.get(
```
sachindshinde, line 1560: Bugs:
- If the directive isn't present here, we should `bail!()` instead of silently ignoring it.

apollo-federation/src/query_graph/build_query_graph.rs
```diff
@@ -1557,36 +1564,41 @@ impl FederatedQueryGraphBuilder {
...
                     subgraph_data
                         .from_context_directive_definition_name
                         .as_str(),
-                ) {
-                    let application = subgraph_data
-                        .federation_spec_definition
-                        .from_context_directive_arguments(dir)?;
-                    let (context, selection) = parse_context(application.field)?;
-
-                    let types_with_context_set = context_name_to_types
```
sachindshinde, line 1570: Bugs:
- There should always be an entry in `context_name_to_types` for the context name, if not we should `bail!()` instead of silently ignore it.

apollo-federation/src/query_graph/build_query_graph.rs
```diff
@@ -2333,42 +2344,66 @@ fn resolvable_key_applications<'doc>(
 }
 
 fn parse_context(field: &str) -> Result<(String, String), FederationError> {
```
sachindshinde, line 2346: Bugs:
- The negative look-aheads in the JS regex were dropped here when we should be accounting for them (my guess is it's because the `regex` crate doesn't natively support them). For this particular case, the negative look-aheads are effectively used to prevent backtracking. We can do that manually without negative look-aheads if we split the regex into smaller regexes.

apollo-federation/src/query_graph/condition_resolver.rs
```diff
@@ -23,9 +23,13 @@ pub(crate) struct ContextMapEntry {
     pub(crate) levels_in_query_path: usize,
     pub(crate) path_tree: Option<Arc<OpPathTree>>,
     pub(crate) selection_set: SelectionSet,
-    pub(crate) param_name: Name,
```
sachindshinde, line 26: Refactors:
- Some renames here to match with `ContextCondition` and `QueryGraph`.

apollo-federation/src/query_graph/graph_path.rs
```diff
@@ -178,11 +173,15 @@ where
     // TODO(@TylerBloom): Add in once defer is supported.
     #[serde(skip)]
     defer_on_tail: Option<DeferDirectiveArguments>,
```
sachindshinde, line 175: Refactors:
- Some renames here to match with `ContextCondition` and `ContextMapEntry`.

apollo-federation/src/query_graph/graph_path.rs
```diff
@@ -1303,61 +1282,47 @@ where
...
             self.merge_edge_conditions_with_resolution(&condition_path_tree, &context_map);
-        let last_parameter_to_context = new_parameter_to_context.last();
+        let last_arguments_to_context_usages = new_arguments_to_context_usages.last();
 
-        if let Some(Some(last_parameter_to_context)) = last_parameter_to_context {
+        if let Some(Some(last_arguments_to_context_usages)) = last_arguments_to_context_usages {
             // TODO: Perhaps it is better to explicitly cast this to `GraphPathTriggerRefMut` and
             // pull out the field from there.
             if let Some(field) = trigger.get_field_mut() {
-                let mut schema = field.schema.schema().clone().into_inner();
```
sachindshinde, line 1320: Bugs:
- Instead of creating a fake modified supergraph schema (which then also has to be validated), we should be switching the field to use the relevant subgraph schema of the edge, that's what declared the `@fromContext` arguments to start with.

apollo-federation/src/query_graph/graph_path.rs
```diff
@@ -976,22 +953,22 @@ pub(crate) trait GraphPathTriggerVariant: Eq + Hash + std::fmt::Debug {
         }
     }
 
-    fn get_field_parent_type<'a>(&'a self) -> Option<FieldDefinitionPosition>
```
sachindshinde, line 979: Refactors:
- This is returning the field definition when it should be returning the parent type.

apollo-federation/src/query_graph/graph_path.rs
```diff
@@ -1686,143 +1657,148 @@ where
                         levels_in_data_path += 1;
                     }
                     let Some(e) = (*e).into() else { continue };
-                    if !was_unsatisfied && !context_map.contains_key(&ctx.named_parameter) {
-                        if let Some(parent_type) = parent_type {
-                            let parent_type = parent_type.parent();
-                            let subgraph_schema =
-                                self.graph.schema_by_source(&ctx.subgraph_name)?;
-                            let mut potential_matches: IndexSet<Name> = Default::default();
-                            ctx.types_with_context_set.iter().for_each(|pos| {
```
sachindshinde, line 1695: Bugs:
- The supergraph schema should be used here instead of the subgraph schema.
- The union logic here is inverted, I've added a comment here to better explain it.

apollo-federation/src/query_graph/graph_path.rs
```diff
@@ -1686,143 +1657,148 @@ where
...
+                                return Ok(true);
+                            }
+                        }
+                        Ok(false)
+                    })? {
+                        continue;
+                    }
 
-                            // get the selection set for the first match that parses
-                            let selection_set =
```
sachindshinde, line 1741: Bugs:
- The selection set should be parsed against the supergraph schema, not the subgraph schema.
- Duck typing semantics means the `@fromContext` selection set may not validate against any of the types individually. The fix here is to parse without validation against the parent type in the supergraph schema, remove any "impossible" type conditions, and then validate.

apollo-federation/src/query_graph/graph_path.rs
```diff
@@ -1865,6 +1841,9 @@ where
             excluded_conditions,
             None,
         )?;
+        if matches!(resolution, ConditionResolution::Unsatisfied { .. }) {
```
sachindshinde, line 1844: Bugs:
- We should early return here if the edge's conditions are unsatisfiable.

apollo-federation/src/query_graph/graph_path.rs
```diff
@@ -1916,10 +1895,12 @@ where
             }
         }
         if let ConditionResolution::Satisfied {
+            cost,
             context_map: ctx_map,
             ..
         } = &mut resolution
         {
+            *cost += total_cost;
```
sachindshinde, line 1903: Bugs:
- The cost of context conditions was being ignored when there were also edge conditions.

apollo-federation/src/query_graph/graph_path.rs
```diff
@@ -2350,6 +2331,12 @@ where
         ) -> Option<TEdge>,
         override_conditions: &EnabledOverrideConditions,
     ) -> Result<Option<NodeIndex>, FederationError> {
+        // TODO: Temporary fix to avoid optimization if context exists, a permanent fix is here:
+        //       https://github.com/apollographql/federation/pull/3017#pullrequestreview-2083949094
+        if self.graph.is_context_used() {
```
sachindshinde, line 2336: Bugs:
- This optimization should be disabled when `@fromContext` is in use, as it is in JS QP.

apollo-federation/src/query_graph/graph_path.rs
```diff
@@ -2686,8 +2673,11 @@ impl OpGraphPath {
             runtime_types_before_tail_if_last_is_cast: None,
             // TODO: The JS codebase copied this from the current path, which seems like a bug.
             defer_on_tail: self.defer_on_tail.clone(),
-            context_to_selection: self.context_to_selection.clone(),
```
sachindshinde, line 2689: Bugs:
- These vectors should be truncated similarly to how `edges` is (really we should consider moving to a vector of structs, but that's a follow-up). Note this bug is also present in JS QP.

apollo-federation/src/query_graph/path_tree.rs
```diff
@@ -365,8 +369,18 @@ where
                         (Some(cond_a), Some(cond_b)) => cond_a.equals_same_root(cond_b),
                         _ => false,
                     }
-                    && a.context_to_selection == b.context_to_selection
```
sachindshinde, line 368: Bugs:
- `None` should be considered the same as an empty collection, similar to JS. (Really we should probably remove the `Option` wrapper in the data structures, but will leave for a follow-up.)

apollo-router/tests/snapshots/set_context__set_context_union_rust_qp.snap
```diff
@@ -47,15 +47,6 @@ expression: response
                       "scopes": []
                     },
                     "contextRewrites": [
-                      {
```
sachindshinde, line 50: Bugs:
- This snapshot differs from `apollo-router/tests/snapshots/set_context__set_context_union.snap`, and this PR change brings them back into alignment. Not sure why we allowed the difference there.

apollo-federation/src/link/join_spec_definition.rs
```diff
@@ -49,7 +49,7 @@ pub(crate) const JOIN_OVERRIDE_LABEL_ARGUMENT_NAME: Name = name!("overrideLabel"
...
+pub(crate) const JOIN_CONTEXTARGUMENTS_ARGUMENT_NAME: Name = name!("contextArguments");
```
lrlna, line 52: typo?
```suggestion
pub(crate) const JOIN_CONTEXT_ARGUMENTS_ARGUMENT_NAME: Name = name!("contextArguments");
```

sachindshinde: When I looked at other const names in this file for multi-word GraphQL arguments, it seemed to be joining them into one word (e.g. `JOIN_USEROVERRIDDEN_ARGUMENT_NAME` for `userOverridden`, `JOIN_UNIONMEMBER_DIRECTIVE_NAME_IN_SPEC` for `unionMember`, `JOIN_ISINTERFACEOBJECT_ARGUMENT_NAME` for `isInterfaceObject`). What's in the PR follows that current convention, but I'm also down to change that convention if we think it's less confusing.

apollo-federation/src/query_graph/mod.rs
```diff
@@ -548,6 +565,26 @@ impl QueryGraph {
...
+        argument: &ObjectFieldArgumentDefinitionPosition,
+    ) -> Result<&Name, FederationError> {
+        self.arguments_to_context_ids_by_source
+            .get(source)
+            .and_then(|r| r.get(argument))
+            .ok_or_else(|| {
+                SingleFederationError::Internal {
+                    message: "context ID unexpectedly missing for @fromContext argument".to_owned(),
+                }
+                .into()
```
lrlna, line 580: should this be `internal_error!()`?

apollo-federation/src/schema/field_set.rs
```diff
@@ -166,6 +166,38 @@ pub(crate) fn collect_target_fields_from_field_set(
...
+    // string.
+    Ok(FieldSet::parse(
+        schema.schema(),
+        parent_type_name,
+        field_value,
+        "field_set.graphql",
+    )?)
+}
+
+pub(crate) fn validate_field_value(
```
lrlna, line 184: This can probably use a little note on why we need to strip check for absence of aliases.

apollo-federation/src/subgraph/spec.rs
```diff
@@ -51,6 +51,7 @@ pub const REQUIRES_DIRECTIVE_NAME: Name = name!("requires");
 pub const SHAREABLE_DIRECTIVE_NAME: Name = name!("shareable");
 pub const TAG_DIRECTIVE_NAME: Name = name!("tag");
 pub const FIELDSET_SCALAR_NAME: Name = name!("FieldSet");
+pub const CONTEXTFIELDVALUE_SCALAR_NAME: Name = name!("ContextFieldValue");
```
lrlna, line 54: a typo?
```suggestion
pub const CONTEXT_FIELD_VALUE_SCALAR_NAME: Name = name!("ContextFieldValue");
```

apollo-federation/src/subgraph/spec.rs
```diff
@@ -320,6 +321,15 @@ impl FederationSpecDefinitions {
         }
     }
 
+    /// scalar ContextFieldValue
+    pub fn contextfieldvalue_scalar_definition(&self, alias: &Option<Name>) -> ScalarType {
```
lrlna, line 325: probably also a typo?
```suggestion
    pub fn context_field_value_scalar_definition(&self, alias: &Option<Name>) -> ScalarType {
```

apollo-federation/src/supergraph/mod.rs
```diff
@@ -392,12 +395,14 @@ fn extract_subgraphs_from_fed_2_supergraph(
     Ok(())
 }
 
+#[allow(clippy::too_many_arguments)]
```
lrlna, line 398: don't love this, but I understand that it'll take a lot more time to refactor properly

lrlna (commented):
It looks like there are a few typos that probably got in via a search and replace. It'd be good to address those before this gets merged in.

I'd like to make sure the diagnostics language stays the same - this PR uses a past tense for a couple of messages. Can we change those to present tense to align with apollo-rs diagnostics? I am not going to "request changes" so this can land while I am asleep, but this is the main thing I'd like addressed before the merge.

apollo-federation/src/link/join_spec_definition.rs
lrlna: oh interesting, that's really hard to read (for me personally). but let's not change that convention in this PR, it can be done later on.

apollo-federation/src/query_graph/graph_path.rs
```diff
@@ -178,11 +173,15 @@ where
...
     defer_on_tail: Option<DeferDirectiveArguments>,
-    /// At the point where a `@context` is set, we will have fields to select
-    context_to_selection: Vec<Option<ContextToSelection>>,
-    /// Where a context is used (i.e. `@fromContext`) there will exist a ContextAtUsageEntry map
-    /// (1 for each parameter)
-    parameter_to_context: Vec<Option<ParameterToContext>>,
+    // PORT_NOTE: This field was renamed because the JS name (`contextToSelection`) implied it was
+    // a map to selections, which it isn't.
+    /// The IDs of contexts set at the head of an edge, for each edge in the path.
+    set_context_ids: Vec<Option<SetContextIds>>,
```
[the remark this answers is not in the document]
sachindshinde: The intended implication with `set_` here isn't that the data structure is a set, but rather that the edge (specifically the head) "sets" the context.

"Set" might be too common of a word here, maybe a different word like `matching_context_ids` might help (we do already have the concept of finding a "match" on the graph path for a given context).

apollo-federation/src/subgraph/spec.rs
sachindshinde: This case is [similar to above](https://github.com/apollographql/router/pull/6380#discussion_r1868170997), where we're following an existing convention (e.g. `FIELDSET_SCALAR_NAME` for `FieldSet`). Definitely down to change the convention across the crate in a follow-up.

apollo-federation/src/subgraph/spec.rs
sachindshinde: This case is [similar to above](https://github.com/apollographql/router/pull/6380#discussion_r1868170997), where we're following an existing convention (e.g. `fieldset_scalar_definition()` for `FieldSet`). Definitely down to change the convention across the crate in a follow-up.

apollo-federation/src/link/join_spec_definition.rs
```diff
@@ -81,7 +81,9 @@ impl<'doc> TryFrom<&'doc Value> for ContextArgument<'doc> {
             value: &'a Value,
         ) -> Result<(), FederationError> {
             if let Some(first_value) = field {
-                bail!("Duplicate contextArgument for '{name}' field: {first_value} and {value}")
+                bail!(
+                    r#"Input field "{name}" in contextArguments was repeated with value "{value}" (previous value was "{first_value}")"#
```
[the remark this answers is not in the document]
sachindshinde: Good catch on the tense fixes!

apollo-federation/src/schema/field_set.rs
sachindshinde: Addressed in [259fe72](https://github.com/apollographql/router/pull/6380/commits/259fe723715f9466bbdc6c7a9430766df73e98d7).

apollo-federation/src/query_graph/mod.rs
sachindshinde: Addressed in [259fe72](https://github.com/apollographql/router/pull/6380/commits/259fe723715f9466bbdc6c7a9430766df73e98d7).

apollo-federation/src/query_graph/graph_path.rs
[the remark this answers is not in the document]
sachindshinde: Addressed in [259fe72](https://github.com/apollographql/router/pull/6380/commits/259fe723715f9466bbdc6c7a9430766df73e98d7).

apollo-federation/src/link/join_spec_definition.rs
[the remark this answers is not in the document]
sachindshinde: Addressed in [a38626f](https://github.com/apollographql/router/pull/6380/commits/a38626fefff7048e4f99d876a73bc974d25d7be8).

apollo-federation/src/link/join_spec_definition.rs
```diff
@@ -94,36 +96,36 @@ impl<'doc> TryFrom<&'doc Value> for ContextArgument<'doc> {
...
                     FederationError::internal(format!(
-                        "'{field_name}' field was missing from contextArgument"
+                        r#"Input field "{field_name}" was missing from contextArguments"#
                     ))
                 })?
                 .as_str()
                 .ok_or_else(|| {
                     FederationError::internal(format!(
-                        "'{field_name}' field of contextArgument was not a string"
+                        r#"Input field "{field_name}" in contextArguments was not a string"#
```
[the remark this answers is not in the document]
sachindshinde: Addressed in [a38626f](https://github.com/apollographql/router/pull/6380/commits/a38626fefff7048e4f99d876a73bc974d25d7be8).

apollo-federation/src/link/join_spec_definition.rs
```diff
@@ -94,36 +96,36 @@ impl<'doc> TryFrom<&'doc Value> for ContextArgument<'doc> {
...
                     ))
                 })
         }
 
-        let Value::Object(names) = value else {
-            bail!("Item in contextArgument list is not an object {value}")
+        let Value::Object(input_object) = value else {
+            bail!(r#"Item "{value}" in contextArguments list was not an object"#)
```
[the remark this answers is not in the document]
sachindshinde: Addressed in [a38626f](https://github.com/apollographql/router/pull/6380/commits/a38626fefff7048e4f99d876a73bc974d25d7be8).

TylerBloom (commented):
Looks good overall. I have a few comment and several very minor things.

apollo-federation/src/query_graph/build_query_graph.rs
```diff
@@ -2333,42 +2344,66 @@ fn resolvable_key_applications<'doc>(
...
+    // There's a fancy_regex crate which does support this, but in the specific case above, the
+    // negative lookaheads are just used to ensure strict *-greediness for the preceding expression
+    // (i.e., it guarantees those *-expressions match greedily and won't backtrack).
+    //
+    // We can emulate that in this case by matching a series of regexes instead of a single regex,
+    // where for each regex, the relevant *-expression doesn't backtrack by virtue of the rest of
+    // the haystack guaranteeing a match. Also note that Rust has (?s:.) to match all characters
+    // including newlines, which we use in place of JS's common regex workaround of [\s\S].
+    fn strip_leading_ignored_tokens(input: &str) -> Result<&str, FederationError> {
+        let leading_pattern = Regex::new(r#"^(?:[\n\r\t ,]|#[^\n\r]*)*((?s:.)*)$"#).unwrap();
```
dariuszkuc, line 2359: nit: should we make this regex static so we don't need to recreate it?

apollo-federation/src/link/federation_spec_definition.rs
sachindshinde: Mainly it's just to prevent people from forgetting to account for spec version in these kinds of functions (or forgoing it because its too much extra work).

As an example, imagine in a future spec bump that this directive allows users to specify more information in a backwards compatible way (e.g. a new argument, or a new syntax for an existing string-ish argument). It would be very tempting in such cases to keep the method static, and to just use the "latest" directive's semantics to parse (after all, its "backwards compatible"). However, this leads to bugs where code can accidentally permit that new information with old specs when it shouldn't. Some of this is prevented by the GraphQL type system itself (e.g. new arguments via directive definition validation), but things beyond that type system won't be (e.g. new syntaxes for existing string-ish arguments).

apollo-federation/src/query_graph/path_tree.rs
sachindshinde: There gist is that in this case, there's no distinction, and we should remove the `Option` (both for matching context IDs, and for maps of arguments to context usages).

apollo-federation/src/lib.rs
```diff
@@ -94,7 +101,22 @@ pub(crate) fn validate_supergraph(
             ),
         }.into());
     };
-    Ok((link_spec_definition, join_spec_definition))
+    let context_spec_definition = metadata.for_identity(&Identity::context_identity()).map(|context_link| {
+        context_versions.find(&context_link.url.version).map_or_else(|| {
```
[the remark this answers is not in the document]
sachindshinde: Addressed in [d71dff0](https://github.com/apollographql/router/pull/6380/commits/d71dff0f9845031253f11fd2e5aaf6d4d0573279).

apollo-federation/src/lib.rs
```diff
@@ -94,7 +101,22 @@ pub(crate) fn validate_supergraph(
...
+            Err(SingleFederationError::InvalidFederationSupergraph {
+                message: format!(
+                    "Invalid supergraph: uses unsupported context spec version {} (supported versions: {})",
+                    context_link.url.version,
+                    context_versions.versions().map(|v| v.to_string()).collect::<Vec<_>>().join(", "),
```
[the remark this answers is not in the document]
sachindshinde: Addressed in [d71dff0](https://github.com/apollographql/router/pull/6380/commits/d71dff0f9845031253f11fd2e5aaf6d4d0573279).

apollo-federation/src/link/context_spec_definition.rs
```diff
@@ -30,13 +39,26 @@ impl ContextSpecDefinition {
...
+        &self,
+        schema: &'schema FederationSchema,
+    ) -> Result<&'schema Node<DirectiveDefinition>, FederationError> {
+        self.directive_definition(schema, &CONTEXT_DIRECTIVE_NAME_IN_SPEC)?
+            .ok_or_else(|| {
+                SingleFederationError::Internal {
+                    message: "Unexpectedly could not find context spec in schema".to_owned(),
+                }
+                .into()
+            })
```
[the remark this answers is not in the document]
sachindshinde: Addressed in [d71dff0](https://github.com/apollographql/router/pull/6380/commits/d71dff0f9845031253f11fd2e5aaf6d4d0573279).

apollo-federation/src/query_graph/build_query_graph.rs
sachindshinde: No worries! A test makes sense, added it in [d71dff0](https://github.com/apollographql/router/pull/6380/commits/d71dff0f9845031253f11fd2e5aaf6d4d0573279).

apollo-federation/src/query_plan/fetch_dependency_graph.rs
```diff
@@ -685,7 +685,7 @@ impl FetchDependencyGraphNodePath {
 }
 
 /// If the `iter` yields a single element, return it. Else return `None`.
-fn iter_into_single_item<T>(mut iter: impl Iterator<Item = T>) -> Option<T> {
+pub(crate) fn iter_into_single_item<T>(mut iter: impl Iterator<Item = T>) -> Option<T> {
```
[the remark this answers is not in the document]
sachindshinde: Addressed in [d71dff0](https://github.com/apollographql/router/pull/6380/commits/d71dff0f9845031253f11fd2e5aaf6d4d0573279).

apollo-federation/src/query_graph/graph_path.rs
```diff
@@ -916,11 +915,13 @@ impl TryFrom<GraphPathTrigger> for Arc<QueryGraphEdgeTransition> {
     }
 }
 
+#[derive(derive_more::From)]
```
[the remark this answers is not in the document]
sachindshinde: No worries, a lot of macros to keep track of haha.

apollo-federation/src/query_graph/graph_path.rs
```diff
@@ -1450,48 +1417,55 @@ where
...
+        debug_assert_eq!(self.edges.len(), self.arguments_to_context_usages.len());
         self.edges
             .iter()
             .copied()
             .zip(&self.edge_triggers)
             .zip(&self.edge_conditions)
-            .zip(&self.context_to_selection)
-            .zip(&self.parameter_to_context)
+            .zip(&self.matching_context_ids)
+            .zip(&self.arguments_to_context_usages)
```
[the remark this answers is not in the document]
sachindshinde: Oh nice, didn't know about `izip!`, it removes the nested tuples too. Updated it in [d71dff0](https://github.com/apollographql/router/pull/6380/commits/d71dff0f9845031253f11fd2e5aaf6d4d0573279).

Also while looking at the iterator, I noticed we were doing some collection cloning, I've changed those to `.as_ref()`s and moved the cloning to users of the iterator.

apollo-federation/src/query_graph/build_query_graph.rs
sachindshinde: We can make it work with `lazy_static!`, addressed in [d71dff0](https://github.com/apollographql/router/pull/6380/commits/d71dff0f9845031253f11fd2e5aaf6d4d0573279).

apollo-federation/src/supergraph/mod.rs
clenfest: Isn't that why we were using `new_unchecked`?

clenfest (approved):
Sorry it took me longer than expected, but I've gone through the changes at a high level and they all make sense to me. I know there is some outstanding feedback that you're looking at, but I'm going to approve so we can get it merged when you're ready.

apollo-federation/src/supergraph/mod.rs
sachindshinde: Right, but the main thing is that QP shouldn't assert that subgraph names are valid GraphQL names, since that's a breaking change. Subgraph names can have type `Arc<str>` instead.

sachindshinde:
Added one last change to align context ID generation with the JS code. (It reduces mismatches, and I figure we can relax those changes when we relax the other ordering changes we made specifically to align with JS.)

@clenfest 
It looks like pushing that last commit auto-dismissed your review, didn't mean for that. I think I've addressed all feedback, and the PR should be ready to merge.

lrlna (approved):

PR #6382, opened 2024-12-03
regular merge of dev into next

garypen:
Merge dev at commit: 7eb9b0b52af483d3b56ded11b82c49d169a7a37e

Notes:
1) I wasn't totally sure how to resolve this conflict:
```
apollo-federation/src/schema/position.rs

<<<<<<< HEAD
            return Err(SingleFederationError::Internal {
                message: format!("Cannot insert reserved enum value \"{}\"", self),
            }
            .into());
=======
            bail!(r#"Cannot insert reserved enum value "{self}""#);
>>>>>>> dev
```

I examined the code and the history and concluded that taking `dev` looked like the correct answer. @lennyburdette Can you let me know what your opinion is?

2) After merging, the code failed to compile because:

```
apollo-federation/src/supergraph/mod.rs

fn insert_directive(
    pos: &CompositeTypeDefinitionPosition,
    directive: Component<Directive>,
)
```

There was no `Directive` and it recommended that I: `use apollo_compiler::ast::Directive;`. I did that in it compiled fine. I think this is ok, but I'm not sure why it compiled on dev, but not on next.


goto-bus-stop:
Keeping the `bail!()` version is the right way to go, yes

lennyburdette (approved):
I also don't know why the imports changed but that's the solution I went with too

PR #6390, opened 2024-12-04
Handle escape sequences in GraphQL strings

pubmodmatt:
As defined in the [GraphQL spec](https://spec.graphql.org/October2021/#sec-String-Value), standard strings may contain escape sequences, and block strings have their whitespace modified to remove any common indent or leading whitespace lines.

The Apollo Connectors feature parses various GraphQL strings in the `@connect` and `@source` directives, and provides messages with location information when there are problems within the content of those strings. To prevent each parser from having to handle escape sequences, the parsers should execute on the strings returned by `apollo-compiler`, which processes the escape sequences and block string whitespace. This requires mapping the locations returned by the parsers back to the corresponding locations in the original file by reversing the rules for handling escape sequences and whitespace. These changes provide that mapping.

## Limitations

The following known limitations still exist:

* The GraphQL spec allows line endings to be `\n`, `\r\n`, or just  a single `\r` character. The location mapping does not function properly with the last case. The `line-col` dependency used to convert offsets to line/column values only handles `\n` newlines, and as such, this code does not attempt to handle that case either.
* Unicode surrogate pairs cannot be used in escape sequences, due to an [open issue in the `apollo-rs` dependency](https://github.com/apollographql/apollo-rs/issues/657). Strings containing these sequences will not be recognized as GraphQL strings by the compiler. An example is the codepoint pair `\uD83E\uDD80`, which corresponds to the 🦀 character.
* Some of the parsers, such as the JSON Selection parser, do not appear to handle `\b` (backspace) and `\f`(form feed) escape sequences allowed by the GraphQL spec. Strings containing those characters will result in parse errors.

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [x] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[CNN-528]: https://apollographql.atlassian.net/browse/CNN-528?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apollo-federation/src/sources/connect/json_selection/parser.rs
+++ b/apollo-federation/src/sources/connect/json_selection/parser.rs
@@ -116,11 +116,7 @@ impl std::error::Error for JSONSelectionParseError {}
 
 impl Display for JSONSelectionParseError {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(
-            f,
-            "{}: {}",
-            self.message, self.fragment
-        )
+        write!(f, "{}: {}", self.message, self.fragment)
     }
 }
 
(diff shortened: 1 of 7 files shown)


lennyburdette (approved):
this is awesome, you should show this to the apollo-rs folks!

PR #6399, opened 2024-12-04
Cleanup unused bypass QP functionality

dariuszkuc:
Cleanup query planner code by removing unused bypass QP functionality. This was a gateway only feature that was never supported on the router.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-federation/cli/src/main.rs
+++ b/apollo-federation/cli/src/main.rs
@@ -35,10 +35,6 @@ struct QueryPlannerArgs {
     /// Set the `debug.paths_limit` option.
     #[arg(long)]
     paths_limit: Option<u32>,
-    /// If the supergraph only represents a single subgraph, pass through queries directly without
-    /// planning.
-    #[arg(long, default_value_t = false)]
-    single_subgraph_passthrough: bool,
 }
 
 /// CLI arguments. See <https://docs.rs/clap/latest/clap/_derive/index.html>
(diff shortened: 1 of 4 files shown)


duckki (approved):
Looks good!

dariuszkuc:
Looks like this one also impacts the QP redis cache tests -> I'll wait for the https://github.com/apollographql/router/pull/6354 before fixing this

goto-bus-stop:
Love to see it, thanks for doing that 😁

lrlna:
> Looks like this one also impacts the QP redis cache tests -> I'll wait for the https://github.com/apollographql/router/pull/6354 before fixing this

Yea, QueryPlannerConfig is part of the cache key (which is generally good, but bad for when we change the config)

dariuszkuc:
@lrlna @SimonSapin am I missing something around redis cache key tests? I'd expect that this change should have also affected the `query_planner_redis_update_reuse_query_fragments` test but it seems like it didn't update the cache key? Just want to double check whether it is expected -> I'm guessing it didn't change the expected cache key there as in order to reuse fragments we have to explicitly switch back to the legacy QP and this config change only affects the native QP.

lrlna:
we talked about this in a huddle, but for historical sake and anyone observing from a distance:

* `query_planner_redis_update_reuse_query_fragments` did not need its cache keys changed
   * it's using the legacy query planner explicitly with the legacy query planner's config
   * legacy query planner config was not affected as part of these changes

PR #6401, opened 2024-12-04
validation for explicit `@key`s

lennyburdette:
If the author defines a resolvable `@key`, they are declaring that there's an entity resolver for that type.

When creating fetch targets for each connector, we create edge conditions from the variables used in the connector, ignoring the author-defined `@key`s.

Defining `@key`s is helpful to avoid needing `@shareable` (keys are always shareable), and to silence hints about inconsistent keys across subgraphs.

This validation avoids confusion if a key is defined that doesn't match the implementation. Defining a resolvable `@key` requires that the entity resolver is implemented either as:

- An explicit `entity: true` connector
- An "implicit" entity resolver for resolving the non-root parent type of a connected field

```graphql
type T @key(fields: "id") {
  id: ID!
  # this creates an implicit entity resolver so that we can "enter" the subgraph via T to reach T.f
  f: String @connect(http: { GET: "http://x/${this.id}" } selection: "$")
}
```

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-federation/src/sources/connect/validation/entity/keys.rs
```diff
@@ -0,0 +1,215 @@
...
+        })
+        .unique()
+        .collect_vec();
+
+    if params.is_empty() {
+        return Ok(None);
+    }
+
+    let mut keys = Vec::with_capacity(params.len());
+    for VariableReference { path, .. } in params {
```
lennyburdette, line 156: feedback from @nicholascioli — let's merge the variables here (`$args.a.b.c` + `$args.a.b.c` = `a { b { c d } }`) and then we can get rid of the optimize_field_sets fn entirely!

apollo-federation/src/sources/connect/validation/extended_type.rs
```diff
@@ -311,6 +341,55 @@ fn validate_field(
...
+                    .flatten(),
+            )
+            .chain(
+                url_template
+                    .as_ref()
+                    .map(|(u, _)| u.variables().cloned())
+                    .into_iter()
+                    .flatten(),
+            )
+            .collect_vec();
```
pubmodmatt, line 364: Should we be including headers here? I'm thinking an API might take a parameter as a header instead of in the URL path.

apollo-federation/src/sources/connect/validation/entity/keys.rs
```diff
@@ -0,0 +1,215 @@
...
+struct FieldAndSelection<'a> {
+    field_name: &'a str,
+    sub_selection: String,
+}
+
+impl<'a> FieldAndSelection<'a> {
+    /// Extract from a path like `a.b.c` into `a` and `b { c }`
+    fn from_path<I: IntoIterator<Item = &'a str>>(parts: I) -> Self {
+        let mut parts = parts.into_iter().peekable();
+        let field_name = parts.next().unwrap_or_default();
```
pubmodmatt, line 189: This seems like silent failure if the path was empty - it will end up with a `FieldAndSelection` whose `field_name` is the empty string.

apollo-federation/src/sources/connect/validation/entity/keys.rs
```diff
@@ -0,0 +1,215 @@
...
+use crate::sources::connect::Namespace;
+
+/// Collects keys and entity connectors for comparison and validation.
+#[derive(Default)]
+pub(crate) struct EntityKeyChecker<'schema> {
+    /// Any time we see `type T @key(fields: "f")` (with resolvable: true)
+    resolvable_keys: Vec<(Valid<FieldSet>, &'schema Node<Directive>, &'schema Name)>,
+    /// Any time we see either:
+    /// - `type Query { t(f: X): T @connect(entity: true) }` (Explicit entity resolver)
+    /// - `type T { f: X g: Y @connect(... $this.f ...) }``  (Implicit entity resolver)
```
pubmodmatt, line 29: Nit: you have an extra ` character here

apollo-federation/src/sources/connect/validation/extended_type.rs
```diff
@@ -346,3 +425,15 @@ fn get_missing_connect_directive_message(
         locations: field.line_column_range(source_map).into_iter().collect(),
     }
 }
+
+/// Extract all seen parameters from a JSONSelection
+///
+/// TODO: this is copied from expand/mod.rs
```
pubmodmatt, line 431: It seems like this should be fairly trivial to add to `JSONSelection` - is there a reason it can't be fixed in this PR?

apollo-federation/src/sources/connect/validation/extended_type.rs
lennyburdette: definitely, good call. we need to update expansion for this too. this is a good reason to consolidate that code. nick and i came up with a plan yesterday

apollo-federation/src/sources/connect/validation/entity/keys.rs
lennyburdette: i have a plan to get rid of this code entirely

apollo-federation/src/sources/connect/validation/entity/keys.rs
lennyburdette, line 29:
```suggestion
    /// - `type T { f: X g: Y @connect(... $this.f ...) }`  (Implicit entity resolver)
```

apollo-federation/src/sources/connect/validation/extended_type.rs
```diff
@@ -274,7 +304,7 @@ fn validate_field(
                 schema,
             ));
 
-            if let Some((template, coordinate)) = url_template {
+            if let Some((template, coordinate)) = url_template.clone() {
```
nicholascioli, line 307: Does this need to be cloned? Would `Option::as_ref` work otherwise?

apollo-federation/src/sources/connect/validation/extended_type.rs
```diff
@@ -288,7 +318,7 @@ fn validate_field(
                     })
                 }
             }
-        } else if let Some((template, coordinate)) = url_template {
+        } else if let Some((template, coordinate)) = url_template.clone() {
```
nicholascioli, line 321: Does this need to be cloned? Would Option::as_ref work otherwise?

nicholascioli (commented):
Looks good! I left a few comments.

nicholascioli:
Some of my old reviews stuck around; please ignore!

apollo-federation/src/sources/connect/validation/extended_type.rs
lennyburdette: killed this function entirely

apollo-federation/src/sources/connect/validation/extended_type.rs
lennyburdette: reverted this

apollo-federation/src/sources/connect/validation/extended_type.rs
lennyburdette: reverted this

apollo-federation/src/sources/connect/validation/extended_type.rs
lennyburdette: this very valid request turned into a major exercise. i didn't want to replicate source/connect header merging logic, so i redid a bunch of stuff to allow using the Connector model struct in validations. once i did that, i was able to somewhat simplify the expansion code by moving `@key` generation into the model layer.

apollo-federation/src/sources/connect/expand/tests/snapshots/it_expand_supergraph@simple.graphql-3.snap
```diff
@@ -47,7 +47,7 @@ enum join__Graph {
   GRAPHQL @join__graph(name: "graphql", url: "https://graphql")
 }
 
-type User @join__type(graph: CONNECTORS_QUERY_USER_0, key: "id") @join__type(graph: CONNECTORS_QUERY_USERS_0) @join__type(graph: CONNECTORS_USER_D_1, key: "b c") @join__type(graph: GRAPHQL, key: "id") {
+type User @join__type(graph: CONNECTORS_QUERY_USER_0, key: "id") @join__type(graph: CONNECTORS_QUERY_USERS_0) @join__type(graph: CONNECTORS_USER_D_1, key: "c b") @join__type(graph: GRAPHQL, key: "id") {
```
nicholascioli, line 50: This smells like non-determinism. Should we always sort the keys?

apollo-federation/src/sources/connect/models.rs
```diff
@@ -91,72 +105,122 @@ impl Connector {
...
+        connect: ConnectDirectiveArguments,
+        source_arguments: &[SourceDirectiveArguments],
+    ) -> Result<(ConnectId, Self), FederationError> {
+        let source = if let Some(source_name) = connect.source {
+            source_arguments
+                .iter()
+                .find(|source| source.name == source_name)
+        } else {
+            None
+        };
```
nicholascioli, line 125: Small nit: might be nice to use `Option::and_then` here:

```suggestion
        let source = connect.source.as_ref()
            .and_then(|name| source_arguments.iter()
                .find(|s| s.name == name));
```

apollo-federation/src/sources/connect/models.rs
```diff
@@ -91,72 +105,122 @@ impl Connector {
...
 
-                let entity_resolver = match (args.entity, on_root_type) {
-                    (true, _) => Some(EntityResolver::Explicit),
-                    (_, false) => Some(EntityResolver::Implicit),
+        let source_name = source.map(|s| s.name.clone());
+        let connect_http = connect.http.expect("@connect http missing");
```
nicholascioli, line 128: Is it OK to panic here?

apollo-federation/src/sources/connect/models.rs
```diff
@@ -91,72 +105,122 @@ impl Connector {
...
+                else {
+                    return Ok(None);
+                };
+                make_key_field_set_from_variables(
+                    schema,
+                    output_type,
+                    self.variable_references(),
+                    EntityResolver::Explicit,
+                )
+                .map_err(|_| ())
```
nicholascioli, line 213: Is it OK to throw away the error here (and on the other match branch)?

apollo-federation/src/sources/connect/models.rs
```diff
@@ -91,72 +105,122 @@ impl Connector {
...
+        self.transport.variable_references().chain(
+            self.selection
+                .external_var_paths()
+                .into_iter()
+                .flat_map(PathSelection::variable_reference),
+        )
+    }
+
+    /// Create a field set for a `@key` using $args and $this variables.
+    pub(crate) fn resolvable_key(&self, schema: &Schema) -> Result<Option<Valid<FieldSet>>, ()> {
```
nicholascioli, line 194: Super small nit: it seems weird to use `Result` here and not have any data on the error case, especially if the caller of this method always maps the (empty) error into a custom error. Maybe consider making an enum for the three possible states?

Or this is overkill and just maybe consider adding some docs here as to what an empty error means in this context.

apollo-federation/src/sources/connect/models.rs
```diff
@@ -91,72 +105,122 @@ impl Connector {
...
+            None => Ok(None),
+            Some(EntityResolver::Explicit) => {
+                let Ok(output_type) = self
+                    .id
+                    .directive
+                    .field
+                    .get(schema)
+                    .map(|f| f.ty.inner_named_type())
+                else {
+                    return Ok(None);
```
nicholascioli, line 205: Is it OK to not fail if the schema does not actually include the field?

apollo-federation/src/sources/connect/models.rs
```diff
@@ -227,21 +291,26 @@ impl HttpJsonTransport {
...
-        let url_variables = self
-            .connect_template
-            .variables()
-            .map(|v| v.namespace.namespace);
+        self.variable_references()
+            .map(|var_ref| var_ref.namespace.namespace)
+    }
+
+    fn variable_references(&self) -> impl Iterator<Item = VariableReference<Namespace>> + '_ {
+        let url_variables = self.connect_template.variables().map(Clone::clone);
```
nicholascioli, line 299: [Iterator::cloned](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.cloned)

```suggestion
        let url_variables = self.connect_template.variables().cloned();
```

apollo-federation/src/sources/connect/models.rs
```diff
@@ -227,21 +291,26 @@ impl HttpJsonTransport {
...
                 HeaderSource::From(_) => None,
-                HeaderSource::Value(source) => {
-                    Some(source.variable_references().map(|v| v.namespace.namespace))
-                }
+                HeaderSource::Value(source) => Some(source.variable_references()),
             })
-            .flatten();
-        let body_variables = self.body.iter().flat_map(|b| b.external_variables());
+            .flatten()
+            .map(Clone::clone);
```
nicholascioli, line 308:
```suggestion
            .cloned();
```

nicholascioli (commented):
Looks good! Just a few nits; feel free to ignore.

apollo-federation/src/sources/connect/models.rs
lennyburdette: ah yeah i was being lazy here. i should use some kind of internal error

apollo-federation/src/sources/connect/models.rs
lennyburdette: i'll switch this to an internal error

apollo-federation/src/sources/connect/models.rs
```diff
@@ -198,23 +205,29 @@ impl Connector {
                     .get(schema)
                     .map(|f| f.ty.inner_named_type())
                 else {
-                    return Ok(None);
+                    return Err(internal_error!(
```
nicholascioli, line 208: Super pedantic, but maybe a `Result::map_err` could be used here and then no need to do the whole `let Ok(output_type) = ...`.

nicholascioli (approved):
The tiniest of nits, but otherwise looks great!

PR #6402, opened 2024-12-05
feat: update include_subgraph_errors with extensions customization

Samjin:
WIP - test cases are not updated yet. Need apollo team's opinion over the proposed config below.

This is not a breaking change, the default behavior of `include_subgraph_errors` remains the same.

```.yaml
include_subgraph_errors:
  all: true # Same as before, it will NOT redact anything

include_subgraph_errors: 
  all: false # Same as before, redact all subgraph errors.

# `redact_extensions_keys` is default list of keys to redact from error extensions when `all: false` is set.
include_subgraph_errors: 
  all: false # Will redact error message and extensions keys from redact_extensions_keys.
  redact_extensions_keys: # if this is not defined, it'll just propagate all errors to client.
    - "service"
    - "exception"

include_subgraph_errors: 
  all: true # Will NOT redact error message and extensions even tho the keys are specified below.
  redact_extensions_keys:
    - "service"
    - "exception"
  subgraphs:
    product: false # will redact error message and extensions keys based on default config. 
    reviews:
      - "service" # Never redact keys listed here, but error message and other extension keys from default config will be redacted.

include_subgraph_errors:
  all: false # Will redact error message and extensions.
  redact_extensions_keys:
    - "service"
    - "exception"
  subgraphs:
    product: true # Will NOT redact anything.
    reviews:
      - "service" # Never redact keys listed here, but error message and other extension keys from default config will be redacted.

 ```

Reasons we need this:
1. Some subgraphs could want to pass non-vulnerable data to client via error extensions.
2. Some clients depend on certain data in error extensions to render UI. Although they shouldn't do it, migrating away from it would take time, especially from natives. While waiting their migration, it allows the original error message and selected keys to be redacted or propagated per subgraph.

There is no issue number yet. 

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/plugins/include_subgraph_errors.rs
+++ b/apollo-router/src/plugins/include_subgraph_errors.rs
@@ -20,11 +20,23 @@ register_plugin!("apollo", "include_subgraph_errors", IncludeSubgraphErrors);
 #[derive(Clone, Debug, JsonSchema, Default, Deserialize)]
 #[serde(rename_all = "snake_case", deny_unknown_fields, default)]
 struct Config {
-    /// Include errors from all subgraphs
+    /// Redact all subgraphs errors by default
     all: bool,
 
-    /// Include errors from specific subgraphs
-    subgraphs: HashMap<String, bool>,
+    /// Default keys to remove from error extensions for all subgraphs.
+    redact_extensions_keys: Vec<String>,
+
+    /// Override default configuration for specific subgraphs
+    subgraphs: HashMap<String, SubgraphConfig>,
+}
+
+#[derive(Clone, Debug, JsonSchema, Deserialize)]
+#[serde(deny_unknown_fields, untagged)]
+enum SubgraphConfig {
+    /// Enable or disable redaction for a subgraph
+    Included(bool),
+    /// Always reveal specific error extensions keys from a subgraph
+    PropagateExtensionsKeys(Vec<String>),
 }
 
 struct IncludeSubgraphErrors {
@@ -42,8 +54,14 @@ impl Plugin for IncludeSubgraphErrors {
     }
 
     fn subgraph_service(&self, name: &str, service: subgraph::BoxService) -> subgraph::BoxService {
-        // Search for subgraph in our configured subgraph map. If we can't find it, use the "all" value
-        let include_subgraph_errors = *self.config.subgraphs.get(name).unwrap_or(&self.config.all);
+        let subgraph_config = self.config.subgraphs.get(name).cloned();
+        let redact_extensions_keys = self.config.redact_extensions_keys.clone();
+
+        let include_subgraph_errors = match subgraph_config {
+            Some(SubgraphConfig::Included(enabled)) => enabled,
+            Some(SubgraphConfig::PropagateExtensionsKeys(_)) => false,
+            None => self.config.all,
+        };
 
         let sub_name_response = name.to_string();
         let sub_name_error = name.to_string();
@@ -58,11 +76,40 @@ impl Plugin for IncludeSubgraphErrors {
                                 .entry("service")
                                 .or_insert(sub_name_response.clone().into());
                         }
+                        return response;
                     } else {
                         tracing::info!("redacted subgraph({sub_name_response}) errors");
-                        for error in errors.iter_mut() {
+
+                        for error in response.response.body_mut().errors.iter_mut() {
+                            let mut new_extensions = error.extensions.clone();
+                            match subgraph_config.clone() {
+                                Some(SubgraphConfig::PropagateExtensionsKeys(keys)) => {
+                                    // Remove default keys, but show subgraph specific keys
+                                    for key in &redact_extensions_keys {
+                                        if !keys.contains(key) {
+                                            new_extensions.remove(key.as_str());
+                                        }
+                                    }
+                                }
+                                Some(SubgraphConfig::Included(enabled)) if enabled => {
+                                    // Do not redact any keys
+                                    for key in &redact_extensions_keys {
+                                        new_extensions.remove(key.as_str());
+                                    }
+                                }
+                                _ => {
+                                    if redact_extensions_keys.is_empty() {
+                                        new_extensions = Object::default();
+                                    } else {
+                                        for key in &redact_extensions_keys {
+                                            new_extensions.remove(key.as_str());
+                                        }
+                                    }
+                                }
+                            }
+
                             error.message = REDACTED_ERROR_MESSAGE.to_string();
-                            error.extensions = Object::default();
+                            error.extensions = new_extensions;
                         }
                     }
                 }

### CODE REVIEW

File: apollo-router/src/plugins/include_subgraph_errors.rs [hunk 1]
```diff
@@ -58,11 +76,40 @@ impl Plugin for IncludeSubgraphErrors {
...
                                 .or_insert(sub_name_response.clone().into());
                         }
+                        return response;
                     } else {
                         tracing::info!("redacted subgraph({sub_name_response}) errors");
-                        for error in errors.iter_mut() {
+
+                        for error in response.response.body_mut().errors.iter_mut() {
+                            let mut new_extensions = error.extensions.clone();
+                            match subgraph_config.clone() {
```
bnjjj, line 85: I don't think we need to clone it.
```suggestion
                            match &subgraph_config {
```

bnjjj (changes requested):
For configuration I think we could find a more explicit configuration and I think @BrynCooke will have better ideas than me about configuration

IvanGoncharov (changes requested):
I think it should be the opposite `include_extensions_keys`, where by default all extension keys are exposed but if you specify some values only those values are allowed.
Otherwise, we are in a situation where each new extension is added to errors by the router forcing everyone to update their config to prevent information leaking.

The same goes for subgraphs. 
If a subgraph adds a new extension (this can happen with a GraphQL library/framework update), it immediately leaks until someone updates the config.

IvanGoncharov:
> I think it should be the opposite of `include_extensions_keys`, which exposes all extension keys by default. However, if you specify some extension names, only those extensions are allowed. Otherwise, we are in a situation where each new extension adds to the errors caused by the router, forcing everyone to update their config to prevent information from leaking.
> 
> The same goes for subgraphs. If a subgraph adds a new extension (this can happen with a GraphQL library/framework update), it immediately leaks until someone updates the config.
> 
> Update: This discussion is a good example of why we need to be able to add new extensions in the router [#6359 (comment)](https://github.com/apollographql/router/pull/6359#issuecomment-2535766180)

I think the ideal solution would be to have `include_extensions_keys` support both the area of names and boolean values, where `true` (the default is 1.x) allows all extensions and `false` (the default in `2.x`) removes all extensions.
This would be the safest option, in my opinion, and allow us to improve the debuggability of the router (by adding extensions) without being worried that we are leaking private information.

@bnjjj What do you think?

bnjjj:
I agree @IvanGoncharov

abernix:
@Samjin I like where this is going but I do agree with the suggestions above from @IvanGoncharov and I think they would cater to important use cases from other users.  Do you think you'd be able to introduce those changes being suggesting here?

In terms of release scheduling, if we can get this all fixed and to a state that we're all mutually happy with, and get that done by middle of January, then this should be able to make it into our v1.60.0 release which ships at the end of January.  To call that done, we'd be looking at:

* Getting the changes in place.
* Getting tests written (the team can help advise here if needed.)
* Changelog (we can help polish this, but a starting point is lovely) (see [How to create a Changelog entry](https://github.com/apollographql/router/blob/dev/.changesets/README.md#how-to-create-a-changelog-entry)

Could that work?

abernix:
@BrynCooke Tagging you on this too.  Could you take a look?

BrynCooke:
I spent a bit of time thinking about this:

```

include_subgraph_errors:
  all: true # Propagate all errors and all extension keys (current)
  
include_subgraph_errors:
  all:
    allow_extensions_keys: # Propagate all errors and only the specified extension keys
      - "service"
      - "exception" 
    
include_subgraph_errors:
  subgraphs:
    products: true # Propagate all errors and all extension keys from products subgraph (current)
      
include_subgraph_errors:
  subgraphs:
    products:
      allow_extensions_keys: # Propagate all errors and only the specified extension keys
        - "service"
        - "exception"
        
```
We would use an enum to allow true or a nested allow_extension_keys.
```
enum ErrorMode {
  Global(bool)  // true|false to maintain backward compatibility
  Allow {
    allow_extension_keys: Vec<String> // Allow keys but deny all others
  }
}
```
The advantage of this config over what is there in the PR is that it doesn't have rules that the user has to know. e.g. if something it set it ignores this other setting.

@Samjin Would this be config suit your needs? If not can you let us know why a deny list would be better for you?
The above config structure does not preclude us from adding deny support.

Samjin:
> @Samjin Would this be config suit your needs? If not can you let us know why a deny list would be better for you?
The above config structure does not preclude us from adding deny support.

I think the proposal is fine. I'll try to implement it this/next week. Thanks.

Samjin:
Is `path:[xx, xx]` intentionally not redacted from errors?

Samjin:
1. Is `path: [xx, xx]` intentionally ignored? 
image


2. Should subgraph `allow_extensions_keys` propagated together with existing global keys? This way subgraphs don't need to repeatedly define same keys. However, the downside of that is subgraph can't override a global key. For example, it's impossible for A to exclude `service` from propagation. 
```
include_subgraph_errors:
  all:
    allow_extensions_keys:
      - "service"
      - "exception"
  subgraphs:
    A:
      allow_extensions_keys:
        - "code"
    B: false
```

BrynCooke:
> 1. Is `path: [xx, xx]` intentionally ignored?
> 
> image

I think this is a bug. Everything should be redacted. But I'll ask the team to be sure.

> 2. I propose propagating the `allow_extensions_keys` from the global configuration to subgraphs. This way, subgraphs don't have to repeatedly define the same keys. Additionally, subgraphs should have the ability to override global keys.
> 
> ```yaml
> include_subgraph_errors:
>   all:
>     allow_extensions_keys:
>       - "service"
>       - "exception"
>   subgraphs:
>     A:
>       allow_extensions_keys: # Additional allow keys
>         - "code"
>       deny_extensions_keys: # Override a global allow key
>         - "exception"
>     B: false
> ```
> 

The general rule should be that anything in global is the default, but setting at the subgraph level will override. So Agree.

> 3. There could several operational challenges for allowlist approach:
> 
> * All subgraphs must propagate errors to begin with, because gateway owner cannot determine which extension keys are already dependent by client(this is not even including data used in client metrics or logging). However, in our case, we already know specific keys that must be redacted globally for security or compliance reasons.
> * it's error prone as it requires subgraphs to work closely with clients(including external clients) to accurately identify dependent keys to avoid breaking clients. The overall process can be time consuming as it needs effort from many teams.
> 
> To mitigate the challenges above, we could try adding a deny list. deny or allow list in global config can be mutual exclusive.
> 
> ```yaml
> include_subgraph_errors:
>   all:
>     deny_extensions_keys: # Propagate all errors except specified keys
>       - "sensitive_key"
>     redact_message: true # To redact message
>   subgraphs:
>     A:
>       deny_extensions_keys: # Additional deny keys
>         - "stack"
>       allow_extensions_keys: # Override a global deny key
>         - "sensitive_key"
>     B: false
> ```

I'm not opposed to having an allow and deny list, but we'd need to have extensive tests for all the permutations.

Samjin:
Closing this in favor of https://github.com/apollographql/router/pull/6862
Please take a look when you got chance. Thanks

PR #6406, opened 2024-12-05
Distributed query plan cache keys include the Router version number

SimonSapin:
If you have enabled [Distributed query plan caching](https://www.apollographql.com/docs/router/configuration/distributed-caching/#distributed-query-plan-caching), starting with this release and going forward you should anticipate additional cache regeneration cost when updating between any Router versions.

More often than not, an Apollo Router release may contain changes that affect what query plans are generated or how they’re represented. To avoid using outdated entries from distributed cache, the cache key includes a counter that was manually incremented with relevant data structure or algorithm changes. Instead the cache key now includes the Router version number, so that different versions will always use separate cache entries.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/.changesets/config_simon_router_version_in_cache_key.md
+++ b/.changesets/config_simon_router_version_in_cache_key.md
@@ -1,7 +1,7 @@
 ### Distributed query plan cache keys include the Router version number ([PR #6406](https://github.com/apollographql/router/pull/6406))
 
-If you have enabled [Distributed query plan caching](https://www.apollographql.com/docs/router/configuration/distributed-caching/#distributed-query-plan-caching), starting with this release and going forward you should anticipate additional cache regeneration cost when updating between any Router versions.
-
 More often than not, an Apollo Router release may contain changes that affect what query plans are generated or how they’re represented. To avoid using outdated entries from distributed cache, the cache key includes a counter that was manually incremented with relevant data structure or algorithm changes. Instead the cache key now includes the Router version number, so that different versions will always use separate cache entries.
+
+If you have enabled [Distributed query plan caching](https://www.apollographql.com/docs/router/configuration/distributed-caching/#distributed-query-plan-caching), starting with this release and going forward you should anticipate additional cache regeneration cost when updating between any Router versions.
 
 By [@SimonSapin](https://github.com/SimonSapin) in https://github.com/apollographql/router/pull/6406
--- a/.changesets/config_simon_router_version_in_cache_key.md
+++ b/.changesets/config_simon_router_version_in_cache_key.md
@@ -1,0 +1,7 @@
+### Distributed query plan cache keys include the Router version number ([PR #6406](https://github.com/apollographql/router/pull/6406))
+
+If you have enabled [Distributed query plan caching](https://www.apollographql.com/docs/router/configuration/distributed-caching/#distributed-query-plan-caching), starting with this release and going forward you should anticipate additional cache regeneration cost when updating between any Router versions.
+
+More often than not, an Apollo Router release may contain changes that affect what query plans are generated or how they’re represented. To avoid using outdated entries from distributed cache, the cache key includes a counter that was manually incremented with relevant data structure or algorithm changes. Instead the cache key now includes the Router version number, so that different versions will always use separate cache entries.
+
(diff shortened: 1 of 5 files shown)


.changesets/config_simon_router_version_in_cache_key.md
```diff
@@ -0,0 +1,7 @@
+### Distributed query plan cache keys include the Router version number ([PR #6406](https://github.com/apollographql/router/pull/6406))
```
SimonSapin, line 1: I put this (though filename prefix) in the "Configuration" section of the changelog, but none of the options in `xtask changeset create` seemed like a great fit. 

`CHANGELOG.md` has some `> [!IMPORTANT]` sections, is that done manually when editing for a release?

.changesets/config_simon_router_version_in_cache_key.md
```diff
@@ -0,0 +1,7 @@
...
+
+If you have enabled [Distributed query plan caching](https://www.apollographql.com/docs/router/configuration/distributed-caching/#distributed-query-plan-caching), starting with this release and going forward you should anticipate additional cache regeneration cost when updating between any Router versions.
```
SimonSapin, line 3: Should this discuss the possibility a staged rollout to start warming up new cache entries while the previous version still serves most traffic? I didn’t want to sound like "draw the rest of the owl"

.changesets/config_simon_router_version_in_cache_key.md
```diff
@@ -0,0 +1,7 @@
...
+
+More often than not, an Apollo Router release may contain changes that affect what query plans are generated or how they’re represented. To avoid using outdated entries from distributed cache, the cache key includes a counter that was manually incremented with relevant data structure or algorithm changes. Instead the cache key now includes the Router version number, so that different versions will always use separate cache entries.
```
SimonSapin, line 5: I think this doesn’t need to mention `CACHE_KEY_VERSION` v.s. `FEDERATION_VERSION` so I tried to pick a relevant level of detail.

.changesets/config_simon_router_version_in_cache_key.md
lrlna: we do that manually in the release PR, since it has to go in the top of the changelog

.changesets/config_simon_router_version_in_cache_key.md
lrlna: I think "drawing the rest of the owl" can be added to distributed query plan caching docs themselves. It will be a nice addition, and people can search the docs with a bit more ease than a changelog.

.changesets/config_simon_router_version_in_cache_key.md
lrlna: I would put this paragraph first. The point is that almost every release has had a cache key change, which comes with the additional cache regeneration cost. If people read the first paragraph, they'll think this is _more_ than what's currently happening, but it effectively isn't.

apollo-router/tests/integration/redis.rs
```diff
@@ -51,7 +51,10 @@ async fn query_planner_cache() -> Result<(), BoxError> {
     }
     // If this test fails and the cache key format changed you'll need to update the key here.
     // Look at the top of the file for instructions on getting the new cache key.
-    let known_cache_key = "plan:cache:1:federation:v2.9.3:8c0b4bfb4630635c2b5748c260d686ddb301d164e5818c63d6d9d77e13631676:opname:3973e022e93220f9212c18d0d0c543ae7c309e46640da93a4a0314de999f5112:metadata:8f8ce6ad09f15c3d567a05f1c3d7230ab71b3366fcaebc9cc3bbfa356d55ac12";
+    let known_cache_key = &format!(
+        "plan:router:{}:8c0b4bfb4630635c2b5748c260d686ddb301d164e5818c63d6d9d77e13631676:opname:3973e022e93220f9212c18d0d0c543ae7c309e46640da93a4a0314de999f5112:metadata:8f8ce6ad09f15c3d567a05f1c3d7230ab71b3366fcaebc9cc3bbfa356d55ac12",
```
lrlna, line 55: appreciate the `format!` with env var so we don't have to update these tests with every release 🙏

PR #6420, opened 2024-12-09
Improve error messaging when QP encounters unmergeable fields

sachindshinde:
This PR updates the error messaging when query planner encounters unmergeable fields (indicating a known bug in query planner), and increments an internal metric when this error kind is encountered. Note that query planner does not check all such cases of this, as doing so is computationally expensive. The bug itself is not fixed in this PR, as it requires substantial effort to fix.

### DIFF
--- a/apollo-federation/src/operation/merging.rs
+++ b/apollo-federation/src/operation/merging.rs
@@ -47,14 +47,14 @@ impl<'a> FieldSelectionValue<'a> {
                 return Err(SingleFederationError::InternalUnmergeableFields {
                     message: format!(
                         "Cannot merge field selection for field \"{}\" into a field selection for \
-                        field \"{}\". This is a known query planning bug that was silently ignored \
-                        in the old  Javascript query planner. The new Rust-native query planner \
-                        does not fix this bug at this time, but in some cases does catch when this \
-                        bug occurs. If you're seeing this message, this bug was likely triggered \
-                        by one of the field selections mentioned previously having an alias that \
-                        was the same name as the field in the other field selection. The \
-                        recommended workaround is to change this alias to a different one in your \
-                        operation.",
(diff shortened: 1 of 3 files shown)


apollo-federation/src/operation/merging.rs
```diff
@@ -42,12 +43,23 @@ impl<'a> FieldSelectionValue<'a> {
...
-                other_field.field_position,
-                self_field.field_position,
-            );
+            if other_field.field_position != self_field.field_position {
+                return Err(SingleFederationError::InternalUnmergeableFields {
+                    message: format!(
+                        "Cannot merge field selection for field \"{}\" into a field selection for \
+                        field \"{}\". This is a known query planning bug that was silently ignored \
+                        in the old  Javascript query planner. The new Rust-native query planner \
+                        does not fix this bug at this time, but in some cases does catch when this \
```
TylerBloom, line 52: A minor rephrasing:
```suggestion
                        field \"{}\". This is a known query planning bug in the old, Javascript query \
                        planner that was silently ignored. The Rust-native query planner does not \
                        fix fully address this bug at this time, but in some cases does catch when this \
```

TylerBloom (approved):

apollo-federation/src/operation/merging.rs
sachindshinde: The first sentence change sounds fine (I don't think we need the comma between "old" and "Javascript" here though since they're acting as cumulative adjectives instead of coordinate adjectives). For the second sentence, I'm hesitant to replace "does not fix" with "does not fully address", since that implies we partially address it (which we don't). I'm fine with changing it to "does not address" though.

TylerBloom: That works for me

PR #6421, opened 2024-12-09
Improve backwards compatibility of `FederationError`s

TylerBloom:
This PR makes a few changes to how errors that originate in the query planner are served to our users. This largely affects two categories of errors.

For errors that originate from `Import::from_value`, the AST value is now included in the error message.

For `GetOperationError`s (originating from the first line of `build_query_plan`), we need to ensure that their status code is `GRAPHQL_VALIDATION_ERROR` rather than `INTERNAL_SERVER_ERROR` (which all errors from the query planner were being marked as). On the router's side, `FederationError`s no longer get turned into strings. The error is checked to see if it is one of the notable variants. If so, its variant is noted so it can be turned into a GraphQL error. All other variants are still turned into a string. Note that this is a temporary solution. For router 2.0, these errors are likely going to be `INTERNAL_SERVER_ERROR`s, but the Rust QP needs to rework it error handling and the new system will look similar to this (internal and external error types).

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-federation/src/link/argument.rs
```diff
@@ -100,22 +100,6 @@ pub(crate) fn directive_optional_boolean_argument(
...
-        SingleFederationError::Internal {
-            message: format!(
-                "Required argument \"{}\" of directive \"@{}\" was not present.",
-                name, application.name
-            ),
-        }
-        .into()
-    })
-}
-
```
TylerBloom, line 118: This was not part of the ticket, but I stumbled across this dead code that was not clean up during the `@context` PR, so I removed it.

apollo-federation/src/query_plan/query_planner.rs
```diff
@@ -325,14 +325,18 @@ impl QueryPlanner {
...
+            .map_err(|_| {
+                if operation_name.is_some() {
+                    SingleFederationError::UnknownOperation
+                } else {
+                    SingleFederationError::OperationNameNotProvided
+                }
+            })?;
         if operation.selection_set.is_empty() {
             // This should never happen because `operation` comes from a known-valid document.
             // We could panic here but we are returning a `Result` already anyways, so shrug!
```
goto-bus-stop, line 338:
```suggestion
```

Let's remove this line (`bail!` technically doesn't panic, but it should be used as if it does)

goto-bus-stop (approved):
LGTM, for reference tests validating the behaviour already existed at https://github.com/apollographql/router/blob/5f72471d05deb6379610d1f661b2d84b4c65e10a/apollo-router/tests/integration/operation_name.rs#L121-L172

PR #6422, opened 2024-12-09
prep release: v2.0.0-preview.3

pubmodmatt:
Prep for Connectors release 2.0.0-preview.3.

In order to update the license file, this requires upgrading to `cargo-about` version 0.6.6 locally to pick up [this fix](https://github.com/EmbarkStudios/cargo-about/releases/tag/0.6.6). However, that version of `cargo-about` cannot be installed with Rust 1.76, which the router has pinned, so the circleci `config.yml` has not been updated to this version.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/helm/chart/router/README.md
+++ b/helm/chart/router/README.md
@@ -2,7 +2,7 @@
 
 [router](https://github.com/apollographql/router) Rust Graph Routing runtime for Apollo Federation
 
-![Version: 2.0.0-preview.1](https://img.shields.io/badge/Version-2.0.0--preview.1-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: v2.0.0-preview.1](https://img.shields.io/badge/AppVersion-v2.0.0--preview.1-informational?style=flat-square)
+![Version: 2.0.0-preview.3](https://img.shields.io/badge/Version-2.0.0--preview.3-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: v2.0.0-preview.3](https://img.shields.io/badge/AppVersion-v2.0.0--preview.3-informational?style=flat-square)
 
 ## Prerequisites
 
@@ -11,7 +11,7 @@
 ## Get Repo Info
 
 ```console
-helm pull oci://ghcr.io/apollographql/helm-charts/router --version 2.0.0-preview.1
+helm pull oci://ghcr.io/apollographql/helm-charts/router --version 2.0.0-preview.3
 ```
 
 ## Install Chart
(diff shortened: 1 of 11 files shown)


helm/chart/router/README.md
```diff
@@ -81,7 +81,7 @@ helm show values oci://ghcr.io/apollographql/helm-charts/router
 | resources | object | `{}` |  |
 | restartPolicy | string | `"Always"` | Sets the restart policy of pods |
 | rollingUpdate | object | `{}` | Sets the [rolling update strategy parameters](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#rolling-update-deployment).  Can take absolute values or % values. |
-| router | object | `{"args":["--hot-reload"],"configuration":{"health_check":{"listen":"0.0.0.0:8088"},"supergraph":{"listen":"0.0.0.0:4000"}}}` | See https://www.apollographql.com/docs/router/configuration/overview/#yaml-config-file for yaml structure |
+| router | object | `{"args":["--hot-reload"],"configuration":{"health_check":{"listen":"0.0.0.0:8088"},"supergraph":{"listen":"0.0.0.0:4000"}}}` | See https://www.apollographql.com/docs/graphos/reference/router/configuration#yaml-config-file for yaml structure |
```
nicholascioli, line 84: There's an auto-redirect for the old URL, but do we know if there are any other links that need to be updated for this doc change?

(Non-blocking for this PR, just mostly a note for future us)

licenses.html
```diff
@@ -2989,7 +3030,7 @@ <h4>Used by:</h4>
       same &quot;printed page&quot; as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright 2023 The OpenTelemetry Authors
+   Copyright [yyyy] [name of copyright owner]
```
nicholascioli, line 3033: Is this related to the needed update to `cargo-about` that we can't apply due to the pinned version of rust?

nicholascioli (approved):
LGTM! I left a few notes for possible later reference.

licenses.html
pubmodmatt: Ugh, maybe. Looking into it.

helm/chart/router/README.md
pubmodmatt: Not sure how this was updated, it must be something automatic.

licenses.html
pubmodmatt: @lennyburdette pointed out that this same string appears [on main](https://github.com/apollographql/router/blob/main/licenses.html) as well.

pubmodmatt: We believe this was just the non-deterministic ordering in the file.

PR #6423, opened 2024-12-10
fix cargo about

bnjjj:
cargo about and cargo deny have separate configuration files: about.toml and deny.toml.

deny.toml was updated to allow Unicode-3.0 license in #6417 , but about.toml was not updated. This is causing `cargo xtask licenses` to fail.

This PR fixes about.toml.


garypen (approved):

garypen (approved):
This has passed on nightly and looks good.

PR #6429, opened 2024-12-10
Improve docs about enabling invalidation for entity caching

bnjjj:


abernix:
@bnjjj — is this on your radar still?

docs/source/routing/performance/caching/entity.mdx
```diff
@@ -295,6 +295,7 @@ preview_entity_cache:
       redis:
         urls: ["redis://..."]
       invalidation:
+	enabled: true
```
goto-bus-stop:
```suggestion
      invalidation:
        enabled: true
```
correct spacing

PR #6431, opened 2024-12-10
fix(semantic-diff): fixed false mismatches with type conditioned fetching enabled

duckki:
Two issues:

1. JS QP may have empty key roots in response path data, which is ignored by the router. On the other hand, Rust QP does not use such a placeholder. So, the semantic diff ignores the placeholder during comparison.
2. JS QP may have empty condition list, but it was deserialized incorrectly (for example, `k|[]` would be interpreted as a whole key name). This is now fixed in the router's deserializer.

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [x] Manual Tests

### DIFF
--- a/apollo-router/src/json_ext.rs
+++ b/apollo-router/src/json_ext.rs
@@ -37,6 +37,15 @@ fn extract_matched_conditions(caps: &Captures) -> TypeConditions {
         .unwrap_or_default()
 }
 
+fn split_key_and_type_conditions(s: &str) -> (String, Option<TypeConditions>) {
+    let mut type_conditions = None;
+    let key = TYPE_CONDITIONS_REGEX.replace(s, |caps: &Captures| {
+        type_conditions = Some(extract_matched_conditions(caps));
+        ""
+    });
+    (key.to_string(), type_conditions)
+}
+
 macro_rules! extract_key_value_from_object {
     ($object:expr, $key:literal, $pattern:pat => $var:ident) => {{
         match $object.remove($key) {
@@ -857,13 +866,8 @@ impl<'de> serde::de::Visitor<'de> for FlattenVisitor {
     where
         E: serde::de::Error,
     {
-        let mut type_conditions = Default::default();
-        let path = TYPE_CONDITIONS_REGEX.replace(s, |caps: &Captures| {
-            type_conditions = Some(extract_matched_conditions(caps));
-            ""
-        });
-
-        if path == "@" {
+        let (key, type_conditions) = split_key_and_type_conditions(s);
+        if key == "@" {
             Ok(type_conditions)
         } else {
             Err(serde::de::Error::invalid_value(
@@ -913,12 +917,7 @@ impl<'de> serde::de::Visitor<'de> for KeyVisitor {
     where
         E: serde::de::Error,
     {
-        let mut type_conditions = Default::default();
-        let key = TYPE_CONDITIONS_REGEX.replace(s, |caps: &Captures| {
-            type_conditions = Some(extract_matched_conditions(caps));
-            ""
-        });
-        Ok((key.to_string(), type_conditions))
+        Ok(split_key_and_type_conditions(s))
     }
 }
 
@@ -973,26 +972,16 @@ where
 }
 
 fn flatten_from_str(s: &str) -> Result<PathElement, String> {
-    let mut type_conditions = Default::default();
-    let path = TYPE_CONDITIONS_REGEX.replace(s, |caps: &Captures| {
-        type_conditions = Some(extract_matched_conditions(caps));
-        ""
-    });
-
-    if path != "@" {
+    let (key, type_conditions) = split_key_and_type_conditions(s);
+    if key != "@" {
         return Err("invalid flatten".to_string());
     }
     Ok(PathElement::Flatten(type_conditions))
 }
 
 fn key_from_str(s: &str) -> Result<PathElement, String> {
-    let mut type_conditions = Default::default();
-    let key = TYPE_CONDITIONS_REGEX.replace(s, |caps: &Captures| {
-        type_conditions = Some(extract_matched_conditions(caps));
-        ""
-    });
-
-    Ok(PathElement::Key(key.to_string(), type_conditions))
+    let (key, type_conditions) = split_key_and_type_conditions(s);
+    Ok(PathElement::Key(key, type_conditions))
 }
 
 /// A path into the result document.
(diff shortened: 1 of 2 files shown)


apollo-router/src/json_ext.rs
```diff
@@ -26,10 +26,20 @@ pub(crate) type Object = Map<ByteString, Value>;
 const FRAGMENT_PREFIX: &str = "... on ";
 
 static TYPE_CONDITIONS_REGEX: Lazy<Regex> = Lazy::new(|| {
-    Regex::new(r"(?:\|\[)(?<condition>.+?)(?:,\s*|)(?:\])")
+    Regex::new(r"(?:\|\[(?<condition>.+)?\])")
```
sachindshinde, line 29: You probably still want the `.+` to be lazy instead of greedy, so that you capture the first `]` instead of the last one. (This is really just to accommodate future potential syntax changes, and isn't strictly necessary now for correctness.)
```suggestion
    Regex::new(r"(?:\|\[(?<condition>.+?)?\])")
```

apollo-router/src/json_ext.rs
```diff
@@ -852,16 +862,7 @@ impl<'de> serde::de::Visitor<'de> for FlattenVisitor {
...
-                caps.name("condition")
-                    .map(|c| {
-                        c.as_str()
-                            .split(',')
-                            .map(|s| s.to_string())
-                            .collect::<Vec<_>>()
-                    })
-                    .unwrap_or_default(),
-            );
+            type_conditions.extend(extract_matched_conditions(caps).unwrap_or_default());
```
sachindshinde, line 865: Note that `foo|[]` is not equivalent to `foo` here (or rather, empty type conditions are not the same as no type conditions). Previously we were effectively disallowing empty type conditions entirely via the regex. If want to allow them, we should be returning `Some(vec![])` rather `None`.

apollo-router/src/json_ext.rs
```diff
@@ -921,13 +922,7 @@ impl<'de> serde::de::Visitor<'de> for KeyVisitor {
...
         let mut type_conditions = Vec::new();
         let key = TYPE_CONDITIONS_REGEX.replace(s, |caps: &Captures| {
-            type_conditions.extend(
-                caps.extract::<1>()
-                    .1
-                    .map(|s| s.split(',').map(|s| s.to_string()))
-                    .into_iter()
-                    .flatten(),
-            );
+            type_conditions.extend(extract_matched_conditions(caps).unwrap_or_default());
```
sachindshinde, line 925: Similar comment as above.

apollo-router/src/json_ext.rs
```diff
@@ -994,13 +989,7 @@ where
...
     let mut type_conditions = Vec::new();
     let path = TYPE_CONDITIONS_REGEX.replace(s, |caps: &Captures| {
-        type_conditions.extend(
-            caps.extract::<1>()
-                .1
-                .map(|s| s.split(',').map(|s| s.to_string()))
-                .into_iter()
-                .flatten(),
-        );
+        type_conditions.extend(extract_matched_conditions(caps).unwrap_or_default());
```
sachindshinde, line 992: Similar comment as above.

apollo-router/src/json_ext.rs
```diff
@@ -1015,13 +1004,7 @@ fn flatten_from_str(s: &str) -> Result<PathElement, String> {
...
     let mut type_conditions = Vec::new();
     let key = TYPE_CONDITIONS_REGEX.replace(s, |caps: &Captures| {
-        type_conditions.extend(
-            caps.extract::<1>()
-                .1
-                .map(|s| s.split(',').map(|s| s.to_string()))
-                .into_iter()
-                .flatten(),
-        );
+        type_conditions.extend(extract_matched_conditions(caps).unwrap_or_default());
```
sachindshinde, line 1007: Similar comment as above.

sachindshinde (changes requested):
A few minor changes to `json_ext.rs` code, the diffing changes look fine though.

One thing to note is that empty type conditions is probably a bug (which may explain why that regex didn't trigger any red flags till now). If we believe there's a legitimate case for it, we should passing the empty vector along instead of converting it to "no type conditions"/`None`.

apollo-router/src/json_ext.rs
```diff
@@ -26,7 +26,7 @@ pub(crate) type Object = Map<ByteString, Value>;
 const FRAGMENT_PREFIX: &str = "... on ";
 
 static TYPE_CONDITIONS_REGEX: Lazy<Regex> = Lazy::new(|| {
-    Regex::new(r"(?:\|\[(?<condition>.+)?\])")
+    Regex::new(r"\|\[(?<condition>.*?)\]")
```
duckki, line 29: This will match `|[]` as a list with one item of empty string.
I think `"(?:\|\[(?<condition>.+?)?\])` is works better, but None in `extract_matched_conditions) should be interpreted as an empty vec.

sachindshinde: I was about to handle the empty string in `extract_matched_conditions()`, but I think your suggestion here is cleaner, pushed up the change in the [fc894e1](https://github.com/apollographql/router/pull/6431/commits/fc894e1c682449cad4b1e1adf69c0da31a843311). I've also updated `extract_matched_conditions()` to not return an `Option`.

apollo-router/src/json_ext.rs
```diff
@@ -842,31 +858,17 @@ impl<'de> serde::de::Visitor<'de> for FlattenVisitor {
     fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
         write!(
             formatter,
-            "a string that is '@', potentially preceded of followed by type conditions"
+            "a string that is '@', potentially followed by type conditions"
```
sachindshinde, line 861: Slight wording fix here (and similarly for other deserializers).

apollo-router/src/json_ext.rs
```diff
@@ -1119,9 +1072,7 @@ impl Path {
             PathElement::Key(key, type_conditions) => {
                 let mut tc = String::new();
                 if let Some(c) = type_conditions {
-                    if !c.is_empty() {
-                        tc = format!("|[{}]", c.join(","));
-                    }
+                    tc = format!("|[{}]", c.join(","));
```
sachindshinde, line 1075: When the type conditions are empty, we should serialize to `|[]` instead of empty string. (There are some similar cases below.)

sachindshinde (approved):
Pushed up some changes to address feedback I left earlier. I've also made a few other small changes, which I've left comments below to explain more about.

PR LGTM now, should be good to merge!

PR #6432, opened 2024-12-10
Update `ensure!` macro to return its error

sachindshinde:
In v1.58.0, the `internal_error!` macro was updated to no longer return its error for release builds, but the `ensure!` macro (which calls `internal_error!`) wasn't updated accordingly. This means that `ensure!` callsites in the current code are silently ignoring their errors if their check fails in release builds. This PR updates `ensure!` to use the `bail!` macro, which does return its error.


duckki (approved):
👍

PR #6434, opened 2024-12-11
Miscellaneous type-conditioned fetching fixes

sachindshinde:
This PR makes bugfixes to the Rust QP port of type-conditioned fetching. There's a few of these, so I'm going to leave comments on the code with relevant information.

### DIFF
--- a/apollo-federation/src/query_plan/fetch_dependency_graph.rs
+++ b/apollo-federation/src/query_plan/fetch_dependency_graph.rs
@@ -509,15 +509,14 @@ impl FetchDependencyGraphNodePath {
         type_conditioned_fetching_enabled: bool,
         root_type: CompositeTypeDefinitionPosition,
     ) -> Result<Self, FederationError> {
-        let root_possible_types = if type_conditioned_fetching_enabled {
+        let root_possible_types: IndexSet<Name> = if type_conditioned_fetching_enabled {
             schema.possible_runtime_types(root_type)?
         } else {
             Default::default()
         }
         .into_iter()
-        .map(|pos| Ok(pos.get(schema.schema())?.name.clone()))
-        .collect::<Result<IndexSet<Name>, _>>()
-        .map_err(|e: PositionLookupError| FederationError::from(e))?;
+        .map(|pos| Ok::<_, PositionLookupError>(pos.get(schema.schema())?.name.clone()))
+        .process_results(|c| c.sorted().collect())?;
 
         Ok(Self {
             schema,
@@ -577,12 +576,13 @@ impl FetchDependencyGraphNodePath {
                 None => self.possible_types.clone(),
                 Some(tcp) => {
                     let element_possible_types = self.schema.possible_runtime_types(tcp.clone())?;
-                    element_possible_types
+                    self.possible_types
                         .iter()
                         .filter(|&possible_type| {
-                            self.possible_types.contains(&possible_type.type_name)
+                            element_possible_types
+                                .contains(&ObjectTypeDefinitionPosition::new(possible_type.clone()))
                         })
-                        .map(|possible_type| possible_type.type_name.clone())
+                        .cloned()
                         .collect()
                 }
             },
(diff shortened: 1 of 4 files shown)


apollo-federation/src/query_plan/fetch_dependency_graph.rs
```diff
@@ -509,15 +509,14 @@ impl FetchDependencyGraphNodePath {
...
         root_type: CompositeTypeDefinitionPosition,
     ) -> Result<Self, FederationError> {
-        let root_possible_types = if type_conditioned_fetching_enabled {
+        let root_possible_types: IndexSet<Name> = if type_conditioned_fetching_enabled {
             schema.possible_runtime_types(root_type)?
         } else {
             Default::default()
         }
         .into_iter()
-        .map(|pos| Ok(pos.get(schema.schema())?.name.clone()))
```
sachindshinde, line 518: Bugs:
- This should be a sorted list, similar to the JS code. (I've also updated the code to use `process_results()`.)

apollo-federation/src/query_plan/fetch_dependency_graph.rs
```diff
@@ -577,12 +576,13 @@ impl FetchDependencyGraphNodePath {
                 None => self.possible_types.clone(),
                 Some(tcp) => {
                     let element_possible_types = self.schema.possible_runtime_types(tcp.clone())?;
-                    element_possible_types
```
sachindshinde, line 580: Bugs:
- In the JS code, we avoid having to re-sort because we remove elements from `this.possibleTypes` instead of from the type condition's possible runtime types. Otherwise, we have to sort. I've updated the Rust code to similar remove from `self.possible_types`, to avoid the sort.

apollo-federation/src/query_plan/fetch_dependency_graph.rs
```diff
@@ -592,15 +592,11 @@ impl FetchDependencyGraphNodePath {
     }
 
     fn advance_field_type(&self, element: &Field) -> Result<IndexSet<Name>, FederationError> {
-        if !element
```
sachindshinde, line 595: Bugs:
- We shouldn't be silently ignoring errors here when getting the output base type.

apollo-federation/src/query_plan/fetch_dependency_graph.rs
```diff
@@ -616,17 +612,13 @@ impl FetchDependencyGraphNodePath {
                     .schema
                     .possible_runtime_types(typ)?
                     .into_iter()
-                    .map(|ctdp| ctdp.type_name)
```
sachindshinde, line 619: Refactors:
- We don't need to collect into intermediate vectors here, we can just flatten an iterator of iterators.

apollo-federation/src/query_plan/mod.rs
```diff
@@ -249,8 +249,8 @@ pub struct FetchDataKeyRenamer {
 /// elements.
 #[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize)]
 pub enum FetchDataPathElement {
-    Key(Name, Conditions),
```
sachindshinde, line 252: Bugs:
- There's a difference between no type conditions and an empty list of type conditions. The former means no restriction is imposed on `__typename` at the relevant path location, while the latter means there is a restriction imposed on `__typename` at the relevant path location, but there are no values that could satisfy that restriction (i.e the path is "impossible").
  - This latter case should ideally be optimized out by the query planner so that query planner execution doesn't have to deal with it, but we've observed some cases where empty lists of type conditions are generated. Until they are all optimized out, we'll need to allow this in the query plan representation.
- There are various upstream and downstream effects to these changes. The gist is:
  - No type conditions is the new default (instead of empty type conditions).
  - Empty type conditions are serialized as `|[]`.
  - None type conditions are serialized as the empty string.
  - There's some pending changes in https://github.com/apollographql/router/pull/6431 that are also relevant.

apollo-federation/src/query_plan/fetch_dependency_graph.rs
duckki: That's great. It will keep semantic diff simple.

duckki (approved):
Looks good!

PR #6439, opened 2024-12-11
refactor: simplify __typename handling in connectors code

lennyburdette:
three scenarios:

1. connector on a root field: the entire object is handled by json_selection, which does __typename injection. if the return type is abstract, json_selection logic will handle it.
2. explicit entity connector: the entire object is handled by json_selection, which does __typename injection. if the entity type is abstract, the planner decides if it needs __typename (it won't if it's an interfaceObject), and json_selection logic will handle it.
3. field on an entity: the __typename for the entity type is handled outside of json_selection. but there's only two scenarios: it's an object type so we know the __typename and can inject it, or its an interface object and we shouldn't supply the __typename.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-federation/src/sources/connect/json_selection/selection_set.rs
+++ b/apollo-federation/src/sources/connect/json_selection/selection_set.rs
@@ -61,12 +61,10 @@ impl SubSelection {
 
         // When the operation contains __typename, it might be used to complete
         // an entity reference (e.g. `__typename id`) for a subsequent fetch.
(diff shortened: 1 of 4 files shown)


nicholascioli (approved):
LGTM!

PR #6443, opened 2024-12-12
Per-operation and client error metric

bonnici:
This adds a new metric that tracks errors per operation, client, and error code. The metric in incremented whenever an error occurs and includes the following attributes:

* client name
* client version
* apollo operation ID
* error code
* operation name
* operation type (query/mutation/subscription)

This new metric is marked as experimental and disabled by default, since using it in a high-cardinality graph will likely mean that we'll hit OTel cardinality issues.

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/services/router/service.rs
```diff
@@ -778,6 +798,31 @@ impl RouterService {
...
+            let temp_path = match &error.path {
+                None => "".into(),
+                Some(path) => path.to_string()
+            };
+            println!("temp_code: {temp_code}");
+            println!("temp_path: {temp_path}");
+            println!("temp_service: {temp_service}");
+
+            u64_counter!(
+                "temp.nick.apollo.router.operations.error",
```
timbotnik, line 814: 😂  Cool to see a PoC in under 50 lines.

apollo-router/src/services/router/service.rs
```diff
@@ -812,12 +818,47 @@ impl RouterService {
         Ok(graphql_requests)
     }
 
-    fn count_errors(errors: &[graphql::Error]) {
+    fn count_errors(&self, errors: &[graphql::Error], context: &Context) {
+        let codes = errors
+            .iter()
+            .map(|e| e.extensions.get("code").and_then(|c| c.as_str()))
```
timbotnik, line 824: IIUC for errors without a code, it would count against the "empty" code e.g. `code=""`.  I think we could consider defining an explicit `UNKNOWN` code but it seems fine to support this with an empty code.

apollo-router/src/services/router/service.rs
```diff
@@ -812,12 +818,47 @@ impl RouterService {
...
+            .collect();
+        self.count_error_codes(codes, context);
```
timbotnik, line 826: Noting that this implementation will have to change if we add other dimensions here e.g. subgraph, coordinate, path - there could be some errors with the same codes from subgraph A vs. subgraph B, etc.

apollo-router/src/services/router/service.rs
```diff
@@ -812,12 +818,47 @@ impl RouterService {
...
-        for error in errors {
-            let code = error.extensions.get("code").and_then(|c| c.as_str());
+        for code in &codes {
             let entry = map.entry(code).or_insert(0u64);
             *entry += 1;
+
+            if matches!(self.oltp_error_metrics_mode, OtlpErrorMetricsMode::Enabled) {
+                let code_str = code.unwrap_or_default().to_string();
+                u64_counter!(
+                    "apollo.router.operations.error",
```
timbotnik, line 851: I know that the Router team has asked for this to be a private metric - would that mean we would need to call it `apollo_private.router.operations.error`?  Ultimately we don't want it to be, so I'm fine with it not being private at all.

apollo-router/src/services/router/service.rs
bonnici: Yeah this is an empty string. I can easily add a fallback code but I think an empty string is just as good.

apollo-router/src/services/router/service.rs
bonnici: This should pass the "private" filter regex and not pass the "public" filter regex, so it should be private. I don't really know how to test that thought because it is still coming through in prometheus.

```
    pub(crate) fn private<T: Into<MeterProvider>>(delegate: T) -> Self {
        FilterMeterProvider::builder()
            .delegate(delegate)
            .allow(
                Regex::new(
                    r"apollo\.(graphos\.cloud|router\.(operations?|lifecycle|config|schema|query|query_planning|telemetry|instance))(\..*|$)|apollo_router_uplink_fetch_count_total|apollo_router_uplink_fetch_duration_seconds",
                )
                .expect("regex should have been valid"),
            )
            .build()
    }

    pub(crate) fn public<T: Into<MeterProvider>>(delegate: T) -> Self {
        FilterMeterProvider::builder()
            .delegate(delegate)
            .deny(
                Regex::new(r"apollo\.router\.(config|entities|instance|operations\.(connectors|fetch|request_size|response_size)|schema\.connectors)(\..*|$)")
                    .expect("regex should have been valid"),
            )
            .build()
    }
```

timbotnik:
This is looking pretty good to me.

timbotnik (approved):
Good from my side.

SimonSapin:
```rust
error[E0432]: unresolved import `opentelemetry_api`
  --> apollo-router/src/services/router/tests.rs:12:5
   |
12 | use opentelemetry_api::KeyValue;
   |     ^^^^^^^^^^^^^^^^^ use of undeclared crate or module `opentelemetry_api`
```

The dependency on `opentelemetry_api` was removed in https://github.com/apollographql/router/pull/6493. In `dev` I see other files with `use opentelemetry::KeyValue;`, maybe that's what you need?

PR #6451, opened 2024-12-12
Move result handling to initial async block

pubmodmatt:
The Connectors service runs all the REST API requests for an operation as parallel async tasks. However, it was waiting for *all* of them to return a response before processing the results from *any* request.

These changes move the result processing into the same async block as the request, allowing each result to be processed as soon as it is available. These changes don't reduce the total amount of work to be done, but rather change when the work is done to reduce the total latency in some cases. For example, if one task is blocked due to long REST API request latency, we can spend that time processing the results of any requests that have already completed.

The only work that needs to happen after all of the requests complete is the aggregation of the results. The work is now split between two functions, `process_response` and `aggregate_responses`, to allow executing each bit of work at the appropriate time.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

**Notes**

### DIFF
--- a/apollo-router/src/plugins/connectors/handle_responses.rs
+++ b/apollo-router/src/plugins/connectors/handle_responses.rs
@@ -150,7 +150,7 @@ impl RawResponse {
 
 // --- MAPPED RESPONSE ---------------------------------------------------------
 
-enum MappedResponse {
+pub(crate) enum MappedResponse {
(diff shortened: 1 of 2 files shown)


lennyburdette (approved):
💯

PR #6455, opened 2024-12-13
Make `->` method registration more flexible and type-safe

benjamn:
Previously, `->` methods (#5762) were implemented as functions and registered in a `lazy_static!` map for dynamic lookup at runtime.

This PR refactors the system for registering methods, so there is now an `ArrowMethod` enum with a variant for each known method, and each method is now represented by a struct type that has an `::apply` method, whose signature resembles the previous function implementations.

In the future, we can add other methods to these structs, so `->` methods can programmatically define their interactions with the output shape of the JSONSelection, for example.

### DIFF
--- a/apollo-federation/src/sources/connect/json_selection/methods.rs
+++ b/apollo-federation/src/sources/connect/json_selection/methods.rs
@@ -1,6 +1,3 @@
-use apollo_compiler::collections::IndexMap;
-use apollo_compiler::collections::IndexSet;
-use lazy_static::lazy_static;
 use serde_json_bytes::Value as JSON;
 
 use super::immutable::InputPath;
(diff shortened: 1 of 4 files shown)


apollo-federation/src/sources/connect/json_selection/methods.rs
```diff
@@ -24,131 +21,161 @@ mod public;
...
+    And,
+}
 
-        public_methods
+#[macro_export]
+macro_rules! impl_arrow_method {
+    ($struct_name:ident, $trait_name:ident, $impl_fn_name:ident) => {
+        #[derive(Debug, Clone, Copy, PartialEq, Eq)]
+        pub(super) struct $struct_name;
+        impl $trait_name for $struct_name {
```
lennyburdette, line 59: why is trait_name dynamic? just curious, not a blocker

lennyburdette (approved):

apollo-federation/src/sources/connect/json_selection/methods.rs
```diff
@@ -24,131 +21,161 @@ mod public;
...
+        tail: &WithRange<PathList>,
+    ) -> (Option<JSON>, Vec<ApplyToError>);
+}
 
-        methods
-    };
+// This Deref implementation allows us to call .apply(...) directly on the
+// ArrowMethod enum.
+impl std::ops::Deref for ArrowMethod {
+    type Target = dyn ArrowMethodImpl;
```
dylan-apollo, line 90: Oh nice, don't think I've ever done a `dyn` `Deref` before, good to know that's possible.

apollo-federation/src/sources/connect/json_selection/methods.rs
```diff
@@ -24,131 +21,161 @@ mod public;
...
+    pub(super) fn lookup(name: &str) -> Option<Self> {
+        let method_opt = match name {
+            "echo" => Some(Self::Echo),
+            "map" => Some(Self::Map),
+            "eq" => Some(Self::Eq),
+            "match" => Some(Self::Match),
+            // As this case suggests, we can't necessarily provide a name()
+            // method for ArrowMethod (the opposite of lookup), because method
+            // implementations can be used under multiple names.
+            "matchIf" | "match_if" => Some(Self::MatchIf),
```
dylan-apollo, line 137: We could make the design decision _not_ to have multiple names, though. Seems like we should standardize on camelCase and not snake_case.

That would also mean we could use `strum` to impl both `FromStr` and `Display` for all the methods without all the boilerplate.

dylan-apollo: I guess `strum` would make it harder to do the public/future split. Might end up having to add a `#[cfg(test)]` to every future method...

dylan-apollo (approved):
Very excited for the enum!!! My only suggestion is that _maybe_ `strum` would make to/from name conversions easier? But also maybe not...

apollo-federation/src/sources/connect/json_selection/methods.rs
benjamn: I'm open to ideas, though not previously familiar with `strum`.

I'd be ok with sticking to one method name/style, but I think it's worthwhile to keep passing `method_name` as an argument into the method `apply`/`shape` implementations, so there's only one spot in the code we need to change if we want to rename a method (compared to each method implicitly "knowing" its own name and encoding it into error messages, for example).

PR #6456, opened 2024-12-13
[JSONSelection] Support variadic `$(...)` function

benjamn:
When it takes a single argument, the `$(...)` function (aka [`ExprPath` syntax](https://github.com/apollographql/router/blob/909e897200c0f2ccf9187aa6c861b6a14fa42aa3/apollo-federation/src/sources/connect/json_selection/README.md#exprpath-), introduced in #5994), is useful for embedding literal values into selection syntax.

This PR generalizes the `$(...)` function to take one or more arguments, where the first argument that evaluates to `Some(JSON)` instead of `None` is returned, allowing default expressions like `$(possiblyEmptyList->first, {})` or `$($args.possibly.missing, null)`. A benefit of this syntax is that it's insensitive to whether the first expression failed because `$args.possibly` was missing or because `$args.possibly.missing` was missing. The whole expression evaluated (in the `apply_to_path` sense) to `None`, for whatever reason, so the `$(...)` function moves on to the next option.

Passing more than two arguments is also possible, as in `$(maybe.this, maybe.that, false)`.

Like many `->` functions, the `$(...)` syntax evaluates its arguments lazily, so any arguments passed _after_ the one returned will remain fully unevaluated. This makes it possible to use a (hypothetical) `->withError(message)` method to report errors only when an expression is evaluated:
```graphql
$(possiblyEmptyList->first, $(null)->withError("list was empty"))
```

It may also be worth noting that methods like `->match` can evaluate to `None` if they do not have an infallible `[@, ...]` catch-all case, or if the matched value expression evaluates to `None`. In principle, you could use the `$(...)` function to provide a catch-all default, like so:
```graphql
$(kind->match(
  ["book", "Book"],
  ["movie", "Film"],
), "Product")
```
The `->match` method will normally produce a runtime `ApplyToError` when it exhausts all its cases, but the `$(...)` function hides those errors as long as one of its options evaluates to `Some(JSON)`. If they all fail (i.e. evaluate to `None`), then all the original runtime `ApplyToError`s will be returned.

### DIFF
--- a/apollo-federation/src/sources/connect/json_selection/apply_to.rs
+++ b/apollo-federation/src/sources/connect/json_selection/apply_to.rs
@@ -420,9 +420,39 @@ impl ApplyToInternal for WithRange<PathList> {
                     )
                 }
             }
-            PathList::Expr(expr, tail) => expr
-                .apply_to_path(data, vars, input_path)
-                .and_then_collecting_errors(|value| tail.apply_to_path(value, vars, input_path)),
+
+            PathList::Expr(expr_seq, tail) => {
+                // In case none of the expressions in expr_seq successfully
+                // evaluate to a JSON value, we'll want to see all the errors
+                // that could be responsible for that failure. If any expression
+                // succeeds, all_errors will be ignored.
+                let mut all_errors = Vec::new();
+
+                for expr in expr_seq.as_ref() {
+                    let (expr_value, expr_errors) = expr.apply_to_path(data, vars, input_path);
+                    if let Some(value) = expr_value {
+                        // If any expression evaluates to Some(JSON), apply the
+                        // tail to that value and return errors pertaining only
+                        // to this specific expression or its tail, ignoring any
+                        // errors previously added to all_errors.
+                        return tail
+                            .apply_to_path(&value, vars, input_path)
+                            .prepend_errors(expr_errors);
+                    }
+                    // As long as the expressions keep evaluating to None, keep
+                    // accumulating errors in all_errors.
+                    all_errors.extend(expr_errors);
+                }
+
+                all_errors.push(ApplyToError::new(
+                    "No $(...) expression evaluated successfully".to_string(),
+                    input_path.to_vec(),
+                    expr_seq.range(),
+                ));
+
+                (None, all_errors)
+            }
+
             PathList::Method(method_name, method_args, tail) => {
                 let method_path =
                     input_path.append(JSON::String(format!("->{}", method_name.as_ref()).into()));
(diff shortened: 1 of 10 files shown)


apollo-federation/src/sources/connect/json_selection/apply_to.rs
```diff
@@ -420,9 +420,39 @@ impl ApplyToInternal for WithRange<PathList> {
...
+                        // tail to that value and return errors pertaining only
+                        // to this specific expression or its tail, ignoring any
+                        // errors previously added to all_errors.
+                        return tail
+                            .apply_to_path(&value, vars, input_path)
+                            .prepend_errors(expr_errors);
+                    }
+                    // As long as the expressions keep evaluating to None, keep
+                    // accumulating errors in all_errors.
+                    all_errors.extend(expr_errors);
```
lennyburdette, line 444: so if all subexpressions in `$(a, b, c)`, we'd see four errors? (one for each subexpression plus the `No $(...) expression evaluated successfully`)

is there value in the errors for each subexpression? seems like the last error gives you all the information you need, right?

lennyburdette (approved):

apollo-federation/src/sources/connect/json_selection/apply_to.rs
benjamn: I think there's some value in the subexpression errors because they're the ones that will tell you which property specifically was missing, if the expression could fail in multiple ways.

That said, there definitely is a sort of hierarchical structure to the errors, so maybe it doesn't make sense to lump them all into one array, but instead store the subexpression errors as children of the final error somehow?

apollo-bot2:

###  Issues Detected 
| Insecure WebSocket Detected. WebSocket Secure (wss) should be used for all WebSocket connections. | Yes | <a href="https://github.com/apollographql/router/blob/dfeb558df858135e009c4e6324c035dee3ff244b/apollo-router/src/protocols/websocket.rs?plain=1#L958" target="_blank">Link</a> | apollo-  router/src/protocols/websocket.rs | rules.providers.semgrep.security.ja  vascript.lang.security.detect-  insecure-websocket |
| Insecure WebSocket Detected. WebSocket Secure (wss) should be used for all WebSocket connections. | Yes | <a href="https://github.com/apollographql/router/blob/dfeb558df858135e009c4e6324c035dee3ff244b/apollo-router/src/protocols/websocket.rs?plain=1#L1025" target="_blank">Link</a> | apollo-  router/src/protocols/websocket.rs | rules.providers.semgrep.security.ja  vascript.lang.security.detect-  insecure-websocket |
<hr />
<details>

```python
#!/usr/bin/env python3

def my_function():
```

<details>

</details>

</details>
<hr />
<details> 


</details>
<hr />
<details>
</details>

abernix:
Feel free to re-open this, but I'll close this for now for focusing the number of active PRs downward some.  Thanks!

PR #6457, opened 2024-12-13
Implement `path->withError(message)` for inline error reporting

benjamn:
This PR proposes an implementation of the `->withError` method I alluded to in #6456.

PR #6458, opened 2024-12-13
Implement `ApplyToInternal::compute_output_shape` and `JSONSelection::shape` returning `shape::Shape`

benjamn:
This draft PR uses [our newly published `shape` crate](https://crates.io/crates/shape) to implement the `ApplyToInternal::compute_output_shape` trait method, which returns a [`shape::Shape`](https://apollographql.github.io/shape-rs/shape/struct.Shape.html), which is an immutable and reference-counted Rust data structure that describes a set of constraints on JSON data. The "shape" terminology is a close synonym for "type" but I'm using it to emphasize the immutable, reference-free, value-typed nature of the JSON domain.

As an important part of this `::compute_output_shape` system, all `->` methods now define a `method_shape` function that allows them to reject unexpected input shapes and compute their own output shape in terms of a given input shape and variable shapes. In other words, the _type signatures_ of `->` methods are expressed using Rust code which ultimately returns a `shape::Shape`, which then gets further processed by the `compute_output_shape` logic, allowing `->` methods to blend seamlessly into the shape processing system.

This new ability to compute a static output `Shape` for a given JSONSelection string allows us to relax the constraint that inline `PathSelection` items must have a static `{...}` subselection at the end. The static subselection was previously our only means of statically determining the output fields of the `PathSelection`, but now we can statically/programmatically examine the output `Shape`, which doesn't distinguish between static subselections and `->` method results (as long as they have the same computed object shape). We are also now requiring a `...` token before inline `PathSelection` items without static subselections, and encouraging the `...` token for any inline `PathSelection`, even if it has a subselection (though enforcing it in that case would be a breaking change).

Finally, this PR proposes a way of setting the `__typename` in situations where abstract GraphQL types (interfaces and unions) are involved: by writing a `{ <Type> ... }` annotation at the beginning of any subselection, developers can indicate to both the `Shape` system and the runtime system that the `__typename` of this output object should have the literal shape `"Type"`, and the object should include `__typename: "Type"` at runtime. The underlying `Shape` representation actually does use a `__typename` field to model this information, but that's a choice made by the JSONSelection code, not something embedded in the `Shape` library.

I wish I could have broken this PR up more, though I was able to split out #6455, #6456, and #6457.

I don't believe this PR should introduce any logical runtime differences of behavior in the connectors system, since it's basically adding the `compute_output_shape` method and then only using it in tests. Once I address a few `TODO`s, the output shapes will play a role in checking the validity of `...`-inlined `PathSelection` items, but the `...` syntax is new with this PR, so changing it is still easy.


apollo-federation/src/sources/connect/json_selection/README.md
```diff
@@ -82,9 +82,10 @@ below.
...
 ```ebnf
 JSONSelection        ::= PathSelection | NamedSelection*
-SubSelection         ::= "{" NamedSelection* "}"
+SubSelection         ::= "{" OutputTypeAnnotation? NamedSelection* "}"
+OutputTypeAnnotation ::= "<" Identifier ">"
 NamedSelection       ::= NamedPathSelection | PathWithSubSelection | NamedFieldSelection | NamedGroupSelection
-NamedPathSelection   ::= Alias PathSelection
+NamedPathSelection   ::= (Alias | "...") PathSelection
```
benjamn, line 90: Here's the only new JSONSelection syntax we need: `{ <Type> ... }` annotations and `...` before inlined `NamedPathSelection` items.

apollo-federation/src/sources/connect/json_selection/parser.rs
```diff
@@ -324,16 +331,67 @@ impl NamedSelection {
...
+                Err(nom::Err::Failure(e)) => Err(nom::Err::Failure(e)),
+                Err(_) => Err(nom_error_message(
+                    input,
+                    "Path selection alias must be followed by a path",
+                )),
+            }
+        } else if let Ok((remainder, _dots)) =
+            tuple((spaces_or_comments, ranged_span("...")))(input)
+        {
+            // TODO Enforce path has a static output shape with known object keys.
```
benjamn, line 353: Here's a remaining `TODO`.

apollo-federation/src/sources/connect/json_selection/parser.rs
```diff
@@ -362,6 +425,11 @@ impl NamedSelection {
                         name_set.extend(selection.names());
                     }
                     name_set.into_iter().collect()
+                } else if *inline {
+                    // TODO If the path is inlined with ... and has no explicit
+                    // SubSelection, we should use the computed Shape of the
+                    // path to determine the output fields.
```
benjamn, line 431: Another `TODO` I'm planning to get to before taking this PR out of draft.

apollo-federation/src/sources/connect/json_selection/parser.rs
```diff
@@ -969,6 +1062,11 @@ impl SubSelection {
                         // names derived from the SubSelection, which need to be
                         // recursively collected.
                         selections.extend(sub.selections_iter());
+                    } else if *inline {
+                        // If the PathSelection is inlined with ... but does not
+                        // have a subselection, we can use the computed Shape of
+                        // the path to synthesize the output selections.
+                        // TODO Implement this.
```
benjamn, line 1069: Another `TODO` for the list.

apollo-federation/src/sources/connect/json_selection/README.md
```diff
@@ -82,9 +82,10 @@ below.
 
 ```ebnf
 JSONSelection        ::= PathSelection | NamedSelection*
-SubSelection         ::= "{" NamedSelection* "}"
+SubSelection         ::= "{" OutputTypeAnnotation? NamedSelection* "}"
+OutputTypeAnnotation ::= "<" Identifier ">"
 NamedSelection       ::= NamedPathSelection | PathWithSubSelection | NamedFieldSelection | NamedGroupSelection
-NamedPathSelection   ::= Alias PathSelection
+NamedPathSelection   ::= (Alias | "...") PathSelection
```
lennyburdette, line 88: does this allow for future keywords like `if` or `match` directly after the `...`?

apollo-federation/src/sources/connect/json_selection/apply_to.rs
```diff
@@ -69,6 +73,44 @@ impl JSONSelection {
 
         (value, errors.into_iter().collect())
     }
+
+    pub fn shape(&self) -> Shape {
```
lennyburdette, line 77: what's the reason for having shape code in this file? it feels different enough to warrant a separate file and maybe a separate trait, even if it's structured basically the same

apollo-federation/src/sources/connect/json_selection/methods/public.rs
```diff
@@ -157,8 +230,68 @@ fn match_method(
...
+    dollar_shape: Shape,
+    named_var_shapes: &IndexMap<&str, Shape>,
+) -> Shape {
+    if let Some(MethodArgs { args, .. }) = method_args {
+        let mut result_union = Vec::new();
+        let mut has_infallible_case = false;
+
+        for pair in args {
+            if let LitExpr::Array(pair) = pair.as_ref() {
+                if pair.len() == 2 {
```
lennyburdette, line 246: should we return an error if the length isn't exactly 2?

lennyburdette: do you have an example of error handling? since this doesn't return a `Result`, what should a caller do with a `Shape::error_with_range`?

apollo-bot2:

###  Issues Detected 
| Insecure WebSocket Detected. WebSocket Secure (wss) should be used for all WebSocket connections. | Yes | <a href="https://github.com/apollographql/router/blob/70486c1d08e6813dd785e637a75ac3c956b51907/apollo-router/src/protocols/websocket.rs?plain=1#L958" target="_blank">Link</a> | apollo-  router/src/protocols/websocket.rs | rules.providers.semgrep.security.ja  vascript.lang.security.detect-  insecure-websocket |
| Insecure WebSocket Detected. WebSocket Secure (wss) should be used for all WebSocket connections. | Yes | <a href="https://github.com/apollographql/router/blob/70486c1d08e6813dd785e637a75ac3c956b51907/apollo-router/src/protocols/websocket.rs?plain=1#L1025" target="_blank">Link</a> | apollo-  router/src/protocols/websocket.rs | rules.providers.semgrep.security.ja  vascript.lang.security.detect-  insecure-websocket |
<hr />
<details>

```python
#!/usr/bin/env python3

def my_function():
```

<details>

</details>

</details>
<hr />
<details> 


</details>
<hr />
<details>
</details>

apollo-federation/src/sources/connect/json_selection/methods/public.rs
benjamn: I imagine we'll add a way to collect/traverse/visit any `ShapeCase::Error` variants within a given `Shape`, to handle them programmatically or surface them in the debugger, but for now the main point of errors is to fail validation against expected GraphQL schema types.

Since you can embed `ShapeCase::Error` within a larger shape, shape errors are similar to the GraphQL errors-as-data pattern, where you're able to get back some partial shape information along with contextual errors.

apollo-federation/src/sources/connect/json_selection/README.md
benjamn: If the `PathSelection` started with an `if` token, that would be a potential concern for making `if` a keyword in the future, but if we stick with the the idea of a parenthesized test (i.e. `if (test) ...`), the `(` character should prevent parsing as a `PathSelection`.

Same goes for `match (value) ...`, as long as we use parentheses, or as long as we're willing to make the `match` field illegal immediately after `...` (a breaking change but not unthinkable).

apollo-federation/src/sources/connect/json_selection/apply_to.rs
benjamn: It made the code a little easier to write to have `apply_to_path` nearby, but there's no harm in moving it to a new file now.

apollo-federation/src/sources/connect/json_selection/methods.rs
```diff
@@ -68,6 +71,24 @@ macro_rules! impl_arrow_method {
...
+
+            fn shape(
+                &self,
+                method_name: &WithRange<String>,
+                method_args: Option<&MethodArgs>,
+                input_shape: Shape,
+                dollar_shape: Shape,
+                named_var_shapes: &IndexMap<&str, Shape>,
+            ) -> Shape {
+                // TODO
```
dylan-apollo, line 83:
```suggestion
```

Unless there's a follow-up task to make for this

dylan-apollo (approved):

apollo-federation/src/sources/connect/json_selection/methods.rs
benjamn: https://github.com/apollographql/router/pull/6458/commits/87a2ecd072d2782b161a973234e6fbbf1481f2f6

benjamn:
The `TODO`s I mentioned above are important, but I'm going to attempt them after merging PR #6513 on top of this PR, rather than addressing them in this PR.

apollo-federation/src/sources/connect/json_selection/README.md
benjamn: Postponing this syntax for a later PR/release.

apollo-federation/src/sources/connect/json_selection/apply_to.rs
benjamn: I'll make this move in a follow-up PR.

PR #6461, opened 2024-12-16
pre-GA connector error cleanup

lennyburdette:
1. switch to plain GraphQLError so we can support future customizations — these errors are never returned in a `Result`, so they don't need to be `FetchError` variants.
2. use `CONNECTOR_FETCH` for `extensions.code` and the subgraph for `extension.service`. this keeps things consistent with other errors.
3. reference the connector using a more opaque "coordinate" — this avoids exposing internal implementation details like URLs while still pointing directly to the relevant connector.

when we start supporting error customizations, i think we'll enforce that `extensions.code`, `extensions.service`, and `extensions.connector.coordinate` are always present by adding them after the customizations.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [x] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-bot2:

###  Issues Detected 
| Insecure WebSocket Detected. WebSocket Secure (wss) should be used for all WebSocket connections. | Yes | <a href="https://github.com/apollographql/router/blob/83a981cc0405382d44e331faa611ba053b05f73d/apollo-router/src/protocols/websocket.rs?plain=1#L958" target="_blank">Link</a> | apollo-  router/src/protocols/websocket.rs | rules.providers.semgrep.security.ja  vascript.lang.security.detect-  insecure-websocket |
| Insecure WebSocket Detected. WebSocket Secure (wss) should be used for all WebSocket connections. | Yes | <a href="https://github.com/apollographql/router/blob/83a981cc0405382d44e331faa611ba053b05f73d/apollo-router/src/protocols/websocket.rs?plain=1#L1025" target="_blank">Link</a> | apollo-  router/src/protocols/websocket.rs | rules.providers.semgrep.security.ja  vascript.lang.security.detect-  insecure-websocket |
<hr />
<details>

```python
#!/usr/bin/env python3

def my_function():
```

<details>

</details>

</details>
<hr />
<details> 


</details>
<hr />
<details>
</details>

apollo-router/src/plugins/connectors/handle_responses.rs
```diff
@@ -126,17 +128,27 @@ impl RawResponse {
...
-                    service: connector.id.label.clone(),
-                    reason: format!(
-                        "{}: {}",
-                        parts.status.as_str(),
-                        parts.status.canonical_reason().unwrap_or("Unknown")
-                    ),
-                }
-                .to_graphql_error(Some((&key).into()))
-                .add_subgraph_name(&connector.id.subgraph_name);
+                let error = graphql::Error::builder()
```
pubmodmatt, line 131: I'm not sure I understand the motivation for using `graphql::Error` here. This will just get converted right back into a `SubrequestHttperror` in `FetchService::map_to_graphql_error`, and then get converted *again* into `graphql::Error`.

lennyburdette: I don’t believe that’s true. These errors are added directly to the “errors” list on the response, not returned in a Result. The FetchError handling and SubrequstHttpError conversion happens in `.map_err` and related methods, which will never see these errors. 

In general, handle_response.rs code only returns Err if something really unexpected went wrong, not if a request fails.

pubmodmatt: Ah, right, I forgot about how it wraps the error in `Ok`.

pubmodmatt (approved):

PR #6470, opened 2024-12-17
fix: cleanup defer handling within the QP

dariuszkuc:
If user explicitly opt-out of the `@defer` support, deferred queries will be rejected by the router with GraphQL validation error (unknown directive). It is not possible to plan a query with a `@defer` in it when defer support is disabled so we can simplify this logic.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-federation/tests/query_plan/mod.rs
```diff
@@ -1,5 +1,3 @@
 #[macro_use]
 mod build_query_plan_support;
 mod build_query_plan_tests;
-// TODO: port query-planner-js/src/__tests__/buildPlan.*.test.ts as new modules here
-mod operation_validations_tests;
```
dariuszkuc, line 5: NOTE: this change is unrelated to the defer cleanup. Those tests were not implemented and their JS counterpart were covering general GraphQL validations that are already handled by `apollo-rs` and reuse fragment logic.

apollo-federation/src/operation/tests/mod.rs
```diff
@@ -17,8 +17,6 @@ use crate::schema::position::InterfaceTypeDefinitionPosition;
 use crate::schema::position::ObjectTypeDefinitionPosition;
 use crate::schema::ValidFederationSchema;
 
-mod defer;
```
dariuszkuc, line 20: those tests were verifying `without_defer` functionality

dariuszkuc:
Was going to remove `QueryPlannerConfig#incremental_delivery` but that is actually used to generate proper API schema. Comment on `QueryPlanIncrementalDeliveryConfig` seems reasonable for "future" use case (i.e. when subgraphs add support for `@defer/@stream`) .... but it might be a while before it happens so I'd just drop it and reintroduce it back WHEN we need it.

We generate API schema twice -> once by the router and once by the QP. Maybe we should update our `QueryPlanner::new` signature to accept `Schema` from the router instead (it contains both the supergraph and api schema)?

apollo-federation/src/operation/directive_list.rs
```diff
@@ -332,6 +335,18 @@ impl DirectiveList {
         inner.rehash();
         Some(item)
     }
+
+    /// Removes @defer directive from self if it has a matching label.
+    pub(crate) fn remove_defer(&mut self, defer_labels: &IndexSet<String>) {
```
dariuszkuc, line 340: This was lifted out of the old `DeferFilter`. We don't need to pass schema anymore as at this stage all `@defer` applications will already have labels normalized.

dariuszkuc:
We'll tackle the duplicate API schema generation issue and cleanup the overall API in other PR.

duckki (approved):
Looks good!

apollo-federation/tests/query_plan/build_query_plan_tests/defer.rs
```diff
@@ -6,71 +6,6 @@ fn config_with_defer() -> QueryPlannerConfig {
     config
 }
 
-#[test]
-fn defer_test_handles_simple_defer_without_defer_enabled() {
```
TylerBloom, line 10: Is there a reason this was removed? It was ported directly from JS, and, IMO, is testing behavior that we care about.

TylerBloom (approved):
Other than the one comment about a test being removed, looks good to me.

apollo-federation/tests/query_plan/build_query_plan_tests/defer.rs
duckki: The changed behavior does not allow `@defer` directive if defer is enabled, while previously `@defer` was removed silently. So, the tested scenario can no longer happen.

However, I just noticed the test would still be able to run. I think this function need to be updated:
https://github.com/apollographql/router/blob/6342406c70edcefbc62bbb6c37bd397a4dc080b9/apollo-federation/tests/query_plan/build_query_plan_support.rs#L98
```rust
pub(crate) fn api_schema_and_planner(
    function_path: &'static str,
    config: QueryPlannerConfig,
    subgraph_names_and_schemas: &[(&str, &str)],
) -> (ValidFederationSchema, QueryPlanner) {
    let supergraph = compose(function_path, subgraph_names_and_schemas);
    let supergraph = apollo_federation::Supergraph::new(&supergraph).unwrap();
    let planner = QueryPlanner::new(&supergraph, config).unwrap();
    let api_schema_config = apollo_federation::ApiSchemaOptions {
        include_defer: true,
        include_stream: false,
    };
    let api_schema = supergraph.to_api_schema(api_schema_config).unwrap();
    (api_schema, planner)
}
```
The config should control `ApiSchemaOptions`'s `include_defer`.

dariuszkuc: thats the thing -> outside of tests it is not possible for this scenario to ever reach QP. Old test was verifying the behavior that we would strip out `@defer` from incoming requests if it was disabled.... but if user explicitly opts out of `@defer` but still send a query with `@defer` it will be rejected during GraphQL validation by `apollo-rs` (as that directive won't exist in the API schema).

So real use cases are
- defer enabled
  - query with defer (need to normalize it)
  - query without defer 
- defer disabled
  - query without defer

dariuszkuc: Fixed the test api schema config in https://github.com/apollographql/router/pull/6470/commits/6d7a1ca018e9ba35ab8b6937bb4fb9112269352d

PR #6476, opened 2024-12-17
Migrate  `value.*` metrics to gauges

goto-bus-stop:
Migrates `tracing::info!(value.)` metrics to otel instruments.

Notes in comments


apollo-router/src/axum_factory/axum_http_server_factory.rs
```diff
@@ -554,6 +567,14 @@ where
         router = router.route("/", get(handle_graphql::<RF>).post(handle_graphql::<RF>));
     }
 
+    // Tie the lifetime of the session count instrument to the lifetime of the router
+    // by moving it into a no-op layer.
+    let instrument = session_count_instrument();
+    router = router.layer(layer_fn(move |service| {
+        let _ = &instrument;
+        service
+    }));
```
goto-bus-stop, line 576: The `session_count_active` instrument is "stored" in an `axum::Router` instance by moving it into a no-op layer. Another option would be to just have one global instrument for the entire lifetime of the program, but I'm not sure if that would work everywhere we need it to (eg. does it do the right thing across reloads, tests, etc?)

apollo-router/src/plugins/telemetry/tracing/apollo_telemetry.rs
```diff
@@ -261,6 +265,47 @@ impl LightSpanData {
...
+///
+/// This *could* be generalised to any kind of gauge, but we should ideally have gauges that can just
+/// observe their accurate value whenever requested. The externally updateable approach is kind of
+/// a hack that happens to work here because we only have one place where the value can change, and
+/// otherwise we might have to use an inconvenient Mutex or RwLock around the entire LRU cache.
+#[derive(Debug)]
+struct SpanLruSizeInstrument {
+    value: Arc<AtomicU64>,
+    _gauge: ObservableGauge<u64>,
+}
```
goto-bus-stop, line 280: As mentioned in the doc comment, this is a hack that kind of mimicks what `tracing::info!(value.)` was doing (storing the value whenever it is updated, and reading the stored value when requested).

I'm not sure about this metric, should we keep it at all in 2.0?

bnjjj: I think we should keep it. It's really helpful to debug and could potentially find the root cause of a leak or so.

bnjjj (approved):

PR #6486, opened 2024-12-18
Provide a backpressure enabled pipeline

garypen:
Most of the functionality is now in place for effective traffic-shaping at the router and subgraph levels.

Please read the review guidance before starting your review.

dev-docs/BACKPRESSURE_REVIEW_NOTES.md

[ROUTER-626]: https://apollographql.atlassian.net/browse/ROUTER-626?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ROUTER-53]: https://apollographql.atlassian.net/browse/ROUTER-53?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ROUTER-774]: https://apollographql.atlassian.net/browse/ROUTER-774?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ROUTER-641]: https://apollographql.atlassian.net/browse/ROUTER-641?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ROUTER-301]: https://apollographql.atlassian.net/browse/ROUTER-301?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[ROUTER-980]: https://apollographql.atlassian.net/browse/ROUTER-980?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apollo-router/src/services/router/service.rs
```diff
@@ -934,23 +950,7 @@ impl RouterCreator {
...
-        ServiceBuilder::new()
-            .layer(self.static_page.clone())
-            .service(
-                self.supergraph_creator
-                    .plugins()
-                    .iter()
-                    .rev()
-                    .fold(router_service.boxed(), |acc, (_, e)| e.router_service(acc)),
-            )
+        self.sb.lock().clone()
```
bnjjj, line 953: We should document why it's ok to lock the mutex here

apollo-router/src/services/router/service.rs
```diff
@@ -916,13 +912,33 @@ impl RouterCreator {
...
+            query_analysis_layer,
+            configuration.batching.clone(),
+        ));
+
+        // NOTE: This is the start of the router pipeline (router_service)
+        let sb = ServiceBuilder::new()
+            // NOTE: Buffer is required because the static_page service is not Clone.
+            // The buffer should be >= concurrency as per the advice
+            // at: https://docs.rs/tower/latest/tower/buffer/struct.Buffer.html#a-note-on-choosing-a-bound
+            .buffer(50_000)
```
bnjjj, line 928: Can't we just apply the buffer only on static_page ?

garypen: Probably. I'm also hopeful that we could modify static_page to be `Clone` and avoid the Buffer completely. We aren't quite at that stage of the work yet. Still trying to get tests to pass.

apollo-router/src/services/router/service.rs
garypen: Yes and in SupergraphCreator. I will do that before marking this PR ready to review.

apollo-router/src/services/router/service.rs
garypen: I looked into this at some length and it's actually SyncCheckpoint which we would need to modify to be `Clone` and we can't, because if we did, we would need to modify all our plugin interfaces to be `BoxCloneService` rather than `BoxService`.

I think this is something we won't be able to fix in router 2.0.

.changesets/feat_garypen_next_backpressure.md
```diff
@@ -0,0 +1,9 @@
+### Enabling back-pressure in the request processing pipeline ([PR #6486](https://github.com/apollographql/router/pull/6486))
+
+In Router 1.x, back-pressure was not maintained. Requests would be accepted by the router. This could cause issue for routers which were accepting high levels of traffic.
```
BrynCooke, line 3: Lets go into more detail about the benefits

apollo-router/src/axum_factory/tests.rs
```diff
@@ -2406,114 +2395,3 @@ async fn test_supergraph_and_health_check_same_port_different_listener() {
         error.to_string()
     );
 }
-
-#[tokio::test]
-async fn test_supergraph_timeout() {
```
BrynCooke, line 2411: This has been replaced by a test in traffic shaping

apollo-router/src/layers/async_checkpoint.rs
```diff
@@ -63,86 +61,11 @@ where
...
             checkpoint_fn: Arc::clone(&self.checkpoint_fn),
-            inner: service,
+            service,
         }
     }
 }
 
-/// [`Service`] for OneShot (single use) Asynchronous Checkpoints. See [`ServiceBuilderExt::oneshot_checkpoint_async()`](crate::layers::ServiceBuilderExt::oneshot_checkpoint_async()).
-#[allow(clippy::type_complexity)]
-pub struct OneShotAsyncCheckpointService<S, Fut, Request>
```
BrynCooke, line 73: This has been removed because it doesn't work this service expects only to be called once. We are now reusing services.

It's pub so we need to add it to the migration guide.

apollo-router/src/layers/async_checkpoint.rs
```diff
@@ -277,21 +156,21 @@ mod async_checkpoint_tests {
...
+            .times(1)
+            .returning(move |req| {
+                Ok(ExecutionResponse::fake_builder()
+                    .label(expected_label.to_string())
+                    .context(req.context)
+                    .build()
+                    .unwrap())
+            });
 
         let service_stack = ServiceBuilder::new()
```
BrynCooke, line 175: Let's create a ticket to migrate custom layers to use tower_test.

apollo-router/src/plugins/authorization/mod.rs
```diff
@@ -550,6 +550,8 @@ impl Plugin for AuthorizationPlugin {
         if self.require_authentication {
             ServiceBuilder::new()
                 .checkpoint(move |request: supergraph::Request| {
+                    // XXX(@goto-bus-stop): Why are we doing this here, as opposed to the
```
BrynCooke, line 553: Let's create a ticket for this

apollo-router/src/plugins/cache/entity.rs
```diff
@@ -518,23 +523,20 @@ impl Service<subgraph::Request> for CacheService {
...
     fn poll_ready(
         &mut self,
-        cx: &mut std::task::Context<'_>,
+        _cx: &mut std::task::Context<'_>,
     ) -> std::task::Poll<Result<(), Self::Error>> {
-        match &mut self.0 {
-            Some(s) => s.service.poll_ready(cx),
-            None => panic!("service should have been called only once"),
-        }
+        Poll::Ready(Ok(()))
```
BrynCooke, line 528: Should be calling `service.poll_ready`

apollo-router/src/plugins/cache/entity.rs
```diff
@@ -548,7 +550,7 @@ impl InnerCacheService {
             .extensions()
             .with_lock(|lock| lock.contains_key::<BatchQuery>())
         {
-            return self.service.call(request).await;
+            return self.service.ready().await?.call(request).await;
```
BrynCooke, line 553: Remove the `ready().await?`. Plus other occurences.

apollo-router/src/plugins/limits/layer.rs
```diff
@@ -19,24 +19,36 @@ pub(super) enum BodyLimitError {
...
 
 struct BodyLimitControlInner {
-    limit: AtomicUsize,
+    limit: Arc<AtomicUsize>,
     current: AtomicUsize,
 }
 
 /// This structure allows the body limit to be updated dynamically.
 /// It also allows the error message to be updated
+///
```
BrynCooke, line 28: @BrynCooke TODO. Look at BodyLimit in more detail.

apollo-router/src/plugins/traffic_shaping/deduplication.rs
```diff
@@ -72,7 +72,7 @@ where
     }
 
     async fn dedup(
```
BrynCooke, line 74: @BrynCooke Look in detail

apollo-router/src/plugins/traffic_shaping/deduplication.rs
```diff
@@ -187,19 +182,20 @@ where
...
+        let inner = std::mem::replace(&mut self.service, service);
 
         if request.operation_kind == OperationKind::Query {
             let wait_map = self.wait_map.clone();
 
-            Box::pin(async move { Self::dedup(service, wait_map, request).await })
+            Box::pin(async move { Self::dedup(inner, wait_map, request).await })
         } else {
-            Box::pin(async move { service.oneshot(request).await })
+            Box::pin(async move { inner.oneshot(request).await })
```
BrynCooke, line 198:
```suggestion
            Box::pin(async move { inner.call(request).await })
```

apollo-router/src/plugins/traffic_shaping/mod.rs
```diff
@@ -340,59 +318,89 @@ impl TrafficShaping {
...
+                                        .error(error)
+                                        .context(ctx)
+                                        .build())
+                                }
+                                Err(err) => Err(err),
+                            }
+                        }
+                    },
+                )
+                .load_shed()
```
BrynCooke, line 368: @BrynCooke Let's try and stick to the pattern in the router service

apollo-router/src/query_planner/tests.rs
```diff
@@ -93,9 +95,9 @@ async fn mock_subgraph_service_withf_panics_should_be_reported_as_service_closed
     };
 
     let mut mock_products_service = plugin::test::MockSubgraphService::new();
-    mock_products_service.expect_call().times(1).withf(|_| {
```
BrynCooke, line 96: Possibly points to a bug. The original assertion placement was correct because the original service should be consumed.

apollo-router/src/services/router/service.rs
```diff
@@ -860,11 +877,7 @@ pub(crate) fn process_vary_header(headers: &mut HeaderMap<HeaderValue>) {
 #[derive(Clone)]
 pub(crate) struct RouterCreator {
     pub(crate) supergraph_creator: Arc<SupergraphCreator>,
-    static_page: StaticPageLayer,
-    apq_layer: APQLayer,
-    pub(crate) persisted_query_layer: Arc<PersistedQueryLayer>,
-    query_analysis_layer: QueryAnalysisLayer,
-    batching: Batching,
+    sb: Arc<parking_lot::Mutex<router::BoxCloneService>>,
```
BrynCooke, line 880: @BrynCooke Maybe this could be just buffer rather than in a mutex.

apollo-router/tests/integration/traffic_shaping.rs
```diff
@@ -155,9 +161,12 @@ async fn test_router_timeout_custom_metric() -> Result<(), BoxError> {
     let (_trace_id, response) = router.execute_default_query().await;
     assert_eq!(response.status(), 504);
     let response = response.text().await?;
-    assert!(response.contains("REQUEST_TIMEOUT"));
+    assert!(response.contains("GATEWAY_TIMEOUT"));
```
BrynCooke, line 164: @BrynCooke Fixup these tests to use http.server.response metrics on status.

BrynCooke (changes requested):

apollo-router/src/services/router/service.rs
```diff
@@ -260,11 +262,18 @@ impl RouterService {
                     .await
                 {
                     Err(response) => response,
-                    Ok(request) => self.supergraph_creator.create().oneshot(request).await?,
+                    Ok(request) => {
+                        // This is safe to do because we cloned the entire router_service before
+                        // calling it.
+                        self.supergraph_service.oneshot(request).await?
```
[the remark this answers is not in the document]
goto-bus-stop: I'll write down how I understand it right now after our discussion:
`self.supergraph_service` here is a clone of the service that was readied in `RouterService::poll_ready`. Clones are un-ready by default, so this `self.supergraph_service` is actually *not* ready, which is why we need to oneshot it here. That technically breaks backpressure, but because we are still readying the supergraph service before calling into the router service, backpressure is actually still exerted at that point--there's just potential for some requests to slip through the cracks and end up queueing up at this `.oneshot()` call.

apollo-router/src/services/router/service.rs
BrynCooke: We can get rid of the mutex. It requires an extra box, either on the service after it's cloned or on the future. I don't have strong feeling here.

apollo-router/src/layers/async_checkpoint.rs
garypen: https://apollographql.atlassian.net/browse/ROUTER-984

apollo-router/src/plugins/authorization/mod.rs
garypen: https://apollographql.atlassian.net/browse/ROUTER-985

apollo-router/src/plugins/cache/entity.rs
garypen: There is no `service` to call here, since this is the end of the chain. This is correct if you look at what the call function does.

It basically clones itself, then uses the clone to drive the service.

.changesets/feat_garypen_next_backpressure.md
garypen: Done, but most of the details will be in the migration guide.

apollo-router/src/plugins/traffic_shaping/deduplication.rs
garypen: Done

apollo-router/src/services/hickory_dns_connector.rs
```diff
@@ -47,6 +47,7 @@ impl Service<Name> for AsyncHyperResolver {
 
     fn call(&mut self, name: Name) -> Self::Future {
         let resolver = self.0.clone();
```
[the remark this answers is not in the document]
garypen: I've put the resolver in an Arc.

apollo-router/src/services/http/service.rs
```diff
@@ -248,11 +248,23 @@ impl tower::Service<HttpRequest> for HttpClientService {
 
         #[cfg(unix)]
         let client = match schema_uri.scheme().map(|s| s.as_str()) {
-            Some("unix") => Either::Right(self.unix_client.clone()),
-            _ => Either::Left(self.http_client.clone()),
```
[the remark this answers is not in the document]
garypen: https://apollographql.atlassian.net/browse/ROUTER-986

apollo-router/src/services/layers/apq.rs
```diff
@@ -73,6 +73,17 @@ impl APQLayer {
         Self { cache: None }
     }
 
+    /// Supergraph service implementation for Automatic Persisted Queries.
```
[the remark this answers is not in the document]
garypen: done.

apollo-router/src/services/layers/content_negotiation.rs
```diff
@@ -113,7 +129,12 @@ where
     }
 }
 
-/// [`Layer`] for Content-Type checks implementation.
+/// A layer for the supergraph service that populates the Content-Type response header.
+///
+/// The content type is decided based on the [`ClientRequestAccepts`] context value, which is
+/// populated by the content negotiation [`RouterLayer`].
+// XXX(@goto-bus-stop): this feels a bit odd. It probably works fine because we can only ever respond
+// with JSON, but maybe this should be done as close as possible to where we populate the response body..?
```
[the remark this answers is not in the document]
garypen: https://apollographql.atlassian.net/browse/ROUTER-987

apollo-router/src/services/router/service.rs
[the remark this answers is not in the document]
garypen: I've added this comment to the source.

apollo-router/src/services/router/service.rs
garypen: done: 9b7542300e03fc18de7d0a2447703af322169a55

apollo-router/src/plugins/cache/entity.rs
garypen: Ignore my comment. I fixed this: 017b0f1e3fc88855b14ddcfc513fa2415bedbc11

apollo-router/src/query_planner/tests.rs
goto-bus-stop: The change is because service "creation" is now actually a clone (inside MakeSubgraphService::make). So that seems fine. The test as a whole is a bit odd and I need a bit longer to wrap my head around what it's really trying to do.

apollo-router/src/plugins/limits/layer.rs
BrynCooke: Fixed in [f5d3bdb](https://github.com/apollographql/router/pull/6486/commits/f5d3bdb4bd96fbdc18ab16a9c00c44b0345a1428)

apollo-router/tests/integration/traffic_shaping.rs
BrynCooke: [8ea3123](https://github.com/apollographql/router/pull/6486/commits/8ea31233935e9b5d717d61e07dddf0b191cec2ef)

apollo-router/src/plugins/traffic_shaping/deduplication.rs
BrynCooke: I think Dedupe is OK. We can tackle the performance issues it causes in a follow-up

BrynCooke (approved):

apollo-router/src/plugins/traffic_shaping/mod.rs
BrynCooke: This can be done later.

PR #6495, opened 2024-12-20
Migrate `apollo_router_session_count_total` metric to an OTel gauge

goto-bus-stop:
Somehow missed this in #6476

No changeset because it's in #6476 : )

### DIFF
--- a/apollo-router/src/axum_factory/listeners.rs
+++ b/apollo-router/src/axum_factory/listeners.rs
@@ -14,6 +14,8 @@ use futures::channel::oneshot;
 use futures::prelude::*;
 use hyper::server::conn::Http;
 use multimap::MultiMap;
+use opentelemetry::metrics::MeterProvider;
+use opentelemetry::KeyValue;
 #[cfg(unix)]
 use tokio::net::UnixListener;
 use tokio::sync::mpsc;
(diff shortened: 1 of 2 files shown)


tninesling (approved):

apollo-router/src/axum_factory/listeners.rs
```diff
@@ -234,16 +262,10 @@ pub(super) fn serve_router_on_listen_addr(
...
+
                             // We only want to recognise sessions if we are the main graphql port.
-                            if main_graphql_port {
-                                let session_count = SESSION_COUNT.fetch_add(1, Ordering::Acquire)+1;
-                                tracing::info!(
-                                    value.apollo_router_session_count_total = session_count,
-                                    listener = &address
-                                );
-                            }
+                            let _guard = main_graphql_port.then(TotalSessionCountGuard::start);
```
goto-bus-stop, line 267: Note this contains a mistake. The guard must be moved into the `task::spawn` call below, but it isn't, so the total session count metric is broken with this PR (it is immediately decremented before the next incoming request is handled.)

https://github.com/apollographql/router/pull/6527 will fix this problem.

goto-bus-stop: That PR will be redone in a different way. https://github.com/apollographql/router/pull/6539 fixes just this problem.

PR #6498, opened 2024-12-21
feat(federation): query plan correctness checker

duckki:
This PR implements the first version of the query plan correctness checker.
It implemented the completeness part of the plan. I'll create a separate PR for the soundness part.

Status
- [x] ResponseShape computation from operations
- [x] ResponseShape computation from query plans
- [x] ResponseShape comparison
- [x] Handle all PathElements.
- [x] Handle Defer/Subscription nodes
- [x] Subgraph constraints handling
- [x] Corpus run


apollo-federation/src/query_plan/correctness/response_shape_test.rs
```diff
@@ -0,0 +1,518 @@
...
+    "#;
+    // Note: The set {R, S} has no corresponding named type definition in the schema, while
+    //       `U ∧ J` is just the same as `U`.
+    insta::assert_snapshot!(response_shape(op_str), @r###"
+    {
+      test_u -----> test_u {
+        data -may-> data(arg: 0) on I ∧ U = {R, S}
+        data -may-> data(arg: 0) on U
+      }
+    }
```
lrlna, line 155: This needs a bit of clarification, or more clearly named types. It becomes confusing to reason about single letter types in sets and intersections 😅 . For example, we can use the `Pet` schema that graphql js uses, or expand on the startstuff schema to add more unions?

Also, what does the `----->` actually denote in these snapshots? I take it `-may->` is an option? So, `test_u` has two possible options?

duckki: I wrote some design doc ([google doc](https://docs.google.com/document/d/1qlBLEfu4wRyvLvUrrWqR7MGnap5nsTivRao9cZAFVfI/edit?usp=drive_link)).

duckki:
I have finished the first version (the completeness checking part) and rebased it on more recent `dev` branch commit.
This is now ready for review.

apollo-federation/src/correctness/response_shape_compare.rs
```diff
@@ -0,0 +1,387 @@
...
+                $a,
+                $b,
+            );
+            return Err(CheckFailure::new(message));
+        }
+    };
+}
+
+// Path-specific type constraints on top of GraphQL type conditions.
+pub(crate) trait PathConstraint<'a>
```
TylerBloom, line 33: What is the lifetime parameter doing?

apollo-federation/src/correctness/response_shape_compare.rs
```diff
@@ -0,0 +1,387 @@
...
+            def.conditional_variants()
+                .iter()
+                .try_for_each(|variant| result.insert_variant(variant.clone()))
+                .map_err(|e| {
+                    CheckFailure::new(format!(
+                        "collect_definitions_for_type_condition failed for {filter_cond}\ntype_cond: {type_cond}\nerror: {e}",
+                    ))
+                })?;
+        }
+    }
```
TylerBloom, line 118: This pattern is generally a code smell to me (starting with `None`, finding something, and then using it in a loop) as it can be hard to follow. It seems that `result` is the first `def` whose type condition is implied by the filter condition. IMO, it is much clearer to write that.
```rust
let mut iter = defs
        .iter()
        .filter(|(type_cond, _)| filter_cond.implies(type_cond));
    let Some((type_cond, result)) = iter.next() else {
        return Err(CheckFailure::new(format!(
            "no definitions found for type condition: {filter_cond}"
        )));
    };
    let mut digest = result.clone();
    std::iter::once((type_cond, result)).chain(iter).try_for_each(|(type_cond, def)|
            def.conditional_variants()
                .iter()
                .try_for_each(|variant| digest.insert_variant(variant.clone()))
                .map_err(|e| {
                    CheckFailure::new(format!(
                        "collect_definitions_for_type_condition failed for {filter_cond}\ntype_cond: {type_cond}\nerror: {e}",
                    ))
                })
        )?;
    Ok(digest)
```

apollo-federation/src/correctness/response_shape_compare.rs
```diff
@@ -0,0 +1,387 @@
...
+        },
+    )?;
+    this.conditional_variants()
+        .iter()
+        .try_for_each(|this_def| {
+            // search the same boolean clause in other
+            let found = other
+                .conditional_variants()
+                .iter()
+                .fallible_any(|other_def| {
```
TylerBloom, line 244: I think the `fallible_find` method was removed during cleanup, but this would be an excellent place to bring it back.

apollo-federation/src/correctness/subgraph_constraint.rs
```diff
@@ -0,0 +1,155 @@
...
+    let mut resolvable_key_applications = Vec::new();
+    for directive in ty_def.directives.get_all(&key_directive_definition.name) {
+        let key_directive_application =
+            federation_spec_definition.key_directive_arguments(directive)?;
+        if !key_directive_application.resolvable {
+            continue;
+        }
+        resolvable_key_applications.push(key_directive_application);
+    }
+    Ok(!resolvable_key_applications.is_empty())
```
TylerBloom, line 49: I don't think you need a `Vec` here. We only need the first resolvable key application. `map().find_ok().map(Option::is_some)` should suffice.

apollo-federation/src/correctness/query_plan_analysis.rs
```diff
@@ -0,0 +1,650 @@
...
+}
+
+fn interpret_fetch_node(
+    schema: &ValidFederationSchema,
+    _state: &ResponseShape,
+    conditions: &[Literal],
+    fetch: &FetchNode,
+) -> Result<ResponseShape, String> {
+    let mut result = if let Some(_requires) = &fetch.requires {
+        // TODO: check requires
```
TylerBloom, line 360: Is this needed for this PR?

TylerBloom (approved):
The core looks very good and inline with the design docs. I have some general cleanup comments

duckki:
Thanks, @TylerBloom . The refactored code looks so much better!

duckki:
@mergify queue

duckki:
@mergify backport 1.x

PR #6500, opened 2025-01-01
chore: Update jsonschema usage

Stranger6667:
Hey!

This PR updates the `jsonschema` crate and applies its newer APIs.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1305,24 +1305,24 @@ dependencies = [
 
 [[package]]
 name = "bit-set"
-version = "0.5.3"
+version = "0.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0700ddab506f33b20a03b13996eccd309a48e5ff77d0d95926aa0210fb4e95f1"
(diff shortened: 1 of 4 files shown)


goto-bus-stop:
That's awesome, thank you!

It looks like some of our generated `definitions` have naming that isn't allowed in a `uri-reference` anymore (things like `#/definitions/apollo_router::Extendable<RouterStage>`).

Does `jsonschema` support URL-encoded references? Should it work if we produced something like this instead?
```json
{
  "definitions": { "apollo_router::Extendable<RouterStage>": { /* etc */ } }
}
// elsewhere...
{
  "$ref": "#/definitions/apollo_router%3A%3AExtendable%3CRouterStage%3E"
}
```

Stranger6667:
@goto-bus-stop 

I think it should support them, yep :) I haven't tried it myself, though

Is there a test failure because of this?

Stranger6667:
@abernix I don't have the bandwidth this week, but I can take a look a bit later

goto-bus-stop:
I'll rebase this after https://github.com/apollographql/router/pull/7893 lands.

PR #6285, opened 2024-11-19
Remove experimental_when_header feature

bnjjj:
If you used the `experimental_when_header` feature previously like this for example:

```yaml title="router.previous.yaml"
telemetry:
  exporters:
    logging:
      # If one of these headers matches we will log supergraph and subgraphs requests/responses
      experimental_when_header: # REMOVED
        - name: apollo-router-log-request
          value: my_client
          headers: true # default: false
          body: true # default: false
```

This feature no longer exists and can be replaced with another one included in custom telemetry. Here is how you would configure custom telemetry to achieve the same result:

```yaml title="router.yaml"
telemetry:
  instrumentation:
    events:
      router:
        request: # Display router request log
          level: info
          condition:
            eq:
            - request_header: apollo-router-log-request
            - my_client
        response: # Display router response log
          level: info
          condition:
            eq:
            - request_header: apollo-router-log-request
            - my_client
      supergraph:
        request: # Display supergraph request log
          level: info
          condition:
            eq:
            - request_header: apollo-router-log-request
            - my_client
        response:
          level: info
          condition:
            eq:
            - request_header: apollo-router-log-request
            - my_client
      subgraph:
        request: # Display subgraph request log
          level: info
          condition:
            eq:
            - supergraph_request_header: apollo-router-log-request
            - my_client
        response: # Display subgraph response log
          level: info
          condition:
            eq:
            - supergraph_request_header: apollo-router-log-request
            - my_client
```

[ROUTER-786]: https://apollographql.atlassian.net/browse/ROUTER-786?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #6507, opened 2025-01-03
docs: add debugging client request overview

Meschreiber:
Adds an overview page for debugging client requests


docs/source/routing/observability/debugging-client-requests.mdx
```diff
@@ -0,0 +1,82 @@
...
+          body: true # default: false
+```
+
+### Debugging router logs
+
+By default, the router uses the `info` level for its logging. [Enabling other logging levels](/graphos/reference/router/telemetry/log-exporters/overview) can help debug specific scenarios. Using non-`info` level configurations is only recommended for local or non-production environments.
+
+### Rhai scripts and coprocessors
+
+Hooking into the `RouterService` layer with either [Rhai scripts](/graphos/routing/customization/rhai) or [coprocessors](/graphos/routing/customization/coprocessor) gives you access to the full HTTP request and response before processing occurs. You can use either Rhai scripts or coprocessors to add custom logic for what to log and when.
```
Meschreiber, line 57: @smyrick , the `before processing occurs` in this line confuses me. What processing are we talking about?

Meschreiber:
Beyond some copyedits/restructuring this is largely the same as the original PR. LMK if it looks good to you @smyrick !

docs/source/routing/observability/debugging-client-requests.mdx
```diff
@@ -0,0 +1,82 @@
+---
+title: Debugging Client Requests to GraphOS Router
+subtitle: Options for analyzing incoming requests to help debug problematic requests
```
shorgi, line 3:
```suggestion
subtitle: Options for analyzing and debugging incoming requests
```

docs/source/routing/observability/debugging-client-requests.mdx
smyrick: The `RouterService` external hook is called before the Router does any GraphQL parsing, Query Planning, or other features like Auth checks, so this is the first place you can log out the exact request before anything could error out

smyrick:
I responded to the comment. This look good to me. Not sure if we need a `.changeset`

smyrick (approved):

docs/source/routing/observability/debugging-client-requests.mdx
```diff
@@ -0,0 +1,82 @@
...
+description: Learn how to use GraphOS router telemetry and GraphOS Insights to inspect and debug incoming HTTP client requests.
+---
+
+By default, the GraphOS Router doesn't generate logs for incoming HTTP requests nor export other telemetry metrics. The high request volume your router likely processes can generate significant telemetry data, increasing observability costs. For this reason, you can [configure](/graphos/routing/observability/telemetry) exactly the telemetry you want your router to collect and export.
```
shorgi, line 7: Copyedited and moved this paragraph into the router-specific section

```suggestion
Learn to use GraphOS Insights and/or configure GraphOS Router to analyze and debug client requests. 
```

docs/source/routing/observability/debugging-client-requests.mdx
Meschreiber: Gotcha, the "response" part of `gives you access to the full HTTP request and _response_ before processing occurs` was what was giving me pause. Will just remove that.

```suggestion
Hooking into the `RouterService` layer with either [Rhai scripts](/graphos/routing/customization/rhai) or [coprocessors](/graphos/routing/customization/coprocessor) gives you access to the full HTTP request before processing occurs. You can use either Rhai scripts or coprocessors to add custom logic for what to log and when.
```

Meschreiber:
Thanks @smyrick , since this targets`main`, I don't think I changeset would be picked up in any case. 

@shorgi, when you take a look at this, could you confirm a docs change of this size (adding a whole new page) can skate by without a changeset? TIA

docs/source/routing/observability/debugging-client-requests.mdx
```diff
@@ -0,0 +1,82 @@
...
+
+<Tip>
+
+[Learn how to ensure client IDs are included in all requests.](/graphos/routing/observability/client-id-enforcement)
+
+</Tip>
+
+## Enabling additional telemetry
+
+Use the methods below if you need information outside of what's presented in GraphOS Studio to debug client requests.
```
shorgi, line 21:
```suggestion
To get data outside of GraphOS Studio, you can configure the GraphOS Router to generate logs and export telemetry metrics about incoming HTTP client requests. Because the likely high volume of client requests can generate significant telemetry data and incur steep observability costs, the router supports [configuration](/graphos/routing/observability/telemetry) of the exact telemetry you want your router to collect and export. 
```

shorgi:
> @shorgi, when you take a look at this, could you confirm a docs change of this size (adding a whole new page) can skate by without a changeset? TIA

We've been inconsistent on changesets for docs 😅. A new page though usually gets one.

Other issue is that resolving docs conflicts between `main` and `dev` aren't always smooth. @BrynCooke should we retarget this docs change to `dev` first?

docs/source/routing/observability/debugging-client-requests.mdx
Meschreiber: Since this page is in the Routing section, I think it make sense to open with this.

Meschreiber:
@shorgi , my worry with retargeting to `dev` is that this won't be published until the next router release ([scheduled for Jan 22](https://apollographql.atlassian.net/projects/ROUTER?selectedItem=com.atlassian.jira.jira-projects-plugin%3Arelease-page)

Is there a way to publish this before then but include a changset for `1.60`?

docs/source/routing/observability/debugging-client-requests.mdx
shorgi: This paragraph doesn't mention Studio and its Insights, so the `Using GraphOS Insights` section below feels unexpected. The above edit is meant to include it. 

If keeping this paragraph, I'd still propose some copyedits:

```suggestion
You can configure the GraphOS Router to generate logs and export telemetry metrics about incoming HTTP client requests. Because the likely high volume of client requests can generate significant telemetry data and incur steep observability costs, the router supports [configuration](/graphos/routing/observability/telemetry) of the exact telemetry you want your router to collect and export. 
```

And another option is to make all of `Using GraphOS Insights` into a Tip, as it's all an aside about non-router config.

Meschreiber: > This paragraph doesn't mention Studio and its Insights, so the Using GraphOS Insights section below feels unexpected.

Good point. I don't think of GraphOS Insights as separate from GraphOS Router--rather the router reports directly to insights w/o any additional configuration.

I'll make that clarification and add a mention of Insights here so the next section doesn't feel so jarring.

Meschreiber: https://github.com/apollographql/router/pull/6507/commits/f50662c53d0427842c7f36ddbc1e9a07e8244640

docs/source/routing/observability/debugging-client-requests.mdx
```diff
@@ -0,0 +1,86 @@
...
+### Request and response logging
+
+You can conditionally include the request body in your logs using the [`experimental-when-header`](/graphos/reference/router/telemetry/log-exporters/overview#requestresponse-logging) configuration option.
+
+```yaml title="router.yaml"
+telemetry:
+  exporters:
+    logging:
+      # If one of these headers matches we will log supergraph and subgraphs requests/responses
+      experimental_when_header:
```
BrynCooke, line 48: Let's change this to use the event mechanism. `experimental_when_header` is actually going away https://github.com/apollographql/router/pull/6285

Users should use:
```
telemetry:
  instrumentation:
    spans:
      mode: spec_compliant
    events:
      supergraph:
        "resp":
          level: info
          attributes:
            graphql.document: true
            response:
              response_data: $.*
          on: event_response
          message: "logged request"
```

BrynCooke (changes requested):

docs/source/routing/observability/debugging-client-requests.mdx
Meschreiber: Thanks @BrynCooke , GTK.

I'm a little confused about this config. If the intent is to log the GraphQL operation, would it make more sense to use `telemetry.instrumentation.events.supergraph.request`?

E.g.

```
telemetry:
  instrumentation:
    spans:
      mode: spec_compliant
    events:
      supergraph:
        request:
          level: info
          attributes:
            graphql.document: true
          on: event_request
          message: "logged request"
```

Meschreiber: @BrynCooke , could you take a review this code sample to see if it makes sense?

Meschreiber:
Closing in favor of https://github.com/apollographql/router/pull/6563

docs/source/routing/observability/debugging-client-requests.mdx
Meschreiber: Thanks @devfreddy-apollo . From the [conversation in the open PR](https://github.com/apollographql/router/pull/6563#discussion_r1943611841), do you think this is still worth documenting given it's not conditionally done? (Which I believe is why @smyrick  was documenting it in the first place.)

PR #6509, opened 2025-01-03
Eagerly init subgraph operation for subscription primary nodes

tninesling:
When subgraph operations are deserialized, typically from a query plan cache, they are not automatically parsed into a full document. Each node needs to initialize its operation(s) prior to execution, and the primary node inside SubscriptionNode was missed in the first pass.

This PR also includes a test for demand control scoring. The new test does not reproduce this original issue, but it does provide some basic coverage to ensure scoring handles this operation type.

---

**Checklist**

- Tests added and passing[^3]
    - [X] Unit Tests
    - [ ] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


goto-bus-stop (approved):
Nice! Is it reasonable to also add a test for the case that used to panic?

tninesling:
> Nice! Is it reasonable to also add a test for the case that used to panic?

Hey! Thanks for the review. I'm going to spend some more time today trying to do so, but I spent the better part of an afternoon on it last week and couldn't get it to panic. I figured it would be caused by enabling Subscriptions + Rust QP + QP caching, but that didn't seem to do the trick.

PR #6510, opened 2025-01-06
Priority queue: use crossbeam-channel "biased select"

SimonSapin:
This new functionality in crossbeam allows us to remove a loop that is redundant with a similar loop inside the library.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1863,9 +1863,9 @@ dependencies = [
 
 [[package]]
 name = "crossbeam-channel"
-version = "0.5.13"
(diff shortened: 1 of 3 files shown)


garypen (approved):
I'm approving, but I would prefer it if we had a test that ensured this change was doing the right thing.

SimonSapin:
Is the existing test enough?

https://github.com/apollographql/router/blob/56b1af122600bce45ab69899d43a699ca4d82609/apollo-router/src/ageing_priority_queue.rs#L126-L147

PR #6514, opened 2025-01-06
Connector request service and mapping problem selectors

pubmodmatt:
Initial implementation of a Connector Request Service, which handles individual requests to an Apollo Connector.

The `Request` and `Response` types for this new service contain information about the Connector, as well as how the request or response was mapped, including any mapping problems.

The Connector Service makes one or more calls to the Connector Request Service, and the Connector Request Service in turn makes one call to the underlying connector transport, currently HTTP through the HTTP Client Service.

The transport details are separated out into separate `TransportRequest` and `TransportResponse` enum variants. Currently, there is only an `Http` variant, but this allows for easily expanding to more transport variants in the future without modifying the `Request` and `Response` types.

The Connector Request Service contributes a `connector_request_service` method to the `PluginPrivate` trait. This method is implemented by the telemetry and authentication plugins.

New selectors are added to the telemetry plugin to allow selection of mapping problems from the request and response. 

Existing connectors selectors have been re-implemented at the `connector_request_service` level. This resolves an issue where using the HTTP response selectors would previously cause the data for the last HTTP response to be added to the `connect` span. The response data will now correctly be added to each `connect_request` span.

Standard events are now emitted correctly and contain the `http.request.body` (for request events) and `http.response.body` (for response events).

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-933]: https://apollographql.atlassian.net/browse/ROUTER-933?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apollo-router/src/services/connector_service.rs
```diff
@@ -194,160 +159,107 @@ impl tower::Service<ConnectRequest> for ConnectorService {
...
+        .extensions()
+        .with_lock(|lock| lock.get::<Arc<Mutex<ConnectorContext>>>().cloned());
+
+    let tasks = make_requests(request, &context, connector, debug)
+        .map_err(BoxError::from)?
+        .into_iter()
+        .map(move |request| {
+            // TODO: remove this once we have a source aware query planner. Putting this in the
+            //  context avoids having to expose the internal subgraph name on the Request type.
+            request
```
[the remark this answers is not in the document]
pubmodmatt: I moved this to the `Request` type. Since this is only exposed on `PrivatePlugin` for now, it won't expose the internal subgraph name publicly.

apollo-router/src/services/connector/request_service.rs
```diff
@@ -0,0 +1,297 @@
...
+                    ConnectorRequestService {
+                        http_client_service_factory: self.http_client_service_factory.clone(),
+                    }
+                    .boxed(),
+                    |acc, (_, e)| e.connector_request_service(acc),
+                ),
+            )
+            .boxed()
+    }
+}
```
garypen, line 195: This is the logic in the router that is being changed in the back-pressure branch and so this must not merge to the `next` branch as is.

If you look at the draft PR for back-pressure: https://github.com/apollographql/router/pull/6486 you can see how  we've modified the implementation of `ServiceFactory` so that we only create one instance of the service stack and then use `boxed_clone` to store that on our struct (under an Arc shared Mutex) and then use clone to clone the service stack as required.

The explanation for the change is that the router wasn't observing the various rules of tower and that meant it was pretty much impossible to provide back-pressure under the "create a new set of services for each request" model.

apollo-router/src/services/connector/request_service.rs
```diff
@@ -0,0 +1,297 @@
...
+                let result = match request.transport_request {
+                    TransportRequest::Http(http_request) => {
+                        debug_request = http_request.debug;
+                        if let Some(http_client_service_factory) = http_client_service_factory
+                            .get(&request.service_name)
+                            .cloned()
+                        {
+                            http_client_service_factory
+                                .create(&original_subgraph_name)
+                                .oneshot(crate::services::http::HttpRequest {
```
garypen, line 243: Using oneshot is potentially problematic in a tower service when trying to maintain back-pressure. In this case if we modify the creation so that it returns a cloned service, then we can't use oneshot and should use call.

We need to look at the implementation of `poll_ready()` here and be sure that we are doing the right thing and not creating too many `HttpClients` without respecting backpressure.

garypen (commented):
I think we need to setup a sync up meeting to bring this in line with the back-pressure changes we are making to the router for 2.0.

I don't think it's a huge amount of work, but talking through the required changes and answering questions about why the changes are important will help us make faster progress.

apollo-router/src/services/connector/request_service.rs
pubmodmatt: We had a meeting about this, and the connectors team will commit to making connectors services "backpressure ready" for GA, but after this PR.

apollo-router/src/services/connector/request_service.rs
pubmodmatt: We had a meeting about this, and the connectors team will commit to making connectors services "backpressure ready" for GA, but after this PR.

apollo-router/src/plugins/connectors/handle_responses.rs
```diff
@@ -389,19 +417,31 @@ async fn deserialize_response<T: HttpBody>(
...
+    })?;
+
+    // Here the body is cloned and used to reconstruct the HTTP response. This allows both the
+    // deserialized data and the original response body to be used later, for example to be able
+    // to include the http.response.body on telemetry (which cannot be done here because there
+    // could be a condition on the telemetry using a selector that requires mapping data, which is
+    // not available until after the deserialized response is mapped to GraphQL).
+    let response =
+        http::Response::from_parts(parts.clone(), router::body::from_bytes(body.clone()));
+
```
pubmodmatt, line 434: I'm currently reworking this to avoid the clone here. It will mean that we can't use the mapping selector as a condition on standard events, but that seems like a good trade-off.

apollo-router/src/plugins/telemetry/config_new/connector/selectors.rs
```diff
@@ -39,6 +43,15 @@ impl From<&ConnectorValue> for InstrumentValue<ConnectorSelector> {
         }
     }
 }
+
+#[derive(Deserialize, JsonSchema, Clone, Debug, PartialEq)]
+#[serde(deny_unknown_fields, rename_all = "snake_case")]
+pub(crate) enum MappingProblems {
+    Problems,
+    MaxLevel,
```
pubmodmatt, line 51: I need to rework this a bit, `max_level` currently doesn't work the way I intended. There needs to be a way to do greater than, and since it's a string there isn't.

apollo-router/src/plugins/connectors/tracing.rs
```diff
@@ -117,9 +118,12 @@ mod tests {
             let _factory = ConnectorServiceFactory::new(
                 schema.into(),
                 Arc::default(),
-                Arc::default(),
                 Default::default(),
                 Arc::default(),
+                Arc::new(ConnectorRequestServiceFactory::new(
```
bnjjj, line 123: As we're giving default values maybe it would be worth it to implement defaut only for tests ?

apollo-router/src/plugins/telemetry/config_new/connector/events.rs
```diff
@@ -86,104 +85,30 @@ impl Instrumented
     type EventResponse = ();
 
     fn on_request(&self, request: &Self::Request) {
+        // Any condition on the request is NOT evaluated here. It must be evaluated later when
+        // getting the ConnectorEventRequest from the context. The request context is shared
+        // between all connector requests, so any request could find this ConnectorEventRequest in
+        // the context. Its presence on the context cannot be conditional on an individual request.
```
bnjjj, line 91: Make sure `is_active` in `Selector` trait is well configured then. Because `is_active` is useful to know when we can compute a condition for a specific selector

apollo-router/src/plugins/telemetry/config_new/fixtures/connector/custom_counter_with_conditions/test.yaml
```diff
@@ -1,23 +1,30 @@
...
+  - - connector_request:
+        subgraph_name: users
+        source_name: user_api
+        http_method: GET
+        url_template: "/user/{$this.userid}"
         uri: "/user/1"
-        method: GET
-    - http_response:
+    - connector_response:
+        subgraph_name: users
```
bnjjj, line 10: Just to make sure I understand. In the context of a connector_request, what does `subgraph_name` means ? Because I thought it was not going through the subgraph_service at all once it was a connector.

apollo-router/src/plugins/telemetry/snapshots/apollo_router__plugins__telemetry__fmt_layer__tests__text_logging_with_custom_events_with_instrumented.snap
```diff
@@ -5,7 +5,7 @@ expression: buff.to_string()
 [timestamp] INFO  http.response.body={"foo": "bar"} http.response.body.size=125 my message kind=my_custom_event
 [timestamp] ERROR  Hello from test http.method="GET"
 [timestamp] INFO  http.request.body.size=0 my event message kind=my.request_event
-[timestamp] INFO  http.response.body={"data": {"data": "res"}} http.response.headers={"content-length": "25", "custom-header": "val1", "x-log-request": "log"} http.response.status=200 OK http.response.version=HTTP/1.1  kind=router.response
+[timestamp] INFO  http.response.headers={"content-length": "25", "custom-header": "val1", "x-log-request": "log"} http.response.status=200 OK http.response.version=HTTP/1.1  kind=router.response
```
bnjjj, line 8: Why don't we have the response body anymore on router.response event ?

apollo-router/src/services/connector/request_service.rs
```diff
@@ -0,0 +1,409 @@
...
+
+/// Request type for a single connector request
+#[derive(Debug)]
+#[non_exhaustive]
+pub(crate) struct Request {
+    /// The request context
+    pub(crate) context: Context,
+
+    /// The connector associated with this request
+    // TODO: if this service moves into the public API, consider whether this exposes too much
```
bnjjj, line 63: Is it still a TODO or just info ?

apollo-router/src/services/connector/request_service.rs
```diff
@@ -0,0 +1,409 @@
...
+            parts,
+            router::body::from_bytes(body_bytes),
+        ))
+    } else {
+        Ok(request)
+    }
+}
+
+/// Replace the internal subgraph name in an error with the connector label
+fn replace_subgraph_name(err: BoxError, connector: &Connector) -> BoxError {
```
bnjjj, line 393: I'm wondering if we could try to force it using a specific type to make sure wedon't forget to use this ?

apollo-router/src/plugins/connectors/http_json_transport.rs
```diff
@@ -113,13 +114,19 @@ pub(crate) fn make_request(
...
         }
     });
 
-    Ok((request, debug_request))
+    Ok((
+        TransportRequest::Http(HttpRequest {
+            inner: request,
+            debug: debug_request,
+        }),
+        aggregate_apply_to_errors(&apply_to_errors),
```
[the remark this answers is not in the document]
pubmodmatt: Good call, I've moved this method call earlier and just store `Problem` directly in the debug code, so we can only do this once. There still needs to be a clone of the problems if debug is enabled, though.

apollo-router/src/plugins/connectors/plugin/debug.rs
```diff
@@ -123,7 +122,7 @@ struct ConnectorDebugSelection {
     source: String,
     transformed: String,
     result: Option<serde_json_bytes::Value>,
-    errors: Vec<serde_json_bytes::Value>,
+    errors: Vec<Problem>,
```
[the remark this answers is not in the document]
pubmodmatt: I prefer calling these problems instead of errors, as errors implies failure, and these do not fail the request. Later, we may provide a way for customers to customize what causes failure. The Connectors Debugger UI still calls these errors, but that may be something to consider changing.

apollo-router/src/plugins/connectors/tracing.rs
pubmodmatt: Good call, there was already an `empty` method defined in `test`, but it wasn't creating the instrument. I fixed that and now use the method.

apollo-router/src/plugins/telemetry/config_new/connector/events.rs
pubmodmatt: Good catch, I forgot to update that method. Done in the next commit.

apollo-router/src/plugins/telemetry/config_new/fixtures/connector/custom_counter_with_conditions/test.yaml
pubmodmatt: Connectors are still defined in a subgraph, but they don't execute as a subgraph at runtime.

apollo-router/src/plugins/telemetry/snapshots/apollo_router__plugins__telemetry__fmt_layer__tests__text_logging_with_custom_events_with_instrumented.snap
pubmodmatt: Good catch, this was a mistake I made in the test. Fixed.

apollo-router/src/services/connector/request_service.rs
pubmodmatt: This should be a temporary situation until we have the source-aware query planner. Until then, we have this issue in many places where we have to avoid exposing the internal subgraph name. We could probably have handled that better, but this would require too many changes at this point, and should be temporary.

lennyburdette (approved):
LGTM after a synchronous review! just needs to land on dev, not next :grin:

PR #6516, opened 2025-01-07
fix: include mutation root type definition in connectors

lennyburdette:
this fixes a query planning panic when using `@defer` within a mutation.

the root issue is that we never added `schema { mutation: Mutation }` when necessary. you'd think that would be a bigger problem, but apparently the query planner doesn't mind a missing root type definition. if `Mutation` is defined, it'll use it — this is probably a bug!

however, some part of defer bookkeeping cares about the root type definition, so when it's missing, it wasn't able to populate a map of deferred information. this lead to a panic in this very specific case.

in addition, i fixed a TODO about using the router's api schema config within the expansion code.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [x] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/plugins/connectors/testdata/mutation.yaml
+++ b/apollo-router/src/plugins/connectors/testdata/mutation.yaml
@@ -4,7 +4,7 @@ subgraphs:
     schema:
       sdl: |
         extend schema
-          @link(url: "https://specs.apollo.dev/federation/v2.7")
+          @link(url: "https://specs.apollo.dev/federation/v2.10")
           @link(
             url: "https://specs.apollo.dev/connect/v0.1"
             import: ["@connect", "@source"]
@@ -19,15 +19,23 @@ subgraphs:
         type User {
           id: ID!
           name: String
+          email: String
         }
 
         type Query {
           users: [User]
             @connect(source: "json", http: { GET: "/users" }, selection: "id name")
+          user(id: ID!): User
+            @connect(
+              source: "json"
+              http: { GET: "/users/{$$args.id}" }
+              selection: "id name email"
+              entity: true
+            )
         }
 
         type Mutation {
-          createUser(name: String!): User
+          createUser(name: String!): CreateUserPayload!
             @connect(
               source: "json"
               http: {
@@ -37,8 +45,16 @@ subgraphs:
                 """
               }
               selection: """
-                id
-                name: username
+                success: $(true)
+                user: {
+                  id
+                  name: username
+                }
               """
             )
+        }
+
+        type CreateUserPayload {
+          success: Boolean!
+          user: User!
         }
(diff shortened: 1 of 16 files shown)


apollo-router/tests/samples/enterprise/connectors-defer/README.md
```diff
@@ -0,0 +1 @@
+Example testing the connectors feature
```
pubmodmatt, line 1: NIt that this should describe the case being tested

apollo-router/tests/samples/enterprise/connectors-defer/supergraph.graphql
```diff
@@ -0,0 +1,83 @@
+schema
+  @link(url: "https://specs.apollo.dev/link/v1.0")
+  @link(url: "https://specs.apollo.dev/join/v0.5", for: EXECUTION)
+  @link(url: "https://specs.apollo.dev/connect/v0.1", for: EXECUTION)
+  @join__directive(graphs: [ONE], name: "link", args: {url: "https://specs.apollo.dev/connect/v0.1", import: ["@connect", "@source"]})
+  @join__directive(graphs: [ONE], name: "source", args: {name: "test", http: {baseURL: "http://localhost:4007/"}})
+{
+  query: Query
+  mutation: Mutation
```
pubmodmatt, line 9: It may be good to have a test schema that changes the root operation type names, to make sure nothing in expansion depends on the default names:

```graphql
schema {
  query: SomeTypeName
  mutation: SomeOtherName
}
```

lennyburdette: federated composition normalizes the names so it's not really important, but i can add a test anyway

lennyburdette: lol there's a separate validation bug that crops up when you try to rename query and compose. i'll leave this for now and file an issue

lennyburdette: https://apollographql.atlassian.net/browse/CNN-548

pubmodmatt (approved):

apollo-router/src/plugins/connectors/testdata/mutation.yaml
```diff
@@ -19,15 +19,23 @@ subgraphs:
...
+          email: String
         }
 
         type Query {
           users: [User]
             @connect(source: "json", http: { GET: "/users" }, selection: "id name")
+          user(id: ID!): User
+            @connect(
+              source: "json"
+              http: { GET: "/users/{$$args.id}" }
```
smyrick, line 31: Typo??

PR #6518, opened 2025-01-07
connectors debugging snapshot test

lennyburdette:
because the debugger captures the host/port of the http API, i had to add the ability to specify a port. this could be a problem if another test chooses the same port, but that's pretty unlikely given the number of snapshot tests we have.

also, you have to use the subgraph name "connectors" for snapshot tests to work.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -510,7 +510,7 @@ commands:
             # TODO: remove this workaround once we update to Xcode >= 15.1.0
             # See: https://github.com/apollographql/router/pull/5462
             RUST_LIB_BACKTRACE: 0
-          command: xtask test --workspace --locked --features ci
+          command: xtask test --workspace --locked --features ci,snapshot
       - run:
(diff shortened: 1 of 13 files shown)


pubmodmatt (approved):

nicholascioli:
This looks good! As a warning, the router's test harness starts up after all subgraphs get started, so tests might fail for somewhat unrelated reasons if a specific test reserves the exact port given to the router (which is also somewhat randomly assigned).

See the previous discussion on this for more context: #4722

nicholascioli:
There was talk of modifying the test harness to use sockets instead, but testing the networking aspect is part of why the test harness exists in the first place, so I'm not sure. @abernix might have some better ideas.

PR #6523, opened 2025-01-08
Add experimental non-reloadable config for worker pool size

goto-bus-stop:
1.58.0 introduced a thread pool for compute-heavy work like GraphQL
validation and query planning: https://github.com/apollographql/router/blob/bd12e44f0e47b0e5e31a75494a6336165c8c91e6/CHANGELOG.md#move-heavy-computation-to-a-thread-pool-with-a-priority-queue-pr-6247

The worker pool uses all available parallelism. Some customers are
reporting concerns that this leads to CPU saturation when a lot of
queries need to be planned, causing other requests to time out.

This branch exists to let select customers experiment with the worker
pool size, and gather data so we can decide on a better default or
introduce useful configuration in a future release.

> [!NOTE]
> This branch is based on Router v1.58.1, *not* the latest release, to
allow customers who are on the 1.58.x line to try it out without pulling
in additional changes from 1.59.x.

This introduces a new configuration option:

```yaml
experimental_worker_pool:
  size: 2 # number of threads
```

To avoid significantly refactoring the worker pool implementation,
this configuration is *not* runtime-reloadable, so users must restart
their routers (eg. recycle pods) for changes to take effect.

If we want to release this feature properly, we should do the
refactoring to make the worker pool resizeable at runtime.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [x] Manual Tests


apollo-router/src/compute_job.rs
```diff
@@ -14,13 +15,28 @@ use crate::metrics::meter_provider;
...
 
+static CONFIGURED_POOL_SIZE: AtomicUsize = AtomicUsize::new(0);
+
+pub(crate) fn experimental_set_thread_pool_size(size: usize) {
+    CONFIGURED_POOL_SIZE.store(size, std::sync::atomic::Ordering::Release);
+}
+
 /// Let this thread pool use all available resources if it can.
 /// In the worst case, we’ll have moderate context switching cost
 /// as the kernel’s scheduler distributes time to it or Tokio or other threads.
```
bgmerrell, line 26: Did you want to update this comment in this PR?

bgmerrell: I tested this manually and it appears to be working as intended. Thanks!

goto-bus-stop: Ha, good spot :) Thanks for testing it out, and it's great to know that it has the intended effect. I neglected to update code comments and the like.
We'll need to figure out if we want to put this configuration option in a release (with testing & docs & a resizeable pool to support hot reload) or if we want to do something else (like using a non-configurable formula to set a sensible max pool size) sometime in the next 10 days or so.

PR #6525, opened 2025-01-08
connnectors: feature warnings on incompatible plugins

nicholascioli:
This commit adds a few informative warnings when using connectors with certain plugins which are currently incompatible. These warnings should better inform router operators of certain features which aren't yet supported when used in tandem with connectors.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [x] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[CNN-343]: https://apollographql.atlassian.net/browse/CNN-343?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apollo-router/tests/integration/connectors.rs
```diff
@@ -0,0 +1,41 @@
...
+        .build()
+        .await;
+
+    router.start().await;
+
+    // Make sure that we have the warnings we expect
+    let plugins = ["coprocessor", "headers", "telemetry", "traffic_shaping"];
+    for plugin in plugins {
+        let msg = format!("plugin `{plugin}` is enabled for connector-enabled subgraphs, which is not yet supported");
+        router.assert_log_contains(&msg).await;
```
lennyburdette, line 37: can we test for subgraph names as well to avoid an regressions in the internal name conversion?

apollo-router/src/plugins/connectors/configuration.rs
```diff
@@ -96,3 +105,102 @@ pub(crate) fn apply_config(config: &Configuration, mut connectors: Connectors) -
...
+                .copied()
+                .collect()
+        } else {
+            &HashSet::new()
+        };
+
+        if !incompatible_subgraphs.is_empty() {
+            tracing::warn!(
+                subgraphs = incompatible_subgraphs.iter().join(","),
+                "plugin `{plugin_name}` is enabled for connector-enabled subgraphs, which is not yet supported. See https://go.apollo.dev/INSERT_DOCS_LINK for more info"
```
lennyburdette, line 189: "which is not yet supported" — in some cases, we might never support a feature with connectors. i wonder if "which is currently unsupported" is slightly better. "yet" kinda implies that it will be some day

apollo-router/src/plugins/connectors/configuration.rs
```diff
@@ -96,3 +105,242 @@ pub(crate) fn apply_config(config: &Configuration, mut connectors: Connectors) -
...
+                subgraphs = incompatible.iter().join(","),
+                message = msg(plugin)
+            );
+        }
+    }
+
+    // There are a few plugins which influence all subgraphs, regardless
+    // of configuration, so we warn about these statically here if we have
+    // any connector-enabled subgraphs.
+    let incompatible_plugins = ["demand_control", "rhai"];
```
pubmodmatt, line 337: It seems too broad to warn if the customer is using Rhai at all. It's not really incompatible - the only issue is that there is no way for them to hook into the connectors requests, which isn't something we need to warn about since they can't do it. They might mistakenly think that the subgraph lifecycle hook will be executed for a subgraph containing connectors, when it won't. But the other lifecycle hooks are still executed, and in fact we suggest that they use these for auth. I think warning here would be more confusing than anything.

My bar for a warning is that there should be a way for the customer to resolve the issue, or at least suppress it. Here, there isn't - if they need to use Rhai for auth or any other reason, they'd see this warning always.

apollo-router/tests/fixtures/connectors/incompatible.router.yaml
```diff
@@ -0,0 +1,71 @@
...
+  subgraph:
+    all:
+      enabled: false
+    subgraphs:
+      connectors:
+        enabled: true
+coprocessor:
+  url: http://127.0.0.1:8081
+  subgraph:
+    all:
```
pubmodmatt, line 32: The `all` case concerns me. I mentioned in another comment that for any warning we give, the customer should have a way to resolve the issue. The only way to resolve this would be to replace `all` with the same config copy and pasted for every single subgraph that doesn't contain connectors. In large supergraphs, that could be very unwieldy. And similar to the Rhai case, there really isn't an incompatibility with coprocessors, there just isn't a way to execute a coprocessor at the connector level - it still works for router, supergraph, subgraph, etc.

apollo-router/tests/fixtures/connectors/incompatible.router.yaml
```diff
@@ -0,0 +1,71 @@
...
+      aws_sig_v4:
+        default_chain:
+          profile_name: "my-test-profile"
+          region: "us-east-1"
+          service_name: "lambda"
+          assume_role:
+            role_arn: "test-arn"
+            session_name: "test-session"
+            external_id: "test-id"
```
pubmodmatt, line 19: We actually do have full support for this with connectors. It just needs to be configured at the connector source level:

```yaml
authentication:
  connector:
    sources:
      aws.lambda:
        aws_sig_v4:
          default_chain:
            profile_name: "default"
            region: "us-east-1"
            service_name: "lambda"
            assume_role:
              role_arn: "arn:aws:iam::396913720934:role/helloexecute"
              session_name: "graphql"
```

Maybe if they have authentication settings on `subgraph` but don't have a `connector` setting, there could be an INFO level message pointing them to how to configure it?

The ideal solution would be to detect if any connector base URL points to AWS, and warn if they haven't configured the corresponding source with `authentication`.

lennyburdette: Does `authentication.subgraph.all` apply to connector sources as well?

pubmodmatt (approved):

PR #6527, opened 2025-01-09
Add `schema_id`, `config_hash` attributes to the `apollo_router_session_count_total` metric

goto-bus-stop:
This adds attributes to the `apollo_router_session_count_total` metric, so customers can see if they have sessions that are keeping alive references to large amounts of previous router state after a configuration or schema change. This would help diagnose potential memory usage issues and inform our approach to dealing with that.

To do this the `session_count_total` instrument is refactored a bit; it was previously using a single global state value, now there can be multiple total session counts with different attributes alive at the same time.

For the `schema_id`, I just had to thread it through several layers of functions. For the config, I "invented" a new config hash (sha256 of JSON-serialized value).

This also fixes a bug that was introduced in https://github.com/apollographql/router/commit/797a3be1b545d4f5a494e577fc2fb27bd65bdddd. In that commit the total session count is changed from a manual counter to a guard, but the guard was not moved into the tokio task properly, so it would immediately decrement. That change has not made it to a release luckily.

I'll mark this as a draft to add some tests for the metric to validate both this PR and the basic behaviour.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-952]: https://apollographql.atlassian.net/browse/ROUTER-952?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


goto-bus-stop:
Interestingly this metric no longer exists in router 2.0. Instead there is a `http.server.active_requests` metric that we could likely attach these attributes to.

goto-bus-stop:
I'm going to redo this after fixing the `apollo_router_session_count_total` metric.

PR #6528, opened 2025-01-09
Update randomization of query prewarming order to also apply to persisted queries

andrewmcgivery:
Update randomization of query prewarming order to also apply to persisted queries

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


goto-bus-stop:
I think it makes sense, but the comment on persisted queries says that they are done *first* so they end up being the *least recently used* queries in the cache on startup. The thinking is that queries that are pre-warmed because they were the *most recently used* in the old cache are more likely to be needed  again than persisted queries are.

So it sounds to me like we should shuffle PQs and queries from the previous cache separately, and then concatenate the two shuffled lists.

BrynCooke (approved):

BrynCooke:
We're going to have to move this PR into the router repo.

BrynCooke:
Closing as an internal PR has been created
https://github.com/apollographql/router/pull/6669

PR #6530, opened 2025-01-09
Upgrade to axum 0.8

tninesling:
Upgrades `axum` to 0.8.1, along with related dependencies `axum-extra` and `tokio-tungstenite`.

The main cosmetic change is that websocket messages are created using `::text` instead of `::Text` because the latter now expects `Utf8Bytes` instead of `String`. The smaller but more important change is that the underlying `matchit` crate made a breaking change which is leaking out of the `axum` API. This changes how bound parameters and wildcards are matched in route strings.

Previously, named parameters were prefixed with `:`, but now they are wrapped in braces (ie. `/foo/:my_param` -> `/foo/{my_param}`). Wildcards are also wrapped in braces (ie. `/foo/*my_wildcard` -> `/foo/{*my_wildcard}`).

We could let this format leak as well and change our documentation, but for now I'm using regex to wrap the affected parameters in the required braces before passing the strings into `axum`. 

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/axum_factory/tests.rs
+++ b/apollo-router/src/axum_factory/tests.rs
@@ -713,7 +713,7 @@ async fn response_with_root_wildcard() -> Result<(), ApolloRouterError> {
     let conf = Configuration::fake_builder()
         .supergraph(
             crate::configuration::Supergraph::fake_builder()
-                .path(String::from("/*"))
+                .path(String::from("/{*rest}"))
                 .build(),
         )
         .build()
@@ -863,7 +863,7 @@ async fn response_with_custom_prefix_endpoint() -> Result<(), ApolloRouterError>
     let conf = Configuration::fake_builder()
         .supergraph(
             crate::configuration::Supergraph::fake_builder()
-                .path(String::from("/:my_prefix/graphql"))
+                .path(String::from("/{my_prefix}/graphql"))
                 .build(),
         )
         .build()
@@ -928,7 +928,7 @@ async fn response_with_custom_endpoint_wildcard() -> Result<(), ApolloRouterErro
     let conf = Configuration::fake_builder()
         .supergraph(
             crate::configuration::Supergraph::fake_builder()
-                .path(String::from("/graphql/*"))
+                .path(String::from("/graphql/{*rest}"))
                 .build(),
         )
         .build()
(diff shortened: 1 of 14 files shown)


garypen (commented):
Great work Taylor!

One of the motivations to get this in before 2.0 releases is so that we could include the incompatible change. Exposing the change from axum is fine and I don't think it's a goal for us to maintain backwards compatibility here with router 1.x.

We are maintaining a MIGRATION guide and putting those details in there is the right thing to do: docs/source/reference/migration/from-router-v1.mdx

tninesling:
> Great work Taylor!
> 
> One of the motivations to get this in before 2.0 releases is so that we could include the incompatible change. Exposing the change from axum is fine and I don't think it's a goal for us to maintain backwards compatibility here with router 1.x.
> 
> We are maintaining a MIGRATION guide and putting those details in there is the right thing to do: docs/source/reference/migration/from-router-v1.mdx

Thanks, @garypen. I've updated the relevant tests and docs to use the new axum format.

apollo-router/src/axum_factory/tests.rs
```diff
@@ -713,7 +713,7 @@ async fn response_with_root_wildcard() -> Result<(), ApolloRouterError> {
     let conf = Configuration::fake_builder()
         .supergraph(
             crate::configuration::Supergraph::fake_builder()
-                .path(String::from("/*"))
+                .path(String::from("/{*rest}"))
```
garypen, line 716: NIT:
I'm a fan of eliminating unnecessary variations. It makes grep searches easier and prevents future confusion for maintainers. Here, you have elected to use `*rest` to represent what used to be `*`. Elsewhere you used `*anything`. I think `*rest` is a better name and if there's no difference between what these labels represent, then you should standardise on one of them.

garypen (approved):
Looks great, I particularly like the docs additions.

I'm approving, but please consider my comment before you merge.

apollo-router/src/axum_factory/tests.rs
tninesling: I agree. Updated to use `*rest`, thanks!

PR #6537, opened 2025-01-13
merge dev into next

garypen:
Merge dev to next.

This was a particularly tricky merge, so careful reviews are advised.


goto-bus-stop (approved):
it all looks right to the best of my knowledge

PR #6539, opened 2025-01-13
fix(router, only 1.x): fix the `apollo_router_session_count_total` metric

goto-bus-stop:
In #6495 I migrated this metric to an otel ObservableGauge. However the lifetime of the count guard was too short, so the value of the metric would almost always just be 0; and the lifetime of the instrument was too short, so even the wrong value of 0 would not actually be reported. This change was not part of a release so there is no user impact as long as we land this :)

Now the lifetime of the instrument is captured by all requests, so the instrument is kept even post-reload if a request is still being completed from the old schema.

Also, a test verifies that the count increments and decrements as expected.

This does **not need to be ported to Router 2.0**.

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [x] Manual Tests


tninesling (approved):
LGTM! Does this need a changeset, or are you reusing one that's already on dev?

goto-bus-stop:
Since it's just a fix to a bug that has never reached a release, I think it doesn't justify a changeset!

PR #6540, opened 2025-01-13
POC: Make supergraph creator clone a BoxCloneService

garypen:
This breaks all the http_post_mutation tests because of changes in expectation.

It's an interesting experiment though and worth discussing.


IvanGoncharov (approved):
Looks good, the only thing I'm worried is `buffer`.
I think it's ok in this particular PR as incremental change but I think we need to have a separate task to remove all `buffer` calls (outside of tests) that we added during this refactoring

PR #6541, opened 2025-01-13
Add decompression failure tests

tninesling:
Adds tests to record the behavior of the decompression layer's new type signature.

---

**Checklist**

- Tests added and passing[^3]
    - [X] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/axum_factory/tests.rs
+++ b/apollo-router/src/axum_factory/tests.rs
@@ -575,20 +575,8 @@ async fn it_decompress_request_body() -> Result<(), ApolloRouterError> {
 async fn unsupported_compression() -> Result<(), ApolloRouterError> {
     let original_body = json!({ "query": "query { me { name } }" });
     let compressed_body = gzip(original_body).await;
-    let expected_response = graphql::Response::builder()
-        .data(json!({"response": ""}))
-        .build();
-    let router_service = router::service::from_supergraph_mock_callback(move |req| {
-        assert_eq!(
-            req.supergraph_request.into_body().query.unwrap(),
-            "query { me { name } }"
-        );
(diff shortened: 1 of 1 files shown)


apollo-router/src/axum_factory/tests.rs
```diff
@@ -567,6 +571,78 @@ async fn it_decompress_request_body() -> Result<(), ApolloRouterError> {
...
 
+#[tokio::test]
+async fn unsupported_compression() -> Result<(), ApolloRouterError> {
+    let original_body = json!({ "query": "query { me { name } }" });
+    let compressed_body = gzip(original_body).await;
+    let expected_response = graphql::Response::builder()
+        .data(json!({"response": ""}))
+        .build();
+    let router_service = router::service::from_supergraph_mock_callback(move |req| {
+        assert_eq!(
```
goto-bus-stop, line 582: This assertion is not actually executed, right? The test asserts that the request is rejected *before* this code is reached. Maybe the body of this lambda should be replaced by an `unreachable!()` statement? The same goes for the `mismatched_compression_header` test.

goto-bus-stop (changes requested):
One question :)

apollo-router/src/axum_factory/tests.rs
tninesling: Good point. Since we're never reaching the callback, I just replace that whole block with `router::service::empty()` in [7706100](https://github.com/apollographql/router/pull/6541/commits/7706100ca2acaf979982f886282bbf8c324fec07). Looks much cleaner, thanks!

PR #6543, opened 2025-01-14
fix directive conflicts in cache control header

### LINKED ISSUE #6441: Entity Caching - Router can propagate unnecessary cache-control directives to client
## Describe the bug
I encountered a situation where someone has a custom Java library that intercepts the response from a subgraph before it goes out to the router and inserts a cache-control header into that response.

The problem is:
- With entity caching enabled it seems that it allows subgraphs to insert additional directives programmatically, but the router does not filter out unnecessary values, which is a confusing experience for clients.

The router does disallow invalid values and there does seem to be some logic to validate the directives present (ex. when passing `no-cache` it changes it in the final response to `no_cache`)

## To Reproduce
1. I've created a reproduction subgraph here, you'll need to clone it and run through the README: https://github.com/FineAndDanD/springboot-dgs-content-cache-repro
2. I've also created a graph in Studio. I didn't want to expose the API key so please feel free to reach out if you want to avoid creating your own graph. Since this is an enterprise feature a graph ref and API key will be needed.
3. Ensure you have a cache running with the basic entity caching router config from here: https://www.apollographql.com/docs/graphos/routing/performance/caching/entity#configure-router-for-entity-caching
4. Start the subgraph, the cache and the router and send a request! I've added a screenshot of what you should see in the headers.

## Expected behavior
The expected behaviour is that the router will filter out unnecessary values from the `cache-control` header when the request resolves.  So if there's `max-age=10, no-cache, must-revalidate, no-store`, the expected value for the `cache-control` header would simply be `no-store`. Please see the MDN docs for justification of this reasoning: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#preventing_storing

## Output
Screenshot 2024-12-11 at 2 06 34 PM

## Desktop (please complete the following information):
 - OS: [e.g. iOS] MacOS
 - Version [e.g. 22] Sonoma 14.7

## Additional context
N/A

bnjjj:
Unnnecessary cache-control directives are created in cache-control header.  The router will now filter out unnecessary values from the `cache-control` header when the request resolves. So if there's `max-age=10, no-cache, must-revalidate, no-store`, the expected value for the cache-control header would simply be `no-store`. Please see the MDN docs for justification of this reasoning: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#preventing_storing

Fixes #6441

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-929]: https://apollographql.atlassian.net/browse/ROUTER-929?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apollo-router/src/plugins/cache/cache_control.rs
```diff
@@ -178,6 +178,16 @@ impl CacheControl {
         let mut s = String::new();
         let mut prev = false;
         let now = now_epoch_seconds();
+        if self.no_store {
+            write!(&mut s, "no-store")?;
+            // Early return to avoid conflicts https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#preventing_storing
+            return Ok(s);
+        }
+        if self.no_cache {
+            write!(&mut s, "no_cache")?;
```
IvanGoncharov, line 187:
```suggestion
            write!(&mut s, "no-cache")?;
```
I think it should have dash, or I missing something?

IvanGoncharov: Here it has dash https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#no-cache_2

bnjjj: Oh wow ! Good catch !

IvanGoncharov (changes requested):
I think we should consider moving https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#preventing_storing
logic here:
https://github.com/apollographql/router/blob/f92329847d7e8859ba781dbdb7c72ee35b907c6d/apollo-router/src/plugins/cache/cache_control.rs#L268

Because:
> In theory, if directives are conflicted, the most restrictive directive should be honored.

So we should handle other conflicting directives like: `max-age=0, max-age=60`
And already have this code here:
https://github.com/apollographql/router/blob/f92329847d7e8859ba781dbdb7c72ee35b907c6d/apollo-router/src/plugins/cache/cache_control.rs#L275-L278

So I think the same function should also have logic that `no-cache` and `no-store` wipe out other directives.
I think it would be better solution since we have some logic around `CacheControl` and also store it in context that means we apply `no-cache` and `no-store` more consistently across the plugin.

Because in current version we apply `no-cache` and `no-store` only before we send response back to a client.

IvanGoncharov (approved):

IvanGoncharov:
@bnjjj I think we should have changeset entry for it.

IvanGoncharov:
@Mergifyio backport 1.x

PR #6544, opened 2025-01-14
chore(deps): Update dev dependencies with `npm audit fix`

abernix:
This updates several non-load-bearing "dev dependencies" that we use in our testing framework to resolve security warnings that appear when running `npm audit`.  Due to the nature of the packages' use inside of this reposoitory, none of these updates respresents a risk to users of the router: they are
not included in its compiled distribution nor used at runtime.


goto-bus-stop (approved):
All the updates are express-related packages, and es-shims in one case, those ecosystems are both super robust so we can be pretty confident

PR #6545, opened 2025-01-15
Point developers to `.with_metrics()` in metrics assertions

goto-bus-stop:
It's easy to forget `.with_metrics()` in an async metrics test, and you have to read dev-docs/metrics.md to know about this.

So let's add the link to dev-docs/metrics.md in the doc comments for the metrics assertion macros, and also call it out in the panic message that you get if you forget to do this.

### DIFF
--- a/apollo-router/src/metrics/mod.rs
+++ b/apollo-router/src/metrics/mod.rs
@@ -149,7 +149,9 @@ pub(crate) mod test_utils {
     pub(crate) fn collect_metrics() -> Metrics {
         let mut metrics = Metrics::default();
         let (_, reader) = meter_provider_and_readers();
-        reader.collect(&mut metrics.resource_metrics).unwrap();
+        reader
+            .collect(&mut metrics.resource_metrics)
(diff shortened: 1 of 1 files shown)


apollo-router/src/metrics/mod.rs
```diff
@@ -149,7 +149,9 @@ pub(crate) mod test_utils {
     pub(crate) fn collect_metrics() -> Metrics {
         let mut metrics = Metrics::default();
         let (_, reader) = meter_provider_and_readers();
-        reader.collect(&mut metrics.resource_metrics).unwrap();
+        reader
+            .collect(&mut metrics.resource_metrics)
+            .expect("Failed to collect metrics. Did you forget to use `async{}.with_metrics()`? See dev-docs/metrics.md");
```
goto-bus-stop, line 154: Since this is in the test_utils module and router users never see it, it's fine to have a developer-facing message.

bnjjj (approved):

PR #6553, opened 2025-01-15
update CircleCI base image

nicholascioli:
This commit fixes an issue with a downstream dependency (`aws-lc-sys`) caused by using an old version of GCC by updating the base CircleCI image to `current`.

This was realistically a long-time coming because the previous tag of `stable` was deprecated in favor of `current` back in 2022.

Below is a quote from the [docker hub page](https://hub.docker.com/r/cimg/base) explaining the change to the tag names:

> current - This image tag points to the latest, production ready base image. This is a replacement for the old stable tag. This image should be used by projects that want a decent level of stability but would like to get occasional software updates. It is typically updated once a month.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -237,6 +237,7 @@ commands:
                 command: |
                   if [[ ! -d "$HOME/.deb" ]]; then
                     mkdir $HOME/.deb
+                    sudo apt-get update
                     sudo apt-get --download-only -o Dir::Cache="$HOME/.deb" -o Dir::Cache::archives="$HOME/.deb" install libssl-dev libdw-dev
                   fi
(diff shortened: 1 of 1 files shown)


nicholascioli:
A required dependency was not found during setup, so I've added a `sudo apt-get update` to the setup instructions to repopulate the package cache. This does add some non-determinism to the build process, since packages can update between invocations of CircleCI, but it should be mitigated by the fact that only security and patch updates are pushed to LTS versions of Ubuntu.

lennyburdette (approved):

PR #6559, opened 2025-01-16
replace traceid::maybe_new with traceid::current

tapaderster:
Irrespective of whether the span is sampled, we should be able to get a trace id. 
When using TraceId::maybe_new() with sampler set to a value < 1.0, returns `None` at times. 
We should be using to TraceId::current instead.
Removing the `maybe_new` method and making the `TraceId::currrent()` method public so that it can be used from application code.

----

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


tapaderster:
should i make it a breaking change since I am removing a public method?

BrynCooke:
@tapaderster Thanks for this PR. I've taken a look and there's some slightly more extensive work that needs to be done.

I'm going to deprecate TraceID and remove all internal use. Users can use the Otel and Tracing APIs directly if they need this.

I'm also going to move the code to a different branch so that we can get CI to run.

tapaderster:
in that case, I am going to close this pr

PR #6561, opened 2025-01-16
Aaron/bgs 186/license into plugins

aaronArinder:
this is part of the TPS effort (see the router-inbox channel, but also DM for a rough draft of a design doc [not sure what links I should add to the description])

this is just a draft for sanity checking part of that work: getting the license down into the plugin ecosystem so that we can use some soon-to-exist claims in configuring plugins; very wippy, very drafty, but open for comments to help keep me from shooting myself in the foot

some questions I have are how to write tests for this, how to sort out all those FIXMEs (or if cloning is acceptable for License), and if I'm missing anything big

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/uplink/license_enforcement.rs
```diff
@@ -503,20 +503,24 @@ impl Display for LicenseEnforcementReport {
 
 /// License controls availability of certain features of the Router. It must be constructed from a base64 encoded JWT
 /// This API experimental and is subject to change outside of semver.
-#[derive(Debug, Clone, Default)]
+#[derive(Debug, Clone, Default, Eq, PartialEq)]
```
clenfest, line 506: We are doing this because the plugin will need to use the claims and send them somewhere, is that right? There's no additional state explosion that we can do initially?

apollo-router/src/state_machine.rs
```diff
@@ -325,15 +328,18 @@ impl<FA: RouterSuperServiceFactory> State<FA> {
         // Check the license
         let report = LicenseEnforcementReport::build(&configuration, &schema);
 
-        match license {
-            LicenseState::Licensed => {
+        let better_name_license = match license {
```
clenfest, line 331: Yes, better name please :)

apollo-router/src/router_factory.rs
```diff
@@ -483,7 +495,15 @@ pub async fn create_test_service_factory_from_yaml(schema: &str, configuration:
...
     let service = YamlRouterFactory
-        .create(is_telemetry_disabled, Arc::new(config), schema, None, None)
+        .create(
+            is_telemetry_disabled,
+            Arc::new(config),
+            schema,
+            None,
+            None,
+            // FIXME: default? feels wrong; needs fixture
+            LicenseState::default(),
```
nmoutschen, line 505: __Question__: Does this mean we'll have to reload the Router whenever the license state changes?

And additional question: do we reload the router whenever the `LicenseState` changes?

apollo-router/src/router_factory.rs
```diff
@@ -866,6 +894,8 @@ mod test {
                 Arc::new(schema),
                 None,
                 None,
+                // FIXME: fixture?
+                LicenseState::default(),
```
nmoutschen, line 898: __NIT__: for a lot of tests, you can simplify with `Default::default()` to avoid importing `LicenseState` everywhere.

apollo-router/src/uplink/license_enforcement.rs
```diff
@@ -501,22 +501,48 @@ impl Display for LicenseEnforcementReport {
...
+/// Transactions Per Second limits. We talk as though this will be in seconds, but the Duration
+/// here is actually given to us in milliseconds via the License's JWT's claims
+#[derive(Builder, Copy, Clone, Debug, Eq, PartialEq)]
+pub(crate) struct TpsLimit {
+    capacity: usize,
+    duration: Duration,
+}
+
+// FIXME: better defaults set with some producty insight and move then to consts
+impl Default for TpsLimit {
```
nmoutschen, line 520: Does it make sense to have any defaults for this?

apollo-router/src/uplink/license_enforcement.rs
```diff
@@ -501,22 +501,48 @@ impl Display for LicenseEnforcementReport {
...
+    /// Transaction Per Second limits
+    tps: TpsLimit,
+}
+
 /// Licenses are converted into a stream of license states by the expander
 #[derive(Debug, Copy, Clone, Eq, PartialEq, Default, Display)]
 pub(crate) enum LicenseState {
     /// licensed
-    Licensed,
+    Licensed(Option<LicenseLimits>),
```
nmoutschen, line 541: Should we do the opposite instead?

E.g. `LicenseLimits` is always present, but `tps` should be an `Option<TpsLimit>`.

apollo-router/src/router_factory.rs
aaronArinder: yes and yes!

apollo-router/src/router_factory.rs
aaronArinder: good call; will update

apollo-router/src/uplink/license_enforcement.rs
aaronArinder: I think so, but I could be wrong; this seems like a good way to set the product defaults to fall back on if they're not available in the jwt. What do you think? Maybe the LicenseLimits could make TpsLimits optional so that None is the default (which is the case for, say, free tier)

apollo-router/src/uplink/license_enforcement.rs
aaronArinder: ahh, yeah! I replied above and this makes me think that it's the right idea; I'll make TpsLimits optional in LicenseLimits and use _that_ as the default

aaronArinder: updated

apollo-router/src/uplink/license_enforcement.rs
aaronArinder: updated to remove the Default and use Option<TpsLimits> instead

apollo-router/src/uplink/license_enforcement.rs
aaronArinder: talked in slack; updated! new struct, tighter contract, better derives

BrynCooke:
This looks like a good direction to go!

aaronArinder:
superseded by #6598

PR #6566, opened 2025-01-17
Remove the tracing `MetricsLayer`

goto-bus-stop:
Note: This is a layer for tracing-subscriber, not for tower!

MetricsLayer converts tracing events with magic prefixes to otel metrics. We no longer use this mechanism for metrics, so we can delete support for it.
Custom Rust plugins could, in the past, use the `tracing` macros to emit metrics. The migration path will be to use the opentelemetry Rust crates like we do in the router. The new `apollo_router::metrics::meter_provider()` API facilitates this.

When a tracing event is emitted following the old system, it is now not treated specially, so it will end up in the router's log output. Additionally, an error event will be emitted for each tracing event that attempts to use the legacy metrics system.

[ROUTER-297]: https://apollographql.atlassian.net/browse/ROUTER-297?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bnjjj:
I think I would keep the `layer.rs` file for a bit, remove the logics and just add logs inside to warn our users who are still trying to use these kind of metrics.

apollo-router/src/plugins/telemetry/reload.rs
```diff
@@ -185,3 +165,33 @@ where
...
+            field
+                .name()
+                .starts_with(LEGACY_METRIC_PREFIX_MONOTONIC_COUNTER)
+                || field.name().starts_with(LEGACY_METRIC_PREFIX_COUNTER)
+                || field.name().starts_with(LEGACY_METRIC_PREFIX_HISTOGRAM)
+                || field.name().starts_with(LEGACY_METRIC_PREFIX_VALUE)
+        }) {
+            tracing::error!(
+                metric_name = field.name(),
+                "Detected unsupported legacy metrics reporting, remove or migrate to opentelemetry"
```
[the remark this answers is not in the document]
goto-bus-stop: yeah, was going to follow up with that after adding the upgrade guide entry and making sure this error log actually happens (i can't get it to show up yet 🤔 but might just be a skill issue)

docs/source/reference/migration/from-router-v1.mdx
```diff
@@ -411,7 +411,18 @@ For more information about [configuring traffic shaping](../../routing/performan
...
+Previously, Rust plugins could use the Router's internal metrics system using the `tracing` macros. This is no longer supported, and the opentelemetry crates should be used instead.
+
+`tracing` field names starting with these strings are no longer treated specially:
+- `counter.`
+- `histogram.`
+- `monotonic_counter.`
+- `value.`
+
+Instead, use the [OpenTelemetry](https://docs.rs/opentelemetry/latest/opentelemetry/) crates.
+You can use the new `apollo_router::metrics::meter_provider()` API to access the router's global meter provider to register your instruments.
```
garypen, line 425: I think you have renamed this to meter_provider_internal() unless this is referencing something else...

docs/source/routing/customization/native-plugins.mdx
```diff
@@ -246,53 +246,32 @@ plugins:
...
-info!(value.qux = 1);
-info!(value.abc = -1);
-info!(value.def = 1.1);
-
-let caller = "router";
-tracing::info!(
-    histogram.loading_time = loading_time.elapsed().as_secs_f64(),
-    kind = %caller, // Custom attribute for the metrics
-);
+use apollo_router::metrics::meter_provider;
```
garypen, line 265: Same comment as above.

garypen (commented):
I think you renamed the meter_provider() fn to meter_provider_internal() and need to update the docs, but I may have mis-read what is going on.

docs/source/reference/migration/from-router-v1.mdx
goto-bus-stop: I renamed the existing `meter_provider()` function to `meter_provider_internal()`, because it returns a non-public type with custom feature extensions.

There is also a new `fn meter_provider()` that returns an `impl MeterProvider`, so we *only* expose the API of the meter provider type from the `opentelemetry` crate, and not our own implementation details.

garypen (approved):

Issue #4746, opened 2024-03-05
Document Context keys


Geal:
the request context is used a lot to carry data between parts of the router, there should be a central place in documentation to list them

theJC:
@bnjjj / @Geal -- Can we count on this documentation be made available alongside the 2.0 release? 🙏

goto-bus-stop:
Now available in https://www.apollographql.com/docs/graphos/routing/customization/overview#request-context

PR #6572, opened 2025-01-17
Update context key names for consistency and document them

tninesling:
Documentation and naming refactor for context keys. The new unified taxonomy for context keys is formatted at `apollo::` + router stage or plugin name + `::` + key descriptor.

Selected examples:
`operation_name` -> `apollo::supergraph::operation_name`
`apollo_authentication::JWT::claims` -> `apollo::authentication::jwt_claims`

A full compendium of available context keys in this new format can be found in the request lifecycle section of [docs/source/routing/customization/overview.mdx](https://github.com/apollographql/router/blob/tninesling/context-docs/docs/source/routing/customization/overview.mdx#request-context).

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


garypen (approved):
Thanks very much @tninesling !! This is going to be really useful for developers and customers.

bnjjj (commented):
This is great ! I would like to raise some concerns about this but it should not block this PR at all. This PR is like a big breaking change actually. And I'm wondering how we could do to make this migration for our customers smoother. I think it also needs to be documented in the migration guide. Which names changed, like `operation_name` is pretty common and used I think.
I also would like to make sure that in the future we don't end up again by introducing bad context key names again. How can we enforce this ? Same for documentation about these context keys ?

Also I know some of our customers are using the context keys in telemetry config using [context selectors ](https://www.apollographql.com/docs/graphos/reference/router/telemetry/instrumentation/selectors) and maybe we could try to craft a new migration in our migrations to make sure it will change the deprecated context keys to the new ones ? 

I also remember a discussion we had with @IvanGoncharov who suggested to add a new configuration for coprocessors in order to opt-in for context keys to propagate instead of propagating all context keys by default. I know we won't have time to do this but that's also something we have to keep in mind I think.

tninesling:
> Also I know some of our customers are using the context keys in telemetry config using [context selectors ](https://www.apollographql.com/docs/graphos/reference/router/telemetry/instrumentation/selectors) and maybe we could try to craft a new migration in our migrations to make sure it will change the deprecated context keys to the new ones ?

Great points @bnjjj. I added a full mapping of the old -> new values in the migration guide. Do you have a good example to follow for the config migration? I haven't done one of those before, and I think it would help avoid a lot of cases where folks don't realize they need to change the strings.

goto-bus-stop:
I'd also like to record my unease about the name changes (even though i think the new taxonomy is actually really great). I think this could be a significant burden for anyone using coprocessors or rhai scripts. We can't automatically warn them about this change and it might fail in invisible ways, so it requires careful manual fixing. I think we should have a discussion about whether to include the name changes in 2.0 at all. It's very similar to how we're removing old metrics in 2.0, and having two of those tricky updates in one release may be too much in my opinion.

lennyburdette:
We should at minimum make some explicit recommendations about how to handle this in advance in their scripts. In a JS coprocessor, you'd have to write something like 

```js
const value = req.context["the_old_key"] ?? req.context["the_new_key"];
```

i'm not really sure what the rhai equivalent is, but we could even provide some copy-pasteable snippets in the migration guide.

theJC:
@tninesling -- Probably can kill off https://github.com/apollographql/router/issues/4746 with your great work in this MR

goto-bus-stop:
@theJC thanks for calling that out, I assume that was not intentional or that the docs team is looking for a different place to put it. I'll follow up to make sure we have that information easily accessible *somewhere* in the regular docs for 2.x.

PR #6576, opened 2025-01-20
Promote `next` branch to `dev`; make 2.x our default development branch

abernix:
> tl;dr:  Make `next` (currently 2.x) into our `dev` (currently 1.x) branch and put 1.x work onto a new `1.x` branch.  After this PR lands, we'll follow-up immediately by opening a `1.x` branch with the commit _just prior_ to this merge.

Router 2.x has been under active development for nearly a year now and the `next` branch on this repository has been the place in which development has been taking place; while not always obvious, [198 PRs opened on this repository over the last many months](https://github.com/apollographql/router/pulls?q=sort%3Aupdated-desc+is%3Apr+base%3Anext+is%3Aclosed) (this link may change after this PR merges, but the number is right!) have been targeting `next`.

During that 2.x development, the `dev` branch (the default branch on this repository) has remained the location that 1.x work was taking place.  We have continually been merging work from `dev` (1.x) into `next` (2.x) to ensure relevant work manifests in both places.

As work on 2.x is becoming closer to "release ready" and 1.x is getting closer to moving into our "[maintenance LTS](https://www.apollographql.com/docs/graphos/reference/router-release-lifecycle#maintenance)".  This means that the substantive portion of work has pivoted; the vast majority of development is desired to take place on 2.x rather than 1.x and we want to update our branches to reflect that — as has always been the eventual plan.

With that context in mind, the time has come to:

1. Move 1.x work into a `1.x` branch
2. Let the `dev` branch become 2.x work (and the new "default" target).

To make that outcome true, this PR is the **~second~ third of ~three~ four steps** which we are taking:

- [x] https://github.com/apollographql/router/pull/6578 — ensure `main` is up to date with `dev`.
- [ ] _(🙅 **not this** PR)_ Merge `dev` into `next`, resolving any merge conflicts to fit into the "2.x outcome" satisfactorily.
   * This step will happen soon, however this PR will remain as draft until it is done.
- [ ] _(**this** PR ✅)_ Merge `next` into `dev` as a **true merge**, dissolving the (current) existence of `next` (indefinitely for now, though it may be back when 3.x comes around!l) and causing PRs targeting `next` to instead re-base themselves on `dev`.
- [x] (to be completed **after** this PR) Create a new `1.x` branch which will be the target for patches which become necessary to land on the 1.x   This likely won't manifest as a PR since it'll effectively be pushing a new branch.  There will, however, be some manual adjustments to our branch protection rules to enforce protections on that branch which also exist on `next` and `dev`.

[...]

PR #6587, opened 2025-01-20
Empty commit to allow a cascading re-targeting of PRs.

abernix:
This follows #6576 which **didn't** successfully re-target existing PRs, as we expected.  We have a theory that this is because of branch protection rules that prevented the deletion of the `next` branch.

Therefore, this is an empty commit that will try the same PR merge again and hopefully trigger the magical GitHub bots that re-target branches and save us from doing it manually.

Worst case, we will go through and re-target them all manually, but there are a fair dozen of them.


goto-bus-stop (approved):

abernix:
For posterity, that worked.  You're welcome.

PR #6597, opened 2025-01-20
DRAFT: router limits plugin

aaronArinder:
draft PR to check whether I'm heading in the right direction with a plugin for enforcing some product limits in the router (beginning with a transactions-per-second plugin); lots of FIXMEs and TODOs, but an early sanity check would be super useful

big questions I'm hoping to get help answers:
- is a separate plugin than the TrafficShaping one the right strategy? This duplicates a good chunk of that logic, but registers it as a private plugin that only gets added as a plugin if there are tps limits defined
- does the plumbing look correct? I _think_ this is how to register and then use a plugin, but I might be missing something big
- do you have a better name than `better_name`? `limits` is taken; I'm open to suggestions that don't sound like we're kneecapping the router on purpose
- what, apart from tests, is needed to get this out of draft and into review?

see #6561 

[product doc](https://apollographql.atlassian.net/wiki/spaces/Router/pages/edit-v2/1086881851); mostly outdated [design doc](https://apollographql.atlassian.net/wiki/spaces/BGS/pages/1184006257/2025-01-14+TPS+Rate+Limiting+in+the+Router), but still a useful read

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/plugins/better_name/mod.rs
+++ b/apollo-router/src/plugins/better_name/mod.rs
@@ -1,0 +1,130 @@
+// TODO: docs
+pub(crate) mod tps;
+
+use std::num::NonZeroU64;
+use std::time::Duration;
+
+use futures::future::BoxFuture;
+use futures::FutureExt;
+use http::StatusCode;
+use schemars::JsonSchema;
+use serde::Deserialize;
+use tower::BoxError;
+use tower::Service;
+use tower::ServiceBuilder;
+
+use crate::graphql;
+use crate::layers::ServiceBuilderExt;
+use crate::plugin::PluginInit;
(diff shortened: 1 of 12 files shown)


BrynCooke:
Sorry @aaronArinder , please take a look at this code which is backpressure aware: https://github.com/apollographql/router/blob/garypen/next-backpressure/apollo-router/src/plugins/traffic_shaping/mod.rs#L201-L302
The new plugin needs to be based on this.

PR #6601, opened 2025-01-21
Limit the depth of introspection queries

SimonSapin:
The [schema-intropsection schema](https://spec.graphql.org/draft/#sec-Schema-Introspection.Schema-Introspection-Schema) is recursive: a client can query the fields of the types of some other fields, and so on arbitrarily deep. This can produce responses that grow much faster than the size of the request.

To protect against abusive Router now refuses to execute introspection queries that nest list fields too deep and returns an error instead. The criteria matches `MaxIntrospectionDepthRule` in graphql-js, but may change in future versions.

In case it rejects legitimate queries, this check can be disabled in Router configuration:

```yaml
supergraph:
  introspection: true  # Needed to enable schema introspection in the first place
  introspection_check_max_depth: false  # Defaults to true
```

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [x] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/configuration/mod.rs
+++ b/apollo-router/src/configuration/mod.rs
@@ -627,6 +627,11 @@ pub(crate) struct Supergraph {
     /// Default: false
     pub(crate) introspection: bool,
 
+    /// Limit the depth of nested list fields in introspection queries
+    /// to protect avoid generating huge responses.
+    /// Default: false
+    pub(crate) introspection_check_max_depth: bool,
+
     /// Enable QP generation of fragments for subgraph requests
     /// Default: true
     pub(crate) generate_query_fragments: bool,
@@ -670,6 +675,7 @@ impl Supergraph {
         listen: Option<ListenAddr>,
         path: Option<String>,
         introspection: Option<bool>,
+        introspection_check_max_depth: Option<bool>,
         defer_support: Option<bool>,
         query_planning: Option<QueryPlanning>,
         generate_query_fragments: Option<bool>,
@@ -680,6 +686,8 @@ impl Supergraph {
             listen: listen.unwrap_or_else(default_graphql_listen),
             path: path.unwrap_or_else(default_graphql_path),
             introspection: introspection.unwrap_or_else(default_graphql_introspection),
+            introspection_check_max_depth: introspection_check_max_depth
+                .unwrap_or_else(default_graphql_introspection_check_max_depth),
             defer_support: defer_support.unwrap_or_else(default_defer_support),
             query_planning: query_planning.unwrap_or_default(),
             generate_query_fragments: generate_query_fragments
@@ -699,6 +707,7 @@ impl Supergraph {
         listen: Option<ListenAddr>,
         path: Option<String>,
         introspection: Option<bool>,
+        introspection_check_max_depth: Option<bool>,
         defer_support: Option<bool>,
         query_planning: Option<QueryPlanning>,
         generate_query_fragments: Option<bool>,
@@ -709,6 +718,8 @@ impl Supergraph {
             listen: listen.unwrap_or_else(test_listen),
             path: path.unwrap_or_else(default_graphql_path),
             introspection: introspection.unwrap_or_else(default_graphql_introspection),
+            introspection_check_max_depth: introspection_check_max_depth
+                .unwrap_or_else(default_graphql_introspection_check_max_depth),
             defer_support: defer_support.unwrap_or_else(default_defer_support),
             query_planning: query_planning.unwrap_or_default(),
             generate_query_fragments: generate_query_fragments
@@ -1438,6 +1449,10 @@ fn default_graphql_path() -> String {
 
 fn default_graphql_introspection() -> bool {
     false
+}
+
+fn default_graphql_introspection_check_max_depth() -> bool {
+    true
 }
 
 #[derive(Clone, Debug, Default, Error, Display, Serialize, Deserialize, JsonSchema)]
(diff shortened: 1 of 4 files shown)


.changesets/breaking_simon_introspection_max_depth.md
```diff
@@ -0,0 +1,15 @@
+### Limit the depth of introspection queries ([PR #6601](https://github.com/apollographql/router/pull/6601))
+
+The [schema-intropsection schema](https://spec.graphql.org/draft/#sec-Schema-Introspection.Schema-Introspection-Schema) is recursive: a client can query the fields of the types of some other fields, and so on arbitrarily deep. This can produce responses that grow much faster than the size of the request.
+
+To protect against abusive Router now refuses to execute introspection queries that nest list fields too deep and returns an error instead. The criteria matches `MaxIntrospectionDepthRule` in graphql-js, but may change in future versions.
```
goto-bus-stop, line 5:
```suggestion
To protect against abuse Router now refuses to execute introspection queries that nest list fields too deep and returns an error instead. The criteria matches `MaxIntrospectionDepthRule` in graphql-js, but may change in future versions.
```

goto-bus-stop (commented):
LGTM, should we also mention it in the migration guide?

apollo-router/src/introspection.rs
```diff
@@ -138,32 +159,47 @@ impl IntrospectionCache {
...
         let priority = compute_job::Priority::P1; // Low priority
-        let response =
-            compute_job::execute(priority, move || Self::execute_introspection(&schema, &doc))
-                .await
-                .expect("Introspection panicked");
+        let response = compute_job::execute(priority, move || {
+            Self::execute_introspection(max_depth, &schema, &doc)
+        })
+        .await
+        .expect("Introspection panicked");
```
lrlna, line 166: Do we have to keep this `.expect` here, or can it be an error?

lrlna (commented):
This is handled a little bit differently than our existing limits, and I am torn on whether it should be handled in a similar fashion. On one hand, it's definitely a limit configuration, but on the other hand, it's specific to graphql specification, and is closely tied to introspection implementation itself, which will make it hard to decouple the limiting functionality from introspection itself.

The one thing we do have to think about is the error we return regardless of where the actual limiting lives. Is there merit to returning the same `http::StatusCode::PAYLOAD_TOO_LARGE` error as the other limits do?

apollo-router/src/introspection.rs
SimonSapin: I believe the error path here is unreachable. `compute_job::execute` would return `Err` if a job panics, but before it gets a chance to do so the panic handler in `apollo-router/src/executable.rs` would exit the process. This leaves `expect` only for type-checking.

But even if we didn’t have this handler, what makes this path special out of all the places we could potentially catch panics and turn them into errors?

.changesets/breaking_simon_introspection_max_depth.md
SimonSapin: I a word, this was meant to be "To protect against abusive requests". Which do you prefer?

goto-bus-stop: Perfectly happy with abusive requests :)

apollo-router/src/introspection.rs
goto-bus-stop: It would be helpful to add that context about the panic handler as a code comment for future readers

apollo-router/src/configuration/mod.rs
```diff
@@ -627,6 +627,11 @@ pub(crate) struct Supergraph {
     /// Default: false
     pub(crate) introspection: bool,
 
+    /// Limit the depth of nested list fields in introspection queries
+    /// to protect avoid generating huge responses.
+    /// Default: false
+    pub(crate) introspection_check_max_depth: bool,
+
```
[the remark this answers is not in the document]
SimonSapin: Done!

apollo-router/tests/integration/introspection.rs
```diff
@@ -178,16 +178,185 @@ async fn mixed() {
...
+            }
+          ]
+        }
+      ]
+    }
+    "###);
+}
+
+#[tokio::test]
+async fn just_over_max_depth_with_check_disabled() {
```
lrlna, line 286: is there a test for "over max depth with check enabled"? am i just not seeing it and scrolling past it?

SimonSapin: Yes it’s the one above, `async fn just_over_max_depth()`

lrlna: i think it's time to log off 🙃

SimonSapin:
> should we also mention it in the migration guide?

Before this comment I didn’t realize we have this guide already. I’ve added to it now.

apollo-router/src/plugins/limits/mod.rs
```diff
@@ -115,6 +115,11 @@ pub(crate) struct Config {
     /// Default is ~400kib.
     #[schemars(with = "Option<String>", default)]
     pub(crate) http1_max_request_buf_size: Option<ByteSize>,
+
+    /// Limit the depth of nested list fields in introspection queries
+    /// to protect avoid generating huge responses.
+    /// Default: true
+    pub(crate) introspection_max_depth: bool,
```
lrlna, line 122:
```suggestion

    /// Limit the depth of nested list fields in introspection queries
    /// to protect avoid generating huge responses. Returns a GraphQL
    /// error with `{ message: "Maximum introspection depth exceeded" }` 
    /// when nested fields exceed the limit.
    /// Default: true
    pub(crate) introspection_max_depth: bool,
```

lrlna (approved):

PR #6618, opened 2025-01-22
[DRAFT] Test fixes

pubmodmatt:
Fixes for local test failures and attempted fixes for intermittent CI test failures that have become frequent:

* Conditionally ignore a test that always fails when running locally on MacOS
* Add a `--no-fail-fast` option to `cargo xtask dev` to run all tests instead of exiting after the first failure
* Fix an intermittent failure in `oltp` tests by using the `start_kill` function to kill the router process

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/services/layers/persisted_queries/mod.rs
+++ b/apollo-router/src/services/layers/persisted_queries/mod.rs
@@ -413,10 +413,8 @@ mod tests {
 
     use maplit::hashmap;
     use serde_json::json;
-    use tracing::instrument::WithSubscriber;
 
     use super::*;
-    use crate::assert_snapshot_subscriber;
     use crate::configuration::Apq;
     use crate::configuration::PersistedQueries;
     use crate::configuration::PersistedQueriesSafelist;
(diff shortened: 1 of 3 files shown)


xtask/src/commands/test.rs
```diff
@@ -25,6 +25,10 @@ pub struct Test {
     /// Pass --features to cargo test
     #[clap(long)]
     features: Option<String>,
+
+    /// Pass `--no-fail-fast` to `cargo test`
+    #[clap(long)]
+    no_fail_fast: bool,
```
goto-bus-stop, line 31: 💯

apollo-router/src/services/layers/persisted_queries/mod.rs
```diff
@@ -985,11 +983,14 @@ mod tests {
     }
 
     #[tokio::test(flavor = "multi_thread")]
+    #[cfg(not(all(not(feature = "ci"), target_os = "macos")))]
     async fn pq_layer_freeform_graphql_with_safelist_log_unknown_true() {
```
goto-bus-stop, line 987: Maybe we can add it to the list of flaky tests, rather than entirely disabling it?

This doesn't fail all the time. When we looked into it last, it would always pass if you run *only* this test with `cargo t pq_layer_freeform_graphql_with_safelist_log_unknown_true`. But we did not find the root cause despite looking at it pretty deeply.

PR #6620, opened 2025-01-22
ci(Mergify): configuration update

BrynCooke:
This change has been made by @BrynCooke from the Mergify config editor.

This will cause PRs that are raised via the backport label to impersonate the user that created the original PR.

Example on my test repo:

image


abernix (approved):

goto-bus-stop (approved):
oh nice, i'm surprised that that is possible

PR #6621, opened 2025-01-22
Remove `access_json` dependency and deprecated configurations

bnjjj:
Remove `access_json` from our dependencies. It's no longer needed and has been replaced by `jsonpath-rust` which follows the jsonpath specs.

It removes the deprecated `subgraph_response_body` selector from telemetry configuration and change the type of `path` field in `headers` configuration when you want to get a data from request body and insert it in a subgraph header.

+ If you have this configuration for headers:

```yaml
headers:
  all:
    request:
      - insert:
          name: from_app_name
          path: .extensions.metadata[0].app_name
```

you'll have to migrate the `path` field to be jsonpath compliant and just add `$` at the beginning of the `path`, example:

```yaml
headers:
  all:
    request:
      - insert:
          name: from_app_name
          path: $.extensions.metadata[0].app_name
```

+ If you used the `subgraph_response_body` selector for telemetry like this:

```yaml
telemetry:
  instrumentation:
    instruments:
      subgraph:
        http.client.request.duration:
          attributes:
            http.response.status_code:
              subgraph_response_status: code
            my_data_value:
              subgraph_response_body: .data.test
```

You'll have to migrate the `subgraph_response_body` to `subgraph_response_data` selector (or `subgraph_response_errors` if you want to target errors):

```yaml
telemetry:
  instrumentation:
    instruments:
      subgraph:
        http.client.request.duration:
          attributes:
            http.response.status_code:
              subgraph_response_status: code
            my_data_value:
              subgraph_response_data: $.test # Heads up, we removed the .data prefix as it's directly targeting data and we added $
```

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-288]: https://apollographql.atlassian.net/browse/ROUTER-288?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


goto-bus-stop (commented):
Does the router fail to start if you use the old syntaxes? I assume yes but just checkin' :)

Maybe it'd be good to land the `config validate` command for 2.0 and tell people to use that in the migration guide so they can spot issues with that

apollo-router/src/plugins/headers.rs
```diff
@@ -652,7 +663,7 @@ mod test {
         let mut service = HeadersLayer::new(
             Arc::new(vec![Operation::Insert(Insert::FromBody(InsertFromBody {
                 name: "header_from_request".try_into()?,
-                path: JSONQuery::parse(".operationName")?,
+                path: JsonPathInst::from_str("$.operationName").unwrap(),
```
garypen, line 666: Why `unwrap()` and not `?`?

garypen (commented):
Looks good, but I think we need to address the `unwrap()` that I reference.

apollo-router/src/plugins/headers.rs
bnjjj: It's just a test so it's better to panic asap imo

bnjjj:
@goto-bus-stop Actually I thought it would fail to select the right data without `$` as it's highly recommended in JSONPath specs. But I just added a new test to double check if it was working with previous notation (without `$`) and it works so it might be pretty transparent for our users but I think we should still mention it.

apollo-router/src/plugins/headers.rs
garypen: Ah, ok. I just spotted the change and wondered about it.

apollo-router/src/plugins/headers.rs
```diff
@@ -345,25 +345,36 @@ impl<S> HeadersService<S> {
...
+                                        output.to_string()
+                                    };
+                                match HeaderValue::from_str(&header_value) {
+                                    Ok(header_value) => {
+                                        req.subgraph_request
+                                            .headers_mut()
+                                            .insert(&from_body.name, header_value);
+                                    }
+                                    Err(err) => {
+                                        tracing::error!("cannot convert from the body into a header value for header name '{}': {:?}", from_body.name, err);
```
garypen, line 376: NIT, I prefer having the errors separately for json parsing purposes. So:
```suggestion
                                        let header_name = from_body.name;
                                        tracing::error!(header_name, err?, "cannot convert from the body into a header value for header name");
```
I think that syntax is correct, I haven't typed it for a while...

bnjjj: done in https://github.com/apollographql/router/pull/6621/commits/3032728cffe78efd9ece24f7ab2c33259ae44827

garypen (approved):

IvanGoncharov (approved):

PR #6622, opened 2025-01-22
Improve performance of query hashing by using a precomputed schema hash

IvanGoncharov:
This PR reverts the query hashing back to a simpler algorithm, which is faster and has more predictable CPU and memory costs.

## Previous vs. New Algorithm
- Both the old and new algorithms base their hash on the query, operation name, and supergraph schema.
- This hash is critical in various parts of the Router (e.g., caching query plans and entities).

## Key Change
- The new algorithm always hashes the entire schema (as an SDL string).
- The old algorithm tried to hash only the parts of the schema referenced by the query. While it allowed the hash to stay unchanged for irrelevant schema modifications, it introduced significant complexity and risk of collisions (partially fixed in #6205).
- Those fixes increased CPU and memory usage (improved somewhat in #6569), revealing scalability issues for large queries and schemas.

## Benefits of the New Approach
- Simple, predictable, and less prone to collisions (limited by SHA256).
- Since each query depends on the same supergraph schema, the algorithm computes the schema hash only once and reuses it.

## Implementation Details
- A new `SchemaHash` type has been introduced and the existing `QueryHash` type refined.
- `QueryHash` is now created via `SchemaHash::operation_hash` with a query string and optional operation name.
- This ensures the hash always includes the full schema, plus the full query and operation name.

## Future Plan
- This PR still allows some legacy shortcuts in `QueryHash` creation for now, but they will be removed in subsequent PRs.

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [x] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-978]: https://apollographql.atlassian.net/browse/ROUTER-978?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apollo-router/src/spec/schema.rs
```diff
@@ -362,6 +364,118 @@ impl std::ops::Deref for ApiSchema {
...
+///
+/// For a document with two queries A and B, queries A and B will result in a different hash even
+/// if the document text is identical.
+/// If query A is then executed against two different versions of the schema, the hash will be
+/// different again, depending on the [SchemaHash].
+///
+/// A query hash can be obtained from a schema ID using [SchemaHash::operation_hash].
+// FIXME: rename to OperationHash since it include operation name?
+#[derive(Clone, Hash, PartialEq, Eq, Deserialize, Serialize)]
+pub(crate) struct QueryHash(
```
goto-bus-stop, line 427: This is not blocking but something we could address once we have all tests passing and we know for sure that everything is working.
It's a bit odd that we have the QueryHash declared inside the schema module. Maybe we should have a hash.rs module, which would also guarantee the privacy of the internal storage of the hash types.

IvanGoncharov: good idea

goto-bus-stop:
There are many snapshot changes. The only changes we expect are:
- `schemaAwareHash` values in query plans
- Redis cache keys, but *only* in the `hash:` or `query:` section depending on the hash
So review should check against those expectations.

I've just reviewed all the snapshot changes thus far and they all look OK.

apollo-router/src/router_factory.rs
```diff
@@ -160,13 +160,15 @@ impl RouterSuperServiceFactory for YamlRouterFactory {
...
+                    inject_schema_id(schema.schema_id.as_str(), plugin_config);
                     match factory
                         .create_instance(
                             PluginInit::builder()
                                 .config(plugin_config.clone())
                                 .supergraph_sdl(schema.raw_sdl.clone())
-                                .supergraph_schema_id(schema.schema_id.clone())
+                                .supergraph_schema_id(Arc::new(
+                                    schema.schema_id.as_str().to_owned(),
+                                ))
```
goto-bus-stop, line 171:
```suggestion
                                .supergraph_schema_id(schema.schema_id.clone().into_inner())
```

apollo-router/tests/set_context.rs
```diff
@@ -313,6 +313,8 @@ fn setup_from_mocks(
     configuration: serde_json::Value,
     mocks: &[(&'static str, &'static str)],
 ) -> TestHarness<'static> {
+    let _ = rustls::crypto::aws_lc_rs::default_provider().install_default();
+
```
IvanGoncharov, line 317: @goto-bus-stop This looks like an unintentional change introduced by the merge.
Is it correct?

apollo-router/tests/type_conditions.rs
```diff
@@ -354,6 +354,8 @@ fn setup_from_mocks(
     configuration: serde_json::Value,
     mocks: &[(&'static str, &'static str)],
 ) -> TestHarness<'static> {
+    let _ = rustls::crypto::aws_lc_rs::default_provider().install_default();
+
```
IvanGoncharov, line 358: @goto-bus-stop looks like the same happened here

apollo-router/tests/set_context.rs
goto-bus-stop: Tests failed without it. We have to do this in several places after the hyper 1.0 upgrade on dev, but it's a bit weird that these tests were passing on dev before :/

apollo-router/src/query_planner/fetch.rs
```diff
@@ -100,7 +100,18 @@ pub(crate) type SubgraphSchemas = HashMap<String, SubgraphSchema>;
 
 pub(crate) struct SubgraphSchema {
     pub(crate) schema: Arc<Valid<apollo_compiler::Schema>>,
-    pub(crate) implementers_map: HashMap<Name, apollo_compiler::schema::Implementers>,
+    // TODO: Ideally should have separate nominal type for subgraph's schema hash
+    pub(crate) hash: Arc<SchemaHash>,
```
SimonSapin, line 104: The `SchemaHash` struct already contains an `Arc` field, so wrapping it in another `Arc` is redundant

apollo-router/src/query_planner/tests.rs
```diff
@@ -1877,23 +1877,17 @@ fn broken_plan_does_not_panic() {
...
-        query_planner::fetch::SubgraphSchema {
-            implementers_map: subgraph_schema.implementers_map(),
-            schema: Arc::new(subgraph_schema),
-        },
-    );
-    let result = Arc::make_mut(&mut plan.root)
-        .init_parsed_operations_and_hash_subqueries(&subgraph_schemas, "");
-    assert_eq!(
-        result.unwrap_err().to_string(),
-        r#"[1:3] Cannot query field "invalid" on type "Query"."#
```
SimonSapin, line 1897: This seems to remove a test, is that intentional?

goto-bus-stop: Oh, thanks! This can be added back. For a few commits during our exploration/fixing cycles we did not return a Result from `init_parsed_operations_and_hash_subqueries`.

apollo-router/src/query_planner/fetch.rs
IvanGoncharov: good catch! will fix

goto-bus-stop: Oh sorry @IvanGoncharov I already pushed it! 🙈

SimonSapin (approved):

apollo-router/src/spec/schema.rs
```diff
@@ -401,6 +403,118 @@ impl std::ops::Deref for ApiSchema {
...
+impl Display for QueryHash {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", hex::encode(&self.0))
+    }
+}
+
+// FIXME: It seems bad that you can create an empty hash easily and use it in security-critical
+// places. This impl should be deleted outright and we should update usage sites.
+// If the query hash is truly not required to contain data in those usage sites, we should use
+// something like an Option instead.
```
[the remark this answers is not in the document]
BrynCooke: What's the impact of fixing this now? Adding something that is a security risk doesn't sound good.

apollo-router/tests/set_context.rs
BrynCooke: It's because tonic changed. We need to call this.

BrynCooke (approved):

apollo-router/src/spec/schema.rs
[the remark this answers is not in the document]
IvanGoncharov: @BrynCooke Actual call to default happening here:
https://github.com/apollographql/router/pull/6622/files#diff-7108edf02483a7de6d8c37b938df0634afc78590242b4fc94709c9dd2e626fb4R108
Previously, it was just less noticeable, and my change and @goto-bus-stop comments just make it stand out more.
Fixing it will significantly increase scope of this PR but @SimonSapin is right, we need to open an issue about it.
I will do that right now

IvanGoncharov:
@mergifyio backport 1.59.2

goto-bus-stop:
@Mergifyio backport 1.x

BrynCooke:
@mergifyio refresh

PR #6624, opened 2025-01-22
Limit the compute job thread pool size

garypen:
The router has always observed the APOLLO_ROUTER_NUM_CORES environment variable to restrict the size of the main tokio async job scheduler.

We are now enhancing the compute job thread pool to both respect this environment variable and restrict the number of threads in the pool.

If the environment variable is not set, then the size of the pool is computed as a fraction of the total number of cores that the router has determined are available.

If it is set, then the environment variable is taken as the number of available cores.

From this number, let's call it available, the router then uses the following table to size the compute job thread pool:

/// available: 1     pool size: 1
/// available: 2     pool size: 1
/// available: 3     pool size: 2
/// available: 4     pool size: 3
/// available: 5     pool size: 4
/// ...
/// available: 8     pool size: 7
/// available: 9     pool size: 7
/// ...
/// available: 16    pool size: 14
/// available: 17    pool size: 14
/// ...
/// available: 32    pool size: 28
/// etc...

This table should not be relied upon as an explicit interface, since it may change in the future, but is provided here for informational purposes.

### DIFF
--- a/apollo-router/src/compute_job.rs
+++ b/apollo-router/src/compute_job.rs
@@ -14,13 +14,37 @@ use crate::metrics::meter_provider;
 /// reaches `QUEUE_SOFT_CAPACITY_PER_THREAD * thread_pool_size()`
 const QUEUE_SOFT_CAPACITY_PER_THREAD: usize = 20;
 
-/// Let this thread pool use all available resources if it can.
-/// In the worst case, we’ll have moderate context switching cost
-/// as the kernel’s scheduler distributes time to it or Tokio or other threads.
+/// Leave a fraction of CPU cores free to run Tokio threads even if this thread pool is very busy:
+///
+/// available: 1     pool size: 1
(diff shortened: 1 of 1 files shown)


lrlna:
@mergifyio backport 1.x

theJC:
Any chance you would be willing to incorporate the spirit of https://github.com/apollographql/router/pull/6663 into this and log out the effective pool size and queue for diagnostics/info?  

As a customer, I'd really appreciate having this available and be able to see the reality based on the amount of cores dialed in for router in kubernetes deployments.

garypen:
We will add the log lines at some point. Unfortunately, right now, if we add the log lines then unrelated logging tests in the router start to fail. See [this PR](https://github.com/apollographql/router/pull/6746/files) for more detail.

I'm closing this PR, since we are resolving the core problem elsewhere and we will add the log lines soon as we get time to work on them.

PR #6626, opened 2025-01-22
fix: explictly support `extend schema @link` for connectors

lennyburdette:
before this change, if you use `extend schema` to add the link directive and you don't define `Query`, the schema extension is dropped and the validations don't see the link. this means that all connector validations are skipped!

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [x] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


dylan-apollo (approved):
Ah, the ol' "subgraphs don't need to be valid GraphQL schemas" bit 😬

PR #6638, opened 2025-01-23
$request.headers and $response.headers for selection mapping in connectors

andrewmcgivery:
$request.headers and $response.headers for selection mapping in connectors

image

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/services/connector/request_service.rs
```diff
@@ -76,6 +77,8 @@ pub(crate) struct Request {
 
     /// Mapping problems encountered when creating the transport request
     pub(crate) mapping_problems: Vec<Problem>,
+
+    pub(crate) original_request: Arc<connect::Request>,
```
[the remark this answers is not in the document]
andrewmcgivery: Great idea. This cleans up the tests quite a bit. Working on implementing this now!

apollo-router/src/services/connector/request_service.rs
```diff
@@ -76,6 +77,8 @@ pub(crate) struct Request {
 
     /// Mapping problems encountered when creating the transport request
     pub(crate) mapping_problems: Vec<Problem>,
+
```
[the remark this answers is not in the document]
andrewmcgivery: Oops! Good catch!

apollo-router/src/plugins/connectors/make_requests.rs
```diff
@@ -92,6 +95,44 @@ impl RequestInputs {
...
+        // Add headers from the original router request
+        if variables_used.contains(&Namespace::Request) {
+            let headers: Map<ByteString, Value> = supergraph_request
+                .headers()
+                .iter()
+                .map(|(key, value)| {
+                    (
+                        key.as_str().into(),
+                        value.to_str().unwrap_or_default().into(),
+                    )
```
pubmodmatt, line 107: This is potentially cloning a lot of data. I think Lenny and I talked about optimizing this when we got to adding headers. Rather than just tracking which variable namespaces are in use, it could track which individual headers are referenced, and only cloning those values. That might end up being a fair amount of work, however.

lennyburdette: agreed, i think we can get a list of full variable paths and copy only the headers that are referenced. 

either that, or we can change apply_with_vars to take an object that lazily compute values instead of a serde_json_bytes::Value, but that's a bit harder because of how apply_with_vars is written

andrewmcgivery: Updated!

apollo-federation/src/sources/connect/json_selection/parser.rs
```diff
@@ -239,6 +240,32 @@ impl JSONSelection {
             .flat_map(|var_path| var_path.variable_reference())
             .map(|var_ref| var_ref.namespace.namespace)
     }
+
+    /// Get any headers referenced in the variable references by looking at both Request and Response namespaces.
+    pub fn header_references(&self) -> impl Iterator<Item = String> + '_ {
```
lennyburdette, line 245: json selection shouldn't know or care about headers, i think. is there another location for this code, preferably in or around an `http` module?

apollo-router/src/plugins/connectors/make_requests.rs
```diff
@@ -92,6 +97,54 @@ impl RequestInputs {
...
         }
 
+        // Add headers from the original router request.
+        // Only include headers that are actually referenced to save on passing around unused headers in memory.
+        if variables_used.contains(&Namespace::Request) {
+            let headers: Map<ByteString, Value> = supergraph_request
+                .headers()
+                .iter()
+                .filter_map(|(key, value)| {
+                    if headers_used.contains::<String>(&key.as_str().into()) {
```
pubmodmatt, line 107: Cloning the key here partially defeats the purpose. Instead:
```suggestion
                    if headers_used.contains(key.as_str()) {
```

apollo-router/src/plugins/connectors/make_requests.rs
```diff
@@ -92,6 +97,54 @@ impl RequestInputs {
...
+
+        // Add headers from the connectors response
+        // Only include headers that are actually referenced to save on passing around unused headers in memory.
+        if variables_used.contains(&Namespace::Response) {
+            if let Some(response_parts) = response_parts {
+                let headers: Map<ByteString, Value> = response_parts
+                    .headers
+                    .iter()
+                    .filter_map(|(key, value)| {
+                        if headers_used.contains::<String>(&key.as_str().into()) {
```
pubmodmatt, line 131: Same as above:
```suggestion
                        if headers_used.contains(key.as_str()) {
```

pubmodmatt (changes requested):

apollo-router/src/plugins/connectors/make_requests.rs
andrewmcgivery: I could have sworn this did not work before... I dunno what I did before but this works fine now! lol 😓

apollo-federation/src/sources/connect/json_selection/parser.rs
andrewmcgivery: The problem is, the `transport` doesn't contain the connector's JSON selection for the response, which is why I have to grab the variables from the `connect.selection`.

A possible alternative is I could maybe update `external_variables()` to generically return the variables instead of just the namespaces and apply the additional filtering/mapping for both `response_variables` and `response_headers` in `Connector` instead of in `JSONSelection`.

Does that sound like a reasonable alternative?

lennyburdette: i'm trying to remember if there was a specific reason why we didn't expose the full external variables outside of this module. it might have been because they're made up of json selection components that aren't meant for use by other code? @benjamn might have some suggestions here.

i do think it's generally the right approach, but maybe there's an abstraction we want to add that doesn't expose json selection bits.

i was thinking that if we had the whole variable path, we could also do some logging on startup if a connector references `$config.whatever` and it's not present in the YAML!

apollo-router/src/plugins/connectors/make_requests.rs
```diff
@@ -92,6 +97,54 @@ impl RequestInputs {
...
+                            value.to_str().unwrap_or_default().into(),
+                        ));
+                    }
+
+                    None
+                })
+                .collect();
+            let request_object = json!({
+                "headers": Value::Object(headers)
+            });
```
benjamn, line 119: Ah, so here's an example of where we turn actual request headers into the `$request` object. It's noteworthy that the `supergraph_request.headers()` method returns a `HeaderMap` (a multimap), and [`HeaderMap::iter`](https://docs.rs/http/latest/http/header/struct.HeaderMap.html#method.iter) handles multi-value headers by yielding the same header multiple times with different values, which is information that's lost by this `filter_map` code.

benjamn (changes requested):
Thanks so much for tackling this nontrivial improvement @andrewmcgivery!

While I support the general idea, I'm worried the choice to use object-based `$request.headers` and `$response.headers` maps assumes header names will be unique (see comments below), whereas HTTP headers can have multiple values per header.

Is there another representation we can use to expose the headers that takes this HTTP reality into account? Accessing a map where the values are always arrays gets a little less ergonomic (maybe a new `->header(name)` method could help?), but I don't see another way to preserve multiple values separately, so I think we need something closer to a multimap than a key-value object.

We might be able to represent multiple keys by concatenating them with commas (and documenting that policy), but that needs to happen somewhere, instead of letting the last value win.

apollo-router/src/plugins/connectors/make_requests.rs
andrewmcgivery: > We might be able to represent multiple keys by concatenating them with commas (and documenting that policy), but that needs to happen somewhere, instead of letting the last value win.

Comma separated makes sense to me. I can see a world where we provide a "split" function and then select the "n-th" element if there are multiple values and they need to select one.

My gut feeling is that while it is certainly possible to have multiple values for the same HTTP header, most of the time headers used in connectors would likely only have one value... thinking this is an 80/20 rule.

Assuming 80% of the time it'll have one value, I'd rather go with the easiest ergonomics for that case and provide a solution for the 20% that there will be multiple values.

This is definitely assumptions and subjective opinion though so if we know of a very common case where we'll have multiple values, happy to be wrong here :D

benjamn: I think our choices here need to be consistent with our choices elsewhere, and it seems like we mostly respect multimap behavior for HTTP headers elsewhere, so it's a departure to disregard that here.

Even if the single-value case is much more common, it's still disappointing as a developer to realize a tool you're using isn't really aligned with the rules/possibilities of HTTP, for those few cases where it matters. You can speculate about the distribution of use cases, but the weird edge cases (needing to interact with an API that abuses HTTP headers, for example) happen for everyone eventually.

andrewmcgivery: I guess I'm unclear on what the suggestion is here if not the comma separated list route. That sounded like a very reasonable compromise but I'm unclear if we're okay with that tradeoff or if you're suggesting a different approach. 😅 

if not `$request.headers.x`, what does it look like instead? And how does the user access value a versus value b of the same header?

lennyburdette: I'm thinking that we default to arrays in all cases. In common cases like `headers: [{ value: "prefix {$request.headers.'x-foo'}" }]` we'll actually fail the build with an error about using an array in this position (once we add shape info to that variable). users can choose either `->first` or eventually `->join(",")` to handle this.

(i'm also wondering if we'll want `$request.headers.'x-foo'[0]` before too long)

andrewmcgivery: Converted to array instead as suggested!

apollo-federation/src/sources/connect/json_selection/parser.rs
andrewmcgivery: Refactored this into models.rs as suggested. :)

PR #6640, opened 2025-01-24
Remove variable namespace checks from JSON selection

pubmodmatt:
Remove validation of variable namespaces from JSON Selection. These changes work toward a larger principle, which is that JSON Selection should be agnostic to the context it is used in, including what variable names are available. JSON Selection could some day be used in other places besides Connectors. Therefore, these changes remove a direct dependency on the `apollo-federation::sources::connect::variable::Namespace` type, which is the namespace specificlaly used by Connectors.

Instead, `KnownVariable::External` now contains only an owned `String`, rather than any specific namespace. Namespaces must be `FromStr` and `ToString`, so it is always possible to convert the caller's type for or to this string. `KnownVariable` makes no assumptions or guarantees about the value of this string, beyond what is accepted by the parser.

Validation of the specific values of this string are handled by the Connectors validation code, which ensures both that the values are one of the possible set, and also that the value can be used in the specific expression context - for example, `$status` can only be used in the context of a REST response.

A new test namespace is added in JSON Selection test code, which avoids picking up any dependency on a specific external namespace in tests, and ensures that JSON Selection works with any namespace.

The impact of these changes to customers is that error messages are now greatly improved. Instead of a vague "Unknown variable", a more detailed description of what variables are available is presented.

## Before
Screenshot 2025-01-23 at 3 58 44 PM
Screenshot 2025-01-23 at 3 39 05 PM

## After

Screenshot 2025-01-23 at 4 00 03 PM
Screenshot 2025-01-23 at 5 03 54 PM

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[CNN-540]: https://apollographql.atlassian.net/browse/CNN-540?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lennyburdette (approved):
nice work!

PR #6646, opened 2025-01-24
Removed deprecated methods for router@2.x

lrlna:
The following deprecated fields are being removed:
* `services::router::Response::map()`
* `router::event::schema::SchemaSource::File.delay` field
* `router::event::configuration::Configuration::File.delay` field
* `context::extensions::sync::ExtensionMutex::lock()`
* `test_harness::TestHarness::build()`
* `PluginInit::new()`
* `PluginInit::try_new()`

The following unsupported fields are being removed. Since they are "unsupported", I omitted them from the changeset and upgrade docs. 
* `axum_factory::unsupported_set_axum_router_callback`
* `PluginInit::unsupported_supergraph_schema`
* `PluginInit::unsupported_subgraph_schema`
*  `Context::unsupported_executable_document`

The following deprecated metrics are being removed:
- `apollo_require_authentication_failure_count`
- `apollo_authentication_failure_count`
- `apollo_authentication_success_count`
- `apollo_router_deduplicated_subscriptions_total`

Additionally, the callback of context::extensions::sync::ExtensionMutex 's `.with_lock()` function now receives `&mut Extensions` instead of a guard that dereferences to it. This makes a `mut` binding no longer necessary for the callback argument.


apollo-router/src/context/mod.rs
```diff
@@ -266,10 +266,8 @@ impl Context {
         }
     }
 
-    /// Read only access to the executable document. This is UNSTABLE and may be changed or removed in future router releases.
-    /// In addition, ExecutableDocument is UNSTABLE, and may be changed or removed in future apollo-rs releases.
-    #[doc(hidden)]
-    pub fn unsupported_executable_document(&self) -> Option<Arc<Valid<ExecutableDocument>>> {
+    /// Read only access to the executable document for internal router plugins.
+    pub(crate) fn executable_document(&self) -> Option<Arc<Valid<ExecutableDocument>>> {
```
lrlna, line 270: As talked about during the huddle this morning, this was changed to a pub(crate) to let us continue using this for router's own demand control plugin.

apollo-router/src/plugins/authentication/mod.rs
```diff
@@ -589,12 +589,6 @@ fn authenticate(
         status: StatusCode,
     ) -> ControlFlow<router::Response, router::Request> {
         // This is a metric and will not appear in the logs
-        u64_counter!(
-            "apollo_authentication_failure_count",
-            "Number of requests with failed JWT authentication (deprecated)",
-            1,
-            kind = "JWT"
-        );
```
lrlna, line 597: I removed several of these metrics that had `deprecated` notes. Let me know if this shouldn't happen.

docs/source/reference/migration/from-router-v1.mdx
```diff
@@ -188,6 +204,14 @@ telemetry:
 
 - `apollo_router_session_count_total` has been removed and replaced by `http.server.active_requests`.
 
+- `apollo_require_authentication_failure_count` has been removed. Use the `http.server.request.duration` metric's `http.response.status_code` attribute. Requests with authentication failures have HTTP status code 401.
+
+- `apollo_authentication_failure_count` has been removed. Use the `apollo.router.operations.authentication.jwt` metric's `authentication.jwt.failed` attribute.
+
+- `apollo_authentication_success_count` has been removed. Use the `apollo.router.operations.authentication.jwt` metric instead. If the `authentication.jwt.failed` attribute is _absent_ or `false`, the authentication succeeded.
+
+- `apollo_router_deduplicated_subscriptions_total` has been removed. Use the `apollo.router.operations.subscriptions` metric's `subscriptions.deduplicated` attribute
```
lrlna, line 213: I am not certain how docs are going to be built for 1.x vs 2.x. Is dev explicitly tied to 2.x docs now? Should I go ahead and remove/tweak the actual docs referencing these metrics?

SimonSapin (approved):

.changesets/breaking_lrlna_remove_deprecated_interfaces.md
```diff
@@ -26,6 +26,8 @@ succeeded.
 - `apollo_router_deduplicated_subscriptions_total`. Use the
 `apollo.router.operations.subscriptions` metric's `subscriptions.deduplicated`
 attribute
+- `apollo_router_span` has been removed. Use `apollo.router.processing.time`
```
BrynCooke, line 29: We should not recommend this either. We should instead remove processing time also. It's a flawed metric.

BrynCooke:
```
Calculating the overhead of injecting the router into your service stack when making multiple downstream calls is a complex task. We are removing these misleading metrics until we can figure out something better.

It is recommended that you test your workloads with and the router and see if the latency meets your requirements.
```

BrynCooke (changes requested):

docs/source/reference/migration/from-router-v1.mdx
garypen: `dev` is the 2.x docs now. 1.x docs will come from `1.x`. So, yes you can update docs.

garypen (approved):
I think you may have done: https://apollographql.atlassian.net/browse/ROUTER-297 as well (although, I'm not totally sure).
Thanks for the diligent cleanup. Nice to see so much code being deleted...

docs/source/reference/migration/from-router-v1.mdx
lrlna: excellent! done in be696bcb5da1235069a8d910b2a9b13f504375a1

BrynCooke (approved):

PR #6648, opened 2025-01-24
[JSONSelection] Make `->map` always return an array

benjamn:
Prior to this PR, the `->map` method had the runtime behavior of producing an array output when given an array input (mapping each of the elements), but producing only a single (mapped) output value when given a non-array as input.

This behavior took inspiration from the automatic mapping behavior of GraphQL selection sets, which produce only a single object when the input is not an array. However, in practice, this behavior complicates any attempt to provide a static output type/shape for the `->map` method, because the output type/shape has to be a union `One<T, List<T>>` for some `T`. To make matters worse, depending on the input type `@`, the output might be either the result of mapping just `@` itself (which I'll write `@'`, inventing notation), or an array resulting from mapping each of the elements of `@`, when `@` is known to be an array. It's tricky to write the union accurately in terms of `@`, but it's something like `One<@', List<@.*'>>`, where `.*` denotes a union of an array's elements (and `'` maps over that union).

In the `friends: friend_ids->map({ id: @ })` example from the tests, the union would end up being
```graphql
friends: { id: $root.friend_ids } | List<{ id: $root.friend_ids.* }>
```
This duplication feels undesirable, and it's not even exact duplication, because one side has `.*` and the other does not.

**This PR proposes a breaking change**: I believe we should make the `->map` method return a single-element wrapper array even when the input was not an array, allowing us to say `->map` always returns a `List<@.*'>` for input shape `@`. Conveniently, the `.*` path element is already [defined](https://github.com/apollographql/shape-rs/blob/4884961b79032f97cceed667d47bf6dfbb40db21/src/child_shape.rs#L33-L35) to return the input shape itself when it's not an array, so `@.*'` is the same shape as `@'` when `@` is not an array.

This change would allow us to produce output shapes like this instead, with no union or duplication involved:
```graphql
friends: List<{ id: $root.friend_ids.* }>
```

We may not know yet whether `$root.friend_ids` will be an array or not, but `List<{ id: $root.friend_ids.* }>` now accurately describes the output type of `friend_ids->map({ id: @ }`, regardless.

As a final example of the value of `->map` consistently returning an array, consider the common pattern of mapping an input and then slicing it with `input->map(...)->slice(0, 3)` or selecting only the first element with `input->map(...)->first`. With the proposed change, these uses of `->slice` and `->first` always work, even if they are strict and only accept an array or string, because `input->map(...)` now reliably returns an array.


apollo-federation/src/sources/connect/validation/expression.rs
```diff
@@ -305,7 +305,7 @@ mod tests {
     #[case("$args.customScalar")]
     #[case("$args.object.bool")]
     #[case("$args.array->echo(1)")]
-    #[case("$args.int->map(@)")]
+    #[case("$args.int->map(@)->last")]
```
dylan-apollo, line 308: Can we add the old, now-invalid case to the invalid_methods up above? To make sure we're catching `$args.int->map(@)` and `$args.array->map(@)` as errors now?

benjamn: https://github.com/apollographql/router/pull/6648/commits/842f9a9739fa874bd9a3dce0b1f864e80f706592

apollo-federation/src/sources/connect/validation/expression.rs
pubmodmatt, line 1: Can the test case on line 340 be enabled now? It appears to pass now.

benjamn: https://github.com/apollographql/router/pull/6648/commits/d1b1cc66c3878dc3278448f2dfff2680d562804f

dylan-apollo (approved):

PR #6654, opened 2025-01-27
tidy up config migration post js qp removal

lrlna:
We had two migrations at conflict with one other regarding the deleted `experimental_reuse_query_fragments` config. This PR reconciles the conflict, and separates out `0032-removal-legacy-qp.yaml` migration into separate files for each individual config.


abernix:
@mergifyio backport 1.60.0

apollo-router/src/configuration/migrations/0031-experimental_reuse_query_fragments.yaml
```diff
@@ -0,0 +1,4 @@
+description: supergraph.experimental_reuse_query_fragments is deprecated
```
IvanGoncharov, line 1: I don't have a context for those changes, but it might be worth mentioning that `supergraph.generate_query_fragments` is enabled by default and is a safe feature. 
I think it would look less scary than just removing the flag.

IvanGoncharov (approved):

apollo-router/src/configuration/migrations/0031-experimental_reuse_query_fragments.yaml
lrlna: we do have in the [changelog](https://github.com/apollographql/router/releases/tag/v1.58.0), but it's a good call to be extra explicit here.

lrlna, line 1:
```suggestion
description: supergraph.experimental_reuse_query_fragments fragment optimization config is removed and superseded by supergraph.generate_query_fragments
```

apollo-router/src/configuration/migrations/0034-experimental_parallelism.yaml
```diff
@@ -0,0 +1,4 @@
+description: Along with the removal of the legacy, JavaScript, query planner, supergraph.query_planning.experimental_parallelism is also removed.
```
lrlna, line 1:
```suggestion
description: along with the removal of the legacy, JavaScript, query planner, supergraph.query_planning.experimental_parallelism is also removed
```

PR #6657, opened 2025-01-27
Reduce calls to unwrap and expect

tninesling:
Reduces the number of usages of unwrap and expect with the following changes:
1. Pattern match options and json values instead of using `is_none`/`as_array` followed by `expect`
2. Propagate proteus parsing errors as `ConfigurationError::MigrationFailure` instead of panicking in config migrations
3. Replace `Mutex` and `RwLock` from `std::sync` with the equivalents from `parking_lot` to avoid unwraps related to lock poisoning. The [panic handler](https://github.com/apollographql/router/blob/v1.59.1/apollo-router/src/executable.rs#L738-L758) means the process should exit before we get to a point where a poisoned lock can be accessed, so this mostly serves to clean up the code. Note that instances of `tokio::sync::RwLock` have not been changed.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/.changesets/fix_bnjjj_fix_6441.md
+++ b/.changesets/fix_bnjjj_fix_6441.md
@@ -1,0 +1,5 @@
+### Entity cache: fix directive conflicts in cache-control header ([Issue #6441](https://github.com/apollographql/router/issues/6441))
+
+Unnecessary cache-control directives are created in cache-control header.  The router will now filter out unnecessary values from the `cache-control` header when the request resolves. So if there's `max-age=10, no-cache, must-revalidate, no-store`, the expected value for the cache-control header would simply be `no-store`. Please see the MDN docs for justification of this reasoning: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#preventing_storing
+
+By [@bnjjj](https://github.com/bnjjj) in https://github.com/apollographql/router/pull/6543
--- a/apollo-router/src/plugins/cache/cache_control.rs
+++ b/apollo-router/src/plugins/cache/cache_control.rs
@@ -178,6 +178,15 @@ impl CacheControl {
         let mut s = String::new();
         let mut prev = false;
         let now = now_epoch_seconds();
+        if self.no_store {
+            write!(&mut s, "no-store")?;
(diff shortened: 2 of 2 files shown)


apollo-router/src/configuration/mod.rs
```diff
@@ -108,6 +108,46 @@ pub enum ConfigurationError {
...
+
+impl From<proteus::parser::Error> for ConfigurationError {
+    fn from(error: proteus::parser::Error) -> Self {
+        Self::MigrationFailure {
+            error: error.to_string(),
+        }
+    }
+}
+
+impl From<serde_yaml::Error> for ConfigurationError {
```
goto-bus-stop, line 127: The proteus impls are great, but I think these should be manual at the site where the error is propagated. It's conceivable that we could produce errors of these types in config loading in the future in situations where they're not related to migrations at all

tninesling: Good point. Updated in [d3bc2bb](https://github.com/apollographql/router/pull/6657/commits/d3bc2bbf0859888a80cec1376f35b1cf0bdddc8c).

apollo-router/src/configuration/mod.rs
```diff
@@ -108,6 +108,38 @@ pub enum ConfigurationError {
...
+    }
+}
+
+impl From<tracing_core::metadata::ParseLevelError> for ConfigurationError {
+    fn from(error: tracing_core::metadata::ParseLevelError) -> Self {
+        Self::MigrationFailure {
+            error: error.to_string(),
+        }
+    }
+}
```
goto-bus-stop, line 141: I shouldve selected the whole thing :P I have the same concern with these impls as in https://github.com/apollographql/router/pull/6657#discussion_r1931747522.

goto-bus-stop:
everything else LGTM

apollo-router/src/configuration/mod.rs
tninesling: All good! Updated in [f8ca562](https://github.com/apollographql/router/pull/6657/commits/f8ca56292dc7faef8be42369056e2e40c590c109)

goto-bus-stop (approved):
thanks!

PR #6658, opened 2025-01-27
[JSONSelection] Allow iterative refinement of output shapes with `ShapedSelection` wrapper

benjamn:
This PR ~builds on PR #6659 by introducing~ introduces a `ShapedSelection` struct that wraps both a `JSONSelection` and an `IndexMap<String, Shape>` of named shapes, computing and storing `ShapedSelection::output_shape` based on these inputs.

The `ShapedSelection` type also supports `refine` and `replace` methods for adding to the map of named shapes and recomputing the output shape (producing a new `ShapedSelection` object):
```rust
pub(crate) fn refine(&self, new_named_shapes: IndexMap<String, Shape>) -> ShapedSelection;
pub(crate) fn replace(&self, new_named_shapes: IndexMap<String, Shape>) -> ShapedSelection;
```
Though they have the same signature, the key difference between `refine` and `replace` is that `refine` handles name collisions by combining shapes with `Shape::all`, whereas `replace` overwrites existing named shapes in the map.

Typically, the `JSONSelection` will not change as the `ShapedSelection` is refined, but the map of named shapes will grow from one or more external sources, via one or more calls to `refine`.

Finally, this PR demonstrates one place where we might use `ShapedSelection::refine` to specify better variable shapes, but I've temporarily commented this out with a TODO (see below) because it causes a number of (reasonable) test failures.


apollo-federation/src/sources/connect/validation/expression.rs
```diff
@@ -84,7 +84,25 @@ pub(crate) fn validate(expression: &Expression, context: &Context) -> Result<(),
...
+    // TODO Reenable this when we can deal with the consequences better,
+    // using (for example) tools for mapping GraphQL null values to None
+    // when appropriate.
+    // .refine(var_shapes);
+
+    let shape = shaped_selection.output_shape();
+
     let errors: Vec<Error> = shape
         .errors()
         .map(|err| Error {
```
benjamn, line 108: Here's where I temporarily commented out the call to `ShapedSelection::refine` that specifies better variable shapes. @dylan-apollo may want to try un-commenting this code to see what fails.

PR #6659, opened 2025-01-27
[JSONSelection] `ComputOutputShape` refactoring to support `ShapedSelection`

benjamn:
This PR collects a few refactoring changes I wanted to make before attempting PR #6658:
* Updating the `shape` crate to the latest published version, 0.4.2
* Renaming `named_var_shapes` parameters to `named_shapes`, since the map is not just for variables
* Moving the `compute_output_shape` method implementations from the `ApplyToInternal` trait to a new `ComputeOutputShape` trait
* Unifying the handling of `None`, union and intersection shapes, and errors across all `PathList` variants
  * Among other benefits, this means `->` methods never have to worry about receiving `None` or a union or intersection shape as their input shape, since we can handle those shapes automatically without invoking `method.shape(...)`

These changes should be mostly organizational and should have no runtime consequences outside the `compute_output_shape` code.

### DIFF
--- a/apollo-federation/src/sources/connect/json_selection/methods.rs
+++ b/apollo-federation/src/sources/connect/json_selection/methods.rs
@@ -78,14 +78,14 @@ macro_rules! impl_arrow_method {
                 method_args: Option<&MethodArgs>,
                 input_shape: Shape,
                 dollar_shape: Shape,
-                named_var_shapes: &IndexMap<&str, Shape>,
+                named_shapes: &IndexMap<String, Shape>,
             ) -> Shape {
                 $shape_fn_name(
                     method_name,
                     method_args,
                     input_shape,
                     dollar_shape,
-                    named_var_shapes,
+                    named_shapes,
                 )
             }
         }
@@ -123,7 +123,7 @@ pub(super) trait ArrowMethodImpl {
         // Other variable shapes may also be provided here, though in general
         // variables and their subproperties can be represented abstractly using
         // $var.nested.property ShapeCase::Name shapes.
-        named_var_shapes: &IndexMap<&str, Shape>,
+        named_shapes: &IndexMap<String, Shape>,
     ) -> Shape;
 }
 
--- a/apollo-federation/src/sources/connect/json_selection/shape.rs
+++ b/apollo-federation/src/sources/connect/json_selection/shape.rs
@@ -1,0 +1,545 @@
+use apollo_compiler::collections::IndexMap;
+use shape::Shape;
+use shape::ShapeCase;
+
+use super::known_var::KnownVariable;
+use super::lit_expr::LitExpr;
+use super::location::WithRange;
+use super::methods::ArrowMethod;
+use super::JSONSelection;
+use super::NamedSelection;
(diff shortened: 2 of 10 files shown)


apollo-federation/src/sources/connect/json_selection/shape.rs
```diff
@@ -0,0 +1,493 @@
...
+        input_shape: Shape,
+        // Shape of the `$` variable, which is bound to the closest enclosing
+        // subselection object, or the root data object if there is no enclosing
+        // subselection.
+        dollar_shape: Shape,
+        // Shapes of other named variables, with the variable name `String`
+        // including the initial `$` character. This map typically does not
+        // change during the compute_output_shape recursion, and so can be
+        // passed down by immutable reference.
+        named_shapes: &IndexMap<String, Shape>,
```
dylan-apollo, line 66: Not for this PR, but I wonder if in the future we should have this take a single struct, like `&InputShapes` or something, that way the fields of that struct can use the `///` comments, which will show folks these comments at a glance in their editors. _might_ also reduce some boilerplate elsewhere.

dylan-apollo (approved):
LGTM. This will conflict with the location PRs, but I think it's a good refactor, so worth the pain.

apollo-federation/src/sources/connect/json_selection/shape.rs
```diff
@@ -0,0 +1,493 @@
...
+    fn compute_output_shape(
+        &self,
+        // Shape of the `@` variable, which typically changes with each
+        // recursive call to compute_output_shape.
+        input_shape: Shape,
+        // Shape of the `$` variable, which is bound to the closest enclosing
+        // subselection object, or the root data object if there is no enclosing
+        // subselection.
+        dollar_shape: Shape,
+        // Shapes of other named variables, with the variable name `String`
```
pubmodmatt, line 62: This comment seems out of date with the rename from `named_var_shapes` to `named_shapes` - the name may not be a variable name.

apollo-federation/src/sources/connect/json_selection/methods.rs
```diff
@@ -123,7 +123,7 @@ pub(super) trait ArrowMethodImpl {
         // Other variable shapes may also be provided here, though in general
```
pubmodmatt, line 123: Same comment issue here - out of date with renaming from `named_var_shapes`

pubmodmatt (approved):
Just nits on comments

apollo-federation/src/sources/connect/json_selection/shape.rs
benjamn: https://github.com/apollographql/router/pull/6659/commits/7ef08ae160d84253673eb59df0c3284464d9edbf

apollo-federation/src/sources/connect/json_selection/methods.rs
benjamn: https://github.com/apollographql/router/pull/6659/commits/7ef08ae160d84253673eb59df0c3284464d9edbf

PR #6664, opened 2025-01-28
with_metrics() doesn't require `Send`

garypen:
The Router service is required to use UnsyncBoxBody because futures::Stream is not Sync. Changing this to use a different Streams implementation (e.g.: tokio) is a lot of work and not guaranteed to improve things.

This causes issues for `with_metrics()` since it requires the provided closure to be `Send` so that it can be boxed(). Actually, `with_metrics()` doesn't have to box the created closure, it could use `boxed_local()`.

That leads to a simple change to `with_metrics()` which means Router service can be used with it and all existing tests continue to function.

### DIFF
--- a/apollo-router/src/metrics/mod.rs
+++ b/apollo-router/src/metrics/mod.rs
@@ -1193,7 +1193,7 @@ macro_rules! assert_non_zero_metrics_snapshot {
 }
 
 #[cfg(test)]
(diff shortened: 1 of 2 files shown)


BrynCooke (approved):

goto-bus-stop:
ah, is this like a "better" version of https://github.com/apollographql/router/pull/6655? :)

PR #6675, opened 2025-01-28
[WIP] docs: upgrade guides

shorgi:
- Move router upgrade guides from `migration` to `upgrade` path and section
- Fix preview version #


Meschreiber (approved):
Do these file name changes require updates in the _sidebar.yaml file to grab the correct file? Additionally, should we update the sidebar category from "Migration Guides" to something like "Migration and Upgrade Guides"?

Separately, the new paths will also require redirects, right?

shorgi:
> Do these file name changes require updates in the _sidebar.yaml file to grab the correct file? ... new paths will also require redirects, right?

All taken care of in the [docs-rewrite PR](https://github.com/apollographql/docs-rewrite/pull/441/files) that's providing the preview

shorgi:
@mergify queue

shorgi:
@mergify queue

PR #6679, opened 2025-01-28
doc: clarify offline license

shorgi:
Clarify use of schema delivery to maintain schema when using offline license


Meschreiber (approved):
Lgtm!

shorgi:
@mergify queue

PR #6686, opened 2025-01-29
tidy up config migration post js qp removal (backport #6654)

lrlna:
We had two migrations at conflict with one other regarding the deleted `experimental_reuse_query_fragments` config. This PR reconciles the conflict, and separates out `0032-removal-legacy-qp.yaml` migration into separate files for each individual config.<hr>This is an automatic backport of pull request #6654 done by [Mergify](https://mergify.com).


goto-bus-stop (approved):
I cant claim to fully understand the change but I do agree it does the same thing as on dev :))

bnjjj (approved):

PR #6688, opened 2025-01-29
Body selection shape validation

pubmodmatt:
* Use the shape system to validate connectors body selections. This adds validations for arrow methods, and makes it possible to validate that the output type is sensible.
* Add an extension trait to simplify working with offset ranges

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[CNN-581]: https://apollographql.atlassian.net/browse/CNN-581?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apollo-federation/src/sources/connect/validation/selection.rs
+++ b/apollo-federation/src/sources/connect/validation/selection.rs
@@ -11,6 +11,7 @@ use apollo_compiler::schema::ExtendedType;
 use apollo_compiler::schema::ObjectType;
 use apollo_compiler::Node;
 use itertools::Itertools;
+use shape::ShapeCase;
 
 use super::coordinates::ConnectDirectiveCoordinate;
 use super::coordinates::SelectionCoordinate;
@@ -23,6 +24,7 @@ use crate::sources::connect::expand::visitors::GroupVisitor;
 use crate::sources::connect::json_selection::ExternalVarPaths;
 use crate::sources::connect::json_selection::NamedSelection;
 use crate::sources::connect::json_selection::Ranged;
+use crate::sources::connect::range::RangeExt;
 use crate::sources::connect::spec::schema::CONNECT_SELECTION_ARGUMENT_NAME;
 use crate::sources::connect::validation::coordinates::connect_directive_http_body_coordinate;
 use crate::sources::connect::validation::graphql::GraphQLString;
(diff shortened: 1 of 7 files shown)


apollo-federation/src/sources/connect/validation/test_data/body_selection.graphql
```diff
@@ -61,4 +61,28 @@ type Query {
...
+            POST: "http://127.0.0.1",
+            body: "$({ userid: $nosuchvariable })" # INVALID - no such variable
+        }
+        selection: "$"
+    )
+    stringLiteralFromArrowMethod(userid: ID!): String
+    @connect(
+        http: {
+            POST: "http://127.0.0.1",
+            body: "$({ userid: $args.userid })->jsonStringify" # INVALID - results in string literal
```
lennyburdette, line 84: this surprised me but i see the logic in it. we'll probably eventually support string values when you set `content-type: text/plain` or something, but let's assume for now that a non-object/non-array is a mistake. we should document this and probably list it as a breaking change in the preview changelog

apollo-federation/src/sources/connect/validation/test_data/body_selection.graphql
```diff
@@ -61,4 +61,28 @@ type Query {
         }
         selection: "$"
     )
+    invalidArrowMethod(userid: ID!): String
```
lennyburdette, line 64: can't highlight line 52: `# INVALID, but currently allowed - output is not valid JSON` — is that comment still correct?

apollo-federation/src/sources/connect/validation/selection.rs
```diff
@@ -91,65 +93,105 @@ pub(super) fn validate_body_selection(
...
+                .line_col_for_subslice(expression_location.narrow(err.range.as_ref()), schema)
                 .into_iter()
                 .collect(),
-        });
+        }
+    }));
+    match shape.case() {
+        ShapeCase::Name(name, _) if name == "$root" => messages.push(Message {
+            code: Code::InvalidJsonSelection,
+            message: format!("In {coordinate}: invalid body selection - body selections have no input to select from"),
```
lennyburdette, line 155:
```suggestion
            message: format!("In {coordinate}: body selections cannot access the API response with `$`"),
```

or something. i think the word "input" here is potentially confusing

lennyburdette (approved):
a couple of suggestions, but otherwise lgtm!

apollo-federation/src/sources/connect/validation/test_data/body_selection.graphql
pubmodmatt: Good catch - and that does seem inconsistent with the error about returning a string literal from `jsonStringify`...

apollo-federation/src/sources/connect/validation/selection.rs
pubmodmatt: Agreed - I'm not sure what the best way to convey this to customers is, but I'll think about it more.

dylan-apollo: In URIs and headers, we just say that expressions must start with `$args`, `$this`, etc. (whatever is valid there). Technically a lie, because you can also use literals/methods, but if they're trying to use a bare path they _probably_ meant `$this`

apollo-federation/src/sources/connect/validation/selection.rs
```diff
@@ -91,65 +93,105 @@ pub(super) fn validate_body_selection(
...
+                    .collect(),
+            }]
+        }
+    };
+    let selection = match JSONSelection::parse(selection_str.as_str()) {
+        Ok(selection) => selection,
+        Err(err) => {
+            return vec![Message {
+                code: Code::InvalidJsonSelection,
+                message: format!("{coordinate} is not a valid JSONSelection: {err}"),
```
dylan-apollo, line 119:
```suggestion
                message: format!("{coordinate} is not valid mapping syntax: {err}"),
```

Or something? Obviously this change isn't actually from this PR, but we should get rid of user-facing references to "JSONSelection" where we notice them.

dylan-apollo: I guess that means we should also rename that code...

apollo-federation/src/sources/connect/validation/selection.rs
```diff
@@ -91,65 +93,105 @@ pub(super) fn validate_body_selection(
...
+            code: Code::InvalidJsonSelection,
+            message: format!("In {coordinate}: invalid body selection - body selections have no input to select from"),
+            locations: selection_str
+                .line_col_for_subslice(expression_location, schema)
+                .into_iter()
+                .collect(),
+        }),
+        ShapeCase::String(_) | ShapeCase::Int(_) | ShapeCase::Float | ShapeCase::Bool(_) => messages.push(Message {
+            code: Code::InvalidJsonSelection,
+            message: format!("In {coordinate}: body selection cannot be a scalar literal"),
```
dylan-apollo, line 163: Why are we limiting this?

apollo-federation/src/sources/connect/validation/selection.rs
```diff
@@ -91,65 +93,105 @@ pub(super) fn validate_body_selection(
...
+            locations: selection_str
+                .line_col_for_subslice(expression_location, schema)
+                .into_iter()
+                .collect(),
+        }),
+        // TODO: Handle All, One, Array, etc.
```
dylan-apollo, line 169: I think _most_ of these are going to result in `One`, because most GraphQL types are nullable. I think we can re-use the logic from `expression.rs` here

dylan-apollo (changes requested):
I think we can get the types of the variables sorted, too, if we re-use `expression.rs`. Since we're not going forward with the labels & resolver improvements, I can take a stab at that and make a PR into this one.

We're also going to have much better locations soon, so _hopefully_ we don't need narrow

apollo-federation/src/sources/connect/validation/selection.rs
pubmodmatt: This was already here, only the indent has changed in this PR - but it's a good point that we should sanitize anyplace that exposes "JSON Selection" to customers.

apollo-federation/src/sources/connect/validation/selection.rs
pubmodmatt: It's a good question. We're setting the `content-type` to `application/json` for the request, so sending something like "5" seems incorrect. I think there's a larger issue here, that we need support for `text/plain` on both the request and response side - I've hit the same thing on the response side, where the response is just a string, and I want to map that string onto a GraphQL field.

apollo-federation/src/sources/connect/validation/selection.rs
pubmodmatt: I'll take it in hand to do that sanitization in a separate PR, as that'll fan out into a lot of snapshot updates.

PR #6690, opened 2025-01-29
(fix) Header propagation rule passthrough

BrynCooke:
Header propagation contains logic to prevent headers from being propagated more than once. This was broken in https://github.com/apollographql/router/pull/6281 which always considered a header propagated regardless  if a rule actually matched.

This PR alters the logic so that only when a header is populated then the header is marked as fixed.

The following will now work again:
```
headers:
  all:
    request:
      - propagate:
          named: a
          rename: b
      - propagate:
          named: b
```

Note that defaulting a head WILL populate a header, so make sure to include your defaults last in your propagation rules.

```
headers:
  all:
    request:
      - propagate:
          named: a
          rename: b
          default: defaulted # This will prevent any further rule evaluation for header `b`
      - propagate:
          named: b
```
Instead make sure that your headers are defaulted last:

```
headers:
  all:
    request:
      - propagate:
          named: a
          rename: b
      - propagate:
          named: b
          default: defaulted # OK
```

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


goto-bus-stop (approved):
subtle... should we also clarify in the docs that you should put items with `default` last?

BrynCooke:
> subtle... should we also clarify in the docs that you should put items with `default` last?
Good point I'll add it.

BrynCooke:
@mergify queue

BrynCooke:
@shorgi I've made some docs changes so it'd be great if you could take a pass.

abernix (approved):

PR #6692, opened 2025-01-29
create a new health_check plugin

garypen:
This migrates the health checking functionality to a plugin, health_check, which is now responsible for indicating if the router is LIVE and/or READY.

The main benefit of this change is a substantial improvement in the accuracy of live/ready indication (which should help k8s deployments) and a much simpler lifecycle.


apollo-router/src/axum_factory/tests.rs
```diff
@@ -2264,77 +2262,45 @@ async fn send_to_unix_socket(addr: &ListenAddr, method: Method, body: &str) -> S
...
                 .listen(ListenAddr::SocketAddr("127.0.0.1:0".parse().unwrap()))
                 .enabled(true)
                 .build(),
         )
         .supergraph(Supergraph::fake_builder().path("/health").build()) // here be dragons
         .build()
         .unwrap();
+
+    // Manually add the endpoints, since they are only created if the health-check plugin is
+    // enabled and that won't happ in init_with_config()
```
lrlna, line 2279:
```suggestion
    // enabled and that won't happen in init_with_config()
```

apollo-router/src/configuration/snapshots/apollo_router__configuration__tests__schema_generation.snap
lrlna, line 1: this one is a bit difficult to follow. it's just the ordering that changed in this file, right?

apollo-router/src/plugins/healthcheck/mod.rs
```diff
@@ -0,0 +1,506 @@
...
+
+/// Configuration options pertaining to the readiness health sub-component.
+#[derive(Clone, Debug, Deserialize, Serialize, JsonSchema)]
+#[serde(deny_unknown_fields)]
+#[serde(default)]
+pub(crate) struct ReadinessConfig {
+    /// The readiness interval configuration
+    pub(crate) interval: ReadinessIntervalConfig,
+
+    /// How many errors/interval are allowed until unready (default: 100)
```
lrlna, line 74: is `100` the _number_ of errors, or 100 second interval? the `errors/interval` is slightly misleading to me.

apollo-router/src/plugins/healthcheck/mod.rs
```diff
@@ -0,0 +1,506 @@
...
+        let my_recovery_interval = init
+            .config
+            .readiness
+            .interval
+            .unready
+            .unwrap_or(3 * my_sampling_interval);
+        let my_rejected = rejected.clone();
+        let my_ready = ready.clone();
+
+        my_rejected.store(0, Ordering::Relaxed);
```
garypen, line 195: remove

apollo-router/src/plugins/healthcheck/mod.rs
```diff
@@ -0,0 +1,506 @@
...
+                    if query_upper.starts_with("READY") {
+                        let status = if my_ready.load(Ordering::SeqCst) {
+                            HealthStatus::Up
+                        } else {
+                            // It's hard to get k8s to parse payloads. Especially since we
+                            // can't install curl or jq into our docker images because of CVEs.
+                            // So, compromise, k8s will interpret this as probe fail.
+                            status_code = StatusCode::SERVICE_UNAVAILABLE;
+                            HealthStatus::Down
+                        };
```
lrlna, line 261: I can't match up this comment with this if-else. So if `my_ready` is at [AtomicBool] of 1, so health status is up, if at 0, we are down. But what does the payload have to do with the health status being down? is it that for a specific type of payload we declare somewhere else that the health check is not ready?

BrynCooke (approved):
Approved pending changeset and docs.
We did a live walkthough.

apollo-router/src/state_machine.rs
```diff
@@ -274,7 +271,7 @@ impl<FA: RouterSuperServiceFactory> State<FA> {
         new_state.unwrap_or(self)
     }
 
-    async fn shutdown<S>(self, http_server_factory: &S) -> Self
+    async fn shutdown<S>(self, _http_server_factory: &S) -> Self
```
lrlna, line 274: 😅 should we just change the function signature?

lrlna (approved):
None of my comments are blocking!

And I quite like this in a plugin, makes this whole thing more readable 🙏

apollo-router/src/plugins/healthcheck/mod.rs
```diff
@@ -0,0 +1,506 @@
...
+impl Default for Config {
+    fn default() -> Self {
+        Self::builder().build()
+    }
+}
+
+struct HealthCheck {
+    config: Config,
+    live: Arc<AtomicBool>,
+    ready: Arc<AtomicBool>,
```
BrynCooke, line 165: Shame rust doesn't have atomic enum built in.

apollo-router/src/plugins/healthcheck/mod.rs
garypen: It is the number of rejections per interval. I'll phrase that more clearly.

apollo-router/src/plugins/healthcheck/mod.rs
garypen: The payload is the textual representation of the health status, the status code is the HTTP status code. k8s will drive its probe behaviour from the HTTP status code, but other clients may choose to parse the json body that we return. We are basically returning two different indicators of up/down-ness: status code OK or UNAVAILABLE and some body text (JSON) which is UP or DOWN (matching the status codes). I'm not sure this is answering your question...

PR #6695, opened 2025-01-29
Try not to panic when span lookups fail

tninesling:
Primarily updates calls for span lookups to log an error when the lookup doesn't find a span. The OTel layer hooks unfortunately do not return results, but I think a no-op with an error log is preferably than a panic in the case of a missing span. Additionally, this rewrites a few small blocks to avoid calls to unwrap/expect and propagate errors where possible. 

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


garypen (approved):
Great work Taylor! This is another nice contribution to the quality story in Router 2.0.

PR #6697, opened 2025-01-29
docs: add connectors reference/link

Meschreiber:
- Adds a link to connectors doc in the configuration reference
- Mentions Connectors in the router overview

### DIFF
--- a/docs/source/routing/about-router.mdx
+++ b/docs/source/routing/about-router.mdx
@@ -16,16 +16,18 @@ GraphOS Router is the runtime of the GraphOS platform. It executes client operat
   class="hidden dark:block"
 />
 
-Apollo Connectors are a core part of the router, enabling API orchestration with REST APIs.
-
-<img src='../../img/connectors/connector.svg' class="dark:hidden"/>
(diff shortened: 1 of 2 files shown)


docs/source/routing/about-router.mdx
```diff
@@ -16,6 +16,11 @@ GraphOS Router is the runtime of the GraphOS platform. It executes client operat
   class="hidden dark:block"
 />
 
+Apollo Connectors are a core part of the router, enabling API orchestration with REST APIs.
```
shorgi, line 19: - A header should improve scannability and searchability
- If agree to the header, suggest moving it to after Runtime of GraphOS platform

```suggestion

### Enables Apollo Connectors

Apollo Connectors are a core part of the router, enabling API orchestration with REST APIs.
```

shorgi (approved):

docs/source/routing/about-router.mdx
Meschreiber: https://github.com/apollographql/router/pull/6697/commits/5bd86563cf43ca791c7efcde14f85f22ea0db11c

PR #6707, opened 2025-01-30
Catch invalid fields of literal objects

dylan-apollo:
Causes a composition error in URIs/headers/body when a literal object has a property accessed which doesn't exist.

This is taken from the name resolution code, but has no impact on non-literals right now since there's no resolution yet.

### DIFF
--- a/apollo-federation/src/sources/connect/json_selection/apply_to.rs
+++ b/apollo-federation/src/sources/connect/json_selection/apply_to.rs
@@ -406,7 +406,7 @@ impl ApplyToInternal for NamedSelection {
                 let output_key = alias_opt
                     .as_ref()
                     .map_or(key.as_str(), |alias| alias.name());
-                let field_shape = dollar_shape.field(key.as_str(), key.shape_location(source_id));
+                let field_shape = field(&dollar_shape, key, source_id);
                 output.insert(
                     output_key.to_string(),
                     if let Some(selection) = selection {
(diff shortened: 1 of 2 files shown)


benjamn (approved):
More tests would be great, but this change makes sense to me.

dylan-apollo:
@benjamn I added cases for subselection, in addition to the path that was there, and both sub and path within a nested array.

I _think_ that covers the main places this new function is used, but are there any others you can think of we should cover?

apollo-federation/src/sources/connect/json_selection/apply_to.rs
```diff
@@ -989,6 +989,28 @@ impl ApplyToInternal for SubSelection {
...
+        return Shape::none();
+    }
+    let field_shape = shape.field(key.as_str(), key.shape_location(source_id));
+    if field_shape.is_none() {
+        return Shape::error(
+            format!("field `{field}` not found", field = key.as_str()),
+            key.shape_location(source_id),
+        );
+    }
+    field_shape
```
pubmodmatt, line 1011: This is tangential, but in several reviews now I've thought it might be more "Rusty" to add an `or` method on `Shape`:
```rust
    pub fn or(self, other: Shape) -> Shape {
        if self.is_none() {
            other
        } else {
            self
        }
    }
```
Which would let you change code like this to:
```rust
    shape.field(key.as_str(), key.shape_location(source_id)).or(Shape::error(
        format!("field `{field}` not found", field = key.as_str()),
        key.shape_location(source_id),
    ))
```

pubmodmatt (approved):

PR #6715, opened 2025-01-31
Disable type checking of arrow methods

dylan-apollo:
Since these checks are incomplete until we finish name resolution, we've decided to disable them entirely for consistency. Will re-enable in a future version with an opt-in from users.


apollo-federation/src/sources/connect/json_selection/apply_to.rs
```diff
@@ -703,47 +703,9 @@ impl ApplyToInternal for WithRange<PathList> {
...
-                        tail.compute_output_shape(
-                            method_result_shape,
-                            dollar_shape.clone(),
-                            named_var_shapes,
-                            source_id,
-                        )
-                    }
+            PathList::Method(method_name, _method_args, _tail) => {
+                if let Some(_method) = ArrowMethod::lookup(method_name) {
+                    Shape::unknown(method_name.shape_location(source_id))
```
benjamn, line 708: Could we add a TODO comment here about eventually calling `_method.shape` to find out the method's result shape?

PR #6727, opened 2025-02-05
Error codes on trace spans

bonnici:
Adds new span events to trace spans that include errors generated by connectors and the demand control plugin. These error events are added to the trace but not logged in the console. The plan is to add more events for more error types once we have confirmed that this is the right way to implement this functionality.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/plugins/connectors/handle_responses.rs
```diff
@@ -182,6 +183,16 @@ impl RawResponse {
             }
         };
 
+        if let MappedResponse::Error {
```
pubmodmatt, line 186: I don't think it affects what you're doing here, but just be aware that what is considered an error here may change going forward. Currently any HTTP response status other than 2XX is an error, but in the future, there are plans to make this more customizable by the customer. For example, a 404 may just result in a `null` field, not an error.

lennyburdette: the `extensions.code` property will also be user-customizable, if that matters

.changesets/feat_trace_span_error_code.md
```diff
@@ -0,0 +1,5 @@
+### Brief but complete sentence that stands on its own ([PR #6727](https://github.com/apollographql/router/pull/6727))
```
abernix, line 1: 😄

abernix (changes requested):

abernix (dismissed):
Only reviewed the changeset; feel free to dismiss once this is fixed, but don't want this to get missed and only show up at release time. 😄

apollo-router/src/plugins/demand_control/mod.rs
```diff
@@ -342,16 +343,19 @@ impl Plugin for DemandControl {
...
-                                .errors(
-                                    err.into_graphql_errors()
-                                        .expect("must be able to convert to graphql error"),
-                                )
-                                .context(req.context.clone())
-                                .build()
-                                .expect("Must be able to build response"),
-                        ),
+                        Err(err) => {
+                            emit_error_event(err.code(), "Demand control execution error");
```
BrynCooke, line 347: We'll now emit a log for every request that is rejected. I think this could/will cause a flood of logs to the users' APM. I don't think we can go with this as is. I'm going to do a bit of reading around what the latest recommendations are around logging and comment further.

BrynCooke (changes requested):

apollo-router/src/plugins/demand_control/mod.rs
bonnici: By log do you mean to standard out? Because the changes to `fmt_layer.rs` will prevent that.

BrynCooke: Ahh I see!

BrynCooke (approved):

PR #6737, opened 2025-02-06
Traffic shaping for connectors

andrewmcgivery:
Example config:

```
traffic_shaping:
  connector:
    sources:
        global_rate_limit:
          capacity: 20
          interval: 1s
        experimental_http2: http2only
        timeout: 1s
```

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [x] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/services/supergraph/service.rs
```diff
@@ -619,6 +620,17 @@ async fn subscription_task(
                         .and_then(|plugin| (*plugin.1).as_any().downcast_ref::<Subscription>())
                         .map(|p| p.config.clone());
 
+                    let connector_sources: HashSet<String> = execution_service_factory.schema
+                        .connectors
+                        .as_ref()
+                        .map(|c| {
+                            c.by_service_name
+                                .iter()
+                                .map(|(_, connector)| format!("{}.{}", connector.id.subgraph_name.clone(), connector.id.source_name.clone().unwrap_or_default()))
```
lennyburdette, line 629: this should be a getter on the connector model, i'm thinking

lennyburdette: especially since it should be an option — if there's no source, there shouldn't be a source name (otherwise you'll have `subgraph_name.` for a bunch of connectors)

andrewmcgivery: I'd be down with that... what should it be called? 🤔

lennyburdette: it's internal so it doesn't really matter, but maybe `fn source_config_key(&self) -> Option<String>`?

apollo-router/src/plugins/traffic_shaping/mod.rs
```diff
@@ -153,6 +159,8 @@ pub(crate) struct Config {
     all: Option<SubgraphShaping>,
     /// Applied on specific subgraphs
     subgraphs: HashMap<String, SubgraphShaping>,
+    /// Applied on specific subgraphs
+    sources: HashMap<String, SubgraphShaping>,
```
lennyburdette, line 163: spoke offline about using a different struct here that doesn't include deduplication, and have a From impl that converts to the standard one

andrewmcgivery: Donezo!

apollo-router/src/services/supergraph/service.rs
andrewmcgivery: So one small thing to point out just to make sure we're on the same page... there needs to be some kind of "default" value if there is no source_name because the services and http factories are created and looked up by this value as part of this PR... so if I have no value at all... it'll just break connectors lol

Would we prefer the default to be just `connector.id.subgraph_name` or some other default value?

andrewmcgivery: I moved this into a method on Connector but kept it as `String` instead of `Option<String>` because the code necessitates that this always has some kind of a value due to creating http clients.

When there is no source name, it ends up being `subgraph_name.anonymous`.

lennyburdette: this would conflict with `@source(name: "anonymous")` right? and we're letting users configure traffic shaping for source-less connectors by using `subgraph_name.anonymous` in their yaml?

what's a way we can create and lookup clients without exposing this detail?

andrewmcgivery: It would result in the same connector http client and if you set any config if you used anonymous, yea. Which... wouldn't break anything but perhaps have unintended consequences if they configured `subgraph_name.anonymous` in their yaml.

> and we're letting users configure traffic shaping for source-less connectors by using subgraph_name.anonymous in their yaml?

Setting config on `all`, instead of `sources`, would target all connectors and subgraphs, even connectors without a source.

I wouldn't expect `anonymous` it to be used for sourceless connectors. The intent would not be to document this and it instead be an internal detail. Given that... maybe just making it a less likely overlapping name would suffice like `__anonymous__`?

andrewmcgivery: Lenny and I resolved this discussion via a phone call and basically decided to use existing "id" methods to construct a unique identifier instead of just "anonymous". :)

apollo-router/src/configuration/snapshots/apollo_router__configuration__tests__schema_generation.snap
```diff
@@ -2060,6 +2067,37 @@ snapshot_kind: text
...
     },
+    "ConnectorShaping": {
+      "additionalProperties": false,
+      "properties": {
+        "compression": {
+          "$ref": "#/definitions/Compression",
+          "description": "#/definitions/Compression",
+          "nullable": true
+        },
+        "dns_resolution_strategy": {
```
lennyburdette, line 2078: am i correct in reading that `dns_resolution_strategy` and `experimental_http2` aren't tested?

apollo-router/src/plugin/test/mock/connector.rs
lennyburdette, line 1: were there things you had to test this way (using this new mock service and the tests in plugins/traffic_shaping/mod.rs) that warrant two ways of testing? i'm wondering if we can kill this in favor or tests/integration/traffic_shaping.rs?

apollo-federation/src/sources/connect/models.rs
```diff
@@ -231,6 +231,16 @@ impl Connector {
             .map_err(|_| internal_error!("Failed to create key for connector {}", self.id.label)),
         }
     }
+
+    /// Create an identifier for this connector that can be used for configuration and service identification
```
lennyburdette, line 235: can you document the behavior when source is None and why?

apollo-router/src/plugins/connectors/handle_responses.rs
```diff
@@ -49,7 +49,7 @@ pub(crate) enum HandleResponseError {
 
 // --- RAW RESPONSE ------------------------------------------------------------
 
-enum RawResponse {
+pub(crate) enum RawResponse {
```
pubmodmatt, line 52: Why `pub(crate)`?

apollo-router/src/configuration/snapshots/apollo_router__configuration__tests__schema_generation.snap
andrewmcgivery: I did not explicitly write tests for them because the existing traffic shaping also doesn't.

My interpretation as to why is because while these settings are in traffic shaping, the actual logic/functionality for them is used elsewhere in other parts of the codebase.

apollo-router/src/plugin/test/mock/connector.rs
andrewmcgivery: I'm be fine with doing just the integration tests.

I did both ways mostly to match what was already being done in the existing tests.... but I'm not married to it. I tend to monkey-see-monkey-do when touching existing code. :D

apollo-router/src/router_factory.rs
```diff
@@ -423,6 +424,31 @@ pub(crate) async fn create_http_services(
         let http_service_factory = HttpClientServiceFactory::new(http_service, plugins.clone());
         http_services.insert(name.clone(), http_service_factory);
     }
+
+    // Also create client service factories for connector sources
```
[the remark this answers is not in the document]
andrewmcgivery: Would creating a list of "connector subgraph names" prior to that loop and doing a comparison of the subgraph name versus the connector subgraph name suffice?

Basically... if we have a subgraph name that matches a connector subgraph name... don't add it.

apollo-router/src/plugins/connectors/http_json_transport.rs
```diff
@@ -103,7 +103,7 @@ pub(crate) fn make_request(
                 transport.body.as_ref().map(|body| SelectionData {
                     source: body.to_string(),
                     transformed: body.to_string(), // no transformation so this is the same
-                    result: json_body,
+                    result: json_body.clone(),
```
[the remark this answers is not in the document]
andrewmcgivery: Got left in from a previous iteration of this PR.. removed!

apollo-router/src/services/connector/request_service.rs
```diff
@@ -169,51 +249,54 @@ impl ErrorExtension for Error {
...
-                self.plugins.iter().rev().fold(
-                    ConnectorRequestService {
-                        http_client_service_factory: self.http_client_service_factory.clone(),
-                    }
-                    .boxed(),
-                    |acc, (_, e)| e.connector_request_service(acc),
-                ),
-            )
-            .boxed()
+    pub(crate) fn create(&self, source_name: String) -> BoxService {
```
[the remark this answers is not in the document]
andrewmcgivery: Ah crap, definitely accidentally nuked it. I'll look at adding it back!

apollo-router/src/plugins/connectors/handle_responses.rs
andrewmcgivery: Got left in from a previous iteration of this PR.. removed!

apollo-federation/src/sources/connect/models.rs
andrewmcgivery: Added some more comment!

apollo-router/src/services/connector/request_service.rs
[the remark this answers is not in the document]
andrewmcgivery: Added it back as suggested!

apollo-router/src/plugins/traffic_shaping/mod.rs
```diff
@@ -390,6 +430,91 @@ impl Plugin for TrafficShaping {
...
+        service: crate::services::connector::request_service::BoxService,
+        source_name: String,
+    ) -> crate::services::connector::request_service::BoxService {
+        let all_config = self.config.all.as_ref();
+        let source_config = self
+            .config
+            .sources
+            .get(&source_name)
+            .map(|connector_config| connector_config.clone().into());
+        let final_config = Self::merge_config(all_config, source_config.as_ref());
```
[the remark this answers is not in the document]
andrewmcgivery: Yep! This was a problem because before the `all` config was the subgraph config which could contain `deduplicate_query` whereas now the merge will be two connector configs.

apollo-router/src/plugins/connectors/handle_responses.rs
```diff
@@ -203,7 +203,7 @@ impl RawResponse {
 }
 
 // --- MAPPED RESPONSE ---------------------------------------------------------
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
lennyburdette, line 206: remind me why this is clone? maybe even add a comment for future readers?

andrewmcgivery: I must have accidentally left it in from a previous iteration of code 😓 removed!

apollo-router/src/services/supergraph/service.rs
```diff
@@ -940,6 +953,17 @@ impl PluggableSupergraphServiceBuilder {
             .and_then(|plugin| (*plugin.1).as_any().downcast_ref::<Subscription>())
             .map(|p| p.config.clone());
 
+        let connector_sources: HashSet<String> = schema
+            .connectors
```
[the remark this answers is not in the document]
lennyburdette: yeah, both the set of subgraphs and set of source config keys can be calculated once at schema load time.

lennyburdette (approved):

PR #6743, opened 2025-02-06
Add dhat test for connectors validation

pubmodmatt:
Add a test to catch increases in the maximum heap usage used by connectors valiadtion. This would have quickly caught a very large increase in heap usage that led to an OOM case in composition.

This adds `dhat` as a test dependency, and running tests will produce a `dhat-heap.json` file, which is added to `.gitignore`. The test is added as an integration test, as described in the `dhat` documention. The `apollo-federation` integration tests are intentionally compiled into a single binary, but `dhat` tests need to run in their own binary to prevent concurrent heap usage from interfering with the test. So the new test is broken out into its own isolated binary.

In the event of failure, the `dhat-heap.json` can be used to find the problem, with something like:

```sh
cargo install dhat-to-flamegraph\
  && dhat-to-flamegraph dhat-heap.json > /tmp/out.svg\
  && open /tmp/out.svg
```

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

**Notes**

### DIFF
--- a/apollo-federation/tests/connectors/validation.rs
+++ b/apollo-federation/tests/connectors/validation.rs
@@ -1,0 +1,17 @@
+#[global_allocator]
+pub(crate) static ALLOC: dhat::Alloc = dhat::Alloc;
+
+#[test]
+fn valid_large_body() {
+    const SCHEMA: &str = "src/sources/connect/validation/test_data/valid_large_body.graphql";
+    const MAX_BYTES: usize = 204_800; // 200 KiB
+
+    let schema = std::fs::read_to_string(SCHEMA).unwrap();
+
+    let _profiler = dhat::Profiler::builder().testing().build();
+
+    apollo_federation::sources::connect::validation::validate(&schema, SCHEMA);
+
+    let stats = dhat::HeapStats::get();
+    dhat::assert!(stats.max_bytes < MAX_BYTES);
+}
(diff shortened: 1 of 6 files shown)


apollo-federation/tests/connectors/validation.rs
```diff
@@ -0,0 +1,17 @@
...
+    const MAX_BYTES: usize = 204_800; // 200 KiB
+
+    let schema = std::fs::read_to_string(SCHEMA).unwrap();
+
+    let _profiler = dhat::Profiler::builder().testing().build();
+
+    apollo_federation::sources::connect::validation::validate(&schema, SCHEMA);
+
+    let stats = dhat::HeapStats::get();
+    dhat::assert!(stats.max_bytes < MAX_BYTES);
```
lrlna, line 16: It _may_ be interesting to get the total number of allocations (`stats.total_blocks`) as well to catch a change that could allocate [and deallocate] a lot, but still not hit the max heap size number.

lrlna (approved):
I am _really_ into this idea. We should have this sort of tests in more places!

apollo-federation/tests/connectors/validation.rs
pubmodmatt: Added. The current number is quite high for this test (`20,391 blocks`), but that's another good thing about adding these tests - they provide an easy way to test changes to impove these numbers.

PR #6745, opened 2025-02-07
prep release: v2.0.0-preview.6

abernix:
Bumps versions in preparation for v2.0.0-preview.6


abernix (commented):
Confirmed this is just version bumps, and will correctly land back into `dev`.

garypen (approved):

PR #6746, opened 2025-02-07
Add $APOLLO_ROUTER_COMPUTE_THREADS environment variable

SimonSapin:
Router creates a thread pool dedicated to "compute jobs" to avoid blocking Tokio threads. By default this pool has as many threads as available CPUs. This can now be changed by setting the `$APOLLO_ROUTER_COMPUTE_THREADS` environment variable to an integer value. This feature is intentionally undocumented and should be considered "experimental".

Also rename the pre-existing (also undocumented) `$APOLLO_ROUTER_NUM_CORES` to `$APOLLO_ROUTER_IO_THREADS` to better reflect what it does.

Drive-by unrelated change: make `cargo run` default to the main router executable.

----

Manual testing instructions: in a terminal, run:

```sh
APOLLO_ROUTER_COMPUTE_THREADS=4 cargo run -- \
  -c router.yaml -s examples/graphql/supergraph.graphql --dev \
  --log error,apollo_router=info,apollo_router::compute_job=debug
```

After seeing a log line with `GraphQL endpoint exposed at http://127.0.0.1:4000/`, in another terminal, run:

```sh
curl -s -H 'accept: application/json' -H 'content-type: application/json' 'http://127.0.0.1:4000/' -d '{"query":"{__typename}"}'
```

(This step is needed because the thread pool is created lazily on first use.) The first terminal should now have a log line with `Compute thread pool size: $APOLLO_ROUTER_COMPUTE_THREADS = 4`.

---

**Checklist**

  * Undocumented `$APOLLO_ROUTER_NUM_CORES` no longer works
- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/compute_job.rs
+++ b/apollo-router/src/compute_job.rs
@@ -23,14 +23,11 @@ fn thread_pool_size() -> usize {
         .ok()
         .and_then(|value| value.parse::<usize>().ok())
     {
-        tracing::info!(threads, type="compute", source="env", "thread pool");
         threads
     } else {
-        let threads = std::thread::available_parallelism()
+        std::thread::available_parallelism()
             .expect("available_parallelism() failed")
-            .get();
-        tracing::info!(threads, type="compute", source="available_parallelism", "thread pool");
-        threads
+            .get()
     }
 }
 


apollo-router/Cargo.toml
```diff
@@ -11,6 +11,7 @@ license = "Elastic-2.0"
 rust-version = "1.83.0"
 edition = "2021"
 build = "build/main.rs"
+default-run = "router"
```
SimonSapin, line 14: This sets the default for `cargo run`, so `cargo run --bin router` isn’t needed.

apollo-router/src/compute_job.rs
```diff
@@ -14,13 +14,25 @@ use crate::metrics::meter_provider;
...
+        .ok()
+        .and_then(|value| value.parse::<usize>().ok())
+    {
+        tracing::debug!("Compute thread pool size: $APOLLO_ROUTER_COMPUTE_THREADS = {size}");
+        size
+    } else {
+        let size = std::thread::available_parallelism()
+            .expect("available_parallelism() failed")
+            .get();
+        tracing::debug!("Compute thread pool size: available_parallelism() = {size}");
```
SimonSapin, line 33: This log can also help makes sure cgroup limits affect `available_parallelism()` like we expect when benchmarking

apollo-router/src/compute_job.rs
```diff
@@ -14,13 +14,25 @@ use crate::metrics::meter_provider;
...
-    std::thread::available_parallelism()
-        .expect("available_parallelism() failed")
-        .get()
+    // This environment variable is intentionally undocumented.
+    // See also APOLLO_ROUTER_IO_THREADS in apollo-router/src/executable.rs
+    if let Some(size) = std::env::var("APOLLO_ROUTER_COMPUTE_THREADS")
+        .ok()
+        .and_then(|value| value.parse::<usize>().ok())
+    {
+        tracing::debug!("Compute thread pool size: $APOLLO_ROUTER_COMPUTE_THREADS = {size}");
```
garypen, line 27: Better format for JSON parseability:
```
tracing::info!(size, type="compute", source="env", "thread pool");
```
I'd make it info since it will not be flooding the logs and it seems like useful information

apollo-router/src/compute_job.rs
garypen:
```
tracing::info!(size, type="compute", source="calculated", "thread pool");
```

apollo-router/src/executable.rs
```diff
@@ -327,12 +327,16 @@ pub fn main() -> Result<()> {
...
     builder.enable_all();
-    if let Some(nb) = std::env::var("APOLLO_ROUTER_NUM_CORES")
+
+    // This environment variable is intentionally undocumented.
+    // See also APOLLO_ROUTER_COMPUTE_THREADS in apollo-router/src/compute_job.rs
+    if let Some(nb) = std::env::var("APOLLO_ROUTER_IO_THREADS")
         .ok()
         .and_then(|value| value.parse::<usize>().ok())
     {
         builder.worker_threads(nb);
```
garypen, line 337: Worth logging these details as well:
```
tracing::info!(size=nb, type="io", source="env", "thread pool");
```

garypen (commented):
Formatting up strings in logs makes them harder to consume or work with from JSON logs.

apollo-router/src/executable.rs
SimonSapin: Can `tracing` be used before we create a Tokio runtime? (If not, I can move things around to log after we’re in async context)

apollo-router/src/executable.rs
```diff
@@ -677,6 +681,11 @@ impl Executable {
             }
         };
 
+        let threads = tokio::runtime::Handle::current().metrics().num_workers();
+        tracing::info!(threads, type="io", "thread pool");
```
SimonSapin, line 685: This is done here after we’ve initialized enough of the logging infrastructure. Unlike for compute threads there is no `source` here. The source could be:
* `$APOLLO_ROUTER_IO_THREADS`, if specified and `apollo_router::executable::main` is used
* Anything, if a custom binary defines its own `main` creating a Tokio runtime
* `$TOKIO_WORKER_THREADS` if specified and nothing set Tokio’s `worker_threads()`
* `available_parallelism()` is the eventual default

apollo-router/src/executable.rs
SimonSapin: When I added this log it was captured in the snapshot for test `pq_layer_freeform_graphql_with_safelist_log_unknown_true` which made the outcome of the test dependent of number of available CPUs. Rather than spend time on figuring out how to make that test filter out that specific log I’ve removed the log for now.

apollo-router/src/compute_job.rs
```diff
@@ -14,18 +14,30 @@ use crate::metrics::meter_provider;
...
-/// Let this thread pool use all available resources if it can.
+/// By default, let this thread pool use all available resources if it can.
 /// In the worst case, we’ll have moderate context switching cost
 /// as the kernel’s scheduler distributes time to it or Tokio or other threads.
 fn thread_pool_size() -> usize {
-    std::thread::available_parallelism()
-        .expect("available_parallelism() failed")
-        .get()
+    // This environment variable is intentionally undocumented.
+    // See also APOLLO_ROUTER_IO_THREADS in apollo-router/src/executable.rs
```
garypen, line 22:
```suggestion
```

apollo-router/src/compute_job.rs
```diff
@@ -14,18 +14,30 @@ use crate::metrics::meter_provider;
...
+            .get();
+        tracing::info!(threads, type="compute", source="available_parallelism", "thread pool");
+        threads
+    }
 }
 
 type Job = Box<dyn FnOnce() + Send + 'static>;
 
-fn queue() -> &'static AgeingPriorityQueue<Job> {
+pub(crate) fn queue() -> &'static AgeingPriorityQueue<Job> {
```
garypen, line 40: Deos this need to be `pub(crate)` now?

SimonSapin: Good points!

garypen (approved):

apollo-router/src/compute_job.rs
```diff
@@ -14,13 +14,24 @@ use crate::metrics::meter_provider;
...
+        .and_then(|value| value.parse::<usize>().ok())
+    {
+        tracing::info!(threads, type="compute", source="env", "thread pool");
+        threads
+    } else {
+        let threads = std::thread::available_parallelism()
+            .expect("available_parallelism() failed")
+            .get();
+        tracing::info!(threads, type="compute", source="available_parallelism", "thread pool");
+        threads
```
lrlna, line 33: should we emit a metric for the number of threads available and the number of threads being set by the env var?

PR #6747, opened 2025-02-07
Add ability to specify path as regex in snapshots

pubmodmatt:
Add the ability to specify a `regex` for the URL path in a shapshot file. This is useful when constructing tests that may result in a number of HTTP requests where responses can be reused, or where the path is dynamically generated during the test (such as a randomly generated identifier).

```json
    "request": {
        "method": "GET",
        "path": null,
        "regex": "users/.*/posts",
        "body": null
    },
```

The snapshot server first looks for an exact match to the URL path in the map, then sequentially applies each `regex` if no exact match if found. Performance could degrade if there were a lot of `regex` snapshots, but this should be reasonable for a small number of snapshots.

The regular expression provided will need to escape any special characters used by the `regex` crate, such as the `?` in query parameters, using `\\`.

The `regex` is never set when taking snapshots from the upstream API. The only way to use it is to manually edit a snapshot file, typically by converting an existing snapshot by changing `path` to `regex` and replacing the variable part of the path with wilcards.

These changes are backward compatible with existing snapshot files, so there is no need to update any source controlled snapshots.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

**Notes**

### DIFF
--- a/apollo-router/src/test_harness/http_snapshot.rs
+++ b/apollo-router/src/test_harness/http_snapshot.rs
@@ -18,11 +18,22 @@
 //!
 //! Any requests made to the snapshot server will be proxied on to the given base URL, and the
 //! responses will be saved to the given file. The next time the snapshot server receives the
-//! same request (same relative path, HTTP method, and request body), it will respond with the
-//! response recorded in the file rather than sending the request to the upstream server.
+//! same request, it will respond with the response recorded in the file rather than sending the
+//! request to the upstream server.
 //!
 //! The snapshot file can be manually edited to manipulate responses for testing purposes, or to
 //! redact information that you don't want to include in source-controlled snapshot files.
+//!
+//! Requests are matched to snapshots based on the URL path, HTTP method, and base64 encoding of
+//! the request body (if there is one). If the snapshot specifies the `path` field, the URL path
+//! must match exactly. Alternatively, a snapshot containing a `regex` field will match any URL
+//! path that matches the regular expression. A snapshot with an exact `path` match takes
+//! precedence over a snapshot with `regex`. Snapshots recorded by the server will always specify
+//! `path`. The only way to use `regex` is to manually edit a snapshot file. A typical pattern is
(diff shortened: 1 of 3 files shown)


apollo-router/src/test_harness/http_snapshot.rs
```diff
@@ -320,24 +342,37 @@ fn map_headers<F: Fn(&str) -> bool>(
...
-    Ok(snapshots
-        .into_iter()
-        .map(|snapshot| (snapshot.key(), snapshot))
-        .collect())
+    let mut snapshots_by_key: BTreeMap<String, Snapshot> = Default::default();
+    let mut snapshots_by_regex: Vec<Snapshot> = Default::default();
+    for snapshot in snapshots.into_iter() {
+        if snapshot.request.regex.is_some() {
+            snapshots_by_regex.push(snapshot);
+        } else {
```
dylan-apollo, line 371: My understanding from the PR description is that you could specify _both_ path and regex, and that path would take precedence?

apollo-router/src/test_harness/http_snapshot.rs
```diff
@@ -263,23 +271,37 @@ fn response_from_snapshot(
...
+                        }
+                    }
+                }
+                None
+            })
+            .and_then(|snapshot| {
+                debug!(
+                    url = %uri,
+                    method = %method,
+                    "Found existing snapshot"
```
dylan-apollo, line 297: Should this specify which snapshot was selected, in case multiple match?

dylan-apollo (approved):
A couple questions, but LGTM

apollo-router/src/test_harness/http_snapshot.rs
pubmodmatt: I've made it an error for a snapshot to specify both `path` and `regex`. If a request matches to more than one snapshot, the one with the exact path takes precedence.

Issue #5739, opened 2024-07-29
External Coprocessor: Support for connections to be made with Unix Domain Sockets


ShaunPhillips:
#### Is your feature request related to a problem? Please describe.
Support calling the External Coprocessor over a Unix domain socket when the endpoint is deployed as a sidecar container.
https://github.com/apollographql/router/pull/4757

#### Describe the solution you'd like
Expose support for Unix domain socket configuration options for the External Coprocessor plugin via the router.yaml file.

PR #6752, opened 2025-02-10
Add better context config for coprocessor to help migration

bnjjj:
### Context configuration

You can configure what you want to send from the context to your coprocessor. By default it's disabled (`none`) and doesn't send any context keys to the coprocessor. You now have the ability to give an array
containing all the context keys you want to send to your coprocessor. Sending only a selective list of context keys will increase the performances because it will lower the size of payloads between the router and coprocessor.
You can also set `context: deprecated` if you still want to use deprecated context key names from router 1.x. If you want to pass all context keys with new naming just set `context: all`.

Example:

```yaml title="router.yaml"
coprocessor:
  url: http://127.0.0.1:3000 # mandatory URL which is the address of the coprocessor
  supergraph:
    request:
      headers: true
      context: # It will only send these 2 context keys to your coprocessor
        - apollo::supergraph::operation_name
        - apollo::demand_control::actual_cost
      body: true
    response:
      headers: true
      context: all # It will send all context keys with new names (2.x version)
      body: true
  subgraph:
    all:
      request:
        context: deprecated # It will send all the context keys with deprecated names (1.x version)
```

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


bnjjj:
@lennyburdette good catch ! I think yes that would probably be enough to just include the context key in the coprocessor config. But for now the default config is to pass all context keys. Could it be an issue if your connector keys are passed by default ?

lennyburdette:
we add the key only if you set that connectors-specific config value: https://github.com/apollographql/router/blob/dev/apollo-router/src/plugins/connectors/plugin.rs#L147-L175 ... we could switch this logic to read from your new config?

theJC:
I like this idea, but I would believe a prerequisite for your customers would be fro Apollo to supply a compendium of all possible context keys in the Router documentation.

This was added to the docs in docs/source/routing/customization/overview.mdx in this MR: https://github.com/apollographql/router/pull/6572/files#diff-e95018a62736abd100a39b77182d90ffa70565ec35cad5cb0cd366b1535f306a

But then subsequent document updates seemed to have removed it.

apollo-router/src/context/mod.rs
```diff
@@ -246,6 +303,66 @@ impl Default for Context {
...
+        COST_ESTIMATED_KEY => DEPRECATED_COST_ESTIMATED_KEY.to_string(),
+        COST_ACTUAL_KEY => DEPRECATED_COST_ACTUAL_KEY.to_string(),
+        COST_RESULT_KEY => DEPRECATED_COST_RESULT_KEY.to_string(),
+        COST_STRATEGY_KEY => DEPRECATED_COST_STRATEGY_KEY.to_string(),
+        ENABLED_CONTEXT_KEY => DEPRECATED_ENABLED_CONTEXT_KEY.to_string(),
+        FORMATTED_QUERY_PLAN_CONTEXT_KEY => DEPRECATED_FORMATTED_QUERY_PLAN_CONTEXT_KEY.to_string(),
+        QUERY_PLAN_CONTEXT_KEY => DEPRECATED_QUERY_PLAN_CONTEXT_KEY.to_string(),
+        PERSISTED_QUERY_CACHE_HIT => DEPRECATED_PERSISTED_QUERY_CACHE_HIT.to_string(),
+        PERSISTED_QUERY_REGISTERED => DEPRECATED_PERSISTED_QUERY_REGISTERED.to_string(),
+        _ => key,
```
garypen, line 332: I think it would be better to return `Result` or `Option`. Otherwise silent failure is possible. Or make the "keys" members of an `Enum` so it can't be invoked with a bad key.

apollo-router/src/context/mod.rs
```diff
@@ -246,6 +303,66 @@ impl Default for Context {
...
+        DEPRECATED_COST_ESTIMATED_KEY => COST_ESTIMATED_KEY.to_string(),
+        DEPRECATED_COST_ACTUAL_KEY => COST_ACTUAL_KEY.to_string(),
+        DEPRECATED_COST_RESULT_KEY => COST_RESULT_KEY.to_string(),
+        DEPRECATED_COST_STRATEGY_KEY => COST_STRATEGY_KEY.to_string(),
+        DEPRECATED_ENABLED_CONTEXT_KEY => ENABLED_CONTEXT_KEY.to_string(),
+        DEPRECATED_FORMATTED_QUERY_PLAN_CONTEXT_KEY => FORMATTED_QUERY_PLAN_CONTEXT_KEY.to_string(),
+        DEPRECATED_QUERY_PLAN_CONTEXT_KEY => QUERY_PLAN_CONTEXT_KEY.to_string(),
+        DEPRECATED_PERSISTED_QUERY_CACHE_HIT => PERSISTED_QUERY_CACHE_HIT.to_string(),
+        DEPRECATED_PERSISTED_QUERY_REGISTERED => PERSISTED_QUERY_REGISTERED.to_string(),
+        _ => key,
```
garypen, line 362: Same as above.

apollo-router/src/plugins/mod.rs
```diff
@@ -26,8 +26,8 @@ pub(crate) mod cache;
 pub(crate) mod connectors;
 mod coprocessor;
 pub(crate) mod csrf;
-mod demand_control;
-mod expose_query_plan;
+pub(crate) mod demand_control;
+pub(crate) mod expose_query_plan;
```
garypen, line 30: Would this still be required if we centralized the deprecated keys?

docs/source/reference/upgrade/from-router-v1.mdx
```diff
@@ -346,6 +346,37 @@ The router request context is used to share data across stages of the request pi
 - `persisted_query_hit` -> `apollo::apq::cache_hit`
 - `persisted_query_register` -> `apollo::apq::registered`
 
+### Coprocessor
+
+As we renamed several context keys it could impact your coprocessor logic and you might need to update the codebase. In order to smoothly migrate to the new context keys we changed the `context` configuration on coprocessors.
+So by default if you didn't specify the `context` configuration nothing will change. But you can now specify either if you have to still use the deprecated names (`deprecated`) or the new ones (`all`).
+We also added a new configuration, if you specify an array in `context` you'll have to specify only the context keys you want to send to your coprocessors. It will lower the body size of the request/response and then improve performances.
```
garypen, line 353:
```suggestion
We have renamed several context keys which could impact your coprocessor logic. If you don't specify the `context` configuration nothing will change. If, however, you rely on any of the impacted keys. You should update your coprocessor logic to reflect the change. If this is difficult, you can postpone this migration by relying on a new context configuration feature for coprocessors.

Prior to this change, context was a boolean argument indicating that all of context (true) or no context (false) should be passed to the coprocessor. Now it can take one of several values indicating which names, deprecated or new, you wish to pass to your coprocessor.

Finally, if you specify an array in `context` you'll have to specify only the context keys you want to send to your coprocessors. This will lower the body size of the request/response and may improve performance.
```

garypen (commented):
General Comment:

I don't like having DEPRECATED const values scattered around the code base. I'd prefer if they were isolated to a single deprecated source file.

I have a question about the selective option. Does it only work with "new" names or can you use "deprecated" names? Also, is it possible to mix and match them within the same array? Would this lead to duplication and confusion?

bnjjj:
@garypen 
> I have a question about the selective option. Does it only work with "new" names or can you use "deprecated" names? Also, is it possible to mix and match them within the same array? Would this lead to duplication and confusion?

No it's only new names

apollo-router/src/context/mod.rs
bnjjj: I'm not sure to understand the idea here. Basically it's used directly to convert a deprecated key to a current one. If the key is not listed as deprecated then I just use the original one so it's not an error it's just a conversion and converting the context key to the original itself is not an error IMHO

bnjjj:
Good catch @theJC I added back the docs for context keys you linked

bnjjj:
@lennyburdette if it's an opt-in option that's even better. I think you're right it should check the new configuration for context and add it if it's included in the context keys list

bnjjj:
@garypen I agree with you about the deprecated context keys. I think it's better to put everything into the same module/file. Can we just consider doing this in a follow up PR ?

apollo-router/src/context/mod.rs
garypen: I was thinking it might be possible to provide a key which was a typo and so you'd get yourself back and not notice there was a problem.

garypen:
> @garypen I agree with you about the deprecated context keys. I think it's better to put everything into the same module/file. Can we just consider doing this in a follow up PR ?

It could be done in a follow-up.

apollo-router/src/context/mod.rs
bnjjj: Hmm got it. I think that's something we should do at the jsonschema level for the selective configuration. This function is only used when `deprecated` is enabled and so you don't need to provide any context key, it's directly coming from our context itself.

bnjjj:
@garypen Added in https://github.com/apollographql/router/pull/6752/commits/05bc550ad4f3b5f6fb4ef0bb75ccd142ee8c0655

apollo-router/src/context/mod.rs
garypen: Ok. It feels like a "dangerous" function. Could you add a comment that captures what you've just written to the conversion functions? That way if anyone tries to use them in future they'll be aware of the constraints.

goto-bus-stop:
Is/should this also be available for Rhai scripts?

apollo-router/src/context/mod.rs
goto-bus-stop: It'd be hard to enforce context keys that actually exist because users can add their own, but maybe we could do it for `apollo::` namespaced ones

bnjjj:
@goto-bus-stop Yes but we will do this in a followup PR because it's taking more time. And I think it's also easier to spot in a rhai script when you're using a deprecated key

apollo-router/src/context/mod.rs
bnjjj: fixed in https://github.com/apollographql/router/pull/6752/commits/50fa970adefa94c1bc67b8a5c2919f29ec73801e

garypen (approved):

theJC:
While focusing on optimizing the performance around coprocessor, can I please recommend prioritizing looking into the possibility of refactoring/DRYing as much as possible of the httpclientservice used for outbound calls to subgraphs and the similar/duplicative work done for outbound calls for the coprocessor.  That way advances/improvements/changes there are available and leveraged for both pieces of communication (as well as hopefully sharing common configuration)

Not having this consistent/common between these areas causes and will continue to cause additional efforts and asks from your customers to get them aligned as well as duplicative work for you.  

Several examples of efforts and asks to get coprocessor calling caught up with subgraph calling capabilities:

- enabling http2/h2c - I contributed this last year - https://github.com/apollographql/router/pull/5300 
- having common observability - I just sent this MR this week: https://github.com/apollographql/router/pull/6776
- supporting unix domain sockets for coprocessor - a request Indeed made last year https://github.com/apollographql/router/issues/5739

PR #6757, opened 2025-02-11
Adjust plugin order to correct interaction with telemetry & traffic shaping 

goto-bus-stop:
- Update the layer inventory document
  - This doesn't fully bring it up to date with the current state of 2.x, I think, but the plan for today will be to use this to audit the order of plugins and to document an understanding of the way requests flow through the router and the constraints on the placement of various layers.
- Move CSRF to the router service, and after traffic shaping
  - It being at the supergraph service was likely a leftover from the pre-router service era.
  - Moving this to the router service prevents doing unnecessary work like GraphQL parsing and validation if we're just going to reject the request anyways.
  - Previously, telemetry "worked" for CSRF-rejected requests because CSRF worked at the supergraph service. Now, 
CSRF happens at the router service, so it must be applied *after* telemetry in the plugin list.
  - This results in a change to some error responses. Especially, if a request sends a wrong content-type, it may now be rejected by CSRF instead of being rejected by content negotiation.
- Move body size limit to after traffic shaping.

Follow-ups (Maybe for router 2.0.1 if necessary):
- Move CORS handling to a plugin so it can be done in-between telemetry and traffic shaping
- Disable Router span creation for CORS requests


apollo-router/src/plugins/csrf/mod.rs
```diff
@@ -27,6 +26,7 @@ pub(crate) struct CSRFConfig {
     /// set unsafe_disabled = true to disable the plugin behavior
     /// Note that setting this to true is deemed unsafe.
     /// See <https://developer.mozilla.org/en-US/docs/Glossary/CSRF>.
+    /// TODO rename this to enabled. This is in line with the other plugins and will be less confusing.
```
bnjjj, line 29: Should it be done before merging this PR ?

goto-bus-stop: No, 2.n or 3.0

apollo-router/src/router_factory.rs
```diff
@@ -675,10 +696,12 @@ pub(crate) async fn create_plugins(
             }
         }
     }
-    add_mandatory_apollo_plugin!("limits");
     add_mandatory_apollo_plugin!("license_enforcement");
     add_mandatory_apollo_plugin!("health_check");
+    // TODO: Introduce a CORS plugin here
```
goto-bus-stop, line 701: This will be a follow-up, just recording our decision on where to put it as a code comment

apollo-router/src/configuration/snapshots/apollo_router__configuration__tests__schema_generation.snap
```diff
@@ -342,7 +343,7 @@ expression: "&schema"
     },
     "CSRFConfig": {
       "additionalProperties": false,
-      "description": "CSRF Configuration.",
+      "description": "CSRF protection configuration.\n\nSee <https://owasp.org/www-community/attacks/csrf> for an explanation on CSRF attacks.",
```
garypen, line 346: Nice improvement!

garypen (approved):

PR #6761, opened 2025-02-11
Fix backpressure for compute jobs

SimonSapin:
Since version 1.58.0, Apollo Router has a dedicated thread pool for computation-heavy jobs with a priority queue. When the queue was full above a certain threshold, it attempted to exert backpressure by returning `Poll::Pending` in `tower::Service::poll_ready` for the query planner service. However this was not sufficient since `LoadShed` was not used together with `Poll::Pending`.

This PR moves to using error `Result`s to represent backpressure. As a consequence, we can now have a precise queue capacity checked when inserting instead of a "soft" capacity checked separately. Much of the diff is related to separating "permanent" errors that should be cached from "temporary" backpressure errors that should not be cached.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/ageing_priority_queue.rs
+++ b/apollo-router/src/ageing_priority_queue.rs
@@ -14,6 +14,9 @@ pub(crate) enum Priority {
     P8,
 }
 
+#[derive(Debug, Clone)]
+pub(crate) struct QueueIsFullError;
+
 const INNER_QUEUES_COUNT: usize = Priority::P8 as usize - Priority::P1 as usize + 1;
 
 /// Indices start at 0 for highest priority
@@ -34,7 +37,7 @@ where
     inner_queues:
         [(crossbeam_channel::Sender<T>, crossbeam_channel::Receiver<T>); INNER_QUEUES_COUNT],
     queued_count: AtomicUsize,
-    soft_capacity: usize,
+    capacity: usize,
 }
 
 pub(crate) struct Receiver<'a, T>
@@ -49,12 +52,12 @@ impl<T> AgeingPriorityQueue<T>
 where
     T: Send + 'static,
 {
-    pub(crate) fn soft_bounded(soft_capacity: usize) -> Self {
+    pub(crate) fn bounded(capacity: usize) -> Self {
         Self {
             // Using unbounded channels: callers must use `is_full` to implement backpressure
             inner_queues: std::array::from_fn(|_| crossbeam_channel::unbounded()),
             queued_count: AtomicUsize::new(0),
-            soft_capacity,
+            capacity,
         }
     }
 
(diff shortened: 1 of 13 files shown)


apollo-router/src/ageing_priority_queue.rs
```diff
@@ -49,28 +52,29 @@ impl<T> AgeingPriorityQueue<T>
...
+    pub(crate) fn send(&self, priority: Priority, message: T) -> Result<(), QueueIsFullError> {
+        self.queued_count
+            .fetch_update(Ordering::Relaxed, Ordering::Relaxed, |previous_count| {
+                (previous_count < self.capacity).then_some(previous_count + 1)
+            })
+            .map_err(|_| QueueIsFullError)?;
         let (inner_sender, _) = &self.inner_queues[index_from_priority(priority)];
-        inner_sender.send(message).expect("disconnected channel")
+        inner_sender.send(message).expect("disconnected channel");
+        Ok(())
```
garypen, line 77: I know this isn't new `expect`, but now that return an error from the `send` function, could we not return an error here rather than `expect` the `send` to work?

apollo-router/src/lib.rs
```diff
@@ -55,6 +55,7 @@ mod apollo_studio_interop;
 pub(crate) mod axum_factory;
 mod batching;
 mod cache;
+#[macro_use]
```
garypen, line 58: Delete this line.

SimonSapin: You’re right, this was for a previous iteration where I had a custom `Result`-like type and a `try!-like macro (with the role of the `?` operator). This did not pan out because Tower really wants `Result`.

apollo-router/src/compute_job.rs
```diff
@@ -23,9 +24,47 @@ fn thread_pool_size() -> usize {
...
+    /// Doing the same request again later would result in the same error (e.g. invalid query).
+    ///
+    /// This error can be cached.
+    PermanentError(E),
+
+    /// Doing the same request again later might work.
+    ///
+    /// This error must not be cached.
+    TemporaryError(ComputeBackPressureError),
+}
```
lrlna, line 42: We should definitely have a metric for both of these. Even just for us to know how many of the requests get turned away because the job queue is full so we can validate this work.

And speaking of, should we also have a metric to count the number of jobs in the queue? Also for our own "selfish" reasons to see if our assumptions are correct.

SimonSapin: The `PermanentError` case is generic and used in a lot of existing places, so there isn’t a single thing to count. I’m adding a counter metric for when `compute_job::execute()` returns `ComputeBackPressureError`.

We already have a `apollo.router.compute_jobs.queued` gauge metric.

apollo-router/src/ageing_priority_queue.rs
SimonSapin: I’ve made this change, but the caller in `compute_jobs.rs` now calls `expect` instead. There’s a comment explaining why this can never panic.

apollo-router/src/compute_job.rs
lrlna: thank you!!

BrynCooke (approved):

garypen (approved):

PR #6765, opened 2025-02-11
expose number of evaluated query plan options

dariuszkuc:
Before we start generating a query plan, we need to first evaluate all possible ways (options) that given query can be resolved. This PR exposes number of options that were considered to generate a final query plan.

### DIFF
--- a/apollo-federation/src/error/mod.rs
+++ b/apollo-federation/src/error/mod.rs
@@ -489,7 +489,9 @@ impl SingleFederationError {
                 ErrorCode::InterfaceKeyMissingImplementationType
             }
             SingleFederationError::DeferredSubscriptionUnsupported => ErrorCode::Internal,
-            SingleFederationError::QueryPlanComplexityExceeded { .. } => ErrorCode::QueryPlanComplexityExceededError,
+            SingleFederationError::QueryPlanComplexityExceeded { .. } => {
+                ErrorCode::QueryPlanComplexityExceededError
+            }
         }
     }
 }
--- a/apollo-federation/src/error/mod.rs
+++ b/apollo-federation/src/error/mod.rs
@@ -296,6 +296,8 @@ pub enum SingleFederationError {
     InterfaceKeyMissingImplementationType { message: String },
     #[error("@defer is not supported on subscriptions")]
     DeferredSubscriptionUnsupported,
+    #[error("{message}")]
+    QueryPlanComplexityExceeded { message: String },
 }
 
 impl SingleFederationError {
@@ -487,6 +489,7 @@ impl SingleFederationError {
                 ErrorCode::InterfaceKeyMissingImplementationType
             }
             SingleFederationError::DeferredSubscriptionUnsupported => ErrorCode::Internal,
+            SingleFederationError::QueryPlanComplexityExceeded { .. } => ErrorCode::QueryPlanComplexityExceededError,
         }
     }
 }
@@ -1453,6 +1456,15 @@ static UNSUPPORTED_FEDERATION_DIRECTIVE: LazyLock<ErrorCodeDefinition> = LazyLoc
     )
 });
 
+static QUERY_PLAN_COMPLEXITY_EXCEEDED: LazyLock<ErrorCodeDefinition> = LazyLock::new(|| {
+    ErrorCodeDefinition::new(
+        "QUERY_PLAN_COMPLEXITY_EXCEEDED".to_owned(),
+        "Indicates that provided query has too many possible ways to generate a plan and cannot be planned in a reasonable amount of time"
+            .to_owned(),
+        None,
+    )
+});
+
 #[derive(Debug, strum_macros::EnumIter)]
 pub enum ErrorCode {
     Internal,
@@ -1534,6 +1546,7 @@ pub enum ErrorCode {
     InterfaceKeyMissingImplementationType,
     UnsupportedFederationVersion,
     UnsupportedFederationDirective,
+    QueryPlanComplexityExceededError,
 }
 
 impl ErrorCode {
@@ -1633,6 +1646,7 @@ impl ErrorCode {
             }
             ErrorCode::UnsupportedFederationVersion => &UNSUPPORTED_FEDERATION_VERSION,
             ErrorCode::UnsupportedFederationDirective => &UNSUPPORTED_FEDERATION_DIRECTIVE,
+            ErrorCode::QueryPlanComplexityExceededError => &QUERY_PLAN_COMPLEXITY_EXCEEDED,
         }
     }
 }
--- a/apollo-federation/src/query_plan/query_planner.rs
+++ b/apollo-federation/src/query_plan/query_planner.rs
@@ -167,7 +167,7 @@ impl Default for QueryPlannerDebugConfig {
 #[derive(Debug, PartialEq, Default, Serialize)]
 pub struct QueryPlanningStatistics {
     pub evaluated_plan_count: Cell<usize>,
-    pub evaluated_plan_options: Cell<usize>,
+    pub evaluated_plan_paths: Cell<usize>,
 }
 
 #[derive(Debug, Default, Clone)]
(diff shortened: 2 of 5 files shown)


apollo-federation/src/error/mod.rs
```diff
@@ -296,6 +296,8 @@ pub enum SingleFederationError {
     InterfaceKeyMissingImplementationType { message: String },
     #[error("@defer is not supported on subscriptions")]
     DeferredSubscriptionUnsupported,
+    #[error("{message}")]
+    QueryPlanComplexityExceeded { message: String },
```
TylerBloom, line 300: I like this!

apollo-federation/src/query_plan/query_planning_traversal.rs
```diff
@@ -417,14 +418,21 @@ impl<'a: 'b, 'b> QueryPlanningTraversal<'a, 'b> {
                 no_followups = true;
                 break;
             }
+
+            // capture options in statistics
+            let options_count = &self.parameters.statistics.evaluated_plan_options;
+            options_count.set(options_count.get() + followups_for_option.len());
```
TylerBloom, line 424: Nit: you can call `.update(|count| count + followups_for_option.len())` rather than `.set`

dariuszkuc: sadly it is still experimental -> https://doc.rust-lang.org/std/cell/struct.Cell.html#method.update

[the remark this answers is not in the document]
sachindshinde: Specifically, you'd be looking to compute
```rust
followups_for_option.iter().map(|p| p.len()).sum()
```
instead of
```rust
followups_for_option.len()
```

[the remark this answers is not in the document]
dariuszkuc: [e0b353f](https://github.com/apollographql/router/pull/6765/commits/e0b353f664a3aa9d66833d83938b408b76124fac)

lrlna (commented):
ooooo can we have a metric in the router that tracks these as well? it can probably follow the example of max_evaluated_plans

apollo-federation/src/query_plan/query_planning_traversal.rs
TylerBloom: Ah, darn. I missed that :(

dariuszkuc:
@lrlna new metric is added to the `query_planner_service` as `apollo.router.query_planning.plan.evaluated_query_options`

apollo-router/src/query_planner/query_planner_service.rs
```diff
@@ -324,6 +327,11 @@ impl QueryPlannerService {
                 "Number of query plans evaluated for a query before choosing the best one",
                 evaluated_plan_count
             );
+            u64_histogram!(
+                "apollo.router.query_planning.plan.evaluated_query_options",
```
lrlna, line 331: to match the one right above this
```suggestion
                "apollo.router.query_planning.plan.evaluated_options",
```

TylerBloom (approved):

apollo-router/src/query_planner/query_planner_service.rs
```diff
@@ -560,6 +568,7 @@ pub(crate) struct QueryPlanResult {
     pub(super) formatted_query_plan: Option<Arc<String>>,
     pub(super) query_plan_root_node: Option<Arc<PlanNode>>,
     pub(super) evaluated_plan_count: u64,
+    pub(super) evaluated_plan_options: u64,
```
sachindshinde, line 571: Similar name change here.

sachindshinde (approved):
Just some nits below, but after that it should be good to merge.

apollo-federation/src/query_plan/query_planning_traversal.rs
```diff
@@ -417,14 +418,23 @@ impl<'a: 'b, 'b> QueryPlanningTraversal<'a, 'b> {
                 no_followups = true;
                 break;
             }
+
+            let evaluated_options_count = &self.parameters.statistics.evaluated_plan_options;
+            let simultaneous_indirect_path_count: usize =
+                followups_for_option.iter().map(|p| p.paths.0.len()).sum();
+            evaluated_options_count
+                .set(evaluated_options_count.get() + simultaneous_indirect_path_count);
+
```
duckki, line 427: A few lines below, we have `new_options.len() > options_limit`. Maybe we should limit `simultaneous_indirect_path_count > options_limit`? (Also, we may rename `options_limit` to `paths_limit`, too.)

apollo-router/src/query_planner/query_planner_service.rs
dariuszkuc: [4670a4f](https://github.com/apollographql/router/pull/6765/commits/4670a4f23affdc34fb6b4df65e0ee152a441c96f)

apollo-router/src/query_planner/query_planner_service.rs
```diff
@@ -324,6 +327,11 @@ impl QueryPlannerService {
                 "Number of query plans evaluated for a query before choosing the best one",
                 evaluated_plan_count
             );
+            u64_histogram!(
+                "apollo.router.query_planning.plan.evaluated_options",
+                "Number of different ways to plan a query evaluated before starting to generate a plan",
```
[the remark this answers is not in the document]
dariuszkuc: [4670a4f](https://github.com/apollographql/router/pull/6765/commits/4670a4f23affdc34fb6b4df65e0ee152a441c96f)

apollo-federation/src/query_plan/query_planner.rs
```diff
@@ -167,6 +167,7 @@ impl Default for QueryPlannerDebugConfig {
 #[derive(Debug, PartialEq, Default, Serialize)]
 pub struct QueryPlanningStatistics {
     pub evaluated_plan_count: Cell<usize>,
+    pub evaluated_plan_options: Cell<usize>,
```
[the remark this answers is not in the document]
duckki: +1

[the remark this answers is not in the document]
dariuszkuc: [4670a4f](https://github.com/apollographql/router/pull/6765/commits/4670a4f23affdc34fb6b4df65e0ee152a441c96f)

apollo-federation/src/query_plan/query_planning_traversal.rs
dariuszkuc: I wouldn't change existing functionality in this PR. Those limits are magic numbers that we shouldn't expose to users. Instead we should be looking into more "natural" limits like time.

from the docs
```rs
    /// Before creating query plans, for each path of fields in the query we compute all the
    /// possible options to traverse that path via the subgraphs. Multiple options can arise because
    /// fields in the path can be provided by multiple subgraphs, and abstract types (i.e. unions
    /// and interfaces) returned by fields sometimes require the query planner to traverse through
    /// each constituent object type. The number of options generated in this computation can grow
    /// large if the schema or query are sufficiently complex, and that will increase the time spent
    /// planning.
    ///
    /// This config allows specifying a per-path limit to the number of options considered. If any
    /// path's options exceeds this limit, query planning will abort and the operation will fail.
    ///
    /// The default value is None, which specifies no limit.
    pub paths_limit: Option<u32>,
```

sachindshinde: I mentioned it [here](https://github.com/apollographql/router/pull/6765#discussion_r1951217354), but that would be a breaking change.

We could impose a new limit on `simultaneous_indirect_path_count`, but we'd have to do some checking to see what an appropriate limit value would be.

sachindshinde: As @dariuszkuc mentioned as well, it's not a number that users intuitively know how to set.

duckki: 👍

dariuszkuc:
@mergify backport 1.x

PR #6766, opened 2025-02-11
Remove leading 0s before parsing into `serde_json::Number`

benjamn:
We previously decided to allow leading 0s, per [this discussion](https://github.com/apollographql/router/pull/5762#discussion_r1711807550).

However, since we use the parsing logic of `serde_json::Number` to produce the internal representation, and that logic fails given leading zeros, we need to normalize the input `&str` a bit before calling `number.parse()`. We already perform some similar normalizations, e.g. ensuring the fractional part is at least a 0 (not empty after the `.`), so there's precedent.

Thanks to @nicholascioli for finding this problem using fuzz testing!

### DIFF
--- a/apollo-federation/src/sources/connect/json_selection/lit_expr.rs
+++ b/apollo-federation/src/sources/connect/json_selection/lit_expr.rs
@@ -96,7 +96,14 @@ impl LitExpr {
                         );
 
                         let mut s = String::new();
-                        s.push_str(int.fragment());
+                        let int_str = int.fragment();
+                        // Remove leading zeros to avoid failing the stricter
+                        // number.parse() below, but allow a single zero.
+                        if let Some(first_non_zero) = int_str.chars().position(|c| c != '0') {
+                            s.push_str(&int_str[first_non_zero..]);
+                        } else {
+                            s.push('0');
+                        }
 
                         let full_range = if let Some((_, dot, _, frac)) = frac {
                             let frac_range = merge_ranges(
@@ -277,6 +284,7 @@ mod tests {
     use crate::sources::connect::json_selection::location::new_span;
     use crate::sources::connect::json_selection::PathList;
 
+    #[track_caller]
     fn check_parse(input: &str, expected: LitExpr) {
         match LitExpr::parse(new_span(input)) {
             Ok((remainder, parsed)) => {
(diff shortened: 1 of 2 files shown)


apollo-federation/src/sources/connect/json_selection/lit_expr.rs
```diff
@@ -317,6 +325,24 @@ mod tests {
             "-123.",
             LitExpr::Number(serde_json::Number::from_f64(-123.0).unwrap()),
         );
+        check_parse("00", LitExpr::Number(serde_json::Number::from(0)));
+        check_parse(
+            "-00",
+            LitExpr::Number(serde_json::Number::from_f64(-0.0).unwrap()),
+        );
```
benjamn, line 332: I guess negative zero has to be a float, since an integer wouldn't be able to represent the negativity?

apollo-federation/src/sources/connect/json_selection/lit_expr.rs
```diff
@@ -96,7 +96,14 @@ impl LitExpr {
                         );
 
                         let mut s = String::new();
-                        s.push_str(int.fragment());
+                        let int_str = int.fragment();
+                        // Remove leading zeros to avoid failing the stricter
+                        // number.parse() below, but allow a single zero.
+                        if let Some(first_non_zero) = int_str.chars().position(|c| c != '0') {
```
nicholascioli, line 102: This is a super nit, but apparently `Iterator::position` [can panic](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.position) if the string is long enough to overflow `usize::MAX`. I don't think that we need to handle this, but if so we should probably make use of [Iterator::skip_while](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.skip_while) and the like.

apollo-federation/src/sources/connect/json_selection/lit_expr.rs
nicholascioli: That sounds right!

apollo-federation/src/sources/connect/json_selection/lit_expr.rs
```diff
@@ -317,6 +325,24 @@ mod tests {
...
+            LitExpr::Number(serde_json::Number::from_f64(-0.0).unwrap()),
+        );
+        check_parse(" 00 ", LitExpr::Number(serde_json::Number::from(0)));
+        check_parse(" 0 ", LitExpr::Number(serde_json::Number::from(0)));
+        check_parse(
+            " - 0 ",
+            LitExpr::Number(serde_json::Number::from_f64(-0.0).unwrap()),
+        );
+        check_parse("001", LitExpr::Number(serde_json::Number::from(1)));
+        check_parse("0010", LitExpr::Number(serde_json::Number::from(10)));
```
nicholascioli, line 345: Just in case, could we add a test case for leading zeros on fractional numbers? Like `-007.` (actually generated in one of the fuzz cases) or `000.10`?

nicholascioli (approved):
LGTM! I added a few nits, but feel free to ignore.

apollo-federation/src/sources/connect/json_selection/lit_expr.rs
benjamn: https://github.com/apollographql/router/pull/6766/commits/bfccf2a3a0f8fa7ebfbc495a4c8e616ddb3443c5

apollo-federation/src/sources/connect/json_selection/lit_expr.rs
benjamn: https://github.com/apollographql/router/pull/6766/commits/bfccf2a3a0f8fa7ebfbc495a4c8e616ddb3443c5

PR #6771, opened 2025-02-11
Add more observability around WS handshake failures

john-markham:
*Description here*

Fixes #**issue_number**

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/services/subgraph_service.rs
+++ b/apollo-router/src/services/subgraph_service.rs
@@ -1,6 +1,7 @@
 //! Tower fetcher for subgraphs.
 
 use std::collections::HashMap;
+use std::error::Error;
 use std::sync::atomic::AtomicBool;
 use std::sync::atomic::Ordering::Relaxed;
 use std::sync::Arc;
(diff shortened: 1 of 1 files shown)


apollo-router/src/services/subgraph_service.rs
```diff
@@ -628,9 +628,62 @@ async fn call_websocket(
...
+
+            other_error => format!("{other_error}"),
+        };
+
+        tracing::warn!(
+            error.type   = "websocket_connection_failed",
+            error.details= %error_details,
+            error.source = %std::any::type_name_of_val(&err),
+            "WebSocket connection failed"
+        );
```
bnjjj, line 681: It looks good but I don't think this part is needed as we're returning the `SubrequestWsError` we should be able to display it using custom telemetry configuration. Also might be worth to add the `source` to `SubrequestWsError`

john-markham: We can leave adding additional metadata to `SubrequestWsError` to a followup

bnjjj:
Thanks @john-markham closing this PR in favor of https://github.com/apollographql/router/pull/6798

PR #6772, opened 2025-02-11
Set up stubs for SchemaBlueprint

tninesling:
Begin porting blueprints from JS. The JS version extends FederationBlueprint from SchemaBlueprint. With this setup, we'll have to call the DefaultBlueprint from the FederationBlueprint in lieu of `super`.

Omitted `validationRules(): readonly SDLValidationRule[]` because apollo-rs doesn't allow passing in a set of rules to validate in the same way as graphql-js. Additionally, type signatures using `GraphQLError` or `GraphQLError[]` use `FederationError`, which can represent multiple errors. The JS code marks success with an empty error array, so I've used `Result<(), FederationError>` for `GraphQLError[]`.

JS SchemaBlueprint: https://github.com/apollographql/federation/blob/main/internals-js/src/definitions.ts#L891
JS FederationBlueprint: https://github.com/apollographql/federation/blob/main/internals-js/src/federation.ts#L1549

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


tninesling:
General question: should we be operating on `apollo_compiler::Schema` here, `apollo_federation::schema::FederationSchema`, or is this where we'll introduce the staged schema type we talked about during the kickoff?

clenfest (approved):

clenfest (approved):

PR #6777, opened 2025-02-12
Remove the ability to auto-upgrade a router configuration at runtime

garypen:
In Router 1.x, during configuration parsing/validation the router
upgrade migrations could be applied "auto-magically" to generate a valid
runtime representation of a config for the life of the executing
process.
    
This can be problematic, since it makes de-bugging and support more
challenging than we'd like.
    
We want to retain the ability to upgrade a configuration using
migrations, but we don't want to auto-magically do this during runtime.
It should be the user's decision to make an explicit upgrade.
    
This is what this PR does. Various impacted tests are updated.
    
I also updated the link to the "migration" guide to be a link to the
"upgrade" guide at its new location.

### DIFF
--- a/docs/source/reference/router/configuration.mdx
+++ b/docs/source/reference/router/configuration.mdx
@@ -1251,12 +1251,7 @@ After you generate the schema, configure your text editor. Here are the instruct
 
 New releases of the router might introduce breaking changes to the [YAML config file's](#yaml-config-file) expected format, usually to extend existing functionality or improve usability.
 
-**If you run a new version of your router with a configuration file that it no longer supports:**
-
-1. The router emits a warning on startup.
-2. The router attempts to translate your provided configuration to the new expected format.
-   - If the translation succeeds without errors, the router starts up as usual.
-   - If the translation fails, the router terminates.
+**If you run a new version of your router with a configuration file that it no longer supports, it emits a warning on startup and terminates.** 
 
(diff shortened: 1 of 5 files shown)


BrynCooke (approved):

docs/source/reference/router/configuration.mdx
```diff
@@ -1251,12 +1251,7 @@ After you generate the schema, configure your text editor. Here are the instruct
 
 New releases of the router might introduce breaking changes to the [YAML config file's](#yaml-config-file) expected format, usually to extend existing functionality or improve usability.
 
-**If you run a new version of your router with a configuration file that it no longer supports:**
-
-1. The router emits a warning on startup.
-2. The router attempts to translate your provided configuration to the new expected format.
-   - If the translation succeeds without errors, the router starts up as usual.
-   - If the translation fails, the router terminates.
+**If you run a new version of your router with a configuration file that it no longer supports, it emits a warning on startup and terminates.**
```
lrlna, line 1254:
```suggestion
**If you run a new version of your router with a configuration file that is no longer supports, it emits a warning on startup and terminates.** 
```

lrlna (approved):

docs/source/reference/router/configuration.mdx
garypen: I think `it` is correct. The alternative would be `is no longer supported`.

PR #6782, opened 2025-02-12
Fix backpressure for compute jobs (backport #6761)

SimonSapin:
Since version 1.58.0, Apollo Router has a dedicated thread pool for computation-heavy jobs with a priority queue. When the queue was full above a certain threshold, it attempted to exert backpressure by returning `Poll::Pending` in `tower::Service::poll_ready` for the query planner service. However this was not sufficient since `LoadShed` was not used together with `Poll::Pending`.

This PR moves to using error `Result`s to represent backpressure. As a consequence, we can now have a precise queue capacity checked when inserting instead of a "soft" capacity checked separately. Much of the diff is related to separating "permanent" errors that should be cached from "temporary" backpressure errors that should not be cached.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [x] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

<hr>This is an automatic backport of pull request #6761 done by [Mergify](https://mergify.com).

### DIFF
--- a/apollo-router/Cargo.toml
+++ b/apollo-router/Cargo.toml
@@ -366,6 +366,12 @@ name = "samples"
 path = "tests/samples_tests.rs"
 harness = false
 
+[[test]]
+# This test is separated because it needs to run in a dedicated process.
+# nextest does this by default, but using a separate [[test]] also makes it work with `cargo test`.
+name = "compute_backpressure"
+path = "tests/compute_backpressure.rs"
+
 [[bench]]
 name = "huge_requests"
 harness = false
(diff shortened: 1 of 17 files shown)


SimonSapin:
Closing as benchmarking this PR v.s. the current `1.x` branch yielded mixed results: latency improved for some schemas but worsened for others. The status quo since 1.58 seems to be working well enough in practice. The path to improvement is Router 2.x, where the equivalent PR was merged: https://github.com/apollographql/router/pull/6761.

PR #6783, opened 2025-02-12
Hide leading `...` for inline `PathWithSubSelection`

benjamn:
My PR #6458 originally included support for conditional inline selections, which would have required a leading `...` for clarity, unlike the existing [`PathWithSubSelection`](https://github.com/apollographql/router/blob/dev/apollo-federation/src/sources/connect/json_selection/README.md#pathwithsubselection-) syntax.

We decided to hold back conditional selection support for the GA release, so I removed it from that PR, but I missed a spot in the pretty-printing code. I've now cleaned that up and added a regression test, with some comments about what we might do in the future, when we get back to implementing conditional selections.

Thanks to @nicholascioli for catching this oversight with fuzz testing.

### DIFF
--- a/apollo-federation/src/sources/connect/json_selection/pretty.rs
+++ b/apollo-federation/src/sources/connect/json_selection/pretty.rs
@@ -293,14 +293,17 @@ impl PrettyPrintable for NamedSelection {
                     result.push_str(sub.as_str());
                 }
             }
-            Self::Path {
-                inline,
-                alias,
-                path,
-            } => {
-                if *inline {
-                    result.push_str("... ");
-                }
+            Self::Path { alias, path, .. } => {
+                // TODO Once we reintroduce conditional selections, I believe we
+                // should print the ... even for PathWithSubSelection selections
+                // which were not originally written with the ... (but for which
+                // *inline is nevertheless true), because the version with ...
+                // will be equivalent to the version without, and using the ...
+                // makes it much more obvious that the output of the selection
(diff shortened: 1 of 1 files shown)


.changesets/fix_uniform_buoy_manuscript_bathing.md
```diff
@@ -0,0 +1,9 @@
+### Hide leading `...` for inline `PathWithSubSelection` ([PR #6783](https://github.com/apollographql/router/pull/6783))
+
+My PR #6458 originally included support for conditional inline selections, which would have required a leading `...` for clarity, unlike the existing [`PathWithSubSelection`](https://github.com/apollographql/router/blob/dev/apollo-federation/src/sources/connect/json_selection/README.md#pathwithsubselection-) syntax.
```
dylan-apollo, line 3: I don't think we want a changeset for this because this will end up in the user-facing changelog / GitHub release for 2.0.0.

dylan-apollo (changes requested):

.changesets/fix_uniform_buoy_manuscript_bathing.md
dylan-apollo: That check shouldn't be required, should just provide the suggestion. I've been ignoring it when the change isn't user-facing in the router (since the composition changelog also goes elsewhere)

dylan-apollo (approved):

PR #6784, opened 2025-02-12
Allow naked `LitExpr::LitPath` in expression parsing context

benjamn:
PR #5994 introduced the `$(...)` wrapper for injecting `LitExpr` JSON expression literals into selection syntax. Along with `->` method arguments, the `$(...)` wrapper is one of two ways to enter an expression parsing context, where the syntax to be parsed is closer to JSON than to GraphQL, and thus more convenient for crafting new/input data, as compared with the selection syntax, which is oriented towards selecting/transforming existing data.

Once parsing has entered the expression context, you shouldn't need to keep using `$(...)` to wrap JSON expressions. For many literal expressions, this is true: for example, you can write `$({ a: 1, b: 2 })` rather than `$({ a: $(1), b: $(2) })`. However, if you wanted to apply a `->` method or grab a `.key` from a JSON literal expression, the grammar previously required you to wrap that expression with `$(...)`, as in `$({ slice: $("asdf")->slice(0, 2) })`. This PR allows that expression to be written without the inner `$("asdf")` wrapper, giving `$({ slice: "asdf"->slice(0, 2) })`, which is arguably easier to read.

This new syntax works for any JSON literal, not just strings, so you can have `1234->add(1111)` or even `{ a: 1, b: 2 }.b` or `[1, 2, 3]->last` as `LitPath` expressions. Two cases here are worth highlighting:

1. Previously, a string literal expression with a path, like `"asdf"->first`, would have been interpreted/parsed as a quoted field selection, equivalent to `$.asdf->first`, meaning the `"asdf"` is not really a string literal, and `"asdf"->first` probably would not return the string `"a"`. The quoted field syntax is useful for non-identifier field names, but in my opinion the string literal interpretation is much less surprising (in the `LitExpr` parsing mode) than the field interpretation. This PR changes the meaning of that syntax, but only within the expression parsing context. If you want to select a quoted field now (in an expression context), you can still write `$."quoted field"->last`, using the `$.` to disambiguate, just as you can do in a selection context. Three other JSON literal have a similar ambiguity: `null->...`, `true->...` and `false->...` refer to input _fields_ called `null`, `true`, and `false` (respectively) in the outer selection syntax, but now refer to the corresponding JSON values in an expression context (again, much less surprising than before, IMO).

2. JSON number literals are allowed to be negative, using the `-` unary operator like most programming languages. However, if you write `-1->add(10)` in an expression context, it's important to know the `->` has a lower precedence than the `-`, so the `-1` gets parsed first, and then `->add(10)` operates on that negative value, giving 9 rather than -11. You might expect -11 if you're used to `-` having a lower precedence, negating the whole expression at the end, but that's not how it works here. The `LitPath` expression `-1->add(10)` is equivalent to `$(-1)->add(10)`, in other words.

Thanks to @nicholascioli for noticing that `"quoted field"->first` was previously mis-parsing in an expression context, because our greedy parser was parsing `"quoted field"` as a string literal and then failing on the unexpected `->first` suffix. Among the other benefits, this PR should fix that problem.

### DIFF
--- a/apollo-federation/src/sources/connect/json_selection/README.md
+++ b/apollo-federation/src/sources/connect/json_selection/README.md
@@ -99,7 +99,8 @@ PathStep             ::= "." Key | "->" Identifier MethodArgs?
 Key                  ::= Identifier | LitString
 Identifier           ::= [a-zA-Z_] NO_SPACE [0-9a-zA-Z_]*
 MethodArgs           ::= "(" (LitExpr ("," LitExpr)* ","?)? ")"
-LitExpr              ::= LitPrimitive | LitObject | LitArray | PathSelection
+LitExpr              ::= LitPath | LitPrimitive | LitObject | LitArray | PathSelection
+LitPath              ::= (LitPrimitive | LitObject | LitArray) PathStep+
 LitPrimitive         ::= LitString | LitNumber | "true" | "false" | "null"
 LitString            ::= "'" ("\\'" | [^'])* "'" | '"' ('\\"' | [^"])* '"'
 LitNumber            ::= "-"? ([0-9]+ ("." [0-9]*)? | "." [0-9]+)
@@ -672,25 +673,46 @@ The `->echo` method is still useful when you want to do something with the input
 value (which is bound to `@` within the echoed expression), rather than ignoring
 the input value (using `@` nowhere in the expression).
 
-The `$(...)` syntax can be useful within a `LitExpr` as well:
+#### The difference between `array.field->map(...)` and `$(array.field)->map(...)`
 
-```graphql
-# $(-1) needs wrapping in order to apply the ->mul method
-suffix: results.slice($(-1)->mul($args.suffixLength))
+When you apply a field selection to an array (as in `array.field`), the field
+selection is automatically mapped over each element of the array, producing a
+new array of all the field values.
 
-# Instead of something like this:
-# suffix: results.slice($->echo(-1)->mul($args.suffixLength))
+If the field selection has a further `->method` applied to it (as in
+`array.field->map(...)`), the method will be applied to each of the resulting
+field values _individually_, rather than to the array as a whole, which is
+probably not what you want given that you're using `->map` (unless each field
+value is an array, and you want an array of all those arrays, after mapping).
+
+The `$(...)` wrapping syntax can be useful to control this behavior, because it
+allows writing `$(array.field)->map(...)`, which provides the complete array of
+field values as a single input to the `->map` method:
+
+```json
+// Input JSON
+{
+  "array": [
+    { "field": 1 },
+    { "field": 2 },
+    { "field": 3 }
+  ]
+}
 ```
 
-In fairness, due to the commutavity of multiplication, this particular case
-could have been written as `suffix: results.slice($args.suffixLength->mul(-1))`,
-but not all methods allow reversing the input and arguments so easily, and this
-syntax works in part because it still parenthesizes the `-1` literal value,
-forcing `LitExpr` parsing, much like the new `ExprPath` syntax.
+```graphql
+# Produces [2, 4, 6] by doubling each field value
+doubled: $(array.field)->map(@->mul(2))
 
-When you don't need to apply a `.key` or `->method` to a literal value within a
-`LitExpr`, you do not need to wrap it with `$(...)`, so the `ExprPath` syntax is
-relatively uncommon within `LitExpr` expressions.
+# Produces [[2], [4], [6]], since ->map applied to a non-array produces a
+# single-element array wrapping the result of the mapping expression applied
+# to that individual value
+nested: array.field->map(@->mul(2))
+```
+
+In this capacity, the `$(...)` syntax is useful for controlling
+associativity/grouping/precedence, similar to parenthesized expressions in other
+programming languages.
 
 ### `PathStep ::=`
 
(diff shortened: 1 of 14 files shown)


apollo-federation/src/sources/connect/json_selection/README.md
```diff
@@ -99,7 +99,8 @@ PathStep             ::= "." Key | "->" Identifier MethodArgs?
 Key                  ::= Identifier | LitString
 Identifier           ::= [a-zA-Z_] NO_SPACE [0-9a-zA-Z_]*
 MethodArgs           ::= "(" (LitExpr ("," LitExpr)* ","?)? ")"
-LitExpr              ::= LitPrimitive | LitObject | LitArray | PathSelection
+LitExpr              ::= LitPath | LitPrimitive | LitObject | LitArray | PathSelection
+LitPath              ::= (LitPrimitive | LitObject | LitArray) PathStep+
```
benjamn, line 103: It's important that `LitPath` is tried first in the list of `LitExpr` alternatives, because a `LitPath` always has a prefix that looks like one of the other JSON value types.

benjamn: It's also important that we never allow a `PathSelection` to appear at the beginning of the `LitPath` (only `(LitPrimitive | LitObject | LitArray)`), because then the boundary between the `PathSelection` and the `PathStep+` suffix would be ambiguous.

dylan-apollo:
We'll have to block this on 0.2 now, not sure how we want to handle switching that in JSONSelection. I _think_ since this is additive/compatible, we won't _have_ to branch in the runtime, only in composition, but composition will have to make sure that if composing for 0.1 this isn't allowed. Might be easier to just branch in both... but what do you think?

apollo-federation/src/sources/connect/json_selection/lit_expr.rs
```diff
@@ -39,11 +42,39 @@ pub(crate) enum LitExpr {
...
+    }
+
+    // This method memorializes the way we (mis)parsed literal expressions in
+    // connect/v0.1, which was buggy because "hello"->slice(0, 2) would not
+    // successfully parse as a PathSelection referring to a field called
+    // "hello", but instead would parse as a string literal followed by junk
+    // (the `->slice(0, 2)` part). We should avoid changing this method
+    // behaviorally, but we can change it to adjust for different Span and
+    // ParseResult types, as long as the behavior remains the same.
+    fn parse_v0_1(input: Span) -> ParseResult<WithRange<Self>> {
```
benjamn, line 77: @dylan-apollo Great callout—I've now included ground work in this PR to allow parsing changes (like the ones this PR introduces) to be fully `ConnectSpec`-dependent, with tests that memorialize the v0.1 parsing errors that will be fixed in v0.2.

Subjectively, I don't mind this version branching style/discipline, since it affects only the parts of the parser that have version-dependent behavior, and we can keep changing `parse_v0_2` as long as we keep `parse_v0_1` unchanged.

apollo-federation/src/sources/connect/string_template.rs
```diff
@@ -52,6 +52,8 @@ impl<Const: FromStr> StringTemplate<Const> {
                     });
                 }
                 offset += 1; // Account for opening brace
+                // TODO This should call JSONSelection::parse_with_spec with a
+                // ConnectSpec, but we don't have that information handy.
                 let parsed = JSONSelection::parse(&expression).map_err(|err| {
```
benjamn, line 57: Here's a spot where I don't know yet how to obtain the `ConnectSpec` to allow calling `JSONSelection::parse_with_spec`.

apollo-federation/src/sources/connect/json_selection/lit_expr.rs
benjamn: In case you're wondering, yes, `get_connect_spec` is extracting the `ConnectSpec` from the `Span` we use as the input to all parsing functions. We don't have much control over the signatures of the parsing functions, since they may be called by `nom` parser combinators, so we cram multiple kinds of input into one parameter type (the `Span`). Fortunately, we have precedent for this kind of data bundling:
* #6089

The difference here is that we're storing a `SpanExtra` data type that implements `Clone` but not `Copy`, so we have to call `.clone()` in a few more places.

apollo-federation/src/sources/connect/string_template.rs
dylan-apollo: You can just add a parameter to `StringTemplate::parse` because its going to have to diverge its _own_ behavior soon based on spec as well, to support more types of expressions (e.g., disabling percent encoding)

apollo-federation/src/sources/connect/json_selection/lit_expr.rs
dylan-apollo: > get_connect_spec is extracting the ConnectSpec from the Span

Cool!

apollo-federation/src/sources/connect/string_template.rs
benjamn: https://github.com/apollographql/router/pull/6784/commits/e90c8a3582f55447d8a9752b91c3cba6071650e5

benjamn:
@lennyburdette I've rebased this and removed the `ConnectSpec` versioning logic, since there's no older behavior we want to preserve here. We can bring that back for other changes, of course, but this PR doesn't seem to need it.

apollo-federation/src/sources/connect/json_selection/README.md
lennyburdette: is this a comment worth leaving in the readme? or is it sufficiently covered in code comments?

lennyburdette (approved):

PR #6791, opened 2025-02-13
Bf/fix router logging

briannafugate408:
*Description here*

Fixes #**issue_number**

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/plugins/coprocessor/execution.rs
+++ b/apollo-router/src/plugins/coprocessor/execution.rs
@@ -104,7 +104,7 @@ impl ExecutionStage {
                     .map_err(|error| {
                         succeeded = false;
                         tracing::error!(
-                            "external extensibility: execution request stage error: {error}"
+                            "coprocessor: execution request stage error: {error}"
                         );
                         error
                     });
@@ -145,7 +145,7 @@ impl ExecutionStage {
                     .map_err(|error| {
                         succeeded = false;
                         tracing::error!(
-                            "external extensibility: execution response stage error: {error}"
+                            "coprocessor: execution response stage error: {error}"
                         );
                         error
                     });
--- a/apollo-router/src/plugins/coprocessor/mod.rs
+++ b/apollo-router/src/plugins/coprocessor/mod.rs
@@ -479,7 +479,7 @@ impl RouterStage {
                     .map_err(|error| {
                         succeeded = false;
                         tracing::error!(
-                            "external extensibility: router request stage error: {error}"
+                            "coprocessor: router request stage error: {error}"
                         );
                         error
                     });
@@ -518,7 +518,7 @@ impl RouterStage {
                     .map_err(|error| {
                         succeeded = false;
                         tracing::error!(
-                            "external extensibility: router response stage error: {error}"
+                            "coprocessor: router response stage error: {error}"
                         );
                         error
                     });
@@ -617,7 +617,7 @@ impl SubgraphStage {
                     .map_err(|error| {
                         succeeded = false;
                         tracing::error!(
-                            "external extensibility: subgraph request stage error: {error}"
+                            "coprocessor: subgraph request stage error: {error}"
                         );
                         error
                     });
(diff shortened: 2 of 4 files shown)


.changesets/fix_router-external-extensibility-deprecation.md
```diff
@@ -0,0 +1,4 @@
+### Brief but complete sentence that stands on its own ([Issue #ISSUE_NUMBER](https://github.com/apollographql/router/issues/ISSUE_NUMBER))
```
duckki, line 1: Please update the header `###` text :D
We can skip the Issue number if it's not associated with a public issue.

duckki (changes requested):

.changesets/fix_router-external-extensibility-deprecation.md
```diff
@@ -0,0 +1 @@
+###  We have deprecated the external extensibility and now have updated error log statements to support coprocessor. By [@briannafugate408](https://github.com/briannafugate408) in https://github.com/apollographql/router/pull/6791
```
duckki, line 1: It looks like now everything is in the headline. That is now too long.
Let's make a shorter headline with PR link (make it shorter using a hyperlink).
And keep the longer description and your name in the body.

duckki (changes requested):

abernix:
Just a very call-out (particularly for anyone who is publicly following along): external extensibility is **not** deprecated, it's merely called "coprocessor" now (as it has been since we launched it officially).

What **is** changing – and what this PR tries to fix — is we want to get rid of any remaining relics of the old name itself, which was still showing up in a couple places in our logs and our documentation.

.changesets/fix_router-external-extensibility-deprecation.md
```diff
@@ -0,0 +1,3 @@
+###  Deprecating the external extensibility (PR #6791)
+
+ We have now deprecated the external extensibility and now have updated error log statements to support coprocessor. By [@briannafugate408](https://github.com/briannafugate408)
```
abernix, line 3: Phrasing: The way this is written sounds like we are deprecating external extensibility.  This is **not** true and should be adjusted.

All this PR does is use the correct name for a feature that we are **not** deprecating, which is the feature formerly known as "external extensibility". 

I broadcast a similar message in https://github.com/apollographql/router/pull/6791#issuecomment-2737225414.

abernix (changes requested):

.changesets/fix_router-external-extensibility-deprecation.md
abernix, line 3: Also, with whatever wording you land on for the above, there's a deviation from our normal formatting above to correct — basically, this:

```suggestion
We have now deprecated the external extensibility and now have updated error log statements to support coprocessor.

By [@briannafugate408](https://github.com/briannafugate408)
```

(No leading space and `By`-line on its own line.)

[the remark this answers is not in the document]
lrlna: jinx with @abernix !

lrlna (approved):

BrynCooke (changes requested):
I think if we're going to do this we should use the `target` functionality of tracing. This will integrate better with APMs and allow log messages to be searched more easily. 

See:
https://docs.rs/tracing/latest/tracing/#configuring-attributes
https://docs.rs/tracing/latest/tracing/struct.Metadata.html#method.target

I think it would be a reasonable choice to have the target for log messages match the plugin name.

Having the target as a prefix in error messages was probably a mistake.

BrynCooke:
After discussion with the team we will raise a separate ticket to fix the target attribute of our log messages.

BrynCooke (approved):

duckki (approved):

PR #6798, opened 2025-02-14
Add more observability around WS handshake failures

bnjjj:
Follow up of #6771 to make CI works properly.

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-1083]: https://apollographql.atlassian.net/browse/ROUTER-1083?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apollo-router/src/services/subgraph_service.rs
+++ b/apollo-router/src/services/subgraph_service.rs
@@ -1,6 +1,7 @@
 //! Tower fetcher for subgraphs.
 
 use std::collections::HashMap;
+use std::error::Error;
 use std::sync::atomic::AtomicBool;
 use std::sync::atomic::Ordering::Relaxed;
 use std::sync::Arc;
--- a/apollo-router/src/services/subgraph_service.rs
+++ b/apollo-router/src/services/subgraph_service.rs
@@ -1,7 +1,6 @@
 //! Tower fetcher for subgraphs.
 
 use std::collections::HashMap;
-use std::error::Error;
 use std::sync::atomic::AtomicBool;
 use std::sync::atomic::Ordering::Relaxed;
 use std::sync::Arc;
--- a/apollo-router/src/services/subgraph_service.rs
+++ b/apollo-router/src/services/subgraph_service.rs
@@ -2723,10 +2723,10 @@ mod tests {
             )
             .await
             .unwrap_err();
-        assert_eq!(
-            err.to_string(),
-            "Websocket fetch failed from 'test': cannot connect websocket to subgraph: HTTP error: 400 Bad Request".to_string()
-        );
+
+        let err_str = err.to_string();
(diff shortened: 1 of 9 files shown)


apollo-router/src/services/subgraph_service.rs
```diff
@@ -628,9 +629,54 @@ async fn call_websocket(
...
     }
-    .map_err(|err| FetchError::SubrequestWsError {
-        service: service_name.clone(),
-        reason: format!("cannot connect websocket to subgraph: {err}"),
+    .map_err(|err| {
+        let error_details = match &err {
+            tokio_tungstenite::tungstenite::Error::Utf8 => {
+                if let Some(utf8_err) = err
+                    .source()
+                    .and_then(|e| e.downcast_ref::<std::str::Utf8Error>())
```
goto-bus-stop, line 637: Since the  `Utf8` branch holds no data, i don't think that `.source()` can ever return `Some()` here..? am I missing something?

goto-bus-stop (commented):
LGTM at first glance, just not sure that the `Utf8` `err.source()` code can do anything

apollo-router/src/services/subgraph_service.rs
bnjjj: I think it returns Some if it's the right type. It doesn't care about the data

goto-bus-stop: I meant `err.source()`, not the downcast ^^

bnjjj: I don't think it changes anything if Utf8 variant doesn't hold data

goto-bus-stop (approved):
i'm quite sure the `Utf8Error` code is unreachable but the general improvements lgtm

apollo-router/src/services/subgraph_service.rs
goto-bus-stop: `Error::Utf8` is constructed in this file: https://github.com/snapview/tungstenite-rs/blob/8cfd027674a53620f08b6ad3af255f5638eb2f51/src/protocol/message.rs#L30
there's not a `std::str::Utf8Error` involved at any point, and `Error::Utf8` does not carry a source error

bnjjj: I got your point sorry I misread. I think you're right. I'll take a deeper look at this

BrynCooke (approved):
+1 except for the utf8 thing

Issue #6673, opened 2025-01-28
Entity caching uses stringified representations within entity cache key


lleadbet:
## Describe the bug

Enttiy caching works by calculating a cache key that can be repeated for easier invalidation. Unfortunately the code: https://github.com/apollographql/router/blob/7a015fdadf416495f961fe621569580b8cf1181a/apollo-router/src/plugins/cache/entity.rs#L1337

Uses the entity _representation_, meaning that if a type uses `@requires`, it'd be included as part of the cache key. 

Furthermore, this representation is hashed as provided by either the router or user, meaning that the hash key is dependent on the order of a representation's values. For example:

```json
{
    "id": "1234",
    "id2": "5678"
}
```

is not the same as:

```json
{
  "id2": "5678",
  "id": "1234"
}
```

For purposes of hash calculation. 

This poses two problems: 

- Invalidation is unable to be easily used, as `@requires` are dynamically required as part of the request (and can be quite large payloads), meaning that users have to include all required inputs beyond keys, as well as matching the original order in which the router sent them
- The cache key includes data that isn't related to the actual key of the entity, so invalidating a single entity type isn't possible easily

## To Reproduce

For literal representation in the invalidation API: 

1. Configure the router and entity caching (with invalidation)with a Redis server
2. Use a `monitor` command to monitor the Redis commands issued by the router
3. Run a invalidation request, ordering the keys in different ways. You'll note the hash value in the SCAN command is different based on ordering

For literal representations within the cache:

1. Configure the router and entity caching (with invalidation)with a Redis server
2. Use a `monitor` command to monitor the Redis commands issued by the router
3. Have a schema that includes a type that has a `@requires` statement, and issue a request for that entity that includes the field that has `@requires` applied
5. issue a request for that entity that includes the field that has `@requires` applied
6. Issue a second request _without_ the field that has the `@requires` applied
7. Note that the hash prefixes differ

## Expected behavior

Keys are calculated without any required field data as well as ignoring ordering in invalidation requests.

bnjjj:
I'm surprised by the ordering issue, because we're using `serde_json_bytes::Value` with [`Map`](https://docs.rs/serde_json_bytes/latest/serde_json_bytes/map/struct.Map.html) which is an `IndexMap` if `preserve_order` feature is set as specified [here](https://github.com/apollographql/router/blob/dev/Cargo.toml#L72) so it should preserve the order.

bnjjj:
[...]

PR #6799, opened 2025-02-14
Entity caching: sort entity keys before hashing the key

bnjjj:
Part of https://github.com/apollographql/router/issues/6673

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/plugins/cache/entity.rs
```diff
@@ -1362,7 +1363,41 @@ fn extract_cache_keys(
 pub(crate) fn hash_entity_key(representation: &Value) -> String {
     // We have to hash the representation because it can contains PII
     let mut digest = Sha256::new();
-    digest.update(serde_json::to_string(&representation).unwrap().as_bytes());
```
BrynCooke, line 1365: Is there an impact due to the extra heap allowation because of the maps being created?

apollo-router/src/plugins/cache/entity.rs
```diff
@@ -1362,7 +1363,41 @@ fn extract_cache_keys(
 pub(crate) fn hash_entity_key(representation: &Value) -> String {
```
BrynCooke, line 1363: Meta: the implementation changed but no unit tests changed. I think we need somthing that shows that two identical representations with different orderings product the same digest.

BrynCooke (changes requested):

bnjjj:
I found a better implementation but it's waiting for https://github.com/apollographql/serde_json_bytes/pull/20 to be merged

apollo-router/src/plugins/cache/entity.rs
bnjjj: I don't create a new map anymore

apollo-router/src/plugins/cache/entity.rs
bnjjj: I added unit test

apollo-router/src/plugins/cache/invalidation.rs
```diff
@@ -230,7 +230,7 @@ pub(crate) enum InvalidationRequest {
 }
 
 impl InvalidationRequest {
-    fn key_prefix(&self) -> String {
+    fn key_prefix(&mut self) -> String {
```
goto-bus-stop, line 233: I think this requires a comment to explain why it's mut. OR, maybe even better, sort when the InvalidationRequest is constructed?
```suggestion
    /// Compute a cache key prefix. For entity keys, this destructively sorts all objects.
    fn key_prefix(&mut self) -> String {
```

apollo-router/src/plugins/cache/entity.rs
```diff
@@ -1359,9 +1359,10 @@ fn extract_cache_keys(
     Ok(res)
 }
 
-pub(crate) fn hash_entity_key(representation: &Value) -> String {
+pub(crate) fn hash_entity_key(representation: &mut Value) -> String {
```
goto-bus-stop, line 1362: Same comment as the other `&mut` site ^^
```suggestion
/// Returns a hash for the given representation, independent of field order. Destructively sorts any objects in the input.
pub(crate) fn hash_entity_key(representation: &mut Value) -> String {
```

SimonSapin (approved):

goto-bus-stop (approved):

BrynCooke (approved):

PR #6809, opened 2025-02-17
chore: Delegate setting "latest" to RELEASE_CHECKLIST.md, rather than CI.

abernix:
We have a notion of "Latest" which is represented by GitHub.  This indicates what appears on our GitHub Repo homepage, but also what our cURL installer installs when you use `latest`.  (No surprise as it's built on GitHub artifacts which also use that as [its "latest"](https://github.com/apollographql/router/releases/latest))

Because we now have two different "major" versions on the Router repository, we need to make sure they don't both compete for which is "latest".  Removing this from CI and moving this to an explicit step in `RELEASE_CHECKLIST.mdd` is a temporary solution and we can figure out how to operationalize it further next. (Which could actually involve re-introducing this commit, but only on the 2.x line, or something, but there are also other signals we could automatically react to.)

image

[ROUTER-1014]: https://apollographql.atlassian.net/browse/ROUTER-1014?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -967,17 +967,10 @@ jobs:
                   cd artifacts && sha1sum *.tar.gz > sha1sums.txt
             - run:
                 name: Create GitHub Release
+                # We always create the version as a "pre-release" version.
+                # It is promoted to latest in a separate step in RELEASE_CHECKLIST.md
                 command: >
-                  case "$VERSION" in
-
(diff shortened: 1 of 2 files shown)


abernix:
@mergify backport 2.0.0

PR #6816, opened 2025-02-18
Use explicit default on omitted telemetry endpoints

IvanGoncharov:
In the router config, if option `telemetry.exporters.metrics.otlp.endpoint` omitted or set as `default` then endpoint URI is populated by [oltp crate](https://github.com/open-telemetry/opentelemetry-rust/tree/main/opentelemetry-otlp)

This creates subtle issues because router defaults are:
https://github.com/apollographql/router/blob/5a4ee0c9198e5a4428ebe42a64f72d2a83f5b398/apollo-router/src/plugins/telemetry/otlp.rs#L29-L32

But [oltp crate](https://github.com/open-telemetry/opentelemetry-rust/tree/main/opentelemetry-otlp) defaults are:

https://github.com/open-telemetry/opentelemetry-rust/blob/1f0e361dff504e2fd0344445533a0f653292db81/opentelemetry-otlp/src/exporter/mod.rs#L59-L60

So I think it makes sense to always use router defaults in all places.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/plugins/telemetry/endpoint.rs
```diff
@@ -64,7 +64,9 @@ impl UriEndpoint {
 
             Uri::from_parts(parts)
                 .expect("uri cannot be invalid as it was constructed from existing parts")
-        })
+        } else {
+            default_endpoint.clone()
```
IvanGoncharov, line 68: It's a main change, rest is just refactoring after removing Option wrapper

apollo-router/src/plugins/telemetry/endpoint.rs
```diff
@@ -20,8 +20,8 @@ pub(crate) struct UriEndpoint {
 
 impl UriEndpoint {
     /// Converts an endpoint to a URI using the default endpoint as reference for any URI parts that are missing.
-    pub(crate) fn to_uri(&self, default_endpoint: &Uri) -> Option<Uri> {
-        self.uri.as_ref().map(|uri| {
+    pub(crate) fn extend_to_full_uri(&self, default_endpoint: &Uri) -> Uri {
```
IvanGoncharov, line 23: I tried to find a better name for something that this function is doing:
> Converts an endpoint to a URI using the default endpoint as reference for any URI parts that are missing.

apollo-router/src/plugins/telemetry/otlp.rs
```diff
@@ -189,39 +159,25 @@ fn header_map(gen: &mut schemars::gen::SchemaGenerator) -> schemars::schema::Sch
 }
 
 impl GrpcExporter {
-    // Return a TlsConfig if it has something actually set.
-    pub(crate) fn try_from(
```
IvanGoncharov, line 193: `try_from` is confusing name since it's `GrpcExporter` and `try_from` returning `ClientTlsConfig`

apollo-router/src/plugins/telemetry/otlp.rs
```diff
@@ -189,39 +159,25 @@ fn header_map(gen: &mut schemars::gen::SchemaGenerator) -> schemars::schema::Sch
...
-        if let Some(endpoint) = endpoint {
-            let endpoint = endpoint.to_string().parse::<Url>().map_err(|e| {
-                BoxError::from(format!("invalid GRPC endpoint {}, {}", endpoint, e))
-            })?;
-            let domain_name = self.default_tls_domain(&endpoint);
-
-            if self.ca.is_some()
-                || self.key.is_some()
-                || self.cert.is_some()
-                || domain_name.is_some()
```
IvanGoncharov, line 206: If we check that they are Some then it doesn't make sense to use `with` and `try_with`

apollo-router/src/plugins/telemetry/otlp.rs
```diff
@@ -78,84 +76,56 @@ impl Config {
...
-                            }
-                        }
-                    }
-                }
-                None => {
-                    endpoint_parts.path_and_query =
-                        Some(PathAndQuery::from_static(DEFAULT_HTTP_ENDPOINT_PATH));
-                }
-            }
+fn add_missing_traces_path(uri: Uri) -> Result<Uri, BoxError> {
```
IvanGoncharov, line 109: renamed it since it only operates on traces

apollo-router/src/plugins/telemetry/endpoint.rs
IvanGoncharov: Also, we already have this code in other places so it make sense to move it into a common function.

https://github.com/apollographql/router/blob/cb40e7a70429c4a279ffdddc436a9e3376640822/apollo-router/src/plugins/telemetry/tracing/jaeger.rs#L133-L134

Also it's weird that stuff like port number (if omitted) is populated from default but if you omit the entire URI or provide `default` is is populated by OLTP library and from Router's default.

apollo-router/src/plugins/telemetry/endpoint.rs
lrlna: i think `to_uri` still makes sense as a name. it's a conversion, and [mostly] follows rust's [naming conventions](https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv). the only thing that's different is that we do pass in a default value. `extend_to_full_uri` seems more confusing to me.

apollo-router/src/plugins/telemetry/otlp.rs
lrlna: hmmm I wonder if we'd be better off with a `TryFrom` implementation instead? It would require making sure the `endpoint` is also on the struct (which I think makes sense given that it looks like we almost always need it?). But I think overall would make for a more standard implementation.

For example:
```rust
impl TryFrom<GrpcExporter> for ClientTlsConfig {
    type Error = BoxError;

    fn try_from(value: GrpcExporter) -> Result<ClientTlsConfig, BoxError> {
        let endpoint = value.endpoint
            .to_string()
            .parse::<Url>()
            .map_err(|e| BoxError::from(format!("invalid GRPC endpoint {}, {}", endpoint, e)))?;
        let domain_name = value.default_tls_domain(&endpoint);

        if let (Some(ca), Some(key), Some(cert), Some(domain_name)) =
            (value.ca, value.key, value.cert, domain_name)
        {
            Ok(ClientTlsConfig::new()
                .with_native_roots()
                .domain_name(domain_name)
                .ca_certificate(Certificate::from_pem(ca.clone()))
                .identity(Identity::from_pem(cert.clone(), key.clone())))
        } else {
            // This was a breaking change in tonic where we now have to specify native roots.
            Ok(ClientTlsConfig::new().with_native_roots())
        }
    }
}
```

lrlna (approved):

apollo-router/src/plugins/telemetry/endpoint.rs
IvanGoncharov: @lrlna just to clarify, this function extends stuff based on default endpoint.
For example, you can create `Uri` without a port, but this function ensures that port is always copied from default.

For me this is surprising behaviour, since I expect `to_uri` just convert stuff to `Uri` but in reality it returns "full Uri" by extending Uri with parts from default.

apollo-router/src/plugins/telemetry/otlp.rs
BrynCooke: I forget why I didn't use TryFrom in the first place, possibly I was just new to Rust.
I think we should make the suggested change as it is the Rust way.

BrynCooke (changes requested):
Let's apply @lrlna 's suggestion

apollo-router/src/plugins/telemetry/endpoint.rs
```diff
@@ -216,53 +218,50 @@ mod test {
     }
 
     #[test]
-    fn test_to_url() {
+    fn test_extend_to_full_uri() {
+        assert_eq!(
+            UriEndpoint::default()
+                .extend_to_full_uri(&Uri::from_static("http://localhost:9411/path2")),
+            Uri::from_static("http://localhost:9411/path2")
+        );
```
IvanGoncharov, line 226: > LGTM but did you add a test to check the different behaviors ?

@bnjjj Tested here.
I also wanted to add an integration test, but I can't run mock server on the default port inside integration test :(
If `127.0.0.1:4317` is already occupied, the test will fail, and this would mean we can't run integration tests in parallel.

P.S. It would be great to run our integration tests inside docker for isolation.

apollo-router/src/plugins/telemetry/otlp.rs
IvanGoncharov: Sounds good I will make it as `TryFrom`

apollo-router/src/plugins/telemetry/endpoint.rs
bnjjj: maybe just `to_full_uri` . Usually in rust std lib when it's an `extend_*` method it takes `&mut self` instead. So either you change the prototype to take `&mut self` and keep `extend` or the other solution

IvanGoncharov: I like `to_full_uri` and agree that `extend_` is confusing 👍

lrlna: `to_full_uri` wfm

apollo-router/src/plugins/telemetry/otlp.rs
IvanGoncharov: @lrlna Thanks for writing the code. It helped a lot 👍 

I faced an issue when I tried to use it since this function accepted `endpoint` as the second argument.
I tried to put `endpoint` inside `GrpcExporter`.
However, it's problematic since this type is created by the side from `config.yaml`, and `endpoint` is one level up in the config.

I did a quick Google, and I don't see any simple method that allows me to put `endpoint` into `GrpcExporter` without complex workarounds.

Maybe we can use this function `to_tls_config` instead and accept `endpoint` as an argument?

P.S. I don't think adding complex logic to the types created to parse config YAML with Serde is a good idea. I tried to change it but quickly learned it's not something I can change in the scope of this PR 😢

lrlna: I _think_ you should be able to use `deserialize_with` on the config struct:

```
#[derive(Debug, Clone, Deserialize, JsonSchema, Default)]
#[serde(deny_unknown_fields)]
pub(crate) struct Config {
    /// gRPC configuration settings
    #[serde(default)]
    #[serde(deserialize_with = "deserialize_grpc_exporter")]
    pub(crate) grpc: GrpcExporter,
    /// ... the rest of the Config struct
}

fn deserialize_grpc_exporter<'de, D>(deserializer: D) -> Result<GrpcExporter, D::Error>
where
    D: Deserializer<'de>,
{
    Deserialize::deserialize(deserializer).map(|Config { endpoint, .. }| GrpcExporter {
        endpoint,
        ..Default::default()
    })
}
```

You'll have to double check with all the tests that this does what's expected, or otherwise tweak it a little bit. I'll add the TryFrom as a `suggestion` in a separate comment, but what this should effectively mean is that `fn exporter` now has this:

```
impl Config {
    pub(crate) fn exporter<T: From<HttpExporterBuilder> + From<TonicExporterBuilder>>(
        &self,
        kind: TelemetryDataKind,
    ) -> Result<T, BoxError> {
        match self.protocol {
            Protocol::Grpc => {
                let endpoint = self.endpoint.extend_to_full_uri(&DEFAULT_GRPC_ENDPOINT);
                let tls_config: ClientTlsConfig = self.grpc.clone().try_into()?;
                // the rest of the owl
            }
        }
    }
}
```

lrlna: Ok I can't apply suggestions to deleted lines, but effectively using a `deserialize_with` means you can have a GrpcExporter struct have `UriEndpoint` and use that in the `TryFrom`

```rust
impl TryFrom<GrpcExporter> for ClientTlsConfig {
    type Error = BoxError;

    fn try_from(value: GrpcExporter) -> Result<ClientTlsConfig, BoxError> {
        let endpoint = value.endpoint.extend_to_full_uri(&DEFAULT_GRPC_ENDPOINT);
        let endpoint = endpoint
            .to_string()
            .parse::<Url>()
            .map_err(|e| BoxError::from(format!("invalid GRPC endpoint {}, {}", endpoint, e)))?;
        let domain_name = value.default_tls_domain(&endpoint);

        if let (Some(ca), Some(key), Some(cert), Some(domain_name)) =
            (&value.ca, &value.key, &value.cert, domain_name)
        {
            Ok(ClientTlsConfig::new()
                .with_native_roots()
                .domain_name(domain_name)
                .ca_certificate(Certificate::from_pem(ca.clone()))
                .identity(Identity::from_pem(cert.clone(), key.clone())))
        } else {
            // This was a breaking change in tonic where we now have to specify native roots.
            Ok(ClientTlsConfig::new().with_native_roots())
        }
    }
}

#[derive(Debug, Clone, Deserialize, Serialize, Default, JsonSchema)]
pub(crate) struct GrpcExporter {
    /// The optional domain name for tls config.
    /// Note that domain name is will be defaulted to match the endpoint is not explicitly set.
    pub(crate) domain_name: Option<String>,
    /// The optional certificate authority (CA) certificate to be used in TLS configuration.
    pub(crate) ca: Option<String>,
    /// The optional cert for tls config
    pub(crate) cert: Option<String>,
    /// The optional private key file for TLS configuration.
    pub(crate) key: Option<String>,

    /// gRPC metadata
    #[serde(with = "http_serde::header_map")]
    #[schemars(schema_with = "header_map", default)]
    pub(crate) metadata: http::HeaderMap,

    /// gRPC endpoint
    #[serde(skip)]
    pub(crate) endpoint: UriEndpoint,
}
```

lrlna: I would just triple check with some tests that this does what we need it to do if no grpc values are set, and the grpc protocol is not set.

IvanGoncharov: ⬆️ It compiles but doesn't work as expected when I did manual tests (validation failing on the `grpc` field)
I'm pretty deep in Serde right now since I can't find good docs for `deserialize_with`.

But before spending more time, I want to raise concern that even if we fix this code it is still non-obvious how it works.

Also, this solution adds significant bulk and duplicates `endpoint` into two structures.
And all this is done to make a method that naturally accepts the second argument conform to the `TryFrom` trait. 
Also, note that this method is called in a single place.

I'm not sure it is worth the effort to add so much complexity and suggest that we would just name this method `to_tls_config` and accept `endpoint` as an argument.

@lrlna @BrynCooke What do you think about this solution?

lrlna: That makes sense @IvanGoncharov, thank you for giving this a try. I think you're right - `to_tls_config` is the function we go with.

bnjjj (approved):

lrlna (approved):

BrynCooke (approved):

IvanGoncharov:
@mergify backport 1.x

IvanGoncharov:
@mergify backport 1.x

PR #6821, opened 2025-02-18
docs: tweak language around GraphOS + OTLP trace reporting

goto-bus-stop:
In our docs, we claim that "usage reporting" to GraphOS can happen through OpenTelemetry. In reality, we have various reporting mechanisms, and only traces can use OpenTelemetry: operation metrics are for now still using our own format.

This is probably not a huge deal (and it's maybe arguable if this should even be documented in such detail..?), but these tweaks should make things more accurate.


timbotnik (approved):
Looks good thank you!

PR #6825, opened 2025-02-18
docs: add back LTS info

Meschreiber:
Only merge once 1.61 is released


abernix:
@Meschreiber you _could_ merge this into the `1.61.0` branch (now) if you want.   It'll then merge into `1.61.0` at the time that we promote 1.61.0 to official.

Is that useful?  (I am assuming that it is, potentially incorrectly, so I re-targeted this PR's base branch to `1.61.0` — which is recently created.  If it's not, feel free to change it back to `1.x`, or that'll happen automatically when the release merges on its own — thanks to GitHub's PR stacking mechanism.)  Thanks!

Meschreiber:
Thank you @abernix ! Yes, that would be useful, indeed. Would you mind approving so I can merge? Ty!

abernix (approved):
Oh. Yes. lgtm.

PR #6826, opened 2025-02-18
docs: remove request retry mention from traffic-shaping

Meschreiber:
A user noted that since `experimental_retry` was [removed from traffic shaping](https://github.com/apollographql/router/pull/6338) we should no longer mention it as a task in the [ordering section](https://www.apollographql.com/docs/graphos/routing/performance/traffic-shaping#ordering).


goto-bus-stop (approved):
thanks :)

Meschreiber:
@mergifyio backport 1.x

PR #6833, opened 2025-02-19
connectors: add parsing fuzzing

nicholascioli:
This commit adds a fuzzing target for the connector selection mapping language.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


fuzz/fuzz_targets/selection_parsing.rs
lrlna, line 1: maybe `connector_json_selection.rs` or `connector_selection.rs`? so this is not confused with selection set parsing.

fuzz/fuzz_targets/selection_parsing.rs
```diff
@@ -0,0 +1,112 @@
...
+    // seeding and iterating the Rng.
+    let Some(selection) = input.0 else {
+        return Corpus::Reject;
+    };
+
+    let parsed = JSONSelection::parse(&selection).unwrap();
+    drop(parsed);
+
+    Corpus::Keep
+});
```
lrlna, line 25: this thought would have to be a different target entirely and is non-blocking: it would interesting to see how this would behave in addition to a generated supergraph and operation. this only fuzzes a specific part of a whole - but how will it behave all together?

PR #6836, opened 2025-02-20
Upgrade to Rust 1.85

SimonSapin:
This version brings async closures, the 2024 edition, and more: https://blog.rust-lang.org/2025/02/20/Rust-1.85.0.html 

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/tests/common.rs
+++ b/apollo-router/tests/common.rs
@@ -377,7 +377,6 @@ impl Telemetry {
 #[buildstructor]
 impl IntegrationTest {
     #[builder]
-    #[allow(clippy::too_many_arguments)] // not typically used directly, only defines the builder
     pub async fn new(
         config: String,
         telemetry: Option<Telemetry>,
--- a/apollo-router/tests/common.rs
+++ b/apollo-router/tests/common.rs
@@ -553,7 +553,9 @@ impl IntegrationTest {
                         level: String,
                         message: String,
                     }
-                    let log = serde_json::from_str::<Log>(&line).expect("line: '{line}' isn't JSON, might you have some debug output in the logging?");
+                    let Ok(log) = serde_json::from_str::<Log>(&line) else {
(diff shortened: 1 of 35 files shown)


apollo-router/tests/common.rs
```diff
@@ -553,7 +552,9 @@ impl IntegrationTest {
                         level: String,
                         message: String,
                     }
-                    let log = serde_json::from_str::<Log>(&line).expect("line: '{line}' isn't JSON, might you have some debug output in the logging?");
+                    let Ok(log) = serde_json::from_str::<Log>(&line) else {
+                        panic!("line: '{line}' isn't JSON, might you have some debug output in the logging?");
+                    };
```
SimonSapin, line 557: This one was a legit bug! `{line}` was clearly intended as string formatting which `expect()` doesn’t do.

Velfi: Shouldn't there be an `if` somewhere? Can you use `else` on options/results now?

Velfi (approved):

apollo-router/tests/common.rs
SimonSapin: No, this use of [let-else](https://doc.rust-lang.org/beta/rust-by-example/flow_control/let_else.html) is intended. `let` could originally only be used with irrefutable patterns, but was then extended to allow refutable patterns like `Ok(log)` if an `else` is provided that "diverges" (never reaches the end of the else block)

The difference with `if let` is that the `log` binding belong to the outer block

Velfi: Never mind, I looked at this in the playground and it's fine. TIL.

TylerBloom (approved):

PR #6840, opened 2025-02-20
Ensure `build_query_plan()` cancels when the request cancels

SimonSapin:
When a request is cancelled, resources (compute time/memory) allocated to that request should be released. However, query planning is an exception, and currently keeps running after the request is cancelled. This PR adds cooperative cancellation to `build_query_plan()`, to ensure that when the request is cancelled, query planning errors shortly after.

### DIFF
--- a/apollo-federation/src/operation/tests/mod.rs
+++ b/apollo-federation/src/operation/tests/mod.rs
@@ -13,6 +13,7 @@ use super::SelectionKey;
 use super::SelectionSet;
 use super::normalize_operation;
 use crate::error::FederationError;
+use crate::error::SingleFederationError;
 use crate::query_graph::graph_path::OpPathElement;
 use crate::schema::ValidFederationSchema;
 use crate::schema::position::InterfaceTypeDefinitionPosition;
@@ -65,8 +66,14 @@ pub(super) fn parse_and_expand(
         fragments,
         schema,
         &Default::default(),
-        &|| Ok(()),
+        &never_cancel,
     )
+}
+
+/// The `normalize_operation()` function has a `check_cancellation` parameter that we'll want to
+/// configure to never cancel during tests. We create a convenience function here for that purpose.
+pub(crate) fn never_cancel() -> Result<(), SingleFederationError> {
+    Ok(())
 }
 
 #[test]
@@ -106,7 +113,7 @@ type Foo {
             NamedFragments::new(&executable_document.fragments, &schema),
             &schema,
             &IndexSet::default(),
-            &|| Ok(()),
+            &never_cancel,
         )
         .unwrap();
         normalized_operation.named_fragments = Default::default();
@@ -161,7 +168,7 @@ type Foo {
             NamedFragments::new(&executable_document.fragments, &schema),
             &schema,
             &IndexSet::default(),
-            &|| Ok(()),
+            &never_cancel,
         )
         .unwrap();
         normalized_operation.named_fragments = Default::default();
@@ -204,7 +211,7 @@ type Query {
             NamedFragments::new(&executable_document.fragments, &schema),
             &schema,
             &IndexSet::default(),
-            &|| Ok(()),
+            &never_cancel,
         )
         .unwrap();
 
@@ -240,7 +247,7 @@ type T {
             NamedFragments::new(&executable_document.fragments, &schema),
             &schema,
             &IndexSet::default(),
-            &|| Ok(()),
+            &never_cancel,
         )
         .unwrap();
         let expected = r#"query Test {
@@ -284,7 +291,7 @@ type T {
             NamedFragments::new(&executable_document.fragments, &schema),
             &schema,
             &IndexSet::default(),
-            &|| Ok(()),
+            &never_cancel,
         )
         .unwrap();
         let expected = r#"query Test($skipIf: Boolean!) {
@@ -331,7 +338,7 @@ type T {
             NamedFragments::new(&executable_document.fragments, &schema),
             &schema,
             &IndexSet::default(),
-            &|| Ok(()),
+            &never_cancel,
         )
         .unwrap();
         let expected = r#"query Test($skipIf: Boolean!) {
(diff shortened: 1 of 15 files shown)


apollo-federation/src/query_plan/query_planning_traversal.rs
```diff
@@ -339,6 +354,7 @@ impl<'a: 'b, 'b> QueryPlanningTraversal<'a, 'b> {
     )]
     fn find_best_plan_inner(&mut self) -> Result<Option<&BestQueryPlanInfo>, FederationError> {
         while !self.open_branches.is_empty() {
+            self.parameters.check_cancellation()?;
```
SimonSapin, line 357: Checks like this would need to be added to all loops where the planner can spend significant amount of time.

apollo-router/src/compute_job.rs
```diff
@@ -32,6 +33,31 @@ fn thread_pool_size() -> usize {
...
+impl<T> JobStatus<'_, T> {
+    /// Checks whether the oneshot receiver for the result of the job was dropped,
+    /// which means nothing is expecting the result anymore.
+    ///
+    /// This can happen if the Tokio task owning it is cancelled,
+    /// such as if a supergraph client disconnects or if a request times out.
+    ///
+    /// In this case, a long-running job should try to cancel itself
+    /// to avoid needless resource consumption.
+    pub(crate) fn check_for_cooperative_cancellation(&self) -> ControlFlow<()> {
```
dariuszkuc, line 49: Pretty cool! While tying it to the client request (disconnect/time out) is a great start - if we were to expose this feature as a user specified config value it sounds like it should be as simple as wrapping those QP tasks with a [timeout](https://docs.rs/tokio/latest/tokio/time/fn.timeout.html)?

apollo-federation/src/query_plan/query_planning_traversal.rs
duckki: I think `handle_open_branch` is the one. Other functions won't take too long. Even if `advance_with_operation_element` created a million options, it returns reasonably quickly. It's the `handle_open_branch` that actually visits them. One decision point is, `handle_open_branch` is called recursively, so we may decide whether it checks only at the top-level or always.

apollo-router/src/compute_job.rs
lrlna: I don't think we should be exposing this specifically as a feature. Cancellation at the query planning level should work in the same way as backpressure in other parts of the router. That is, query planning should adhere to the router's existing `traffic_shaping.router.timeout` configuration.

dariuszkuc: I agree that for now this feature should be tied to the request lifecycle. That being said, sometime in the future, I think we should revisit whether it would make sense to further expand on this functionality as there are some potential additional usecases (eg folks might want to still plan the query even though their request timed out with the assumption that be subsequent request would hit the cached plan, similarly we might want to have some limits when doing qp cache prewarming etc).

duckki (commented):
My concern is calling `check_for_cooperative_cancellation` too often, since It tries access memory that is not in the cpu cache. But, I may be wrong because CPU architecture might have changed since the last time I studied it :)

duckki:
Ok, I've done some reading and I'm convinced this won't slowdown our QP as it currently stands. However, once we add a timeout in the cancelation check, that will add up some inefficiently in QP. I don't have a strong opinion, but we may cut calls later as necessary.

apollo-federation/src/query_plan/query_planner.rs
```diff
@@ -348,6 +367,11 @@ impl QueryPlanner {
             NamedFragments::new(&document.fragments, &self.api_schema),
             &self.api_schema,
             &self.interface_types_with_interface_objects,
+            &|| {
+                QueryPlanningParameters::check_cancellation_with(
+                    &options.check_for_cooperative_cancellation,
+                )
```
duckki, line 373: I was wondering about `check_cancellation_with` at first. Then, I realized `normalize_operation` has to be called before  `QueryPlanningParameters` is constructed. That's unfortunate.

duckki (approved):
Approved. Only a minor suggestion to refactor some obscure closures in tests.

apollo-federation/src/operation/tests/mod.rs
```diff
@@ -60,7 +60,13 @@ pub(super) fn parse_and_expand(
         .expect("must have anonymous operation");
     let fragments = NamedFragments::new(&doc.fragments, schema);
 
-    normalize_operation(operation, fragments, schema, &Default::default())
+    normalize_operation(
+        operation,
+        fragments,
+        schema,
+        &Default::default(),
+        &|| Ok(()),
```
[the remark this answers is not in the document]
SimonSapin: I agree. A named function can be used directly like `&never_cancel` with:

```rust
fn never_cancel() -> Result<(), SingleFederationError> {
    Ok(())
}
```

apollo-federation/src/query_plan/fetch_dependency_graph.rs
```diff
@@ -3488,6 +3488,7 @@ pub(crate) fn compute_nodes_for_tree(
     initial_node_path: FetchDependencyGraphNodePath,
     initial_defer_context: DeferContext,
     initial_conditions: &OpGraphPathContext,
+    check_cancellation: &impl Fn() -> Result<(), SingleFederationError>,
```
SimonSapin, line 3491: Using `impl Fn` here makes the whole method generic and monomorphized separately for every distinct callback that can be passed. This has some compilation time and code size cost.

Assuming that `check_cancellation` isn’t called in very hot loop, would it be better to use `&dyn Fn()`? (Here and in other places with a similar argument)

apollo-federation/src/operation/tests/mod.rs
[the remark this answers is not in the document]
sachindshinde: Makes sense, I've shifted test files to using `&never_cancel` in https://github.com/apollographql/router/pull/6840/commits/083ebc1a0b4be20a826d4791b68b6dadbba49533

PR #6844, opened 2025-02-20
Update connectors request to use a String instead of RouterBody

andrewmcgivery:
Update connectors request to use a String instead of RouterBody and convert to RouterBody just-in-time to be consistent with how subgraphs are handled.

Why?

1. Simplify code
2. Unblock traffic shaping
3. Unblock coprocessors
4. Make it consistent with the existing subgraph code

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/plugins/connectors/http_json_transport.rs
+++ b/apollo-router/src/plugins/connectors/http_json_transport.rs
@@ -712,7 +712,7 @@ mod tests {
         )
         .unwrap();
 
-        assert_debug_snapshot!(req, @r###"
+        assert_debug_snapshot!(req, @r#"
         (
             Http(
                 HttpRequest {
@@ -724,14 +724,14 @@ mod tests {
                             "content-type": "application/json",
                             "content-length": "8",
                         },
-                        body: UnsyncBoxBody,
+                        body: "{\"a\":42}",
                     },
                     debug: None,
                 },
             ),
             [],
         )
-        "###);
+        "#);
 
         let TransportRequest::Http(HttpRequest { inner: req, .. }) = req.0;
         let body = body::into_string(req.into_body()).await.unwrap();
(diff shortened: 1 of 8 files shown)


apollo-router/src/services/connector/request_service.rs
```diff
@@ -274,21 +273,24 @@ impl tower::Service<Request> for ConnectorRequestService {
...
-                        )
-                        .await?;
+                        );
 
                         if let Some(http_client_service_factory) = http_client_service_factory
                             .get(&request.service_name)
                             .cloned()
                         {
+                            let (parts, body) = http_request.inner.into_parts();
+                            let final_request =
```
pubmodmatt, line 287: Minor, but I prefer just shadowing the same name in Rust if it's really a different version of the same thing, so I'd just use `http_request` again here instead of `final_request`.

apollo-router/src/plugins/connectors/http_json_transport.rs
```diff
@@ -64,19 +63,21 @@ pub(crate) fn make_request(
                         .map_err(HttpJsonTransportError::FormBodySerialization)?;
                     form_body = Some(encoded.clone());
                     let len = encoded.bytes().len();
-                    (router::body::from_bytes(encoded), len)
+                    let body_string = encoded;
```
pubmodmatt, line 66: Nit, but this temporary variable is unnecessary, just return `(encoded, len)`

pubmodmatt (approved):
Just a couple of nits, but looks good otherwise.

apollo-router/src/services/connector/request_service.rs
andrewmcgivery: Can you tell I come from a Javascript background where shadowing is a big no-no? 😆 

Updating this now!

PR #6850, opened 2025-02-21
Avoid setting env variables (not thread-safe) and other Rust 2024 edition compat

SimonSapin:
In most libc implementations and therefore also in Rust on non-Windows platforms, environment variables are a process-wide global resource without any synchronization. They are therefore not thread-safe. That is, setting or removing an environment variable while another thread is accessing them in any way (including reading) Undefined Behavior.

To reflect this, Rust is [making](https://doc.rust-lang.org/edition-guide/rust-2024/newly-unsafe-functions.html) std::env::set_var an unsafe fn in its language edition 2024, which has just become available in [version 1.85](https://blog.rust-lang.org/2025/02/20/Rust-1.85.0.html). But the underlying libc issue has always been there.

This PR fixes `set_var` related issues as well as other compatibility issues with the 2024 edition in the `apollo_router` crate. However it does not yet actually switch to the new edition, that is blocked on https://github.com/GREsau/schemars/issues/376 being fixed.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/plugins/telemetry/tracing/datadog_exporter/exporter/model/unified_tags.rs
```diff
@@ -83,41 +83,3 @@ impl UnifiedTagEnum {
         }
     }
 }
-
-#[cfg(test)]
-mod tests {
```
SimonSapin, line 88: These were effectively tests for `std::env::set_var()` and `std::env::var()`. If they were removed or mocked out, there would be so little business logic left that it's not worth IMO trying to have unit tests for it.

apollo-router/src/configuration/expansion.rs
```diff
@@ -81,16 +97,27 @@ impl Expansion {
...
+        let builder = builder.mocked_env_vars(mocked_env_vars);
+        let listen_override = Override::builder()
+            .config_path("supergraph.listen")
+            .value_type(ValueType::String);
+        let listen = *crate::executable::APOLLO_ROUTER_LISTEN_ADDRESS.lock();
+        let listen_override = if let Some(listen) = listen {
+            listen_override.value(listen.to_string()).build()
+        } else {
+            listen_override.build()
+        };
```
Velfi, line 119: Consider using a slightly simpler version:
```suggestion
        let mut listen_override = Override::builder()
            .config_path("supergraph.listen")
            .value_type(ValueType::String);
        if let Some(listen) = *crate::executable::APOLLO_ROUTER_LISTEN_ADDRESS.lock() {
            listen_override.value = listen_override.value(listen.to_string());
        }
        listen_override.build()
```

apollo-router/src/configuration/mod.rs
```diff
@@ -23,7 +23,7 @@ use regex::Regex;
 use rustls::pki_types::CertificateDer;
 use rustls::pki_types::PrivateKeyDer;
 use rustls::ServerConfig;
-use schemars::gen::SchemaGenerator;
+use schemars::r#gen::SchemaGenerator;
```
Velfi, line 26: If we can't change this now then we should create a TODO (or GitHub issue/Jira task) to rename it to `generator` in a later version.

Velfi (approved):

apollo-router/src/configuration/expansion.rs
SimonSapin: The return type of `Override::builder()` is not `Override` but a struct generated by the [`buildstructor`](https://crates.io/crates/buildstructor) proc macro crate. The builder type does not have public fields:

```rust
error[E0615]: attempted to take value of method `value` on type `configuration::expansion::__override_new_builder::__OverrideBuilder<(configuration::expansion::__override_new_builder::__Set<std::string::String>, configuration::expansion::__override_new_builder::__Optional<std::option::Option<std::string::String>>, configuration::expansion::__override_new_builder::__Optional<std::option::Option<serde_json::Value>>, configuration::expansion::__override_new_builder::__Set<configuration::expansion::ValueType>)>`
   --> apollo-router/src/configuration/expansion.rs:124:29
    |
124 |             listen_override.value = listen_override.value(listen.to_string());
    |                             ^^^^^ method, not a field
    |
    = help: methods are immutable and cannot be assigned to

PR #6854, opened 2025-02-24
Reformat apollo-router crate with rustfmt 2024 edition

SimonSapin:
Starting with Rust 1.85, the language edition 2024 is available. With it also comes the [rustfmt style edition 2024](https://doc.rust-lang.org/edition-guide/rust-2024/rustfmt-style-edition.html), with several changes to code formatting such as import sorting. By default the style edition matches the language edition, but both can be configured separately. While we cannot use the new language edition yet in the apollo-router crate (https://github.com/GREsau/schemars/issues/376),
this PR changes to the new style edition.
Aside from `apollo-router/rustfmt.toml` it is entirely generated by this command matching CI configuration:

```
cargo +nightly fmt -- --config imports_granularity=Item,group_imports=StdExternalCrate
```

(Unfortunately, import granularity and grouping config is still unstable in rustfmt.)


IvanGoncharov (approved):
Not a Rust expert but it make sense to me 👍

Velfi (approved):

abernix (commented):
I would propose that we:

* Add the commit from this PR (`10dfa657cc3ddcb356dbdbccd70e8c2741614ab8`) to `.git-blame-ignore-revs` at the root of the repository so `git blame` usage is not impacted: https://docs.github.com/en/repositories/working-with-files/using-files/viewing-and-understanding-files#ignore-commits-in-the-blame-view
* DON'T Squash this PR, but instead True-Merge.

SimonSapin:
As expected this has conflicts now so I’m gonna:

* Start over with a new branch from dev
* Reformat in a first commit
* Add that commit hash `.git-blame-ignore-revs` (and 3f62ebe8ab0adfb819a24af8bc9f2861f1ad9318 for https://github.com/apollographql/router/pull/6839) in a second commit
* Force-push to this PR
* True-merge it before it conflicts again, based on the spirit of approval of the current diff even though it’ll be a different diff

Issue #6859, opened 2025-02-25
Router v2: Gauge are not exporter to prometheus


yanns:
## Describe the bug

I'm migrating some rust plugins from router v1 to router v2.
For metrics, I'm using the new approach with:
```rust
let meter = meter_provider().meter("apollo/router");
```
I'm using counter, histograms and gauge.
Counters and histograms are exported in `/metrics` used by prometheus, as expected.
Gauges are not exported.

I still don't understand what is wrong only with Gauges.
By looking at the code, I've found it that `Gauge` is being commented out in https://github.com/apollographql/router/blob/4c7114d6af86934d46a2562430c5f9a77c170a30/apollo-router/src/plugins/telemetry/config_new/instruments.rs#L1281, but I'm not sure it's related.

## Desktop (please complete the following information):
 - Version 2.0.0

yanns:
Here is a test probably reproducing the issue https://github.com/apollographql/router/pull/6860

yanns:
In https://github.com/apollographql/router/pull/6860/commits/911780cbc46cccdae14e642912dc85633994ff87, I've tried to add support for `SyncGauge`, but the test is still failing.

yanns:
I could finally fix by adding https://github.com/apollographql/router/pull/6860/commits/b76d71bf0a0cf9c3372c1b92039ccedcdc48e6d6

PR #6860, opened 2025-02-25
reproduce issue with Gauge

yanns:
Reproduce https://github.com/apollographql/router/issues/6859


yanns:
I've tested this version of the router, and I confirm it's fixing the export of the Gauges.

goto-bus-stop:
Thanks, I think this is great, I've pushed it to a branch on the router repository so CI can run!

PR #6861, opened 2025-02-25
Implement `Deserialize` for apollo-federation query plans

SimonSapin:
Query plan data structures in the `apollo-federation` crate currently implement the `Serialize` trait but not `Deserialize`. This limitation came up twice recently, in the context of prototyping a GRPC remote planner and in the context of planner correctness testing. Fixing it is also a necessary step for making `apollo-router` use these data structures directly instead of having almost-identical ones and a conversion step.

`apollo-federation/src/query_plan/mod.rs` contains this comment on the `QueryPlan` struct:

```
// NOTE: This type implements `Serialize` for debugging purposes; however, it should not implement
// `Deserialize` until two requires are met.
// 1) `SelectionId`s and `OverrideId`s are only unique per lifetime of the application. To avoid
//    problems when caching, this needs to be changes.
// 2) There are several types transatively used in the query plan that are from `apollo-compiler`.
//    They are serialized as strings and use the `serialize` methods provided by that crate. In
//    order to implement `Deserialize`, care needs to be taken to deserialize these correctly.
//    Moreover, how we serialize these types should also be revisited to make sure we can and want
//    to support how they are serialized long term (e.g. how `DirectiveList` is serialized can be
//    optimized).
```

For these two points:

1. Today’s query plan data structures do not contain `SelectionId` nor `OverrideId`, so this comment is no longer relevant.
2. In the case of `FetchNode::operation_document`, the same problem exists in `apollo-router`, and is solved with a `SubgraphDocument` struct that can contain either or both of two representations: a string in GraphQL syntax, or a parsed `apollo_compiler::ExecutableDocument`. The former is used for (de)serialization. When initialized with a string, creating the parsed document requires an explicit call that provides the relevant `&Valid<Schema>`. This can be moved to apollo-federation. Other cases are similar, or a `String` can be used instead.

This PR’s commits each represent a meaningful step in the changes being done, and each should compile and pass tests. Hopefully this makes review easier.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-federation/src/correctness/query_plan_analysis.rs
```diff
@@ -356,12 +356,15 @@ fn interpret_fetch_node(
     conditions: &[Literal],
     fetch: &FetchNode,
 ) -> Result<ResponseShape, String> {
-    let mut result = if let Some(_requires) = &fetch.requires {
+    let mut result = if !fetch.requires.is_empty() {
         // TODO: check requires
-        compute_response_shape_for_entity_fetch_operation(&fetch.operation_document, schema)
-            .map(|rs| rs.add_boolean_conditions(conditions))
+        compute_response_shape_for_entity_fetch_operation(
+            fetch.operation_document.as_parsed().unwrap(),
```
garypen, line 362: Is it better to `unwrap()` here or could we meaningfully propagate an error (`String`) ?

apollo-federation/src/correctness/query_plan_analysis.rs
```diff
@@ -356,12 +356,15 @@ fn interpret_fetch_node(
...
+            schema,
+        )
+        .map(|rs| rs.add_boolean_conditions(conditions))
     } else {
-        compute_response_shape_for_operation(&fetch.operation_document, schema)
+        compute_response_shape_for_operation(fetch.operation_document.as_parsed().unwrap(), schema)
```
garypen, line 367: Same question

apollo-federation/src/query_plan/display.rs
```diff
@@ -364,6 +357,69 @@ fn write_selections(
     state.write("}")
 }
 
+fn write_requires_selections(
+    state: &mut State<'_, '_>,
+    mut selections: &[requires_selection::Selection],
+) -> fmt::Result {
+    if let Some(requires_selection::Selection::Field(field)) = selections.first() {
```
duckki, line 364: Requires selections won't have `_entities`. So, this if statement won't be necessary.

duckki (approved):
Looks great! I have a minor suggestion above.
One more suggestion: we can shorten this line in some `impl From<>` for defer nodes.
```rust
   subselection: sub_selection.as_ref().map(|s| s.to_string()),
```
to 
```rust
   subselection: sub_selection.clone(),
```

apollo-router/src/query_planner/convert.rs
```diff
@@ -195,18 +190,15 @@ impl From<&'_ next::DeferredDeferBlock> for plan::DeferredNode {
...
-                        next::QueryPathElement::InlineFragment(inline) => {
-                            inline.type_condition.as_ref().map(|cond| {
-                                crate::graphql::JsonPathElement::Fragment(cond.to_string())
-                            })
+                    .map(|e| match e {
+                        next::QueryPathElement::Field { response_key } =>
+                        // TODO: type conditioned fetching once it s available in the rust planner
+                        {
+                            crate::graphql::JsonPathElement::Key(response_key.to_string(), None)
+                        }
```
lrlna, line 198: Unrelated to this PR: looks like we are not handling type conditioned here, and probably should be.

lrlna (approved):

apollo-router/src/query_planner/convert.rs
duckki: Filed https://apollographql.atlassian.net/browse/ROUTER-1143 for that.

PR #6862, opened 2025-02-25
Subgraph include errors with allow and deny list

Samjin:
WIP:

Fixes #**issue_number**

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


IvanGoncharov:
> There could several operational challenges for allowlist approach:
> * All subgraphs must propagate errors to begin with, because gateway owner cannot determine which extension keys are already dependent by client(this is not even including data used in client metrics or logging). However, in our case, we already know specific keys that must be redacted globally for security or compliance reasons.
> * it's error prone as it requires subgraphs to work closely with clients(including external clients) to accurately identify dependent keys to avoid breaking clients. The overall process can be time consuming as it needs effort from many teams.

@Samjin Continuing on discussion in #6402 
I still think we should be "secure by default" and provide only allowlist functionality.
Otherwise companies who use router, will plug information leaks only after they discovered.

However, your argument about existing client is totally valid and 100% should be solved.
I just want to get your feedback on alternative solution:
What if we add new metric (histogram of `key inside extensions => number of times used`) that will count how many times each extension was used in errors returned from subgraph service.
That will allow companies to run Router for some period of time (e.g. a week) collect all the keys that are currently used and add this to first version of allowlist.

That way existing clients is not broken but new extensions are not exposed by default.
What do you think about this solution

BrynCooke:
@IvanGoncharov I think that the config *is* secure by default, and this PR just provides more granular behaviour for what we already have.

The reasons for this are:
1. The default behaviour has not changed. We still deny error messages and globally allowing must be explicitly enabled.
2. The existing config allowed users to let error messages globally propagate or on subgraph basis. This PR adds some fine grained control over this.

I see your point that users could shoot themselves in the foot. However, I don't think we can predict the error handling policy for organizatons, and the current config is not expressive enough to capture their needs. It's also worth pointing out that not every router is exposed to the internet, and it may be acceptable for some users to expose errors except certain sensitive keys, especially if they have hundreds of teams developing subgraphs. 

Expecting centralised config to be updated every time a subgraph adds a new key seems unlikely to scale. And I don't think metrics are going to cut it.

How can we mitigate your concerns? A couple of thoughts:
1. We call out explicitly in the docs what the risks are for using a deny list and state that we don't recommend this.
2. We log a warning on router startup that a deny list is in use and that users MUST have an error handling policy for their subgraphs.

IvanGoncharov:
@BrynCooke, I agree with your points 👍 

@Samjin We had an internal discussion, and the consensus is that we go with `deny_extensions_keys.`
To address my concerns, we just need to document that in future releases, Router can add its own extensions to subgraph errors in minor releases (but we would definetly mention it in release notes).
Can you please add such note to the docs?

Samjin:
Im not sure why many telemetry test cases and gileaks are failing? Are they even relate to my change?

abernix:
@Samjin I _don't_ think they're related to your change, I think they're related to external contributor PRs + CircleCI limitation in how we have things setup.  I think @BrynCooke is going to look at this next week, but we may need to clone this PR before it lands.

abernix:
@Samjin Just for transparency, this has fallen a bit behind internally in no fault because of your PR.  We're still looking to move it forward, but some other things have taken priority and @BrynCooke may not be able to get this reviewed this week.  (+ @IvanGoncharov is currently out of office).

BrynCooke:
There are quite a few changed required to get this onto the 2.0 branch. We won't backport this to 1.x.

New WIP PR is at https://github.com/apollographql/router/pull/7164

I think it's close. Just need to get the tests to pass and add a few more tests.

BrynCooke:
Closing this PR as the new WIP PR against the dev branch has significant changes.

abernix:
@BrynCooke Your comment indicates to me that you meant to "close" this, so I'll do that. 😄

PR #6863, opened 2025-02-25
Fix missing content type

Samjin:
Batching intermittently throw error. Debug log says resp is missing content type. In our case, it looks like istio returned raw resp somehow which doesn't contain content-type header. 

Fix is to make sure body is always read and processed, even if the `Content-Type` header is missing when the response is chunked and the Content-Type header is missing.

image

Fixes #**issue_number**

---

**Checklist**

- [x ] Changes are compatible[^1]
- Tests added and passing[^3]
    - [ x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/services/subgraph_service.rs
+++ b/apollo-router/src/services/subgraph_service.rs
@@ -1557,7 +1557,6 @@ async fn do_fetch(
     } else {
         get_graphql_content_type(service_name, &parts)
     };
-    // let content_type = get_graphql_content_type(service_name, &parts);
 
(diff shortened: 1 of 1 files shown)


Samjin:
Converted to draft. need to look at how subgraph is randomly not returning content-type header. If apollo wants to gracefully handle this, the PR maybe still useful?

Samjin:
downstream racing issue.

PR #6881, opened 2025-02-28
docs: fix values for `apollo::supergraph::operation_kind` context key

goto-bus-stop:
The operation kind is actually lowercase (and always has been).


lrlna (approved):
(and always has been)

PR #6884, opened 2025-02-28
[docs] Reduce duplicate config sections and move to top

smyrick:
Docs upgrade. We had duplicate sections explaining what `./router config ...` does and the important sections were at the bottom of the page.

This bumps up and makes the two important sections right to the top, before the long list of YAML config file options

* Configuration awareness in editor
* Upgrading your config


smyrick:
I did intentionally delete the section that was on the top since it really just said the same thing as the sections below

PR #6885, opened 2025-02-28
chore: update federation tests to apply correctness checks

dariuszkuc:
Update federation tests:
* operation normalization/fragment genereation unit tests now verify that original and generated operation are equal
* query planner integration tests now explicitly verify that generated plans are correct


apollo-federation/src/correctness/mod.rs
```diff
@@ -73,6 +73,9 @@ pub fn check_plan(
                 "Failed to compute the response shape from query plan:\n{e}"
             ))
         })?;
+    println!(
+        "check_plan:\nOperation response shape: {op_rs}\nQuery plan response shape: {plan_rs}"
+    );
```
lrlna, line 78: This is not needed. When you set `RUST_LOG=debug` the tracing::debug invocation below is sent to stdout.

For example:
```
$ RUST_LOG=debug cargo run -p apollo-federation-cli --release -- plan /path/to/operation /path/to/schema

DEBUG check_plan:
Operation response shape: {
```

lrlna (commented):
I like this! I think this makes for a nice small step towards catching inconsistencies early! And given the CI, it seems like it might have caught some already.

apollo-federation/src/operation/tests/mod.rs
```diff
@@ -21,30 +21,59 @@ use crate::schema::position::ObjectTypeDefinitionPosition;
...
+}
+
+macro_rules! assert_equal_ops {
+    ($schema: expr, $first: expr, $second: expr) => {
+        let original_document: Valid<ExecutableDocument> =
+            $first.try_into().expect("valid document");
+        let minified_document: Valid<ExecutableDocument> =
+            $second.try_into().expect("valid document");
+        compare_operations($schema, &original_document, &minified_document)
+            .expect("documents are the same");
```
duckki, line 50: Similarly to the above comment, `expect("The first document includes the second")`.

apollo-federation/src/operation/tests/mod.rs
```diff
@@ -2031,6 +1941,7 @@ fn reuse_fragments_with_non_intersecting_types() {
         }
     "#;
 
+    // dropping unnecessary __typename selection from fragment
```
duckki, line 1944: I think the correctness checker needs to understand when `__typename` should be dropped. I'll follow-up on this.

duckki (commented):
It looks great! Thanks for wiring the correctness checker!
Please exclude those failing tests from the check. We might need a new macro like `assert_plan_without_check` -- something that can be searched.
I'll follow-up with the checker issues soon after.

duckki:
> I like this! I think this makes for a nice small step towards catching inconsistencies early! And given the CI, it seems like it might have caught some already.

The test failures are actually bugs in the correctness checker. I'm working to fix them.

apollo-federation/src/correctness/mod.rs
dariuszkuc: yep it's just a leftover from my local debugging stuff (pushed this PR to show the failing correctness checker cases)

duckki (approved):
Looks good. Thanks, again!

apollo-federation/Cargo.toml
```diff
@@ -55,6 +55,8 @@ tempfile.workspace = true
 pretty_assertions = "1.4.0"
 rstest = "0.22.0"
 dhat = "0.3.3"
+# workaround for https://github.com/rust-lang/cargo/issues/2911
+apollo-federation = { path = ".", features = ["correctness"] }
```
goto-bus-stop, line 59: huh, cool!

PR #6888, opened 2025-03-03
fix(entity_caching): do not include other fields from representation variable than the entity keys

### LINKED ISSUE #6673: Entity caching uses stringified representations within entity cache key

bnjjj:
Separate entity keys and representation variable value in the cache key to avoid issues with `@requires` for example.

close #6673 

 

[ROUTER-1007]: https://apollographql.atlassian.net/browse/ROUTER-1007?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apollo-router/src/plugins/cache/entity.rs
```diff
@@ -1344,22 +1410,55 @@ fn extract_cache_keys(
...
+                representation.insert(key, val);
+            });
         res.push(key);
     }
     Ok(res)
 }
 
-pub(crate) fn hash_entity_key(representation: &Value) -> String {
+/// Use it specifically for keys field set because we don't check for fragment spread
+fn get_root_field_names(selection_set: &SelectionSet) -> Vec<Name> {
```
SimonSapin, line 1432: With apollo-compiler 1.27.0 released today `selection_set.root_fields()` returns an iterator that can replace this function

apollo-router/src/plugins/cache/tests.rs
```diff
@@ -152,6 +152,9 @@ impl Mocks for MockStore {
 
 #[tokio::test]
 async fn insert() {
+    let mut parser = apollo_compiler::parser::Parser::new();
+    let schema = parser.parse_schema(SCHEMA, "test.graphql").unwrap();
+    let valid_schema = Arc::new(schema.validate().unwrap());
```
SimonSapin, line 157: If you’re not changing `Parser` parameters you can use `Schema::parse_and_validate()`

Same a few more times below

SimonSapin (approved):
Looks good so far, but since it’s draft I assume you want to make more changes?

bnjjj:
> Looks good so far, but since it’s draft I assume you want to make more changes?
@SimonSapin yes tests

apollo-router/src/plugins/cache/entity.rs
```diff
@@ -500,10 +514,33 @@ impl EntityCache {
...
+    let mut subgraph_enums = HashMap::new();
+    if let Some(graph_enum) = supergraph_schema.get_enum("join__Graph") {
+        subgraph_enums.extend(graph_enum.values.iter().filter_map(
+            |(enum_name, enum_value_def)| {
+                let subgraph_name = enum_value_def
+                    .directives
+                    .get("join__graph")
+                    .and_then(|directive| directive.specified_argument_by_name("name"))
+                    .and_then(|arg| arg.as_str())?
+                    .to_string();
```
SimonSapin, line 534: Nit (feel free to ignore): I think this might work:
```suggestion
                    .get("join__graph")?
                    .specified_argument_by_name("name")?
                    .as_str()?
                    .to_string();
```

SimonSapin (approved):

abernix:
@bnjjj The test failure here looks potentially related, so I'm going to disable auto-merge while you take a look.

bnjjj:
Good catch @abernix . I fixed it :) thank you :)

PR #6896, opened 2025-03-03
make request decompression into a plugin

Velfi:
*Description here*

This PR takes the existing `RequestDecompressionLayer` and turns it into a plugin.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


BrynCooke (changes requested):
As new config has been added these need to be added to the docs.

goto-bus-stop:
Turning it into a plugin also requires adding it to our Big List Of Plugins: https://github.com/apollographql/router/blob/bdeb1b64b839c714099e580a7e45543accc0cf09/apollo-router/src/router_factory.rs#L663-L721

That's how it gets instantiated and (eventually) added to the pipeline.

apollo-router/src/plugins/request_decompression.rs
```diff
@@ -0,0 +1,90 @@
...
+                    .deflate(self.deflate),
+            )
+            .map_request(router::Request::from)
+            .map_response(http::Response::from)
+            .service(service)
+            .boxed()
+    }
+}
+
+register_private_plugin!(
```
BrynCooke, line 59: This needs to also be added to the mandatory plugin list. Search for `add_mandatory_apollo_plugin`. It should go somewhere after traffic shaping as limits should be applied before decompression.

apollo-router/src/plugins/request_decompression.rs
```diff
@@ -0,0 +1,90 @@
+use crate::plugin::{PluginInit, PluginPrivate};
+use crate::services::router;
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+use tower::{ServiceBuilder, ServiceExt};
+
+#[derive(Debug, Default, Deserialize, JsonSchema, Serialize)]
+struct RequestDecompressionConfig {
```
BrynCooke, line 8: Prefer implementing Default rather than using option. There's history to this where option causes unexpected behaviour in serde for complex objects. There's details in the dev docs (in repo) for config design and implementation.

apollo-router/src/plugins/request_decompression.rs
```diff
@@ -0,0 +1,90 @@
...
+    RequestDecompressionPlugin
+);
+
+#[cfg(test)]
+mod test {
+    use crate::plugin::{PluginInit, PluginPrivate};
+    use super::{RequestDecompressionConfig, RequestDecompressionPlugin};
+
+    #[tokio::test]
+    async fn test_config() {
```
BrynCooke, line 71: We should add tests for what happens when a request is invalid. Do we get the right response?

apollo-router/src/plugins/request_decompression.rs
```diff
@@ -0,0 +1,90 @@
...
+        })
+    }
+
+    fn router_service(&self, service: router::BoxService) -> router::BoxService {
+        // RequestDecompressionLayer works on `http` types so we have to perform the conversions to
+        // make it compatible with `BoxService`.
+        ServiceBuilder::new()
+            .map_request(http::Request::from)
+            .map_response(router::Response::from)
+            .layer(
```
BrynCooke, line 46: We're probably going to need an `map_err` here. All responses should look graphql shaped.

BrynCooke (changes requested):

apollo-router/src/plugins/request_decompression.rs
Velfi: Let's talk soon about how to do this the best way. I'm not familiar enough with this yet to do it on my own.

apollo-router/src/plugins/request_decompression.rs
Velfi: I've added tests: three for success and three for failure. The (deflate, success) case is currently failing for reasons I don't understand so I'll have to figure that out before we can consider this done.

apollo-router/src/plugins/request_decompression.rs
Velfi: done

Velfi:
I'm going to close this PR since a different approach is required to solve the issue.

PR #6900, opened 2025-03-03
Fix nested input type reporting bug

meryl-c:
When there are input types in a nested query, they are not getting reported. This PR fixes the bug by processing a field's selection set.
This PR also creates an integration test that makes a query with an enum subquery in order to replicate the scenario in [PULSR-1429](https://apollographql.atlassian.net/browse/PULSR-1429).

Fixes #**PULSR-1429**

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[PULSR-1429]: https://apollographql.atlassian.net/browse/PULSR-1429?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BrynCooke (commented):
The core of this change looks good to me. 

I'm not about performance implications, if the logic is required for correctness then this may trump performance. However, I do know that recursively visiting documents is expensive. And this did come up in demand control CC @tninesling.

You should bench the difference for a large query against a large schema and query.

apollo-router/tests/fixtures/supergraph.graphql
```diff
@@ -1,66 +1,46 @@
 schema
   @link(url: "https://specs.apollo.dev/link/v1.0")
-  @link(url: "https://specs.apollo.dev/inaccessible/v0.2", for: SECURITY)
-  @link(url: "https://specs.apollo.dev/join/v0.3", for: EXECUTION) {
```
BrynCooke, line 4: You probably don't want to modify this fixture. It will break a bunch of other tests. Probably create a new fixture.

BrynCooke (dismissed):

apollo-router/tests/fixtures/supergraph.graphql
meryl-c: Got it, I'll make a new one. Thanks Bryn!

apollo-router/src/apollo_studio_interop/snapshots/apollo_router__apollo_studio_interop__tests__extended_references_nested_query.snap.new
```diff
@@ -0,0 +1,58 @@
+---
```
bonnici, line 1: You'll need to review the snapshots again to accept this. We should never commit a `.snap.new` file, as those are the "unaccepted" version of the snapshot.

apollo-router/src/services/layers/persisted_queries/snapshots/apollo_router__services__layers__persisted_queries__tests__pq_layer_freeform_graphql_with_safelist_log_unknown_true@logs.snap.new
```diff
@@ -0,0 +1,7 @@
+---
```
bonnici, line 1: Ditto for this file - but I'm not sure why this one is new as it looks unrelated to the code change.

bonnici (changes requested):

bonnici:
> The core of this change looks good to me.
> 
> I'm not about performance implications, if the logic is required for correctness then this may trump performance. However, I do know that recursively visiting documents is expensive. And this did come up in demand control CC @tninesling.
> 
> You should bench the difference for a large query against a large schema and query. Alternatively is this logic sampled in ay way?

The logic isn't sampled and it is required for correctness. This impacts operation checks related to unused enum values, so we need to know when an enum is used no matter where it's used in the query.

apollo-router/src/apollo_studio_interop/snapshots/apollo_router__apollo_studio_interop__tests__extended_references_nested_query.snap.new
meryl-c: fixing-- cargo wasn't working for me so I got blocked

apollo-router/src/services/layers/persisted_queries/snapshots/apollo_router__services__layers__persisted_queries__tests__pq_layer_freeform_graphql_with_safelist_log_unknown_true@logs.snap.new
meryl-c: I was confused about that. Tim said he's seen the same thing locally

bonnici:
> You should bench the difference for a large query against a large schema and query.

@BrynCooke can we use router-scale for this, or will we need to use the corpus?

apollo-router/tests/fixtures/supergraph.graphql
meryl-c: ended up deleting this test

meryl-c:
I tried to do router-scale perf testing but it just hung for me. Since we already do some recursion, I don't think this should affect performance.

bonnici (approved):

PR #6904, opened 2025-03-03
Ensure `build_query_plan()` cancels when the request cancels (backport #6840)

SimonSapin:
When a request is cancelled, resources (compute time/memory) allocated to that request should be released. However, query planning is an exception, and currently keeps running after the request is cancelled. This PR adds cooperative cancellation to `build_query_plan()`, to ensure that when the request is cancelled, query planning errors shortly after.<hr>This is an automatic backport of pull request #6840 done by [Mergify](https://mergify.com).


lrlna (approved):

goto-bus-stop (changes requested):
I think we should manually backport the entire feature + tests once it's ready, as they're spread out across various PRs.

duckki:
I created a manual backport and resolved conflicts: https://github.com/apollographql/router/pull/7006

goto-bus-stop:
Will close this and we can consider backporting it when the feature is complete on 2.x

PR #6905, opened 2025-03-04
Count value completion errors in GraphQL error metrics

timbotnik:
Several customers, including ourselves, have run into issues with the silent failures associated with value completion.  Ultimately we hope that these will one day be elevated to proper GraphQL errors, which would make the code path introduced here obsolete.  That is a bigger issue that has had some history [1](https://github.com/apollographql/router/pull/5787), [2](https://github.com/apollographql/router/pull/6314).  As a stop-gap to fixing the root issue (which would likely be a breaking change), we'll introduce a non-breaking change to at least capture some telemetry when this happens.

We will start tracking this as part of our new error metrics using the code `RESPONSE_VALIDATION_FAILED`, and we'll now include a `severity` attribute as well that designates this particular error as a warning, since it's technically _not_ a GraphQL error in the Router right now.  

Note (and request for feedback) - the implementation currently will also will result in counting these value completion errors as part of the existing user-facing instrument `apollo.router.graphql_error` (also using code `RESPONSE_VALIDATION_FAILED`).  I think there are benefits to allowing users to observe these, however it could be misleading in that they are not hard GraphQL errors from the client perspective.  

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[OE-602]: https://apollographql.atlassian.net/browse/OE-602?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bonnici (approved):
Looks good to me. Can you create a story to ingest the error severity into trace_path_errors_refs if we want to ingest this?

goto-bus-stop (approved):
I think this is a great idea :)

bnjjj (approved):

apollo-router/src/services/router/service.rs
```diff
@@ -877,15 +887,21 @@ impl RouterService {
                 .and_then(|s| s.as_str())
                 .unwrap_or_default()
                 .to_string();
+            let severity = error.extensions.get("severity").and_then(|s| s.as_str());
             let path = match &error.path {
                 None => "".into(),
                 Some(path) => path.to_string(),
             };
+            // TBD(tim): do we want to also transmit warnings to the user-facing "apollo.router.graphql_error" instrument?
+            // Can see arguments for both sides...
```
goto-bus-stop, line 896: Right now the PR *does* count it, right?

I think counting it in graphql_error is good, but we should call it out in the changelog so users can understand why their errors are increasing.

timbotnik: Sounds good to me.  I'll remove the TBD.

PR #6912, opened 2025-03-04
Install Rust-based tools on CI from pre-compiled binaries

SimonSapin:
Installing from source can take up to 8 minutes:
https://app.circleci.com/pipelines/github/apollographql/router/31305/workflows/792ab4bc-cfcd-4120-bfbb-d86405cf5964/jobs/230568/steps

Installing binaries takes 5 seconds:
https://app.circleci.com/pipelines/github/apollographql/router/31322/workflows/dcca4c65-e4bb-46ad-a05a-c0b9f6c77e11/jobs/230688/steps

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -369,20 +369,20 @@ commands:
       - run:
           name: Install cargo deny, about, edit
           command: |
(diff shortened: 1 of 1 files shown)


lrlna (approved):
was there a reason we _didn't_ do this before? 8m vs 5s is a huge difference!

PR #6930, opened 2025-03-05
Update JWKS handling

### LINKED ISSUE #6180: Support JWT authorization without failing request upon validation failure
Today, if a JWT is present but validation of the JWT fails, the router rejects the request.

We require when a JWT is present, that it be validated, but if that fails, allow the request to continue, but obviously the context `apollo_authentication::JWT::claims` should not be populated.

A router configuration setting, allowing the the processing to continue even if JWT validation fails.

Perhaps a context value `apollo_authentication::JWT::status` to be provided to allow rhai/coprocessor to differentiate when no JWT was provided versus when a JWT could not be validated.

Having to write our own JWT processing code and not be able to leverage Router's built-in support

Velfi:
This PR updates JWT-handling in the `AuthenticationPlugin`;

- JWT-related code has been moved around a bit for organizational purposes
- Users may now set a new a new config option `config.authentication.router.jwt.on_error`. When set to the default `Error`, JWT-related errors will be returned to users (the current behavior.) When set to `Continue`, JWT errors will instead be ignored. When a JWT error is encountered, JWT claims will not be set in the request context.
- When JWTs are processed, whether processing fails or succeeds, request context will contain a new variable `apollo::authentication::jwt_status` which notes the result of processing.
- I applied a few lints that RustRover highlighted.

Fixes #6180 

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


dupham-indeed:
Can we make sure in the `continue` option to send the `jwt_claims` context over to the coprocessor no matter whether the token has expired or not? We might need that for observability purposes

theJC:
> Can we make sure in the `continue` option to either send an error code value or send the `jwt_claims` context over to the coprocessor no matter whether the token has expired or not? We might need that for observability purposes

Historically when we had a case where we had to dig into incorrectly minted JWT's, having access to the claims (even when invalid or expired in some way) to be able to provide details via metrics or logs was instrumental into pinpointing what was wrong.

BrynCooke:
@dupham-indeed You'll be able to configure the coprocessor to send the new context key over.

apollo-router/src/plugins/authentication/mod.rs
```diff
@@ -150,6 +98,13 @@ struct JWTConf {
     /// Alternative sources to extract the JWT
     #[serde(default)]
     sources: Vec<Source>,
+    /// Control the behavior when an error occurs during the authentication process.
+    ///
+    /// Defaults to `Error`. When set to `Continue`, requests that fail JWT authentication will
+    /// continue to be processed by the router, but without the JWT claims in the context. When set
+    /// to `Error`, requests that fail JWT authentication will be rejected with a HTTP 403 error.
+    #[serde(default = "OnError::default")]
```
BrynCooke, line 106: If JWTConf and OnError implementsDefault you don't need this
```suggestion
```
Same for the other fields

apollo-router/src/plugins/authentication/mod.rs
```diff
@@ -133,7 +74,14 @@ struct AuthenticationPlugin {
...
 
-#[derive(Clone, Debug, Deserialize, JsonSchema, serde_derive_default::Default)]
+#[derive(Clone, Debug, Deserialize, JsonSchema, Default, PartialEq)]
+enum OnError {
+    Continue,
+    #[default]
+    Error,
+}
+
+#[derive(Clone, Debug, Deserialize, JsonSchema, Default)]
```
BrynCooke, line 84: We should implement Default for this manually

BrynCooke (changes requested):
Minor config stuff.
Docs need to be added.
Shame to not use the new test harness, but I understand why. I'll refactor the tests separately.

apollo-router/src/plugins/authentication/jwks.rs
```diff
@@ -256,3 +273,419 @@ impl Iterator for Iter<'_> {
...
+    headers: &'b HeaderMap,
+) -> Option<Result<&'b str, AuthenticationError>> {
+    match source {
+        Source::Header { name, value_prefix } => {
+            // The http_request is stored in a `Router::Request` context.
+            // We are going to check the headers for the presence of the configured header
+            let jwt_value_result = match headers.get(name) {
+                Some(value) => value.to_str(),
+                None => return None,
+            };
```
bnjjj, line 466: NIT: 
```suggestion
            let jwt_value_result = headers.get(name)?.to_str();
```

apollo-router/src/plugins/authentication/jwks.rs
```diff
@@ -256,3 +273,419 @@ impl Iterator for Iter<'_> {
...
+                candidates
+                    .into_iter()
+                    .map(|(_score, candidate)| candidate)
+                    .collect(),
+            )
+        }
+    }
+}
+
+pub(super) fn extract_jwt<'a, 'b: 'a>(
```
bnjjj, line 454: I think it would be useful to document what's the meaning of the return type here. Like what does that mean when it's `None` or `Some(Err)`  and so one

apollo-router/src/plugins/authentication/jwks.rs
```diff
@@ -256,3 +273,419 @@ impl Iterator for Iter<'_> {
...
+        validation.validate_aud = false;
+
+        match decode::<serde_json::Value>(jwt, &decoding_key, &validation) {
+            Ok(v) => return Ok((issuer, v)),
+            Err(e) => {
+                tracing::trace!("JWT decoding failed with error `{e}`");
+                error = Some((
+                    AuthenticationError::CannotDecodeJWT(e),
+                    StatusCode::UNAUTHORIZED,
+                ));
```
bnjjj, line 601: `error` could potentially be overridden by another one here. Is it expected ? Should we have a Vec of error instead to keep all of them ?

apollo-router/src/plugins/authentication/jwks.rs
```diff
@@ -256,3 +273,419 @@ impl Iterator for Iter<'_> {
...
+        .get(APOLLO_AUTHENTICATION_JWT_CLAIMS)
+        .map_err(|err| tracing::error!("could not read JWT claims: {err}"))
+        .ok()
+        .flatten();
+    let ts_opt = claims.as_ref().and_then(|x: &serde_json::Value| {
+        if !x.is_object() {
+            tracing::error!("JWT claims should be an object");
+            return None;
+        }
+        let claims = x.as_object().expect("claims should be an object");
```
bnjjj, line 632: As we try as much as possible to get rid of `expect` and `unwrap` in our code, maybe that could be worth it to use pattern matching here instead of an expect ?

apollo-router/src/plugins/authentication/jwks.rs
```diff
@@ -256,3 +273,419 @@ impl Iterator for Iter<'_> {
...
+                .expect("we should not run before EPOCH")
+                .as_secs() as i64;
+            if now < ts {
+                Duration::from_secs((ts - now) as u64)
+            } else {
+                Duration::ZERO
+            }
+        }
+        None => Duration::MAX,
+    }
```
bnjjj, line 653: Maybe it could also be interesting to comment this behavior in function's doc. To document what are our assumptions here. Mainly explain that if there is no `exp` in the claim or if it's not a number then we set it to MAX for example

BrynCooke (changes requested):
Docs need updating with new functionality here: 
https://github.com/apollographql/router/blob/dev/docs/source/routing/security/jwt.mdx

and possibly here: 
https://github.com/apollographql/router/blob/dev/docs/source/routing/security/router-authentication.mdx

apollo-router/src/plugins/authentication/jwks.rs
Velfi: I moved this code but did not write it. As such, I'm not sure what the best thing to do is.

apollo-router/src/plugins/authentication/jwks.rs
Velfi: I'll rewrite this as a `match`

BrynCooke (approved):

BrynCooke:
@Velfi I've approved, but please update the docs to also include the shape of what will end up in context for the user, otherwise they won't know how to interact with it in Rhai.

apollo-router/src/plugins/authentication/jwks.rs
bnjjj: Oh ok, sorry. Unfortunately it hasn't be done before :p now it's like dark magic

apollo-router/src/plugins/authentication/jwks.rs
bnjjj: Oh ok, sorry. Unfortunately it hasn't be done before :p now it's like dark magic

bnjjj (approved):

Velfi:
> @Velfi I've approved, but please update the docs to also include the shape of what will end up in context for the user, otherwise they won't know how to interact with it in Rhai.

Ah yes, that would be important 😅. I'll add that before merging.

PR #6936, opened 2025-03-06
Update team names in CODEOWNERS

abernix:
Updating these to new names that we're using internally.  I will backport this to a couple places to ensure they're not used still in actively enforced places.

PR #6937, opened 2025-03-06
Remove now unused CODEOWNERS line for apollo-router-scaffold

abernix:
Per comment from @goto-bus-stop  in https://github.com/apollographql/router/pull/6936#discussion_r1983194113


abernix:
Stacked PR will re-target to `dev` after #6936 lands.  (Feel free to mark as ready for review once that happens.)

PR #6943, opened 2025-03-06
Pattern for iterating over directive applications

tninesling:
Implements a pattern for iterating over all applications of a particular directive. In JS, this is called as `directive.applications()` on some directive definition. In Rust, the directive definition itself doesn't have a pointer back to the `FederationSchema`, so the schema will return the iterators. There is a different iterator for each directive because the contained fields don't have useful common types, and this tries to provide a simply, strongly-typed directive representation for consumption.

---

**Checklist**

- Tests added and passing[^3]
    - [X] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-federation/src/schema/mod.rs
+++ b/apollo-federation/src/schema/mod.rs
@@ -396,6 +396,7 @@ impl FederationSchema {
                     schema_directive: provides_directive_application,
                     sibling_directives: directives,
                     target: field_definition_position,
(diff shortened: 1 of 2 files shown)


lrlna (commented):
I feel like there is room for `apollo-compiler`'s `Directive` to have a `.applications()` method, which may perhaps make this easier 🤔

PR #6949, opened 2025-03-07
Propagate request cancellation from the router service to the query planner

goto-bus-stop:
Follow-up to #6840 with a missing piece, and some tests.

This uses `scopeguard` to abort the QueryPlannerService future when
the CachingQueryPlanner future aborts.

By continuing to use a tokio task, and a separate abort handle, we can
adjust the abort conditions later (such as introducing a separate QP
timeout, independent of request cancellation).

**Todos**:
- [ ] An integration-type test (at least testing the entire pipeline starting at the RouterService)

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/query_planner/caching_query_planner.rs
+++ b/apollo-router/src/query_planner/caching_query_planner.rs
@@ -515,11 +515,7 @@ where
                 .plan_options(caching_key.plan_options)
                 .build();
 
-            // some clients might timeout and cancel the request before query planning is finished,
-            // so we execute it in a task that can continue even after the request was canceled and
-            // the join handle was dropped. That way, the next similar query will use the cache instead
-            // of restarting the query planner until another timeout
-            tokio::task::spawn(
+            let qp_task = tokio::spawn(
                 async move {
                     let service = match self.delegate.ready().await {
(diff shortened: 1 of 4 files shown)


apollo-router/src/query_planner/caching_query_planner.rs
```diff
@@ -1041,4 +1047,98 @@ mod tests {
...
+            .with_lock(|lock| lock.insert::<ParsedDocument>(doc));
+
+        let fut = planner.call(query_planner::CachingRequest::new(
+            "query Me { me { name { first } } }".to_string(),
+            Some("".into()),
+            context.clone(),
+        ));
+
+        let handle = tokio::task::spawn(fut);
+        tokio::time::sleep(Duration::from_millis(25)).await;
```
goto-bus-stop, line 1127: unfortunately it's a time-based test the way i've written this; 25ms might be a bit tight here.

i could probably use a few oneshot channels to make it more robust

PR #6950, opened 2025-03-07
Replace Rhai specific hot-reload functionality with general hot-reload

garypen:
The Rhai hot-reload functionality in 1.x only worked because 1.x created the entire service stack for each request.
    
In 2.0, we no longer do that and so the Rhai hot-reload functionality does not work.
    
The solution is to remove the rhai specific hot-reload code and to add rhai source code into the set of things watched by the router which will trigger a hot reload event.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [x] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/uplink/license_stream.rs
```diff
@@ -533,7 +533,7 @@ mod test {
                 }
                 Event::UpdateLicense(_) => SimpleEvent::UpdateLicense,
                 Event::NoMoreLicense => SimpleEvent::NoMoreLicense,
-                Event::Reload => SimpleEvent::ForcedHotReload,
+                Event::Reload | Event::RhaiReload => SimpleEvent::ForcedHotReload,
```
garypen, line 536: Reviewers: I'm not 100% certain that this is the best thing to do here, but it didn't seem worth creating a new `SimpleEvent` for the new `RhaiReload` event. Happy to change this though if required.

BrynCooke (approved):

apollo-router/src/uplink/license_stream.rs
BrynCooke: I like it

PR #6956, opened 2025-03-07
Divide individual `@connect` from whole-schema validations

dylan-apollo:
Two new modules `connect` and `schema` replace comprised collectively of code that used to be in `extended_type`, `entity`, `selection`, and `mod.rs`. Validations which mostly work on a single connector at a time are in `connect`, whereas validations which operate on the entire schema as a whole (e.g., missing files, key resolvability) are in `schema`.

Depends on https://github.com/apollographql/router/pull/6940


apollo-federation/src/sources/connect/validation/connect.rs
```diff
@@ -19,39 +21,45 @@ use super::coordinates::FieldCoordinate;
...
 use crate::sources::connect::spec::schema::CONNECT_SOURCE_ARGUMENT_NAME;
 use crate::sources::connect::spec::schema::HTTP_ARGUMENT_NAME;
 use crate::sources::connect::validation::graphql::SchemaInfo;
 
-pub(super) fn validate_extended_type(
-    extended_type: &ExtendedType,
+mod entity;
+mod selection;
+
+pub(super) fn fields_seen_by_all_connects(
```
lennyburdette, line 35: the names of these functions threw me — these functions do the validation, and then return either messages or the lists of seen fields for later validation? that feels off to me. at the very least, the fn name should talk about validation, right? 

won't we have warning messages that we'd want to emit _with_ seen fields? using Result for this feels weird to me, baking the logic of what gets validated into the signatures of all these functions 

i've also wondered if we should address the noisy seen fields problem by emitting only a single Message with a list of source locations.

lennyburdette (commented):
definitely like the connect/schema split, but i'm not quite sure about the validation/seen fields organization. we can chat about it tomorrow.

and it looks like i should wait for this to land since it overlaps with my PRs (spec in SchemaInfo specifically)

lennyburdette (approved):

PR #6959, opened 2025-03-10
chore: refactor federation to avoid referencing fragment spreads

dariuszkuc:
In order to simplify the query planning process, we first "normalize" the `apollo-rs` Operation into a form that deduplicates all selections and expands all fragment spreads. As a result, during query planning we only have to handle field selections and inline fragments. This PR refactors federation code to remove all the fragment spread references. Since operations are normalized before query planning, that logic should never be executed anyway. As a side effect fragment generation logic now generates `apollo-rs` Operation instead.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-federation/src/operation/contains.rs
+++ b/apollo-federation/src/operation/contains.rs
@@ -1,7 +1,6 @@
 use apollo_compiler::executable;
 
 use super::FieldSelection;
-use super::FragmentSpreadSelection;
 use super::HasSelectionKey;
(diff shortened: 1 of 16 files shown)


dariuszkuc:
closed one from the fork in favor of the direct branch (as I think some CI might not work properly against forks)

PR #6960, opened 2025-03-10
chore: refactor federation to avoid referencing fragment spreads

dariuszkuc:
In order to simplify the query planning process, we first "normalize" the `apollo-rs` Operation into a form that deduplicates all selections and expands all fragment spreads. As a result, during query planning we only have to handle field selections and inline fragments. This PR refactors federation code to remove all the fragment spread references. Since operations are normalized before query planning, that logic should never be executed anyway. As a side effect fragment generation logic now generates `apollo-rs` Operation instead.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-federation/src/operation/contains.rs
+++ b/apollo-federation/src/operation/contains.rs
@@ -1,7 +1,6 @@
 use apollo_compiler::executable;
 
 use super::FieldSelection;
-use super::FragmentSpreadSelection;
 use super::HasSelectionKey;
 use super::InlineFragmentSelection;
 use super::Selection;
@@ -61,14 +60,9 @@ impl Selection {
             (Selection::Field(self_field), Selection::Field(other_field)) => {
                 self_field.containment(other_field, options)
             }
-            (
-                Selection::InlineFragment(self_fragment),
-                Selection::InlineFragment(_) | Selection::FragmentSpread(_),
-            ) => self_fragment.containment(other, options),
-            (
-                Selection::FragmentSpread(self_fragment),
-                Selection::InlineFragment(_) | Selection::FragmentSpread(_),
-            ) => self_fragment.containment(other, options),
+            (Selection::InlineFragment(self_fragment), Selection::InlineFragment(_)) => {
+                self_fragment.containment(other, options)
+            }
             _ => Containment::NotContained,
         }
     }
@@ -105,24 +99,6 @@ impl FieldSelection {
                 );
                 Containment::NotContained
             }
-        }
-    }
-}
-
-impl FragmentSpreadSelection {
-    pub(crate) fn containment(
-        &self,
-        other: &Selection,
-        options: ContainmentOptions,
-    ) -> Containment {
-        match other {
-            // Using keys here means that @defer fragments never compare equal.
-            // This is a bit odd but it is consistent: the selection set data structure would not
-            // even try to compare two @defer fragments, because their keys are different.
-            Selection::FragmentSpread(other) if self.spread.key() == other.spread.key() => self
-                .selection_set
-                .containment(&other.selection_set, options),
-            _ => Containment::NotContained,
         }
     }
 }
@@ -208,7 +184,7 @@ impl SelectionSet {
 mod tests {
     use super::Containment;
     use super::ContainmentOptions;
-    use crate::operation::Operation;
+    use crate::operation::NormalizedOperation;
     use crate::schema::ValidFederationSchema;
 
     fn containment_custom(left: &str, right: &str, ignore_missing_typename: bool) -> Containment {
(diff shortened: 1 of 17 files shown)


apollo-federation/src/operation/mod.rs
```diff
@@ -190,56 +190,45 @@ impl ArgumentList {
 /// - Stores the schema that the operation is queried against.
 /// - Swaps `operation_type` with `root_kind` (using the analogous apollo-federation type).
 /// - Encloses collection types in `Arc`s to facilitate cheaper cloning.
-/// - Stores the fragments used by this operation (the executable document the operation was taken
-///   from may contain other fragments that are not used by this operation).
+/// - Expands all named fragments into inline fragments.
+/// - Deduplicates all selections within its selection sets.
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub struct Operation {
+pub struct NormalizedOperation {
```
duckki, line 196: Do we need to add `Normalized`? `Selection`, `SelectionSet`,etc. are all normalized in this module.

apollo-federation/src/operation/mod.rs
```diff
@@ -3728,25 +3100,123 @@ impl Display for InlineFragment {
...
             OperationElement::InlineFragment(inline_fragment) => inline_fragment.fmt(f),
-            OperationElement::FragmentSpread(fragment_spread) => fragment_spread.fmt(f),
         }
     }
 }
 
+/// Holds normalized selection sets of provided fragments.
+#[derive(Default)]
+pub(crate) struct FragmentSpreadCache {
+    fragment_selection_sets: Arc<HashMap<Name, SelectionSet>>,
```
duckki, line 3115: It is worth noting that we are dropping directives on the declaration of named fragments.
That wasn't used before. But, we may want to keep that comment somewhere.
```rust
        // PORT_NOTE: The JS codebase combined the fragment spread's directives with the fragment
        // definition's directives. This was invalid GraphQL as those directives may not be applicable
        // on different locations. While we now keep track of those references, they are currently ignored.
        pub(crate) fragment_directives: DirectiveList,
```

apollo-federation/tests/query_plan/build_query_plan_tests/fragment_autogeneration.rs
```diff
@@ -450,8 +450,8 @@ fn works_with_key_chains() {
         Fetch(service: "Subgraph1") {
           {
             t {
-              __typename
               id1
```
duckki, line 453: I'd like to see `__typename` at the beginning, since it's easier to spot when debugging.

duckki (commented):
Looking pretty good. It feelds so good when so much code is (safely) removed :)
I have a few questions. I'll take one more look tomorrow.

apollo-federation/src/operation/mod.rs
dariuszkuc: I wanted to have a distinction between `Operation` which includes field, inline fragment + spread selection and this custom one that doesn't expose spreads. I can revert the name change though it if it clearer.

apollo-federation/src/operation/contains.rs
```diff
@@ -247,8 +223,8 @@ mod tests {
         )
         .unwrap();
         let schema = ValidFederationSchema::new(schema).unwrap();
-        let left = Operation::parse(schema.clone(), left, "left.graphql", None).unwrap();
-        let right = Operation::parse(schema.clone(), right, "right.graphql", None).unwrap();
+        let left = NormalizedOperation::parse(schema.clone(), left, "left.graphql").unwrap();
```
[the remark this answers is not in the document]
dariuszkuc: its just a unit test so yeah we expect those unwraps to always succeed. I can change it to expect (but end result is still the same)

apollo-federation/tests/query_plan/build_query_plan_tests/fragment_autogeneration.rs
dariuszkuc: yeah i'll look into this as we shouldn't be modifying the plan

dariuszkuc:
The failing tests are from the reordering of the `__typename` field in the final selection set. It appears that our `apollo-federation` `SelectionSet` performs some optimization to ensure that `__typename` field is serialized first. Since we are now using `apollo-rs` types directly in `generate_fragments` this optimization is no longer hit, i.e.

```
// serialized selection set
{ __typename id1 }

// iterating over selections in the selection set
id1
__typename
```

apollo-federation/tests/query_plan/build_query_plan_tests/fragment_autogeneration.rs
dariuszkuc: fixed in https://github.com/apollographql/router/pull/6960/commits/00458168bbb07fd1a396059acd18288f8a51476e

apollo-federation/src/query_graph/path_tree.rs
```diff
@@ -718,10 +717,12 @@ mod tests {
 
         let normalized_operation = normalize_operation(
             operation,
-            Default::default(),
+            &Default::default(),
             &schema,
             &Default::default(),
-            &never_cancel,
+            &||
```
duckki, line 723: Can we keep `never_cancel` here?

duckki (approved):
Looks great! Just one nit in a test.

apollo-federation/src/query_graph/path_tree.rs
dariuszkuc: https://github.com/apollographql/router/pull/6960/commits/8643ec5b03071127634cd238ed797d63856ad595

PR #6961, opened 2025-03-10
Improve our Renovate config

goto-bus-stop:
**Remove renovate grouping of < 1.0 packages**  
A pre 1.0 minor package update can be a breaking change. We have many
pre 1.0 dependencies that are totally independent, so the grouped up PR
can almost never be merged, and actually functions more as a garbage bin
for us to forget about these dependencies altogether.

This change will cause more renovate PRs to be opened, and also more PRs
that we can't merge at all. But we'll also be much more able to stay up
to date on the dependencies that we *can* upgrade.

**Remove router-bridge update config**  
We no longer have this dependency 🎉

### DIFF
--- a/renovate.json5
+++ b/renovate.json5
@@ -99,15 +99,6 @@
       "groupSlug": "rust-apollo-rs-updates",
       "matchManagers": ["cargo"],
       "automerge": false
-    },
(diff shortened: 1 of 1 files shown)


garypen (approved):
I don't think it's controversial. It makes sense.

PR #6966, opened 2025-03-10
Error categorization: touch-ups around configuration

timbotnik:
In 2.0.0, an experimental option `telemetry.apollo.errors.experimental_otlp_error_metrics` was introduced to track errors with additional attributes. A few related changes are included here:

- Sending these metrics now also respects the subgraph's `send` flag e.g. `telemetry.apollo.errors.subgraph.[all|(subgraph name)].send`.
- A new configuration option `telemetry.apollo.errors.subgraph.[all|(subgraph name)].redaction_policy` has been added. This flag only applies when `redact` is set to `true`. When set to `ErrorRedactionPolicy.Strict` (the default), error redaction will behave as it has in the past. Setting this to `ErrorRedactionPolicy.Extended` will allow the `extensions.code` value from subgraph errors to pass through redaction and be sent to Studio.
- A warning about incompatibility of error telemetry with connectors will be suppressed when this feature is enabled, since it _does_ support connectors when using the new mode.

We are planning a larger docs change around this feature so only addressing the autogenerated docs in this PR, but more to come on that front.

Review by commit as they are broken out semantically.

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [x] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/plugins/telemetry/apollo.rs
```diff
@@ -142,13 +142,17 @@ pub(crate) struct ErrorConfiguration {
...
+    /// even when `redact` is set to `true`.  Has no effect when `redact` is false.
+    pub(crate) redaction_policy: ErrorRedactionPolicy,
 }
 
 impl Default for ErrorConfiguration {
     fn default() -> Self {
         Self {
             send: true,
             redact: true,
+            redaction_policy: ErrorRedactionPolicy::Strict,
```
bonnici, line 155: When we mark the config as non-experimental, should we also update this default to be extended?

bonnici (approved):
Looks good to me - just added a question.

rregitsky (approved):
lgtm too after Nick's question is addressed

apollo-router/src/plugins/telemetry/apollo.rs
timbotnik: I'm on the fence here - I think it would ideally be on by default, but I'd love to confirm that with someone on the Router side.  Will ask them for their thoughts before shipping the non-experimental flag.

bnjjj: I think right now our policy in the router would probably to keep Strict as default to be more conservative by default.
Also I would suggest to use:
```suggestion
            redaction_policy: ErrorRedactionPolicy::default(),
```

It would be easier for us to change the default behavior in the future by just changing the default impl of `ErrorRedactionPolicy`

apollo-router/src/plugins/telemetry/tracing/apollo_telemetry.rs
```diff
@@ -1053,6 +1055,27 @@ pub(crate) fn extract_ftv1_trace(
     None
 }
 
+fn perform_exteneded_redaction(error_json: &str) -> String {
```
bnjjj, line 1058:
```suggestion
fn perform_extended_redaction(error_json: &str) -> String {
```

apollo-router/src/plugins/telemetry/tracing/apollo_telemetry.rs
```diff
@@ -1063,7 +1086,14 @@ fn preprocess_errors(
             t.error.iter_mut().for_each(|err| {
                 err.message = String::from("<redacted>");
                 err.location = Vec::new();
-                err.json = String::new();
+                err.json = if matches!(
+                    error_config.redaction_policy,
+                    ErrorRedactionPolicy::Extended
+                ) {
+                    perform_exteneded_redaction(&err.json)
```
bnjjj, line 1093: Sorry maybe I don't understand something, but why do we keep the whole error in json instead of just the extension code as we only use extension code ? It would prevent us from deserializing it from json everytime

apollo-router/src/plugins/connectors/incompatible/telemetry.rs
```diff
@@ -4,40 +4,48 @@ use itertools::Itertools as _;
...
+        Some(TelemetryIncompatPlugin {
+            config: Conf::apollo(config).errors,
+        })
     }
 }
 
 impl IncompatiblePlugin for TelemetryIncompatPlugin {
     fn is_applied_to_all(&self) -> bool {
-        self.config.all.send || self.config.all.redact
+        self.config.subgraph.all.send
```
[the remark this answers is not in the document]
timbotnik: It's just extra noise, if `send` is false then we'll never look at `redact`.  Also, it was checking the wrong bool value since the default is `redact: true`.

apollo-router/src/plugins/telemetry/tracing/apollo_telemetry.rs
timbotnik: I think mainly the reason is that the data model for errors in this spot comes from protobuf, and the only place to add the `code` is in the `json` field.  I don't think it's an easy refactor to change this but maybe I'm wrong.
```
    pub struct Error {
        /// required
        #[prost(string, tag = "1")]
        pub message: ::prost::alloc::string::String,
        #[prost(message, repeated, tag = "2")]
        pub location: ::prost::alloc::vec::Vec<Location>,
        #[prost(uint64, tag = "3")]
        pub time_ns: u64,
        #[prost(string, tag = "4")]
        pub json: ::prost::alloc::string::String,
    }
```

bnjjj: Ah ok sorry I didn't see it was coming from protobuf.

bnjjj (approved):
LGTM, just to make sure @timbotnik I assume you don't introduce breaking changes in the configuration ?

PR #6982, opened 2025-03-11
allow @connect on OBJECT

lennyburdette:
- easiest to review commit-by-commit. i'm trying to sequence changes so that they're easier to review
- does not allow @connect on types for connect/v0.1 via a validation rule, so it's safe to merge to dev and could be released
    - i'm not thrilled by the validation code, but that can be addressed in the larger validation refactoring
- most of the work is changing the `ConnectId` to contain references to types, not just fields
- i'll wait until the feature is complete to add a changelog entry

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/services/connector_service.rs
```diff
@@ -146,7 +146,7 @@ impl tower::Service<ConnectRequest> for ConnectorService {
                 "otel.kind" = "INTERNAL",
                 "apollo.connector.type" = CONNECTOR_TYPE_HTTP,
                 "apollo.connector.detail" = tracing::field::Empty,
-                "apollo.connector.field.name" = %connector.field_name(),
+                "apollo.connector.coordinate" = %connector.id.coordinate(),
```
dylan-apollo, line 149: Are we relying on these field names at all yet? Gonna mess up any queries?

Also, does that `otel.kind = "INTERNAL"` mean that these fields are not exposed to users, so they can't be relying on them?

apollo-federation/src/sources/connect/spec/type_and_directive_specifications.rs
```diff
@@ -38,6 +38,12 @@ use crate::sources::connect::spec::schema::HTTP_HEADER_MAPPING_NAME_ARGUMENT_NAM
 use crate::sources::connect::spec::schema::HTTP_HEADER_MAPPING_VALUE_ARGUMENT_NAME;
 use crate::sources::connect::spec::schema::SOURCE_BASE_URL_ARGUMENT_NAME;
 
+const CONNECT_V0_1_LOCATIONS: &[DirectiveLocation] = &[DirectiveLocation::FieldDefinition];
+const CONNECT_V0_2_LOCATIONS: &[DirectiveLocation] = &[
+    DirectiveLocation::FieldDefinition,
+    DirectiveLocation::Object,
+];
```
dylan-apollo, line 45: Feels like this should be a func on `ConnectSpec` and we should use `ConnectSpec` below

apollo-federation/src/sources/connect/validation/test_data/v0_1/connect_on_object.graphql
```diff
@@ -0,0 +1,19 @@
...
+    import: ["@source", "@connect"]
+  )
+
+type Query {
+  products: [Product]
+    @connect(http: { GET: "http://localhost:4001/products" }, selection: "id")
+}
+
+type Product
+  @connect(
```
dylan-apollo, line 14: Can you also add a happy-path case for 0.2?

apollo-federation/src/sources/connect/id.rs
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    pub(crate) fn parent_type_name(&self) -> Option<Name> {
+        match self {
+            ConnectorPosition::Field(pos) => Some(pos.field.type_name().clone()),
+            ConnectorPosition::Type(_) => None,
+        }
+    }
+
+    pub(crate) fn root_type_name(&self, schema: &Schema) -> Option<NamedType> {
```
dylan-apollo, line 72: This function name is a bit confusing to me, since "root type" typically refers to something like `Query` or `Mutation`. Maybe this is `return_type_name` (if I understand the intent)?

apollo-federation/src/sources/connect/id.rs
```diff
@@ -0,0 +1,118 @@
...
+use apollo_compiler::schema::ExtendedType;
+
+use crate::error::FederationError;
+use crate::schema::position::ObjectOrInterfaceFieldDirectivePosition;
+
+#[derive(Debug, Clone, Hash, PartialEq, Eq)]
+pub(crate) struct ObjectTypeDefinitionDirectivePosition {
+    type_name: Name,
+    directive_name: Name,
+    directive_index: usize,
```
dylan-apollo, line 17: This is specifically the connect directive index, right? Not the index into all applied directives? I don't think we necessarily need a longer field name here, but maybe a doc comment to specify?

apollo-federation/src/sources/connect/id.rs
```diff
@@ -0,0 +1,118 @@
...
+
+    pub(crate) fn synthetic_name(&self) -> String {
+        match self {
+            ConnectorPosition::Field(pos) => format!(
+                "{}_{}_{}",
+                pos.field.type_name(),
+                pos.field.field_name(),
+                pos.directive_index,
+            ),
+            ConnectorPosition::Type(pos) => format!("{}_{}", pos.type_name, pos.directive_index,),
```
dylan-apollo, line 115:
```suggestion
            ConnectorPosition::Type(pos) => format!("{}_{}", pos.type_name, pos.directive_index),
```
When are they going to teach `rustfmt` about `format!`? 😆

apollo-federation/src/sources/connect/models.rs
```diff
@@ -131,7 +130,12 @@ impl Connector {
 
         let transport = HttpJsonTransport::from_directive(connect_http, source_http)?;
 
-        let parent_type_name = connect.position.field.type_name().clone();
+        let parent_type_name = connect.position.parent_type_name().ok_or_else(|| {
```
dylan-apollo, line 133: So this isn't called for `@connect` on `type`?

apollo-router/src/services/connector_service.rs
lennyburdette: I think internal refers to the kind of computation being performed: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md

| `spanKind` | Call direction | Communication style |
| --------- | --------- | -------------- |
| CLIENT | outgoing | request/response |
| SERVER | incoming | request/response |
| PRODUCER | outgoing | deferred execution |
| CONSUMER | incoming | deferred execution |
| INTERNAL | |

apollo-federation/src/sources/connect/models.rs
lennyburdette: not til a later PR

apollo-federation/src/sources/connect/id.rs
lennyburdette: "return type" make sense for connectors on fields, but not for connectors on objects. this is the "root" of the connector — the base type the `selection` starts from. i'm fine with another word — i'll ask an llm for suggestions

lennyburdette: - base_type_name
- primary_type_name
- core_type_name
- main_type_name
- origin_type_name
- source_type_name
- fundamental_type_name
- principal_type_name

"base type" might be the best?

apollo-federation/src/sources/connect/id.rs
lennyburdette: this is consistent with the other Position structs — it's the position in the full list of directives. it doesn't really matter because it's just an identifier, not information that we use anywhere

apollo-router/src/services/connector_service.rs
lennyburdette: Screenshot 2025-03-13 at 10 21 03 AM

i guess we at least have a localized label for this field. i'll keep looking into it

dylan-apollo (approved):

apollo-router/src/uplink/license_enforcement.rs
```diff
@@ -508,7 +508,7 @@ impl LicenseEnforcementReport {
                 spec_url: "https://specs.apollo.dev/connect".to_string(),
                 version_req: semver::VersionReq {
                     comparators: vec![semver::Comparator {
-                        op: semver::Op::Exact,
+                        op: semver::Op::GreaterEq,
```
dylan-apollo, line 511: Should we have a semver check here at all if we still need to do other enforcement of connect spec version later?

If we _do_ need some check here, it should probably be <= 0.2

apollo-router/src/uplink/feature_gate_enforcement.rs
```diff
@@ -0,0 +1,299 @@
...
+struct ParsedLinkSpec {
+    version: semver::Version,
+    spec_url: String,
+    url: String,
+}
+
+impl ParsedLinkSpec {
+    fn from_join_directive_args(
+        args: &[(Name, Node<ast::Value>)],
+    ) -> Option<Result<Self, url::ParseError>> {
```
dylan-apollo, line 34: Is it possible to use the `Link` type from `apollo-federation` so we don't have more definitions of that URI format? Might take some refactoring to reuse for join directive...

Surely we also have code for parsing these in extract subgraphs, but maybe that's a tangle?

apollo-router/src/uplink/feature_gate_enforcement.rs
```diff
@@ -0,0 +1,299 @@
...
+    fn validate_schema(
+        schema: &Schema,
+        schema_restrictions: &Vec<FeatureRestriction>,
+        configuration: &Configuration,
+    ) -> Vec<FeatureGateViolation> {
+        let link_specs_in_join_directive = schema
+            .supergraph_schema()
+            .schema_definition
+            .directives
+            .get_all("join__directive")
```
dylan-apollo, line 93: In theory, the `@link` will have the highest spec version, so we could just check that today, but we will have a minimum version at some point, too (e.g., removing 0.1 from Router 3.0) so I think it's correct to check them all (like you are).

Might be worth putting a comment in, though, as to why we check `join__directive` and not `link`?

lennyburdette: yeah i should add a comment. the top-level link is removed at some point in the expansion process, so all we have here are the join__directive links

apollo-router/src/uplink/feature_gate_enforcement.rs
lennyburdette: i should at least extract this code from license_enforcement, it's a copy-paste job

apollo-router/src/uplink/license_enforcement.rs
lennyburdette: remove the version requirement entirely and tested that license key enforcement still works as expected

PR #6986, opened 2025-03-11
connect runtime: find @connect on types

lennyburdette:
this is safe to merge because there's no active version of federation that allows connect on types

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-federation/src/sources/connect/id.rs
+++ b/apollo-federation/src/sources/connect/id.rs
@@ -12,9 +12,9 @@ use crate::schema::position::ObjectOrInterfaceFieldDirectivePosition;
 
 #[derive(Debug, Clone, Hash, PartialEq, Eq)]
 pub(crate) struct ObjectTypeDefinitionDirectivePosition {
-    type_name: Name,
-    directive_name: Name,
-    directive_index: usize,
+    pub(super) type_name: Name,
+    pub(super) directive_name: Name,
+    pub(super) directive_index: usize,
 }
 
 /// Stores information about the position of the @connect directive, either
@@ -118,5 +118,29 @@ impl ConnectorPosition {
             ),
             ConnectorPosition::Type(pos) => format!("{}_{}", pos.type_name, pos.directive_index),
         }
+    }
+
+    pub(super) fn on_query(&self, schema: &Schema) -> bool {
+        schema
+            .schema_definition
+            .query
+            .as_ref()
+            .map(|query| match self {
+                ConnectorPosition::Field(pos) => *pos.field.type_name() == query.name,
+                ConnectorPosition::Type(_) => false,
+            })
+            .unwrap_or_default()
+    }
+
+    pub(super) fn on_mutation(&self, schema: &Schema) -> bool {
+        schema
+            .schema_definition
+            .mutation
+            .as_ref()
+            .map(|mutation| match self {
+                ConnectorPosition::Field(pos) => *pos.field.type_name() == mutation.name,
+                ConnectorPosition::Type(_) => false,
+            })
+            .unwrap_or_default()
     }
 }
--- a/apollo-federation/src/sources/connect/id.rs
+++ b/apollo-federation/src/sources/connect/id.rs
@@ -1,3 +1,6 @@
+use std::fmt;
+use std::fmt::Display;
+use std::fmt::Formatter;
 use std::hash::Hash;
 
 use apollo_compiler::Name;
@@ -24,19 +27,6 @@ pub(crate) enum ConnectorPosition {
     Field(ObjectOrInterfaceFieldDirectivePosition),
     #[allow(unused)]
     Type(ObjectTypeDefinitionDirectivePosition),
-}
-
-/// Reifies the connector position into schema definitions
-#[derive(Debug)]
-pub(crate) enum ConnectedElement<'schema> {
-    Field {
-        parent_type: &'schema ExtendedType,
-        field_def: &'schema Component<FieldDefinition>,
-    },
-    Type {
-        #[allow(unused)]
-        type_def: &'schema ExtendedType,
-    },
 }
 
 impl ConnectorPosition {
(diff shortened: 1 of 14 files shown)


apollo-federation/src/sources/connect/id.rs
```diff
@@ -26,19 +29,6 @@ pub(crate) enum ConnectorPosition {
     Type(ObjectTypeDefinitionDirectivePosition),
```
dylan-apollo, line 29: I think you can remove the `#[allow(unused)]` now

apollo-federation/src/sources/connect/id.rs
```diff
@@ -119,4 +109,78 @@ impl ConnectorPosition {
...
+    pub(super) fn on_query(&self, schema: &Schema) -> bool {
+        schema
+            .schema_definition
+            .query
+            .as_ref()
+            .map(|query| match self {
+                ConnectorPosition::Field(pos) => *pos.field.type_name() == query.name,
+                ConnectorPosition::Type(_) => false,
+            })
+            .unwrap_or_default()
```
dylan-apollo, line 122:
```suggestion
            .is_some_and(|query| match self {
                ConnectorPosition::Field(pos) => *pos.field.type_name() == query.name,
                ConnectorPosition::Type(_) => false,
            })
```

apollo-federation/src/sources/connect/id.rs
```diff
@@ -119,4 +109,78 @@ impl ConnectorPosition {
...
+    pub(super) fn on_mutation(&self, schema: &Schema) -> bool {
+        schema
+            .schema_definition
+            .mutation
+            .as_ref()
+            .map(|mutation| match self {
+                ConnectorPosition::Field(pos) => *pos.field.type_name() == mutation.name,
+                ConnectorPosition::Type(_) => false,
+            })
+            .unwrap_or_default()
```
dylan-apollo, line 134:
```suggestion
            .is_some_and(|mutation| match self {
                ConnectorPosition::Field(pos) => *pos.field.type_name() == mutation.name,
                ConnectorPosition::Type(_) => false,
            })
```

apollo-federation/src/sources/connect/id.rs
```diff
@@ -119,4 +109,78 @@ impl ConnectorPosition {
...
+    }
+
+    pub(super) fn on_mutation(&self, schema: &Schema) -> bool {
+        schema
+            .schema_definition
+            .mutation
+            .as_ref()
+            .map(|mutation| match self {
+                ConnectedElement::Field { parent_type, .. } => *parent_type.name() == mutation.name,
+                ConnectedElement::Type { .. } => todo!(),
```
dylan-apollo, line 182:
```suggestion
                ConnectedElement::Type { .. } => false,
```

Also, looks like `on_query` and `on_mutation` are only used to check `on_root_type`, so should they be a single function to reduce code?

apollo-federation/src/sources/connect/id.rs
```diff
@@ -119,4 +109,78 @@ impl ConnectorPosition {
...
+    pub(super) fn on_query(&self, schema: &Schema) -> bool {
+        schema
+            .schema_definition
+            .query
+            .as_ref()
+            .map(|query| match self {
+                ConnectedElement::Field { parent_type, .. } => *parent_type.name() == query.name,
+                ConnectedElement::Type { .. } => false,
+            })
+            .unwrap_or_default()
```
dylan-apollo, line 172:
```suggestion
            .is_some_and(|query| match self {
                ConnectedElement::Field { parent_type, .. } => *parent_type.name() == query.name,
                ConnectedElement::Type { .. } => false,
            })
```

apollo-federation/src/sources/connect/models.rs
```diff
@@ -255,6 +233,22 @@ fn make_label(
     format!("{}{} {}", subgraph_name, source, transport.label())
 }
 
+fn determine_entity_resolver(
+    connect: &ConnectDirectiveArguments,
+    schema: &Schema,
+) -> Option<EntityResolver> {
+    let on_root_type = connect.position.on_query(schema) || connect.position.on_mutation(schema);
+
+    match connect.position {
```
dylan-apollo, line 242: If those functions, which match on position internally, are only used right before we match on position... I wonder if we should just include the is_root_type logic in this match here? Maybe that gets too messy.

apollo-federation/src/sources/connect/spec/directives.rs
```diff
@@ -96,6 +98,33 @@ pub(crate) fn extract_connect_directive_arguments(
...
         })
+        .chain(
+            // connect on types
+            schema
+                .types
+                .iter()
+                .filter_map(|(_, ty)| match ty {
+                    apollo_compiler::schema::ExtendedType::Object(node) => Some(node),
+                    _ => None,
+                })
```
dylan-apollo, line 109:
```suggestion
                .filter_map(|(_, ty)| ty.as_object())
```

apollo-federation/src/sources/connect/spec/directives.rs
```diff
@@ -96,6 +98,33 @@ pub(crate) fn extract_connect_directive_arguments(
...
+                .flat_map(|ty| {
+                    ty.directives
+                        .iter()
+                        .enumerate()
+                        .filter(|(_, directive)| directive.name == *name)
```
dylan-apollo, line 114:
```suggestion
                        .filter(|directive| directive.name == *name)
                        .enumerate()
```

apollo-federation/src/sources/connect/spec/directives.rs
```diff
@@ -96,6 +98,33 @@ pub(crate) fn extract_connect_directive_arguments(
...
+                        .map(move |(i, directive)| {
```
dylan-apollo, line 115:
```suggestion
                        .map(|(i, directive)| {
```

apollo-federation/src/sources/connect/validation/connect/selection.rs
```diff
@@ -61,30 +57,55 @@ pub(super) fn get_seen_fields_from_selection(
...
+                        code: Code::GraphQLError,
+                        message: format!("{coordinate} is valid only on object types"),
+                        locations: coordinate
+                            .directive
+                            .line_column_range(&schema.sources)
+                            .into_iter()
+                            .collect(),
+                    });
+                }
+            };
```
dylan-apollo, line 78: Can we store a `& Node<ObjectType>` in `ConnectedElement::Field` instead of `ExtendedType`? Should remove the need for checks in a few other places too, I think

apollo-federation/src/sources/connect/validation/expression.rs
```diff
@@ -41,36 +42,39 @@ impl<'schema> Context<'schema> {
...
+                Self {
+                    schema,
+                    var_lookup,
+                    source,
+                    code,
+                }
+            }
+            ConnectedElement::Type { .. } => Self {
+                schema,
+                var_lookup: Default::default(), // TODO
```
dylan-apollo, line 74: Maybe we should tag all these related TODOs with something so we can easily search for them when we're ready to fill out validations and such?

dylan-apollo (approved):
Looks good to me as another incremental step. Left a few suggestions around, only one blocker which is removing a `todo!` panic, I think

apollo-federation/src/sources/connect/validation/connect/selection.rs
lennyburdette: thinking ahead to interfaces

lennyburdette: though maybe we want a separate enum for interface/object so we don't have to deal with union/input/etc

apollo-federation/src/sources/connect/spec/directives.rs
lennyburdette: this seems right, but i think i need to figure out how to make this not FnMut? 
```
error: captured variable cannot escape `FnMut` closure body
  --> apollo-federation/src/sources/connect/spec/directives.rs:69:17
   |
67 |           .flat_map(|(type_name, fields, is_interface)| {
   |                                          ------------ variable defined here
68 |               fields.iter().flat_map(move |(field_name, field_def)| {
   |                                                                   - inferred to be a `FnMut` closure
69 | /                 field_def
70 | |                     .directives
71 | |                     .iter()
72 | |                     .enumerate()
...  |
75 | |                         let field_pos = if is_interface {
   | |                                            ------------ variable captured here
...  |
97 | |                         ConnectDirectiveArguments::from_position_and_directive(position, directive)
98 | |                     })
   | |______________________^ returns a closure that contains a reference to a captured variable, which then escapes the closure body
   |
   = note: `FnMut` closures only have access to their captured variables while they are executing...
   = note: ...therefore, they cannot allow references to captured variables to escape
help: consider adding 'move' keyword before the nested closure
   |
74 |                     .map(move |(i, directive)| {
   |                          ++++
```

apollo-federation/src/sources/connect/models.rs
lennyburdette: i'll look into this more when i fill out the logic for $batch. it certainly is wasteful to check for root type if the position is `::Type`

PR #6998, opened 2025-03-13
Add Metric to Measure Cardinality Overflow Frequency

rregitsky:
Adds a new counter metric, `apollo.router.telemetry.metrics.cardinality_overflow`, that is incremented when the [cardinality overflow log](https://github.com/open-telemetry/opentelemetry-rust/blob/d583695d30681ee1bd910156de27d91be3711822/opentelemetry-sdk/src/metrics/internal/mod.rs#L134) from [opentelemetry-rust](https://github.com/open-telemetry/opentelemetry-rust) occurs. This log means that a metric in a batch has reached a cardinality of > 2000 and that any excess attributes will be ignored.

This also moves the telemetry `error_handler` logic into its own file, `error_handler.rs`, in attempt to help clean up the module.

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


theJC:
This cardinality limit feature has been removed in the latest version of opentelemetry-rust.

Maybe best to upgrade the crate to the latest version, and not worry about this until the limit feature is re-introduced?

https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/CHANGELOG.md#0280

> Removed Metrics Cardinality Limit feature. This was originally introduced in https://github.com/open-telemetry/opentelemetry-rust/pull/1066 with a hardcoded limit of 2000 and no ability to change it. This feature will be re-introduced in a future date, along with the ability to change the cardinality limit.

rregitsky:
> This cardinality limit feature has been removed in the latest version of opentelemetry-rust.
> 
> Maybe best to put effort into upgrade the crate to the latest version, and not worry about this until the limit feature is re-introduced?
> 
> https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/CHANGELOG.md#0280
> 
> > Removed Metrics Cardinality Limit feature. This was originally introduced in [open-telemetry/opentelemetry-rust#1066](https://github.com/open-telemetry/opentelemetry-rust/pull/1066) with a hardcoded limit of 2000 and no ability to change it. This feature will be re-introduced in a future date, along with the ability to change the cardinality limit.

Hey @theJC -- yep we're aware of the fix in newer versions. There are some significant changes necessary to upgrade to v0.28 that will slow down that upgrade, but definitely still the intention long-term. We're looking to gather data on how often users are hitting this limit to inform the priority of that upgrade, especially in light of the high cardinality `experimental_otlp_error_metrics` introduced in router 2.0.

bonnici (approved):
Looks good to me but checking against an error string seems a bit brittle. Maybe just checking if the string contains "Maximum data points for metric stream exceeded" is slightly better? I'll leave the decision to you as you've definitely seen more of this error than I have.

rregitsky:
> Looks good to me but checking against an error string seems a bit brittle. Maybe just checking if the string contains "Maximum data points for metric stream exceeded" is slightly better? I'll leave the decision to you as you've definitely seen more of this error than I have.

Yeah, there's certainly a bit of brittleness here. I was following the pattern that already existed the error handler. My thinking is that this metric is temporary until we upgrade versions and at that point the log will no longer be emitted by the library anyway. I'm ok leaving it as is.

PR #7000, opened 2025-03-13
update the deny.toml to work with latest version of cargo deny

Velfi:
*Description here*
Exactly what it says on the tin. The changes make our `deny.toml` more restrictive.

- I removed the keys that `cargo deny` no longer supports.
- I had to add licenses to the examples. For those, I just pointed to the license file at the repo's root.
- I had to allowlist three unmaintained crates that we rely on. I'd like to address this in a separate PR b/c I think the changes required will be extensive.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


deny.toml
```diff
@@ -64,9 +60,6 @@ allow = [
     "Unicode-3.0",
     "Zlib"
 ]
-copyleft = "warn"
-allow-osi-fsf-free = "neither"
```
goto-bus-stop, line 68: What does removing this mean? Is it also making something stricter?

Velfi: I believe yes, this is more strict now. The tool no longer allows you to `warn`; Something is either an error or it isn't. Currently, we're only going to error on licenses of crates that we directly depend on, and we don't depend on any copyleft, osi/fsf free stuff. If we do start depending on those in the future, we can allowlist them.

deny.toml
```diff
@@ -36,15 +30,17 @@ ignore = [
     # a `[bans]` entry below. in the prometheus crates, only the protobuf
     # encoder is used, while only the decoder is affected by this advisory.
     "RUSTSEC-2024-0437",
+
+    # The following crates are unmaintained
```
duckki, line 34: Do we have a Jira tick to track this TODO?

deny.toml
goto-bus-stop: okay, i agree based on this:
https://github.com/EmbarkStudios/cargo-deny/blob/1c8b8908d9db080ca4e226905c25cf8dff38bc01/docs/src/checks/licenses/cfg.md?plain=1#L59

goto-bus-stop (approved):
Thanks a lot for picking this up 😁

deny.toml
Velfi: https://apollographql.atlassian.net/browse/ROUTER-1189

PR #7001, opened 2025-03-13
refactor: 1.x branch style edition 2024

duckki:
* added `rustfmt.toml` file with `style_edition = "2024"` under `apollo-federation` and `apollo-router`.
* ran `cargo fmt` as usual.
* to keep difference between `dev` and `1.x` as small as we can.


abernix (dismissed):
sweet. can we please make sure to use [this](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-and-understanding-files#ignore-commits-in-the-blame-view) feature?

briefly to-use:

1. add the commit `0bc12718d6f200956fecf4195c15fea7733ddb9a` to a `.git-blame-ignore-revs` at the root.
2. commit that change to this PR.
3. **true-merge** the PR (**don’t squash**. this is important. :wink: ) 

**feel free to reject this request if applied.**

SimonSapin (approved):

PR #7003, opened 2025-03-13
Partial implementation for programmatic subgraph schema construction

tninesling:
Implements [these two tests](https://github.com/apollographql/federation/blob/main/internals-js/src/__tests__/federation.test.ts) from JS Federation, which simply check whether a schema is a Fed 1 subgraph or a Fed 2 subgraph. The `SubgraphMetadata` tests already asserted this would work correctly when given a schema with all the appropriate directive definitions filled in, however, many of our tests take a very small schema and embellish it programmatically with various definitions.

The constructor for `FederationSchema` is now broken down into a few parts, primarily because we need to inject intermediate steps for adding definitions. When building the schema, we do the shallow definition pass, then we inject a definition for `@link` if it's missing and add the `@link` to the link spec at the top of the schema. From there, the existing code can properly bootstrap the metadata from the injected link information.

This is a partial implementation of the schema building code, following the steps in [buildSchemaFromAST](https://github.com/apollographql/federation/blob/main/internals-js/src/buildSchema.ts#L73-L181). I also updated the `SpecDefintion` trait to auto-delegate to the existing `checkOrAdd` functionality for filling in definitions. All specs other than the link specification have their directive and type specs stubbed out. In the future, we will need to update the signature of `checkOrAdd` to take in a `Link` (or alternative representation of JS's `CoreFeature`) so we can account for renamed imports when generating the definitions.

---

**Checklist**

- Tests added and passing[^3]
    - [X] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


clenfest (approved):
I'm not sure about the MultipleFederationErrors class, but we can revisit that later.

PR #7005, opened 2025-03-13
Add configuration option to limit maximum batch size

carodewig:
Add an optional `maximum_size` parameter to the batching configuration. 
* When specified, the router will reject requests which contain more than `maximum_size` queries in the client batch.
* When unspecified, the router performs no size checking (the current behavior).

I've also added tests for the configuration and router, and updated the documentation to reflect the new parameter.

NB: I ended up using HTTP code 422 (`Unprocessable Content`) as the error code when a batch is rejected due to the maximum size limit. My original plan was to use 413 (`Content Too Large`), but the error ended up being mangled by the `limit` plugin.

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-1096]: https://apollographql.atlassian.net/browse/ROUTER-1096?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


garypen (commented):
Looks great. My only comment is to update the changeset to document the new error path.

.changesets/feat_limit_client_batch_size.md
```diff
@@ -0,0 +1,8 @@
+### Add configuration option to limit maximum batch size ([PR #7005](https://github.com/apollographql/router/pull/7005))
+
+Add an optional `maximum_size` parameter to the batching configuration.
+
+* When specified, the router will reject requests which contain more than `maximum_size` queries in the client batch.
+* When unspecified, the router performs no size checking (the current behavior).
+
```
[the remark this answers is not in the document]
carodewig: Good point - addressed in b1d7683a8.

apollo-router/src/configuration/mod.rs
```diff
@@ -1409,6 +1409,10 @@ pub(crate) struct Batching {
 
     /// Subgraph options for batching
     pub(crate) subgraph: Option<SubgraphConfiguration<CommonBatchingConfig>>,
+
+    /// Maximum size for a batch
+    #[serde(default)]
+    pub(crate) maximum_size: Option<usize>,
```
BrynCooke, line 1415: Should we have a reasonable default max batch size?

BrynCooke (approved):

apollo-router/src/services/router/service.rs
```diff
@@ -682,6 +682,19 @@ impl RouterService {
                 }
             }
         };
+
+        if is_batch && self.batching.exceeds_batch_size(&result) {
+            return Err(TranslateError {
+                status: StatusCode::UNPROCESSABLE_ENTITY,
+                extension_code: "CONTENT_TOO_LARGE".to_string(),
```
BrynCooke, line 689:
```suggestion
                extension_code: "BATCH_LIMIT_EXCEEDED".to_string(),
```

BrynCooke (changes requested):

apollo-router/src/services/router/service.rs
goto-bus-stop: (don't forget to update the changeset when applying this change 😇 )

apollo-router/src/configuration/mod.rs
carodewig: I wondered about this too, but didn't for two reasons:
* I wasn't sure what a reasonable batch size would be, and
* I was concerned about the user experience if someone upgraded and started seeing batch size errors, for a batch size they didn't set

apollo-router/src/services/router/service.rs
carodewig: Good call! Addressed in 8831824 (to bundle all the changes in one)

BrynCooke (approved):

goto-bus-stop (approved):
Great 🥳

carodewig:
@mergifyio backport 1.x

PR #7006, opened 2025-03-13
Ensure build_query_plan() cancels when the request cancels (manual backport #6840)

duckki:
This is my manual backport of https://github.com/apollographql/router/pull/6840.
It depends on `https://github.com/apollographql/router/pull/7001` to make the changeset a bit smaller.

### Original PR description
When a request is cancelled, resources (compute time/memory) allocated to that request should be released. However, query planning is an exception, and currently keeps running after the request is cancelled. This PR adds cooperative cancellation to build_query_plan(), to ensure that when the request is cancelled, query planning errors shortly after.


goto-bus-stop:
Thanks for doing this, but I think we should not backport this individually as it doesn't work yet, and rather backport the whole feature once it's complete

PR #7020, opened 2025-03-18
Add more flexible subgraph mocks

SimonSapin:
Many Router tests rely on `MockSubgraph` a plugin that intercepts subgraph requests about to be sent over the network and instead returns pre-determined response directly. It requires specifying exactly what subgraph queries are expected:

```rust
MockSubgraph::builder().with_json(
    serde_json::json!{{
        "query": "query($representations:[_Any!]!){_entities(representations:$representations){...on Contact{__typename country}}}",
        "variables": {
            "representations": [
                {
                    "__typename":"Contact",
                    "id":"0",
                    "displayName": "Max",
                }
            ]
        }
    }},
    serde_json::json!{{"data": {
        "_entities": [{
            "__typename":"Contact",
            "country": "Fr"
        }]
    }}}
).with_json(
    serde_json::json!{{
        "query": "query($representations:[_Any!]!){_entities(representations:$representations){...on Contact{country}}}",
        "variables": {
            "representations": [
                {
                    "__typename":"Contact",
                    "id":"0",
                    "displayName": "Max",
                }
            ]
        }
    }},
    serde_json::json!{{"data": {
        "_entities": [{
            "country": "Fr"
        }]
    }}}
).build()
```

Exact subgraph queries are difficult to predict because they are generated by the query planner. So writing a new test involves starting with an empty `MockSubgraph`, letting it fail when the generated subgraph query is unexpected, hacking some logging to exfiltrate the expected string somehow, and copy pasting it into the right place. Overall not a pleasant experience.

This PR adds a new plugin that intercepts subgraph requests and responds with mocked data. Data is specified in Router YAML configuration. Configuring the same data as above looks like this:

```yaml
subgraph_mocks:
  subgraph1:
    entities:
      - __typename: "Contact"
        id: "0"
        displayName: "Max"
        country: "Fr"
```

There is no specification of expected queries. The new plugin is based on a full-featured GraphQL execution engine that can respond to any request. (The plugin does not yet support field arguments other than Federation’s `_entities(representations)` but that limitation exists only in data configuration, not in the execution engine.)

The execution engine is a copy of the one that apollo-compiler contains to support schema introspection, but is duplicated because apollo-compiler does not expose a public API for it. This may change if we figure out a nice rustic API for specifying resolver functions.

To avoid bloating the `apollo_router` crate, this plugin is only available in tests reachable from `apollo-router/tests/integration_tests.rs`.


apollo-router/tests/integration/subgraph_mocks_plugin/mod.rs
```diff
@@ -0,0 +1,270 @@
...
+struct MockResolver<'a> {
+    mocks: &'a JsonMap,
+}
+
+impl<'a> RootResolver<'a> {
+    fn find_entities(&self, representation: &JsonMap) -> Option<&'a JsonMap> {
+        self.entities.iter().find(|entity| {
+            representation
+                .iter()
+                .all(|(k, v)| entity.get(k).is_some_and(|value| value == v))
```
SimonSapin, line 180: This linear search would not be appropriate for a production server with many entities, but mocks typically have very few

apollo-router/src/router_factory.rs
```diff
@@ -765,6 +765,15 @@ pub(crate) async fn create_plugins(
...
     add_user_plugins!();
 
+    if apollo_plugin_factories.contains_key("apollo.subgraph_mocks") {
+        // Defined in apollo-router/tests/integration/subgraph_mocks_plugin/mod.rs,
+        // only present when running that test executable.
+        // Needs to come after user plugins
+        // but before `__internal_for_test_harness_prevent_all_subgraph_requests`.
+        add_optional_apollo_plugin!("subgraph_mocks");
+    }
+    add_optional_apollo_plugin!("__internal_for_test_harness_prevent_all_subgraph_requests");
```
SimonSapin, line 775: I’m not really happy with hard-coding so much here but I don’t know how else to manage plugin ordering.

Going through `TestHarness::extra_plugin` instead of `register_plugin!` does give control over ordering, but requires an already-instanciated plugin which didn’t go through `Plugin::new` and did get access to `PluginInit` and through it subgraph schemas. Exfiltrating subgraph schemas from `TestHarness` does not look easy since there are many layers of nested types (including a `Box<dyn Service>`)

apollo-router/tests/integration/entity_cache.rs
```diff
@@ -0,0 +1,985 @@
...
+                    "__typename": "Organization",
+                    "id": "1",
+                    "creatorUser": {
+                        "__typename": "User",
+                        "id": 2
+                    }
+                }
+            ]
+        },
+    });
```
SimonSapin, line 88: Parsing a YAML string could be nicer (less punctuation heavy), I stuck with `json!` macro here because that’s what we do everywhere for router config in `TestHarness` tests

apollo-router/tests/integration/subgraph_mocks_plugin/mod.rs
```diff
@@ -0,0 +1,270 @@
...
+///         field: [42, 7]
+/// ```
+///
+/// Example with mocked data in separate `mocked_subgraph_data.yaml` file:
+/// TODO: what’s the base path for this relative filename?
+///
+/// ```yaml
+/// subgraph_mocks: ${file.mocked_subgraph_data.yaml}
+/// ```
+type Config = HashMap<String, Arc<SubgraphConfig>>;
```
lrlna, line 46: I was going to raise this during the presentation you were giving, but we ran out of time. Could you write either a confluence doc or a dev doc in the repo on how you'd like people to use this new and improved way of mocking subgraphs? I think if a month from now I'll go to write a test like this, I am going to forget which files to look for examples in 😅

SimonSapin:
Superseded by https://github.com/apollographql/router/pull/7060

PR #7023, opened 2025-03-18
Add apollo.router.open_connections metric

BrynCooke:
To help users to diagnose when connections are keeping pipelines hanging around the following metric has been added:
- `apollo.router.open_connections` - The number of request pipelines active in the router
  - `schema.id` - The Apollo Studio schema hash associated with the pipeline.
  - `launch.id` - The Apollo Studio launch id associated with the pipeline (optional).
  - `config.hash` - The hash of the configuration.
  - `server.address` - The address that the router is listening on.
  - `server.port` - The port that the router is listening on if not a unix socket.
  - `state` - Either `active` or `terminating`.

Connections can be held open by clients via keepalive or even just a long running request, so it's useful to know when this is happening.

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [x] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/plugins/telemetry/config_new/instruments.rs
```diff
@@ -985,6 +987,54 @@ impl InstrumentsConfig {
                     .init(),
             ),
         );
+
+        // This new metric will also have the pipeline information.
+        // The reason that this instrument has a different name to the one above is that we wish to also implement this in router 2.0
```
goto-bus-stop, line 992: This entire comment can probably go on 2.x since the other metric doesn't exist here.
```suggestion
        // This new metric will also have the pipeline information.
        // The reason that this instrument has a different name to the one above is that we wish to also implement this in router 2.0
```

apollo-router/src/plugins/telemetry/config_new/instruments.rs
```diff
@@ -985,6 +987,54 @@ impl InstrumentsConfig {
...
+            OPEN_CONNECTIONS_METRIC.to_string(),
+            StaticInstrument::GaugeU64(
+                meter
+                    .u64_observable_gauge(OPEN_CONNECTIONS_METRIC)
+                    .with_description("Number of currently connected clients")
+                    .with_callback(move |gauge| {
+                        let connections =
+                            crate::axum_factory::connection_handle::connection_counts();
+                        for (connection, count) in connections.iter() {
+                            let mut attributes = Vec::with_capacity(5);
```
goto-bus-stop, line 1003: oh, the # of attributes when listening on a port is actually 5 or 6. should we use 6 here? (`server.address`, `server.port`, `schema.id`, `config.hash`, `state` is 5, + optionally `launch.id`)
```suggestion
                            let mut attributes = Vec::with_capacity(6);
```

apollo-router/src/plugins/telemetry/config_new/instruments.rs
BrynCooke, line 992:
```suggestion
```

BrynCooke: [a99b852](https://github.com/apollographql/router/pull/7023/commits/a99b8522422458273297bf098ff2e62c9108d049)

goto-bus-stop (approved):

apollo-router/src/plugins/telemetry/config_new/instruments.rs
BrynCooke: I also fixed this on the 1.x branch

PR #6906, opened 2025-03-04
Feature/6897 - Allow scheme for subgraph to be inspected and updated

starJammer:
In my organization's case, we have multi-region services. When accessing services within the same region we want to use the `http` scheme and hostnames local to the region which allow quicker communication. When accessing services outside the region we want to use a different host name which uses the `https`. With current functionality the scheme cannot be updated via a simple Rhai script.

This patch adds the ability to set the scheme vie `request.subgraph.uri.scheme`.

Fixes #6897

---

**Checklist**

Complete the checklist (and note appropriate exceptions) before the PR is marked ready-for-review.

- [x] Changes are compatible[^1]
- [x] Documentation[^2] completed
- [ ] Performance impact assessed and acceptable
- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[^1]: It may be appropriate to bring upcoming changes to the attention of other (impacted) groups. Please endeavour to do this before seeking PR approval. The mechanism for doing this will vary considerably, so use your judgement as to how and when to do this.
[^2]: Configuration is an important part of many changes. Where applicable please try to document configuration examples.
[^3]: Tick whichever testing boxes are applicable. If you are adding Manual Tests, please document the manual testing (extensively) in the Exceptions.

Issue #6897, opened 2025-03-03
Rhai: Allow setting or overriding request scheme


starJammer:
## Is your feature request related to a problem? Please describe.
In my organization's case, we have multi-region services. When accessing services within the same region we want to use the `http` scheme and hostnames local to the region which allow quicker communication. When accessing services outside the region we want to use a different host name which uses the `https`. With current functionality the scheme cannot be updated via a simple Rhai script.

## Describe the solution you'd like
The ability to set the scheme vie `request.subgraph.uri.scheme`.

## Describe alternatives you've considered
I looked at coprocessors but they also add more latency and complexity. Rhai scripts are 'simpler' and already provide some ability to change the host and port values.

Another alternative was having the GraphQL router configured to point to a proxy which will then forward onto the correct regionalized subgraph. This is similar to the coprocessor and will be less performant and introduce more complexity and fragility to the system. The proxy I tested with was the Spring Cloud Gateway (SCG). I pointed the router to a subgraph specific URL on the Spring Cloud Gateway and the SCG forwarded the request to the correct regionalized service with the correct http scheme using the requests JWT token. This introduced a lot of latency per subgraph per request, however.

starJammer:
thanks @goto-bus-stop . Any further direction on how to get this reviewed/resolved/merged would be appreciated from my side. Thank you for linking the issue and the pull request. I'll keep that in mind in the future.

PR #7025, opened 2025-03-18
Support to get/set URI scheme in Rhai

goto-bus-stop:
This pulls in https://github.com/apollographql/router/pull/6906 so we can run GraphOS tests.

Closes https://github.com/apollographql/router/pull/6906
Fixes https://github.com/apollographql/router/issues/6897


abernix (approved):

abernix:
Thank you for the contribution! @goto-bus-stop did the majority of the work. 😄 

Re: backport to 1.x — per our policy for Router 1.x going into LTS, we would **no longer add** anything but critical patches to 1.x.  Unfortunately, this would not be eligible for backporting under that policy.

PR #7027, opened 2025-03-18
use spec version to determine allowed headers

lennyburdette:
---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-federation/src/sources/connect/models.rs
+++ b/apollo-federation/src/sources/connect/models.rs
@@ -18,7 +18,6 @@ use apollo_compiler::parser::SourceSpan;
 use apollo_compiler::validation::Valid;
 use either::Either;
 use http::HeaderName;
-use http::header;
 use keys::make_key_field_set_from_variables;
 use serde_json::Value;
 use url::Url;
@@ -46,6 +45,7 @@ use crate::sources::connect::spec::schema::HEADERS_ARGUMENT_NAME;
 use crate::sources::connect::spec::schema::HTTP_HEADER_MAPPING_FROM_ARGUMENT_NAME;
 use crate::sources::connect::spec::schema::HTTP_HEADER_MAPPING_NAME_ARGUMENT_NAME;
 use crate::sources::connect::spec::schema::HTTP_HEADER_MAPPING_VALUE_ARGUMENT_NAME;
+use crate::sources::connect::spec::versions::SpecVersionConfig;
 
 // --- Connector ---------------------------------------------------------------
 
@@ -100,10 +100,10 @@ impl Connector {
         };
 
         let source_name = ConnectSpec::source_directive_name(&link);
-        let source_arguments = extract_source_directive_arguments(schema, &source_name)?;
+        let source_arguments = extract_source_directive_arguments(schema, &source_name, &spec)?;
 
         let connect_name = ConnectSpec::connect_directive_name(&link);
-        let connect_arguments = extract_connect_directive_arguments(schema, &connect_name)?;
+        let connect_arguments = extract_connect_directive_arguments(schema, &connect_name, &spec)?;
 
         connect_arguments
             .into_iter()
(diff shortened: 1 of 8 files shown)


apollo-federation/src/sources/connect/spec/versions.rs
```diff
@@ -0,0 +1,49 @@
...
+
+impl SpecVersionConfig for ConnectSpec {
+    /// These headers are not allowed to be defined by connect directives at all.
+    /// Copied from Router's plugins::headers
+    /// Headers from https://datatracker.ietf.org/doc/html/rfc2616#section-13.5.1
+    /// These are not propagated by default using a regex match as they will not make sense for the
+    /// second hop.
+    /// In addition, because our requests are not regular proxy requests content-type, content-length
+    /// and host are also in the exclude list.
+    fn header_name_is_reserved(&self, header_name: &HeaderName) -> bool {
```
dylan-apollo, line 19: Just to sketch out what I was referring to on a call—I wonder if we should make a new struct out of this and then use the type as an input to the validation function.

So rather than `Header` requiring a `ConnectSpec` to build, and then calling a specific function which branches, `Header` could require an `AllowedHeaders` struct. At the very start of validation, we create an `AllowedHeaders` based on _whatever_ inputs we want—today that might just be `ConnectSpec` but maybe in the future we're considering other things as well, like router config?

So we calculate the `AllowedHeaders` _once_, then use that everywhere headers are validated rather than branching based on spec for every single `@connect` and `@source`. A little more efficient, and hopefully models the dependencies of the stages a little better.

The other thing to consider is code organization and two different requirements seem to be at odds. On the one hand, we might want to see all the rules for a single connect spec in a single place. On the other, we may want to see all the rules for a single validation in a single place, regardless of spec. Personally I'm in favor of the latter—I think it's more likely when reading the code that we're trying to debug or modify "the rules for headers" rather than "the rules for the entirety of 0.2".

lennyburdette: I refactored things to use an AllowedHeaders struct. Right now I'm leaving it in versions.rs, but I can see how it could be nice to have it colocated with the other headers code. Does this match what you're thinking?

apollo-federation/src/sources/connect/models.rs
```diff
@@ -395,11 +398,12 @@ impl<'a> Header<'a> {
     /// Get a list of headers from the `headers` argument in a `@connect` or `@source` directive.
     pub(crate) fn from_headers_arg(
         node: &'a Node<ast::Value>,
+        spec: &AllowedHeaders,
```
dylan-apollo, line 401: May want to rename this var now

apollo-federation/src/sources/connect/models.rs
```diff
@@ -409,7 +413,10 @@ impl<'a> Header<'a> {
     }
 
     /// Build a single [`Self`] from a single entry in the `headers` arg.
-    fn from_single(node: &'a Node<ast::Value>) -> Result<Self, HeaderParseError<'a>> {
+    fn from_single(
+        node: &'a Node<ast::Value>,
+        spec: &AllowedHeaders,
```
dylan-apollo, line 418: And this one

apollo-federation/src/sources/connect/spec/versions.rs
```diff
@@ -0,0 +1,64 @@
+use http::HeaderName;
+use http::header;
+
+use super::ConnectSpec;
+
+impl ConnectSpec {
+    pub(crate) fn version_info(&self) -> VersionInfo {
+        VersionInfo::new(self)
+    }
+}
```
dylan-apollo, line 10: Should this be `impl From<ConnectSpec> for VersionInfo`? `ConnectSpec` should also be pass by value not reference since it's `Copy` I think.

dylan-apollo (approved):
LGTM, left a couple of suggestions

PR #7034, opened 2025-03-18
Support units in `metric!` macro

carodewig:
### Changes
- Update the `metric!` macro to optionally take a `unit` parameter
- Deduplicate some of the attribute parsing into a new macro `parse_attributes!`, which reduces a lot of duplication within each metric macro
- Add new `*_with_unit!` macros to support providing units and begin deprecating the existing macros
  I had hoped to be able to simply add a macro pattern to `f64_histogram!`, but was unable to get it to work properly. This was mostly due to `local ambiguity: multiple parsing options` errors where the macro couldn't work out which pattern to use:
  ```rust
  // the two patterns below cannot be disambiguated for the callsite below
  macro_rules! f64_histogram {  
      ($name:literal, $description:literal, $value: expr, $($attrs:tt)*) => {
          metric!(f64, histogram, record, $name, $description, $value, parse_attributes!($($attrs)*));
      };  
      
      ($name:literal, $description:literal, $unit:literal, $value: expr) => {
          metric!(f64, histogram, record, $name, $description, $unit, $value, []);
      };
  }
  f64_histogram!("test", "desc", "s", 1.0);
  ```
- ~~Replace many of the `f64_histogram!` callsites with `f64_histogram_with_unit!`, including the `query_planning.plan.duration` metric that spawned this effort~~

### Open questions

- [x] ~~Is it worth adding more `_with_unit` macros now? The vast majority of typed metrics seem to use `f64_histogram`, but it might be nice to have others ready to go.~~ 
  We decided to go ahead and add the new metrics now.
- [x] ~~By default, adding units will [change the names of Prometheus metrics](https://prometheus.io/docs/practices/naming/) (ie `apollo_router_cache_hit_time` -> `apollo_router_cache_hit_time_seconds`). My preference is to call out the specific affected metrics in a changeset, but we could also revert the change that modified those metrics.~~
  I did not change any existing metrics as those would be breaking changes. However I did mark the existing macros as `#[deprecated(since = "TBD")]` to suggest that new macros use the `_with_unit` variants.

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [x] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


bnjjj:
Hmm I didn't know about the renaming of prometheus metrics but it's considered as a breaking change and we don't want to introduce breaking changes in minor/patch versions. I'm wondering if this renaming really breaks the prometheus queries we could write or is it still valid using the old names without the unit ?

carodewig:
> Hmm I didn't know about the renaming of prometheus metrics but it's considered as a breaking change and we don't want to introduce breaking changes in minor/patch versions. I'm wondering if this renaming really breaks the prometheus queries we could write or is it still valid using the old names without the unit ?

Unfortunately it really does break the Prometheus queries - I had to change the test to accommodate the name change on account of the units.

I'll revert the changes I made to the metrics to avoid causing breaking changes. I think it's probably still worth including the metrics macro simplification and the new `f64_histogram_with_unit!` macro, but I'm also happy to close this PR if it no longer seems useful.

bnjjj:
I agree with you, we should keep the macros `*_with_unit` and we should even go further by marking the other ones as deprecated if we can. To make sure no ones will create new metrics without units. And add a comment on why it's deprecated, why we can't migrate the older ones. It would create clarity. Please could you also update this [doc](https://github.com/apollographql/router/blob/dev/dev-docs/metrics.md) and add why we should not use older ones and so one ? Also add it in comments in the codebase

carodewig:
> I agree with you, we should keep the macros `*_with_unit` and we should even go further by marking the other ones as deprecated if we can. To make sure no ones will create new metrics without units. And add a comment on why it's deprecated, why we can't migrate the older ones. It would create clarity. Please could you also update this [doc](https://github.com/apollographql/router/blob/dev/dev-docs/metrics.md) and add why we should not use older ones and so one ? Also add it in comments in the codebase

@bnjjj I think this is done! I added a bunch of new `_with_unit` macros, deprecated 'in the future' the existing macros, and added a bunch of documentation to `crate::metrics` and `dev-docs/metrics.md`.

While I was working on the documentation, I noticed that a lot of it was duplicated, so I moved a bunch of it to the module docs rather than duplicating it on each macro. But if you prefer, I can put it back - just thought it might be easier to maintain this way!

bnjjj (approved):

PR #7035, opened 2025-03-18
refactor: remove `alias` field from `requires_selection::Field`

duckki:
* The `alias` field is never used (always set `None`). This PR removes it.
* This is a follow-up of the de-serializable `QueryPlan` PR (#6861).

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

This is a pure refactoring.


dariuszkuc:
I think this was a known limitation that should be fixed instead?

duckki:
Indeed, there is a possible future where we may actually use the `alias` field. I'll close this for now.

PR #7036, opened 2025-03-18
Make `@connect(source:)` validation more self contained

dylan-apollo:
`validate_source_name` now returns the `SourceName` so it's more clear what's happening, and that this check has to pass before we can run checks which require it. Also moves the header check up, since that _doesn't_ require the source name.

### DIFF
--- a/apollo-federation/src/sources/connect/validation/connect.rs
+++ b/apollo-federation/src/sources/connect/validation/connect.rs
@@ -2,9 +2,9 @@
 
 use apollo_compiler::Name;
 use apollo_compiler::Node;
-use apollo_compiler::ast::Argument;
 use apollo_compiler::ast::FieldDefinition;
 use apollo_compiler::schema::Component;
+use apollo_compiler::schema::Directive;
 use apollo_compiler::schema::ExtendedType;
 use apollo_compiler::schema::ObjectType;
 use itertools::Itertools;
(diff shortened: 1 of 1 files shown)


apollo-federation/src/sources/connect/validation/connect.rs
```diff
@@ -260,16 +262,22 @@ fn fields_seen_by_connector(
...
+    let Some(source_name) = directive
+        .arguments
+        .iter()
+        .find(|arg| arg.name == CONNECT_SOURCE_ARGUMENT_NAME)
+    else {
+        return Ok(None);
+    };
 
-    if source_names.iter().all(|name| name != &source_name.value) {
+    source_names.iter().find(|name| **name == source_name.value).map(Some).ok_or_else(|| {
```
lennyburdette, line 280: this line took me a while to parse — the double `Some` is kinda weird, right? but i'm not liking any suggested options enough to demand a change :grin:

lennyburdette (approved):

apollo-federation/src/sources/connect/validation/connect.rs
dylan-apollo: Yeah, I think the intent of the code wasn't clear. I rewrote it in a less functional style and added a comment which I _think_ results in more clear code. What do you think?

lennyburdette: yeah i think that'll be easier for future us to read someday!

PR #7044, opened 2025-03-19
Add configuration option to limit maximum batch size (backport #7005)

carodewig:
Add an optional `maximum_size` parameter to the batching configuration. 
* When specified, the router will reject requests which contain more than `maximum_size` queries in the client batch.
* When unspecified, the router performs no size checking (the current behavior).

I've also added tests for the configuration and router, and updated the documentation to reflect the new parameter.

NB: I ended up using HTTP code 422 (`Unprocessable Content`) as the error code when a batch is rejected due to the maximum size limit. My original plan was to use 413 (`Content Too Large`), but the error ended up being mangled by the `limit` plugin.

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-1096]: https://apollographql.atlassian.net/browse/ROUTER-1096?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ<hr>This is an automatic backport of pull request #7005 done by [Mergify](https://mergify.com).

### DIFF
--- a/apollo-router/src/configuration/mod.rs
+++ b/apollo-router/src/configuration/mod.rs
@@ -1473,6 +1473,10 @@ pub(crate) struct Batching {
 
     /// Subgraph options for batching
     pub(crate) subgraph: Option<SubgraphConfiguration<CommonBatchingConfig>>,
+
+    /// Maximum size for a batch
+    #[serde(default)]
+    pub(crate) maximum_size: Option<usize>,
 }
 
 /// Common options for configuring subgraph batching
@@ -1505,6 +1509,13 @@ impl Batching {
                         .is_some_and(|x| x.enabled)
                 }
             }
+            None => false,
+        }
+    }
+
+    pub(crate) fn exceeds_batch_size<T>(&self, batch: &[T]) -> bool {
+        match self.maximum_size {
+            Some(maximum_size) => batch.len() > maximum_size,
             None => false,
         }
     }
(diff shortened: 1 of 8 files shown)


apollo-router/src/configuration/mod.rs
```diff
@@ -1473,6 +1473,10 @@ pub(crate) struct Batching {
 
     /// Subgraph options for batching
     pub(crate) subgraph: Option<SubgraphConfiguration<CommonBatchingConfig>>,
+
+    /// Maximum size for a batch
+    #[serde(default)]
```
bnjjj, line 1478: I know it's not you and was already bad but I'm wondering if we could refactor this struct to follow [this guideline](https://github.com/apollographql/router/blob/dev/dev-docs/yaml-design-guidance.md#use-serdedefault-on-struct-instead-of-fields-when-possible) ? It could be part of a follow-up PR

bnjjj: Also, should we set a safe default value ? To make sure by default it's still safe ?

bnjjj:
Oops, it's a backport. My bad I didn't see it. You can ignore my comments

apollo-router/src/configuration/mod.rs
carodewig: That's a good idea! I'll probably do it as a follow-up PR on the non-backport side.
RE the safe default - I didn't add one due to concerns about users suddenly encountering errors, but it might be a good idea for follow up as well? (Comment [here](https://github.com/apollographql/router/pull/7005#discussion_r2001495242))

PR #7045, opened 2025-03-19
expand connectors on types

lennyburdette:
---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-federation/src/sources/connect/expand/mod.rs
```diff
@@ -743,10 +720,54 @@ mod helpers {
             }
         }
 
-        /// Insert a query root for a connect field
+        /// Insert the parent type and field definition for a connector
+        fn insert_object_and_field(
```
lennyburdette, line 724: hopefully this refactoring makes sense — i separated out the code for adding the connected field and parent type from the code that adds the `schema { query: Query }` and `schema { mutation: Mutation }` (the former involves a dummy field if there are no query root fields in the schema)

apollo-federation/src/sources/connect/expand/tests/snapshots/it_expand_supergraph@batch.graphql.snap
```diff
@@ -0,0 +1,16 @@
+---
+source: apollo-federation/src/sources/connect/expand/tests/mod.rs
+expression: api_schema
+input_file: apollo-federation/src/sources/connect/expand/tests/schemas/expand/batch.graphql
+---
+directive @defer(label: String, if: Boolean! = true) on FRAGMENT_SPREAD | INLINE_FRAGMENT
```
dylan-apollo, line 6: Do you know if it's possible to name these snapshots something other than `-2`, `-3` and such? Glad the `expression` field tells me what I'm looking at, but might be nice for the file name to be descriptive as well.

apollo-federation/src/sources/connect/expand/mod.rs
```diff
@@ -330,161 +330,138 @@ mod helpers {
...
 
-                let field_pos = ObjectFieldDefinitionPosition {
-                    type_name: query_alias.clone(),
-                    field_name: field_def.name.clone(),
-                };
+                        _ => {
+                            return Err(FederationError::internal(
+                                "connect directives on interfaces, unions, enums, scalars, or input objects is not supported",
+                            ));
+                        }
```
dylan-apollo, line 451:
```rust
let ExtendedType::Object(object) = type_def else { return Err(...) }
```

_might_ be a little cleaner, would reduce some nesting.

dylan-apollo (approved):
Looks good to me, I think that refactor clarifies the code a bunch, too 👍

apollo-federation/src/sources/connect/expand/mod.rs
lennyburdette: the nesting was kinda intentional since we'll eventually have a branch for Interface, but sure!

apollo-federation/src/sources/connect/expand/tests/snapshots/it_expand_supergraph@batch.graphql.snap
lennyburdette: i can give the snapshot a name, which replaces the test name. so we'd end up with
```
apischema@batch.graphql.snap
apischema@carryover.graphql.snap
...
connectormap@batch.graphql.snap
connectormap@carryover.graphql.snap
...
supergraph@batch.graphql.snap
supergraph@carryover.graphql.snap
```

so they're sorted by snapshot, not by test input. that's probably fine?

PR #7047, opened 2025-03-19
support preserving key fields in connector response mapping

lennyburdette:
the query planner doesn't ask for key fields when it makes a batch entity request (because it already has them). but we'll need key fields in order to match the response objects with the entity references and ensure that the batch result is consistent with the request

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-federation/src/sources/connect/json_selection/selection_set.rs
+++ b/apollo-federation/src/sources/connect/json_selection/selection_set.rs
@@ -19,6 +19,7 @@
 use apollo_compiler::ExecutableDocument;
 use apollo_compiler::Node;
 use apollo_compiler::executable::Field;
+use apollo_compiler::executable::FieldSet;
 use apollo_compiler::executable::Selection;
 use apollo_compiler::executable::SelectionSet;
 use apollo_compiler::name;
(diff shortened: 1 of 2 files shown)


apollo-federation/src/sources/connect/json_selection/selection_set.rs
```diff
@@ -669,4 +691,53 @@ mod tests {
...
+        }
+        "###,
+            "./",
+        )
+        .unwrap();
+
+        let (document, selection_set) = selection_set(&schema, "{ product { name price } }");
+
+        let keys =
+            FieldSet::parse_and_validate(&schema, name!(Product), "id store { id }", "").unwrap();
```
benjamn, line 731: We still need a general way of extracting these key `FieldSet`s from a `JSONSelection`, right?

benjamn (approved):
Looks good. Is there any way to use this outside of the tests, yet?

apollo-federation/src/sources/connect/json_selection/selection_set.rs
lennyburdette: yeah, this currently only works with something like `{$this.id}/{$this.store.id}` so we still need a way to support `$this { id store { id } }`

lennyburdette:
i'll use it in the next PR!

PR #7058, opened 2025-03-20
Connection shutdown timeout 1.x

BrynCooke:

This PR introduces a configurable timeout from signal to terminate to actual termination defaulted to 60 seconds. After the timeout is reached the connection is forcibly terminated.

```
supergraph:
  connection_shutdown_timeout: 12s
```

```
supergraph:
  early_cancel: true
```

```
traffic_shaping:
  router:
    timeout: 10s
```

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


lrlna (approved):

.changesets/fix_bryn_connection_shutdown_timeout_1_x.md
```diff
@@ -0,0 +1,30 @@
+### Connection shutdown timeout 1.x ([PR #7058](https://github.com/apollographql/router/pull/7058))
+
+When a connection is closed we call `graceful_shutdown` on hyper and then await for the connection to close.
+
+Hyper 0.x has various issues around shutdown that may result in us waiting for extended periods for the connection to eventually be closed.
+
+This PR introduces a configurable timeout from signal to terminate to actual termination defaulted to 60 seconds. After the timeout is reached the connection is forcibly terminated.
```
carodewig, line 7: Nit: this took me a few reads to puzzle out - perhaps something like this would be clearer?
```suggestion
This PR introduces a configurable timeout from the termination signal to actual termination, defaulted to 60 seconds. The connection is forcibly terminated after the timeout is reached.
```

apollo-router/src/axum_factory/listeners.rs
```diff
@@ -217,9 +218,55 @@ pub(super) async fn get_extra_listeners(
...
+            // the shutdown receiver was triggered first,
+            // so we tell the connection to do a graceful shutdown
+            // on the next request, then we wait for it to finish
+            _ = connection_shutdown.notified() => {
+                connection_handle.shutdown();
+                connection.as_mut().graceful_shutdown();
+                // if the connection was idle and we never received the first request,
+                // hyper's graceful shutdown would wait indefinitely, so instead we
+                // close the connection right away
+                if received_first_request.load(Ordering::Relaxed) {
```
carodewig, line 244: I could very well be misunderstanding something, but this comment and condition seem to be at odds?

The comment suggests that we need to immediately close the connection if `received_first_request` is false, since the graceful shutdown will never terminate (hence the `connection.timeout().await`) below. But I don't see where that is being handled -- the code is checking if `received_first_request` is true.

BrynCooke: It's super confusing. I just copied it from the original code, but let's improve it. It's actually describing the opposite of what is in the block as the connection gets closed right away by not awaiting and just dropping the connection.

BrynCooke: [d212e23](https://github.com/apollographql/router/pull/7058/commits/d212e236665b0b2cbad0cfa202fab88193b2de6e)

docs/source/reference/router/configuration.mdx
```diff
@@ -1176,6 +1176,25 @@ supergraph:
   experimental_log_on_broken_pipe: true
 ```
 
+### Connection shutdown timeout
+
+When the Router schema or configuration updates all connections must be closed for the existing resources to be freed.
+To ensure that long-lived connections do hang on to these resources a maximum graceful shutdown timeout can be configured.
```
carodewig, line 1182: Quite possibly my misunderstanding again, but should this be the opposite? Isn't the graceful shutdown timeout in place to _prevent_ long-lived connections from hanging onto resources after they shouldn't?

.changesets/fix_bryn_connection_shutdown_timeout_1_x.md
```diff
@@ -0,0 +1,30 @@
...
+supergraph:
+  connection_shutdown_timeout: 60s
+```
+
+Note that even after connections have been terminated the router will still hang onto pipelines if `early_cancel` has not been configured to true. The router is trying to complete the request. 
+
+Users can either set `early_cancel` to `true` 
+```
+supergraph:
+  early_cancel: true
```
bnjjj, line 20: Should it be set to `true` by default ?

apollo-router/src/axum_factory/listeners.rs
lrlna: (this was copied over from below, but) Bryn and I were sitting there for solid 10 minutes last week saying exactly what you just wrote!! That's 3 people now!

apollo-router/src/axum_factory/listeners.rs
```diff
@@ -217,9 +218,55 @@ pub(super) async fn get_extra_listeners(
...
+                            "connection shutdown exceeded, forcing close",
+                        );
+                    }
+                }
+            }
+        }
+    };
+}
+
+#[allow(clippy::too_many_arguments)]
```
lrlna, line 264: not a fan of this, but i hear you, it's difficult to do anything else reasonably

lrlna (approved):

.changesets/fix_bryn_connection_shutdown_timeout_1_x.md
```diff
@@ -0,0 +1,30 @@
...
+```
+
+AND/OR use traffic shaping timeouts:
+```
+traffic_shaping:
+  router:
+    timeout: 60s
+```
```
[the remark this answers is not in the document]
BrynCooke: We should take about this as a team. I actually don't think the `early_cancel` option should exist. We should always cancel requests where the client has dropped.
Mutations could be a special case, but it's not as if we have a transactional system, so it seems unsafe to rely on us completing the request.

apollo-router/src/axum_factory/listeners.rs
```diff
@@ -217,9 +218,55 @@ pub(super) async fn get_extra_listeners(
     Ok(listeners_and_routers)
 }
 
+// This macro unifies the logic tht deals with connections.
+// Ideally this would be a function, but the generics proved too difficult to figure out.
+macro_rules! handle_connection {
```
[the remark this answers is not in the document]
BrynCooke: I tried to make it a function. I was not certain it was possible in the end. At least the compiler didn't give me the right clues to figure it out.

apollo-router/src/axum_factory/listeners.rs
BrynCooke: Agree. This needs a larger refactoring though.

.changesets/fix_bryn_connection_shutdown_timeout_1_x.md
BrynCooke: See my comment below. [We need to talk about `early_cancel`](https://github.com/apollographql/router/pull/7058#discussion_r2011709391)

docs/source/reference/router/configuration.mdx
BrynCooke: [54130cb](https://github.com/apollographql/router/pull/7058/commits/54130cb7c400d850a6525c16687016186ef94998)

abernix:
@mergifyio backport 1.61.3

PR #7059, opened 2025-03-20
Inject federation definitions when building test schemas

tninesling:
The `build_subgraph` utility takes in a partial schema and expands it to include definitions for `@link` as well as federation directives. Note that this utility is still private to the blueprint tests, but it will soon be ready to use widely in composition tests. After collecting links metadata for the schema, it now applies the blueprint's `on_directive_definition_and_schema_parsed` hook, which ultimately uses the federation spec's `add_elements_to_schema` functionality.

JS source for blueprint hook in [federation.ts](https://github.com/apollographql/federation/blob/3566777b82cc7a5e011a096764f8ee2884524658/internals-js/src/federation.ts#L2113-L2240)

Some notes on the blueprint:
As we go along, we'll probably find that the code in the blueprint is either misnamed or obsolete. This PR also removes the base trait for `SchemaBlueprint` because the only nontrivial implementation in the JS base class is handling `graphql-js` error messages which won't apply in Rust. This code now uses `FederationBlueprint` directly, which removes some unnecessary self references previously needed for dispatch. Additionally, the hook names sometimes refer to the "parsing" that was done to convert from `graphql-js` ASTs into usable object formats, which is done for us by `apollo-compiler`. I have left the names as-is for now, but I have created a follow-up ticket for us to make these more declarative later on.

---

**Checklist**

- Tests added and passing[^3]
    - [X] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-federation/src/link/federation_spec_definition.rs
+++ b/apollo-federation/src/link/federation_spec_definition.rs
@@ -42,6 +42,8 @@ pub(crate) const FEDERATION_SHAREABLE_DIRECTIVE_NAME_IN_SPEC: Name = name!("shar
 pub(crate) const FEDERATION_OVERRIDE_DIRECTIVE_NAME_IN_SPEC: Name = name!("override");
 pub(crate) const FEDERATION_CONTEXT_DIRECTIVE_NAME_IN_SPEC: Name = name!("context");
 pub(crate) const FEDERATION_FROM_CONTEXT_DIRECTIVE_NAME_IN_SPEC: Name = name!("fromContext");
+pub(crate) const FEDERATION_COMPOSEDIRECTIVE_DIRECTIVE_NAME_IN_SPEC: Name =
+    name!("composeDirective");
 
 pub(crate) const FEDERATION_FIELDS_ARGUMENT_NAME: Name = name!("fields");
 pub(crate) const FEDERATION_RESOLVABLE_ARGUMENT_NAME: Name = name!("resolvable");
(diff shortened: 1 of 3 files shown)


apollo-federation/src/link/federation_spec_definition.rs
```diff
@@ -539,6 +553,178 @@ impl FederationSpecDefinition {
...
+            FEDERATION_SHAREABLE_DIRECTIVE_NAME_IN_SPEC,
+            &[],
+            self.version().gt(&Version { major: 2, minor: 1 }), // repeatable in Fed 2.2 or later
+            &[DirectiveLocation::FieldDefinition],
+            false,
+            None,
+        )
+    }
+
+    fn override_directive_specification() -> DirectiveSpecification {
```
duckki, line 655: This seems to be the pre-2.7 version of override. It seems worth commenting about that.

apollo-federation/src/link/federation_spec_definition.rs
```diff
@@ -539,6 +553,178 @@ impl FederationSpecDefinition {
...
+            &[
+                DirectiveArgumentSpecification {
+                    base_spec: ArgumentSpecification {
+                        name: FEDERATION_FROM_ARGUMENT_NAME,
+                        get_type: |_| Ok(ty!(String!)),
+                        default_value: None,
+                    },
+                    composition_strategy: None,
+                },
+                DirectiveArgumentSpecification {
```
duckki, line 667: No `label` argument for this version.

apollo-federation/src/link/federation_spec_definition.rs
```diff
@@ -539,6 +553,178 @@ impl FederationSpecDefinition {
...
+                    base_spec: ArgumentSpecification {
+                        name: FEDERATION_OVERRIDE_LABEL_ARGUMENT_NAME,
+                        get_type: |_| Ok(ty!(String!)),
```
duckki, line 698: The `label` argument should be nullable.

apollo-federation/src/link/federation_spec_definition.rs
```diff
@@ -539,6 +553,178 @@ impl FederationSpecDefinition {
...
+        )
+    }
+
+    fn compose_directive_directive_specification() -> DirectiveSpecification {
+        DirectiveSpecification::new(
+            FEDERATION_COMPOSEDIRECTIVE_DIRECTIVE_NAME_IN_SPEC,
+            &[DirectiveArgumentSpecification {
+                base_spec: ArgumentSpecification {
+                    name: FEDERATION_NAME_ARGUMENT_NAME,
+                    get_type: |_| Ok(ty!(String!)),
```
duckki, line 717: This argument seems to be nullable.

duckki (changes requested):
Besides a few comments and corrections, I'm excited to have this :)

apollo-federation/src/link/federation_spec_definition.rs
```diff
@@ -539,6 +553,178 @@ impl FederationSpecDefinition {
...
+            false,
+            None,
+        )
+    }
+
+    fn shareable_directive_specification(&self) -> DirectiveSpecification {
+        DirectiveSpecification::new(
+            FEDERATION_SHAREABLE_DIRECTIVE_NAME_IN_SPEC,
+            &[],
+            self.version().gt(&Version { major: 2, minor: 1 }), // repeatable in Fed 2.2 or later
```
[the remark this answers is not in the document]
tninesling: JS code uses >= 2.2, but this is using > 2.1. Left the comment to clarify that at first glance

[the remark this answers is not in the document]
duckki: Oh, I see. IMO, `>=` is easier to understand when it comes to versions.

apollo-federation/src/link/federation_spec_definition.rs
tninesling: From what I can tell, name should be non-null in [composeDirective](https://www.apollographql.com/docs/graphos/reference/federation/directives#composedirective)

duckki: Indeed. Then, [the JS code](https://github.com/apollographql/federation/blob/3566777b82cc7a5e011a096764f8ee2884524658/internals-js/src/specs/federationSpec.ts#L156) might be a bug.

conwuegb (commented):
LGTM

apollo-federation/src/schema/blueprint.rs
```diff
@@ -182,6 +106,52 @@ impl SchemaBlueprint for FederationBlueprint {
     fn apply_directives_after_parsing() -> bool {
         todo!()
     }
+
+    fn remove_federation_definitions_broken_in_known_ways(
```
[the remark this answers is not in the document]
tninesling: I brought the comments over in the latest commit

apollo-federation/src/schema/blueprint.rs
```diff
@@ -182,6 +106,52 @@ impl SchemaBlueprint for FederationBlueprint {
...
+            FEDERATION_REQUIRES_DIRECTIVE_NAME_IN_SPEC,
+        ] {
+            if let Some(pos) = schema.get_directive_definition(directive_name) {
+                let directive = pos.get(schema.schema())?;
+                if directive.arguments.is_empty()
+                    || (directive.arguments.len() == 1
+                        && directive
+                            .argument_by_name(&FEDERATION_FIELDS_ARGUMENT_NAME)
+                            .is_some_and(|fields| {
+                                fields.ty.inner_named_type() == "String"
```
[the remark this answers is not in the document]
tninesling: Updated this to use `ty!` because it captures both the nullability and list/non-list aspect of the check. I think these shouldn't be used anywhere but here, and it's a bit easier to grok with them inlined. Definitely agree on constants for shared values though.

apollo-federation/src/schema/blueprint.rs
```diff
@@ -182,6 +106,52 @@ impl SchemaBlueprint for FederationBlueprint {
...
+        Ok(())
+    }
+
+    fn expand_known_features(schema: &mut FederationSchema) -> Result<(), FederationError> {
+        let Some(links_metadata) = schema.metadata() else {
+            return Ok(());
+        };
+
+        for link in links_metadata.links.clone() {
+            if link.url.identity == Identity::join_identity() {
```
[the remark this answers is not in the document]
tninesling: I missed that it explicitly skipped join spec and was using that as an example of how to add it. We'll need a way to to pick out known features, so I've updated this to be an empty block in the loop. The router-side of things has a static registry for plugins, so we could consider a similar pattern for feature registration.

apollo-federation/src/link/federation_spec_definition.rs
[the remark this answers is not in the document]
tninesling: I was looking for `.gte` and couldn't find it 🤦 Will update to `.ge`

apollo-federation/src/link/federation_spec_definition.rs
tninesling: Fixed in latest commit

PR #7060, opened 2025-03-20
Add more flexible subgraph mocks, take 2

SimonSapin:
This PR supersedes https://github.com/apollographql/router/pull/7020.

I suggest reviewing by commit since the “_Import GraphQL execution engine_” commit is a large copy-paste from apollo-compiler.

---

When testing the Router, spinning up subgraph servers can be more involved than we’d like. `apollo_router::test_harness::MockedSubgraphs` has long allowed tests to run a modified router that intercepts subgraph requests and respond with fixed JSON data. It is configured with a map of expected requests to corresponding responses. For example:

```rust
// Don’t do this going forward! See below

let subgraphs = MockedSubgraphs([
    (
        "subgraph_name",
        MockSubgraph::builder()
        .with_json(
            serde_json::json!({
                "query": "query($representations:[_Any!]!){_entities(representations:$representations){...on Contact{__typename country}}}",
                "variables": {
                    "representations": [
                        {
                            "__typename":"Contact",
                            "id":"0",
                            "displayName": "Max",
                        }
                    ]
                }
            }),
            serde_json::json!({"data": {
                "_entities": [{
                    "__typename":"Contact",
                    "country": "Fr"
                }]
            }})
        ).with_json(
            serde_json::json!({
                "query": "query($representations:[_Any!]!){_entities(representations:$representations){...on Contact{country}}}",
                "variables": {
                    "representations": [
                        {
                            "__typename":"Contact",
                            "id":"0",
                            "displayName": "Max",
                        }
                    ]
                }
            }),
            serde_json::json!({"data": {
                "_entities": [{
                    "country": "Fr"
                }]
            }})
        ).build(),
    )
].collect());

let service = TestHarness::builder()
    .configuration_json(serde_json::json!({
        "include_subgraph_errors": { "all": true },
    }))
    .unwrap()
    .schema(SOME_TESTING_SCHEMA)
    .extra_plugin(subgraphs)
    .build_supergraph()
    .await
    .unwrap();
```

The map keys contain the `query` strings of subgraph requests generated by the query planner, which can be hard to predict.

Instead, the comparatively newer `mock_subgraphs` is a built-in plugin based on a full-featured GraphQL executor. There is no need configure expected requests as the executor should be able to handle any request. Only the data for responses need to be configured, and a request’s selection set can select any subset of it.

```rust
// Recommended:

let subgraphs = serde_json::json!({
    "subgraph_name": {
        "entities": [
            {
                "__typename": "Contact",
                "id": "0",
                "displayName": "Max",
                "country": "Fr",
            },
        ],
    },
});

let service = TestHarness::builder()
    .configuration_json(serde_json::json!({
        "include_subgraph_errors": { "all": true },
        "experimental_mock_subgraphs": subgraphs,
    }))
    .unwrap()
    .schema(SOME_TESTING_SCHEMA)
    .build_supergraph()
    .await
    .unwrap();
```

Compared to before, note the additional `configuration_json` line and the lack of `extra_plugin`.

The full plugin configuration simplifies to:

```rust
type MockSubgraphPluginConfig = Map<SubgraphName, ConfigForOneSubgraph>;
type SubgraphName = String;

struct ConfigForOneSubgraph {
    /// Entities that can be queried through Federation’s special `_entities` field
    entities: Vec<JsonMap>,
    /// Data for `query` operations (excluding the special `_entities` field)
    query: JsonMap,
    /// Data for `mutation` operations
    mutation: Option<JsonMap>,
    /// HTTP headers for the subgraph response
    headers: Map<String, String>,
}
```

As of this writing, this plugin is only intended for the Router’s own tests but is always available. It is excluded from the JSON Schema for Router configuration (see `HIDDEN_FROM_CONFIG_JSON_SCHEMA`) and prefixed `experimental_`, but nothing actually prevents its use externally.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/plugins/cache/snapshots/apollo_router__plugins__cache__tests__no_cache_control-4.snap
```diff
@@ -5,5 +5,15 @@ expression: response
...
+  "errors": [
+    {
+      "message": "subgraph mock not configured",
+      "path": [],
+      "extensions": {
+        "code": "SUBGRAPH_MOCK_NOT_CONFIGURED",
+        "service": "user"
+      }
+    }
+  ]
```
IvanGoncharov, line 18: @SimonSapin Is it intentional?
From the code, it looks like snapshots should stay the same 🤔

IvanGoncharov (changes requested):

apollo-router/src/plugins/cache/snapshots/apollo_router__plugins__cache__tests__no_cache_control-4.snap
SimonSapin: Yes, this new error comes from the commit [TestHarness subgraph intercept: proper plugin that returns errors](https://github.com/apollographql/router/pull/7060/commits/f6de468a23aa0ed0b953116721f9ff78c7599a1a). Before that, `TestHarness::subgraph_network_requests == false` would cause a subgraph hook to respond with an empty JSON object

IvanGoncharov: In this case, I don't understand why we have an error originating in the mocking code inside the snapshot of cache tests.
Is it an issue with the test that had this problem but before this change making it didn't trigger any errors?
And you planning to fix those tests in the follow up PR?
Or am I totally missing context here, and is it normal for this error to be inside a snapshot?

IvanGoncharov (approved):
It appears to be a good iterative improvement on our current mocking system 👍 

My only question is about mocking related errors within snapshots. 
However, this is not a blocking concern since it is an intentional change.

SimonSapin:
@mergify backport 1.x

PR #7070, opened 2025-03-21
Add support of ignored_headers for subscription deduplication

bnjjj:
Add support for `ignored_headers` for subscription deduplication which is a way to ignore some specific headers in deduplication because for example if you include some transaction ID in headers or so , then you can’t benefit from subscription dedup even if it doesn't change the data returned by susbscription.

Here is an example of configuration:

```yaml
subscription:
  enabled: true
  deduplication:
    enabled: true # optional, default: true
    ignored_headers: # (optional) List of ignored headers when deduplicating subscriptions
    - x-transaction-id
    - custom_header_name
```

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/plugins/subscription.rs
+++ b/apollo-router/src/plugins/subscription.rs
@@ -68,13 +68,33 @@ pub(crate) struct SubscriptionConfig {
     pub(crate) enabled: bool,
     /// Select a subscription mode (callback or passthrough)
     pub(crate) mode: SubscriptionModeConfig,
-    /// Enable the deduplication of subscription (for example if we detect the exact same request to subgraph we won't open a new websocket to the subgraph in passthrough mode)
-    /// (default: true)
-    pub(crate) enable_deduplication: bool,
+    /// Configure the deduplication of subscriptions (for example if we detect the exact same request to subgraph we won't open a new websocket to the subgraph in passthrough mode)
+    pub(crate) deduplication: DeduplicationConfig,
     /// This is a limit to only have maximum X opened subscriptions at the same time. By default if it's not set there is no limit.
     pub(crate) max_opened_subscriptions: Option<usize>,
     /// It represent the capacity of the in memory queue to know how many events we can keep in a buffer
     pub(crate) queue_capacity: Option<usize>,
+}
+
+/// Deduplication configuration
+#[derive(Debug, Clone, Deserialize, Serialize, JsonSchema)]
+#[serde(deny_unknown_fields, default)]
+pub(crate) struct DeduplicationConfig {
+    /// Enable the deduplication of subscriptions (for example if we detect the exact same request to subgraph we won't open a new websocket to the subgraph in passthrough mode)
+    /// (default: true)
+    pub(crate) enabled: bool,
+    /// List of headers to ignore for deduplication, for example if you forward 'user-agent' header but doesn't interfere with data returned by the subscription we can just ignore it
+    /// to not create a new subscription connection
+    pub(crate) ignored_headers: HashSet<String>,
+}
+
+impl Default for DeduplicationConfig {
+    fn default() -> Self {
+        Self {
+            enabled: true,
+            ignored_headers: Default::default(),
+        }
+    }
 }
 
 impl Default for SubscriptionConfig {
@@ -82,7 +102,7 @@ impl Default for SubscriptionConfig {
         Self {
             enabled: true,
             mode: Default::default(),
-            enable_deduplication: true,
+            deduplication: DeduplicationConfig::default(),
             max_opened_subscriptions: None,
             queue_capacity: None,
         }
@@ -1426,7 +1446,7 @@ mod tests {
         .unwrap();
 
         assert!(sub_config.enabled);
-        assert!(sub_config.enable_deduplication);
+        assert!(sub_config.deduplication.enabled);
         assert!(sub_config.max_opened_subscriptions.is_none());
         assert!(sub_config.queue_capacity.is_none());
     }
--- a/apollo-router/src/services/subgraph.rs
+++ b/apollo-router/src/services/subgraph.rs
@@ -1,5 +1,6 @@
 #![allow(missing_docs)] // FIXME
 
+use std::collections::HashSet;
 use std::fmt::Display;
 use std::pin::Pin;
 use std::sync::Arc;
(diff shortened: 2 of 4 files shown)


apollo-router/src/configuration/migrations/0038-ignored-headers-subs-dedup.yaml
```diff
@@ -0,0 +1,5 @@
+description: "`subscription.enable_deduplication` was a true/false before but now has child properties `enabled` and `ignored_headers`"
```
goto-bus-stop, line 1: Can we still do this? I thought we now prevent starting the router with non-migrated config, so this would be a breaking change https://github.com/apollographql/router/pull/6777

goto-bus-stop (changes requested):

apollo-router/src/configuration/migrations/0038-ignored-headers-subs-dedup.yaml
bnjjj: Waiting for https://github.com/apollographql/router/pull/7162 to be merged before

bnjjj:
Waiting for https://github.com/apollographql/router/pull/7162

apollo-router/src/plugins/subscription.rs
```diff
@@ -68,21 +68,41 @@ pub(crate) struct SubscriptionConfig {
...
 
+/// Deduplication configuration
+#[derive(Debug, Clone, Deserialize, Serialize, JsonSchema)]
+#[serde(deny_unknown_fields, default)]
+pub(crate) struct DeduplicationConfig {
+    /// Enable the deduplication of subscriptions (for example if we detect the exact same request to subgraph we won't open a new websocket to the subgraph in passthrough mode)
+    /// (default: true)
+    pub(crate) enabled: bool,
+    /// List of headers to ignore for deduplication, for example if you forward 'user-agent' header but doesn't interfere with data returned by the subscription we can just ignore it
+    /// to not create a new subscription connection
```
goto-bus-stop, line 87:
```suggestion
    /// List of headers to ignore for deduplication. Even if these headers are different, the subscription request is considered identical.
    /// For example, if you forward the "User-Agent" header, but the subgraph doesn't depend on the value of that header,
    /// adding it to this list will let the router dedupe subgraph subscriptions even if the header value is different.
```

apollo-router/src/services/subgraph.rs
```diff
@@ -422,15 +426,71 @@ impl Request {
         }
         for (var_name, var_value) in &body.variables {
             hasher.update(var_name.inner());
-            // TODO implement to_bytes() for value in serde_json_bytes
-            hasher.update(var_value.to_string().as_bytes());
+            hasher.update(var_value.to_bytes());
         }
         for (name, val) in &body.extensions {
             hasher.update(name.inner());
             // TODO implement to_bytes() for value in serde_json_bytes
```
goto-bus-stop, line 433:
```suggestion
```

docs/source/routing/operations/subscriptions/index.mdx
```diff
@@ -341,6 +341,22 @@ The router considers subscription operations **identical** if all of the followi
 - The operations sent to the subgraph have identical GraphQL selection sets (i.e., requested fields).
 - The operations provide identical values for all headers that the router sends to the subgraph.
```
goto-bus-stop, line 342:
```suggestion
- The operations provide identical values for all headers that the router sends to the subgraph, except those listed in `ignored_headers`.
```

docs/source/routing/operations/subscriptions/index.mdx
```diff
@@ -341,6 +341,22 @@ The router considers subscription operations **identical** if all of the followi
...
 
+### Configuring deduplication
+
+You can ignore specific headers when doing the subscription deduplication. If you propagate `user-agent` header to subgraphs and you still want to deduplicate subscriptions and so ignore this header
+in the deduplication algorithm you can do it by adding the following to your router's YAML config file:
```
goto-bus-stop, line 347:
```suggestion
You can ignore specific headers when deduplicating subscriptions.

If you propagate a header with varying values to your subgraph (for example, `User-Agent`), but you still want to deduplicate subscriptions and you can accept that only one out of many values will be propagated, you can configure `ignored_headers` like this:
```

I'm not sure this is the best way to phrase it, but I think it'd be good to explicitly state that you lose some detail by doing this. Maybe @apollographql/docs can help or follow up with an improvement PR?

apollo-router/src/configuration/testdata/migrations/minor/health_check.yaml
```diff
@@ -1,2 +0,0 @@
-health-check:
```
goto-bus-stop, line 1: I'm a bit confused by this file 🤔  It doesn't look like this change is related to the PR?

goto-bus-stop (commented):
The implementation LGTM, I have mostly some copy edit suggestions

I do wonder about this feature though. If you enable this for a particular header, you lose the detail that header provides. eg., if the first client out of 10 000 identical requests is IE 11, while the other 9 999 are Chrome, then your User-Agent data is going to be poor quality. The same applies maybe even more strongly for transaction IDs. If it doesn't matter to a user that those values are low quality, does it even make much sense to propagate them at all? (the answer might be yes if low quality data is better than no quality data 😇 )

e; after reading some of the customer requests for this feature, it looks like they *have* made this consideration already and do prefer low quality over no quality, so that's probably okay!

goto-bus-stop (approved):
Since i pretty much answered my concern by reading the associated ticket ( 💀 ), i'm happy to approve it, just needs a test fix 🙈

apollo-router/src/configuration/testdata/migrations/minor/health_check.yaml
bnjjj: It's mainly because I changed the test because now I have a minor migration so it changed

bnjjj:
@goto-bus-stop yes basically we let our users to decide if it's worth it or not. And in this case it would only affect subscriptions, not other queries. I also thought about a mechanism using `vary` header but it doesn't work because the client would be able to modify these headers and we don't want to let clients/users interacting with this because they could be malicious and try to deduplicate a specific subscription to get events they're not allowed. So I think this is the right strategy.

goto-bus-stop:
Looking at the CI failure...maybe there's a tabs/spaces mixup in a config somewhere?

chandrikas:
Thanks for this @bnjjj - we actually had a roadmap item around this later in Q4! However, I see that we are only addressing deduplication for subscriptions, deduplicating for queries is still a big use case. Is that a fast follow or do we have other concerns there?

bnjjj:
Thanks @chandrikas the configuration will be the same but the implementation is a little bit different but can be done as a follow up.

PR #7071, opened 2025-03-21
fix JWT expiry logging bug

Velfi:
I introduced a bug when I changed the JWT expiry sourcing code. This PR fixes that bug and adds tests.

Fixes #**issue_number**

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/plugins/authentication/jwks.rs
```diff
@@ -617,40 +617,44 @@ pub(super) fn decode_jwt(
...
-                tracing::error!("expected JWT claims to be an object");
-                None
+        .unwrap_or_else(|err| {
+            tracing::error!("could not read JWT claims: {err}");
+            None
+        })
+        .flatten()
+        .and_then(|claims_value: Option<serde_json::Value>| {
+            let claims_obj = claims_value.as_ref()?.as_object();
+            // Extract the expiry claim from the JWT, returning immediately if it's not present
```
goto-bus-stop, line 629: this seems a little untrue since we're not returning immediately, just propagating `None`, maybe we can just omit that?
```suggestion
            // Extract the expiry claim from the JWT
```

goto-bus-stop (approved):
LGTM!

bnjjj (approved):

PR #7084, opened 2025-03-21
Move public connectors arrow methods to their own separate files

andrewmcgivery:
- Move public connectors arrow methods to their own separate files
- Colocate tests for each method into their test file
- Split out some of the larger tests into singular more granular tests that test one thing

Worth noting for reviewers, this is almost completely a copy/paste. The method code itself is copy/pasted as is and the tests are mostly just refactored into separate test functions. That being said, the comments at the top of the method field definitely use a look for accuracy!

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [x] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-federation/src/sources/connect/json_selection/methods.rs
```diff
@@ -141,15 +141,15 @@ impl std::ops::Deref for ArrowMethod {
...
-            Self::JsonStringify => &public::JsonStringifyMethod,
+            Self::Echo => &public::echo::EchoMethod,
+            Self::Map => &public::map::MapMethod,
+            Self::Match => &public::r#match::MatchMethod,
+            Self::First => &public::first::FirstMethod,
+            Self::Last => &public::last::LastMethod,
+            Self::Slice => &public::slice::SliceMethod,
+            Self::Size => &public::size::SizeMethod,
+            Self::Entries => &public::entries::EntriesMethod,
+            Self::JsonStringify => &public::json_stringify::JsonStringifyMethod,
```
benjamn, line 152: Can we possibly keep the structure of the `public::` module, so this code doesn't have to change?

andrewmcgivery: Possibly? But it would likely make `public/mod.rs` kind of ugly and annoying to maintain. I want sort of optimizing to make that file as simple as possible. Is the concern here just the verbosity/repetitiveness of these lines or am I missing something else here? 😅 

@dylan-apollo and I discussed the exporting a bit cause I had started with an uglier syntax before landing on this so I'm curious if they have any thoughts!

dylan-apollo: Ah yeah, my recommendation basically came from a desire to _use_ the module name rather than repeat it. I think we decided that switching that style (and updating the macro) would be better in a follow-up rather than in this PR.

Basically best practice ([this Clippy lint turns on in pedantic mode](https://rust-lang.github.io/rust-clippy/master/index.html#module_name_repetitions)) would be to say `echo::shape` and `echo::method` rather than `echo::echo_method` and `echo::echo_shape`. I _think_ if the modules each have a consistently named function, we won't need to generate the wrapper struct at _all_ and use a much simpler macro helper.

So with the current code, yes re-exporting would be cleaner. But longer term I think we can simplify more instead.

benjamn: To answer the question about what to put in `public/mod.rs`, I think it would just be a list of
```rust
pub(crate) use echo::EchoMethod;
pub(crate) use entries::EntriesMethod;
// And so on....
```
without re-exporting the module names (so no `echo`, `entries`). I consider the individual module names an implementation detail, especially since the names are not 1:1 with actual method names (see `json_stringify.rs` and `->jsonStringify` for example). So unless we need the modules separately, I think we can keep the `mod.rs` file to the same number of lines as before?

And, sure, we could adjust the protocol that arrow method implementations expose, in the future. I don't have strong preferences, but I would say it's useful for `ArrowMethod::lookup` to have a single type of thing to return, hence he `ArrowMethodImpl` trait that associates the `apply` and `shape` methods together for each method. I would hope that any strategy we adopt here is another implementation detail that we can change freely behind the scenes.

Can we preserve the previous module structure unless there's an actively better way to structure the code?

andrewmcgivery: Updated!

benjamn (approved):
Looks good so far! Feel free to include the `future` methods in this PR or a follow-up.

andrewmcgivery:
> Looks good so far! Feel free to include the `future` methods in this PR or a follow-up.

Thank you! Will attack the `future` methods in my next PR. :)

Will wait for checks to pass and then will merge!

PR #7088, opened 2025-03-24
Rename metric attribute `state` to `http.connection.state`

BrynCooke:
In this PR we rename the `state` attribute on `apollo.router.open_connections` to `http.connection.state`

This enables us to use the [Otel convention] (https://opentelemetry.io/docs/specs/semconv/attributes-registry/http/#http-connection-state)

Technically this attribute is on a custom instrument, so there is no requirement to call it the same as the convention, but it's probably better from the user perspective to reuse a well known attribute.

Note that we don't support `idle` as a value yet, but we should add this in future.

<--ROUTER-1129-->

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/.changesets/feat_bryn_connection_metrics_2_0.md
+++ b/.changesets/feat_bryn_connection_metrics_2_0.md
@@ -7,8 +7,8 @@ To help users to diagnose when connections are keeping pipelines hanging around 
     - `config.hash` - The hash of the configuration.
     - `server.address` - The address that the router is listening on.
     - `server.port` - The port that the router is listening on if not a unix socket.
-    - `state` - Either `active` or `terminating`.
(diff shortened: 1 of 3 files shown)


BrynCooke:
@mergifyio backport 2.1.0

goto-bus-stop (approved):
`terminating` isn't a value in the spec, but the spec does allow custom values if `idle` or `active` don't apply, which seems true enough here

BrynCooke:
Yeah I read that `terminating` is OK for us to have.

PR #7094, opened 2025-03-24
Restore error accumulation in connectors parsing

dylan-apollo:
Uses `MultiTry` to get back to accumulating as many errors as possible during connectors parsing instead of exiting when the first error is encountered.

Also removes an unused dependency I noticed while adding the new one.


lennyburdette (approved):
very cool! thanks for the comments, that helps a lot

PR #7095, opened 2025-03-24
Connection shutdown timeout 2.x

BrynCooke:
Forward port of #7058 

This PR introduces a configurable timeout from signal to terminate to actual termination defaulted to 60 seconds. After the timeout is reached the connection is forcibly terminated.

```
supergraph:
  connection_shutdown_timeout: 12s
```

```
supergraph:
  early_cancel: true
```

```
traffic_shaping:
  router:
    timeout: 10s
```

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/axum_factory/listeners.rs
```diff
@@ -201,6 +202,50 @@ pub(super) async fn get_extra_listeners(
...
+                // Only wait for the connection to close gracfully if we recieved a request.
+                // On hyper 0.x awaiting the connection would potentially hang forever if no request was recieved.
+                if received_first_request.load(Ordering::Relaxed) {
+                    // The connection may still not shutdown so we apply a timeout from the configuration
+                    // Connections stuck terminating will keep the pipeline and everything related to that pipeline
+                    // in memory.
+
+                    if let Err(_) = connection.timeout(connection_shutdown_timeout).await {
+                        tracing::warn!(
+                            timeout = connection_shutdown_timeout.as_millis() / 1000,
```
Velfi, line 234: Why not use `as_secs()` here?

Velfi (approved):

apollo-router/src/axum_factory/listeners.rs
BrynCooke: I should have divided by 1000.0f the purpose was to get a float. Will fix in the morning

PR #7096, opened 2025-03-24
minimal validations to support connect on type

lennyburdette:
---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-federation/src/sources/connect/validation/connect.rs
+++ b/apollo-federation/src/sources/connect/validation/connect.rs
@@ -40,7 +40,17 @@ pub(super) fn fields_seen_by_all_connects(
         .filter(|ty| !ty.is_built_in())
         .for_each(|extended_type| {
             if let ExtendedType::Object(node) = extended_type {
-                match fields_seen_by_object_connectors(
+                match fields_seen_by_connectors_on_types(
+                    extended_type,
+                    node,
+                    schema,
+                    all_source_names,
+                ) {
+                    Ok(fields) => seen_fields.extend(fields),
+                    Err(errs) => messages.extend(errs),
+                }
+
+                match fields_seen_by_connectors_on_fields(
                     extended_type,
                     node,
                     schema,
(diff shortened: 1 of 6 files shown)


apollo-federation/src/sources/connect/validation/connect.rs
```diff
@@ -94,6 +109,65 @@ fn fields_seen_by_object_connectors(
...
+            }
+        };
+        messages.extend(http.type_check(schema).err().into_iter().flatten());
+    }
+
+    if messages.is_empty() {
+        Ok(seen_fields)
+    } else {
+        Err(messages)
+    }
```
dylan-apollo, line 161: Is this all identical to the field version? Don't have to deal with it now, but when I extract this into `Connect::parse` and then `Connect::type_check`, can that be shared between the locations or am I missing a subtle difference?

apollo-federation/src/sources/connect/validation/connect.rs
```diff
@@ -66,20 +76,25 @@ pub(crate) enum ObjectCategory {
     Other,
 }
 
-/// Make sure that any `@connect` directives on object fields are valid
-fn fields_seen_by_object_connectors(
+/// Make sure that any `@connect` directives on types are valid
+fn fields_seen_by_connectors_on_types(
```
dylan-apollo, line 80: Do you have a follow-up to prevent this on root types?

dylan-apollo (approved):

apollo-federation/src/sources/connect/validation/connect.rs
lennyburdette: there will be differences around variable use and keys, but it's not impossible it could be abstracted to handle that. i'll take a note

apollo-federation/src/sources/connect/validation/connect.rs
lennyburdette: yeah, there are other tickets for further validations

PR #7101, opened 2025-03-24
federation/composition: added a `Link` argument to `ArgumentSpecification`'s `get_type` closure field

duckki:
The `link` argument is now used in `ArgumentSpecification`.

* Refactor: `get_type`'s return type is changed to use `FederationError`, instead of `SingleFederationError`, which simplifies code.
* ~This PR fails the `detects_federation_2_subgraphs_correctly` due to incomplete subgraph schema builder implementation, which is expected at this time. So, `should_panic` macro was added.~ (fixed)
* Updated `on_missing_directive_definition` and `on_directive_definition_and_schema_parsed` to keep tests working.
    - partially ported `addToSchema` and `addDefinitionsToSchema`
    - specialized `add_elements_to_schema` for `LinkSpecDefinition`
    - updated some directive/type references to use the "in-schema" names.


apollo-federation/src/schema/blueprint.rs
```diff
@@ -213,6 +213,9 @@ mod tests {
     }
 
     #[test]
+    #[should_panic(
```
tninesling, line 216: Is there a way to do this change while keeping this test intact? I'm not quite understanding why this fails in this way.

duckki: I think what it takes is to update `on_directive_definition_and_schema_parsed` to use `add_to_schema`, which I'm working on in the [other PR](https://github.com/apollographql/router/pull/7103). Let me see if I can fix the error with that.

duckki: I pushed a fix.

apollo-federation/src/schema/type_and_directive_specification.rs
```diff
@@ -440,44 +459,11 @@ impl DirectiveSpecification {
...
-                                let arg_strategy_strings: Vec<String> = arg_strategies_capture2
-                                    .iter()
-                                    .map(|(arg_name, strategy)| format!("{arg_name}: {}", strategy.name()))
-                                    .collect();
-                                format!("{{ {} }}", arg_strategy_strings.join(", "))
-                            }
-                        }),
-                    })
-                }));
+                argument_merger = Some(directive_argument_merger(
```
duckki, line 462: This is just a pure refactoring. `cargo fmt` was adding another indentation level, so I factored out the whole block.

apollo-federation/src/schema/type_and_directive_specification.rs
```diff
@@ -290,6 +305,10 @@ pub(crate) struct EnumTypeSpecification {
 }
 
 impl TypeAndDirectiveSpecification for EnumTypeSpecification {
+    fn name(&self) -> &Name {
```
duckki, line 308: Useful for debugging. JS also has this field.

tninesling (approved):
Nice! Glad to see those tests re-enabled

PR #7110, opened 2025-03-25
Remove `UnwindSafe` bound and panic `Result` from `compute_job::execute`

SimonSapin:
This PR is expected to not change any run-time behavior, but makes it easier to use the “compute” thread pool:

* The return type no longer includes a `Result` related to the callback having panicked. Instead `resume_unwind` is used to propagate any panic to the caller. This is not a behavior change because every single caller was using `.expect()` on that `Result`, and because we also install a panic hook that exits the process so unwinding never starts this whole code path is unreachable.

* The callback no longer needs to implement `UnwindSafe`. With the above change, `compute_job::execute(job)` is effectively the same as `job()` regarding unwind safety.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/compute_job.rs
+++ b/apollo-router/src/compute_job.rs
@@ -1,6 +1,5 @@
 use std::future::Future;
 use std::ops::ControlFlow;
-use std::panic::UnwindSafe;
 use std::sync::OnceLock;
 
 use opentelemetry::metrics::MeterProvider as _;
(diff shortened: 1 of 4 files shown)


lrlna (approved):
🙏 🙏

PR #7111, opened 2025-03-25
debug query planning span with creation of child spans

lrlna:
This commit introduces inner, child spans inside the existing query planning span. This lets us differentiate between different processes in this part of the codebase, which is especially needed to debug this 1.58.1 release.

The current query planning span looks like this:
```
query myQuery
│  parse_query
└─── supergraph
│    └─── query_planning
└───execute
```

With this change, the query planning span can have three distinct possible children depending on the codepath.

1. If the requested operation is already in the cache and has been planned, this is the span structure:

    ```
    query myQuery
    │  parse_query
    └─── supergraph
    │    └─── query_planning
    │         └─── cache_lookup
    └───execute
    ```
2. If the requested operation is not in the cache and we need to do actual query planning, this is the span structure:
    ```
    query myQuery
    │  parse_query
    └─── supergraph
    │    └─── query_planning
    │         │    cache_lookup
    │         │    caching_query_planner_wrapper
    │         └─── parallelism_pool
    │              └─── invoke_planner
    │                   └─── worker_pool
    │                        └─── plan
    └───execute
    ```
3. If the requested operation is not in the cache, **but a different connection is already planning it**, this is the span structure:

    ```
    query myQuery
    │  parse_query
    └─── supergraph
    │    └─── query_planning
    │         │    cache_lookup
    │         └─── waiting_for_cache
    └───execute
    ```


SimonSapin (approved):

BrynCooke (changes requested):
Let's make these new spans debug level and document how to turn them on.

PR #7114, opened 2025-03-25
Move `ObjectCategory` into `ConnectedElement`

dylan-apollo:
This is going to make a different refactor easier. Possible we could use this information in other ways as well, like maybe in place of some lookups that `ConnectorPosition` has to do dynamically? But I didn't want to go digging.


apollo-federation/src/sources/connect/validation/connect.rs
```diff
@@ -209,7 +203,7 @@ fn fields_seen_by_connector(
     let mut seen_fields = vec![(object.name.clone(), field.name.clone())];
 
     // direct recursion isn't allowed, like a connector on User.friends: [User]
-    if matches!(category, ObjectCategory::Other) && &object.name == field.ty.inner_named_type() {
+    if &object.name == field.ty.inner_named_type() {
```
dylan-apollo, line 206: I could put this back—the check just didn't seem to be needed, since you can't return `Query` or `Mutation` anyway... and if you could we'd still want to prevent circular references.

lennyburdette (approved):

Issue #7122, opened 2025-03-26
Entity caching subgraph overrides not working


FahimSiddiquiDev:
## Describe the Bug

Entity caching behaves inconsistently when both global TTL and per-subgraph TTLs are defined. Specifically:

- Entity caching works correctly when only global TTL is defined or when only per-subgraph TTL is defined.
- However, when both are configured simultaneously, the global TTL always takes precedence, and subgraph-specific TTL values are ignored.

## Reference

According to the [Apollo documentation](https://www.apollographql.com/docs/graphos/routing/performance/caching/entity#configure-router-for-entity-caching), entity caching should respect subgraph-specific TTL values if provided:

> Configure entity caching per subgraph, with overrides per subgraph for disabling entity caching and TTL.

Router Configuration Example

```yaml
preview_entity_cache:
  enabled: true
  expose_keys_in_context: true
  subgraph:
    all:
      enabled: true
      redis:
        urls: ["redis://host.docker.internal:6379”]
        timeout: 2s
        ttl: 120s
    subgraphs:
      product-service:
        ttl: 60s  # Time-to-live for cached entities from product-service
      inventory-service:
        ttl: 90s  # Time-to-live for cached entities from inventory-service
```
Note: The global TTL can be either higher or lower than the individual subgraph TTLs.

## Steps to Reproduce

1. Configure the router with the above YAML configuration.
2. Trigger caching for entities from product-service and inventory-service.
3. Observe the cache expiration times.

## Expected Behavior

The TTL values for each subgraph should take precedence over the global TTL:

- Entities from product-service should have a 60s TTL.
- Entities from inventory-service should have a 90s TTL.

## Actual Behavior

- All entities are cached for 120s (the global TTL), ignoring the subgraph-specific values.
- Cache respones (checked a few seconds later):
```bash 
127.0.0.1:6379> TTL version:1.0:subgraph:product-service:type:Query:hash:8ab7bc4e9de4c14e35426675620bb984c40a3b992c4395c7815d43816ed23a32:data:faa1c49292e3ee1b183ec6e23ec2340be08d8cbef64cf90f656910a83ecfdbf4
(integer) 117
[...]

PR #7123, opened 2025-03-26
fix(entity_cache): compute TTL and cache enabled properly regarding the subgraph

bnjjj:
Related to #7122 

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/plugins/cache/entity.rs
```diff
@@ -518,6 +511,29 @@ impl EntityCache {
...
     }
+
+    // Returns boolean to know if cache is enabled for this subgraph
+    fn subgraph_enabled(&self, subgraph_name: &str) -> bool {
+        self.enabled
+            // If subgraphs.all is set but there's no override for the subgraph
+            && ((self.subgraphs.all.enabled.unwrap_or(true) // Because enabled is true by default
+                && matches!(self.subgraphs.get(subgraph_name).enabled, Some(true) | None))
+                // It overrides subgraphs.all
+                || (self.subgraphs.all.enabled == Some(false) && matches!(self.subgraphs.get(subgraph_name).enabled, Some(true))))
```
SimonSapin, line 522: Sorry but I find it hard to follow the logic here. What do you think of reorganizing this into a `match`? I _think_ this is equivalent:

```rust
if !self.enabled {
    return false
}
match (
    self.subgraphs.all.enabled,
    self.subgraphs.get(subgraph_name).enabled,
) {
    (_, Some(x)) => x, // explicit per-subgraph setting overrides the `all` default
    (Some(true) | None, None) => true, // unset defaults to true
    (Some(false), None) => false,
}

File: apollo-router/src/plugins/cache/entity.rs [hunk 2]
```diff
@@ -1765,3 +1781,158 @@ impl Ord for CacheKeyStatus {
...
+        let map = [
+            (
+                "user".to_string(),
+                Subgraph {
+                    redis: None,
+                    private_id: Some("sub".to_string()),
+                    enabled: None,
+                    ttl: None,
+                    ..Default::default()
+                },
```
SimonSapin, line 1806: “Real” users never construct struct literals like this. Since these two test is all about configuration, I think they should instead parse configuration from a YAML string or a `json!` macro in order to exercise things like `#[serde(default)]`

apollo-router/src/plugins/cache/entity.rs
bnjjj: I agree with you, I planned to come back to this once the tests were good but I forgot

apollo-router/src/plugins/cache/entity.rs
bnjjj: Done, thanks for the review

SimonSapin (approved):

FahimSiddiquiDev:
@bnjjj  could we please merge the PR since approval is in place. I would like to verify whether it fixes https://github.com/apollographql/router/issues/7122. Many thanks 👍

bnjjj:
@FahimSiddiquiDev done. Let me know if it fixes everything but I'm not sure

bnjjj:
@FahimSiddiquiDev Were you able to try ?

Issue #6542, opened 2025-01-14
OTEL bottleneck, ability to decide programmatically decide which spans we want to create


samuelAndalon:
## Describe the solution you'd like
Apollo router provides the capability of high throughput, with very few instances the router can handle very large workloads, however, we have identified a bottleneck in [Open Telemetry and the batch processor](https://www.apollographql.com/docs/graphos/reference/router/telemetry/trace-exporters/datadog#batch_processor)

the `query_planning` span wraps the whole query_planning services, which includes cache lookups, those takes microseconds, and as a result the router is exporting hundreds of thousands of spans causing the OpenTelemetry channel to not being able to send all spans that are actually created (bummer that OTEL does not implement backpressure)

<img width="763" alt="image" src="https://github.com/user-attachments/assets/76b6e09a-d13b-422f-864c-e47c9c1a6b3d" />

for our use case, these spans are completely useless, they don't give us any useful insight and they just consume memory and CPU of the node where our router instance is running, and memory and CPU of the datadog agent that sends the spans to the datadog api/

Its not only `query_planning` there are other spans that are useless for our use cases:
1. execution
2. http_request
3. fetch
4. parallel
5. secuence

pretty much all query plan nodes.

## Describe alternatives you've considered
We have tried changing the log level of modules using the `RUST_LOG` env variable, which works for most of use cases, but it breaks the propagation of headers when changing the log level of `apollo_router::services::http::service`.

After doing some debugging found that the issue is here:

https://github.com/apollographql/router/blob/dev/apollo-router/src/services/http/service.rs#L276

the propagation of headers relies on the log level of modules. Which IMHO should be decoupled.

BrynCooke:
I think it would be possible to fix the propagation even if the http span is suppressed.

PR #7124, opened 2025-03-26
Tracing: add child spans to the existing `query_planning` span

lrlna:
This is #7111, but for `dev`, as 7111 won't land in an existing release. 

There are several processes happening before an operation is being sent to the query planner for planning. The existing `query_planning` span encapsulates all of it without a detailed breakdown:
```
query myQuery
│  parse_query
└─── supergraph
│    └─── query_planning
└───execute
```

With this change, we are adding more details to `query_planning` span to have a fuller story in the traces around query planning. There are now three possible children spans: `cache_lookup`, `process_and_plan` and `waiting_for_cache`.

**Cache Lookup**. The following span structure is expected if the requested operation is already in the cache and has been planned:

```
query myQuery
│  parse_query
└─── supergraph
│    └─── query_planning
│         └─── cache_lookup
└───execute
```
**Process and Plan**. The following span structure is expected if the requested operation is not in the cache, and we need to do planning:
```
query myQuery
│  parse_query
└─── supergraph
│    └─── query_planning
│         │    cache_lookup
│         └─── process_and_plan
│              └─── worker_pool
│                   └─── plan
└───execute
```
**Waiting for cache**. The following span structure is expected if the requested operation is not in the cache, **but a different connection is already planning it**, making the current connection wait for that result to be written to cache before using it:

```
query myQuery
│  parse_query
└─── supergraph
│    └─── query_planning
│         │    cache_lookup
│         └─── waiting_for_cache
└───execute
```

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [x] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/.changesets/lrlna_query_planner_router_spans.md
+++ b/.changesets/lrlna_query_planner_router_spans.md
@@ -1,0 +1,49 @@
+### Tracing: add child spans to the existing `query_planning` span ([PR #7123](https://github.com/apollographql/router/pull/7123)) 
+
+There are several processes happening before an operation is being sent to the query planner for planning. The existing `query_planning` span encapsulates all of it without a detailed breakdown:
+```
+query myQuery
+│  parse_query
+└─── supergraph
+│    └─── query_planning
+└───execute
+```
+
+
+With this change, we are adding more details to `query_planning` span to have a fuller story in the traces around query planning. There are now three possible children spans: `cache_lookup`, `process_and_plan` and `waiting_for_cache`.
+
+**Cache Lookup**. The following span structure is expected if the requested operation is already in the cache and has been planned:
+
+    ```
+    query myQuery
+    │  parse_query
+    └─── supergraph
+    │    └─── query_planning
+    │         └─── cache_lookup
+    └───execute
+    ```
+**Process and Plan**. The following span structure is expected if the requested operation is not in the cache, and we need to do planning:
+    ```
+    query myQuery
+    │  parse_query
+    └─── supergraph
+    │    └─── query_planning
+    │         │    cache_lookup
+    │         └─── process_and_plan
+    │              └─── worker_pool
+    │                   └─── plan
+    └───execute
+    ```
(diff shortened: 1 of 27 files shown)


lrlna:
> **@lrlna, please consider creating a changeset entry in /.changesets/. [These instructions](https://github.com/apollographql/router/blob/HEAD/.changesets/README.md) describe the process and tooling.**
The bot is incorrect, my changes came with a changeset >.>

BrynCooke (changes requested):
As discussed in the huddle, let's make all the new spans debug, and then provide documentation on how to turn these on.

We may want to consider using a custom trace target rather than module name, for instance `query_planning_internals` which would mean that we can refactor the code without the target changing in future.

See https://docs.rs/tracing/latest/tracing/#configuring-attributes

lrlna:
> and then provide documentation on how to turn these on.

i can't find this in our docs, actually. how do we expect users to turn on debug _tracing_? do we have it set up the same way as debug _logging_? If so, I am not sure if it's going to be useful in debugging prod, since then users would be expected to turn on debug logging, which nobody is going to do 🤔

samuelAndalon:
please allow me to chime in in this PR, I have a couple of thoughts.

on router instances with very high throughput Otel spans could represent a memory bottleneck specially during export, while I think an span for pure QP creation is absolutely needed, (specially if contains operation name as resource name), an span for cache lookup will exponentially increase the amount of spans that will be exported

A while ago I created https://github.com/apollographql/router/issues/6542 mentioning that query planning spans with cache lookups are tacking microseconds and our router instances are exporting hundreds of thousands of those.


It would be great if we can decide programmatically which spans we want through the router config, for now our solution is to use the `RUST_LOG=apollo_router::query_planner::execution=error` env variable

lrlna:
Closing this since we've made better changes in #7236 when fixing #7184.

PR #7125, opened 2025-03-26
Revert: chore: Delegate setting "latest" to RELEASE_CHECKLIST.md, rather than CI.

abernix:
This reverts the change introduced in:

- https://github.com/apollographql/router/pull/6809

This will also need to be backported to 1.x to revert the similarly applied
https://github.com/apollographql/router/pull/6811.

It turns out that this change, while precautionarily a great idea, is not
necessary! This is only noted very very discreetly in this part of the
easily missed [GitHub documentation].

[GitHub documentation]: https://docs.github.com/en/repositories/releasing-projects-on-github/managing-releases-in-a-repository#about-release-management:~:text=Optionally%2C%20select%20Set%20as%20latest%20release.%20If%20you%20do%20not%20select%20this%20option%2C%20the%20latest%20release%20label%20will%20automatically%20be%20assigned%20based%20on%20semantic%20versioning.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -970,10 +970,17 @@ jobs:
                   cd artifacts && sha1sum *.tar.gz > sha1sums.txt
             - run:
                 name: Create GitHub Release
-                # We always create the version as a "pre-release" version.
-                # It is promoted to latest in a separate step in RELEASE_CHECKLIST.md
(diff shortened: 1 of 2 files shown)


abernix:
@mergifyio backport 1.x

PR #7128, opened 2025-03-26
Create parse & type check phases for `@connect`

dylan-apollo:
A few things in here:

1. Unifies the way we validate a single `@connect` whether on type or field
2. Carries the parse/validate structure up another level
3. Creates a new `ResolvedField` struct to add some semantics to seen fields. We'll want to use that in more places as refactors continue

### DIFF
--- a/apollo-federation/src/sources/connect/validation/connect.rs
+++ b/apollo-federation/src/sources/connect/validation/connect.rs
@@ -60,6 +60,78 @@ pub(super) fn fields_seen_by_all_connects(
     }
 }
 
+/// A parsed `@connect` directive
+struct Connect<'schema> {
+    selection: Selection<'schema>,
+    http: Http<'schema>,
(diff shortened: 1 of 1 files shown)


lennyburdette (approved):
Would it be accurate to say that “parse” is for structure, and “type check” is for semantics?

dylan-apollo:
> Would it be accurate to say that “parse” is for structure, and “type check” is for semantics?

Only somewhat—because, for example, checking that `source:` matches the virtual enum of `@source`s is part of this phase. I really am trying to divide between the bare minimum required to run and all the other stuff—so that we can share the first phase between runtime and validation.

So maybe the phases are like:
1. Connectors can be run
2. Preventing runtime errors

Though phase 2 _also_ includes preventing later composition errors which would have unclear messages (e.g., satisfiability) 😵‍💫

PR #7134, opened 2025-03-27
Fix flakey connectors test

dylan-apollo:
Connectors assertions weren't properly handling parallel requests, which can technically happen in any order.


lennyburdette (approved):
very clever!

PR #7137, opened 2025-03-27
Root type renaming and validation

tninesling:
During validation, the Federation blueprint checks if there's a custom type for any of the root operation types. If so, that type is renamed to the default name for that corresponding operation (ie. `MyCustomQuery` is renamed to `Query`). If there is some other non-root type which already has the default operation name, the validation produces an error.

The code ported validating and renaming corresponds to [this block in federation.ts](https://github.com/apollographql/federation/blob/main/internals-js/src/federation.ts#L1585-L1618).

The corresponding tests came from [this section of subgraphValidations.ts](https://github.com/apollographql/federation/blob/main/internals-js/src/__tests__/subgraphValidation.test.ts#L590-L656).

BONUS: This also includes an implementation for renaming a type at a given position, since it did not already exist. This tries to directly mutate any objects holding the name to be replaced as well as update all maps that may use the type as a key.

---

**Checklist**

- Tests added and passing[^3]
    - [X] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-federation/src/schema/blueprint.rs
```diff
@@ -416,18 +469,191 @@ mod tests {
...
+            }
+
+            type Query {
+                g: Int
+            }
+            "#,
+            "test.graphqls",
+        )
+        .expect("parses schema");
+        let error = build_subgraph("S", &schema, true).expect_err("fails validation");
```
conwuegb, line 491: I think the messages passed to [`expect()`](https://doc.rust-lang.org/std/result/enum.Result.html#method.expect) and [`expect_err()`](https://doc.rust-lang.org/std/result/enum.Result.html#method.expect_err) could be clearer. The "Recommended Message Style" from the Rust docs suggests using the word "should" and describing the reason _why_ you expect the `Result` to be `Ok` or _why_ you expect an error.

```suggestion
        .expect("The schema should parse successfully");
        let error = build_subgraph("S", &schema, true).expect_err("The invalid schema should fail validation");
```

I'd say the code reads more intuitively this way. Thoughts?

tninesling: I think that makes sense if you're going to panic and propagate an error to an end-user, but I usually keep these short in tests because I actually care about the content of the `Err` that's being unwrapped. To me, it's more of a tag/line number indicator so I know which assertion is failing when I'm developing the test.

apollo-federation/src/schema/position.rs
```diff
@@ -1250,6 +1295,21 @@ impl ScalarTypeDefinitionPosition {
         };
         directive_referencers.scalar_types.shift_remove(self);
     }
+
+    fn rename(&self, schema: &mut FederationSchema, new_name: Name) -> Result<(), FederationError> {
```
duckki, line 1299: This seems macro worthy :)

tninesling: I realized I needed to implement these changes for the various referencers of each type/field, which vary in subtle but important ways for each type. I've tried to factor out some common stuff to DRY things out, but I think it's different enough that a macro may cause more problems than it's worth.

apollo-federation/src/subgraph/mod.rs
```diff
@@ -353,6 +355,24 @@ impl From<ValidFederationSubgraph> for ValidSubgraph {
     }
 }
 
+/// Currently, this is making up for the fact that we don't have an equivalent of `addSubgraphToErrors`.
+/// In JS, that manipulates the underlying `GraphQLError` message to prepend the subgraph name. In Rust,
+/// it's idiomatic to have strongly typed errors which defer conversion to strings via `thiserror`, so
+/// for now we wrap the underlying error until we figure out a longer-term replacement that accounts
+/// for missing error codes and the like.
+#[allow(dead_code)]
```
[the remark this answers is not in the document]
tninesling: Makes sense. I think this is a good enough home for now, since I don't expect it to be long-lived. I think it'll be totally replaced when we create a proper composition error representation, but I needed a wrapper so I could do a 100% matching port for the test assertions.

PR #7141, opened 2025-03-28
coprocessor: fix parsing of graphql responses with null as `data`

IvanGoncharov:
Coprocessor deserialized GraphQL response using serde_json directly that resulted in `data: null` were ignored.
After my change, it's using `Response::from_value`, which applies the same steps for parsing GraphQL response that we are already using for parsing subgraph response.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-1219]: https://apollographql.atlassian.net/browse/ROUTER-1219?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apollo-router/src/graphql/request.rs
```diff
@@ -211,20 +210,20 @@ impl Request {
...
-            let bytes = serde_json::to_vec(value)?;
+            let bytes = serde_json::to_vec(&value)?;
             result.push(Request::deserialize_from_bytes(&bytes.into())?);
         }
         Ok(result)
     }
 
-    fn process_query_values(value: &serde_json::Value) -> Result<Vec<Request>, serde_json::Error> {
-        let mut result = Request::allocate_result_array(value);
+    fn process_query_values(value: Value) -> Result<Vec<Request>, serde_json::Error> {
```
bnjjj, line 223: I'm curious to know why you changed this function to take ownership of value here instead of ref as previously ?

apollo-router/src/graphql/request.rs
```diff
@@ -239,42 +238,36 @@ impl Request {
...
-            .extensions(extensions);
-
-        let request = if let Some(query_str) = query {
-            request_builder.query(query_str).build()
-        } else {
-            request_builder.build()
-        };
-
+            .extensions(extensions)
+            .build();
```
bnjjj, line 270: Is this change related to the fix or just a refactor ?

apollo-router/src/graphql/mod.rs
```diff
@@ -125,41 +125,39 @@ impl Error {
...
+    pub(crate) fn from_value(value: Value) -> Result<Error, MalformedResponseError> {
+        let mut object = ensure_object!(value).map_err(|error| MalformedResponseError {
+            reason: format!("invalid error within `errors`: {}", error),
+        })?;
 
         let extensions =
             extract_key_value_from_object!(object, "extensions", Value::Object(o) => o)
-                .map_err(|err| FetchError::SubrequestMalformedResponse {
-                    service: service_name.to_string(),
+                .map_err(|err| MalformedResponseError {
```
bnjjj, line 135: Why do you use a different error ? We're loosing related service here

bnjjj (commented):
It contains a lot of changes unrelated to the fix itself (formatting fixes), I tried to read everything carefully to make sure I didn't miss anything but feel free to tell me in which files changes are really important to look at.

apollo-router/src/graphql/request.rs
IvanGoncharov: was part of serde_json => serde_json_bytes switch, but I will test if works with reference

IvanGoncharov:
> It contains a lot of changes unrelated to the fix itself (formatting fixes), I tried to read everything carefully to make sure I didn't miss anything but feel free to tell me in which files changes are really important to look at.

@bnjjj I will add comments on all important changes.

apollo-router/src/graphql/mod.rs
```diff
@@ -125,41 +125,39 @@ impl Error {
...
                 .unwrap_or_default();
-        let message = extract_key_value_from_object!(object, "message", Value::String(s) => s)
-            .map_err(|err| FetchError::SubrequestMalformedResponse {
-                service: service_name.to_string(),
+        let message = match extract_key_value_from_object!(object, "message", Value::String(s) => s)
+        {
+            Ok(Some(s)) => Ok(s.as_str().to_string()),
+            Ok(None) => Err(MalformedResponseError {
+                reason: "missing required `message` property within error".to_owned(),
+            }),
```
IvanGoncharov, line 144: This is a new behaviour for subgraph response parsing. 
Previously, if the description was missing, it was set up by `unwrap_or_default`, which resulted in an empty string.
Since the coprocessor used serde for parsion, it was erroring on missing `message`.
Since the GraphQL spec requires `message` I just added validation here.

.changesets/fix_fix_coprocessor_data_null.md
```diff
@@ -0,0 +1,9 @@
+### Fix Parsing of Coprocessor GraphQL Responses ([PR #7141](https://github.com/apollographql/router/pull/7141))
+
+Standardized GraphQL response parsing and validation between coprocessors and subgraphs to ensure consistent behavior throughout the Router.
```
BrynCooke, line 3: I think we need a user focused changset entry. Something like.

Previously coprocessors allowed null body in graphql responses, hoever this is not allowed in the graphql spec. The router now correctly enforces that the body of a coprocessor response is valid, and will reject such responses.

If your coprocessor relied on this behaviour then you must update your coprocessor to return a calid graphql response.

BrynCooke (changes requested):

apollo-router/src/graphql/response.rs
```diff
@@ -3,18 +3,25 @@ use std::time::Instant;
...
 use crate::json_ext::Object;
 use crate::json_ext::Path;
 use crate::json_ext::Value;
 
+#[derive(thiserror::Error, Display, Debug, Eq, PartialEq)]
+#[error("GraphQL response was malformed: {reason}")]
+pub(crate) struct MalformedResponseError {
+    /// The reason the deserialization failed.
+    pub(crate) reason: String,
+}
```
IvanGoncharov, line 23: Previously, we reported all errors related to parsing/validation graphql response as `FetchError::SubrequestMalformedResponse` which made this code only usable for subgraph responses. Since now we have generic error it could be reused for coprocessor responses.

apollo-router/src/plugins/coprocessor/mod.rs
```diff
@@ -1137,29 +1138,28 @@ where
         let code = control.get_http_status()?;
 
         let res = {
-            let graphql_response: crate::graphql::Response =
-                match co_processor_output.body.unwrap_or(serde_json::Value::Null) {
-                    serde_json::Value::String(s) => crate::graphql::Response::builder()
```
IvanGoncharov, line 1142: This part is weird, I think it would be better to just have custom error saying you need to provide `body` if you return Break. Instead we just push null instead of `body` that result in parsing error.
But I decided to stop myself here, since I already went to deep into refactoring this code.

.changesets/fix_fix_coprocessor_data_null.md
IvanGoncharov: Make sense 👍

apollo-router/src/graphql/mod.rs
bnjjj: Is it potentially a breaking change or could it break existing implementation ? I think it's a bug fix but as it's more strict maybe it could be interesting to add notes about this in the changelog to avoid confusions

BrynCooke (approved):

IvanGoncharov:
@mergify backport 1.x

PR #7142, opened 2025-03-28
Tracing deadlock investigation

carodewig:
I've been trying to look into a deadlock on `tracing-subscriber::registry::Extensions`. The function `extract_dd_trace_id` is a relatively good candidate for the lock trigger due to the fact that it's looking at the root span extensions, rather than those of the current span.

For more information about when reentrant locking may cause a deadlock see: https://doc.rust-lang.org/std/sync/struct.RwLock.html

```
let _rg1 = lock.read();  |
                         |  // will block
                         |  let _wg = lock.write();
// may deadlock          |
let _rg2 = lock.read();  |
```

To test this we ran an instrumented version of the router that would log out if re-entrant locking of span extensions was detected.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


carodewig:
NB: this comment is interesting (and perhaps relevant) - locking multiple spans at a time could cause deadlocks. https://github.com/tokio-rs/tracing/issues/639#issuecomment-749683552

I reorganized the code in 6b4c178 to remove simultaneously-held locks in a few functions, including `on_new_span` which appeared in some of the deadlocks.

BrynCooke:
This looks great!

BrynCooke:
I did eventually manage to run the router against an instrumented version of tracing subscriber. This checks that an RWLock never tried to access the same lock from the same thread. This would immediately fail on the dev branch, but on this branch it doesn't error.

Nice work!

BrynCooke:
@carodewig To get this merged I think we need a changelog.

carodewig:
> @carodewig To get this merged I think we need a changelog.

@BrynCooke Done (a7e231e), thank you!

apollo-router/src/plugins/telemetry/formatters/json.rs
```diff
@@ -357,7 +357,7 @@ fn extract_dd_trace_id<'a, 'b, T: LookupSpan<'a>>(span: &SpanRef<'a, T>) -> Opti
     if let Some(root_span) = root.next() {
         let ext = root_span.extensions();
         // Extract dd_trace_id, this could be in otel data or log attributes
-        if let Some(otel_data) = root_span.extensions().get::<OtelData>() {
+        if let Some(otel_data) = ext.get::<OtelData>() {
```
bnjjj, line 360: ouch. Good catch !

bnjjj (approved):

PR #7143, opened 2025-03-28
[docs] Update in-memory cache warm up

smyrick:
Adding this info to an appropriate place so it is not hidden in just the federation docs
: https://github.com/apollographql/federation/pull/3226/files

### DIFF
--- a/docs/source/routing/performance/caching/in-memory.mdx
+++ b/docs/source/routing/performance/caching/in-memory.mdx
@@ -80,16 +80,18 @@ then look at `apollo.router.schema.loading.time` and `apollo.router.query_planni
 
 If the router is using distributed caching for query plans, the warm-up phase will also store the new query plans in Redis. Since all Router instances might have the same distributions of queries in their in-memory cache, the list of queries is shuffled before warm-up, so each Router instance can plan queries in a different order and share their results through the cache.
 
+<MinVersion version="1.61.0">
+
 #### Cache warm-up with headers
 
-With [the Apollo Router Core](https://github.com/apollographql/router) v1.61.0+  or v2.x+, you can pass the following header to return the query plans in the GraphQL response extensions:
+</MinVersion>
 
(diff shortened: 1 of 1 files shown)


docs/source/routing/performance/caching/in-memory.mdx
```diff
@@ -80,6 +80,16 @@ then look at `apollo.router.schema.loading.time` and `apollo.router.query_planni
...
+#### Cache warm-up with headers
+
+With [the Apollo Router Core](https://github.com/apollographql/router) v1.61.0+  or v2.x+, you can pass the following header to return the query plans in the GraphQL response extensions:
+
+- The header `Apollo-Expose-Query-Plan` must be set with one of the following options:
+  - A value of `true` returns a human-readable string and JSON blob of the query plan
+  - A value of `dry-run` will generate the query plan and then abort execution. This can be helpful if you want to warm up any internal or external query plan caches as after generation it will save it to your configured cache locations as normal.
+
+This will allow you to warm up cache instances with real, mirrored, or similar to production operations that you direct to the Router before you move traffic to new instances.
+
```
Meschreiber, line 92: Just some suggestions for conciseness and clarity:

```suggestion
<MinVersion version="1.61.0">

#### Cache warm-up with headers

</MinVersion>

With router v1.61.0+  and v2.x+, you can pass the `Apollo-Expose-Query-Plan` header to return query plans in the GraphQL response extensions. You must set the header to one of the following values:

- `true`: Returns a human-readable string and JSON blob of the query plan.
- `dry-run`: Generates the query plan without executing the query.  After generation, query plans are saved to your configured cache locations, making it useful for pre-warming internal or external caches.

This lets you warm up query plan caches with real, mirrored, or similar to production operations before transitioning traffic to new router instances.
```

Meschreiber (approved):

theJC:
To be able to do this, doesnt the experimental.expose_query_plan need to be explicitly need to be enabled?  If so, perhaps adding this detail would be helpful here.

PR #7144, opened 2025-03-28
Simplify `->` method tail handling, ensuring left associativity

benjamn:
This PR removes the `tail: &WithRange<PathList>` parameter from the `ArrowMethodImpl::apply` trait method that each arrow method must implement, which is a significant simplification because arrow method implementations are now responsible for returning only their immediate value, not also invoking the rest of the path (the `tail`) to return a final value recursively.

For reference, the `tail` of an arrow method invocation is `input->method(arg).THIS.PART`, so it's a little strange that the implementation of each `->method` was previously responsible for continuing to evaluate its own tail. Instead, the runtime code that invokes methods will now call `tail.apply_to_path` automatically on behalf of any method that returns a non-`None` value, enforcing the same protocol for every method. Theoretically, this slightly reduces the power of arrow methods, since they can no longer evaluate their own tails in exotic ways. We do not have any use cases for that today, and arrow methods are not developer-extensible, so this doesn't seem like much of a loss.

As part of these changes, I changed the handling of `PathList::Key` items when mapped over input arrays. Previously, if `array.key` had a `->method` after it (so `array.key->method`), the `->method` would be applied to each individual `key` value, rather than receiving the entire array as a whole. After this PR, path evaluation is reliably _left associative_, so everything preceding the `->` will be evaluated before invoking the method, and the method will be invoked only once, rather than participating in the mapping of `array.key`. This is generally more predictable and convenient, since (for example) it allows you to guarantee a final string by slapping `->jsonStringify` at the end of any path, without having to worry it may be mapped over individual values and recomposed into a larger (non-string) array.

It may be possible to gate these changes so they work the old way(s) in `connect/v0.1`, but it would be ugly. For now, I'm waiting on the team to decide if we need to preserve old behavior for a change like this (where the old behavior was buggy).

### DIFF
--- a/apollo-federation/src/sources/connect/json_selection/apply_to.rs
+++ b/apollo-federation/src/sources/connect/json_selection/apply_to.rs
@@ -540,42 +540,58 @@ impl ApplyToInternal for WithRange<PathList> {
                 }
             }
             PathList::Key(key, tail) => {
-                if let JSON::Array(array) = data {
-                    return self.apply_to_array(array, vars, input_path);
-                }
-
                 let input_path_with_key = input_path.append(key.to_json());
 
-                if !matches!(data, JSON::Object(_)) {
-                    return (
-                        None,
-                        vec![ApplyToError::new(
-                            format!(
-                                "Property {} not found in {}",
-                                key.dotted(),
-                                json_type_name(data),
-                            ),
-                            input_path_with_key.to_vec(),
-                            key.range(),
-                        )],
-                    );
-                }
+                if let JSON::Array(array) = data {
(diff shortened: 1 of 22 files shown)


lennyburdette:
To get this out in a patch release, we'll need to merge it to a branch cut from https://github.com/apollographql/router/tree/v2.1.1 (which doesn't exist yet). It is difficult to isolate this change from the other PR?

lennyburdette:
great! i'll make a branch we can both target (i have to backport a different fix too) towards the end of the week

lennyburdette (approved):

PR #7147, opened 2025-03-28
docs: Apollo-Expose-Query-Plan header

shorgi:
Update `Apollo-Expose-Query-Plan` header doc


Meschreiber (commented):
@shorgi , do you think this is a better or just additional spot for this info? (also in https://github.com/apollographql/router/pull/7143)

dariuszkuc:
Looks like this was updated in the federations docs https://github.com/apollographql/federation/pull/3226 instead of the router docs. Can you update this PR to match the federation one (as it is more accurate)?

dariuszkuc (approved):

PR #7151, opened 2025-03-31
add tests for entity key validations and batch connectors

lennyburdette:
also changes a doc link that points to an anchor that will probably change

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-federation/src/sources/connect/validation/snapshots/validation_tests@keys_and_entities__invalid__mismatch_batch.graphql.snap
```diff
@@ -0,0 +1,14 @@
+---
+source: apollo-federation/src/sources/connect/validation/mod.rs
+expression: "format!(\"{:#?}\", result.errors)"
+input_file: apollo-federation/src/sources/connect/validation/test_data/keys_and_entities/invalid/mismatch_batch.graphql
+---
+[
+    Message {
+        code: InvalidBody,
+        message: "In `@connect(http: {body:})` on `Product`: `$batch` doesn't have a field named `id`",
```
dylan-apollo, line 9: Not for this PR, but in the future we should probably come up with a standard way to refer to the type as well for `$batch` and `$this`.

apollo-federation/src/sources/connect/validation/snapshots/validation_tests@keys_and_entities__invalid__mismatch_composite_key_batch.graphql.snap
```diff
@@ -0,0 +1,14 @@
+---
+source: apollo-federation/src/sources/connect/validation/mod.rs
+expression: "format!(\"{:#?}\", result.errors)"
+input_file: apollo-federation/src/sources/connect/validation/test_data/keys_and_entities/invalid/mismatch_composite_key_batch.graphql
+---
+[
+    Message {
+        code: InvalidBody,
+        message: "In `@connect(http: {body:})` on `Product`: `Country` doesn't have a field named `id`",
```
dylan-apollo, line 9: And here we don't refer to the problematic expression, which is the opposite problem. Actually I think this will be easier when precomputing inputs, because we could get the list of expressions that build a given input type (i.e., `$batch`) and then _while_ resolving the actual type we should have both bits of info.

apollo-federation/src/sources/connect/validation/snapshots/validation_tests@keys_and_entities__invalid__multiple_keys_not_all_resolved_batch.graphql.snap
```diff
@@ -0,0 +1,14 @@
+---
+source: apollo-federation/src/sources/connect/validation/mod.rs
+expression: "format!(\"{:#?}\", result.errors)"
+input_file: apollo-federation/src/sources/connect/validation/test_data/keys_and_entities/invalid/multiple_keys_not_all_resolved_batch.graphql
+---
+[
+    Message {
+        code: MissingEntityConnector,
+        message: "Entity resolution for `@key(fields: \"id store { id country { key_id region } }\")` on `Product` is not implemented by a connector. See https://go.apollo.dev/connectors/directives/#rules-for-entity-true",
```
dylan-apollo, line 9: Looks like this link needs updating still.

apollo-federation/src/sources/connect/validation/test_data/keys_and_entities/valid/entity_connector_matches_non_resolvable_key_batch.graphql
```diff
@@ -0,0 +1,13 @@
+extend schema
+  @link(url: "https://specs.apollo.dev/federation/v2.11", import: ["@key"])
+  @link(url: "https://specs.apollo.dev/connect/v0.2", import: ["@connect"])
+
+type Product
+  @key(fields: "id", resolvable: false)
+  @connect(
+    http: { POST: "http://127.0.0.1:8000/v1/products/", body: "$batch.id" }
+    selection: "id name"
```
dylan-apollo, line 9: Should the test here be with a connector that _doesn't_ resolve the key?

lennyburdette: i'm thinking we'll validate against that for now, and i'm planning to add that validation rule when i have the batch shape to work with.

we can relax the validation when we implement `assumedOrdered: true` or whatever we end up calling it

apollo-federation/src/sources/connect/validation/snapshots/validation_tests@keys_and_entities__invalid__mismatch_batch.graphql.snap
lennyburdette: Yeah, I think we'll revisit a bunch of this when we move all validation to be shape-based.

dylan-apollo (approved):

PR #7152, opened 2025-03-31
Add support for connector header propagation via YAML config

andrewmcgivery:
Add support for connector header propagation via YAML config.

Added support for connector header propagation via YAML config. All of the existing header propagation in the Router now works for connectors by using
`headers.connector.all` to apply rules to all connectors or `headers.connector.sources.*` to apply rules to specific sources.

Note that if one of these rules conflicts with a header set in your schema, either in `@connect` or `@source`, the value in the Router config will take priority and be treated as an override.

```
headers:
  connector:
    all:
    request:
        - insert:
            name: "x-inserted-header"
            value: "hello world!"
        - propagate:
            named: "x-client-header"
    sources:
        request:
          - insert:
              name: "x-inserted-header"
              value: "hello world!"
          - propagate:
              named: "x-client-header"
```

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


.changesets/feat_feature_connectorsyamlheaders.md
```diff
@@ -0,0 +1,29 @@
...
+`headers.connector.all` to apply rules to all connectors or `headers.connector.sources.*` to apply rules to specific sources.
+
+Note that if one of these rules conflicts with a header set in your schema, either in `@connect` or `@source`, the value in your Router config will
+take priority and be treated as an override.
+
+```
+headers:
+  connector:
+    all:
+    request:
```
lennyburdette, line 13:
```suggestion
      request:
```

apollo-router/src/plugins/headers/mod.rs
```diff
@@ -179,6 +182,18 @@ enum Propagate {
...
 }
 
+#[derive(Clone, JsonSchema, Default, Deserialize)]
+#[serde(rename_all = "snake_case", deny_unknown_fields, default)]
+struct ConnectorHeadersConfiguration {
+    /// Map of subgraph_name.connector_source_name to configuration
+    #[serde(default)]
+    sources: HashMap<String, HeadersLocation>,
+
+    /// Options applying to all sources
```
lennyburdette, line 192:
```suggestion
    /// Options applying to all sources across all subgraphs
```

apollo-router/src/plugins/headers/mod.rs
```diff
@@ -307,159 +350,234 @@ where
...
+            }
+            Operation::Remove(remove) => remove.process_header_rules(headers_mut),
+            Operation::Propagate(propagate) => {
+                propagate.process_header_rules(already_propagated, supergraph_headers, headers_mut)
+            }
+        }
+    }
+}
+
+impl Insert {
```
lennyburdette, line 438: am i reading it correctly that `insert` doesn't look at `RESERVED_HEADERS`, so it's possible to insert `Host` and `Content-Type`?

lennyburdette (approved):
very nice!

apollo-router/src/plugins/headers/mod.rs
andrewmcgivery: It would appear that was the existing logic for subgraphs, yes. 😅 

Do we think this is a problem or are we okay to proceed?

lennyburdette: i just wanted to make sure that users can set host and content-type because we've allowed those with the directive API. we're ok!

PR #7164, opened 2025-04-02
feat: Add allow/deny lists and redaction to subgraph errors

BrynCooke:
`include_subgraph_errors` with new allow and deny list configuration that's compatible with [existing feature](https://www.apollographql.com/docs/graphos/routing/observability/subgraph-error-inclusion). 
Details were discussed in https://github.com/apollographql/router/pull/6402 and #6862

Reasons we need the capability:

1. Some subgraphs need to pass non-vulnerable data to client via error extensions. So a simple boolean wouldn't work in this case.
2. Some clients depend on certain data from errors object. Although they shouldn't do it, migrating away from it would take time for existing subgraphs, especially for native apps. While waiting for their migration, this PR would allow the original error message and extension keys to be redacted or propagated per subgraph configuration.

### Example:
```yaml
include_subgraph_errors:
  all:
    redact_message: true
    allow_extensions_keys:
      - code
  subgraphs:
    product:
      redact_message: false  # Propagate original error messages
      allow_extensions_keys: # Extend global allow list - `code` and `reason` will be propagated
        - reason
      exclude_global_keys:   # Exclude `code` from global allow list - only `reason` will be propagated.
        - code
    account:
      deny_extensions_keys:  # Overrides global allow list
        - classification
    review: false            # Redact everything.

    # Undefined subgraphs inherits default global settings from `all`
``` 

For reviewers:

The way that this works is that error redaction has been moved to the supergraph service. 
At the subgraph and connectors services the subgraph name is attached to errors.
At the supergraph response for each error we extract the subgraph and redact accordingly.

One thing I am not sure about is the `exclude_global_keys` config option. This is could be either `deny_extension_keys` or `allow_extension_keys`.

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


BrynCooke:
Deferred responses aren't working still

.changesets/config_subgraph_include_errors allow_and_deny_list.md
```diff
@@ -0,0 +1,29 @@
+### `include_subgraph_errors` fine grained control ([Issue #6402](https://github.com/apollographql/router/pull/6402)
+
+Update `include_subgraph_errors` with additional configuration options for both global and subgraph levels. This update provides finer control over error messages and extension keys for each subgraph. 
+For more details, please read the [subgraph-error-inclusion](https://www.apollographql.com/docs/apollo-server/observability/subgraph-error-inclusion/).
```
Velfi, line 4:
```suggestion
For more details, please read [subgraph-error-inclusion](https://www.apollographql.com/docs/apollo-server/observability/subgraph-error-inclusion/).
```

apollo-router/src/plugins/include_subgraph_errors/config.rs
```diff
@@ -0,0 +1,227 @@
...
+use serde::de::Deserializer;
+use serde::de::MapAccess;
+use serde::de::Visitor;
+use serde::de::{self};
+
+/// Configuration for exposing errors that originate from subgraphs
+#[derive(Clone, Debug, JsonSchema, Default, Deserialize)]
+#[serde(rename_all = "snake_case", deny_unknown_fields, default)]
+pub(crate) struct Config {
+    /// Configuration for all subgraphs, and handles subgraph errors
```
Velfi, line 16: This sentence reads weird with the `, and handles subgraph errors` at the end. Could you make this more explicit?

apollo-router/src/plugins/include_subgraph_errors/config.rs
```diff
@@ -0,0 +1,227 @@
...
+    pub(crate) all: ErrorMode,
+
+    /// Override default configuration for specific subgraphs
+    pub(crate) subgraphs: HashMap<String, SubgraphConfig>,
+}
+
+#[derive(Clone, Debug, JsonSchema, Serialize)]
+#[serde(untagged)]
+pub(crate) enum ErrorMode {
+    /// Propagate original error or redact everything
```
Velfi, line 26:
```suggestion
    /// When `true`, Propagate the original error as is. Otherwise, redact it.
```

apollo-router/src/plugins/include_subgraph_errors/config.rs
```diff
@@ -0,0 +1,227 @@
...
+        /// Allow specific extension keys
+        allow_extensions_keys: Vec<String>,
+        /// redact errors messages for all subgraphs
+        redact_message: bool,
+    },
+    /// Deny specific extension keys with required redact_message
+    Deny {
+        /// Deny specific extension keys
+        deny_extensions_keys: Vec<String>,
+        /// redact errors messages for all subgraphs
```
Velfi, line 39:
```suggestion
        /// redact error messages for all subgraphs
        redact_message: bool,
    },
    /// Deny specific extension keys with required redact_message
    Deny {
        /// Deny specific extension keys
        deny_extensions_keys: Vec<String>,
        /// redact error messages for all subgraphs
```

apollo-router/src/plugins/include_subgraph_errors/mod.rs
```diff
@@ -0,0 +1,169 @@
...
+                .unwrap_or(&config.default);
+
+            if !effective_config.include_errors {
+                tracing::debug!(
+                    "Redacting errors for subgraph '{}' based on config: include_errors=false",
+                    subgraph_name
+                );
+                // Redact fully if errors should not be included
+                error.message = REDACTED_ERROR_MESSAGE.to_string();
+                error.extensions = Object::new(); // Clear all extensions
```
Velfi, line 119: Do we make it clear in the docs that redaction of errors includes all error extensions?

apollo-router/src/plugins/include_subgraph_errors/tests.rs
```diff
@@ -0,0 +1,507 @@
...
+) -> Result<PluginTestHarness<IncludeSubgraphErrors>, BoxError> {
+    let mut config = Map::new();
+    config.insert("include_subgraph_errors".to_string(), plugin_config.clone());
+    let config = serde_yaml::to_string(&config).expect("config to yaml");
+    PluginTestHarness::builder().config(&config).build().await
+}
+
+async fn run_test_case(
+    config: &Value,
+    mock_responses: &[&str], // Raw bytes representing the *original* supergraph response
```
Velfi, line 41: Technically, if these are "raw bytes" then shouldn't the type be `&[u8]`?

apollo-router/src/services/fetch.rs
```diff
@@ -18,7 +18,7 @@ use crate::query_planner::fetch::Variables;
 use crate::query_planner::subscription::SubscriptionHandle;
 use crate::query_planner::subscription::SubscriptionNode;
 
-const FETCH_SUBGRAPH_NAME_EXTENSION_KEY: &str = "fetch_subgraph_name";
+const SUBGRAPH_NAME_EXTENSION_KEY: &str = "apollo.subgraph.name";
```
Velfi, line 21: Is this a breaking change? How do we know no user is doing anything with this?

.changesets/config_subgraph_include_errors allow_and_deny_list.md
abernix, line 4: The link is to Apollo Server right now, but [this](https://www.apollographql.com/docs/graphos/routing/observability/subgraph-error-inclusion) looks like the link you want:

```suggestion
For more details, please read [subgraph error inclusion](https://www.apollographql.com/docs/graphos/routing/observability/subgraph-error-inclusion).
```

apollo-router/src/services/fetch.rs
BrynCooke: @pubmodmatt WDYT? The reason I renamed this was to make clear that it is an Apollo thing. I don't believe it is meant for general consumption.

.changesets/config_subgraph_include_errors allow_and_deny_list.md
BrynCooke: Fixed in [0e287a1](https://github.com/apollographql/router/pull/7164/commits/0e287a1bf63cb62602432965f4745fc582208b8e)

apollo-router/src/plugins/include_subgraph_errors/tests.rs
BrynCooke: Fixed in [b4456d8](https://github.com/apollographql/router/pull/7164/commits/b4456d80e3053413a6ef10d3a58bac87c3b86f72)

apollo-router/src/plugins/include_subgraph_errors/mod.rs
BrynCooke: [e549e03](https://github.com/apollographql/router/pull/7164/commits/e549e038c362c07786f35c6739b7c96446571e1b)

apollo-router/src/services/fetch.rs
pubmodmatt: It's just added and immediately removed as a mechanism to propagate data up the stack, so the name is arbitrary.

BrynCooke: I spoke with @pubmodmatt and this is an internal name that was introduced for connectors. I think it's OK to rename it as it is generally new and also undocumented.
I'm going to add an explicit doc saying that it may change in the future.

BrynCooke: I'm also going to stick private in the name just to make sure.

apollo-router/src/plugins/include_subgraph_errors/config.rs
BrynCooke: [e549e03](https://github.com/apollographql/router/pull/7164/commits/e549e038c362c07786f35c6739b7c96446571e1b)

apollo-router/src/services/fetch.rs
BrynCooke: [8e18cc5](https://github.com/apollographql/router/pull/7164/commits/8e18cc5139459892b0f630dd4bcfc408eddb4c84)

PR #7165, opened 2025-04-02
implement ->joinNotNull method

lennyburdette:
---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-federation/src/sources/connect/json_selection/methods/public/join.rs
```diff
@@ -0,0 +1,288 @@
...
+            ));
+            &EMPTY_STRING
+        });
+
+    fn to_string(value: &JSON) -> String {
+        match value {
+            JSON::Bool(b) => b.then_some("true").unwrap_or("false").to_string(),
+            JSON::Number(number) => number.to_string(),
+            JSON::String(byte_string) => byte_string.as_str().to_string(),
+            JSON::Null | JSON::Array(_) | JSON::Object(_) => "".to_string(),
```
dylan-apollo, line 54: I think we should append to `warnings`, so it's clear why this didn't work. Otherwise, users get an empty string with no indication of what happened.

We may also want to return `None` like other methods so processing stops?

apollo-federation/src/sources/connect/json_selection/methods/public/join.rs
```diff
@@ -0,0 +1,288 @@
...
+        .unwrap_or_else(|| {
+            warnings.push(ApplyToError::new(
+                format!(
+                    "Method ->{} requires a string argument",
+                    method_name.as_ref()
+                ),
+                input_path.to_vec(),
+                method_name.range(),
+            ));
+            &EMPTY_STRING
```
dylan-apollo, line 46: Should this return `None` at this point instead of continuing with an empty string? That's what happens when most methods fail, I think, so processing doesn't continue down the line with an unexpected value.

apollo-federation/src/sources/connect/json_selection/methods/public/join.rs
```diff
@@ -0,0 +1,288 @@
...
+
+    use super::*;
+    use crate::selection;
+
+    #[rstest::rstest]
+    #[case(json!(["a","b","c"]), ", ", json!("a, b, c"))]
+    #[case(json!([1, 2, 3]), "|", json!("1|2|3"))]
+    #[case(json!([1.00000000000001, 2.9999999999999, 0.3]), "|", json!("1.00000000000001|2.9999999999999|0.3"))]
+    #[case(json!([true, false]), " and ", json!("true and false"))]
+    #[case(json!([null, "a", 1]), ", ", json!(", a, 1"))]
```
dylan-apollo, line 168: Seems logical for people to want to omit nulls—as I think is the case with the RFC 6570 stuff (gotta double check). Probably makes sense for this to require an explicit `->filter`, though, rather than make it an implicit behavior here.

apollo-federation/src/sources/connect/json_selection/methods/public/join.rs
```diff
@@ -0,0 +1,288 @@
...
+    #[case(json!([null, null]), ", ", json!(", "))] // two nulls should be joined with a comma
+    #[case(json!(1), ", ", json!("1"))]
+    #[case(json!("a"), ", ", json!("a"))]
+    #[case(json!(true), ", ", json!("true"))]
+    #[case(json!(null), ", ", json!(""))]
```
dylan-apollo, line 173: Need tests for objects, arrays of objects, and multi-dimensional arrays as well. Probably should wait until we decide on whether they produce errors and/or `None`, though.

apollo-federation/src/sources/connect/json_selection/methods/public/join.rs
benjamn: I was about to cite other programming languages as precedent for `join` defaulting to `,` for the separator string, but it turns out that's only JavaScript. Ruby and Perl default to an empty string (which makes more sense than `,` IMO), and many other languages require a separator.

Given that we probably don't want people struggling with the distinction between `->join` and `->join()`, I'd vote we require the argument and return `None` here to avoid leading anyone to think "Ah, I got a `String` back, so it must be the string I wanted."

apollo-federation/src/sources/connect/json_selection/methods/public/join.rs
benjamn: I disagree that `null` should be omitted by default, and strongly feel it should stringify to `null`.

apollo-federation/src/sources/connect/json_selection/methods/public/join.rs
```diff
@@ -0,0 +1,288 @@
...
+    if !(input_shape.is_unknown() || matches!(input_shape.case(), ShapeCase::Name(_, _))) {
+        let mismatches = input_shape_contract.validate(&input_shape);
+        if !mismatches.is_empty() {
+            return Shape::error(
+                format!(
+                    "Method ->{} requires an array of scalars values as input",
+                    method_name.as_ref()
+                ),
+                [],
+            );
```
benjamn, line 113: Can we give any idea of what the `mismatches` were with this error?

apollo-federation/src/sources/connect/json_selection/methods.rs
```diff
@@ -147,6 +148,7 @@ impl std::ops::Deref for ArrowMethod {
             Self::Size => &public::SizeMethod,
             Self::Entries => &public::EntriesMethod,
             Self::JsonStringify => &public::JsonStringifyMethod,
+            Self::Join => &public::JoinMethod,
```
benjamn, line 151: Please move this to the `::future` section until we've ironed out all the policy decisions involved in this method's design. There are several other future methods I feel better about than `->join`, so I don't want it skipping the line. It needs time to bake.

benjamn (changes requested):

apollo-federation/src/sources/connect/json_selection/methods/public/join.rs
andrewmcgivery: I also think filtering out nulls implicitly is a confusing behaviour. I'd much rather require them use `filter` if they want to remove them.

It would appear that most languages, including javascript, output an empty string for `null` values.

```
[null, 'hello', null, 'hello2'].join(',')
',hello,,hello2'
```

benjamn: Wow, you're right about JS (see [docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/join#description)). That behavior seems so arbitrary to me, but maybe there's a reason?

If the argument is that you sometimes need to do some conditional logic within a single array element expression that may result in an empty string, I would much rather developers actually use the empty string for that than `null` or `None`.

Since JS supports "sparse" arrays where not all elements are occupied, and JSON does not, I wonder if another reason is that the holes in a sparse array tend to get filled in with `null` when you JSON-encode the array, so it the designers of `Array.prototype.join` were trying to make sure those instances of `null` still behaved like `undefined` (which turns into an empty string). I don't think that applies to us though, since we're starting with JSON, so none of our arrays can be sparse (nor can they contain `None` elements, I believe). However, if you construct an array where any of the elements evaluates to `None` (due to an error, say), we fill that position with `null` to make sure the rest of the indices are not shifted. So maybe we do have a version of this problem.

Not sure where this leaves me. For now I've crossed out the ~strongly~ in my comment above.

apollo-federation/src/sources/connect/json_selection/methods.rs
lennyburdette: i'm not going to merge it until it's ready, but it will be public

apollo-federation/src/sources/connect/json_selection/methods/public/join.rs
lennyburdette: so one reason i haven't pushed on review or merge for this is that i really want to get validation working for input and argument shapes. then we'll have these scenarios for the input:

1. it's a scalar or list of scalars, passing validation and the runtime works as expected
2. it's an object or list of lists/objects, so it doesn't pass validation and we can't get to the runtime
3. it's an unknown or named input, so it passes validation and the runtime emits empty strings. i agree with Dylan that we should emit warnings — i'll add that.

I don't think we should have an implicit `array->toString` or `object->toString` at this point. we can add it later, but i'd rather emit nothing (with warnings) than something unexpected. Dylan's [original design doc](https://apollographql.atlassian.net/wiki/spaces/graphdev/pages/1451720708/join) (that I had forgotten about when we chatted in slack) also specifies this behavior. Part of my motivation is that if we do introduce backticked `string {templates}`, I don't think we should be jsonifying array and objects there either (instead requiring an explicit stringification like `->join`).

For null being "null", I'm trying to decide which of these is better:

1. null results "", so if you want `null`, you'll need something like:
    ```
    $->map(@->match([null, "null"], [@, @]))->join(",")
    ````
2. null results in `null`, so if you want "", you'll use 
    ```
    $->map(@->match([null, ""], [@, @]))->join(",")`
    ```

I'm leaning towards #1 — it's still data in that something is emitted (and the separator is used), but it's less likely to be junk. I don't expect users to be reading comma-delimited lists as json-encoded values, so `"null"` isn't useful.

apollo-federation/src/sources/connect/json_selection/methods/public/join.rs
lennyburdette: validation will require the separator. you won't be able to compose if you call `->join`. this runtime code is just for the case when the argument shape is unknown/named. i'm cool with returning None

apollo-federation/src/sources/connect/json_selection/methods/public/join.rs
lennyburdette: good question! i'm not sure what to do with this. i'm pretty sure it's giving me one mismatch per shape in the One<>, so i get: 
```
expected: String, received: List<List<String>>
expected: Float, received: List<List<String>>
expected: Bool, received: List<List<String>>
expected: null, received: List<List<String>>
expected: String, received: List<String>
expected: Float, received: List<String>
expected: Bool, received: List<String>
expected: null, received: List<String>
expected: One<String, Float, Bool, null>, received: List<String>
expected: One<String, Float, Bool, null, List<One<String, Float, Bool, null>>>, received: List<List<String>>
```

that's a lot, what should we do with this?

apollo-federation/src/sources/connect/json_selection/methods/public/join.rs
benjamn: I like mapping `null` to `"null"` too. If you see an unwanted `null` in your string, there shouldn't be much mystery about why it's there, and you have a number of tools to deal with it.

Once we have multi-arg `$(...)` and optional chaining with `?`, an even shorter way to replace `null` with `""` would be:
```
$->map($(@?, ""))->join(",")
```

dylan-apollo: I think we should do a sketch of how we expect this to come into play in practice. My gut is saying that most of the time folks will want to filter out nulls, so we should both sketch what filtering will look like and write down some example API designs that might use this.

lennyburdette:
This is now `->joinNotNull`, removing a footgun for batch uses and avoiding the whole topic of stringifying nulls

apollo-federation/src/sources/connect/json_selection/methods/public/join_not_null.rs
```diff
@@ -0,0 +1,349 @@
...
+                if let Some(err) = err {
+                    warnings.push(ApplyToError::new(
+                        err.to_string(),
+                        input_path.to_vec(),
+                        method_name.range(),
+                    ));
+                }
+                value
+            })
+            .collect::<Vec<_>>()
```
dylan-apollo, line 91: We can delete this intermediate Vec if we `use itertools::Itertools;`. That should allow us to `.join()` more efficiently (without collecting first)

apollo-federation/src/sources/connect/json_selection/methods/public/join_not_null.rs
```diff
@@ -0,0 +1,349 @@
...
+        match value {
+            JSON::Bool(b) => (
+                Some(b.then_some("true").unwrap_or("false").to_string()),
+                None,
+            ),
+            JSON::Number(number) => (Some(number.to_string()), None),
+            JSON::String(byte_string) => (Some(byte_string.as_str().to_string()), None),
+            JSON::Null => (None, None),
+            JSON::Array(_) | JSON::Object(_) => (
+                Some("".to_string()),
```
dylan-apollo, line 68: Adding an empty string still feels unexpected to me. If we can't join the thing, we should be erroring out (return `None` up above) instead of proceeding. That's what happens with other methods in their failure cases I think.

apollo-federation/src/sources/connect/json_selection/methods/public/join_not_null.rs
```diff
@@ -0,0 +1,342 @@
...
+    };
+
+    fn to_string(value: &JSON, method_name: &str) -> Result<Option<String>, String> {
+        match value {
+            JSON::Bool(b) => Ok(Some(b.then_some("true").unwrap_or("false").to_string())),
+            JSON::Number(number) => Ok(Some(number.to_string())),
+            JSON::String(byte_string) => Ok(Some(byte_string.as_str().to_string())),
+            JSON::Null => Ok(None),
+            JSON::Array(_) | JSON::Object(_) => Err(format!(
+                "Method ->{} requires an array of scalars values as input",
```
benjamn, line 65:
```suggestion
                "Method ->{} requires an array of scalar values as input",
```

benjamn (approved):
LGTM

PR #7169, opened 2025-04-03
Tracing deadlock investigation (backport #7142)

carodewig:
I've been trying to look into a deadlock on `tracing-subscriber::registry::Extensions`. The function `extract_dd_trace_id` is a relatively good candidate for the lock trigger due to the fact that it's looking at the root span extensions, rather than those of the current span.

For more information about when reentrant locking may cause a deadlock see: https://doc.rust-lang.org/std/sync/struct.RwLock.html

```
let _rg1 = lock.read();  |
                         |  // will block
                         |  let _wg = lock.write();
// may deadlock          |
let _rg2 = lock.read();  |
```

To test this we ran an instrumented version of the router that would log out if re-entrant locking of span extensions was detected.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

<hr>This is an automatic backport of pull request #7142 done by [Mergify](https://mergify.com).

### DIFF
--- a/.changesets/fix_caroline_deadlock.md
+++ b/.changesets/fix_caroline_deadlock.md
@@ -1,0 +1,20 @@
+### Fix potential telemetry deadlock ([PR #7142](https://github.com/apollographql/router/pull/7142))
+
+The `tracing_subscriber` crate uses `RwLock`s to manage access to a `Span`'s `Extensions`. Deadlocks are possible when
+multiple threads access this lock, including with reentrant locks:
+```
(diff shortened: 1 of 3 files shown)


BrynCooke (approved):

abernix:
@mergifyio backport 1.61.3

PR #7171, opened 2025-04-04
Represent subgraph building in typed stages

tninesling:
This change aims to create a strongly-typed contract for processing subgraphs. It replaces the `build_subgraph` utility with a `Subgraph` struct, which uses the [typestate pattern](https://cliffle.com/blog/rust-typestate/).

The states represented are as follows:
- `Raw`: This is the initial state once a schema is parsed.
- `Expanded`: Once a `Raw` subgraph is expanded, directive definitions are populated and subgraph metadata is computed.
- `Validated`: Once all Federation types and directives are validated, the subgraph enters this state.

This aims to ensure the following invariants:
1. After a subgraph is expanded, we should always have `SubgraphMetadata`. Encoding this in a strongly-typed way can avoid many unnecessary existence checks for the `Option<SubgraphMetadata>` currently stored in `FederationSchema`. It should also avoid future confusion around availability of that metadata for supergraph schemas, which will always have `None` there.
2. After a subgraph is validated, we should provide infallible iterators over directive applications. Currently, we provide iterators on `FederationSchema` of type `Result<Vec<Result<DirectiveType, FederationError>>, FederationError>`. Both preparing the iterator and parsing each directive can fail in prior states, but the `Validated` invariant will allow us to expose infallible iterators for use in merging later on.
3. Mutating a `Validated` subgraph invalidates the schema. This is enforced by consuming the owned `Validated` subgraph and returning an `Expanded` subgraph, as shown in `Subgraph<Validated>::invalidate`. The compiler will enforce re-validation prior to passing the subgraph to another stage that requires validation.

Notes for reviewers:
- The diff is largely due to copy-pasting the tests from `apollo-federation/src/schema/blueprint.rs` to `apollo-federation/src/subgraph/typestate.rs`. This allows simpler testing via accessing private fields of `Subgraph`.
- This should eventually replace `crate::subgraph::Subgraph` and its `parse_and_expand` function, but there is still a non-trivial difference between that and the new implementation, and the existing one is being used in connectors composition code.

---

**Checklist**

- Tests added and passing[^3]
    - [X] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-federation/src/subgraph/typestate.rs
```diff
@@ -0,0 +1,660 @@
...
+#[derive(Clone, Debug)]
+pub struct Validated {
+    schema: ValidFederationSchema,
+    metadata: SubgraphMetadata,
+}
+
+pub trait SubgraphState {}
+impl SubgraphState for Raw {}
+impl SubgraphState for Expanded {}
+impl SubgraphState for Validated {}
```
TylerBloom, line 35: This is largely unneeded unless you need to be generic over all (or some) of these typestates. You can't directly construct a `Subgraph` type other than `Subgraph<Raw>` and, from there, you can only create either of these other two states.

Generally, traits are used to group states to reduce duplication. For example, the method `Subgraph<Expanded>::metadata` can be implemented via a grouping trait:
```rust
// Note: this does not *need* to leak out of this module
trait SubgraphMetadataState {
  fn metadata(&self) -> SubgraphMetadata
}

impl SubgraphMetadataState for Expanded { ... }
impl SubgraphMetadataState for Validated { ... }

impl<S: SubgraphMetadataState> Subgraph<S> {
  fn metdata(&self) -> { self.state.metadata() }
}
```

apollo-federation/src/subgraph/typestate.rs
```diff
@@ -0,0 +1,660 @@
...
+///          (mutate/invalidate)
+///  ```
+///
+/// Subgraph states and their invariants:
+/// - `Raw`: The initial state, containing a raw schema. This provides no guarantees about the schema, other than
+///   that it can be parsed.
+/// - `Expanded`: The schema's links have been expanded to include missing directive definitions and subgraph
+///   metadata has been computed.
+/// - `Validated`: The schema has been validated according to Federation rules. Iterators over directives should
+///   be infallible at this stage.
```
TylerBloom, line 55: I really like this documentation. Typestates are meant to help you feel more confident about these invariants. Perhaps we ought to be able to say: "Iterators over directives *are* infallible at this stage" about the last invariant.

apollo-federation/src/subgraph/typestate.rs
```diff
@@ -0,0 +1,660 @@
...
+/// A subgraph represents a schema and its associated metadata. Subgraphs are updated through the
+/// composition pipeline, such as when links are expanded or when fed 1 subgraphs are upgraded to fed 2.
+/// We aim to encode these state transitions using the [typestate pattern](https://cliffle.com/blog/rust-typestate).
+///
+/// ```text
+///   (expand)     (validate)
+/// Raw ──► Expanded ──► Validated
+///            ▲             │
+///            └────────────┘
+///          (mutate/invalidate)
```
TylerBloom, line 46: Out of curiosity, will this grow/get more granular in the future?

TylerBloom (commented):
Looks good overall. I have general comments about the structure of the typestates.

apollo-federation/src/subgraph/mod.rs
```diff
@@ -362,11 +362,20 @@ impl From<ValidFederationSubgraph> for ValidSubgraph {
 /// for missing error codes and the like.
 #[allow(dead_code)]
 #[derive(Clone, Debug)]
-pub(crate) struct SubgraphError {
-    pub(crate) subgraph: Name,
+pub struct SubgraphError {
```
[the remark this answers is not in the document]
tninesling: I think we probably should eventually. Or, we should at least have some representation of a composition error other than the generic FederationError, whether it goes in its own module or not. I put this here just to get the formatting right for the 1:1 port of JS tests, but I expect a longer-term solution to be developed in the merging epic, specifically in FED-411.

apollo-federation/src/schema/schema_upgrader.rs
```diff
@@ -11,17 +11,17 @@ use super::position::InterfaceFieldDefinitionPosition;
...
+use crate::subgraph::typestate::Subgraph;
 use crate::utils::FallibleIterator;
 
 #[derive(Clone, Debug)]
 struct SchemaUpgrader<'a> {
     schema: FederationSchema,
-    original_subgraph: &'a ValidFederationSubgraph,
-    subgraphs: &'a ValidFederationSubgraphs,
+    original_subgraph: &'a Subgraph<Expanded>,
+    subgraphs: &'a Vec<Subgraph<Expanded>>,
```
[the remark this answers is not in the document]
tninesling: Sure. Updated to slice in [e505d8a](https://github.com/apollographql/router/pull/7171/commits/e505d8a2a1350bed4db57f2d1fc5d2e976d623e9)

apollo-federation/src/schema/schema_upgrader.rs
```diff
@@ -89,20 +78,20 @@ pub(crate) fn upgrade_subgraphs_if_necessary(
 impl<'a> SchemaUpgrader<'a> {
     #[allow(unused)]
     fn new(
-        original_subgraph: &'a ValidFederationSubgraph,
-        subgraphs: &'a ValidFederationSubgraphs,
+        original_subgraph: &'a Subgraph<Expanded>,
+        subgraphs: &'a Vec<Subgraph<Expanded>>,
```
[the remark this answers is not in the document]
tninesling: Updated in [e505d8a](https://github.com/apollographql/router/pull/7171/commits/e505d8a2a1350bed4db57f2d1fc5d2e976d623e9)

apollo-federation/src/subgraph/typestate.rs
tninesling: Good point. I was trying to demonstrate that, but the empty trait wasn't quite correct. Updated to use your metadata example to expose common functions for getting `SubgraphMetadata` and `FederationSchema` when available. It's basic, but it at least shows the pattern for folks updating in the future. Thanks!

apollo-federation/src/subgraph/typestate.rs
tninesling: Updated the wording. It was "should" more as a marker to me later to add the iterator itself, but I've added a stub for the infallible iterator here and updated to "are".

apollo-federation/src/subgraph/typestate.rs
tninesling: I think there's room for that. My current vision is that we go through the pipeline as:
`Subgraph<Raw> -> Subgraph<Expanded> -> Subgraph<Validated> -> Supergraph<Raw> -> Supergraph<Satisfiable>` (or something similar).

Right now, that means `SchemaUpgrader` operates on a mutable `Subgraph<Expanded>`. However, that will guarantee an invariant that we're looking at a Fed 2 subgraph, so there's potential for encoding the Federation major version as an invariant at some point. There's also a lot of code on the JS side which has memoized, lazily-computed data which may or may not fit into these stages and may inform new stages as needed.

apollo-federation/src/subgraph/typestate.rs
```diff
@@ -0,0 +1,691 @@
...
+
+        // Now that we have the definition for `@link` and an application, the bootstrap directive detection should work.
+        schema.collect_links_metadata()?;
+
+        FederationBlueprint::on_directive_definition_and_schema_parsed(&mut schema)?;
+
+        // Also, the backfilled definitions mean we can collect deep references.
+        schema.collect_deep_references()?;
+
+        // TODO: Remove this and use metadata from this Subgraph instead of FederationSchema
```
dariuszkuc, line 154: do we have a follow up item to tackle this?

dariuszkuc:
Very nice! Are those TODOs going to be fixed with this PR or there is some follow up work already created to address them?

apollo-federation/src/subgraph/typestate.rs
tninesling: Created a couple tickets to track this work. Effectively it boils down to using this new `Subgraph<Validated>` instead of `FederationSubgraph` or `ValidFederationSchema` where applicable. The other TODOs in the schema upgrader will get tackled as part of my upgrader work this week.

duckki (approved):
Nice!
My only concern is that the `subgraph` module is too crowded with two `Subgraph` implementations in it. They may need to live together for a few quarters. I wish there was a separate directory like `composition` and put this implementation there (like `composition/subgraph.rs`?) even though we may move it to `subgraph` directory later.

apollo-federation/src/subgraph/typestate.rs
```diff
@@ -0,0 +1,691 @@
...
+    metadata: SubgraphMetadata,
+}
+
+#[derive(Clone, Debug)]
+pub struct Validated {
+    schema: ValidFederationSchema,
+    metadata: SubgraphMetadata,
+}
+
+trait SubgraphMetadataState {
```
duckki, line 33: How about a shorter name like `HasMetadata`?

tninesling:
> Nice! My only concern is that the `subgraph` module is too crowded with two `Subgraph` implementations in it. They may need to live together for a few quarters. I wish there was a separate directory like `composition` and put this implementation there (like `composition/subgraph.rs`?) even though we may move it to `subgraph` directory later.

Definitely understand the concern. I'm planning to do a relatively fast follow to replace the existing one with this representation. I just wanted to keep this PR relatively self-contained and reviewable. Swapping that implementation out requires touching a bunch of files that would have distracted from the pattern itself. For context, the other Subgraph implementation is primarily used in connectors composition.

apollo-federation/src/subgraph/typestate.rs
TylerBloom: Good to know. I'm a big fan of making that implicit state as explicit as possible when the time comes!

TylerBloom (approved):

PR #7174, opened 2025-04-07
fix: update tokio to address unsoundness issue

goto-bus-stop:
This addresses [RUSTSEC-2025-0023](https://rustsec.org/advisories/RUSTSEC-2025-0023).

PR #7175, opened 2025-04-07
fix: update tokio to address unsoundness issue (backport #7174)

goto-bus-stop:
This addresses [RUSTSEC-2025-0023](https://rustsec.org/advisories/RUSTSEC-2025-0023).

I don't have a reason to assume that we're affected by this issue (you have to have an unusual `Clone` implementation), so the update is just to unblock our CI, and a changeset is unnecessary.

This is a manual backport of #7174, only increasing the patch version of tokio and not the minor.


BrynCooke (approved):

SimonSapin (approved):

abernix:
@mergifyio backport 1.61.3

PR #7207, opened 2025-04-10
fix(entity-cache): support several key directives on the same type 

bnjjj:

```graphql
  upc: ID!
  sku: ID!
  name: String
}
```

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [x] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-1244]: https://apollographql.atlassian.net/browse/ROUTER-1244?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apollo-router/src/plugins/cache/entity.rs
+++ b/apollo-router/src/plugins/cache/entity.rs
@@ -1426,13 +1426,13 @@ fn extract_cache_keys(
 
         let mut representation_entity_keys = IndexMap::new();
         for entity_key in entity_keys {
+            dbg!(&entity_key);
             // We remove it from original representation to not hash it both in entity_hash_key and representation_hash_key
-            let (key, value) = representation
-                .remove_entry(entity_key.as_str())
-                .ok_or_else(|| FetchError::MalformedRequest {
-                    reason: format!("can't get entity key {entity_key:?} in representations"),
-                })?;
-            representation_entity_keys.insert(key, value);
+            let entry = representation.remove_entry(entity_key.as_str());
+            // In case of several @key directives using different field set sometimes we don't have the related entity keys in representations
+            if let Some((key, value)) = entry {
+                representation_entity_keys.insert(key, value);
+            }
         }
 
         let hashed_representation = if representation.is_empty() {
@@ -1474,7 +1474,7 @@ fn get_entity_keys_from_supergraph_schema(
     subgraph_name: &str,
     supergraph_schema: &Valid<Schema>,
     subgraph_enums: &HashMap<String, String>,
-) -> Result<impl Iterator<Item = Name>, BoxError> {
+) -> Result<HashSet<Name>, BoxError> {
     let entity_keys = supergraph_schema
         .types
         .get(typename)
@@ -1488,7 +1488,6 @@ fn get_entity_keys_from_supergraph_schema(
                 .specified_argument_by_name("graph")
                 .and_then(|arg| arg.as_enum())
                 .and_then(|arg| subgraph_enums.get(arg.as_str()))?;
-
             if schema_subgraph_name == subgraph_name {
                 let mut parser = Parser::new();
                 directive
@@ -1516,7 +1515,7 @@ fn get_entity_keys_from_supergraph_schema(
                 .collect::<Vec<Name>>()
         });
 
-    Ok(entity_keys)
+    Ok(entity_keys.collect())
 }
 
 // Only hash the list of entity keys
--- a/apollo-router/src/plugins/cache/entity.rs
+++ b/apollo-router/src/plugins/cache/entity.rs
@@ -1426,7 +1426,6 @@ fn extract_cache_keys(
 
         let mut representation_entity_keys = IndexMap::new();
         for entity_key in entity_keys {
-            dbg!(&entity_key);
             // We remove it from original representation to not hash it both in entity_hash_key and representation_hash_key
             let entry = representation.remove_entry(entity_key.as_str());
             // In case of several @key directives using different field set sometimes we don't have the related entity keys in representations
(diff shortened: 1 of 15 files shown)


apollo-router/src/testdata/supergraph.graphql
```diff
@@ -68,6 +68,7 @@ enum link__Purpose {
 
 type Product
   @join__type(graph: INVENTORY, key: "upc")
+  @join__type(graph: INVENTORY, key: "upc inStock")
```
goto-bus-stop, line 71: Can we put this in a separate file so other tests don't also have this extra key? It could just literally be a copy of this but specific for your new test. Since existing tests are passing it probably doesn't have a negative effect but I think putting things highly specific for one test in this supergraph file that's shared by loads of tests is going to be risky going forward

apollo-router/src/plugins/cache/tests.rs
```diff
@@ -404,6 +404,133 @@ async fn insert_with_requires() {
     insta::assert_json_snapshot!(response);
 }
 
+#[tokio::test]
+async fn insert_with_several_key_directives() {
+    let valid_schema =
+        Arc::new(Schema::parse_and_validate(SCHEMA_REQUIRES, "test.graphql").unwrap());
+    let query = "query { topProducts { shippingEstimate price } }";
+
+    let subgraphs = MockedSubgraphs([
```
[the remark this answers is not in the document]
bnjjj: I'll do this in a follow up PR but good suggestion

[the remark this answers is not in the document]
goto-bus-stop: yeah that's fine :)

apollo-router/src/testdata/supergraph.graphql
bnjjj: Done

apollo-router/src/plugins/cache/entity.rs
```diff
@@ -1427,12 +1427,11 @@ fn extract_cache_keys(
         let mut representation_entity_keys = IndexMap::new();
         for entity_key in entity_keys {
             // We remove it from original representation to not hash it both in entity_hash_key and representation_hash_key
-            let (key, value) = representation
-                .remove_entry(entity_key.as_str())
-                .ok_or_else(|| FetchError::MalformedRequest {
-                    reason: format!("can't get entity key {entity_key:?} in representations"),
-                })?;
-            representation_entity_keys.insert(key, value);
+            let entry = representation.remove_entry(entity_key.as_str());
```
duckki, line 1430: It appears that the `extract_cache_keys` is trying to partition the `representation` value into the "key" part and "requires" part. But, it doesn't appear to be precise.

* This PR puts all fields that may be in *some* key fields into the key part. But, it will be an over-approximation, since any given entity fetch will only use one particular key directive at a time.
* The `get_entity_keys_from_supergraph_schema` function collects only the root fields from the key fields selection set. Thus, it will be imprecise since a `representation` value may have a field with a mixed sub-selection containing both key selections and requires selections.

It's unclear to me whether this imprecision is fine for entity caching and invalidation purposes.

duckki: Example:
```graphql
type T @key(fields: "k { a }") @key(fields: "k { b }") {
  k: K!
  data: Int! @requires(fields: "k { c }")
}
```
At the moment, we can't plan such a `@require` overlapping with a `@key` due to a bug (FED-507). But, it may be fixed in the future.

bnjjj: Thanks @duckki ! Would you have an example of generated QP for the entity you used in your example ? Because I'm still not sure to understand the part where I should keep the selection set

duckki: Schema & operation:
```graphql
#----------------------------------------------------------
#!subgraph A

type K {
    a: Int!
    b: Int!
}

type T @key(fields: "k { a }") @key(fields: "k { b }") {
  k: K!
  r: Int! @external
  data: Int! @requires(fields: "r")
}

#----------------------------------------------------------
#!subgraph B

type K {
    a: Int! @shareable
    b: Int! @shareable
}

type T @key(fields: "k { a }") @key(fields: "k { b }") {
  k: K! @shareable
  r: Int!
}

type Query {
  start: T!
}

#----------------------------------------------------------
#!operation
{
    start {
        data
    }
}
```

The query plan:
```
QueryPlan {
  Sequence {
    Fetch(service: "B") {
      {
        start {
          __typename
          k {
            a
          }
          r
        }
      }
    },
    Flatten(path: "start") {
      Fetch(service: "A") {
        {
          ... on T {
            __typename
            k {
              a
            }
            r
          }
        } =>
        {
          ... on T {
            data
          }
        }
      },
    },
  },
}
```

duckki: So, we will need to keep the selection set of
```
            k {
              a
            }
```
in the entity key hash.

bnjjj: Ok correct me if I'm wrong but in the current implementation we take the root field which is k here. So in representation I guess we'll have something like "k":{"a": 42}. In our logic we make representation.remove("k") which will result having "k": {"a": 42} and so we would have the whole data including the selection set and that would be good enough right ?

duckki: Here's my attempt: https://github.com/apollographql/router/compare/dev...duckki/entity-cache-matching-key

dariuszkuc: I think @duckki 's changes seem reasonable but we still got an issue that given keys `A` + `A + B` we will always pick just `A` (that's the QP behavior anyway so I guess this should be fine?)

Long term we should update QP to provide `@key` + `@requires` info so you don't need to reparse those field sets and just read those off directly from `FetchNode`

bnjjj:
Closing this one in favor of https://github.com/apollographql/router/pull/7228

PR #7214, opened 2025-04-10
fix(federation): return internal errors for invalid query plans

goto-bus-stop:
When an invalid query plan is generated, we may panic and crash the
router. This was probably introduced before we had the short error
macros :)

Let's use the `bail!()` macro for all these cases and return internal
errors. None of these cases should happen unless there is a bug.

[ROUTER-1230]: https://apollographql.atlassian.net/browse/ROUTER-1230?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BrynCooke (approved):

dariuszkuc (approved):

abernix:
@mergifyio backport 2.1.3

PR #7218, opened 2025-04-10
Add test ensuring we don't cache temporary query planner errors

Velfi:
This PR adds tests and removes an unused error variant.

Fixes #**issue_number**

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [x] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


carodewig (commented):
LGTM! I wonder if there would be a relatively easy way to test that the caching query planner does reject those errors?

apollo-router/src/ageing_priority_queue.rs
```diff
@@ -72,7 +72,11 @@ where
     pub(crate) fn send(&self, priority: Priority, message: T) -> Result<(), SendError> {
         self.queued_count
             .fetch_update(Ordering::Relaxed, Ordering::Relaxed, |previous_count| {
-                (previous_count < self.capacity).then_some(previous_count + 1)
+                if rand::random::<bool>() {
+                    None
+                } else {
+                    (previous_count < self.capacity).then_some(previous_count + 1)
+                }
```
Velfi, line 79: Note to self to remove this before merging

apollo-router/src/query_planner/caching_query_planner.rs
```diff
@@ -970,32 +974,108 @@ mod tests {
         );
     }
 
+    // Expect that if we call the CQP twice, the second call will return cached data
     #[test(tokio::test)]
-    async fn test_introspection_cache() {
+    async fn test_cache_works() {
```
Velfi, line 979: Note to self: bring back the old test

apollo-router/src/query_planner/testdata/starstuff.graphql
```diff
@@ -0,0 +1,98 @@
+schema
```
SimonSapin, line 1: I think we already have a number of copies of this schema. Is this one different? Or could an existing file be reused instead?

SimonSapin (approved):

apollo-router/src/query_planner/testdata/starstuff.graphql
Velfi: I can point to an existing copy. I'll update the code.

PR #7220, opened 2025-04-10
Simplify how URI templates are built

dylan-apollo:
This changes URI templates in a variety of ways both to simplify the code and to lay the foundation for future URI enhancements.

## Bug fixes

Correct some potentially surprising behavior with how we build URIs. We should test against known schemas to see if this will impact any current users.

1. Characters outside of `{ }` expressions will no longer be percent-encoded unless they are completely invalid for a URI. This is intended to better match what users would expect when writing those characters (and aligns better with RFC-6570). Users can opt-in to percent encoding by sticking those characters into expressions.
2. Trailing slashes will now be preserved/respected. Previously, paths were always normalized.

## Features

1. Expressions can now be used anywhere in a URI template _after_ the domain (if present). Previously, we only allowed expressions in very specific locations (a path segment, immediately following a query param name and `=`). This gets us a little closer to string interpolation.

## Refactors

1. `URLTemplate` was removed in favor of directly using `StringTemplate` with a new special URI-encoding interpolation mode
2. All connectors' usage of `url::Url` has been replaced with `http::Uri` which we needed to use for requests anyway. This eliminates the possibility of a mismatch between the two crates and simplifies some code.

## TODO:

- [x] More validation of the percent encoding implementation to ensure completeness
- [x] More thorough percent encoding of constant pieces of the URI
- [x] Don't allow expressions to modify domains
- [x] Document this change
- [x] Special-case a connect template of "/" to _not_ append that slash to source, since there's no other way to avoid a trailing slash.

[CNN-669]: https://apollographql.atlassian.net/browse/CNN-669?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apollo-federation/src/sources/connect/validation/http/url.rs
```diff
@@ -2,23 +2,35 @@ use std::fmt::Display;
...
+        return Err(Message {
+            code: Code::InvalidUrlScheme,
+            message: format!("Base URL for {coordinate} did not start with http:// or https://.",),
+            locations: value
+                .line_column_range(&schema.sources)
+                .into_iter()
+                .collect(),
+        });
+    };
+    if *scheme == Scheme::HTTP || *scheme == Scheme::HTTPS {
```
lennyburdette, line 30: i'm a little surprised no one has asked for unix: — we had to support it for subgraphs/service meshes running on sockets

lennyburdette (approved):
appreciate the simplification of removing urltemplate!

.changesets/feat_connectors_uri_expression_locations.md
```diff
@@ -0,0 +1,15 @@
+### Allow expressions in more locations in Connectors URIs ([PR #7220](https://github.com/apollographql/router/pull/7220))
+
+Previously, we only allowed expressions in very specific locations in Connectors URIs:
+
+1. A path segment, like `/users/{$args.id}`
+2. A query parameter's _value_, like `/users?id={$args.id}`
+
+Expressions can now be used anywhere in or after the path of the URI.
+For example, you can do
+`@connect(http: {GET: "/users?{$args.filterName}={$args.filterValue}#{$args.fragmentStuff}"})`.
```
dylan-apollo, line 10:
```suggestion
`@connect(http: {GET: "/users?{$args.filterName}={$args.filterValue}"})`.
```

PR #7228, opened 2025-04-11
fix(entity-cache): handle multiple key directives

duckki:
This is my suggestion for PR https://github.com/apollographql/router/pull/7207.

```graphql
  upc: ID!
  sku: ID!
  name: String
}
```

[ROUTER-1244]: https://apollographql.atlassian.net/browse/ROUTER-1244?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changesets/fix_duckki_entity_cache_matching_key.md
+++ b/.changesets/fix_duckki_entity_cache_matching_key.md
@@ -1,6 +1,6 @@
 ### Entity-cache: handle multiple key directives ([PR #7228](https://github.com/apollographql/router/pull/7228))
 
-This PR fixes a bug in entity caching introduced by the fix in https://github.com/apollographql/router/pull/6888 for cases where several `@key` directives with different fields were declared on a type as documented [here](https://www.apollographql.com/docs/graphos/schema-design/federated-schemas/reference/directives#managing-types). 
+This PR fixes a bug in entity caching introduced by the fix in https://github.com/apollographql/router/pull/6888 for cases where several `@key` directives with different fields were declared on a type as documented [here](https://www.apollographql.com/docs/graphos/schema-design/federated-schemas/reference/directives#managing-types).
 
 For example if you have this kind of entity in your schema:
 
@@ -11,9 +11,5 @@ type Product @key(fields: "upc") @key(fields: "sku") {
   name: String
 }
 ```
-
-<!-- [ROUTER-1244] -->
-
-[ROUTER-1244]: https://apollographql.atlassian.net/browse/ROUTER-1244?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
 
 By [@duckki](https://github.com/duckki) & [@bnjjj](https://github.com/bnjjj) in https://github.com/apollographql/router/pull/7228
--- a/.changesets/fix_duckki_entity_cache_matching_key.md
+++ b/.changesets/fix_duckki_entity_cache_matching_key.md
@@ -1,0 +1,19 @@
+### Entity-cache: handle multiple key directives ([PR #7228](https://github.com/apollographql/router/pull/7228))
+
+This PR fixes a bug in entity caching introduced by the fix in https://github.com/apollographql/router/pull/6888 for cases where several `@key` directives with different fields were declared on a type as documented [here](https://www.apollographql.com/docs/graphos/schema-design/federated-schemas/reference/directives#managing-types). 
+
+For example if you have this kind of entity in your schema:
+
+```graphql
+type Product @key(fields: "upc") @key(fields: "sku") {
+  upc: ID!
+  sku: ID!
+  name: String
+}
+```
+
+<!-- [ROUTER-1244] -->
+
+[ROUTER-1244]: https://apollographql.atlassian.net/browse/ROUTER-1244?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
+
+By [@duckki](https://github.com/duckki) & [@bnjjj](https://github.com/bnjjj) in https://github.com/apollographql/router/pull/7228
--- a/apollo-router/src/plugins/cache/entity.rs
+++ b/apollo-router/src/plugins/cache/entity.rs
@@ -663,7 +663,6 @@ impl CacheService {
                         );
 
                         let mut response = self.service.call(request).await?;
-
                         let cache_control =
                             if response.response.headers().contains_key(CACHE_CONTROL) {
                                 CacheControl::new(response.response.headers(), self.storage.ttl)?
@@ -771,6 +770,7 @@ impl CacheService {
                                     reason: e.to_string(),
                                 },
                             };
+                            println!("errrooooor !!!! {e:?}");
 
                             let graphql_error = e.to_graphql_error(None);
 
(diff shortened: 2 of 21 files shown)


apollo-router/src/plugins/cache/entity.rs
```diff
@@ -1507,37 +1521,131 @@ fn get_entity_keys_from_supergraph_schema(
...
 // Hash other representation variables except __typename and entity keys
 fn hash_other_representation(
     representation: &mut serde_json_bytes::Map<ByteString, Value>,
 ) -> String {
-    // We had to sort it to be deterministic
-    representation.sort_keys();
-    let mut digest = Sha256::new();
-    digest.update(serde_json::to_string(&representation).unwrap().as_bytes());
-    hex::encode(digest.finalize().as_slice())
+    hash_representation(representation)
```
duckki, line 1648: After this change, hash keys changed as expected.
But, one test failure looked odd. Not sure why "cache_control" changed.
```
Snapshot file: apollo-router/src/plugins/cache/snapshots/apollo_router__plugins__cache__tests__insert_with_requires.snap
Snapshot: insert_with_requires
Source: apollo-router/src/plugins/cache/tests.rs:367
────────────────────────────────────────────────────────────────────────────────
Expression: cache_keys
────────────────────────────────────────────────────────────────────────────────
-old snapshot
+new results
────────────┬───────────────────────────────────────────────────────────────────
    0     0 │ [
    1     1 │   {
    2     2 │     "key": "version:1.0:subgraph:inventory:type:Product:entity:72bafad9ffe61307806863b13856470e429e0cf332c99e5b735224fb0b1436f7:representation:7a26bc9b6cfab31cd9278921710167eaf9cde6c17d5f86102c898a04a06aeb6c:hash:a7b76558fe61b2f89ee9ea8bac8eee0b702ba1f319eaace576ab6ccdd19cdc80:data:d9d84a3c7ffc27b0190a671212f3740e5b8478e84e23825830e97822e25cf05c",
    3     3 │     "status": "new",
    4       │-    "cache_control": "public"
          4 │+    "cache_control": "no-store"
    5     5 │   },
    6     6 │   {
    7     7 │     "key": "version:1.0:subgraph:products:type:Query:hash:0a9c11d2f86fae30d5fdfe748e0be34c3d69dd26fa6520ee683d4d002e36e62a:data:d9d84a3c7ffc27b0190a671212f3740e5b8478e84e23825830e97822e25cf05c",
    8     8 │     "status": "new",
────────────┴───────────────────────────────────────────────────────────────────
To update snapshots run `cargo insta review`
Stopped on the first failure. Run `cargo insta test` to run all snapshots.
test plugins::cache::tests::insert_with_requires ... FAILED
```

bnjjj: fixed, it was caused by a mock

apollo-router/src/plugins/cache/entity.rs
```diff
@@ -772,6 +770,7 @@ impl CacheService {
                                     reason: e.to_string(),
                                 },
                             };
+                            println!("errrooooor !!!! {e:?}");
```
goto-bus-stop, line 773:
```suggestion
```

apollo-router/src/plugins/cache/entity.rs
```diff
@@ -1507,37 +1520,136 @@ fn get_entity_keys_from_supergraph_schema(
...
+    hex::encode(digest.finalize().as_slice())
 }
 
 // Only hash the list of entity keys
 pub(crate) fn hash_entity_key(
-    entity_keys: &IndexMap<ByteString, serde_json_bytes::Value>,
+    entity_keys: &serde_json_bytes::Map<ByteString, serde_json_bytes::Value>,
 ) -> String {
+    #[cfg(test)]
+    dbg!(&entity_keys);
```
duckki, line 1643: Maybe, remove this?

duckki (commented):
I wish we had tests with nested `@key` fields and multiple sub-selections like `@key(fields: "k { a b }") @key(fields: "k { b c }")`.
Also some tests for `hash_representation` being stable over re-ordering of nested fields. But, we'd need a new test supergraph schema. 😢

apollo-router/src/plugins/cache/entity.rs
bnjjj: Actually I would like to keep it, it's pretty easy when running the test to debug if anything goes wrong

.changesets/fix_duckki_entity_cache_matching_key.md
```diff
@@ -0,0 +1,19 @@
...
+type Product @key(fields: "upc") @key(fields: "sku") {
+  upc: ID!
+  sku: ID!
+  name: String
+}
+```
+
+<!-- [ROUTER-1244] -->
+
+[ROUTER-1244]: https://apollographql.atlassian.net/browse/ROUTER-1244?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
```
SimonSapin, line 17:
```suggestion
```

apollo-router/src/plugins/cache/entity.rs
SimonSapin: I think this should not stay as-is because it’d be a hindrance for debugging a test if something _else_ (unrelated to this function) goes wrong. If you want to keep some logging in `dev` it should be opt-in somehow. Maybe with `tracing::trace!`?

apollo-router/src/plugins/cache/entity.rs
```diff
@@ -1458,24 +1448,47 @@ fn extract_cache_keys(
...
+    subgraph_enums: &HashMap<String, String>,
+) -> Result<serde_json_bytes::Map<ByteString, Value>, FetchError> {
+    // find an entry in the `key_field_sets` that matches the `representation`.
+    let matched_key_field_set =
+        collect_key_field_sets(typename, subgraph_name, supergraph_schema, subgraph_enums)?
+        .find(|field_set| {
+            matches_selection_set(representation, &field_set.selection_set)
+        })
+        .ok_or_else(|| FetchError::MalformedRequest {
+            reason: format!("representation does not match any key field set for typename {typename} in subgraph {subgraph_name}"),
```
abernix, line 1474: @bnjjj @SimonSapin @goto-bus-stop @dariuszkuc @duckki — user experience by committee: I wonder if this is really a reasonable error to return to a client who made the query?  It certainly seems useful in terms of discovering that you're encountering an Apollo Router defect, but is this actually a user-serviceable error message? What will the user be able to do with this error message?  Most users won't understand what a representation is nor will they understand the `@key` details, right?

It seems more like "critical internal server error that we didn't expect" that perhaps maybe would be better emitted to the logs, but how do we decide?

dariuszkuc: indeed, while this error is descriptive of whats happening.... there is little action that users can do as this will be auto generated by the QP. +1 to emitting descriptive log message and just returning some generic internal server error

SimonSapin: As far as I understand hitting this error would indicate a bug in Router or composition. The best a user can do is give us a reproduction. So in a sense the contents of the error message doesn’t matter as long as we can recognize it as coming from here.

apollo-router/src/plugins/cache/entity.rs
bnjjj: Replaced with `tracing::trace!`

apollo-router/src/plugins/cache/entity.rs
bnjjj: I changed the error sent to the client and added a log

IvanGoncharov (approved):

abernix:
@mergifyio backport 2.1.3

PR #7236, opened 2025-04-11
Add spans for compute jobs

BrynCooke:

* `compute_job`:
* `compute_job.execution`
  - `job.age`: `P1`-`P8`

As the result of compute jobs are generally cached these extra spans are OK.

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [x] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

We don't currently document spans. This needs to be tackled separately.

**Notes**

### DIFF
--- a/apollo-router/src/compute_job/mod.rs
+++ b/apollo-router/src/compute_job/mod.rs
@@ -8,6 +8,11 @@ use std::time::Instant;
 use opentelemetry::metrics::MeterProvider as _;
 use opentelemetry::metrics::ObservableGauge;
 use tokio::sync::oneshot;
+use tracing::Instrument;
+use tracing::Span;
+use tracing::info_span;
+use tracing_core::Dispatch;
+use tracing_subscriber::util::SubscriberInitExt;
 
 use self::metrics::ActiveComputeMetric;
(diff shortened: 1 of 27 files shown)


apollo-router/src/compute_job/mod.rs
```diff
@@ -129,6 +136,8 @@ impl From<ComputeJobType> for opentelemetry::Value {
 }
 
 pub(crate) struct Job {
+    subscriber: Dispatch,
+    span: Span,
```
bnjjj, line 140: NIT: should we call it `parent_span` to make it more explicit ?

bnjjj (approved):

apollo-router/src/compute_job/mod.rs
BrynCooke: [a5ac86a](https://github.com/apollographql/router/pull/7236/commits/a5ac86ad11bda3a3ae377c63bd872b531566d80b)

BrynCooke:
@mergifyio refresh

BrynCooke:
@Mergifyio backport 1.x

PR #7243, opened 2025-04-11
Add support for $request.headers and $response.headers to connectors mapping

andrewmcgivery:
Add support for `$request.headers` and `$response.headers` to connectors mapping

TODO: Shape validation, versioning

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [x] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-federation/src/sources/connect/json_selection/parser.rs
+++ b/apollo-federation/src/sources/connect/json_selection/parser.rs
@@ -29,6 +29,7 @@ use super::location::WithRange;
 use super::location::merge_ranges;
 use super::location::new_span;
 use super::location::ranged_span;
+use crate::sources::connect::Namespace;
 use crate::sources::connect::variable::VariableNamespace;
 use crate::sources::connect::variable::VariablePathPart;
 use crate::sources::connect::variable::VariableReference;
@@ -231,11 +232,10 @@ impl JSONSelection {
         }
     }
 
-    pub fn external_variables<N: FromStr + ToString>(&self) -> impl Iterator<Item = N> {
+    pub fn variable_references(&self) -> impl Iterator<Item = VariableReference<Namespace>> + '_ {
         self.external_var_paths()
             .into_iter()
             .flat_map(|var_path| var_path.variable_reference())
-            .map(|var_ref| var_ref.namespace.namespace)
     }
 }
 
--- a/apollo-federation/src/sources/connect/validation/expression.rs
+++ b/apollo-federation/src/sources/connect/validation/expression.rs
@@ -3,6 +3,7 @@
 
 use std::ops::Range;
 use std::str::FromStr;
+use std::sync::LazyLock;
 
 use apollo_compiler::collections::IndexMap;
 use apollo_compiler::parser::LineColumn;
@@ -23,6 +24,17 @@ use crate::sources::connect::validation::Message;
 use crate::sources::connect::validation::coordinates::ConnectDirectiveCoordinate;
 use crate::sources::connect::validation::graphql::GraphQLString;
 use crate::sources::connect::validation::graphql::SchemaInfo;
+
+static REQUEST_SHAPE: LazyLock<Shape> = LazyLock::new(|| {
+    Shape::record(
+        [(
+            "headers".to_string(),
+            Shape::dict(Shape::list(Shape::string([]), []), []),
+        )]
+        .into(),
+        [],
+    )
+});
 
 /// Details about the available variables and shapes for the current expression.
 /// These should be consistent for all pieces of a connector in the request phase.
@@ -53,7 +65,7 @@ impl<'schema> Context<'schema> {
                     (Namespace::Args, shape_for_arguments(field_def)),
                     (Namespace::Config, Shape::unknown([])),
                     (Namespace::Context, Shape::unknown([])),
-                    (Namespace::Request, Shape::unknown([])),
+                    (Namespace::Request, REQUEST_SHAPE.clone()),
                 ]
                 .into_iter()
                 .collect();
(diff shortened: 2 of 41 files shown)


apollo-router/src/plugins/connectors/make_requests.rs
```diff
@@ -103,6 +109,61 @@ impl RequestInputs {
             }
         }
 
+        // Add headers from the original router request.
+        // Only include headers that are actually referenced to save on passing around unused headers in memory.
+        if variables_used.contains(&Namespace::Request) {
+            let new_headers = externalize_header_map(supergraph_request.headers())
+                .unwrap_or_default()
+                .iter()
+                .filter_map(|(key, value)| {
```
naomijub, line 118:
```rust
.filter_map(|(key, value)| 
  headers_used.contains(key.as_str())
    .then_some((
      key.as_str().into(),
      value
        .iter()
         .map(|s| Value::String(s.as_str().into()))
          .collect(),
    ))
)
```

apollo-router/src/plugins/connectors/make_requests.rs
```diff
@@ -103,6 +109,61 @@ impl RequestInputs {
...
+        // Add headers from the connectors response
+        // Only include headers that are actually referenced to save on passing around unused headers in memory.
+        if variables_used.contains(&Namespace::Response) {
+            if let Some(response_parts) = response_parts {
+                let new_headers: Map<ByteString, Value> =
+                    externalize_header_map(&response_parts.headers)
+                        .unwrap_or_default()
+                        .iter()
+                        .filter_map(|(key, value)| {
+                            if headers_used.contains(key.as_str()) {
```
naomijub, line 147: `then_some`

apollo-router/src/plugins/connectors/make_requests.rs
naomijub: https://doc.rust-lang.org/std/primitive.bool.html#method.then_some

andrewmcgivery: Slick! Updating now!

apollo-federation/src/sources/connect/validation/expression.rs
```diff
@@ -74,6 +75,7 @@ impl<'schema> Context<'schema> {
                     (Namespace::Batch, Shape::list(Shape::from(type_def), [])),
                     (Namespace::Config, Shape::unknown([])),
                     (Namespace::Context, Shape::unknown([])),
+                    (Namespace::Request, Shape::unknown([])),
```
lennyburdette, line 78: We do know this shape! Let's talk about this synchronously

apollo-federation/src/sources/connect/json_selection/parser.rs
```diff
@@ -231,11 +232,10 @@ impl JSONSelection {
         }
     }
 
-    pub fn external_variables<N: FromStr + ToString>(&self) -> impl Iterator<Item = N> {
+    pub fn variable_references(&self) -> impl Iterator<Item = VariableReference<Namespace>> + '_ {
```
lennyburdette, line 235: why was this change necessary?

apollo-federation/src/sources/connect/validation/expression.rs
```diff
@@ -24,6 +25,17 @@ use crate::sources::connect::validation::coordinates::ConnectDirectiveCoordinate
 use crate::sources::connect::validation::graphql::GraphQLString;
 use crate::sources::connect::validation::graphql::SchemaInfo;
 
+static REQUEST_SHAPE: LazyLock<Shape> = LazyLock::new(|| {
+    Shape::record(
+        [(
+            "headers".to_string(),
```
lennyburdette, line 31: we should have a test for when you try to access something besides `headers` on the object

apollo-router/src/plugins/connectors/make_requests.rs
```diff
@@ -41,12 +43,16 @@ impl RequestInputs {
     /// Creates a map for use in JSONSelection::apply_with_vars. It only clones
     /// values into the map if the variable namespaces (`$args`, `$this`, etc.)
     /// are actually referenced in the expressions for URLs, headers, body, or selection.
+    #[allow(clippy::too_many_arguments)]
```
lennyburdette, line 46: looks like we should make a builder — that would make the tests better too. not necessary for this PR

lennyburdette (commented):
i honestly don't know what will happen when you try `$request.x` or `$response.y` — let's get that sorted and then we're good to go

apollo-federation/src/sources/connect/json_selection/parser.rs
andrewmcgivery: Consistency and allow this to be reused for checking what actual headers are being used instead of having a separate method here. Note that before this method only returned the name of the namespace vs after this change, it returns the whole variable so that we can do additional checks other than just the namespace name.

This was discussed in the previous PR with you and Ben to avoid us slapping a bunch of header-related code into here as we decided this code should have no concept of headers.

Making this function more generic then allows it to be re-used for multiple purposes.

The alternative would be to have two functions here... one that returns the variable references and one that returns the variable namespaces... but my feeling is that would be very duplicative.

apollo-federation/src/sources/connect/validation/expression.rs
andrewmcgivery: Added!

apollo-router/src/plugins/connectors/make_requests.rs
andrewmcgivery: Would you want a "TODO" comment here? Something like "TODO: Consider converting to a builder" or something like that?

apollo-federation/src/sources/connect/json_selection/parser.rs
andrewmcgivery: This thread: https://github.com/apollographql/router/pull/6638#discussion_r1929164201

And this one: https://github.com/apollographql/router/pull/6638#discussion_r1930967266

lennyburdette (approved):

lennyburdette:
approved with the acknowledgment that we might add a version gate before this is released

PR #7250, opened 2025-04-11
Allow disabling subgraphs during query planning

sachindshinde:
This PR updates `build_query_plan()` to optionally accept a set of subgraphs that shouldn't be used for query planning. This may result in no query plan being found; if it does, the `SingleFederationError::NoPlanFoundWithDisabledSubgraphs` error is emitted (normally, composition's satisfiability check guarantees that a plan is always found).

[FED-586]: https://apollographql.atlassian.net/browse/FED-586?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


duckki (approved):
Looks good!

PR #7256, opened 2025-04-14
Add spans for compute jobs (backport #7236)

BrynCooke:

* `compute_job`:
* `compute_job.execution`
  - `job.age`: `P1`-`P8`

As the result of compute jobs are generally cached these extra spans are OK.

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [x] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

We don't currently document spans. This needs to be tackled separately.

**Notes**

<hr>This is an automatic backport of pull request #7236 done by [Mergify](https://mergify.com).


apollo-router/src/compute_job/mod.rs
```diff
@@ -111,34 +132,46 @@ where
...
+            queue_start: Instant::now(),
         };
-        match result {
-            Ok(r) => r,
-            Err(e) => Err(Box::new(e) as Box<dyn Any + Send>),
+        queue().send(compute_job_type.into(), job);
+
+        async move {
+            let result = rx.await;
+            // This local variable MUST exist
```
Velfi, line 161: Why must it exist?

Velfi (approved):

apollo-router/src/compute_job/mod.rs
BrynCooke: Ahh, I should have put a better comment on this.

It's because without the local variable Rust will do a partial move of the job watcher causing it to drop before you expect it to.

It's actually part of Rust: https://github.com/rust-lang/rust/issues/108808

PR #7121, opened 2025-03-26
Improve error message for invalid JWT Header values

IvanGoncharov:
Enhanced parsing error messages for JWT Authorization header values now provide developers with clear, actionable feedback while ensuring that no sensitive data is exposed.

Examples of the updated error messages:
```diff
-         Header Value: '<invalid value>' is not correctly formatted. prefix should be 'Bearer'
+         Value of 'authorization' JWT header should be prefixed with 'Bearer'
```

```diff
-         Header Value: 'Bearer' is not correctly formatted. Missing JWT
+         Value of 'authorization' JWT header has only 'Bearer' prefix but no JWT token
```

Fixes #**issue_number**

---

**Checklist**

Complete the checklist (and note appropriate exceptions) before the PR is marked ready-for-review.

- [x] Changes are compatible[^1]
- [x] Documentation[^2] completed
- [x] Performance impact assessed and acceptable
- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[^1]: It may be appropriate to bring upcoming changes to the attention of other (impacted) groups. Please endeavour to do this before seeking PR approval. The mechanism for doing this will vary considerably, so use your judgement as to how and when to do this.
[^2]: Configuration is an important part of many changes. Where applicable please try to document configuration examples.
[^3]: Tick whichever testing boxes are applicable. If you are adding Manual Tests, please document the manual testing (extensively) in the Exceptions.

[ROUTER-1212]: https://apollographql.atlassian.net/browse/ROUTER-1212?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #7257, opened 2025-04-14
prep release: v1.61.3

abernix:
> **Note**
>
> When approved, this PR will merge into **the `1.61.3` branch** which will — upon being approved itself — merge into `dev`.
>
>  - Version bumps
>  - That it targets the right release branch (`1.61.3` in this case!).
>
---
### Fix potential telemetry deadlock ([PR #7142](https://github.com/apollographql/router/pull/7142))

The `tracing_subscriber` crate uses `RwLock`s to manage access to a `Span`'s `Extensions`. Deadlocks are possible when
multiple threads access this lock, including with reentrant locks:
```
let _rg1 = lock.read();  |
                         |  // will block
                         |  let _wg = lock.write();
// may deadlock          |
let _rg2 = lock.read();  |
```

This fix removes an opportunity for reentrant locking while extracting a Datadog identifier.

There is also a potential for deadlocks when the root and active spans' `Extensions` are acquired at the same time, if
multiple threads are attempting to access those `Extensions` but in a different order. This fix removes a few cases
where multiple spans' `Extensions` are acquired at the same time.

By [@carodewig](https://github.com/carodewig) in https://github.com/apollographql/router/pull/7142

### Connection shutdown timeout 1.x ([PR #7058](https://github.com/apollographql/router/pull/7058))

```
supergraph:
  connection_shutdown_timeout: 60s
```

```
supergraph:
  early_cancel: true
```

```
traffic_shaping:
  router:
    timeout: 60s
```

By [@BrynCooke](https://github.com/BrynCooke) in https://github.com/apollographql/router/pull/7058

### Fix crash when an invalid query plan is generated ([PR #7214](https://github.com/apollographql/router/pull/7214))

When an invalid query plan is generated, the router could panic and crash.
This could happen if there are gaps in the GraphQL validation implementation.
Now, even if there are unresolved gaps, the router will handle it gracefully and reject the request.

By [@goto-bus-stop](https://github.com/goto-bus-stop) in https://github.com/apollographql/router/pull/7214

### Improve Error Message for Invalid JWT Header Values ([PR #7121](https://github.com/apollographql/router/pull/7121))

Enhanced parsing error messages for JWT Authorization header values now provide developers with clear, actionable feedback while ensuring that no sensitive data is exposed.

Examples of the updated error messages:
```diff
-         Header Value: '<invalid value>' is not correctly formatted. prefix should be 'Bearer'
+         Value of 'authorization' JWT header should be prefixed with 'Bearer'
```

```diff
-         Header Value: 'Bearer' is not correctly formatted. Missing JWT
+         Value of 'authorization' JWT header has only 'Bearer' prefix but no JWT token
```

By [@IvanGoncharov](https://github.com/IvanGoncharov) in https://github.com/apollographql/router/pull/7121


CHANGELOG.md
```diff
@@ -4,6 +4,89 @@ All notable changes to Router will be documented in this file.
...
+
+This fix removes an opportunity for reentrant locking while extracting a Datadog identifier.
+
+There is also a potential for deadlocks when the root and active spans' `Extensions` are acquired at the same time, if
+multiple threads are attempting to access those `Extensions` but in a different order. This fix removes a few cases
+where multiple spans' `Extensions` are acquired at the same time.
+
+By [@carodewig](https://github.com/carodewig) in https://github.com/apollographql/router/pull/7142
+
+### Connection shutdown timeout 1.x ([PR #7058](https://github.com/apollographql/router/pull/7058))
```
goto-bus-stop, line 32:
```suggestion
### Connection shutdown timeout ([PR #7058](https://github.com/apollographql/router/pull/7058))
```

goto-bus-stop (approved):

garypen (approved):

PR #7264, opened 2025-04-14
feature: Plugin for Enhanced Client Awareness

calvincestari:
This is the plugin to enable capturing of client library metadata, aka. Enhanced Client Awareness.

**Checklist**

  - I've been working closely with the Insights team and they're aware of this work and helping with the implementation.
- ~Documentation[^2] completed~
  - I don't believe any documentation is necessary because this is not a configurable plugin; it should _always_ execute. It is the clients that are configurable re. whether to send the request metadata or not.
- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [x] Manual Tests

**Manual Testing**
* Tested locally using the [Deepspace Dogfood supergraph with local Router code](https://github.com/mdg-private/deepspace-dogfood-supergraph?tab=readme-ov-file#running-local-router-code)
* Built a simple client using the [feature/client-awareness](https://github.com/apollographql/apollo-ios/tree/feature/client-awareness) branch of Apollo iOS
* Verified that the telemetry report included valid values for the `client_library_name` and `client_library_version` keys. Here's a sample trace from the Router console: `2025-04-22T19:03:45.033236Z DEBUG  submitting report: Report { traces_per_query: {"# ExampleQuery\nquery ExampleQuery{devices{__typename id}randomSensorValue stores{__typename id latitude longitude name planet{__typename name}}}": TracesAndStats { traces: [], stats_with_context: {StatsContext { client_name: "apollo-ios-client", client_version: "(unknown)", operation_type: "query", operation_subtype: "", result: "", client_library_name: "apollo-ios", client_library_version: "1.19.0" }: ContextualizedStats { context: StatsContext { client_name: "apollo-ios-client", client_version: "(unknown)", operation_type: "query", operation_subtype: "", result: "", client_library_name: "apollo-ios", client_library_version: "1.19.0" }, query_latency_stats: QueryLatencyStats ...`

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-1251]: https://apollographql.atlassian.net/browse/ROUTER-1251?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apollo-router/src/plugins/enhanced_client_awareness/mod.rs
+++ b/apollo-router/src/plugins/enhanced_client_awareness/mod.rs
@@ -14,7 +14,7 @@ const CLIENT_LIBRARY_KEY: &str = "clientLibrary";
 const CLIENT_LIBRARY_NAME_KEY: &str = "name";
 const CLIENT_LIBRARY_VERSION_KEY: &str = "version";
 
-#[derive(Debug, Default, Deserialize, JsonSchema)]
+#[derive(Debug, Deserialize, JsonSchema)]
 struct Config {
     #[serde(default = "default_enable_client_library_metrics")]
     enable_client_library_metrics: bool,
--- a/apollo-router/src/plugins/enhanced_client_awareness/mod.rs
+++ b/apollo-router/src/plugins/enhanced_client_awareness/mod.rs
@@ -82,3 +82,6 @@ register_plugin!(
     "enhanced_client_awareness",
     EnhancedClientAwareness
 );
+
+#[cfg(test)]
+mod tests;
--- a/apollo-router/src/plugins/enhanced_client_awareness/mod.rs
+++ b/apollo-router/src/plugins/enhanced_client_awareness/mod.rs
@@ -6,9 +6,9 @@ use tower::ServiceExt;
 
 use crate::plugin::Plugin;
 use crate::plugin::PluginInit;
-use crate::services::supergraph;
 use crate::plugins::telemetry::CLIENT_LIBRARY_NAME;
 use crate::plugins::telemetry::CLIENT_LIBRARY_VERSION;
+use crate::services::supergraph;
 
 const CLIENT_LIBRARY_KEY: &str = "clientLibrary";
 const CLIENT_LIBRARY_NAME_KEY: &str = "name";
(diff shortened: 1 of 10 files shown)


apollo-bot2:

apollo-router/src/plugins/enhanced_client_awareness/tests.rs
```diff
@@ -0,0 +1,112 @@
...
+            "Missing CLIENT_LIBRARY_NAME key/value"
+        );
+        let client_library_name: String = request
+            .context
+            .get(CLIENT_LIBRARY_NAME)
+            .unwrap_or_default()
+            .unwrap_or_default();
+        assert_eq!(
+            client_library_name,
+            serde_json::to_string("apollo-general-client-library")?
```
calvincestari, line 34: I'm not sure if this is what's required here? Without `serde_json::to_string` the comparison doesn't work because the string value pulled from `request.context` contains escaped quotes (`\"`). Please let me know if there is another way I should be doing the comparison.

apollo-router/src/plugins/enhanced_client_awareness/tests.rs
```diff
@@ -0,0 +1,112 @@
...
+        assert!(
+            request.context.contains_key(CLIENT_LIBRARY_VERSION),
+            "Missing CLIENT_LIBRARY_VERSION key/value"
+        );
+        let client_library_version: String = request
+            .context
+            .get(CLIENT_LIBRARY_VERSION)
+            .unwrap_or_default()
+            .unwrap_or_default();
+        assert_eq!(client_library_version, serde_json::to_string("0.1.0")?);
```
calvincestari, line 46: _Same comment as above._

apollo-router/src/plugins/enhanced_client_awareness/mod.rs
```diff
@@ -0,0 +1,87 @@
...
+
+#[async_trait::async_trait]
+impl Plugin for EnhancedClientAwareness {
+    type Config = Config;
+
+    // This is invoked once after the router starts and compiled-in
+    // plugins are registered
+    async fn new(init: PluginInit<Self::Config>) -> Result<Self, BoxError> {
+        Ok(EnhancedClientAwareness {
+            enabled: init.config.enable_client_library_metrics,
```
calvincestari, line 39: I don't actually want this plugin to be configurable by end users. IMO it should always be executed and the configuration is placed at the client where the library metadata can be sent or disabled.

`Config` seems to be a required part of a plugin though?

apollo-router/src/plugins/enhanced_client_awareness/tests.rs
bonnici: This is happening because when you call `to_string` on a JSON serde `Value` without checking if it's a string value, it will format with the extra escaped quotes. If you change the plugin code to something like:

```
                    if let Some(client_library_metadata) = request
                        .supergraph_request
                        .body()
                        .extensions
                        .get(CLIENT_LIBRARY_KEY)
                    {
                        if let Some(name) = client_library_metadata
                            .get(CLIENT_LIBRARY_NAME_KEY)
                            .and_then(|name| name.as_str())
                        {
                            let _ = request.context.insert(CLIENT_LIBRARY_NAME, name.to_string());
                        }

                        if let Some(version) = client_library_metadata
                            .get(CLIENT_LIBRARY_VERSION_KEY)
                            .and_then(|version| version.as_str())
                        {
                            let _ = request.context.insert(CLIENT_LIBRARY_VERSION, version.to_string());
                        }
                    };
```

The string in the context will be populated without the leading characters, and it will skip any context values that are not a string.

calvincestari: That's working great now, thanks for the tip @bonnici!

calvincestari:
@bonnici - the escaped strings issue is now fixed, along with the test.

bonnici (approved):
Looks good to me, but please wait for a router core review before merging

apollo-router/src/plugins/enhanced_client_awareness/mod.rs
BrynCooke: You should be able to use the empty tuple `()` for the config.

goto-bus-stop: After discussing a bit, it looks like `()` actually doesn't work right now. https://github.com/apollographql/router/blob/7041a1a565e79aff6298b890dc7b5bd61ca4dbb9/apollo-router/src/plugins/license_enforcement/mod.rs is another always-on plugin that has no user config: it uses an empty struct.

garypen (approved):

PR #7273, opened 2025-04-15
fix: poll pending compute jobs hang request

BrynCooke:
Compute jobs in the router are used to execute CPU intensive work outside of the main io worker threads, in particular for QueryParsing, QueryPlanning and Introspection.

We currently check in the pipeline if the compute job is full and if it is then it will return `Poll::Pending` in the tower services.
However this will cause requests to hang until timeout.

This PR shifts the logic into `call` and will immediately return a `SERVICE_UNAVAILABLE` response to the user.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [x] Integration Tests
    - [x] Manual Tests

**Exceptions**

No new docs

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/compute_job/mod.rs
+++ b/apollo-router/src/compute_job/mod.rs
@@ -29,14 +29,10 @@ use crate::metrics::meter_provider;
 /// a sizable backlog during spikes.
 fn queue_capacity() -> usize {
     // This environment variable is intentionally undocumented.
-    if let Some(threads) = std::env::var("APOLLO_ROUTER_COMPUTE_QUEUE_CAPACITY_PER_THREAD")
+    std::env::var("APOLLO_ROUTER_COMPUTE_QUEUE_CAPACITY_PER_THREAD")
         .ok()
         .and_then(|value| value.parse::<usize>().ok())
-    {
-        threads
-    } else {
-        1000
-    }
+        .unwrap_or(1000)
 }
 
 /// By default, let this thread pool use all available resources if it can.
--- a/apollo-router/src/compute_job/mod.rs
+++ b/apollo-router/src/compute_job/mod.rs
@@ -20,12 +20,24 @@ use crate::ageing_priority_queue::Priority;
 use crate::metrics::meter_provider;
 
 /// We generate backpressure in tower `poll_ready` when the number of queued jobs
-/// reaches `QUEUE_SOFT_CAPACITY_PER_THREAD * thread_pool_size()`
+/// reaches `APOLLO_ROUTER_COMPUTE_QUEUE_CAPACITY_PER_THREAD * thread_pool_size()`
+///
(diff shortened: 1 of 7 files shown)


apollo-router/src/compute_job/mod.rs
```diff
@@ -20,12 +20,20 @@ use crate::ageing_priority_queue::Priority;
...
 
 /// We generate backpressure in tower `poll_ready` when the number of queued jobs
-/// reaches `QUEUE_SOFT_CAPACITY_PER_THREAD * thread_pool_size()`
+/// reaches `APOLLO_ROUTER_COMPUTE_QUEUE_CAPACITY_PER_THREAD * thread_pool_size()`
+///
+/// The default for APOLLO_ROUTER_COMPUTE_QUEUE_CAPACITY_PER_THREAD is 1000
 ///
 /// This number is somewhat arbitrary and subject to change. Most compute jobs
 /// don't take a long time, so by making the queue quite big, it's capable of eating
 /// a sizable backlog during spikes.
```
Velfi, line 29: should we note the default value?

Velfi (approved):

apollo-router/src/compute_job/mod.rs
BrynCooke: It's in the comment: `The default for APOLLO_ROUTER_COMPUTE_QUEUE_CAPACITY_PER_THREAD is 1000`

apollo-router/src/services/router/service.rs
```diff
@@ -215,15 +216,31 @@ impl Service<RouterRequest> for RouterService {
...
+                            .extension_code("SERVICE_UNAVAILABLE")
+                            .build(),
+                    )
+                    .status_code(StatusCode::SERVICE_UNAVAILABLE)
+                    .header(CONTENT_TYPE, APPLICATION_JSON.essence_str())
+                    .context(context)
+                    .build()
+            }
+            .boxed();
+        }
```
carodewig, line 243: One thought for the future - putting the check here will cause the router to potentially reject queries that wouldn't have had to go through query planning (ie if they're in the cache or in-flight). It matches the existing behavior, but I wonder if it might be helpful to separate out the parsing vs planning queues.

carodewig (approved):

apollo-router/src/services/router/service.rs
BrynCooke: That's a possibility. I'm uncertain at the moment if query parsing is actually cached anymore, so maybe there is no request that does not have at least one compute job.

BrynCooke:
A followup from this will be to forward port the test to dev.

apollo-router/src/services/router/service.rs
goto-bus-stop: Query parsing is cached in-memory https://github.com/apollographql/router/blob/92f9efdfb812a5e4b48be820792fd10719c43f94/apollo-router/src/services/layers/query_analysis.rs#L275-L283

PR #7274, opened 2025-04-15
WIP: Add support for max batch size for connectors

andrewmcgivery:
Add support for max batch size for connectors

IGNORE ME, NOT TESTED!

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-federation/src/sources/connect/spec/directives.rs
+++ b/apollo-federation/src/sources/connect/spec/directives.rs
@@ -15,6 +15,7 @@ use super::schema::HEADERS_ARGUMENT_NAME;
 use super::schema::HTTP_ARGUMENT_NAME;
 use super::schema::SOURCE_BASE_URL_ARGUMENT_NAME;
 use super::schema::SOURCE_NAME_ARGUMENT_NAME;
+use super::schema::SourceBatchArguments;
 use super::schema::SourceDirectiveArguments;
 use super::schema::SourceHTTPArguments;
 use super::versions::VersionInfo;
@@ -234,6 +235,7 @@ impl ConnectDirectiveArguments {
         let mut http = None;
         let mut selection = None;
         let mut entity = None;
+        let mut batch = None;
         for arg in args {
             let arg_name = arg.name.as_str();
 
@@ -249,6 +251,12 @@ impl ConnectDirectiveArguments {
                 ))?;
 
                 http = Some(ConnectHTTPArguments::from_values(http_value, version_info)?);
+            } else if arg_name == "batch" {
+                let http_value = arg.value.as_object().ok_or(internal!(
+                    "`http` field in `@connect` directive is not an object"
+                ))?;
+
+                batch = Some(SourceBatchArguments::from_values(http_value)?);
             } else if arg_name == CONNECT_SELECTION_ARGUMENT_NAME.as_str() {
                 let selection_value = arg.value.as_str().ok_or(internal!(
                     "`selection` field in `@connect` directive is not a string"
@@ -274,6 +282,7 @@ impl ConnectDirectiveArguments {
             http,
             selection: selection.ok_or(internal!("`@connect` directive is missing a selection"))?,
             entity: entity.unwrap_or_default(),
+            batch,
         })
     }
 }
@@ -338,6 +347,28 @@ impl ConnectHTTPArguments {
             body,
             headers: headers.unwrap_or_default(),
         })
+    }
+}
+
+impl SourceBatchArguments {
+    fn from_values(values: &ObjectNode) -> Result<Self, FederationError> {
+        let mut max_size = None;
+        for (name, value) in values {
+            let name = name.as_str();
+
+            if name == "maxSize" {
+                let max_size_int = Some(value.to_i32().ok_or(internal!(
+                    "supplied 'max_size' field in `@connect` directive's `batch` field is not a positive integer"
+                ))?);
+                // Convert the int to a usize since it is used for chunking an array later.
+                // Much better to fail here than at run time.
+                max_size = max_size_int.map(|i| usize::try_from(i).map_err(|_| internal!(
+                    "supplied 'max_size' field in `@connect` directive's `batch` field is not a positive integer"
+                ))).transpose()?;
+            }
+        }
+
+        Ok(Self { max_size })
     }
 }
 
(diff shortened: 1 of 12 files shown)


apollo-federation/src/sources/connect/mod.rs
```diff
@@ -8,7 +8,7 @@ pub mod expand;
 mod header;
 mod id;
 mod json_selection;
-mod models;
+pub mod models;
```
lennyburdette, line 11: let's selectively expose types as pub instead of marking the whole module as pub if possible

apollo-federation/src/sources/connect/spec/directives.rs
```diff
@@ -341,6 +350,28 @@ impl ConnectHTTPArguments {
...
+        let mut max_size = None;
+        for (name, value) in values {
+            let name = name.as_str();
+
+            if name == "maxSize" {
+                let max_size_int = Some(value.to_i32().ok_or(internal!(
+                    "supplied 'max_size' field in `@connect` directive's `batch` field is not a positive integer"
+                ))?);
+                // Convert the int to a usize since it is used for chunking an array later.
+                // Much better to fail here than at run time.
```
lennyburdette, line 364: this code is part of the runtime!

apollo-federation/src/sources/connect/spec/schema.rs
```diff
@@ -55,6 +57,15 @@ pub(crate) struct SourceHTTPArguments {
     pub(crate) headers: IndexMap<HeaderName, HeaderSource>,
 }
 
+/// Settings for the connector when it is doing a $batch entity resolver
+#[cfg_attr(test, derive(Debug))]
+pub(crate) struct SourceBatchArguments {
```
lennyburdette, line 62: why is it called _Source_ BatchArguments?

apollo-router/src/plugins/connectors/handle_responses.rs
```diff
@@ -326,11 +326,21 @@ impl MappedResponse {
...
-                    let entities = key_values
+                    let new_entities = key_values
                         .map(|key| map.remove(&key).unwrap_or(Value::Null))
                         .collect_vec();
 
-                    data.insert(ENTITIES, Value::Array(entities));
+                    // Because we may have multiple batch entities requests, we should add to ENTITIES as the requests come in so it is additive
+                    let entities = data
+                        .entry(ENTITIES)
+                        .or_insert(Value::Array(Vec::with_capacity(count)));
```
lennyburdette, line 336: this count is for the current chunk. i dunno if there's a way to get the original count of entity references here?

apollo-router/src/plugins/connectors/make_requests.rs
```diff
@@ -761,29 +761,44 @@ fn batch_entities_from_request(
...
+                .as_object()
+                .ok_or_else(|| InvalidRepresentations("representation is not an object".into()))?
+                .clone();
+            Ok::<_, MakeRequestError>(obj)
+        })
+        .collect::<Result<Vec<_>, _>>()?;
+
+    // If we've got a max_size set, chunk the batch into smaller batches. Otherwise, we'll default to just a single batch.
+    let max_size = connector.batch_settings.as_ref().and_then(|bs| bs.max_size);
+    let batches = max_size.map_or(vec![batch.clone()], |size| {
```
lennyburdette, line 780: `vec![batch.clone()]` seems like an unnecessary clone, but i can't solve it without rust-analyzer :grin:

apollo-router/src/plugins/connectors/make_requests.rs
```diff
@@ -761,29 +761,44 @@ fn batch_entities_from_request(
...
+        .map(|batch| {
+            let inputs = RequestInputs {
+                batch: batch.to_vec(),
+                ..Default::default()
+            };
+
+            ResponseKey::BatchEntity {
+                selection: selection.clone(),
+                inputs,
+                keys: keys.clone(),
```
lennyburdette, line 796: not for this PR, but we should rename this to `key_field_set` or something

lennyburdette (commented):
i think this warrants an integration test next to the existing batch test (which will require a hand-edited supergraph until we release composition changes). nice work tho!

apollo-federation/src/sources/connect/spec/schema.rs
andrewmcgivery: 🤦 Good call lol

apollo-router/src/plugins/connectors/handle_responses.rs
andrewmcgivery: So... it seems like it's the count of....

```
let count = responses.len();
```

Which is calculated prior to looping through the responses. 😕 Weird.

apollo-router/src/plugins/connectors/make_requests.rs
andrewmcgivery: It seems unnecessary but it complains that it can't be owned by both that line and the closure right below it.... making it borrowed causes all kinds of other errors 😅 I couldn't figure out any alternatives but open to suggestions!

(I tried asking ChatGPT and it said "If you must return owned Vec<T>, then cloning or moving is necessary. 😅 )

apollo-router/src/plugins/connectors/make_requests.rs
andrewmcgivery: In just this function or on `request` and `BatchEntity` too?

lennyburdette: yeah, everywhere. it's a very confusing name

apollo-router/src/plugins/connectors/make_requests.rs
lennyburdette:
```rs
    let batches = if let Some(size) = n {
        v.chunks(size).map(|v| v.to_vec()).collect()  
    } else {
        vec![v]
    };
```

i guess the borrow checker can't know that the closures passed to `map_or` are mutually exclusive. but an `if/else` definitely is~

apollo-federation/src/sources/connect/mod.rs
andrewmcgivery: Done!

apollo-router/src/plugins/connectors/make_requests.rs
andrewmcgivery: Well I suppose I could do it that way! That's actually what I originally started with and it didn't feel rust-y so I re-factored to what I have now 😆

lennyburdette (approved):

PR #7280, opened 2025-04-16
Simplify implementation of telementry's events

IvanGoncharov:
I made a few attempts to make minimal fixes based on #7231. However, it was extremely hard to cover all edge cases, so I decided to do a separate refactoring PR first.

The core issue is that currently, some `on_request`/`on_response` functions exit earlier:
https://github.com/apollographql/router/blob/3ae5f3f2e02464f6208f078a6611ec9f2b002ba8/apollo-router/src/plugins/telemetry/config_new/events.rs#L261-L264

And because of it, bypass `response` and custom events:
https://github.com/apollographql/router/blob/3ae5f3f2e02464f6208f078a6611ec9f2b002ba8/apollo-router/src/plugins/telemetry/config_new/events.rs#L273-L281

Simply removing returns and implementing logic using ifs also doesn't work:
https://github.com/apollographql/router/pull/7231/files#diff-fab80b4af0ee70850af9ac42bf7990d09912b65f75b39456c6e47ca817d4a0d2R262-R269

It results in the standard event without the condition being ignored.
Trying to address it with a more complicated condition would work, but it would require adding it to more than a dozen places and also doesn't guarantee that we don't break some other edge cases.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/plugins/telemetry/config_new/events.rs
+++ b/apollo-router/src/plugins/telemetry/config_new/events.rs
@@ -1158,6 +1158,7 @@ mod tests {
                 spec: ConnectSpec::V0_1,
                 request_variables: Default::default(),
                 response_variables: Default::default(),
+                batch_settings: None,
                 request_headers: Default::default(),
                 response_headers: Default::default(),
             };
@@ -1243,6 +1244,7 @@ mod tests {
                 spec: ConnectSpec::V0_1,
                 request_variables: Default::default(),
                 response_variables: Default::default(),
+                batch_settings: None,
                 request_headers: Default::default(),
                 response_headers: Default::default(),
             };
--- a/apollo-router/src/plugins/telemetry/fmt_layer.rs
+++ b/apollo-router/src/plugins/telemetry/fmt_layer.rs
@@ -842,6 +842,7 @@ connector:
                     spec: ConnectSpec::V0_1,
                     request_variables: Default::default(),
                     response_variables: Default::default(),
+                    batch_settings: None,
                     request_headers: Default::default(),
                     response_headers: Default::default(),
                 });
@@ -1196,6 +1197,7 @@ subgraph:
                     spec: ConnectSpec::V0_1,
                     request_variables: Default::default(),
                     response_variables: Default::default(),
+                    batch_settings: None,
                     request_headers: Default::default(),
                     response_headers: Default::default(),
                 });
--- a/.changesets/fix_curly_think_heap_glue.md
+++ b/.changesets/fix_curly_think_heap_glue.md
@@ -1,0 +1,5 @@
+### Helm: Correct default telemetry `resource` property in `ConfigMap` (copy #6105) ([Issue #6104](https://github.com/apollographql/router/issues/6104))
+
+The Helm chart was using an outdated value when emitting the `telemetry.exporters.metrics.common.resource.service.name` values.  This has been updated to use the correct (singular) version of `resource` (rather than the incorrect `resources` which was used earlier in 1.x's life-cycle).
+
+By [@vatsalpatel](https://github.com/vatsalpatel) in https://github.com/apollographql/router/pull/6105
--- a/.changesets/fix_renee_increase_compute_job_queue.md
+++ b/.changesets/fix_renee_increase_compute_job_queue.md
@@ -1,0 +1,12 @@
+### Increase compute job worker pool queue size ([PR #7205](https://github.com/apollographql/router/pull/7205))
+
+The compute job worker pool is used for CPU-bound tasks, like GraphQL parsing, validation, and query planning. When there are too many jobs to handle in parallel, jobs enter a queue.
+
+We previously set this queue size to 20 (per thread) somewhat arbitrarily. We got some signals that this may be too small.
+
+This patch increases the queue size to 1 000 jobs per thread. For reference, in older router versions before the introduction of the compute job worker pool, the equivalent queue size was *10 000*.
+
+The number is still a bit arbitrary, and subject to more changes in the future as we understand its effects better. Along with some other tweaks to job priorities we expect this to give better behaviour and reject fewer requests needlessly.
+
+
+By [@goto-bus-stop](https://github.com/goto-bus-stop) in https://github.com/apollographql/router/pull/7205
--- a/.changesets/fix_simon_jwt_metrics.md
+++ b/.changesets/fix_simon_jwt_metrics.md
@@ -1,0 +1,9 @@
+### Fix JWT metrics discrepancy ([PR #7258](https://github.com/apollographql/router/pull/7258))
+
+This fixes the `apollo.router.operations.authentication.jwt` counter metric to behave [as documented](https://www.apollographql.com/docs/graphos/routing/security/jwt#observability): emitted for every request that uses JWT, with the `authentication.jwt.failed` attribute set to true or false for failed or successful authentication.
+
+Previously, it was only used for failed authentication.
+
+The attribute-less and accidentally-differently-named `apollo.router.operations.jwt` counter was and is only emitted for successful authentication, but is deprecated now.
+
+By [@SimonSapin](https://github.com/SimonSapin) in https://github.com/apollographql/router/pull/7258
--- a/.github/workflows/update_apollo_protobuf.yaml
+++ b/.github/workflows/update_apollo_protobuf.yaml
@@ -29,3 +29,7 @@ jobs:
           title: 'chore: Update Apollo Protobuf'
           body: |
             This updates the copy of `reports.proto` which this repository relies on with the latest copy fetched via our public endpoint.
+
+            > [!IMPORTANT]
+            > This PR will be continuously force-pushed with the new `reports.proto` copy. If the update requires code changes, apply them
+            > in a separate PR. This PR will be automatically closed if it is no longer necessary.
--- a/.github/workflows/update_uplink_schema.yml
+++ b/.github/workflows/update_uplink_schema.yml
@@ -31,3 +31,7 @@ jobs:
           title: 'chore: Update Uplink schema'
           body: |
             This updates the copy of `uplink.graphql` which this repository relies on with the latest copy fetched via `rover graph introspect`.
+
+            > [!IMPORTANT]
+            > This PR will be continuously force-pushed with the new `uplink.graphql` copy. If the update requires code changes, apply them
+            > in a separate PR. This PR will be automatically closed if it is no longer necessary.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,51 @@
 
 This project adheres to [Semantic Versioning v2.0.0](https://semver.org/spec/v2.0.0.html).
 
+# [2.1.3] - 2025-04-16
+
+## 🐛 Fixes
+
+### Entity-cache: handle multiple key directives ([PR #7228](https://github.com/apollographql/router/pull/7228))
+
+This PR fixes a bug in entity caching introduced by the fix in https://github.com/apollographql/router/pull/6888 for cases where several `@key` directives with different fields were declared on a type as documented [here](https://www.apollographql.com/docs/graphos/schema-design/federated-schemas/reference/directives#managing-types).
+
+For example if you have this kind of entity in your schema:
+
+```graphql
+type Product @key(fields: "upc") @key(fields: "sku") {
+  upc: ID!
+  sku: ID!
+  name: String
+}
+```
+
+By [@duckki](https://github.com/duckki) & [@bnjjj](https://github.com/bnjjj) in https://github.com/apollographql/router/pull/7228
+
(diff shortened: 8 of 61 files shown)


apollo-router/src/plugins/telemetry/config_new/connector/events.rs
```diff
@@ -46,92 +58,67 @@ pub(crate) fn new_connector_events(
...
-        ConnectorAttributes,
-        ConnectorSelector,
-    >
-{
-    type Request = ConnectorRequest;
-    type Response = ConnectorResponse;
-    type EventResponse = ();
-
-    fn on_request(&self, request: &Self::Request) {
+impl CustomEvents<ConnectorRequest, ConnectorResponse, (), ConnectorAttributes, ConnectorSelector> {
```
IvanGoncharov, line 72: I removed the `Instrumented` trait from here, and it worked without any other code modifications.
Doing this allowed me to change `on_request` so it now gets a mutable reference.

apollo-router/src/plugins/telemetry/config_new/connector/events.rs
```diff
@@ -46,92 +58,67 @@ pub(crate) fn new_connector_events(
...
         // Any condition on the request is NOT evaluated here. It must be evaluated later when
         // getting the ConnectorEventRequest from the context. The request context is shared
         // between all connector requests, so any request could find this ConnectorEventRequest in
         // the context. Its presence on the context cannot be conditional on an individual request.
-        if self.request.level() != EventLevel::Off {
-            request
-                .context
-                .extensions()
-                .with_lock(|lock| lock.insert(ConnectorEventRequest(self.request.clone())));
+        if let Some(request_event) = self.request.take() {
```
IvanGoncharov, line 78: I use `Option::take` so event is basically moved into context and we don't need to call clone here.

apollo-router/src/plugins/telemetry/config_new/connector/events.rs
```diff
@@ -46,92 +58,67 @@ pub(crate) fn new_connector_events(
...
+            request.context.extensions().with_lock(|lock| {
+                lock.insert(ConnectorEventRequest {
+                    level: request_event.level,
+                    condition: Arc::new(Mutex::new(request_event.condition)),
```
IvanGoncharov, line 82: As noted earlier I don't think we really need this wrapping, but I don't have time to check so I move it here just to be 100% sure I'm not introducing regression here.

apollo-router/src/plugins/telemetry/config_new/connector/events.rs
```diff
@@ -46,92 +58,67 @@ pub(crate) fn new_connector_events(
...
+            request.context.extensions().with_lock(|lock| {
+                lock.insert(ConnectorEventResponse {
+                    level: response_event.level,
+                    condition: Arc::new(response_event.condition),
+                })
+            });
         }
 
-        for custom_event in &self.custom {
+        for custom_event in &mut self.custom {
```
IvanGoncharov, line 96: Everything is mutable now because I removed the internal wrapping of Arc + Mutex around the condition.

apollo-router/src/plugins/telemetry/config_new/events.rs
```diff
@@ -244,49 +195,39 @@ where
     Attributes: Selectors<Request, Response, EventResponse> + Default,
     Sel: Selector<Request = Request, Response = Response> + Debug,
 {
-    pub(super) request: StandardEvent<Sel>,
-    pub(super) response: StandardEvent<Sel>,
-    pub(super) error: StandardEvent<Sel>,
+    pub(super) request: Option<StandardEvent<Sel>>,
+    pub(super) response: Option<StandardEvent<Sel>>,
+    pub(super) error: Option<StandardEvent<Sel>>,
```
IvanGoncharov, line 200: I normalize data here, previously we always had these values but internally they could have `EventLevel::Off` and this check was noisy.
So I removed `Off` from `EventLevel` and just set `None` instead.

apollo-router/src/plugins/telemetry/config_new/events.rs
```diff
@@ -244,49 +195,39 @@ where
...
     pub(super) custom: Vec<CustomEvent<Request, Response, EventResponse, Attributes, Sel>>,
```
IvanGoncharov, line 201: Also it removes useless code with custom events, because `custom` is filtered to not have events that are `Off` but because `EventLevel` contained this value it resulted in useless checks

apollo-router/src/plugins/telemetry/config_new/events.rs
```diff
@@ -244,49 +195,39 @@ where
...
-    type Response = router::Response;
-    type EventResponse = ();
-
-    fn on_request(&self, request: &Self::Request) {
-        if self.request.level() != EventLevel::Off {
-            if let Some(condition) = self.request.condition() {
-                if condition.lock().evaluate_request(request) != Some(true) {
-                    return;
-                }
+impl CustomEvents<router::Request, router::Response, (), RouterAttributes, RouterSelector> {
```
IvanGoncharov, line 204: Also removed Instrumented here and it just required me to add `pub(crate)` to methods

apollo-router/src/plugins/telemetry/config_new/events.rs
```diff
@@ -244,49 +195,39 @@ where
...
-    fn on_response(&self, response: &Self::Response) {
-        if self.response.level() != EventLevel::Off {
-            if let Some(condition) = self.response.condition() {
-                if !condition.lock().evaluate_response(response) {
-                    return;
-                }
+    pub(crate) fn on_response(&mut self, response: &router::Response) {
+        if let Some(response_event) = &mut self.response {
+            if !response_event.condition.evaluate_response(response) {
+                return;
```
IvanGoncharov, line 230: note: this PR doesn't fix early exit issues I talked about in PR description.
I will fix them in separate PR

apollo-router/src/plugins/telemetry/config_new/events.rs
```diff
@@ -575,66 +513,79 @@ struct SubgraphEventsConfig {
...
-    pub(crate) fn condition(&self) -> Option<&Arc<Mutex<Condition<T>>>> {
-        match self {
-            Self::Level(_) => None,
-            Self::Conditional { condition, .. } => Some(condition),
+impl<T: Clone> StandardEvent<T> {
+    pub(crate) fn from_config(config: &StandardEventConfig<T>) -> Option<Self> {
+        match &config {
+            StandardEventConfig::Level(level) => EventLevel::from_config(level).map(|level| Self {
+                level,
+                condition: Condition::True,
```
IvanGoncharov, line 550: Make it so the condition is always present and true by default.
Since now we have condition executed without Mutex for the most part it means cost of this simplification is pretty low.

apollo-router/src/plugins/telemetry/config_new/events.rs
```diff
@@ -575,66 +513,79 @@ struct SubgraphEventsConfig {
...
     #[default]
     Off,
 }
 
+#[derive(Debug, PartialEq, Clone, Copy)]
+pub(crate) enum EventLevel {
+    Info,
+    Warn,
+    Error,
+}
```
IvanGoncharov, line 576: Main purpose of adding it is to remove `Off`

apollo-router/src/plugins/telemetry/config_new/events.rs
```diff
@@ -690,14 +641,6 @@ pub(crate) enum EventOn {
...
 
 pub(crate) struct CustomEvent<Request, Response, EventResponse, A, T>
-where
-    A: Selectors<Request, Response, EventResponse> + Default,
-    T: Selector<Request = Request, Response = Response> + Debug,
-{
-    pub(super) inner: Mutex<CustomEventInner<Request, Response, EventResponse, A, T>>,
-}
-
-pub(super) struct CustomEventInner<Request, Response, EventResponse, A, T>
```
IvanGoncharov, line 700: I removed "inner" since we no longer need Mutex, we maintain a single copy and mutate it directly.

apollo-router/src/plugins/telemetry/fmt_layer.rs
```diff
@@ -727,7 +726,7 @@ connector:
 
                 error!(http.method = "GET", "Hello from test");
 
-                let router_events = event_config.new_router_events();
+                let mut router_events = event_config.new_router_events();
```
IvanGoncharov, line 729: Everything became explicitly mutable.
Previously it also was mutable but hidden under internal Arc + Mutex

apollo-router/tests/integration/telemetry/fixtures/trace_id_via_header.router.yaml
```diff
@@ -1,9 +1,6 @@
 telemetry:
-
```
IvanGoncharov, line 2: Was using this test as a basis for my new test and cleaned up some stuff in a process

apollo-router/tests/integration/telemetry/propagation.rs
```diff
@@ -13,11 +11,13 @@ async fn test_trace_id_via_header() -> Result<(), BoxError> {
         return Ok(());
     }
     async fn make_call(router: &mut IntegrationTest, trace_id: &str) {
-        let _ = router.execute_query(Query::builder().body(json!({"query":"query {topProducts{name, name, name, name, name, name, name, name, name, name}}","variables":{}})).header("id_from_header".to_string(), trace_id.to_string()).build()).await;
+        let query = Query::builder()
+            .header("id_from_header".to_string(), trace_id.to_string())
+            .build();
```
IvanGoncharov, line 16: we send default query here, just attach custom header to it

apollo-router/src/plugins/telemetry/config_new/connector/events.rs
```diff
@@ -6,25 +8,35 @@ use serde::Deserialize;
...
 use crate::plugins::telemetry::config_new::extendable::Extendable;
-use crate::plugins::telemetry::config_new::instruments::Instrumented;
 
 #[derive(Clone)]
-pub(crate) struct ConnectorEventRequest(pub(crate) StandardEvent<ConnectorSelector>);
+pub(crate) struct ConnectorEventRequest {
+    // XXX(@IvanGoncharov): As part of removing Mutex from StandardEvent I moved it here
+    // I think it's not nessary here but can't verify it right now, so in future can just wrap StandardEvent
+    pub(crate) level: EventLevel,
+    pub(crate) condition: Arc<Mutex<Condition<ConnectorSelector>>>,
```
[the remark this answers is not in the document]
bnjjj: Is condition still optional ?

apollo-router/src/plugins/telemetry/config_new/events.rs
bnjjj: Ok I get the answer of my previous question about conditions not being optional

BrynCooke (approved):
Nice! It's great to have events standardized

apollo-router/src/plugins/telemetry/config_new/connector/events.rs
```diff
@@ -46,92 +58,67 @@ pub(crate) fn new_connector_events(
...
-                    event_on: event_cfg.on,
-                    message: event_cfg.message.clone(),
-                    selectors: event_cfg.attributes.clone().into(),
-                    condition: event_cfg.condition.clone(),
-                    attributes: Vec::new(),
-                    _phantom: Default::default(),
-                }),
-            }),
-        })
+        .filter_map(|(name, config)| CustomEvent::from_config(name, config))
```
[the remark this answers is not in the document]
IvanGoncharov: TryFrom returns Result: https://doc.rust-lang.org/std/convert/trait.TryFrom.html
So it will require us to define some dummy error type.

apollo-router/src/plugins/telemetry/config_new/connector/events.rs
```diff
@@ -46,92 +58,67 @@ pub(crate) fn new_connector_events(
...
     }
 
-    fn on_error(&self, error: &BoxError, ctx: &Context) {
-        if self.error.level() != EventLevel::Off {
-            if let Some(condition) = self.error.condition() {
-                if !condition.lock().evaluate_error(error, ctx) {
-                    return;
-                }
+    pub(crate) fn on_error(&mut self, error: &BoxError, ctx: &Context) {
+        if let Some(error_event) = &mut self.error {
```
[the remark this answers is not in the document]
IvanGoncharov: Technically, we can, if we sure it's the last call (which it probably is).
But I use `take` more as a hack to move event outside of this plugin.
In this case error event is handled inside so `&mut` is enough.

bnjjj (approved):

IvanGoncharov:
@mergify backport 1.x

PR #7295, opened 2025-04-16
Split `apollo-router/src/plugins/telemetry/config_new` code into separate modules

Velfi:


BrynCooke:
@Velfi Moving these files updates the json schema for config, you'll need to update the snapshot.

abernix:
Is this the kind of thing worth backporting to 1.x just to keep "backport-ability" clean?  If so, I think applying the `backport-1.x` label on this now makes sense.

_(Separately, just want to make it clear that I think it's fine to clear the PR template if you don't think that the default steps aren't relevant, or not necessary for your PR. [e.g. if they're not fixing something, remove the `Fixes #issue_number`])_

pubmodmatt:
The connectors ones were already in a [separate file](https://github.com/apollographql/router/blob/f568df79da8e51de3f46ac3d267599448f20f001/apollo-router/src/plugins/telemetry/config_new/connector/selectors.rs#L56) (`ConnectorSelector` is still one of my favorite struct names), but you may consider moving them based on this PR

apollo-router/src/services/layers/persisted_queries/snapshots/apollo_router__services__layers__persisted_queries__tests__pq_layer_freeform_graphql_with_safelist_log_unknown_true@logs.snap
```diff
@@ -2,6 +2,9 @@
 source: apollo-router/src/services/layers/persisted_queries/mod.rs
 expression: yaml
 ---
+- fields: {}
+  level: INFO
+  message: Loaded 2 persisted queries.
```
IvanGoncharov, line 7: @Velfi, is this change related to a PR?
Or it is just an artifact of running tests locally?

IvanGoncharov (changes requested):

Velfi:
> The connectors ones were already in a [separate file](https://github.com/apollographql/router/blob/f568df79da8e51de3f46ac3d267599448f20f001/apollo-router/src/plugins/telemetry/config_new/connector/selectors.rs#L56) (`ConnectorSelector` is still one of my favorite enum names), but you may consider moving them based on this PR

Based on this suggestion I've:
- adopted a different structure for the change
- moved more stuff into submodules

Velfi:
I started a cleaner, easier to review, version of this PR. The changes are broken into separate commits so that you can review each one individually.
https://github.com/apollographql/router/pull/7322

PR #7297, opened 2025-04-16
Add support for connectors `path` and `queryParams`

dylan-apollo:
Implements the `path` and `queryParams` parameters to `http` in both `@source` and `@connect`.

[CNN-738]: https://apollographql.atlassian.net/browse/CNN-738?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apollo-federation/src/sources/connect/spec/directives.rs
```diff
@@ -206,6 +210,18 @@ impl SourceHTTPArguments {
                         .try_collect()
                         .map_err(|err| internal!(err.to_string()))?,
                 );
+            } else if name == PATH_ARGUMENT_NAME.as_str() {
+                let value = value.as_str().ok_or(internal!(format!(
+                    "`{}` field in `@connect` directive's `http` field is not a string",
```
andrewmcgivery, line 215: I assume this should be `@source` instead if `@connect`? :)

apollo-federation/src/sources/connect/spec/directives.rs
```diff
@@ -206,6 +210,18 @@ impl SourceHTTPArguments {
...
+                    PATH_ARGUMENT_NAME
+                )))?;
+                path = Some(JSONSelection::parse(value).map_err(|e| internal!(e.message))?);
+            } else if name == QUERY_PARAMS_ARGUMENT_NAME.as_str() {
+                let value = value.as_str().ok_or(internal!(format!(
+                    "`{}` field in `@connect` directive's `http` field is not a string",
```
andrewmcgivery, line 221: I assume this should be `@source` instead if `@connect`? :)

apollo-federation/src/sources/connect/spec/type_and_directive_specifications.rs
```diff
@@ -168,6 +170,25 @@ pub(super) fn check_or_add(
         },
     ];
 
+    if spec >= &ConnectSpec::V0_2 {
+        connect_http_field_list.extend([
```
andrewmcgivery, line 174: Lenny's original PR had a `ORIGIN_ARGUMENT_NAME` argument... I don't think I have enough background to question if we omitted it in this PR on purpose but I figured I'd throw it out as a difference and just make sure it's expected!

andrewmcgivery (commented):
I noticed also in Lenny's PR that he has a change to connector_service:

```
let method = transport.method_attr();
let url = transport.url_attr();
if let Ok(detail) = serde_json::to_string(&serde_json::json!({ method: url })) {
   span.record("apollo.connector.detail", detail);
}
```

Do we need that in this PR too or is that just really outdated?

apollo-federation/src/sources/connect/spec/type_and_directive_specifications.rs
dylan-apollo: Yup, we dropped it. Good find, though!

dylan-apollo:
> Do we need that in this PR too or is that just really outdated?

We had previously an extra way to set method which required some more hoops there. We dropped that for now, though.

andrewmcgivery (approved):

PR #7299, opened 2025-04-16
Refactor: Move HTTP transport models

dylan-apollo:
This just cuts and pastes a few models into their own module. We're going to add a lot more logic to these in a future PR.


lennyburdette (approved):

apollo-federation/src/sources/connect/models/http_json_transport.rs
```diff
@@ -0,0 +1,321 @@
...
+        } else if let Some(url) = &http.put {
+            (HTTPMethod::Put, url)
+        } else if let Some(url) = &http.delete {
+            (HTTPMethod::Delete, url)
+        } else {
+            return Err(FederationError::internal("missing http method"));
+        };
+
+        #[allow(clippy::mutable_key_type)]
+        // HeaderName is internally mutable, but we don't mutate it
```
naomijub, line 60: I know this was already here, but if we don't need mutability and we only need to preserve the headers from `http.headers`  as the source of truth, we can prevent the early headers cloning by doing:

```rus
let headers = source.map(|source| &source.headers)
            .into_iter()
            .flatten()
            .chain(http.headers.iter())
            .map(|(name, souce)| (name.clone(), souce.clone()))
            .collect::<IndexMap<HeaderName, HeaderSource>>();
```

apollo-federation/src/sources/connect/models/http_json_transport.rs
```diff
@@ -0,0 +1,321 @@
...
+                ))
+            })?,
+            method,
+            headers,
+            body: http.body.clone(),
+        })
+    }
+
+    pub(crate) fn label(&self) -> String {
+        format!("http: {} {}", self.method.as_str(), self.connect_template)
```
naomijub, line 85: `self.method` inside the format doesn't need the `as_str()` as it already implements that in `Display`:

```rust
impl Display for HTTPMethod {
    fn fmt(&self, f: &mut Formatter) -> std::fmt::Result {
        write!(f, "{}", self.as_str())
    }
}
```

naomijub (approved):

PR #7311, opened 2025-04-17
Add top level error mapping for connectors

andrewmcgivery:
Add top level error mapping for connectors

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-federation/src/sources/connect/spec/type_and_directive_specifications.rs
```diff
@@ -229,6 +232,41 @@ pub(super) fn check_or_add(
...
+            ty: Type::Named(json_selection_spec.name.clone()).into(),
+            default_value: None,
+            directives: Default::default(),
+        },
+    ];
+
+    let mut connect_errors_fields = IndexMap::with_hasher(Default::default());
+    for field in connect_errors_field_list {
+        connect_errors_fields.insert(field.name.clone(), Component::new(field));
+    }
```
dylan-apollo, line 257: We can definitely write this without making a vec and then an IndexMap. Here's one idea:

```suggestion
    let connect_errors_fields = [
        ERRORS_MESSAGE_ARGUMENT_NAME,
        ERRORS_EXTENSIONS_ARGUMENT_NAME,
    ]
    .into_iter()
    .map(|name| {
        (
            name.clone(),
            Component::new(InputValueDefinition {
                description: None,
                name,
                ty: Type::Named(json_selection_spec.name.clone()).into(),
                default_value: None,
                directives: Default::default(),
            }),
        )
    })
    .collect();
```

Though I wonder why we don't add descriptions to these...

apollo-federation/src/sources/connect/spec/type_and_directive_specifications.rs
```diff
@@ -381,6 +436,41 @@ pub(super) fn check_or_add(
         type_name: source_http_spec.name.clone(),
     };
 
+    // @source error settings
+
+    let source_errors_field_list = vec![
```
dylan-apollo, line 441: Same thing here, so maybe extract a single function `errors() -> DirectiveArgumentSpecification` ?

apollo-federation/src/sources/connect/validation/connect.rs
```diff
@@ -210,6 +215,7 @@ impl<'schema> Connect<'schema> {
             .selection
             .variables()
             .chain(self.http.variables())
+            .chain(self.errors.variables())
             .collect::<HashSet<_>>();
         if all_variables.contains(&Namespace::Batch) && all_variables.contains(&Namespace::This) {
```
dylan-apollo, line 220: This makes me think... what actually _happens_ when you set a custom error with batch? Wouldn't you end up with arrays in extensions by accident?

We should make sure we have an example of the right way to use this and are validating correctly.

apollo-federation/src/sources/connect/validation/connect/errors.rs
```diff
@@ -0,0 +1,331 @@
...
+    ///
+    /// The order these pieces run in doesn't matter and shouldn't affect the output.
+    pub(super) fn parse(
+        coordinate: ConnectDirectiveCoordinate<'schema>,
+        schema: &'schema SchemaInfo,
+    ) -> Result<Self, Vec<Message>> {
+        let Some((errors_arg, _errors_arg_node)) = coordinate
+            .directive
+            .specified_argument_by_name(&ERRORS_ARGUMENT_NAME)
+            .and_then(|arg| Some((arg.as_object()?, arg)))
```
dylan-apollo, line 48: If it's not an object, maybe we should return a nice, highlighted error here instead of letting JS deal with it?

apollo-federation/src/sources/connect/validation/connect/errors.rs
```diff
@@ -0,0 +1,331 @@
...
+        ErrorsMessage::parse(errors_arg, coordinate, schema)
+            .map_err(|err| vec![err])
+            .and_try(
+                ErrorsExtensions::parse(errors_arg, coordinate, schema).map_err(|err| vec![err]),
+            )
+            .map_err(|nested| nested.into_iter().flatten().collect())
+            .map(|(message, extensions)| Self {
+                message,
+                extensions,
+            })
```
dylan-apollo, line 65: Should be able to do this:

```suggestion
        ErrorsMessage::parse(errors_arg, coordinate, schema)
            .and_try(ErrorsExtensions::parse(errors_arg, coordinate, schema))
            .map(|(message, extensions)| Self {
                message,
                extensions,
            })
```

The only reason we do the Vec messiness elsewhere is because _some_ of the things we're trying return multiple errors, and some return a single error.

dylan-apollo (commented):
Only about halfway through the files, but I figured I'd submit so you can start reading & responding 😁

apollo-federation/src/sources/connect/models.rs
```diff
@@ -139,30 +180,72 @@ impl Connector {
...
+        let errors_extensions_headers = error_settings
+            .as_ref()
+            .map_or(Default::default(), |e| e.extensions.as_ref())
+            .map_or(Default::default(), |m| {
+                extract_header_references(m.variable_references())
+            });
+        let mut response_headers =
+            extract_header_references(connect.selection.variable_references());
+        response_headers.extend(errors_message_headers);
+        response_headers.extend(errors_extensions_headers);
```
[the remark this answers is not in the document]
andrewmcgivery: So to confirm... effectively the same logic but merging the gathering of `errors_message_variables`, `errors_extensions_variables`, and `response_variables` into a chain instead of 3 different HashSet's with the merge?

apollo-federation/src/sources/connect/models.rs
```diff
@@ -74,6 +78,43 @@ impl ConnectorBatchSettings {
...
+pub struct ConnectorErrorsSettings {
+    pub message: Option<JSONSelection>,
+    pub extensions: Option<JSONSelection>,
+}
+
+impl ConnectorErrorsSettings {
+    fn from_directive(
+        connect_errors: Option<&ConnectErrorsArguments>,
+        source_errors: Option<&Option<SourceErrorsArguments>>,
+    ) -> Option<Self> {
```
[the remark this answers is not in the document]
andrewmcgivery: I mostly did this to...

1. Reflect the fact that `erorrs` itself is optional
2. Make testing easier by being able to pass in `None` in the million places we create a raw `Connector` 😅 

However, I could probably also create a `ConnectorErrorsSettings { None, None }` instead that is preferred.

apollo-federation/src/sources/connect/spec/directives.rs
```diff
@@ -240,6 +255,39 @@ impl SourceHTTPArguments {
...
+impl SourceErrorsArguments {
+    fn from_values(values: &ObjectNode) -> Result<Self, FederationError> {
+        let mut message = None;
+        let mut extensions = None;
+        for (name, value) in values {
+            let name = name.as_str();
+
+            if name == ERRORS_MESSAGE_ARGUMENT_NAME.as_str() {
+                let message_value = value.as_str().ok_or(internal!(
+                    "`message` field in `@source` directive's `errors` field is not a string"
```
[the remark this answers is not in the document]
andrewmcgivery: So only concern here I would have is consistency... the `ERRORS_MESSAGE_ARGUMENT_NAME` part does seem to be done in some places (but not all lol) but the `directive_name` part I don't see done anywhere.

Do we keep it consistent, allow it to be different but better, or do a whole lot of updates in this file to make it all consistent? 😅 

(My preference would be to keep consistency and do follow up PRs to clean up it all up at the same time)

apollo-federation/src/sources/connect/spec/directives.rs
```diff
@@ -240,6 +255,39 @@ impl SourceHTTPArguments {
...
+                ))?;
+                message =
+                    Some(JSONSelection::parse(message_value).map_err(|e| internal!(e.message))?);
+            } else if name == ERRORS_EXTENSIONS_ARGUMENT_NAME.as_str() {
+                let extensions_value = value.as_str().ok_or(internal!(
+                    "`extensions` field in `@source` directive's `errors` field is not a string"
+                ))?;
+                extensions =
+                    Some(JSONSelection::parse(extensions_value).map_err(|e| internal!(e.message))?);
```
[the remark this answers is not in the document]
andrewmcgivery: Similar thing here... this is consistent with everything else in this file so if we do want to change this, I'd prefer to do it in a follow on PR and fix it all so it's consistent :D

apollo-federation/src/sources/connect/spec/type_and_directive_specifications.rs
andrewmcgivery: So this code I made identical to what was done previously (See: http_header_mapping_field_list, connect_http_field_list).

Similar to previous comments... I want to be consistent so if we want to update all of these, I'd prefer to do in a follow on PR instead of having my code be a special snowflake. ❄️ 😆

apollo-federation/src/sources/connect/spec/type_and_directive_specifications.rs
andrewmcgivery: Same response as above. :D

apollo-federation/src/sources/connect/validation/connect/errors.rs
andrewmcgivery: Good call. Updated!

apollo-federation/src/sources/connect/validation/connect.rs
andrewmcgivery: Yea, this feels like a documentation problem for sure to highlight that you should have separate error config on `@connect` on `$batch` resolvers.

You wouldn't even end up with an array in extensions... you'd get the default extensions because it would silently fail to convert it to an object.

apollo-federation/src/sources/connect/models.rs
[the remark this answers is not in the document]
dylan-apollo: derive(Default)

apollo-federation/src/sources/connect/spec/directives.rs
```diff
@@ -240,6 +255,39 @@ impl SourceHTTPArguments {
     }
 }
 
+impl SourceErrorsArguments {
```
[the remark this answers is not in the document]
dylan-apollo: Also the validations can probably be a single module/type using an enum to differentiate between source and connect for expression context (see how we validate headers)

andrewmcgivery:
There are a couple of cases I want to talk to @lennyburdette about when he is back (hopefully as a fast follower PR to this PR) around mapping on endpoints that return non-JSON (E.g. blank or HTML).

Currently these don't respect the mapping because they error earlier in the code and a graphql Error type is propagated down. I _could_ try to grab this GraphQL error as it goes by and manipulate it based on the mapping, but that may have other unintended side effects.

Want to discuss further before making a decision there!

apollo-federation/src/sources/connect/models.rs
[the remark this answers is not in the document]
andrewmcgivery: Done!

apollo-federation/src/sources/connect/spec/directives.rs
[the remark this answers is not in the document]
andrewmcgivery: Done!

apollo-federation/src/sources/connect/validation/connect/errors.rs
andrewmcgivery: Done!

apollo-federation/src/sources/connect/validation/errors.rs
```diff
@@ -0,0 +1,370 @@
...
+        Ok(Some(Self {
+            selection,
+            string,
+            coordinate,
+        }))
+    }
+
+    /// Check that the selection of the body matches the inputs at this location.
+    ///
+    /// TODO: check keys here?
```
dylan-apollo, line 314:
```suggestion
    /// Check that the selection only uses allowed variables and evaluates to an object
```

apollo-federation/src/sources/connect/validation/errors.rs
```diff
@@ -0,0 +1,370 @@
...
+}
+
+struct ErrorsExtensions<'schema> {
+    selection: JSONSelection,
+    string: GraphQLString<'schema>,
+    coordinate: ErrorsExtensionsCoordinate<'schema>,
+}
+
+impl<'schema> ErrorsExtensions<'schema> {
+    pub(super) fn parse(
```
dylan-apollo, line 254: This whole function is almost identical to the message version—and probably identical to the same for body/selection.

Makes me pretty sure we need to have a `json_selection_from_arg` helper of some sort that keeps all of them consistent in their expectations of what a "valid mapping expression" looks like.

Will also need that for `path` and `queryParams`.

Doesn't have to be here, but could be a very useful follow-up refactor before we implement validations for ^ 😁

apollo-federation/src/sources/connect/validation/errors.rs
```diff
@@ -0,0 +1,370 @@
...
+            }
+        };
+
+        expression::validate(
+            &Expression {
+                expression: selection,
+                location: 0..string.as_str().len(),
+            },
+            context,
+            &Shape::empty_object([]),
```
dylan-apollo, line 336: I'm actually surprised that this works... I guess an object with properties satisfies an empty object? But I would expect us to be able to type-check that something is _explicitly_ empty in the future if we want to. I think the values here can be _anything_, so maybe this would be more accurate?

```suggestion
            &Shape::dict(Shape::unknown([]), []),
```

apollo-federation/src/sources/connect/validation/expression.rs
```diff
@@ -79,6 +92,66 @@ impl<'schema> Context<'schema> {
...
+                    (Namespace::Context, Shape::unknown([])),
+                    (Namespace::Status, Shape::int([])),
+                    (Namespace::Request, REQUEST_SHAPE.clone()),
+                    (Namespace::Response, RESPONSE_SHAPE.clone()),
+                ]
+                .into_iter()
+                .collect();
+
+                if matches!(parent_category, ObjectCategory::Other) {
+                    var_lookup.insert(Namespace::This, Shape::from(parent_type));
```
dylan-apollo, line 146: And batch? Or are we not allowing that yet until we figure out the user story?

apollo-federation/src/sources/connect/validation/mod.rs
```diff
@@ -231,6 +232,10 @@ pub enum Code {
     InvalidSelection,
     /// The `http.body` provided in `@connect` was not valid.
     InvalidBody,
+    /// The `errors.message` provided in `@connect` was not valid.
+    InvalidErrorsMessage,
+    /// The `errors.extensions` provided in `@connect` was not valid.
+    InvalidErrorsExtensions,
```
dylan-apollo, line 238:
```suggestion
    /// The `errors.message` provided in `@connect` or `@source` was not valid.
    InvalidErrorsMessage,
    /// The `errors.extensions` provided in `@connect` or `@source` was not valid.
    InvalidErrorsExtensions,
```

apollo-federation/src/sources/connect/validation/test_data/errors.graphql
```diff
@@ -0,0 +1,58 @@
...
+    name: "valid_source"
+    http: {
+      baseURL: "http://127.0.0.1"
+      errors: { message: "error.message", extensions: "code: error.code status: $status" }
+    }
+  )
+  @source(
+    name: "invalid_source_message_not_string"
+    http: { baseURL: "http://127.0.0.1" }
+    errors: { message: "message: error.message", extensions: "code: error.code" }
```
dylan-apollo, line 13: Add tests for known shapes as well (`$args` or `$this` in connect) and disallowed variables (those same ones here)

apollo-federation/src/sources/connect/validation/errors.rs
```diff
@@ -0,0 +1,370 @@
...
+            }])
+        }
+    }
+
+    /// Type-check the `@connect(errors:)` or `@source(errors:)` directive.
+    ///
+    /// Does things like ensuring that every accessed variable actually exists and that expressions
+    /// used in the URL and headers result in scalars.
+    ///
+    /// TODO: Return some type checking results, like extracted keys?
```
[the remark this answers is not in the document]
andrewmcgivery: 🤷 I think I'd need other folks's opinions on that lol

apollo-federation/src/sources/connect/validation/errors.rs
```diff
@@ -0,0 +1,370 @@
...
+                        .into_iter()
+                        .collect(),
+                });
+            }
+        };
+        let selection = match JSONSelection::parse(string.as_str()) {
+            Ok(selection) => selection,
+            Err(err) => {
+                return Err(Message {
+                    code: Code::InvalidErrorsMessage,
```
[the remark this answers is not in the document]
andrewmcgivery: That makes sense... does that make sense as a backlog item in Jira? I'm not sure fixing all _our_ error handling is in scope for my PR 😆

apollo-federation/src/sources/connect/validation/errors.rs
```diff
@@ -0,0 +1,370 @@
...
+        if selection.is_empty() {
+            return Err(Message {
+                code: Code::InvalidErrorsMessage,
+                message: format!("{coordinate} is empty"),
+                locations: value
+                    .line_column_range(&schema.sources)
+                    .into_iter()
+                    .collect(),
+            });
+        }
```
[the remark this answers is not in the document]
andrewmcgivery: I did this largely because we do the same check in the other validation modules... if you think it's truly not needed I can take it out, but I assumed it was since we're doing it in other places 😅

apollo-federation/src/sources/connect/validation/errors.rs
andrewmcgivery: Jira item perhaps? :D

apollo-federation/src/sources/connect/validation/expression.rs
andrewmcgivery: I'm... not sure. We also don't have it in `for_connect_request`.. should it be?

apollo-federation/src/sources/connect/validation/coordinates.rs
```diff
@@ -35,6 +35,23 @@ impl Display for ConnectDirectiveCoordinate<'_> {
...
+    pub(super) directive: &'a Node<Directive>,
+}
+
+impl Display for SourceDirectiveCoordinate<'_> {
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
+        let Self { directive } = self;
+        write!(
+            f,
+            "`@{connect_directive_name}`",
+            connect_directive_name = directive.name
```
dylan-apollo, line 50:
```suggestion
            "`@{directive_name}`",
            directive_name = directive.name
```

apollo-federation/src/sources/connect/validation/errors.rs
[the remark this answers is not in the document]
dylan-apollo: I think it'll be covered by https://apollographql.atlassian.net/browse/CNN-611 actually 👍

apollo-federation/src/sources/connect/validation/errors.rs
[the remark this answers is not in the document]
dylan-apollo: I think this was a workaround for weak type checking at some point. Pretty sure you can cut it, but we should have a test for the issue (empty string) regardless, to make sure composition provides a useful error.

apollo-federation/src/sources/connect/validation/test_data/errors.graphql
```diff
@@ -0,0 +1,88 @@
...
+      source: "valid_source"
+      http: { GET: "/" }
+      errors: { message: "error.message", extensions: "code: error.code status: $status" }
+      selection: "$"
+    )
+
+  valid_sourceless: [String!]!
+    @connect(
+      http: { GET: "http://127.0.0.1" }
+      errors: { message: "error.message", extensions: "code: error.code status: $status" }
```
dylan-apollo, line 57: Sorry to keep throwing edge cases at you, but can we get one more happy-path test where the shape will be completely known (doesn't use `$`) just to make sure those shape validations don't throw a fit? Something like

```
errors: { message: """$("An error!")""", extensions: "status: $status"
```

apollo-router/src/plugins/connectors/handle_responses.rs
```diff
@@ -158,25 +156,83 @@ impl RawResponse {
...
+                if let Some(message_selection) = &connector.error_settings.message {
+                    // TODO: In the future, we'll want to add to the debug context. However, we'll need a "v2" debug payload before we can do that.
+                    let (res, _apply_to_errors) = message_selection.apply_with_vars(&data, &inputs);
+
+                    message = res
+                        .unwrap_or_else(|| Value::Null)
+                        .as_str()
+                        .unwrap_or_default()
+                        .to_string();
+                }
```
dylan-apollo, line 183:
```suggestion
                let message = if let Some(message_selection) = &connector.error_settings.message {
                    // TODO: In the future, we'll want to add to the debug context. However, we'll need a "v2" debug payload before we can do that.
                    let (res, _apply_to_errors) = message_selection.apply_with_vars(&data, &inputs);

                    res
                        .unwrap_or_else(|| Value::Null)
                        .as_str()
                        .unwrap_or_default()
                        .to_string()
                } else {
                  "Request failed".to_string()
                }
```

Just to not allocate the default if we don't need it.

apollo-router/src/plugins/connectors/handle_responses.rs
```diff
@@ -158,25 +156,83 @@ impl RawResponse {
...
+
+                // If we have extensions calculated by the JSONSelection, we will need to grab the code + the remaining extensions and map them to the error object
+                // Otherwise, we'll use some defaults
+                // Note: that we set the extension code in this if/else but don't actually set it on the error until after the if/else. This is because the compiler
+                // can't make sense of it in the if/else due to how the builder is constructed.
+                let mut extension_code = "CONNECTOR_FETCH".to_string();
+                if let Some(extensions) = extensions {
+                    if let Some(code) = extensions.get("code") {
+                        extension_code = code.as_str().unwrap_or_default().to_string();
+                    }
```
dylan-apollo, line 208: It looks like the `.extension_code` builder actually handles this for you—the code will be set only if it wasn't already set (we should have a test for this, but that's my read of it).

apollo-router/src/plugins/connectors/handle_responses.rs
```diff
@@ -158,25 +156,83 @@ impl RawResponse {
...
+                                Value::Number(parts.status.as_u16().into()),
+                            )])),
+                        )
+                        .extension(
+                            "connector",
+                            Value::Object(Map::from_iter([(
+                                "coordinate".into(),
+                                Value::String(connector.id.coordinate().into()),
+                            )])),
+                        );
```
dylan-apollo, line 231: I feel like we should set these default ones if they aren't already set by the user. Especially if it helps compatibility with insights.

The trouble is, the user _could_ unset them explicitly if they want to with `extensions: "connector: $(null)"` but has no way to get them _back_ if they didn't want to unset them, just append new ones.

apollo-router/src/plugins/connectors/handle_responses.rs
dylan-apollo: And with that plus the other comment below, I _think_ we can instead just set these directly where we process the mapping and avoid a few branches. So something like:

1. `let mut error ...`
2. `if let Some(extensions_selection) { `... apply each to `error.extension()` }`
3. Apply the defaults... ideally with a value that lazily allocates on `.into()` and not preallocated

That last step I _think_ won't overwrite anything already inserted, though `buildstructor` makes it hard to tell. Either way we'll want a test for it!

apollo-router/src/plugins/connectors/handle_responses.rs
```diff
@@ -158,25 +156,83 @@ impl RawResponse {
...
-                            "coordinate".into(),
-                            Value::String(connector.id.coordinate().into()),
-                        )])),
-                    )
-                    .path::<Path>((&key).into())
+                let mut message = "Request failed".to_string();
+                let mut extensions = None;
+
+                // Do we have errors settings for this connector?
+                let inputs = key.inputs().merge(
```
dylan-apollo, line 163: We don't want to "pay" for this unless there's actually a custom mapping, since it's cloning potentially a ton of data and allocating a new map.

The simplest way is probably to put this under a `if connector.error_settings.message.is_some() || connector.error_settings.extensions.is_some()`, but that only _kinda_ conveys intent.

Better might be a `LazyCell`, but IDK how much you'll have to change later. Worth a shot!

```
let inputs = LazyCell::new(|| {
                    key.inputs().merge(
                        &connector.response_variables,
                        &connector.response_headers,
                        connector.config.as_ref(),
                        context,
                        Some(parts.status.as_u16()),
                        supergraph_request,
                        Some(&parts),
                    )
                });
```

apollo-router/src/plugins/connectors/tests/error_handling.rs
```diff
@@ -0,0 +1,401 @@
...
+        include_str!("../testdata/errors.graphql"),
+        &mock_server.uri(),
+        "query { only_source { id name username } }",
+        Default::default(),
+        Some(override_config.into()),
+        |_| {},
+    )
+    .await;
+
+    // TODO: Why is status not here?
```
dylan-apollo, line 56: ?

apollo-router/src/plugins/connectors/tests/error_handling.rs
```diff
@@ -0,0 +1,401 @@
...
+      "errors": [
+        {
+          "message": "Something blew up!",
+          "path": [
+            "only_source"
+          ],
+          "extensions": {
+            "status": 500,
+            "code": "BIG_BOOM",
+            "service": "connectors"
```
dylan-apollo, line 133: So we _are_ still getting service, just not `connector` or `http`? Why does that happen?

apollo-router/src/plugins/connectors/testdata/errors.yaml
```diff
@@ -0,0 +1,26 @@
...
+          @source(name: "withconfig", http: { baseURL: "http://localhost:4001/api" }, errors: { message: "error.message", extensions: "code: error.code status: $status" } )
+          @source(name: "withoutconfig", http: { baseURL: "http://localhost:4001/api" })
+
+        type Query {
+          only_source: [User!]!
+            @connect(source: "withconfig", http: { GET: "/users" }, selection: "id name username")
+          only_connect: [User!]!
+            @connect(source: "withoutconfig", http: { GET: "/users" }, errors: { message: "error.message", extensions: "code: error.code status: $status" }, selection: "id name username")
+          both_source_and_connect: [User!]!
+            @connect(source: "withconfig", http: { GET: "/users" }, errors: { message: "error.message", extensions: "code: error.code" }, selection: "id name username")
```
dylan-apollo, line 18: So the reason we know this one overrode source is because there's no status? Might make that more explicit somehow so no one breaks it without it being super obvious

apollo-router/src/plugins/connectors/tests/error_handling.rs
```diff
@@ -0,0 +1,401 @@
...
+        include_str!("../testdata/errors.graphql"),
+        &mock_server.uri(),
+        "query { both_source_and_connect { id name username } }",
+        Default::default(),
+        Some(override_config.into()),
+        |_| {},
+    )
+    .await;
+
+    // Note that status 500 is NOT included in extensions because connect is overriding source
```
dylan-apollo, line 383: Oh I see, the comment explains it! 😁

apollo-router/src/plugins/connectors/handle_responses.rs
andrewmcgivery: I did this branching because the design indicates that we don't want the defaults to apply if the user has set extensions mapping.

That said, I have a meeting with Tim on Monday to chat about what impact this has on insights, if any, which could change the direction here.

apollo-router/src/plugins/connectors/handle_responses.rs
andrewmcgivery: LazyCell seems like a simpler approach... so went with that. :D

apollo-router/src/plugins/connectors/tests/error_handling.rs
andrewmcgivery: I don't have the faintest clue. I don't even remember typing that. LOL

apollo-router/src/plugins/connectors/tests/error_handling.rs
andrewmcgivery: because this always happens regardless... 😅 

```
.with_subgraph_name(&connector.id.subgraph_name);
```

It's sort of a weird thing cause it gets run _after_ `.build()`... I suppose I could just take it out and have it only get added directly when the user isn't setting extensions mapping.

andrewmcgivery: I'm going to hold on this change until I have a chance to chat with Tim from Insights to make sure this won't break anything. :)

apollo-router/src/plugins/connectors/tests/error_handling.rs
andrewmcgivery: haha yep... but I added another field to make it more clear :)

apollo-router/src/plugins/connectors/tests/error_handling.rs
andrewmcgivery: I remember now. There was a bug but I fixed it and forgot to remove this TODO to myself 😅 lol

apollo-federation/src/sources/connect/validation/errors.rs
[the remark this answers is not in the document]
andrewmcgivery: Yep, added tests and remove the code and this works as expected!

apollo-router/src/plugins/connectors/handle_responses.rs
andrewmcgivery: Will figure this out on Monday with Tim. :)

apollo-federation/src/sources/connect/validation/errors.rs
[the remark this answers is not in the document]
lennyburdette: yeah this is really interesting. we have a couple of options:

1. don't allow `$this` and `$batch` because we don't know what the use cases are. we can always add it later. the downside is that it's an inconsistency that we'd have to maintain in code and document for users.
2. allow `$this` and `$batch`, but only if they refer to a subset of fields in the request stage or selection so that it doesn't have any effects on planning and dependencies. this is the most work and also requires maintaining and documenting the inconsistency.
3. allow `$this` and `$batch` and use them when making the key. this keeps things simple and consistent, but adds a small footgun where the user could force the planner to fetch fields that might be used only in the case of errors

i'm leaning towards #3 and the consistency it affords.

apollo-federation/src/sources/connect/validation/expression.rs
lennyburdette: we should have allow batch `xor` this — not both. not sure how to do that here

apollo-router/src/plugins/connectors/handle_responses.rs
```diff
@@ -158,25 +157,83 @@ impl RawResponse {
...
+                };
+
+                // Do we have a error extensions mapping set for this connector?
+                if let Some(extensions_selection) = &connector.error_settings.extensions {
+                    // TODO: In the future, we'll want to add to the debug context. However, we'll need a "v2" debug payload before we can do that.
+                    let (res, _apply_to_errors) =
+                        extensions_selection.apply_with_vars(&data, &inputs);
+
+                    // TODO: Currently this "fails silently". In the future, we probably add a warning to the debugger info.
+                    extensions = res.and_then(|e| e.as_object().cloned());
```
lennyburdette, line 195: we can avoid this clone with:
```rs
extensions = res.and_then(|e| {
  match e {
    Value::Object(map) => Some(map),
    _ => None
  }
});
```
chatgpt also suggests
```rs
let (res, _) = extensions_selection.apply_with_vars(&data, &inputs);
extensions = res.and_then(|e| e.as_object_mut().map(std::mem::take));
```
but i dunno if that's much better

andrewmcgivery: Good call. I'm going to go with the `match` cause `mem::take` makes me nervous. LOL

apollo-federation/src/sources/connect/validation/expression.rs
andrewmcgivery: I think the `xor` is already handled in the type checking for `connect`:

```
let all_variables = self
            .selection
            .variables()
            .chain(self.http.variables())
            .chain(self.errors.variables())
            .collect::<HashSet<_>>();
        if all_variables.contains(&Namespace::Batch) && all_variables.contains(&Namespace::This) {
            messages.push(Message {
                code: Code::ConnectBatchAndThis,
                message: format!(
                    "In {}: connectors cannot use both $this and $batch",
                    self.coordinate
                ),
                locations: self
                    .coordinate
                    .directive
                    .line_column_range(&self.schema.sources)
                    .into_iter()
                    .collect(),
            });
        }
```

So I think we're okay to just add both into the lookup and let this existing logic catch them both being in use?

apollo-router/src/plugins/connectors/tests/error_handling.rs
andrewmcgivery: Okay, I got this all figured out and squared up with removing some redundant code, adding a new comment, and adding some new tests.

tl;dr... we always call `with_subgraph_name` and it's up to `include_subgraph_errors` to redact it from the final response.

apollo-router/src/plugins/connectors/handle_responses.rs
andrewmcgivery: Okay so after meeting with Tim...

1. We always want to set `extensions.code` because it is used for telemetry
2. We always want to pass the service name (done with `with_subgraph_name`)

So I think this code is now good as it is as the defaults are only applied if the user has not set `extensions` which is the behaviour we want.

apollo-federation/src/sources/connect/models.rs
```diff
@@ -139,30 +177,58 @@ impl Connector {
...
-        let request_headers = extract_header_references(transport.variable_references());
+        let request_headers = extract_header_references(request_references);
 
-        let response_variables: HashSet<Namespace> = connect
+        // Calculate which variables and headers are in use in the response (including errors.message and errors.extensions)
+        let response_references: HashSet<VariableReference<Namespace>> = connect
             .selection
             .variable_references()
+            .chain(
+                error_settings
```
lennyburdette, line 210: optional — make a `ConnectorErrorsSettings::variable_references()` that combines message/extensions to shorten this already very long method

apollo-federation/src/sources/connect/models.rs
```diff
@@ -74,6 +77,41 @@ impl ConnectorBatchSettings {
...
+        } else if let Some(source_errors) = source_errors {
+            source_errors.extensions.clone()
+        } else {
+            None
+        };
+
+        Self {
+            message,
+            extensions,
+        }
```
lennyburdette, line 111:
```suggestion
        let message = connect_errors
            .and_then(|e| e.message.as_ref())
            .or_else(|| source_errors.and_then(|e| e.message.as_ref())).cloned();

        let extensions = connect_errors
            .and_then(|e| e.extensions.as_ref())
            .or_else(|| source_errors.and_then(|e| e.extensions.as_ref())).cloned();

        Self {
            message,
            extensions,
        }
```

i _think_ this is both shorter and fixes a logic bug. if my brain is working, what you have doesn't support this situation:

```
@source(errors: { extensions: "foo" })
@connect(errors: { message: "bar" })
```

Because `connect_errors` is `Some`, we'll try to grab `extensions` off it and get `None`, and we won't even check `source_errors.extensions`.

apollo-router/src/plugins/connectors/handle_responses.rs
lennyburdette: this matches my expectations, thanks for figuring this out with Tim!

apollo-federation/src/sources/connect/models.rs
andrewmcgivery: Good idea... updated!

apollo-federation/src/sources/connect/validation/coordinates.rs
```diff
@@ -29,12 +29,25 @@ impl Display for ConnectDirectiveCoordinate<'_> {
...
+/// The location of a `@source` directive.
+#[derive(Clone, Copy)]
+pub(super) struct SourceDirectiveCoordinate<'a> {
+    pub(super) directive: &'a Node<Directive>,
+}
+
+impl Display for SourceDirectiveCoordinate<'_> {
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
+        let Self { directive } = self;
+        write!(f, "`@{directive_name}`", directive_name = directive.name)
```
dylan-apollo, line 47: Realizing `@source` alone isn't _super_ useful, and we probably want to include the name of the source? 😬

apollo-federation/src/sources/connect/validation/errors.rs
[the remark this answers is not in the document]
dylan-apollo: It does feel weird to not allow users to do it here. I like 3 as well. If anything, we just need good tooling overall to show dependencies.

apollo-federation/src/sources/connect/validation/expression.rs
dylan-apollo: Actually these are in the wrong place! This and batch are only for `ConnectedElement::Type`, which are already handled below. Let's make sure we have tests for this.

apollo-router/src/plugins/connectors/handle_responses.rs
```diff
@@ -158,27 +157,88 @@ impl RawResponse {
...
-                    )
-                    .extension(
-                        "connector",
-                        Value::Object(Map::from_iter([(
-                            "coordinate".into(),
-                            Value::String(connector.id.coordinate().into()),
-                        )])),
+                let mut extensions = None;
+
+                // Do we have errors settings for this connector?
```
dylan-apollo, line 162:
```suggestion
```

apollo-router/src/plugins/connectors/handle_responses.rs
```diff
@@ -158,27 +157,88 @@ impl RawResponse {
...
+
+                // Do we have a error message mapping set for this connector?
+                let message = if let Some(message_selection) = &connector.error_settings.message {
+                    // TODO: In the future, we'll want to add to the debug context. However, we'll need a "v2" debug payload before we can do that.
+                    let (res, _apply_to_errors) = message_selection.apply_with_vars(&data, &inputs);
+
+                    res.unwrap_or_else(|| Value::Null)
+                        .as_str()
+                        .unwrap_or_default()
+                        .to_string()
```
dylan-apollo, line 183:
```suggestion
                    res.and_then(Value::as_str).unwrap_or_default().to_string()
```

dylan-apollo (changes requested):

apollo-federation/src/sources/connect/validation/expression.rs
andrewmcgivery: @dylan-apollo might need to connect on this via a huddle. I'm not following 😅

apollo-router/src/plugins/connectors/handle_responses.rs
andrewmcgivery: This doesn't work 😅 

```
type mismatch in function arguments
expected function signature `fn(serde_json_bytes::Value) -> _`
   found function signature `fn(&serde_json_bytes::Value) -> _`rustc[Click for full compiler diagnostic](rust-analyzer-diagnostics-view:/diagnostic%20message%20[0]?0#file:///Users/andrew/source/repos/router/apollo-router/src/plugins/connectors/handle_responses.rs)
handle_responses.rs(180, 25): required by a bound introduced by this call
option.rs(1449, 12): required by a bound in `std::option::Option::<T>::and_then`
handle_responses.rs(180, 34): consider wrapping the function in a closure: `|arg0: serde_json_bytes::Value| `, `(&arg0)`
```

apollo-federation/src/sources/connect/validation/expression.rs
andrewmcgivery: Ah, yea, those shouldn't have been under field. Removed.

apollo-federation/src/sources/connect/validation/expression.rs
```diff
@@ -79,6 +92,66 @@ impl<'schema> Context<'schema> {
...
+                }
```
dylan-apollo, line 147:
```suggestion
```

apollo-router/src/plugins/connectors/handle_responses.rs
```diff
@@ -158,10 +157,65 @@ impl RawResponse {
...
+                    // TODO: In the future, we'll want to add to the debug context. However, we'll need a "v2" debug payload before we can do that.
+                    let (res, _apply_to_errors) =
+                        extensions_selection.apply_with_vars(&data, &inputs);
+
+                    // TODO: Currently this "fails silently". In the future, we probably add a warning to the debugger info.
+                    connect_extensions = res.and_then(|e| match e {
+                        Value::Object(map) => Some(map),
+                        _ => None,
+                    });
+                }
```
dylan-apollo, line 210: Why not merge these blocks with the if statements later?

dylan-apollo (changes requested):
The bug with `$this` being allowed on fields is still present

apollo-federation/src/sources/connect/validation/expression.rs
andrewmcgivery: Okay stupid question... why would we _not_ allow `this` on fields? I think that isn't stickying to my smooth brain lol

lennyburdette: It's not allowed on fields of the root Query and Mutation types, because `$this` would allow you to refer to other root fields.

apollo-router/src/plugins/connectors/handle_responses.rs
andrewmcgivery: Done!

apollo-federation/src/sources/connect/validation/expression.rs
andrewmcgivery: But isn't that what this code is already doing? Only inserting `$this` if we're looking at "other" which is _not_ Query or Mutation?

apollo-federation/src/sources/connect/models.rs
```diff
@@ -135,30 +184,45 @@ impl Connector {
...
-        let request_headers = extract_header_references(transport.variable_references());
+        let request_headers = extract_header_references(request_references);
 
-        let response_variables: HashSet<Namespace> = connect
+        // Calculate which variables and headers are in use in the response (including errors.message and errors.extensions)
+        let response_references: HashSet<VariableReference<Namespace>> = connect
             .selection
             .variable_references()
+            .chain(error_settings.variable_references())
+            .collect();
```
naomijub, line 217: You don;t seem to need this collect as you are iterating over it in the next line again

apollo-federation/src/sources/connect/spec/directives.rs
```diff
@@ -227,19 +242,54 @@ impl SourceHTTPArguments {
     }
 }
 
+impl ErrorsArguments {
+    fn from_values(values: &ObjectNode, directive_name: &Name) -> Result<Self, FederationError> {
```
naomijub, line 246: This could be a `impl TryFrom<(&ObjectNode, &Name)> for ErrorArguments` not really an issue, but it feels more standard

apollo-federation/src/sources/connect/spec/directives.rs
```diff
@@ -227,19 +242,54 @@ impl SourceHTTPArguments {
...
+        let mut message = None;
+        let mut extensions = None;
+        for (name, value) in values {
+            let name = name.as_str();
+
+            if name == ERRORS_MESSAGE_ARGUMENT_NAME.as_str() {
```
naomijub, line 252: This cam definitively be a `match`, as  if/else-if/else tend to grow to become annoying
```rust
match name {
  ERRORS_MESSAGE_ARGUMENT_NAME => { .. }
  ERRORS_EXTENSIONS_ARGUMENT_NAME => { .. }
  _ => { .. }
}
```

apollo-federation/src/sources/connect/spec/directives.rs
```diff
@@ -255,12 +305,20 @@ impl ConnectDirectiveArguments {
                 })?;
 
                 http = Some(ConnectHTTPArguments::from_values(http_value)?);
-            } else if arg_name == "batch" {
+            } else if arg_name == BATCH_ARGUMENT_NAME.as_str() {
```
naomijub, line 308: Good example on why use a `match` hahah

apollo-federation/src/sources/connect/spec/directives.rs
naomijub: On the other hand, might need some hacky thing because `Name` is a ptr:

```rust 
match &arg.name.as_str() {
    val if val == & CONNECT_SOURCE_ARGUMENT_NAME.as_str()=> {},
  ...
};
```

apollo-federation/src/sources/connect/validation/errors.rs
```diff
@@ -0,0 +1,371 @@
...
+        let Some((_, value)) = errors_arg
+            .iter()
+            .find(|(name, _)| name == &ERRORS_MESSAGE_ARGUMENT_NAME)
+        else {
+            return Ok(None);
+        };
+        let coordinate = ErrorsMessageCoordinate { coordinate };
+
+        // Ensure that the selection is a valid JSONSelection string
+        let string = match GraphQLString::new(value, &schema.sources) {
```
naomijub, line 142: Don't care about error in the error match:
```rust
let Ok(string) = GraphQLString::new(value, &schema.sources)  else {
          return Err(Message {
              code: Code::GraphQLError,
              message: format!("{coordinate} must be a string."),
              locations: value
                  .line_column_range(&schema.sources)
                  .into_iter()
                  .collect(),
          });
  };
```

apollo-federation/src/sources/connect/validation/errors.rs
```diff
@@ -0,0 +1,371 @@
...
+                    code: Code::GraphQLError,
+                    message: format!("{coordinate} must be a string."),
+                    locations: value
+                        .line_column_range(&schema.sources)
+                        .into_iter()
+                        .collect(),
+                });
+            }
+        };
+        let selection = match JSONSelection::parse(string.as_str()) {
```
naomijub, line 155: depending on the data needed a `map_err(|err| ...)?` or  if let else could make it cleaner

apollo-federation/src/sources/connect/validation/errors.rs
```diff
@@ -0,0 +1,371 @@
...
+        let Some((_, value)) = errors_arg
+            .iter()
+            .find(|(name, _)| name == &ERRORS_EXTENSIONS_ARGUMENT_NAME)
+        else {
+            return Ok(None);
+        };
+        let coordinate = ErrorsExtensionsCoordinate { coordinate };
+
+        // Ensure that the selection is a valid JSONSelection string
+        let string = match GraphQLString::new(value, &schema.sources) {
```
naomijub, line 280: `let Ok(..) = x else {}`

apollo-federation/src/sources/connect/spec/directives.rs
andrewmcgivery: For sure, but every other directive in this file uses `from_values` so I'd rather be consistent. Perhaps we can add a Jira item to refactor it but I'd rather not do a bunch of refactoring on this PR cause it's already bene open for 3 weeks with over 100 comments 😅

apollo-federation/src/sources/connect/spec/directives.rs
andrewmcgivery: Similar to above, I think this would be a solid overall refactor but I'd rather not take any major refactors on in this PR. 😅

apollo-federation/src/sources/connect/validation/errors.rs
andrewmcgivery: Going to apply the same suggestion as above for consistency :)

dylan-apollo (approved):

PR #7312, opened 2025-04-17
impl `Default` on `HttpJsonTransport` to reduce test boilerplate

dylan-apollo:
A small chunk of #7221 which just reduces boilerplate for tests. This will become much more important when we add a ton more attributes.


lennyburdette (approved):
💯

PR #7317, opened 2025-04-18
Add subgraph validation for interface objects

sachindshinde:
Port `validateKeyOnInterfacesAreAlsoOnAllImplementations()` and `validateInterfaceObjectsAreOnEntities()` from the JS composition code, i.e. the per-subgraph validations for interface objects. Note that as part of this, `printHumanReadableList()` and `joinStrings()`, were also ported.

### DIFF
--- a/apollo-federation/src/query_graph/build_query_graph.rs
+++ b/apollo-federation/src/query_graph/build_query_graph.rs
@@ -1170,6 +1170,7 @@ impl FederatedQueryGraphBuilder {
                     schema,
                     type_pos.type_name().clone(),
                     application.fields,
+                    true,
                 )?);
 
                 // Note that each subgraph has a key edge to itself (when head == tail below).
@@ -1295,6 +1296,7 @@ impl FederatedQueryGraphBuilder {
                                     .type_name()
                                     .clone(),
                                 application.fields,
+                                true,
                             ) else {
                                 // Ignored on purpose: it just means the key is not usable on this
                                 // subgraph.
@@ -1361,6 +1363,7 @@ impl FederatedQueryGraphBuilder {
                     &self.supergraph_schema,
                     field_definition_position.parent().type_name().clone(),
                     application.fields,
+                    true,
                 )?;
                 all_conditions.push(conditions);
             }
@@ -1709,6 +1712,7 @@ impl FederatedQueryGraphBuilder {
                     schema,
                     field_type_pos.type_name().clone(),
                     application.fields,
+                    true,
                 )?;
                 all_conditions.push(conditions);
             }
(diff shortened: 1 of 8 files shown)


apollo-federation/src/schema/field_set.rs
```diff
@@ -57,15 +57,25 @@ pub(crate) fn parse_field_set(
     schema: &ValidFederationSchema,
     parent_type_name: NamedType,
     field_set: &str,
+    validate: bool,
```
duckki, line 60: How about `parse_field_set` and `parse_and_validate_field_set`?
I find bool arguments like `true` are not so reader friendly. But, it's not a big deal.

sachindshinde: I think most IDEs these days have some form of inlay parameter names, e.g. for vscode
Screenshot 2025-04-19 at 5 56 43 PM
so I haven't worried too much about it historically, but if anyone has a strong preference I'm fine with changing it.

apollo-federation/src/schema/field_set.rs
```diff
@@ -57,15 +57,25 @@ pub(crate) fn parse_field_set(
...
-        parent_type_name,
-        field_set,
-        "field_set.graphql",
-    )?;
+    let field_set = if validate {
+        FieldSet::parse_and_validate(
+            schema.schema(),
+            parent_type_name,
+            field_set,
+            "field_set.graphql",
```
clenfest, line 69: Would it not be better to leave the filename blank?

apollo-federation/src/query_graph/build_query_graph.rs
```diff
@@ -2127,6 +2131,7 @@ impl FederatedQueryGraphBuilder {
                     schema,
                     type_in_supergraph_pos.type_name.clone(),
                     "__typename",
+                    false,
```
clenfest, line 2134: Why do we not validate here?

clenfest (approved):
Since we talked about unit tests today, I feel compelled to mention them, but I won't hold up approval for it.

apollo-federation/src/schema/field_set.rs
sachindshinde: I mainly did it because other `parse_and_validate()` calls were doing it, but the other reason is that the empty filename [isn't technically a legal UNIX filename](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_170), and if `apollo-compiler` at some point uses a library that expects filenames to be valid, you may get an unexpected error.

apollo-federation/src/query_graph/build_query_graph.rs
sachindshinde: In this case, the field set (`__typename`) is trivially valid on `type_in_supergraph_pos` (which we guarantee is an interface in the preceding statement). I can make that clearer by adding a comment there, so at least if it gets changed in the future they'll remember to flip the boolean to `true`.

PR #7318, opened 2025-04-18
Move `make_uri` into `apollo-federation` `HttpJsonTransport`

dylan-apollo:
This makes some tests easier to read and gives us the ability to consume `make_uri` from the playground.

Part of #7221


andrewmcgivery (approved):
LGTM! 🚀

PR #7319, opened 2025-04-18
fix: terminate unused Redis connections on schema update

carodewig:
The router performs a 'hot reload' whenever it detects a schema update. During this reload, it effectively instantiates a new internal router, warms it up (optional), redirects all traffic to this new router, and drops the old internal router.

This PR fixes a bug in that drop process where the Redis connections are never told to terminate, even though the Redis client pool is dropped. This leads to an ever-increasing number of inactive Redis connections, which eats up memory.

The fix is simply to spawn a task which calls `quit` on the client pool within a `Drop` implementation.

TODO:
- [ ] I'm really not sure how to do unit/integration testing on this. I've performed manual testing of the fix by sending repeated schema updates - the fix shows no Redis client count growth (unlike on dev or 1.x).
- [ ] Write up a changeset.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/cache/redis.rs
+++ b/apollo-router/src/cache/redis.rs
@@ -1,5 +1,6 @@
 use std::collections::HashMap;
 use std::fmt;
+use std::ops::Deref;
 use std::pin::Pin;
 use std::sync::Arc;
 use std::time::Duration;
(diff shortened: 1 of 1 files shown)


Velfi (approved):

apollo-router/src/cache/redis.rs
```diff
@@ -51,9 +52,39 @@ pub(crate) struct RedisValue<V>(pub(crate) V)
...
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+impl Drop for DropSafeRedisPool {
+    fn drop(&mut self) {
+        let inner = self.0.clone();
+        tokio::spawn(async move {
+            let result = inner.clone().quit().await;
```
goto-bus-stop, line 77:
```suggestion
            let result = inner.quit().await;
```
This second clone probably isn't necessary, as `inner` is already owned?

goto-bus-stop (approved):

BrynCooke:
@carodewig Could the client events be used to increment/decrement a counter that could be used in a metric?

```
            // spawn tasks that listen for connection close or reconnect events
            let mut error_rx = client.error_rx();
            let mut reconnect_rx = client.reconnect_rx();

            tokio::spawn(async move {
                while let Ok(error) = error_rx.recv().await {
                    tracing::error!("Client disconnected with error: {:?}", error);
                }
            });
            tokio::spawn(async move {
                while reconnect_rx.recv().await.is_ok() {
                    tracing::info!("Redis client reconnected.");
                }
            });
```
If so we can add an integration test.
It'd be nice to have this metric from the router side.

apollo-router/src/cache/redis.rs
carodewig: Good call, that was leftover from testing!

Velfi (approved):

PR #7320, opened 2025-04-20
Simplify condition handling in coprocessor

IvanGoncharov:
Pure refactoring, during work on #7280 I notice low hanging fruits in coprocessor code.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/plugins/coprocessor/execution.rs
```diff
@@ -19,7 +18,7 @@ use crate::plugins::coprocessor::EXTERNAL_SPAN_NAME;
 use crate::services::execution;
 
 /// What information is passed to a router request/response stage
-#[derive(Clone, Debug, Default, Deserialize, PartialEq, Serialize, JsonSchema)]
```
IvanGoncharov, line 22: I notice `condition` field is skip during serialization in multiple places, e.g.:
https://github.com/apollographql/router/blob/f4ff8225124318c22c7748808ee3efbc9bd7eb8f/apollo-router/src/plugins/coprocessor/mod.rs#L316-L317

That made me wonder where we use serialization of this type and why it's valid to skip conditions. I learned that we don't use serialization for those structs so I removed them.

apollo-router/src/configuration/snapshots/apollo_router__configuration__tests__schema_generation.snap
```diff
@@ -5452,8 +5452,7 @@ snapshot_kind: text
         },
         "condition": {
           "$ref": "#/definitions/Condition_for_RouterSelector",
-          "description": "#/definitions/Condition_for_RouterSelector",
-          "nullable": true
```
IvanGoncharov, line 5456: Since I removed the Option around the Selector, it now disallows explicit null values.
I'm not sure if it's something we should worry about?

If needed I can either write a config migration to add `null` or add support of explicit `null` to serde parsing.

bnjjj (approved):

apollo-router/src/plugins/coprocessor/mod.rs
```diff
@@ -268,12 +266,11 @@ pub(super) struct RouterRequestConf {
...
 
 /// What information is passed to a router request/response stage
-#[derive(Clone, Debug, Default, Deserialize, PartialEq, Serialize, JsonSchema)]
+#[derive(Clone, Debug, Default, Deserialize, PartialEq, JsonSchema)]
 #[serde(default, deny_unknown_fields)]
 pub(super) struct RouterResponseConf {
     /// Condition to trigger this stage
-    #[serde(skip_serializing)]
-    pub(super) condition: Option<Condition<RouterSelector>>,
+    pub(super) condition: Condition<RouterSelector>,
```
carodewig, line 273: What happens if the router configuration file doesn't include a condition? I assume there will be cases of this since the field was previously nullable. Should there be a migration to add `condition: true` to those places?

IvanGoncharov: Type has `#[serde(default` so it will be populated by `RouterSelector::default` which will return `RouterSelector::True`.

PR #7324, opened 2025-04-21
Implement key, provides, and requires validation rules

tninesling:
This implements validation rules for `@key`, `@provides`, and `@requires` corresponding to [these three calls](https://github.com/apollographql/federation/blob/main/internals-js/src/federation.ts#L1620-L1679) to `validateAllFieldSet`.

The JS source iterates over all directive applications for a given directive and recursively parses its fields, applying each rule to each selected field. In that setup, rules are selectively applied based on boolean conditions configured at the top level. In Rust, the parsing is handled for us by `apollo-compiler`, which means we should not need to worry about enforcing certain conditions like `allowOnInterface` that are automatically enforced based on the directive definitions present in the schema.

Instead of creating a monolithic function like `validateAllFieldSet`, this PR separates out the various rules and implements them as visitors accepted by a `FieldSet`. The JS sources for those rules are:

- `DenyUnionAndInterfaceFields`: Specific to `@key`, implemented in the field accessor [here](https://github.com/apollographql/federation/blob/main/internals-js/src/federation.ts#L1632-L1638)
- `DenyAliases`: Applies to all three directives, implemented inside `parseFieldSetArgument` [here](https://github.com/apollographql/federation/blob/main/internals-js/src/federation.ts#L2266-L2269)
- `DenyFieldsWithDirectiveApplications*`: Applies to all three directives, implemented inside `validateFieldSetSelections` [here](https://github.com/apollographql/federation/blob/main/internals-js/src/federation.ts#L178-L183)
- `DenyFieldsWithArguments`: Applies to `@key` and `@provides`, implemented inside `validateFieldSetSelections` [here](https://github.com/apollographql/federation/blob/main/internals-js/src/federation.ts#L188-L193)
- `DenyNonExternalLeafFields`: Applies to `@key` and `@provides`, implemented inside `validateFieldSetSelections` [here](https://github.com/apollographql/federation/blob/main/internals-js/src/federation.ts#L195-L234)

Additionally, the validator for `@provides` implements two rules not related to fields, which JS applies in [this targetTypeExtractor](https://github.com/apollographql/federation/blob/main/internals-js/src/federation.ts#L1661-L1675).

Where possible, I've tried to push error serialization up to `thiserror` templating. The tests added in this changeset simply assert that the right error variant is produced for each rule using applicable test cases from the [subgraph validation tests](https://github.com/apollographql/federation/blob/main/internals-js/src/__tests__/subgraphValidation.test.ts). When those tests are fully ported, they will test the full serialization of error messages and associated codes.

---

**Checklist**

- Tests added and passing[^3]
    - [X] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


duckki (approved):
looks great!

PR #7328, opened 2025-04-22
Add a test in connectors batching for handling gaps

dylan-apollo:


apollo-router/src/plugins/connectors/tests/connect_on_type.rs
```diff
@@ -87,6 +87,80 @@ async fn basic_batch() {
...
 
+#[tokio::test]
+async fn batch_missing_items() {
+    let mock_server = MockServer::start().await;
+    Mock::given(method("GET"))
+        .and(path("/users"))
+        .respond_with(ResponseTemplate::new(200).set_body_json(json!([
+        { "id": 3 },
+        { "id": 1 },
+        { "id": 2 }])))
```
andrewmcgivery, line 98: I'd recommend requesting 1-4 (add one more id here) and returning 2 and 3 so we can make sure both before/after boundaries are covered properly. :)

andrewmcgivery (approved):

PR #7337, opened 2025-04-24
chore(deps): update xtask/Cargo.lock h2 dependency

abernix:


deny.toml
```diff
@@ -58,6 +58,7 @@ allow = [
     "LicenseRef-ring",
     "MIT",
     "MPL-2.0",
+    "NCSA",
```
goto-bus-stop, line 61: This should also be included in `about.toml` otherwise it'll fail while doing a release

abernix: ah. okay!

abernix: deaa91bf1

goto-bus-stop (approved):

PR #7342, opened 2025-04-24
More Aggressive linting in connectors

naomijub:
Removes redundant clones!

Applies some more aggressive linting to connectors to improve Rust best practices:

- `ok_or` when **allocating** should use `ok_or_else`.
- Use `unwrap_or_else` instead of `unwrap_or` when calling a function or allocating.
- When `impl` complex types (using lifetimes and generics), like, `SchemaVisitor<'a, Group, GroupType>`, prefer `Self` over explicit type.
- Make const functions const (`const fn`).
- Unless explicitly prohibited, implement `Eq` alongside `PartialEq` is a good practice. Exceptions might be f32, f64.
- `const YOUR_CONSNTANT_NAME: Name` doesn’t need to be cloned.

Fixes #**issue_number**

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-federation/src/operation/mod.rs
```diff
@@ -2566,6 +2566,7 @@ impl InlineFragmentSelection {
         };
 
         let mut remove_defer = false;
+        #[expect(clippy::redundant_clone)]
```
dylan-apollo, line 2569: TIL `expect`!

apollo-federation/src/sources/connect/json_selection/apply_to.rs
```diff
@@ -576,22 +576,23 @@ impl ApplyToInternal for WithRange<PathList> {
...
-                                    "Property {} not found in {}",
-                                    key.dotted(),
-                                    json_type_name(data),
-                                ),
-                                input_path_with_key.to_vec(),
-                                key.range(),
-                            )],
-                        )
-                    }
+                    data.get(key.as_str()).map_or_else(
```
dylan-apollo, line 579: I don't like `map_or_else` 😓 I find the other style easier to read.

apollo-federation/src/sources/connect/json_selection/apply_to.rs
```diff
@@ -720,21 +729,21 @@ impl ApplyToInternal for WithRange<PathList> {
...
-                dollar_shape.clone(),
+                dollar_shape,
                 named_var_shapes,
                 source_id,
             ),
 
-            PathList::Method(method_name, _method_args, _tail) => {
-                if let Some(_method) = ArrowMethod::lookup(method_name) {
-                    // TODO: call method.shape here to re-enable method type-checking
-                    //  call for each inner type of a One
```
dylan-apollo, line 731: This comment was load bearing, because we had a bug when forgetting to call on inner types 🙃. Maybe that should be called out in a Jira issue instead, though?

apollo-federation/src/sources/connect/json_selection/methods/future/get.rs
```diff
@@ -34,187 +34,186 @@ fn get_method(
...
+    let Some(index_literal) = args.first() else {
+        return (
+            None,
+            vec![ApplyToError::new(
+                format!("Method ->{} requires an argument", method_name.as_ref()),
+                input_path.to_vec(),
+                method_name.range(),
+            )],
+        );
+    };
```
dylan-apollo, line 56: Can these errors be merged with `and_then`? Or do we need `args` later?

apollo-federation/src/sources/connect/json_selection/methods/future/has.rs
```diff
@@ -21,58 +21,56 @@ fn has_method(
...
             None,
             vec![ApplyToError::new(
                 format!("Method ->{} requires an argument", method_name.as_ref()),
                 input_path.to_vec(),
                 method_name.range(),
             )],
-        )
+        );
+    };
+    let Some(arg) = args.first() else {
```
dylan-apollo, line 34: Same merge branch Q

apollo-federation/src/sources/connect/json_selection/methods/public/last.rs
```diff
@@ -103,11 +98,7 @@ fn last_shape(
 
         ShapeCase::Array { prefix, tail } => {
             if tail.is_none() {
-                if let Some(last) = prefix.last() {
-                    last.clone()
-                } else {
-                    Shape::none()
-                }
+                prefix.last().map_or_else(Shape::none, |last| last.clone())
```
dylan-apollo, line 101:
```suggestion
                prefix.last().cloned().unwrap_or_else(Shape::none)
```
Is that... better?

apollo-federation/src/sources/connect/json_selection/parser.rs
```diff
@@ -384,25 +367,18 @@ impl NamedSelection {
...
-                    // Flatten and deduplicate the names of the NamedSelection
-                    // items in the SubSelection.
-                    let mut name_set = IndexSet::default();
-                    for selection in sub.selections_iter() {
-                        name_set.extend(selection.names());
-                    }
-                    name_set.into_iter().collect()
+                    sub.selections_iter()
+                        .flat_map(|selection| selection.names())
+                        .collect()
```
dylan-apollo, line 381: This removed the deduplication, but could use https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.unique for the same effect?

apollo-federation/src/sources/connect/json_selection/parser.rs
```diff
@@ -1073,7 +1048,7 @@ impl Key {
         WithRange::new(self, None)
     }
 
-    pub fn is_quoted(&self) -> bool {
+    pub const fn is_quoted(&self) -> bool {
```
dylan-apollo, line 1051: I wouldn't `const` any `pub` functions since it commits us to a contract. Not that we much care about non-router consumers re: stability

apollo-federation/src/sources/connect/validation/connect/selection/variables.rs
```diff
@@ -154,7 +154,7 @@ fn resolve_path(
                             "`{variable_type}` does not have a field named `{nested_field_name}`."
                         ),
                         locations: expression.line_col_for_subslice(
-                            path_component_range.start..path_component_range.end,
```
dylan-apollo, line 157: lol

dylan-apollo (changes requested):
I think the only must-change is restoring deduplication.. unless you're confident we don't need it.

apollo-federation/src/sources/connect/json_selection/apply_to.rs
naomijub: Agreed, TODOs on large functioning  code bases tend to become lost

apollo-federation/src/sources/connect/json_selection/methods/future/get.rs
naomijub: good call

apollo-federation/src/sources/connect/json_selection/parser.rs
naomijub: actually forgot that, was collecting to a hashset and noticed unique does the something but didnt add it. 
Now, I tried adding a test to cover this, but doesn't seem to make a difference because of how parse works. Also, commented on this in the channel

apollo-federation/src/sources/connect/json_selection/parser.rs
naomijub: The ones that are returning `&'static str` I will leave as const, ok? Doesnt really make sense to have them as non const

naomijub: 1f8046fc2

apollo-federation/src/sources/connect/json_selection/parser.rs
naomijub: 1f8046fc2

apollo-federation/src/sources/connect/json_selection/methods/public/last.rs
naomijub: A lot 1f8046fc2

apollo-federation/src/sources/connect/json_selection/methods/future/has.rs
naomijub: 1f8046fc2

apollo-federation/src/sources/connect/json_selection/methods/future/get.rs
naomijub: 1f8046fc2

apollo-federation/src/sources/connect/json_selection/apply_to.rs
naomijub: 1f8046fc2

apollo-federation/src/sources/connect/json_selection/methods/public/last.rs
```diff
@@ -83,14 +80,12 @@ fn last_shape(
 
     match input_shape.case() {
         ShapeCase::String(Some(value)) => {
-            if let Some(last_char) = value.chars().last() {
+            value.chars().last().map_or_else(Shape::none, |last_char| {
```
dylan-apollo, line 83: This one I like! Maybe it's because the fallback case is so simple?

apollo-federation/src/sources/connect/json_selection/parser.rs
```diff
@@ -384,25 +368,19 @@ impl NamedSelection {
...
-                if let Some(alias) = alias {
-                    vec![alias.name.as_str()]
-                } else {
-                    vec![name.as_str()]
-                }
-            }
-            Self::Path { alias, path, .. } => {
+            Self::Field(alias, name, _) => alias
+                .as_ref()
+                .map_or_else(|| vec![name.as_str()], |alias| vec![alias.name.as_str()]),
```
dylan-apollo, line 373: I think part of what hurts my brain is that the `None` case is first. The function is called `map_or_else`, but you have to put the `else` before the `map` 😵‍💫. As far as the functions go, I think this conveys intent better:

```suggestion
            Self::Field(alias, name, _) => alias
                .as_ref().map(|alias| vec![alias.name.as_str()])
                .unwrap_or_else(|| vec![name.as_str()]),
```

apollo-federation/src/sources/connect/json_selection/parser.rs
```diff
@@ -384,25 +368,19 @@ impl NamedSelection {
...
-                    // items in the SubSelection.
-                    let mut name_set = IndexSet::default();
-                    for selection in sub.selections_iter() {
-                        name_set.extend(selection.names());
-                    }
-                    name_set.into_iter().collect()
+                    sub.selections_iter()
+                        .flat_map(|selection| selection.names())
+                        .unique()
+                        .collect()
```
dylan-apollo, line 383: BTW did the non-duplicate version of this fail any tests? If not... we should make sure whatever this behavior is we're testing for it.

apollo-federation/src/sources/connect/json_selection/parser.rs
```diff
@@ -1181,17 +1158,18 @@ pub(crate) fn parse_string_literal(input: Span) -> ParseResult<WithRange<String>
...
+                |remainder| {
+                    Ok((
+                        remainder,
+                        WithRange::new(
+                            chars.iter().collect::<String>(),
+                            Some(start..remainder.location_offset()),
+                        ),
+                    ))
+                },
+            )
```
dylan-apollo, line 1172: I also like the `ok` methods for options, though not sure if it clears anything up...

```suggestion
            remainder_opt.ok_or_else(|| nom_fail_message(input, "Unterminated string literal"))
            .map(|remainder| {
                (
                    remainder,
                    WithRange::new(
                        chars.iter().collect::<String>(),
                        Some(start..remainder.location_offset()),
                    )
                 )
            })
```

apollo-federation/src/sources/connect/json_selection/parser.rs
naomijub: It didn't fail and that is what caused all my comments about testing

apollo-federation/src/sources/connect/json_selection/parser.rs
naomijub: I like it as well!

apollo-federation/src/sources/connect/json_selection/parser.rs
dylan-apollo: Just poking around the code, I don't think it would cause any real bugs for names to be duplicated, it would just cause us to rerun code extra times in expansion and validation when it doesn't need to. So I _think_ deduplicating is a performance optimization?

I was gonna suggest some sort of functional test for whatever this was preventing, but maybe what we need is just a unit test for now?

naomijub: I wasnt able to create a Unit test for this, maybe we can pair on that later

apollo-federation/src/sources/connect/json_selection/parser.rs
```diff
@@ -384,29 +368,19 @@ impl NamedSelection {
...
-                    name_set.into_iter().collect()
-                } else {
-                    vec![]
-                }
-            }
+            Self::Field(alias, name, _) => alias
+                .as_ref()
+                .map(|alias| vec![alias.name.as_str()])
+                .unwrap_or_else(|| vec![name.as_str()]),
+            Self::Path { alias, path, .. } => match (alias, path.next_subselection()) {
```
dylan-apollo, line 375: This one is technically less efficient because it will eagerly call `next_subselection()` instead of only when `alias` is `None`

dylan-apollo (approved):

apollo-federation/src/sources/connect/json_selection/parser.rs
naomijub: True, I didnt realize it was dealing with the Range property
1a3ee169e

PR #7344, opened 2025-04-25
Add _dd.measured to spans when agent sampling is enabled

BrynCooke:
The 2.x branch doesn't seem to be adding the _dd.measured attribute to spans. And the tests look incorrect.

The tests are now fixed and the attribute is added when enabling datadog agent sampling.

This will need backporting as the test is likely incorrect in 1.x.

It seems that this is working in 1.x because the spans are being set to _top_level, which automatically makes them measured.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


lrlna (approved):

carodewig (approved):

apollo-router/src/plugins/telemetry/tracing/datadog/span_processor.rs
```diff
@@ -36,6 +37,7 @@ impl<T: SpanProcessor> SpanProcessor for DatadogSpanProcessor<T> {
             span.span_context.is_remote(),
             span.span_context.trace_state().clone(),
         );
+        span.attributes.push(KeyValue::new("_dd.measured", "1"));
```
bnjjj, line 40: I think for 1.x we should not add this and just update the test to make sure it works properly.

bnjjj (approved):

apollo-router/src/plugins/telemetry/tracing/datadog/span_processor.rs
BrynCooke: Agreed

apollo-router/src/plugins/telemetry/tracing/datadog/span_processor.rs
```diff
@@ -36,6 +37,7 @@ impl<T: SpanProcessor> SpanProcessor for DatadogSpanProcessor<T> {
             span.span_context.is_remote(),
             span.span_context.trace_state().clone(),
         );
+        span.attributes.push(KeyValue::new("_dd.measured", 1));
```
theJC, line 40: This will set _dd.measured on ALL spans?  If so, do we really want that, or just want certain subset of spans to be measured?

REF: https://github.com/DataDog/documentation/issues/22774#issuecomment-2075965455
> As for measured, that is intentionally undocumented as it must be used carefully, typically with help from our Support team. It allows span metrics to be generated for custom traces.

abernix: @theJC this is a good callout.  we're going to revert this PR and for the moment, this configuration can be used:

```yaml
   supergraph:
        attributes:
          "_dd.measured":
            static: 1
```

abernix: @theJC: @BrynCooke has worked with @cgati to reach out to Datadog support to get advice on this.  At the current time though, we believe not returning this option to enabled is preventing the proper functionality of the Datadog APM view on Router 2.x.  Stay tuned (but let us know if you have any additional thoughts! :) )

PR #7346, opened 2025-04-27
rewrite upgrade_subgraphs_if_necessary() to minimize cloning and fix tests.

clenfest:
Rewrote `upgrade_subgraphs_if_necessary()` to minimize cloning. Function will now take ownership of `Subgraph` array and either pass the original one back in the case of fed2 or create a new `Subgraph` in the case of fed1. 

Also made some various bug fixes including:
  - fixing the way field sets are serialized
  - added position helper functions
  - add tag and interfaceObject directives to bootstrapping code
  - recompute metadata when required
  
Two main things need to be done, and I will open tickets for them. 
1. Tests are still currently skipped. They work if I add fed2 directives, but that breaks tests elsewhere, so I've disabled it for now. 
2. We need to analyze where and how we use metadata to see if we need to recompute. Also somehow make this safe so that we can't accidentally modify schema in a way that would affect the precomputed metadata without recomputing it.


apollo-federation/src/link/federation_spec_definition.rs
```diff
@@ -939,6 +979,43 @@ pub(crate) fn add_fed1_link_to_schema(
     )
 }
 
+#[allow(unused)]
+pub(crate) fn add_fed2_link_to_schema(
```
tninesling, line 983: Not sure which one will go in first, but just an fyi this overlaps with Duckki's [PR](https://github.com/apollographql/router/pull/7343)

apollo-federation/src/link/federation_spec_definition.rs
```diff
@@ -939,6 +979,43 @@ pub(crate) fn add_fed1_link_to_schema(
...
+    schema: &mut FederationSchema,
+) -> Result<(), FederationError> {
+    // Insert `@core(feature: "http://specs.apollo.dev/federation/v1.0")`.
+    // We can't use `import` argument here since fed1 @core does not support `import`.
```
tninesling, line 987: Is this comment accurate?

apollo-federation/src/schema/schema_upgrader.rs
```diff
@@ -439,8 +519,9 @@ impl<'a> SchemaUpgrader<'a> {
     }
 
     fn fix_inactive_provides_and_requires(&mut self) -> Result<(), FederationError> {
+        let cloned_schema = self.schema.clone();
         remove_inactive_requires_and_provides_from_subgraph(
-            self.original_subgraph.schema(),
+            &cloned_schema, // TODO: I don't know what this value should be
```
tninesling, line 524: Is this just because we need to iterate over values and mutate at the same time?

apollo-federation/src/subgraph/typestate.rs
```diff
@@ -59,6 +60,7 @@ pub struct Validated {
 trait HasMetadata {
     fn metadata(&self) -> &SubgraphMetadata;
     fn schema(&self) -> &FederationSchema;
+    fn is_fed_1(&self) -> bool;
```
tninesling, line 63: Is there a reason this can't be implemented as `self.metadata().federation_spec_definition().is_fed1()`?

clenfest: So I originally had this because I was going to add fed2 links rather than fed1 ones and use this to disambiguate between the two. I'm not 100% sure of the approach, but I'll figure it out in the next PR and clean this up.

apollo-federation/src/link/federation_spec_definition.rs
clenfest: Similar to comment below. I just deleted this code for now.

apollo-federation/src/schema/schema_upgrader.rs
tninesling: Ah, this is because the existing implementation we're borrowing from the QP rewrite wants the supergraph schema to parse the field sets in requires and provides. I was thinking the next best thing is to pass the subgraph schema there, since we obviously don't have a supergraph at this point, so what you have is probably correct

tninesling (approved):

PR #7347, opened 2025-04-28
Avoid unnecessary cloning in the deduplication plugin

garypen:
We now only duplicate if there is a waiter. I think this also fixes an undiagnosed race condition where an entry could remain in the wait map AFTER the notifications have been sent.

This is also fixed by explicitly waiting for the spawned task to complete before checking for receivers and sending the notification to them.


apollo-router/src/plugins/traffic_shaping/deduplication.rs
```diff
@@ -124,35 +124,39 @@ where
                     let authorization_cache_key = request.authorization.clone();
                     let id = request.id.clone();
                     let cache_key = ((&request.subgraph_request).into(), authorization_cache_key);
-                    let res = {
+                    let (res, hdl) = {
```
goto-bus-stop, line 127: What does hdl mean?

garypen: handle

garypen: You can take the person out of unix, but you can't take the unix out of.... :)

garypen: I did `s/hdl/handle/g` in https://github.com/apollographql/router/pull/7347/commits/e5ecdc863869e801cabca5432c74e810eb6dac71

.changesets/fix_garypen_dedup_less_clone.md
```diff
@@ -0,0 +1,11 @@
+### Avoid unnecessary cloning in the deduplication plugin ([PR #7347](https://github.com/apollographql/router/pull/7347))
+
+The deduplication plugin always cloned responss, even if there were not multiple simultaneous requests that would benefit from the cloned response.
```
bnjjj, line 3:
```suggestion
The deduplication plugin always cloned responses, even if there were not multiple simultaneous requests that would benefit from the cloned response.
```

bnjjj (changes requested):

.changesets/fix_garypen_dedup_less_clone.md
```diff
@@ -0,0 +1,11 @@
...
+
+We now check to see if deduplication will provide a benefit before we clone the subgraph response.
+
+There was also an undiagnosed race condition which mean that a notification could be missed. This would have resulted in additional work being performed as the missed notification would have led to another subgraph request.
+
+<!-- https://apollographql.atlassian.net/browse/ROUTER-973 -->
```
goto-bus-stop, line 9:
```suggestion
```

apollo-router/src/plugins/traffic_shaping/deduplication.rs
```diff
@@ -124,35 +124,42 @@ where
...
-                    // We may get errors here, for instance if a task is cancelled,
-                    // so just ignore the result of send
-                    let _ = tokio::task::spawn_blocking(move || {
-                        tx.send(broadcast_value)
-                    }).await
-                    .expect("can only fail if the task is aborted or if the internal code panics, neither is possible here; qed");
+                        // We may get errors here, for instance if a task is cancelled,
+                        // so just ignore the result of send
+                        let _ = tokio::task::spawn_blocking(move || {
+                            tx.send(broadcast_value)
```
goto-bus-stop, line 160: I'm actually not sure why this is even in a spawn_blocking. `send()` does access a mutex but so does `tx.receiver_count()`. I think it's unnecessary

goto-bus-stop (approved):
LGTM, just non blocking comments for me

apollo-router/src/plugins/traffic_shaping/deduplication.rs
garypen: I don't like changing code that isn't part of the fix is the reason why I left it and I think the motivation was to prevent the send being lost if the task was cancelled. However, the loop takes care of that for us and so I think you are right and we can remove the spawn_blocking. I'll do that. Fixed: https://github.com/apollographql/router/pull/7347/commits/f9590bcf3f82910226e94ef122d43d9d92cb211a

.changesets/fix_garypen_dedup_less_clone.md
garypen: fixed: https://github.com/apollographql/router/pull/7347/commits/f9590bcf3f82910226e94ef122d43d9d92cb211a

.changesets/fix_garypen_dedup_less_clone.md
garypen: fixed: https://github.com/apollographql/router/pull/7347/commits/f9590bcf3f82910226e94ef122d43d9d92cb211a

bnjjj (approved):

BrynCooke (approved):

PR #7349, opened 2025-04-28
Clarify coprocessor config `context: true` behaviour

goto-bus-stop:
`context: true` is how you enabled propagating context through coprocessors in 1.x. Now, you should use `context: deprecated` to use the 1.x context key names, or `context: all` to use the 2.x context key names.

`context: true` is an alias for `context: deprecated` but should not be used. This patch clarifies in the docs that you should not use `context: true`, and logs a runtime warning on startup if you do use it.

### DIFF
--- a/apollo-router/src/plugins/coprocessor/mod.rs
+++ b/apollo-router/src/plugins/coprocessor/mod.rs
@@ -115,6 +115,31 @@ impl Plugin for CoprocessorPlugin<HTTPClientService> {
             builder.wrap_connector(http_connector)
         };
 
+        if matches!(init.config.router.request.context, ContextConf::Deprecated(true)) {
+            tracing::warn!("Configuration `coprocessor.router.request.context: true` is deprecated. Use one of the new values, `deprecated` or `all`");
+        }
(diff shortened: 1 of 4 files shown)


apollo-router/src/plugins/coprocessor/mod.rs
```diff
@@ -115,6 +115,71 @@ impl Plugin for CoprocessorPlugin<HTTPClientService> {
             builder.wrap_connector(http_connector)
         };
 
+        if matches!(
+            init.config.router.request.context,
+            ContextConf::Deprecated(true)
+        ) {
+            tracing::warn!(
+                "Configuration `coprocessor.router.request.context: true` is deprecated. Use one of the new values, `deprecated` or `all`"
```
bnjjj, line 123: Is it intentional to not mention the third option `selective` ? Should we provide a link to the official docs ?

goto-bus-stop: i was trying to keep it simple, but a docs link might be better (especially as `selective` is the _best_ thing to use...)

bnjjj (approved):

PR #7357, opened 2025-04-28
[WIP] Migrate Error Counting to Telemetry Plugin

rregitsky:
*Description here*

Fixes #**issue_number**

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/error.rs
```diff
@@ -154,12 +154,16 @@ impl FetchError {
...
-            message: self.to_string(),
-            locations: Default::default(),
-            path,
-            extensions: value.as_object().unwrap().to_owned(),
-        }
+        Error::builder()
+            .message(self.to_string())
+            .locations(Vec::default())
+            .and_path(path)
+            // Extension code is required, but is only used if extensions doesn't have it. We always
```
timbotnik, line 161: I didn't dig into it but why is it required?  Is it b/c of the constructor param not being optional?

rregitsky: Yeah exactly. I've decided to change it to optional because I don't really see a good reason why it needs to required. I get the idea that we prefer errors use them, but until we force it everywhere I don't think we can require it.

apollo-router/src/graphql/mod.rs
```diff
@@ -201,8 +230,22 @@ impl Error {
             locations,
             path,
             extensions,
+            apollo_id: Uuid::new_v4()
```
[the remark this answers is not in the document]
rregitsky: I'd like to hear more about the history of these errors. Why do we store them in extensions instead of errors? If we keep it in extensions, I don't see a good way to avoid serde

PR #7366, opened 2025-04-29
Build nightly image on every PR

lrlna:
To make it easier to verify changes introduced in a given PR, this work introduces building nightly images for every PR.

Since we'll be creating a few more images than before, I am also adding a workflow for deleting old nightlies (`nightly/router` docker assets)

For the time being, I chose to keep with the nightly package rather than creating a new one just for PRs. Let me know if you are of a different opinion.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

### DIFF
--- a/.github/workflows/delete-oldest-nightlies.yaml
+++ b/.github/workflows/delete-oldest-nightlies.yaml
@@ -4,7 +4,7 @@ on:
   workflow_dispatch:
     inputs:
       num_old_versions_to_delete:
-        description: 'Number of oldest images nightly/router to delete'
+        description: 'Number of oldest `nightly/router` images to delete'
         required: true
         type: number 
 jobs:


.github/workflows/delete-oldest-nightlies.yaml
```diff
@@ -0,0 +1,19 @@
+name: 'Delete Oldest Nightly Images'
+run-name: "{{ github.actor }} is deleting ${{ inputs.num_old_versions_to_delete }} oldest nightlies"
+on:
+  workflow_dispatch:
+    inputs:
+      num_old_versions_to_delete:
+        description: 'Number of oldest images nightly/router to delete'
```
carodewig, line 7:
```suggestion
        description: 'Number of oldest `nightly/router` images to delete'
```

carodewig (commented):
LGTM! Just a few questions:
* What are your thoughts on adding another workflow that runs on PR merge/close to delete the nightly images for that PR?
* Do the images include the PR number or similar identifying information?

carodewig:
> They have a git sha associated with it, but they can have a PR associated with it. Although, we'll need to find out how docker behaves if you're reusing an artefact name (my hope is that it just rewrites it). But having a PR number associated with it will make it really easy to auto delete, and will make sure we don't have too many artefacts being generated. Otherwise every commit in a PR has the potential to have an artefact.

My expectation would be that when you push an image with a tag that already exists in the registry, the registry will untag the existing image (but not delete it) and then use the tag for your new image. The `delete-only-untagged-versions` option from https://github.com/actions/delete-package-versions would take care of that!
FWIW there are already some untagged images in the registry which should probably be cleaned up: https://github.com/apollographql/router/pkgs/container/nightly%2Frouter/versions

lrlna:
Closing until we get this re-prioritised.

PR #7369, opened 2025-04-29
fix: propagate client name and version modifications through telemetry

carodewig:
The router accepts modifications to the client name and version (`apollo::telemetry::client_name` and `apollo::telemetry::client_version`), but those modifications are not currently propagated through the telemetry layers to update spans and traces.

This PR moves where the client name and version are bound to the span, so that the modifications from plugins **on the `router` service** are propagated. Specifically, it moves the binding to within the `async` block of the plugin, which allows the other `router`-service-level plugins' synchronous components to run before the binding.

Other changes in this PR:
* New test `test_plugin_overridden_client_name_is_included_in_telemetry` to validate this specific behavior
* New functionality to the OTLP and Datadog `TraceSpec`s to check attributes on spans

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [x] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


carodewig:
NB: I've applied the backport label but only intend to backport the `attribute` testing changes!

bnjjj (approved):

PR #7371, opened 2025-04-30
tests(cache): add invalidation test for entity cache

bnjjj:


apollo-router/tests/integration/redis.rs
```diff
@@ -579,6 +580,400 @@ async fn entity_cache_basic() -> Result<(), BoxError> {
...
+                    }
+                }
+            },
+            "include_subgraph_errors": {
+                "all": true
+            },
+            "supergraph": {
+                // TODO(@goto-bus-stop): need to update the mocks and remove this, #6013
+                "generate_query_fragments": false,
+            },
```
goto-bus-stop, line 759: You shouldn't need this with `experimental_mock_subgraphs`

SimonSapin (approved):
I’m clicking "Approve" (assuming `generate_query_fragments: false` is removed) because we’re in dire need of test coverage. But rather than positive approval I mean it in the sense of I don’t oppose merging this.

I strongly believe we should avoid writing new tests that rely on exact database entries like this. When anything changes it’s a huge pain to update all of those hashes hard-coded in tests and snapshots. I think we should treat the database as opaque in tests, and instead check what we actually care about: whether a request ends up doing a subgraph fetch or whether that fetch can be skipped thanks to cache. `apollo-router/tests/integration/entity_cache.rs` does this with a `subgraph_hook` that increments a counter on each fetch.

PR #7377, opened 2025-04-30
Fix router enforcement of connect spec version.

dylan-apollo:
In composition, we set the top-level `@link` for Connectors to the latest spec so that older Routers will reject _any_ new updates. This enables us to add new features without a bunch of version checks.

However, before this PR Router was checking the _subgraph_ spec version of Connectors, which may not be the latest.

This corrects that behavior by making three changes:

- Copy over the connect `@link` during expansion
- Validate that the connect `@link` version is known when parsing
- Use the top-level `@link` version in config gate (for enabling experimental versions, like 0.2 right now)

[CNN-751]: https://apollographql.atlassian.net/browse/CNN-751?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apollo-federation/src/lib.rs
```diff
@@ -177,3 +182,40 @@ const _: () = {
...
+mod test_supergraph {
+    use pretty_assertions::assert_str_eq;
+
+    use super::*;
+
+    #[test]
+    fn validates_connect_spec_is_known() {
+        let res = Supergraph::new(
+            r#"
+        extend schema @link(url: "https://specs.apollo.dev/connect/v99.99")
```
lennyburdette, line 196: you don't think we'll get to v99? so pessimistic

apollo-federation/src/sources/connect/expand/carryover.rs
```diff
@@ -211,16 +200,10 @@ pub(super) fn carryover_directives(
             .get_directive(&directive_name)
             .and_then(|referencers| {
                 if referencers.len() > 0 {
-                    insert_link = true;
                     copy_directive_definition(from, to, directive_name.clone())?;
                 }
                 referencers.copy_directives(from, to, &directive_name)
             })?;
-
-        if insert_link {
```
lennyburdette, line 220: there's another `if insert_link` in the `// @context` section that i think you can remove?

dylan-apollo: Got it! I _think_ the one in the loop for compose directive should still be there? Maybe? Not positive, though.

lennyburdette (approved):

PR #7378, opened 2025-05-01
Apollo telemetry: enhance error details for connector errors

timbotnik:
Previously, the error message and path for connectors / demand control errors would not be included with traces sent to GraphOS, which hinders the experience.  This will be an enhancement for Apollo telemetry in Studio.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [x] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-1293]: https://apollographql.atlassian.net/browse/ROUTER-1293?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apollo-router/tests/fixtures/supergraph_connect.yaml
```diff
@@ -0,0 +1,38 @@
+# rover supergraph compose --config apollo-router/tests/fixtures/supergraph_connect.yaml > apollo-router/tests/fixtures/supergraph_connect.graphql
+federation_version: =2.10.0-preview.3
```
dylan-apollo, line 2:
```suggestion
federation_version: =2.10.0
```

bonnici (approved):
Looks good to me once other comments have been addressed.

apollo-router/src/plugins/demand_control/mod.rs
```diff
@@ -379,13 +380,23 @@ impl Plugin for DemandControl {
                     Ok(match strategy.on_execution_request(&req) {
                         Ok(_) => ControlFlow::Continue(req),
                         Err(err) => {
-                            emit_error_event(err.code(), "Demand control execution error");
+                            let graphql_errors = err
+                                .into_graphql_errors()
+                                .expect("must be able to convert to graphql error");
```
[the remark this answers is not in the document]
timbotnik: I feel like making that change is also making a call about whether to allow control flow to continue in the case of an improperly implemented demand control error, or deciding what kind of error to return instead.  Seems a bit out of scope for this PR since it's existing code (and this also happens in another place in the same file).

I had a look at the implementation and every match arm returns `Ok(vec![...])`.  Changing the function signature (to remove the superfluous `Result` wrapper) is also not a great solution since it comes in from the `IntoGraphQLErrors` trait, even though there is not a way today to return an error from `into_graphql_errors()`.

apollo-router/src/plugins/telemetry/tracing/apollo_telemetry.rs
```diff
@@ -204,13 +211,26 @@ const REPORTS_INCLUDE_SPANS: [&str; 16] = [
...
+        );
+    } else {
+        tracing::event!(
+            Level::ERROR,
+            { GRAPHQL_ERROR_EXT_CODE } = error_code,
+            { FIELD_EXCEPTION_MESSAGE } = error_message,
+            { EVENT_ATTRIBUTE_OMIT_LOG } = true,
+            error_message
+        );
+    }
```
lennyburdette, line 233:
```suggestion
    if let Some(path) = error_path {
        tracing::event!(
            Level::ERROR,
            { GRAPHQL_ERROR_EXT_CODE } = error_code,
            { FIELD_EXCEPTION_MESSAGE } = error_message,
            { GRAPHQL_ERROR_PATH } = path.to_string().as_str(),
            { EVENT_ATTRIBUTE_OMIT_LOG } = true,
            error_message
        );
    } else {
        tracing::event!(
            Level::ERROR,
            { GRAPHQL_ERROR_EXT_CODE } = error_code,
            { FIELD_EXCEPTION_MESSAGE } = error_message,
            { EVENT_ATTRIBUTE_OMIT_LOG } = true,
            error_message
        );
    }
```

timbotnik: Applied thanks!

PR #7383, opened 2025-05-02
fix: make sure all input types end up in the expanded supergraph

lennyburdette:
this is particularly annoying because the original types might have join__* directives that refer to subgraphs that no longer exist

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


lennyburdette:
@dylan-apollo for some reason the snapshot results have connect/v0.1 instead of v0.2 in them and i can't figure out how I changed that. any ideas?

apollo-federation/src/sources/connect/expand/carryover/inputs.rs
```diff
@@ -0,0 +1,338 @@
...
+                let node = strip_invalid_join_directives_from_input_type(
+                    node,
+                    &subgraph_enum_replacements,
+                );
+                pos.pre_insert(to).ok();
+                pos.insert(to, node).ok();
+            }
+        }
+        _ => {}
+    });
```
dylan-apollo, line 90:
```suggestion
    for (name, ty) in &from.schema().types {
        if to.schema().types.contains_key(name) {
            continue;
        }
        match ty {
            ExtendedType::Scalar(node) => {
                let pos = ScalarTypeDefinitionPosition {
                    type_name: node.name.clone(),
                };
                let node =
                    strip_invalid_join_directives_from_scalar(node, &subgraph_enum_replacements);
                pos.pre_insert(to).ok();
                pos.insert(to, node).ok();
            }
            ExtendedType::Enum(node) => {
                let pos = EnumTypeDefinitionPosition {
                    type_name: node.name.clone(),
                };
                let node =
                    strip_invalid_join_directives_from_enum(node, &subgraph_enum_replacements);
                pos.pre_insert(to).ok();
                pos.insert(to, node).ok();
            }
            ExtendedType::InputObject(node) => {
                let pos = InputObjectTypeDefinitionPosition {
                    type_name: node.name.clone(),
                };
                let node = strip_invalid_join_directives_from_input_type(
                    node,
                    &subgraph_enum_replacements,
                );
                pos.pre_insert(to).ok();
                pos.insert(to, node).ok();
            }
            _ => {}
        }
    }
```

Don't know if we _want_ to do this, but this would also skip any types which match the name of a different type in `to`. So if someone named a scalar the same as a input somehow, for example.

If we want to error on that, we would put `?` instead of `.ok()` in all the pre/inserts (which we can do in normal for loop but not `for_each`

apollo-federation/src/sources/connect/expand/carryover/inputs.rs
```diff
@@ -0,0 +1,338 @@
...
+                new_directives.push(new_directive.clone());
+            }
+        } else {
+            new_directives.push(d.clone());
+        }
+    }
+    new_directives
+}
+
+/// Unfortunately, there are two different DirectiveList types, so this
```
dylan-apollo, line 234: 😞

apollo-federation/src/sources/connect/expand/tests/schemas/expand/carryover.yaml
```diff
@@ -67,12 +69,52 @@ subgraphs:
...
           id: ID!
         }
+
+        # bug fix: this won't compose until it's fixed and released in federation
+        # the graphql file is currently hand-edited to add these definitions
+        #
+        # @custom appears in both subgraphs, so will be merged appropriately, and it will attributed only to the non-connector subgraph
+        # @custom2 appears in the connector subgraph, so we have to add it and rewrite the join__* directives
+        # @custom3 appears in the non-connector subgraph, so it's composed appropriately
+        #
```
dylan-apollo, line 79: I feel like we should leave them in so we don't get an incorrect `.graphql` in the future? Maybe with a comment in here that it won't compose for < 2.11.0-preview.3?

dylan-apollo (approved):
LGTM, though the code is dense enough that we totally could be missing something

PR #7390, opened 2025-05-02
fix: limit `http.route` to just the path in telemetry

carodewig:
Per the [OpenTelemetry spec](https://opentelemetry.io/docs/specs/semconv/attributes-registry/http/#http-route), the `http.route` should only include "the matched route, that is, the path template used in the format used by the respective server framework."

The router currently sends the full URI in `http.route`, which can be high cardinality (ie `/graphql?operation=one_of_many_values`). After this change, the router will only include the path (`/graphql`).

In order to test this, I added a dependency on [`tracing-mock`](https://docs.rs/tracing-mock/latest/tracing_mock/index.html). It's currently in beta, but it's already used within the `tracing` library itself so I think it's probably pretty stable. However I'm happy to come up with another way to test this if preferred.

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -406,6 +406,7 @@ dependencies = [
  "tracing",
  "tracing-core",
  "tracing-futures",
+ "tracing-mock",
  "tracing-opentelemetry",
  "tracing-serde 0.1.3",
  "tracing-subscriber",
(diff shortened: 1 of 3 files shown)


bnjjj (approved):
Do we have a way to provide the full uri ? An existing selector to make sure if our users would like to have this data they can still have it ?

carodewig:
> Do we have a way to provide the full uri ? An existing selector to make sure if our users would like to have this data they can still have it ?

@bnjjj Good question - it looks like we can provide the URL path and query via instrumentation attributes: https://www.apollographql.com/docs/graphos/routing/observability/telemetry/instrumentation/standard-attributes#router

PR #7398, opened 2025-05-05
add validations for context directives

clenfest:


apollo-federation/src/link/context_spec_definition.rs
```diff
@@ -64,6 +69,83 @@ impl ContextSpecDefinition {
...
+                default_value: None,
+            },
+            composition_strategy: None,
+        }
+    }
+
+    fn context_field_value_type(
+        schema: &FederationSchema,
+    ) -> Result<ScalarTypeDefinitionPosition, FederationError> {
+        let name_in_schema = Self::context_field_value_name(schema)?.unwrap();
```
tninesling, line 90: Is there a way to avoid the `.unwrap()` here?

tninesling (approved):
One small nit about an `unwrap`. Otherwise, lgtm!

PR #7399, opened 2025-05-05
Update FieldSet validations to match expectations from subgraph tests

tninesling:
Enables several subgraph validation tests and updates the validation error messages to match. This largely boils down to adding a "On field X for application Y," preamble to each of the existing messages.

In order to get these messages to match, we have to plumb the directive being validated down to where we create the error message. I've reworked the validation visitor trait to include this. These replace the unit tests from the validation files, which were copied from the same test cases. I've removed those because they're duplicates and it's difficult to create a directive instance with all the necessary coordinates from the schema.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [X] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


duckki (approved):
Nice! I like the refactoring of common rules.

PR #7418, opened 2025-05-06
Preserve formatting of `@link` when auto-upgrading connect spec

dylan-apollo:
This way, if later stages (JS validations, expansion, override checks) throw errors, the line/col shouldn't be messed up.

### DIFF
--- a/apollo-federation/src/sources/connect/validation/mod.rs
+++ b/apollo-federation/src/sources/connect/validation/mod.rs
@@ -131,15 +131,23 @@ pub fn validate(mut source_text: String, file_name: &str) -> ValidationResult {
     }
 
     // Auto-upgrade the schema as the _last_ step, so that error messages from earlier don't have
-    // incorrect line/col info.
+    // incorrect line/col info if we mess this up
     if schema_info.connect_link.spec() == ConnectSpec::V0_1 {
-        if let Some(replace_range) = schema_info.connect_link.directive().location() {
-            let mut new_link = schema_info.connect_link.clone();
-            new_link.set_spec(ConnectSpec::V0_2);
-            source_text.replace_range(
-                replace_range.offset()..replace_range.end_offset(),
-                &new_link.to_string(),
-            );
+        if let Some(version_range) =
+            schema_info
+                .connect_link
+                .directive()
+                .location()
+                .and_then(|link_range| {
+                    let version_offset = source_text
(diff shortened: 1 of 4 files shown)


apollo-federation/src/sources/connect/validation/mod.rs
```diff
@@ -131,15 +131,23 @@ pub fn validate(mut source_text: String, file_name: &str) -> ValidationResult {
...
-            source_text.replace_range(
-                replace_range.offset()..replace_range.end_offset(),
-                &new_link.to_string(),
-            );
+        if let Some(version_range) =
+            schema_info
+                .connect_link
+                .directive()
+                .location()
+                .and_then(|link_range| {
```
naomijub, line 141: What do you think about breaking it in 2 and_then/map, I feel it might improve readability:

```rust
.and_then(|link_range| 
    Some((link_range, source_text
        .get(link_range.offset()..link_range.end_offset())?
        .find(ConnectSpec::V0_1.as_str())?))
 ).map(|(link_range, version_offset)| {
    let start = link_range.offset() + version_offset;
    let end = start + ConnectSpec::V0_1.as_str().len();
    start..end
})
```

dylan-apollo: Subjective, but I find that one _harder_ to read because it's not super clear where `version_offset` is coming from. You have to find the comma and then figure out where the expression ends amidst the parens.

We could make the `and_then` part longer and still explicitly assign `version_offset` to its own var before returning the tuple, but then the code will look almost identical to the old version.

We could break the whole thing up into multiple lines outside of the `if`, maybe that would help? Something like this?

```rust
    let link_range = schema_info
        .connect_link
        .directive()
        .location();
    let version_offset = link_range.and_then(
        |link_range| {
            source_text
                .get(link_range.offset()..link_range.end_offset())?
                .find(ConnectSpec::V0_1.as_str())
        });
    
    if schema_info.connect_link.spec() == ConnectSpec::V0_1 {
        if let (Some(link_range), Some(version_offset)) = (link_range, version_offset) {
            let start = link_range.offset() + version_offset;
            let end = start + ConnectSpec::V0_1.as_str().len();
            source_text.replace_range(start..end, ConnectSpec::V0_2.as_str());
        } else { ... 
```

lennyburdette (approved):

PR #7419, opened 2025-05-06
[JSONSelection] Improve `$batch.a { b c }` input selection analysis

benjamn:
This PR changes `VariableReference::path: Vec<VariablePathPart>` to `VariableReference::selection: SelectionTrie`, introducing a new data structure called `SelectionTrie`, because a single variable reference may involve a nested structure of sub-selections, rather than a single linear path. Acknowledging this complexity allows us to capture more accurate `$batch` usage information, so all essential keys will be included in `@key(fields: "...")`.

There are a number of selection set-related data structures currently in use in this codebase, like `FieldSet` and `apollo_compiler::executable::SelectionSet` and `apollo-federation::operation::SelectionSet`, but `SelectionTrie` seemed worthwhile because it can track location information for all the field names it stores (see `SelectionTrie::key_ranges`), which did not seem to be an off-the-shelf feature of the other options.

Less important than location information but still relevant to this decision, `SelectionTrie` supports more efficient `extend`/`merge` operations than the other options, because it stores `SelectionTrie::selections` as an `IndexMap` rather than a `Vec`, and because the values of that map are `Arc<SelectionTrie>` (note: `Rc` would work too), enabling cheap copy-on-write semantics when merging one trie into another.

This implementation does not yet understand `$batch` usage that happens _after_ `->` method calls, so `$batch->first.id` probably will not analyze correctly (that is, like `$batch.id->first`, producing a selection of `$batch { id }`), until we have a more robust way to record all input names while processing method shapes. If every `Shape` knew what part of the input it came from, we could lean more heavily on that metadata, but that depends on https://github.com/apollographql/shape-rs/pull/25 which we decided should not block the improvements in this PR.

### DIFF
--- a/apollo-federation/src/sources/connect/json_selection/helpers.rs
+++ b/apollo-federation/src/sources/connect/json_selection/helpers.rs
@@ -288,6 +288,10 @@ mod tests {
         assert!(is_identifier("hello"));
         assert!(is_identifier("hello_world"));
         assert!(is_identifier("hello_world_123"));
+        assert!(is_identifier("_hello_world"));
+        assert!(is_identifier("hello_world_"));
+        assert!(is_identifier("__hello_world"));
+        assert!(is_identifier("__hello_world__"));
         assert!(!is_identifier("hello world"));
         assert!(!is_identifier("hello-world"));
         assert!(!is_identifier("123hello"));
--- a/apollo-federation/src/sources/connect/json_selection/helpers.rs
+++ b/apollo-federation/src/sources/connect/json_selection/helpers.rs
@@ -5,6 +5,7 @@ use serde_json_bytes::Map as JSONMap;
 use serde_json_bytes::Value as JSON;
 
 use super::ParseResult;
+use super::is_identifier;
 use super::location::Span;
 use super::location::WithRange;
 
@@ -145,6 +146,21 @@ pub(crate) fn json_merge(a: Option<&JSON>, b: Option<&JSON>) -> (Option<JSON>, V
     }
 }
 
+pub(crate) fn quote_if_necessary(input: &str) -> String {
+    if is_identifier(input)
+        || (
+            // We also allow unquoted variable syntax, including $, @, and
+            // $identifier.
+            input == "@"
+                || input.starts_with('$') && (input.len() == 1 || is_identifier(&input[1..]))
+        )
+    {
+        input.to_string()
+    } else {
+        serde_json_bytes::Value::String(input.into()).to_string()
+    }
+}
+
 /// A helper to call `assert_snapshot!` without prepending the module, since prepending the
 /// module makes all the paths in json_selection tests too long for Windows.
 #[cfg(test)]
@@ -172,6 +188,7 @@ macro_rules! assert_debug_snapshot {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::sources::connect::json_selection::is_identifier;
     use crate::sources::connect::json_selection::location::new_span;
 
     #[test]
@@ -264,5 +281,37 @@ mod tests {
             # Another comment
         ", "
             "));
+    }
+
+    #[test]
+    fn test_is_identifier() {
+        assert!(is_identifier("hello"));
+        assert!(is_identifier("hello_world"));
+        assert!(is_identifier("hello_world_123"));
+        assert!(!is_identifier("hello world"));
+        assert!(!is_identifier("hello-world"));
+        assert!(!is_identifier("123hello"));
+        assert!(!is_identifier("hello@world"));
+        assert!(!is_identifier("$hello"));
+        assert!(!is_identifier("hello$world"));
+        assert!(!is_identifier(" hello"));
+        assert!(!is_identifier("__hello_world  "));
+        assert!(!is_identifier(" hello_world_123 "));
+    }
+
+    #[test]
+    fn test_quote_if_necessary() {
+        assert_eq!(quote_if_necessary("hello"), "hello");
+        assert_eq!(quote_if_necessary("hello world"), "\"hello world\"");
+        assert_eq!(quote_if_necessary("hello-world"), "\"hello-world\"");
+        assert_eq!(quote_if_necessary("123hello"), "\"123hello\"");
+        assert_eq!(quote_if_necessary("$"), "$");
+        assert_eq!(quote_if_necessary("@"), "@");
+        assert_eq!(quote_if_necessary("$hello"), "$hello");
+        assert_eq!(quote_if_necessary("@asdf"), "\"@asdf\"");
+        assert_eq!(quote_if_necessary("as@df"), "\"as@df\"");
+        assert_eq!(quote_if_necessary("hello$world"), "\"hello$world\"");
+        assert_eq!(quote_if_necessary("hello world!"), "\"hello world!\"");
+        assert_eq!(quote_if_necessary("hello world!@#"), "\"hello world!@#\"");
     }
 }
--- a/apollo-federation/src/sources/connect/json_selection/selection_trie.rs
+++ b/apollo-federation/src/sources/connect/json_selection/selection_trie.rs
@@ -22,10 +22,12 @@ impl PathList {
     }
 }
 
+type Ref<T> = std::sync::Arc<T>;
+
 #[derive(Debug, Eq, Clone)]
 pub(crate) struct SelectionTrie {
     /// The top-level sub-selections of this [`SelectionTrie`].
-    selections: IndexMap<String, SelectionTrie>,
+    selections: IndexMap<String, Ref<SelectionTrie>>,
 
     /// Whether the path terminating at this [`SelectionTrie`] node was
     /// explicitly added to the trie.
@@ -105,11 +107,13 @@ impl SelectionTrie {
     }
 
     pub(crate) fn get(&self, key: impl Into<String>) -> Option<&SelectionTrie> {
-        self.selections.get(&key.into())
+        self.selections.get(&key.into()).map(|sub| sub.as_ref())
     }
 
     pub(crate) fn iter(&self) -> impl Iterator<Item = (&str, &SelectionTrie)> {
-        self.selections.iter().map(|(key, sub)| (key.as_str(), sub))
+        self.selections
+            .iter()
+            .map(|(key, sub)| (key.as_str(), sub.as_ref()))
     }
 
     pub(crate) fn key_ranges(&self, key: &str) -> impl Iterator<Item = Range<usize>> {
@@ -178,10 +182,22 @@ impl SelectionTrie {
         self
     }
 
-    pub(crate) fn add_selection_trie(&mut self, other: &SelectionTrie) -> &mut Self {
+    pub(crate) fn extend(&mut self, other: &SelectionTrie) -> &mut Self {
         for (key, sub) in other.selections.iter() {
-            self.add_str_with_ranges(key, other.key_ranges(key))
-                .add_selection_trie(sub);
+            if let Some(existing) = self.selections.get_mut(key) {
+                Ref::make_mut(existing).extend(sub);
+            } else {
+                // Because sub is an Arc, this clone should be much cheaper than
+                // inserting an empty trie and then recursively extending it
+                // while traversing sub.
+                self.selections.insert(key.clone(), sub.clone());
+            }
+            // Whether or not the key already existed, we update self.key_ranges
+            // the same way:
+            self.key_ranges
+                .entry(key.clone())
+                .or_default()
+                .extend(other.key_ranges(key));
         }
         if self.is_used() || other.is_used() {
             self.set_used()
@@ -201,9 +217,11 @@ impl SelectionTrie {
     }
 
     fn add_str(&mut self, key: &str) -> &mut Self {
-        self.selections
-            .entry(key.to_string())
-            .or_insert_with(SelectionTrie::new)
+        if !self.selections.contains_key(key) {
+            self.selections
+                .insert(key.to_string(), Ref::new(SelectionTrie::new()));
+        }
+        Ref::make_mut(self.selections.get_mut(key).expect("should exist"))
     }
 
     fn add_str_with_ranges(
--- a/apollo-federation/src/sources/connect/json_selection/selection_trie.rs
+++ b/apollo-federation/src/sources/connect/json_selection/selection_trie.rs
@@ -6,6 +6,7 @@ use std::ops::Range;
 use apollo_compiler::collections::IndexMap;
 use apollo_compiler::collections::IndexSet;
 
+use super::JSONSelection;
 use super::Key;
 use super::NamedSelection;
 use super::PathList;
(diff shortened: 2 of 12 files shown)


apollo-federation/src/sources/connect/json_selection/helpers.rs
```diff
@@ -265,4 +275,32 @@ mod tests {
...
+        assert!(is_identifier("hello_world"));
+        assert!(is_identifier("hello_world_123"));
+        assert!(!is_identifier("hello world"));
+        assert!(!is_identifier("hello-world"));
+        assert!(!is_identifier("123hello"));
+        assert!(!is_identifier("hello@world"));
+        assert!(!is_identifier("$hello"));
+        assert!(!is_identifier("hello$world"));
+        assert!(!is_identifier(" hello"));
+        assert!(!is_identifier("__hello_world  "));
```
naomijub, line 291: is `assert!(!is_identifier("_hello_world"))` valid? String starting with `_`

apollo-federation/src/sources/connect/json_selection/selection_trie.rs
```diff
@@ -0,0 +1,390 @@
...
+    }
+}
+
+impl PartialEq for SelectionTrie {
+    fn eq(&self, other: &Self) -> bool {
+        self.used == other.used && self.selections == other.selections
+    }
+}
+
+impl Hash for SelectionTrie {
```
naomijub, line 70: Can't you do the following here:
```rust

impl Hash for SelectionTrie {
    fn hash<H: Hasher>(&self, state: &mut H) {
        self.selections.hash(state);
    }
}
```

Should work

apollo-federation/src/sources/connect/json_selection/selection_trie.rs
```diff
@@ -0,0 +1,390 @@
...
+
+    pub(crate) fn keys(&self) -> impl Iterator<Item = &String> {
+        self.selections.keys()
+    }
+
+    pub(crate) fn get(&self, key: impl Into<String>) -> Option<&SelectionTrie> {
+        self.selections.get(&key.into()).map(|sub| sub.as_ref())
+    }
+
+    pub(crate) fn iter(&self) -> impl Iterator<Item = (&str, &SelectionTrie)> {
```
naomijub, line 113: here you can probably just implement the iterator trait https://doc.rust-lang.org/rust-by-example/trait/iter.html

apollo-federation/src/sources/connect/json_selection/selection_trie.rs
```diff
@@ -0,0 +1,390 @@
...
+    pub(crate) fn has_str_path<'a>(&self, path: impl IntoIterator<Item = &'a str>) -> bool {
+        let mut current = self;
+        for key in path {
+            if let Some(sub) = current.get(key) {
+                current = sub;
+            } else {
+                return false;
+            }
+        }
+        current.is_used()
```
naomijub, line 137: I don't know exactly what this function is trying to do, but I guess one of this functions would be more functional:

```rust
pub(crate) fn has_str_path<'a>(&self, path: impl IntoIterator<Item = &'a str>) -> bool {
    path.any(|key| current.get(key).map(|sub| sub.is_used().unwrap_or_default())
}

pub(crate) fn has_str_path<'a>(&self, path: impl IntoIterator<Item = &'a str>) -> bool {
    path.all(|key| current.get(key).map(|sub| sub.is_used().unwrap_or_default())
}
```

apollo-federation/src/sources/connect/json_selection/selection_trie.rs
```diff
@@ -0,0 +1,390 @@
...
+
+    pub(crate) fn key_ranges(&self, key: &str) -> impl Iterator<Item = Range<usize>> {
+        self.key_ranges
+            .get(key)
+            .into_iter()
+            .flat_map(|ranges| ranges.iter())
+            .cloned()
+    }
+
+    #[allow(dead_code)]
```
naomijub, line 127: `#[expect(dead_code)]`, when using expect it will alert you once this is not the case anymore, so you will be able to remove the warning.

apollo-federation/src/sources/connect/json_selection/selection_trie.rs
```diff
@@ -0,0 +1,390 @@
...
+            if let Some(sub) = current.get(key) {
+                current = sub;
+            } else {
+                return false;
+            }
+        }
+        current.is_used()
+    }
+
+    #[allow(dead_code)]
```
naomijub, line 140: expect

apollo-federation/src/sources/connect/json_selection/selection_trie.rs
```diff
@@ -0,0 +1,390 @@
...
+    pub(crate) fn add_str_path<'a>(
+        &mut self,
+        path: impl IntoIterator<Item = &'a str>,
+    ) -> &mut Self {
+        path.into_iter()
```
naomijub, line 145: maybe replace with https://doc.rust-lang.org/std/iter/trait.Extend.html

apollo-federation/src/sources/connect/json_selection/selection_trie.rs
```diff
@@ -0,0 +1,390 @@
...
+            // unhandled PathList element.
+            _ => self.set_used(),
+        }
+    }
+
+    pub(crate) fn add_subselection(&mut self, sub: &SubSelection) -> &mut Self {
+        for selection in sub.selections_iter() {
+            match selection {
+                NamedSelection::Field(_, key, nested_selection) => {
+                    let result = self.add_key(key);
```
naomijub, line 167: where is this used after if else? feels like dead code

apollo-federation/src/sources/connect/models.rs
```diff
@@ -365,29 +365,20 @@ fn determine_entity_resolver(
 fn extract_header_references(
     variable_references: HashSet<VariableReference<Namespace>>,
 ) -> HashSet<String> {
-    let headers: HashSet<String> = variable_references
+    variable_references
         .iter()
-        .filter_map(|var_ref| {
+        .flat_map(|var_ref| {
```
naomijub, line 370:
```rust
.flat_map(|var_ref| var_ref
  .selection.get("headers")
  .map(|headers_subselection| 
    headers_subselection.keys()
      .cloned()
      .collect()
  ).unwrap_or_default()
)
```

apollo-federation/src/sources/connect/models.rs
```diff
@@ -365,29 +365,20 @@ fn determine_entity_resolver(
...
-                .path
-                .first()
-                .is_none_or(|path| path.part != "headers")
-            {
-                return None;
+                vec![]
+            } else if let Some(headers_subselection) = var_ref.selection.get("headers") {
+                headers_subselection.keys().cloned().collect()
+            } else {
+                vec![]
```
naomijub, line 378: generally avoid `vec![]` as it is not const, prefer `Vec::new()` as it is const, https://doc.rust-lang.org/std/vec/struct.Vec.html#method.new

apollo-federation/src/sources/connect/validation/connect/selection/variables.rs
```diff
@@ -234,29 +235,29 @@ impl NamespaceResolver for ArgsResolver<'_> {
...
+                        "`{object}` does not have an argument named `{root}`",
+                        object = self.field.name,
+                    ),
+                    locations: reference
+                        .selection
+                        .key_ranges(root)
+                        .flat_map(|range| expression.line_col_for_subslice(range, schema))
+                        .collect(),
+                })
+                .map(|field| field.ty.clone())?;
```
naomijub, line 256: maybe `map` before the `ok_or_else`? Know the action before defining the error case

apollo-federation/src/sources/connect/variable.rs
```diff
@@ -144,24 +145,25 @@ impl Display for Namespace {
...
 }
 
-impl<N: FromStr + ToString> Display for VariableReference<'_, N> {
+impl<N: FromStr + ToString> Display for VariableReference<N> {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
         f.write_str(self.namespace.namespace.to_string().as_str())?;
-        for part in &self.path {
-            f.write_str(".")?;
-            f.write_str(part.as_str())?;
+        if !self.selection.is_empty() {
```
naomijub, line 163: `write!(f, " {{ {} }} ", self.selection.to_string().as_str())` should work as well

apollo-federation/src/sources/connect/json_selection/helpers.rs
benjamn: Yes, but I'll add some tests.

apollo-federation/src/sources/connect/json_selection/selection_trie.rs
benjamn: That would work if `IndexMap` implemented `Hash`! 😞  See https://github.com/indexmap-rs/indexmap/issues/67 if you don't mind being a little disappointed.

I personally think `IndexMap` should support an implementation of `Hash` out of the box. Even though `IndexMap` preserves insertion order, `IndexMap`'s implementation of `PartialEq` is order-insensitive, so I think it also can/should implement `Hash` with an order-insensitive combination operator (note: XOR/`^` may not be the best choice here). Until the maintainers come around and implement `Hash` natively, my workaround seems necessary.

Now, we don't have to use `IndexMap` here, especially if we end up normalizing tries by sorting their keys (or something). But if we're using `IndexMap` then we need to hash it manually, I think.

apollo-federation/src/sources/connect/json_selection/selection_trie.rs
benjamn: I'm probably misunderstanding the suggestion, but I don't really want to express the iteration in terms of a stateful `next` method. Is there anything problematic about the way I've written it here?

apollo-federation/src/sources/connect/json_selection/selection_trie.rs
benjamn: I'm already using `SelectionTrie::extend(&mut self, other: &SelectionTrie) -> &mut Self` for (destructively) merging `other` into `self`, and think I'm happy with the clarity of `add_str_path`, as it stands. It's an easy method to test, but it's not really the workhorse of this system, because we generally want to have a chance of collecting location information along with the field names stored in the trie, and you can only provide the field names themselves with `add_str_path`.

apollo-federation/src/sources/connect/json_selection/selection_trie.rs
benjamn: Definitely not dead code: `self.add_key(key)` returns a mutable reference to the child `SelectionTrie` that was either found or created, so the `result.add_subselection(nested)` and `result.set_leaf()` calls are updating it.

apollo-federation/src/sources/connect/validation/connect/selection/variables.rs
benjamn: That code structure was already here, and I'm mostly just reindenting it, but I will try putting `map` first.

apollo-federation/src/sources/connect/models.rs
benjamn: How would you include checking that the `Namespace` is either `$request` or `$response`?

apollo-federation/src/sources/connect/json_selection/selection_trie.rs
benjamn: I tried this but it seems `#[expect(dead_code)]` fails because the code is used in tests, so I went with `#[cfg(test)]` instead to reflect that usage more accurately (and to keep the methods from being used otherwise, for now).

apollo-federation/src/sources/connect/json_selection/selection_trie.rs
benjamn: https://github.com/apollographql/router/pull/7419#discussion_r2080086943

apollo-federation/src/sources/connect/json_selection/selection_trie.rs
benjamn: Commit renaming `used` to `is_leaf`: https://github.com/apollographql/router/pull/7419/commits/cc60897a8c58c78463f1b0db4ec5a9ab0ad5a1a4

apollo-federation/src/sources/connect/variable.rs
benjamn: https://github.com/apollographql/router/pull/7419/commits/6bf390b54abe2d4d7e5c9ab6556d54f0b25035d3

apollo-federation/src/sources/connect/validation/connect/selection/variables.rs
benjamn: https://github.com/apollographql/router/pull/7419/commits/b62dfa0d61a4df34dd342adac0de8ad43e1031bf

apollo-federation/src/sources/connect/json_selection/selection_trie.rs
benjamn: https://github.com/apollographql/router/pull/7419/commits/4f0831607a87a0419c1ea5486eb9a122fcaea305

apollo-federation/src/sources/connect/json_selection/helpers.rs
benjamn: https://github.com/apollographql/router/pull/7419/commits/74b26d22201156f083b54de83de2496ff3d313b7

apollo-federation/src/sources/connect/models.rs
benjamn: Ah, I see what you mean: https://github.com/apollographql/router/pull/7419/commits/d7e8d5e70b57f1d94ea5d117506c2cc4b235b031

apollo-federation/src/sources/connect/models.rs
benjamn: https://github.com/apollographql/router/pull/7419/commits/d7e8d5e70b57f1d94ea5d117506c2cc4b235b031

apollo-federation/src/sources/connect/json_selection/selection_trie.rs
```diff
@@ -0,0 +1,448 @@
...
+
+        use super::ExternalVarPaths;
+        for path in self.external_var_paths() {
+            match path.path.as_ref() {
+                PathList::Var(known_var, tail) => {
+                    trie.add_str(known_var.as_str())
+                        .add_path_list(tail.as_ref());
+                }
+                _ => {
+                    // The self.external_var_paths() method should only return
```
Hunter-apollo, line 40: if so, could this be reduced to an if-let (if we dont want to fail on that) or let-else (if we do want to fail on non-var returns)

I.e.
```
if let PathList::Var(known_var, tail)  = path.path.as_ref() {
      trie.add_str(known_var.as_str()).add_path_list(tail.as_ref());
}
```
or 

```
let PathList::Var(known_var, tail)  = path.path.as_ref() else {
      unreachable!("Should never have non-var types");
}
trie.add_str(known_var.as_str()).add_path_list(tail.as_ref());
```

benjamn: Thanks! https://github.com/apollographql/router/pull/7419/commits/4170e584eb2d2ec90fa881ed62d38268137ff003

lennyburdette (approved):

PR #7429, opened 2025-05-07
Add Enabled Features List to Telemetry Plugin

rregitsky:
Adds a list of enabled features to the `Telemetry` plugin which will be emitted with apollo protobuf in a future PR. This was achieved by funneling the full router config through plugin creation _only for the Telemetry plugin_. All other plugins will have an `Option::None` for the `full_config` attribute on `PluginInit`.

This PR also temporarily annotates the `EnabledFeatures` struct to allow for dead code until it can be used in a follow-up PR.

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/plugin/mod.rs
```diff
@@ -85,14 +86,23 @@ pub struct PluginInit<T> {
 
     /// User's license's state, including any limits of use
     pub(crate) license: LicenseState,
+
+    /// The full router configuration json for use by the telemetry plugin ONLY.
+    /// NEVER use this in any other plugin. Plugins should only ever access their pre-defined
+    /// configuration subset.
+    pub(crate) full_config: Option<Value>,
```
goto-bus-stop, line 93: Can this be the typed `Configuration` object? (or is that a circular dependency?)

apollo-router/src/plugins/coprocessor/test.rs
```diff
@@ -61,6 +61,7 @@ mod tests {
                 "thisFieldDoesntExist": true
             }
         });
+        // TODO this will now fail on the configuration build b/c we do validation... Keep test or kill it?
```
goto-bus-stop, line 64: Does that mean `.configuration_json()` now returns an error? in that case i'd say keep the test but check is_err() on the result of `.configuration_json()` without building a full router

apollo-router/src/test_harness.rs
```diff
@@ -153,18 +155,21 @@ impl<'a> TestHarness<'a> {
         self,
         configuration: serde_json::Value,
     ) -> Result<Self, serde_json::Error> {
-        let configuration: Configuration = serde_json::from_value(configuration)?;
+        // Convert from a json Value to yaml str to Configuration so that we can ensure we validate
+        // and populate the Configuration's validated_yaml attribute
```
goto-bus-stop, line 159: A cleaner way to do this may be to split the YAML parsing and the validation parts in `crate::configuration::schema::validate_yaml_configuration`, and add a new `Configuration::from_value` constructor that takes a `serde_yaml::Value`. I think you can do something like `serde_yaml::to_value(serde_json::json!({ "json": "object" }))` to convert between the JSON and YAML types, because the `serde_json::Value` and `serde_yaml::Value` types are both themselves serializable and deserializable.

that said, not a dealbreaker, I think. It's not bad to do a stringify+parse.

goto-bus-stop (commented):
Just some initial feedback, I think this is pretty sensible in general (and I'm maybe not sure that we even have to limit it to the telemetry plugin, but it's better to start limited and expand later than the other way around 😛 )

apollo-router/src/plugin/mod.rs
rregitsky: That was my original implementation, but @BrynCooke suggested using the Value coming from `validated_yaml` in the [ticket description](https://apollographql.atlassian.net/browse/ROUTER-1284). I'm not certain of his reasoning since he's been OOO, but I imagine the validation aspect was what he was aiming for?

apollo-router/src/plugins/coprocessor/test.rs
rregitsky: I'm just not sure we're following the intention of the test if we remove router building. It essentially becomes a unit test of `Configuration::from_str()`. Maybe that's fine 🤷 I'll change it for now, but open to more discussion if what I said sparks anything.

apollo-router/src/test_harness.rs
rregitsky: I started going through the process of splitting things up, but I discovered `validate_yaml_configuration()` is parsing the raw yaml string in two ways: into `serde_json:Value` and into `MarkedYaml`. That makes splitting this a bit too complicated for my taste if we're ok with the alternative approach.

apollo-router/src/plugin/mod.rs
BrynCooke: I had a go at converting to configuration and the fallout was massive. Let's stick with the validated config.

BrynCooke (approved):

goto-bus-stop (approved):
As I mentioned to @BrynCooke in the router huddle, I don't have a real opinion on how this should be done, so if the tests are happy I'm happy. Just some stylistic points in the comments above but nothing blocking

apollo-router/src/plugins/telemetry/metrics/apollo/mod.rs
```diff
@@ -237,31 +235,29 @@ mod test {
...
-        })
-        .await?;
+        let config = r#"
+            telemetry:
+              apollo:
+                endpoint: "http://example.com"
+                client_name_header: "name_header"
+                client_version_header: "version_header"
+                buffer_size: 10000
+                schema_id: "schema_sha"
```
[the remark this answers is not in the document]
BrynCooke: I think it's a separate thing that we should follow up with. The schema is now available via plugin init, so there's no reason to have it anymore.

apollo-router/src/plugins/telemetry/mod.rs
```diff
@@ -1701,6 +1717,15 @@ impl Telemetry {
             }
         }
     }
+
+    fn extract_enabled_features(full_config: &serde_json::Value) -> EnabledFeatures {
```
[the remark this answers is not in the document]
BrynCooke: Yeah, I thought it would panic also, but no it's fine :)

apollo-router/src/plugins/telemetry/metrics/apollo/mod.rs
```diff
@@ -433,30 +429,42 @@ mod test {
...
-            buffer_size: default_buffer_size(),
-            schema_id: "schema_sha".to_string(),
-            ..Default::default()
-        })
+    fn create_telemetry_plugin() -> impl Future<Output = Result<Telemetry, BoxError>> {
+        let config = format!(
+            r#"
+            telemetry:
+              apollo:
+                endpoint: "{endpoint}"
```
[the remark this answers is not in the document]
rregitsky: Are you worried about the removal of the `Url::parse()`? Maybe I'm misunderstanding the problem. Wouldn't we be deserializing a normal config file this way too?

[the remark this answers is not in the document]
BrynCooke: I think let's leave thsi for now. If I follow up to remove schema_id from the config we can fix this properly by externalizing the config to a yaml file.

PR #7431, opened 2025-05-08
deps: update wiremock

goto-bus-stop:
Updates wiremock to 0.6, removing several dependencies.

One test is changed. A connectors test was asserting that inserting a header with comma-separated values was received as _multiple headers_.
This is not an HTTP feature, but was a quirk of wiremock 0.5. It's removed in 0.6.

https://github.com/LukeMathWalker/wiremock-rs/pull/128/files#diff-d66a7ae966da7f4a6abe13626b8c8793b52f2d81fe85799024f57a14150ababcL146-R120

### DIFF
--- a/apollo-router/Cargo.toml
+++ b/apollo-router/Cargo.toml
@@ -227,7 +227,7 @@ url = { version = "2.5.4", features = ["serde"] }
 urlencoding = "2.1.3"
 uuid = { version = "1.9.1", features = ["serde", "v4"] }
 yaml-rust = "0.4.5"
-wiremock = "0.5.22"
+wiremock = "0.6"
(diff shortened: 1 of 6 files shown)


apollo-router/Cargo.toml
```diff
@@ -97,7 +97,6 @@ futures = { version = "0.3.30", features = ["thread-pool"] }
 graphql_client = "0.14.0"
 hex.workspace = true
 http.workspace = true
-http-0_2 = { version = "0.2.12", package = "http" }
```
goto-bus-stop, line 100: turns out this was the only remaining place where we used http 0.2 directly!

lennyburdette (approved):
thank you!

PR #7432, opened 2025-05-08
deps: clean up some duplicate and unused dependencies

goto-bus-stop:
1. Removes an unused hyper 0.14 dependency. Now, only one hyper version exists in our tree!
2. Updates `opentelemetry-proto` to align with other otel dependencies. This removes several duplicate otel and protobuf related dependencies

A snapshot changed because of 2). I can't tell for certain if this is because of insta, prost, or opentelemetry-proto. But it's because the `int_value` proto field is an int64, which is arguably not round-trippable in YAML (...assuming a JSON/JS style number implementation where every number is float64). So sometimes int64s have to be serialized as strings.
I think this is an acceptable change, because it's only asserting how the test parses values from serialized protobuf and then serialized to YAML; if it's parsed into a string, or an int, it doesn't really matter, because both still tell us if the serialized protobuf contained the correct integer value.

Now, the only thing that still uses the http 0.x crates in our dependency tree is the AWS SDK. We can now likely update the AWS SDK to get rid of this. IIRC, the last time i tried to, i ran into some dependency conflict with wiremock, which is resolved as of #7431.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -312,7 +312,7 @@ dependencies = [
  "http-serde",
  "humantime",
  "humantime-serde",
- "hyper 1.6.0",
+ "hyper",
  "hyper-rustls",
  "hyper-util",
  "hyperlocal",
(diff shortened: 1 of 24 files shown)


conwuegb (approved):
LGTM

goto-bus-stop:
Test failures look related. Maybe the int/string difference is material after all. might look next week

carodewig (approved):

PR #7436, opened 2025-05-08
Add FederationError variant for subgraph errors

sachindshinde:
There are some places in composition code that need to create a `FederationError`, but annotated with subgraph name. Ideally we would split federation errors into internal errors, query planning errors, and composition errors, and give composition errors the ability to be annotated with subgraph name. But that's further out, so for now, this PR tries to introduce a small changeset for allowing subgraph name annotation. If there's a nicer approach with similar size/simplicity, let me know in review and I'll switch to it.

### DIFF
--- a/apollo-federation/src/error/mod.rs
+++ b/apollo-federation/src/error/mod.rs
@@ -622,6 +622,8 @@ impl SingleFederationError {
         }
     }
 
+    // TODO: This logic is here to avoid accidentally nesting subgraph name annotations, in the
+    // future we should change the composition error type to make nesting impossible.
     pub(crate) fn unwrap_subgraph_error(self, expected_subgraph: &str) -> Self {
         if let Self::SubgraphError { subgraph, error } = self {
             debug_assert_eq!(
@@ -635,6 +637,8 @@ impl SingleFederationError {
         }
     }
 
+    // TODO: This logic is here to avoid accidentally nesting subgraph name annotations, in the
+    // future we should change the composition error type to make nesting impossible.
     pub(crate) fn add_subgraph(self, subgraph: String) -> Self {
         if let Self::SubgraphError {
             subgraph: existing_subgraph,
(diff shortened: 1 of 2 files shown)


conwuegb (approved):

apollo-federation/src/error/mod.rs
```diff
@@ -126,6 +126,11 @@ pub enum SingleFederationError {
     // This is a known bug that will take time to fix, and does not require reporting.
     #[error("{message}")]
     InternalUnmergeableFields { message: String },
+    #[error("[{subgraph}] {error}")]
+    SubgraphError {
```
TylerBloom, line 130: When we split QP and composition error variants, is this variant only for composition?

apollo-federation/src/error/mod.rs
```diff
@@ -615,6 +621,39 @@ impl SingleFederationError {
             },
         }
     }
+
+    pub(crate) fn unwrap_subgraph_error(self, expected_subgraph: &str) -> Self {
```
TylerBloom, line 625: Can we leave a doc string here to explain that this method is to avoid having the type recur more than once? And that, ideally, the composition error in the future doesn't need to recur at all?

apollo-federation/src/error/mod.rs
sachindshinde: Yep, this is only for composition.

duckki (approved):
Makes sense as a temporary solution. We can review `SingleFederationError::SubgraphError` cases later. I hope we eventually separate them out as `CompositionError`.

apollo-federation/src/error/mod.rs
sachindshinde: Makes sense, added the TODO/doc strings in https://github.com/apollographql/router/pull/7436/commits/7a6e4b4fa192aef65895bb1c8bdc6c6c3f10def8

PR #7451, opened 2025-05-09
Implement post-merge validations

sachindshinde:
This PR implements post-merge validations for composition. Note that:
- These validations are placed in `apollo-federation/src/schema/validators` instead of on `Merger` as would be the case in JS, since (as the TODO notes) there's logic shared with subgraph validations that we'd like to refactor out in the future. (Also, the current Rust `Merger` struct is for the old composition code.)
- Unrelatedly, we also move interface object validations to `apollo-federation/src/schema/validators` to adhere to existing patterns.

### DIFF
--- a/apollo-federation/src/schema/validators/interface_object.rs
+++ b/apollo-federation/src/schema/validators/interface_object.rs
@@ -1,0 +1,167 @@
+use crate::error::FederationError;
+use crate::error::MultipleFederationErrors;
+use crate::error::SingleFederationError;
+use crate::schema::ValidFederationSchema;
+use crate::schema::field_set::parse_field_set;
+use crate::schema::position::InterfaceTypeDefinitionPosition;
+use crate::schema::subgraph_metadata::SubgraphMetadata;
+use crate::utils::human_readable::HumanReadableListOptions;
+use crate::utils::human_readable::HumanReadableListPrefix;
+use crate::utils::human_readable::human_readable_list;
+
+pub(crate) fn validate_interface_object_directives(
+    schema: &ValidFederationSchema,
+    metadata: &SubgraphMetadata,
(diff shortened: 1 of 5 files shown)


tninesling (commented):
Are there any tests we can add for the new validation logic?

apollo-federation/src/schema/validators/interface_object.rs
```diff
@@ -0,0 +1,167 @@
...
+use crate::utils::human_readable::HumanReadableListOptions;
+use crate::utils::human_readable::HumanReadableListPrefix;
+use crate::utils::human_readable::human_readable_list;
+
+pub(crate) fn validate_interface_object_directives(
+    schema: &ValidFederationSchema,
+    metadata: &SubgraphMetadata,
+    errors: &mut MultipleFederationErrors,
+) -> Result<(), FederationError> {
+    validate_keys_on_interfaces_are_also_on_all_implementations(schema, metadata, errors)?;
```
duckki, line 17: Strictly speaking, `validate_keys_on_interfaces_are_also_on_all_implementations` is not about interface object. 😓  Not a big deal, though.

sachindshinde: Interface `@key`s (a.k.a. [entity interfaces](https://www.apollographql.com/docs/graphos/schema-design/federated-schemas/entities/interfaces)) were a feature introduced specifically to facilitate interface objects, which is why they're grouped together (changes to one will need to be considered in tandem with the other).

sachindshinde:
@tninesling 
> Are there any tests we can add for the new validation logic?

There are relevant tests in `compose.test.ts` to port, but we're going to need the merge logic finished to really test it, so I've created FED-613 to track those tests.

PR #7454, opened 2025-05-09
federation: implement composition stubs

dariuszkuc:
Initial version of overall composition process with placeholders for various composition phases.

Composition process consists of following phases
1. expand subgraphs - add missing federation definitions to the schema
2. upgrade subgraphs - upgrade fed v1 subgraphs to be compatible with fed v2
3. validate subgraphs - run subgraph validations
4. merge subgraphs - generates supergraph
4. validate satisfiability - ensure we can generate plan for any query

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-federation/src/composition/mod.rs
+++ b/apollo-federation/src/composition/mod.rs
@@ -3,7 +3,7 @@ mod satisfiability;
 use std::vec;
 
 use crate::error::FederationError;
-use crate::schema::schema_upgrader::upgrade_subgraphs_if_necessary;
+pub use crate::schema::schema_upgrader::upgrade_subgraphs_if_necessary;
 use crate::subgraph::typestate::Expanded;
 use crate::subgraph::typestate::Initial;
 use crate::subgraph::typestate::Subgraph;
--- a/apollo-federation/src/composition/mod.rs
+++ b/apollo-federation/src/composition/mod.rs
@@ -2,6 +2,7 @@ mod satisfiability;
 
 use std::vec;
 
+use crate::composition::satisfiability::validate_satisfiability;
 use crate::error::FederationError;
 pub use crate::schema::schema_upgrader::upgrade_subgraphs_if_necessary;
 use crate::subgraph::typestate::Expanded;
@@ -9,6 +10,7 @@ use crate::subgraph::typestate::Initial;
 use crate::subgraph::typestate::Subgraph;
 use crate::subgraph::typestate::Upgraded;
 use crate::subgraph::typestate::Validated;
+use crate::supergraph::Merged;
 use crate::supergraph::Satisfiable;
 use crate::supergraph::Supergraph;
 
@@ -17,11 +19,10 @@ pub fn compose(
 ) -> Result<Supergraph<Satisfiable>, Vec<FederationError>> {
     let expanded_subgraphs = expand_subgraphs(subgraphs)?;
     let upgraded_subgraphs = upgrade_subgraphs_if_necessary(expanded_subgraphs)?;
-    let _validated_subgraphs = validate_subgraphs(upgraded_subgraphs)?;
+    let validated_subgraphs = validate_subgraphs(upgraded_subgraphs)?;
 
-    todo!("implement magic here")
-    // let supergraph = merge_subgraphs(validated_subgraphs)?;
-    // validate_satisfiability(supergraph)?
+    let supergraph = merge_subgraphs(validated_subgraphs)?;
+    validate_satisfiability(supergraph)
 }
 
 pub fn expand_subgraphs(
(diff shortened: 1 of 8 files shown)


dariuszkuc:
NOTE: whole error handling and the `FederationError` -> `SubgraphError` -> `FederationError` should be fixed with future PRs where we separate the `CompositionError`s from `FederationError`s

apollo-federation/src/composition/mod.rs
```diff
@@ -0,0 +1,57 @@
...
+use crate::subgraph::typestate::Upgraded;
+use crate::subgraph::typestate::Validated;
+use crate::supergraph::Satisfiable;
+use crate::supergraph::Supergraph;
+
+pub fn compose(
+    subgraphs: Vec<Subgraph<Initial>>,
+) -> Result<Supergraph<Satisfiable>, Vec<FederationError>> {
+    let expanded_subgraphs = expand_subgraphs(subgraphs)?;
+    let upgraded_subgraphs = upgrade_subgraphs_if_necessary(expanded_subgraphs)?;
```
dariuszkuc, line 19: should we move this function over here as well so public API is in one place?

apollo-federation/src/composition/mod.rs
```diff
@@ -0,0 +1,57 @@
...
+    let _validated_subgraphs = validate_subgraphs(upgraded_subgraphs)?;
+
+    todo!("implement magic here")
+    // let supergraph = merge_subgraphs(validated_subgraphs)?;
+    // validate_satisfiability(supergraph)?
```
TylerBloom, line 24: Don't `merge_subgraphs` and `validate_satisfiability` both panic already? If so, can we remove this `todo!` and just let those functions panic (and remove the `allow_unused`)?

dariuszkuc: indeed there is an existing `merge_subgraphs` function that was created 1.5 years ago (that connectors was relying on)... but that needs to be updated to use the typestate stuff and didn't want to touch it yet (as it is a live code)

dariuszkuc: I'll create new stub for `merge_subgraphs`

apollo-federation/src/supergraph/mod.rs
```diff
@@ -83,6 +87,64 @@ use crate::schema::type_and_directive_specification::TypeAndDirectiveSpecificati
...
+        &self,
+        options: ApiSchemaOptions,
+    ) -> Result<ValidFederationSchema, FederationError> {
+        api_schema::to_api_schema(self.state.schema.clone(), options)
+    }
+}
+
+#[derive(Clone, Debug)]
+#[allow(unused)]
+pub struct Merged {
```
dariuszkuc, line 115: I liked the original `Initial` -> `Validated` states.... but since we define the same subgraph states then we would have to rename those whenever importing both states....

apollo-federation/src/composition/mod.rs
dariuszkuc: 4ca6564e535b7390a006615389662d66e59648aa

duckki (commented):
Looks reasonable. Just a few nits (not sure why the `link_handling_tests` module keeps popping up).

apollo-federation/src/subgraph/typestate.rs
```diff
@@ -914,3 +966,157 @@ mod tests {
         );
     }
 }
+
+// PORT_NOTE: Corresponds to '@core/@link handling' tests in JS
+#[cfg(test)]
+mod link_handling_tests {
```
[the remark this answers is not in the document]
dariuszkuc: this is probably from a rebase, i'll clean it up

duckki (approved):

apollo-federation/src/schema/schema_upgrader.rs
```diff
@@ -25,308 +23,160 @@ use crate::error::FederationError;
...
-            schema,
-            expanded_info: ExpandedSubgraphInfo {
-                subgraph_name: original_subgraph.name.clone(),
-                subgraph_url: original_subgraph.url.clone(),
-                key_directive_name: original_subgraph.key_directive_name()?.clone(),
-                requires_directive_name: original_subgraph.requires_directive_name()?.clone(),
-                provides_directive_name: original_subgraph.provides_directive_name()?.clone(),
-                extends_directive_name: original_subgraph.extends_directive_name()?.clone(),
-            },
+            .map(|subgraph| (subgraph.name.clone(), subgraph.clone()))
```
clenfest, line 88: On the whole, I like the reorganization, but the downside is that we're introducing an extra clone for all schemas here(even those that are not fed1) which the old code was avoiding.

dariuszkuc: Indeed but it also keeps the code "cleaner". If this turns out a performance problem we can always optimize later.

PR #7469, opened 2025-05-12
[Batch] Add validation that Batch keys exist in output selection

Hunter-apollo:
Adds validation that batch keys match the output selection keys. 

Batch connectors should always have all batch keys (i.e. `$batch.this_is_a_key`) repeated in the output selection. 

For example, if a users specifies a batch connector with the following URL template: `"/users?ids={$batch.id->joinNotNull(',')}"` must have a corresponding `id` key in the connector selection (i.e. `selection: "$.results { id: foo.bar }"`). 

We do not permit the use of external variables such as `$context`, `$request`, etc to provide the required batch keys. The validation in this PR will emit an error message in such cases.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [X] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-federation/src/sources/connect/validation/schema.rs
```diff
@@ -327,14 +345,145 @@ fn advanced_validations(schema: &SchemaInfo, subgraph_name: &str) -> Vec<Message
     entity_checker.check_for_missing_entity_connectors(schema)
 }
 
+fn compute_input_trie(connector: &Connector) -> SelectionTrie {
+    let mut merged = SelectionTrie::new();
+    for param in connector.variable_references() {
+        merged.extend(&param.selection);
+    }
+    merged
+}
```
benjamn, line 354:
```suggestion
fn compute_input_trie(connector: &Connector) -> SelectionTrie {
    let mut merged = SelectionTrie::new();
    for param in connector.variable_references() {
        merged.add_str(param.namespace.as_str()).extend(&param.selection);
    }
    merged
}
```
I may be missing how this is accounted for elsewhere, but I think you want to preserve the name of each variable at the top level of the trie (note `param.selection` only represents the selection set for that single variable, so it does not already include the variable name).

apollo-federation/src/sources/connect/validation/schema.rs
```diff
@@ -327,14 +345,145 @@ fn advanced_validations(schema: &SchemaInfo, subgraph_name: &str) -> Vec<Message
     entity_checker.check_for_missing_entity_connectors(schema)
 }
 
+fn compute_input_trie(connector: &Connector) -> SelectionTrie {
+    let mut merged = SelectionTrie::new();
+    for param in connector.variable_references() {
```
naomijub, line 350: this can probably be refactor in an iterator way:

```rust
connector.variable_references().iter().ma(|param| param.selection).collect::<SelectionTrie<_>>()
```

apollo-federation/src/sources/connect/validation/schema.rs
```diff
@@ -327,14 +345,145 @@ fn advanced_validations(schema: &SchemaInfo, subgraph_name: &str) -> Vec<Message
...
+                continue;
+            }
+
+            let Some(next_shape) = fields.get(key) else {
+                unreachable!("nested shape is expected")
+            };
+
+            // Check that next shape doesn't come from a non-`$root` field.
+            if let ShapeCase::Name(root, _) = next_shape.case() {
+                if root.value != "$root" {
```
naomijub, line 462: I think we have a const for this somewhere, if we don't, create a const for it:
`const ROOT: &str = "$root";`

apollo-federation/src/sources/connect/validation/schema.rs
benjamn: But… why? It's more code. Can't Rust generate efficient code for the `for` loop?

apollo-federation/src/sources/connect/validation/schema.rs
naomijub: I think `param.namespace.as_str()` is not valid

apollo-federation/src/sources/connect/validation/snapshots/validation_tests@batch__batch_missing_key.graphql.snap
```diff
@@ -0,0 +1,12 @@
+---
+source: apollo-federation/src/sources/connect/validation/mod.rs
+expression: "format!(\"{:#?}\", result.errors)"
+input_file: apollo-federation/src/sources/connect/validation/test_data/batch/batch_missing_key.graphql
+---
+[
+    Message {
+        code: ConnectorsUnresolvedField,
+        message: "The `@connect` directive on `User_0` specifies a batch entity resolver, but the key `id` is not a subset of the output shape `{ name: $root.results.*.name, profilePic: $root.results.*.profile_pic }`",
```
lennyburdette, line 9: couple of notes:

* i'm wary of having `"$root"` in messages because it's not a real variable and that could confuse people. 
* "the key" refers to a somewhat internal detail. 
* the output shape could be huge (consider [this selection](https://studio.apollographql.com/graph/TheSpaceDevs/variant/dev/schema/sdl?selectedSchema=docking-events#Query.dockingEvent) from Watson) so some truncation will be necessary.
* Do you know why it's `User_0` and not `User`?

maybe

```
The `@connect` directive on `User` specifies a batch entity resolver, but the fields `id` are not present in the selection. https://go.apollo.dev/connectors/batch
```

https://github.com/apollographql/shortlinks/blob/master/_redirects#L101

apollo-router/src/services/layers/persisted_queries/snapshots/apollo_router__services__layers__persisted_queries__tests__pq_layer_freeform_graphql_with_safelist_log_unknown_true@logs.snap
lennyburdette, line 1: i see this snapshot once in a while and have to manually reject it. not sure what the deal is

apollo-federation/src/sources/connect/models.rs
```diff
@@ -258,7 +258,7 @@ impl Connector {
         )
     }
 
-    /// Create a field set for a `@key` using $args and $this variables.
+    /// Create a field set for a `@key` using `$args` and `$this` variables.
```
lennyburdette, line 261: and `$batch` if you're correcting this! 😁

apollo-federation/src/sources/connect/validation/schema.rs
```diff
@@ -327,14 +345,145 @@ fn advanced_validations(schema: &SchemaInfo, subgraph_name: &str) -> Vec<Message
...
+        "The `@connect` directive on `{connector}` specifies a batch entity resolver, but the key `{unset}` is not a subset of the output shape `{output_shape}`"
+    )]
+    BatchKeyNotSubsetOfOutputShape {
+        connector: String,
+        unset: &'error String,
+        output_shape: &'error Shape,
+    },
+    #[error("Attempted to resolve key on unexpected shape `{shape_str}`")]
+    UnexpectedKeyOnShape { shape_str: String },
+    #[error("Batch keys cannot be mapped from non-`$root` sources (i.e. `$context`, `$this`)")]
```
lennyburdette, line 382: maybe something about how the fields matching the `$batch` variable must come from the API response? (again not exposing the term "$root" to users)

apollo-federation/src/sources/connect/validation/schema.rs
```diff
@@ -327,14 +345,145 @@ fn advanced_validations(schema: &SchemaInfo, subgraph_name: &str) -> Vec<Message
...
+        for (key, sub_selection) in self.trie.iter() {
+            // Object should contain all keys in the selection set.
+            // If not, then the key is unmapped.
+            if !fields.contains_key(key) {
+                self.unmapped_fields.insert(key.to_string());
+                continue;
+            }
+
+            let Some(next_shape) = fields.get(key) else {
+                unreachable!("nested shape is expected")
```
lennyburdette, line 457: maybe `return Err(...)` instead of panicking?

lennyburdette (commented):
this is great! happy to jam on error message wording whenever

apollo-federation/src/sources/connect/validation/schema.rs
benjamn: Sorry, I wrote the code above without testing it locally first. I hope the intent was clear enough, and that you feel empowered to remove `#[cfg(test)]` (or find a similar method) to get it working, according to your judgement.

apollo-federation/src/sources/connect/validation/schema.rs
Hunter-apollo: This is actually `unreachable` b/c of the above `contains_key` check. I think I could simplify the two into one statement for more clarity though

apollo-router/src/services/layers/persisted_queries/snapshots/apollo_router__services__layers__persisted_queries__tests__pq_layer_freeform_graphql_with_safelist_log_unknown_true@logs.snap
Hunter-apollo: Weird. I will remove.

apollo-federation/src/sources/connect/validation/snapshots/validation_tests@batch__batch_missing_key.graphql.snap
Hunter-apollo: - I will remove uses of "$root" in the error messages
- Field might be better.
- Agreed, but I think we may want to implement that on the Shape pretty print implementation rather than here.
- User_0 is because the synthetic name on the `ConnectorPosition` includes the index of the directive. I will add a method to get the "simple" name.

apollo-federation/src/sources/connect/validation/schema.rs
Hunter-apollo: We only actually care about the `$batch` namespace here so it is only necessary to compute the trie for that namespace. I will update to correct.

apollo-federation/src/sources/connect/validation/snapshots/validation_tests@batch__batch_missing_key.graphql.snap
benjamn: Since `$` is always the same as `$root` at the top level, we could replace `$root.` in those messages with `$.`. Still a little surprising if you don't know about `$`, but that's something we expect folks to learn much earlier than they learn about the `$root` shape name.

On the other hand, we definitely _could_ add `$root` as a real `KnownVariable`. That would make it possible to restart recursive execution from the root data value as many times as desired, rather than confining execution to a given subtree of the data (whatever `$` gives you), but each time you'd have to supply a new nested subpath or subselection to use `$root` in some new way, so there should be no way for execution to re-enter higher-level JSONSelection syntax. That's good news, because if execution could keep reentering the same JSONSelection syntax, that would enable unbounded conditional recursion/loops, and then the language becomes accidentally Turing-complete (unless I am happily mistaken), and many questions of static analysis (e.g. termination) become undecidable. This topic will also be relevant when we're designing reusable selection fragments, as we may want to forbid (some kinds of) fragment recursion, given the risks associated with conditional self-recursion.

apollo-federation/src/sources/connect/validation/schema.rs
```diff
@@ -327,14 +346,146 @@ fn advanced_validations(schema: &SchemaInfo, subgraph_name: &str) -> Vec<Message
...
+            trie,
+            unmapped_fields: HashSet::new(),
+        }
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+enum ShapeVisitorError<'error> {
+    #[error(
+        "The `@connect` directive on `{connector}` specifies a `$batch` entity resolver, but the field `{unset}` could not be found in the output selection"
```
lennyburdette, line 377: this is good! i'm wondering if we should be one step more explicit and say 

```
could not be found in `@connect(selection:)`
```

lennyburdette (approved):

apollo-federation/src/sources/connect/validation/schema.rs
```diff
@@ -327,14 +346,146 @@ fn advanced_validations(schema: &SchemaInfo, subgraph_name: &str) -> Vec<Message
     entity_checker.check_for_missing_entity_connectors(schema)
 }
 
+fn compute_batch_input_trie(connector: &Connector) -> SelectionTrie {
```
andrewmcgivery, line 349: Can we get a `///` comment on this method just explaining at a high level what this is computing? I wouldn't assume everyone knows what a `Trie` is and why we'd use one here 😅

apollo-federation/src/sources/connect/validation/schema.rs
```diff
@@ -327,14 +346,146 @@ fn advanced_validations(schema: &SchemaInfo, subgraph_name: &str) -> Vec<Message
...
+    },
+    #[error("Attempted to resolve key on unexpected shape `{shape_str}`")]
+    UnexpectedKeyOnShape { shape_str: String },
+    #[error(
+        "`$batch` fields must be mapped from the API response body. Variables such as `$context` and `$this` are not supported"
+    )]
+    NonRootBatch,
+}
+
+#[allow(clippy::extra_unused_lifetimes)]
```
andrewmcgivery, line 391: Flagging this as we generally don't allow for clippy overrides without a really specific reason and it should be documented in a comment.

Hunter-apollo: I dont think this is even necessary anymore as the borrow it was allowing is no longer used. Will remove.

apollo-federation/src/sources/connect/validation/schema.rs
```diff
@@ -303,6 +311,17 @@ fn advanced_validations(schema: &SchemaInfo, subgraph_name: &str) -> Vec<Message
         entity_checker.add_key(&field_set, directive);
     }
 
+    for (_, connector) in &connectors {
+        if connector.entity_resolver == Some(crate::sources::connect::EntityResolver::TypeBatch) {
```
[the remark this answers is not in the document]
Hunter-apollo: No idea why Rust rover decided that was the way. Ill fix.

apollo-federation/src/sources/connect/validation/schema.rs
Hunter-apollo: As this is a private method used only in the above validation with a well-constrained context, a doc comment here isnt really needed.

Documentation on the usage and purpose of the trie should live on the `SelectionTrie` object.

andrewmcgivery (approved):

PR #7470, opened 2025-05-12
Removes whitespace from internal url strings

naomijub:
When no `source` is found, the multiline URL composition fails with random error, this is caused by `\n` inside the string that is not a handled case for multiline URL Composition with absolute path.

Fixes **Multi-line absolute URLs fail to compose**

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [x] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-federation/src/sources/connect/validation/test_data/uri_templates/valid_connect_absolute_multiline.graphql
```diff
@@ -0,0 +1,25 @@
+extend schema
+  @link(url: "https://specs.apollo.dev/connect/v0.2", import: ["@connect"])
+
+type Query {
+  products(paramName: String, paramValue: String): [Product]
+    @connect(
+      http: { GET: """
+        https://ecommerce.demo-api.apollo.dev
+        /products?{$args.paramName}={$args.paramValue}
```
lennyburdette, line 9: while not super common, there are white-space sensitive urls for APIs like [oData](https://www.odata.org/getting-started/basic-tutorial/): `GET serviceRoot/People?$filter=FirstName eq 'Scott'` — can you add some meaningful whitespace as a regression test?

naomijub: I removed the extra `' '` but I dont know how to test this individually to see if the URL was successfully built, as there seems to be no current unit test for this section

apollo-federation/src/sources/connect/validation/connect/http.rs
```diff
@@ -291,7 +293,9 @@ impl<'schema> Transport<'schema> {
                     .collect(),
             })
             .map_err(|e| vec![e])?;
-        let url = StringTemplate::from_str(url_string.as_str())
+
+        let multiplie_url = url_string.as_str().replace(Self::SPECIAL_WHITE_SPACES, "");
+        let url = StringTemplate::from_str(&multiplie_url)
```
lennyburdette, line 298:
```suggestion
        let multiple_url = url_string.as_str().replace(Self::SPECIAL_WHITE_SPACES, "");
        let url = StringTemplate::from_str(&multiple_url)
```

apollo-federation/src/sources/connect/validation/connect/http.rs
```diff
@@ -291,7 +293,9 @@ impl<'schema> Transport<'schema> {
                     .collect(),
             })
             .map_err(|e| vec![e])?;
-        let url = StringTemplate::from_str(url_string.as_str())
+
+        let multiplie_url = url_string.as_str().replace(Self::SPECIAL_WHITE_SPACES, "");
```
lennyburdette, line 297: will this throw off locations? since we pass this to StringTemplate, it'll have `{expressions}` in it. if we're removing newlines and tabs first, that'll mean locations in expressions be off by lines or spaces, right?

lennyburdette: where does the error originate? is it in `Uri::from_str(first.value.trim())` in `validate_absolute_connect_url`? if that's the case, we could strip the illegal characters there — we don't care about locations at the point

naomijub: The reason I shifted to here is that I considered that the URL would not be constructed correctly elsewhere, specially as the validation was already successful. Would that be a case? If not, it makes sense to move it back to `validate_absolute_connect_url`

naomijub: Moved to string template builder as suggested by Dylan

apollo-federation/src/sources/connect/string_template.rs
```diff
@@ -70,7 +72,7 @@ impl FromStr for StringTemplate {
             } else {
                 let value = chars
                     .by_ref()
-                    .peeking_take_while(|c| *c != '{')
+                    .peeking_take_while(|c| *c != '{' && SPECIAL_WHITE_SPACES.contains(c))
```
dylan-apollo, line 75: We also need a way to _skip_ those characters in the loop, since the loop is only checking `peek()`. So I think adding another `if` branch which is just like:

```rust
if SPECIAL_WHITE_SPACES.contains(next) {
  chars.next();
  offset +=1;
  continue;
}
```

Or something

apollo-federation/src/sources/connect/validation/connect/http.rs
```diff
@@ -396,18 +400,30 @@ fn validate_absolute_connect_url(
...
+                    Part::Constant(Constant { 
+                        value: acc.value + &extra.value, 
+                        location: acc.location.start..extra.location.end 
+                    })
+                },
+                _ => unreachable!("All parts are constant for sure")
+            }
+        }) else {
         return Err(relative_url_error());
     };
```
dylan-apollo, line 416:
```suggestion
    let before_first_expression: String = url
        .parts
        .iter()
        .map_while(|part| {
            if let Part::Constant(constant) = part {
                Some(&constant.value)
            } else {
                None
            }
        })
        .join("");
    if before_first_expression.is_empty() {
        return Err(relative_url_error());
    }
```

I don't think we actually need to accumulate a `Part`, so we could do something like this which I think both simplifies and gets rid of the panic

dylan-apollo: (Could switch to `map_while` and keep `fold` if we do actually need the locations later)

dylan-apollo: It gets rid of the `unreachable!` which I view as a landmine. I'd always rather use the type system to represent what the data is doing so we never get an accidental panic in the future

apollo-federation/src/sources/connect/validation/connect/http.rs
```diff
@@ -396,18 +400,30 @@ fn validate_absolute_connect_url(
...
 
     let Some((_scheme, without_scheme)) = first.value.split_once("://") else {
         return Err(relative_url_error());
     };
 
-    if url.parts.len() > 1
+    if url.parts.iter().skip_while(|part| matches!(part, Part::Constant(_))).count() > 0
```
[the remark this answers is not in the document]
naomijub: still open question on `'#'`
[9922682](https://github.com/apollographql/router/pull/7470/commits/99226825c75319a47acbe26c4fa05b9a729b5987)

apollo-federation/src/sources/connect/validation/connect/http.rs
naomijub: It doesn't really:
`map_while` replaces `take_while`, that only takes when `Part::Constant(_)`, but the `unreachable` is in the fold, which is needed for the locations. I could get rig of the unreachable with the following (BUT IT TAKES A CLONE):

```rust
let first_range = url.parts.first().map(|part| part.location().start).unwrap_or_default();
    let Part::Constant(first) = Part::Constant(url
        .parts
        .iter()
        .map_while(|part| match part {
            Part::Constant(constant) => Some((constant.value.clone(), constant.location.clone())),
            _ => None
        })
        .fold((String::new(), first_range..first_range), |acc, part| {
            (acc.0 + &part.0, acc.1.start..part.1.end)
        }).into())
```

apollo-federation/src/sources/connect/validation/connect/http.rs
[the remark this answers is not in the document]
naomijub: 6994ee6ee

apollo-federation/src/sources/connect/validation/connect/http.rs
dylan-apollo: The `constant` in `Part::Constant` is a struct—so you could return that instead of the tuple? _If_ we need locations... I'm not sure we do?

apollo-federation/src/sources/connect/validation/connect/http.rs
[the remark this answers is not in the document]
naomijub: This test fails without the `'#'` check:

```graphql
extend schema
  @link(
    url: "https://specs.apollo.dev/connect/v0.2"
    import: ["@connect", "@source"]
  )

type Query {
  path: String
    @connect(
      http: { GET: "http://example.com/{$config.path}" }
      selection: "$"
    )
  inQuery: String
    @connect(
      http: { GET: "http://example.com?{$config.query}" }
      selection: "$"
    )
  inFragment: String
    @connect(
      http: { GET: "http://example.com#{$config.fragment}" }
      selection: "$"
    )
}
```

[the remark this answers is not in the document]
naomijub: Should I remove the last line?

apollo-federation/src/sources/connect/header.rs
```diff
@@ -325,21 +325,6 @@ mod test_header_value_parse {
...
     }
 
-    #[test]
-    fn invalid_header_values() {
-        assert_debug_snapshot!(
-            HeaderValue::from_str("\n"),
-            @r###"
-        Err(
-            Error {
-                message: "invalid value `\n`",
```
dylan-apollo, line 335: Is there any other invalid value we could possibly put here instead? So we still have an easy test for this check?

naomijub: I can search for one, but I suppose anything that is not ascii special white char could do it

naomijub: db42a456f

apollo-federation/src/sources/connect/validation/test_data/uri_templates/valid_connect_absolute_multiline.graphql
```diff
@@ -0,0 +1,25 @@
+extend schema
+  @link(url: "https://specs.apollo.dev/connect/v0.2", import: ["@connect"])
+
+type Query {
+  products(paramName: String, paramValue: String): [Product]
+    @connect(
+      http: { GET: """
+        https://ecommerce.demo-api.apollo.dev
+        /products?{$args.paramName}={$args.paramValue} eq 'foo'
```
lennyburdette, line 9: just confirming: whitespace is not preserved at the beginning and endings of lines, right? so if someone tried to write 
```
https://ecommerce.demo-api.apollo.dev
/products?filter=param
eq
{$args.value}
```

the result would be `https://ecommerce.demo-api.apollo.dev/products?filter=parameqvalue`?

i think that's fine — we just want to document it and the workarounds (keep significant whitespace in a line, or use a literal in an expression like `{$(' ')}`)

naomijub: Yes, we could create a function to `untrim` or prevent trimming, but I feel that is a bit overkill and might be better to just use `{$(" eq ")}`

dylan-apollo: The GraphQL string parser is going to remove whitespace around the lines I'm pretty sure without us ever seeing it. Just the line breaks themselves we have to deal with

lennyburdette: hm that doesn't seem to be true, i was able to get 
```
Caused by:
    INVALID_URL: [connectors] In `POST` in `@connect(http:)` on `User`: invalid uri character: http://localhost:4001/api   /users-batch
```

(when i made sure to "save without formatting" since i have prettier reformat things on save)

lennyburdette:
```
  @connect(
    # source: "json"
    http: {
      POST: """
      http://
      localhost:4001
      /api
      /users-batch
      ?$filter=FirstName eq 'John'
      """
      body: "ids: $batch.id"
    }
    selection: "id name username"
  )
```
results in 
```
Caused by:
    INVALID_URL: [connectors] In `POST` in `@connect(http:)` on `User`: invalid uri character: http://localhost:4001/api/users-batch?$filter=FirstName eq 'John'
```

lennyburdette: (this is from building a supergraph binary based on this branch)

dylan-apollo: Hmm okay, seems like we need some line trimming too, then 😢. That's gonna be a little trickier

apollo-federation/src/sources/connect/validation/connect/http.rs
```diff
@@ -396,34 +396,56 @@ fn validate_absolute_connect_url(
...
+                )),
+                _ => None,
+            })
+            .fold((String::new(), first_range..first_range), |acc, part| {
+                (acc.0 + part.0, acc.1.start..part.1.end)
+            })
+            .into(),
+    ) else {
         return Err(relative_url_error());
     };
```
dylan-apollo, line 420:
```suggestion
    let first = url
        .parts
        .iter()
        .map_while(|part| match part {
            Part::Constant(constant) => Some(constant),
            _ => None,
        })
        .fold(Constant::default(), |mut acc, part| {
            acc.value += &part.value;
            acc.location.end = part.location.end;
            acc
        });
    if first.value.is_empty() {
        return Err(relative_url_error());
    };
```

This saves us some code (here and we can remove the `impl From<(String, Range<usize>)>`

apollo-federation/src/sources/connect/string_template.rs
```diff
@@ -34,7 +36,11 @@ impl FromStr for StringTemplate {
         let mut chars = input.chars().peekable();
         let mut parts = Vec::new();
         while let Some(next) = chars.peek() {
-            if *next == '{' {
+            if SPECIAL_WHITE_SPACES.contains(next) {
+                chars.next();
+                offset += 1;
+                continue;
+            } else if *next == '{' {
```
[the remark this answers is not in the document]
naomijub: I don't have strong feelings

apollo-federation/src/sources/connect/validation/connect/http.rs
naomijub: I really liked the pattern matched structure, but you are right, it can reduce a lot of code this way

dylan-apollo (approved):

PR #7474, opened 2025-05-13
Add OCI schema store

jsegaran:
Support fetching the supergraph schema from Apollo OCI Registry, instead of uplink. We also aren't supporting hot-reloading of supergraph schemas as part of this workflow. 

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [x] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-1338]: https://apollographql.atlassian.net/browse/ROUTER-1338?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apollo-router/Cargo.toml
+++ b/apollo-router/Cargo.toml
@@ -83,6 +83,7 @@ cookie = { version = "0.18.0", default-features = false }
 crossbeam-channel = "0.5.15"
 ci_info = { version = "0.14.14", features = ["serde-1"] }
 dashmap = { version = "6.0.0", features = ["serde"] }
+docker_credential = "1.3.1"
 derivative = "2.2.0"
 derive_more = { version = "2.0.0", default-features = false, features = [
     "from",
@@ -127,6 +128,7 @@ notify = { version = "6.1.1", default-features = false, features = [
 ] }
 nu-ansi-term = "0.50"
 num-traits = "0.2.19"
+oci-client = { version = "0.14.0", default-features = false, features = ["rustls-tls", "rustls-tls-native-roots", "trust-dns", "test-registry"] }
 once_cell = "1.19.0"
 
 # Any package that starts with `opentelemetry` needs to be updated with care
--- a/apollo-router/Cargo.toml
+++ b/apollo-router/Cargo.toml
@@ -128,7 +128,7 @@ notify = { version = "6.1.1", default-features = false, features = [
 ] }
 nu-ansi-term = "0.50"
 num-traits = "0.2.19"
-oci-client = { version = "0.14.0", default-features = false, features = ["rustls-tls", "rustls-tls-native-roots", "trust-dns", "test-registry"] }
+oci-client = { version = "0.14.0", default-features = false, features = ["rustls-tls", "rustls-tls-native-roots", "trust-dns"] }
 once_cell = "1.19.0"
 
 # Any package that starts with `opentelemetry` needs to be updated with care
--- a/apollo-router/Cargo.toml
+++ b/apollo-router/Cargo.toml
@@ -128,7 +128,7 @@ notify = { version = "6.1.1", default-features = false, features = [
 ] }
 nu-ansi-term = "0.50"
 num-traits = "0.2.19"
-oci-client = { version = "0.14.0", default-features = false, features = ["rustls-tls", "rustls-tls-native-roots", "trust-dns"] }
+oci-client = { version = "0.15.0", default-features = false, features = ["rustls-tls", "rustls-tls-native-roots", "trust-dns"] }
 once_cell = "1.19.0"
 
 # Any package that starts with `opentelemetry` needs to be updated with care
--- a/apollo-router/src/executable.rs
+++ b/apollo-router/src/executable.rs
@@ -228,7 +228,7 @@ pub struct Opt {
     /// An OCI reference to an image that contains the supergraph schema for the router.
     #[clap(long, env, action = ArgAction::Append)]
     // TODO: Update name to be final public name
-     graph_artifact_reference: Option<String>,
+    graph_artifact_reference: Option<String>,
 
     /// Disable sending anonymous usage information to Apollo.
     #[clap(long, env = "APOLLO_TELEMETRY_DISABLED", value_parser = FalseyValueParser::new())]
@@ -297,7 +297,7 @@ impl Opt {
                 .clone()
                 .ok_or(Self::err_require_opt("APOLLO_KEY"))?,
             reference: self
-                . graph_artifact_reference
+                .graph_artifact_reference
                 .clone()
                 .ok_or(Self::err_require_opt(" GRAPH_ARTIFACT_REFERENCE"))?,
         })
@@ -639,7 +639,7 @@ impl Executable {
                             return Err(anyhow!("Failed to read Apollo key file: {}", err));
                         }
                     };
-                    match opt. graph_artifact_reference {
+                    match opt.graph_artifact_reference {
                         None => SchemaSource::Registry(opt.uplink_config()?),
                         Some(_) => SchemaSource::OCI(opt.oci_config()?),
                     }
@@ -648,7 +648,7 @@ impl Executable {
             (_, None, None, Some(_apollo_key), None) => {
                 tracing::info!("{apollo_router_msg}");
                 tracing::info!("{apollo_telemetry_msg}");
-                match opt. graph_artifact_reference {
+                match opt.graph_artifact_reference {
                     None => SchemaSource::Registry(opt.uplink_config()?),
                     Some(_) => SchemaSource::OCI(opt.oci_config()?),
                 }
--- a/apollo-router/src/executable.rs
+++ b/apollo-router/src/executable.rs
@@ -542,7 +542,7 @@ impl Executable {
         // 2. Env APOLLO_ROUTER_SUPERGRAPH_PATH
         // 3. Env APOLLO_ROUTER_SUPERGRAPH_URLS
         // 4. Env APOLLO_KEY and APOLLO_OCI_REGISTRY_ENDPOINT
-        // 4. Env APOLLO_KEY and APOLLO_GRAPH_REF
+        // 5. Env APOLLO_KEY and APOLLO_GRAPH_REF
         #[cfg(unix)]
         let akp = &opt.apollo_key_path;
         #[cfg(not(unix))]
--- a/apollo-router/src/executable.rs
+++ b/apollo-router/src/executable.rs
@@ -541,6 +541,7 @@ impl Executable {
         // 1. Cli --supergraph
         // 2. Env APOLLO_ROUTER_SUPERGRAPH_PATH
         // 3. Env APOLLO_ROUTER_SUPERGRAPH_URLS
+        // 4. Env APOLLO_KEY and APOLLO_OCI_REGISTRY_ENDPOINT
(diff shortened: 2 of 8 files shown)


apollo-router/Cargo.toml
```diff
@@ -126,6 +127,7 @@ notify = { version = "6.1.1", default-features = false, features = [
 ] }
 nu-ansi-term = "0.50"
 num-traits = "0.2.19"
+oci-client = { version = "0.14.0", default-features = false, features = ["rustls-tls", "rustls-tls-native-roots", "trust-dns", "test-registry"] }
```
garypen, line 130: We don't need `test-registry`, I'll remove that from `towering` as well.

```suggestion
oci-client = { version = "0.14.0", default-features = false, features = ["rustls-tls", "rustls-tls-native-roots", "trust-dns"] }
```

apollo-router/src/executable.rs
garypen, line 1: Somewhere in the `inner_start()` function, we'll need code to raise an error if both uplink and oci registry are configured (since a router can't be getting schema from both). We'll need tests for that as well.

I suspect this is going to be "tricky" because there's a bunch of environment variables (APOLLO_GRAPH_REF, etc..) that combine with configuration to create the final configuration and all possible combinations will have to be examined so that they make sense.

You'll also need to:
 - update the startup message to document the new oci option
 - update various comments, e.g.: lines 540-545 to bring them up to date.

apollo-router/src/registry/mod.rs
```diff
@@ -0,0 +1,175 @@
...
+use thiserror::Error;
+
+/// Configuration for fetching an OCI Bundle
+/// This struct does not change on router reloads - they are all sourced from CLI options.
+#[derive(Debug, Clone)]
+pub struct OCIConfig {
+    /// The Apollo key: `<YOUR_GRAPH_API_KEY>`
+    pub apollo_key: String,
+
+    // Should this be a reference instead?
```
garypen, line 18: Generally, it's best to have a struct own its data. Use a reference when these factors apply:
 - performance/resources are a concern
 - a different struct (with a predictable lifetime) clearly owns the data
 - consistency across entities is an important aspect of the problem

apollo-router/src/registry/mod.rs
```diff
@@ -0,0 +1,175 @@
...
+    pub url: String,
+}
+
+pub(crate) struct OCIResult {
+    pub schema: String,
+}
+
+#[derive(Debug, Error)]
+pub(crate) enum Error {
+    #[error("OCI layer has not title")]
```
garypen, line 28:
```suggestion
    #[error("OCI layer does not have a title")]
```

apollo-router/src/registry/mod.rs
```diff
@@ -0,0 +1,175 @@
...
+    }
+}
+
+impl From<FromUtf8Error> for Error {
+    fn from(value: FromUtf8Error) -> Self {
+        Error::OCILayerParseError(value)
+    }
+}
+
+fn build_auth(reference: &Reference, apollo_key: &String) -> RegistryAuth {
```
garypen, line 56: Never use `&String`. If it is a ref, use `&str` For more details on why: https://users.rust-lang.org/t/string-str-string-str/100611/3

apollo-router/src/registry/mod.rs
```diff
@@ -0,0 +1,175 @@
...
+    let server = reference
+        .resolve_registry()
+        .strip_suffix('/')
+        .unwrap_or_else(|| reference.resolve_registry());
+
+    // Check if the server registry ends with apollographql.com
+    if server.ends_with("apollographql.com") {
```
garypen, line 63: Better style to make important strings be const. e.g.:
```
const APOLLO_REGISTRY: &str = "apollographql.com";
```
which are declared near the top of the file. If you look in other source files, you'll see lots of examples of these.

Do the same for the other important strings in this file.

apollo-router/src/registry/mod.rs
```diff
@@ -0,0 +1,175 @@
...
+        }
+    }
+}
+
+async fn pull_oci(
+    client: &mut ociClient,
+    auth: &RegistryAuth,
+    reference: &Reference,
+) -> Result<OCIResult, Error> {
+    tracing::info!(?reference, "pulling oci bundle");
```
garypen, line 93:
```suggestion
    tracing::debug!(?reference, "pulling oci bundle");
```

apollo-router/src/registry/mod.rs
```diff
@@ -0,0 +1,175 @@
...
+    let supported_types = vec![
+        "application/json",
+        "text/plain",
+        "application/vnd.oci.empty.v1+json",
+    ];
+    let image = client
+        .pull(reference, auth, supported_types.clone())
```
garypen, line 100: You don't need the clone because you aren't writing layers.

```suggestion
        .pull(reference, auth, supported_types)
```

apollo-router/src/registry/mod.rs
```diff
@@ -0,0 +1,175 @@
...
+    let client_config = oci_client::client::ClientConfig {
+        protocol: oci_client::client::ClientProtocol::Https,
+        ..Default::default()
+    };
+
+    let mut client = ociClient::new(client_config);
+
+    let auth = build_auth(&reference, &oci_config.apollo_key);
+
+    // let output = cli.output.unwrap_or("/tmp".to_string());
```
garypen, line 127: Delete dead code.
```suggestion
```

apollo-router/src/router/event/schema.rs
```diff
@@ -154,6 +159,25 @@ impl SchemaSource {
                 .filter_map(|s| async move { s.map(Event::UpdateSchema) })
                 .boxed()
             }
+            SchemaSource::OCI(oci_config) => {
+                futures::stream::once(async move {
+                    // TODO: Fix error handling
```
garypen, line 164: Interestingly, the other mechanisms seem to always return `None` when a schema can't be retrieved. Without digging into that too much, I'm assuming that is handled elsewhere, so your error handling might already be "correct".

We can dig into that when you get a bit further along.

garypen (commented):
Looks great so far. I've put a few "style" comments in which are hopefully helpful. I think the major complication is going to be ensuring that we can't have a router which tries to get schema from both uplink and OCI registry.

apollo-router/src/executable.rs
jsegaran: I think we want to support customers fetching from uplink for entitlements and persisted queries.
I was thinking of use the presence of an OCI endpoint as deciding if we are using uplnk or OCI
```
 match opt.apollo_oci_registry_endpoint {
                        None => SchemaSource::Registry(opt.uplink_config()?),
                        Some(_) => SchemaSource::OCI(opt.oci_config()?),
                    }
```

Looking at the code itself, it seems we return a single schema source, so I don't think it is possible to configure both uplink and OCI for schema source fetching?

garypen: I've spent some time thinking about this and I think you are right about the behaviour of the code.

I still think that various comments, logs and documentation may/will need to be updated to explain clearly what is happening with various combinations of options.

apollo-router/src/registry/mod.rs
```diff
@@ -0,0 +1,322 @@
...
+pub(crate) enum Error {
+    #[error("OCI layer does not have a title")]
+    OCILayerMissingTitle,
+    #[error("Oci Distribution error: {0}")]
+    OCIDistributionError(OciDistributionError),
+    #[error("Oci Parsing error: {0}")]
+    OCIParseError(oci_client::ParseError),
+    #[error("Unable to parse layer: {0}")]
+    OCILayerParseError(FromUtf8Error),
+}
```
Velfi, line 38: Consider this:

```suggestion
pub(crate) enum OciError {
    #[error("OCI layer does not have a title")]
    LayerMissingTitle,
    #[error("Oci Distribution error: {0}")]
    Distribution(OciDistributionError),
    #[error("Oci Parsing error: {0}")]
    Parse(oci_client::ParseError),
    #[error("Unable to parse layer: {0}")]
    LayerParse(FromUtf8Error),
}
```

Velfi (commented):
I have some style comments

apollo-router/src/registry/mod.rs
```diff
@@ -0,0 +1,322 @@
...
+        .expect("url must be valid");
+        let oci_manifest = OciManifest::Image(OciImageManifest {
+            schema_version: 2,
+            media_type: Some(IMAGE_MANIFEST_MEDIA_TYPE.to_string()),
+            config: Default::default(),
+            layers: layer_descriptors,
+            subject: None,
+            artifact_type: None,
+            annotations: None,
+        });
```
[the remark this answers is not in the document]
jsegaran: Given this is only used once in this test file, I'm gonna leave as is?

Side question, is it considered okay practice in rust to implement the default trait (or any other trait) for things that are defined in 3rd party libraries?

apollo-router/src/executable.rs
```diff
@@ -619,13 +639,19 @@ impl Executable {
...
-                    SchemaSource::Registry(opt.uplink_config()?)
+                    match opt.apollo_oci_registry_endpoint {
+                        None => SchemaSource::Registry(opt.uplink_config()?),
+                        Some(_) => SchemaSource::OCI(opt.oci_config()?),
+                    }
                 }
             }
             (_, None, None, Some(_apollo_key), None) => {
                 tracing::info!("{apollo_router_msg}");
                 tracing::info!("{apollo_telemetry_msg}");
```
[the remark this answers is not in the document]
jsegaran: I don't think we would leak things? The OCI url is a specific container ref style format, and will be included as an env var/cli by default so not secret. 

The url needs to follow the reference REGEX which doesn't support authentication in the url ([link](https://github.com/youki-dev/oci-spec-rs/blob/f1161092df9ea14229984cfdb54932e3692a1b01/src/distribution/reference.rs#L16-L18)).

apollo-router/src/executable.rs
```diff
@@ -224,6 +225,11 @@ pub struct Opt {
     // Should be a Vec<Url> when https://github.com/clap-rs/clap/discussions/3796 is solved
     apollo_uplink_endpoints: Option<String>,
 
+    /// The endpoint used to fetch the supergraph schema from the OCI registry.
+    #[clap(long, env, action = ArgAction::Append)]
+    // TODO: Update name to be final public name
+    apollo_oci_registry_endpoint: Option<String>,
```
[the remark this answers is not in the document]
jsegaran: We are still figuring it out in the registry team. I'm leaning to `${noun}-artifact-reference`, and we are still figuring out the noun but probably build or graph. 
Reference is a well defined concept in OCI so I think that makes the most sense here.
Link: https://oras.land/docs/concepts/reference/

Meschreiber (approved):

apollo-router/src/registry/mod.rs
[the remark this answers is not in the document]
BrynCooke: You can't implement an external trait for a third party struct, the compiler will prevent you as the library owner may implement that trait and cause breakage. You can only implement traits that you own for third party structs.

BrynCooke (approved):

PR #7475, opened 2025-05-13
PoC: Require reporting configuration based on entitlements

timbotnik:
- Adds support for a theoretical `restricted` claim, which when set on the license, requires that there are NO license violations, otherwise prevents the Router from starting (effectively allows a license to conditionally behave like the unlicensed path).

- Adds support for a theoretical `usage_reporting` claim, which when set on the license, requires that you don’t change the usage reporting URL.

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

Manual test with a free account using the env var to override:
image

Manual test using additional restricted features:
image

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/state_machine.rs
+++ b/apollo-router/src/state_machine.rs
@@ -327,8 +327,8 @@ impl<FA: RouterSuperServiceFactory> State<FA> {
         match license {
             LicenseState::Licensed { limits } => {
                 let license_is_restricted = limits
-                    .map(|limits| limits.restricted.unwrap_or(false))
-                    .unwrap_or(false);
+                    .map(|limits| limits.restricted.unwrap_or(true))
+                    .unwrap_or(true); // TODO: Revert me.  Unless we were to update all the licenses first, this should default to false.  Useful for testing though.
                 if report.uses_restricted_features() && license_is_restricted {
                     tracing::error!(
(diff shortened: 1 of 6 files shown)


apollo-router/src/state_machine.rs
```diff
@@ -320,27 +319,37 @@ impl<FA: RouterSuperServiceFactory> State<FA> {
...
-        let license_limits = match license {
+        match license {
             LicenseState::Licensed { limits } => {
-                tracing::debug!("A valid Apollo license has been detected.");
-                limits
+                let license_is_restricted = limits
+                    .map(|limits| limits.restricted.unwrap_or(false))
+                    .unwrap_or(false);
+                if report.uses_restricted_features() && license_is_restricted {
+                    tracing::error!(
```
BrynCooke, line 333: I'd like to see messaging delivered as part of the JWT (with defaults in the router). That way we can change what we display from the server side without having to update the router code.

BrynCooke:
I'd also like us to deliver the features that should be checked via JWT. That way when pricing plans change they are reflected across all routers. For instance we may decide to eventually make a paid feature free.

aaronArinder:
stale; reopen if needed

PR #7482, opened 2025-05-13
Properly categorize and surface FieldSet validation messages during directive validation

tninesling:
# Overview

For FieldSet-based directives (key, provides, etc), we surface validation errors from the parser if we do not encounter federation-specific errors but the FieldSet is still invalid. 

## Error rewriting

When parsing FieldSets, apollo-compiler will return a generic error message without information about the source directive. Inside the validation methods for each of the relevant directives, these errors are caught and categorized to the appropriate variant so they are surfaced with the right error code (i.e. `WithErrors -> KeyInvalidFields` instead of `WithErrors -> InvalidGraphQL` as we do in other cases).

We try to adjust the errors themselves to be closer to the original JS messages with `unstable_compat_message`, defaulting to the normal apollo-compiler error format.

### Error captures

I made a small update to `collect_target_fields_from_field_set` to continue collecting leaf fields from the partially-parsed selection when parts of the selection are invalid. This allows us to avoid extraneous `EXTERNAL_UNUSED` errors when we have a valid, used external field next to some other field that fails validation for some reason. This behavior only occurs when invoked with `validate: false`.

---

**Checklist**

- Tests added and passing[^3]
    - [X] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-federation/src/schema/validators/mod.rs
```diff
@@ -351,3 +353,11 @@ pub(crate) fn deny_unsupported_directive_on_interface_field<D: AppliesOnField>(
         );
     }
 }
+
+pub(crate) fn normalize_diagnostic_message(diagnostic: Diagnostic<'_, DiagnosticData>) -> String {
+    diagnostic
+        .error
+        .unstable_compat_message() // Attempt to convert to something closer to the original JS error messages
```
duckki, line 360: 🤯  I didn't know this method existed.

apollo-federation/tests/subgraph/subgraph_validation_tests.rs
```diff
@@ -418,13 +417,19 @@ mod fieldset_based_directives {
             err,
             [(
                 "KEY_INVALID_FIELDS",
-                r#"[S] On type "T", for @key(fields: ":f"): Syntax Error: Expected Name, found ":"."#,
+                r#"[S] On type "T", for @key(fields: ":f"): Syntax error: expected at least one Selection in Selection Set
+   ╭─[ field_set.graphql:1:1 ]
```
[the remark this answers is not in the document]
duckki: I think we need to add a new feature in apollo-compiler to set an arbitrary initial location for the field set parser. Alternatively, we may drop the code quotation part from the error, which matches JS output. But, I wish to keep it.

apollo-federation/tests/subgraph/subgraph_validation_tests.rs
```diff
@@ -471,7 +491,7 @@ mod fieldset_based_directives {
             err,
             [(
                 "REQUIRES_INVALID_FIELDS",
-                r#"[S] On field "T.g", for @requires(fields: "f b"): Cannot query field "b" on type "T" (if the field is defined in another subgraph, you need to add it to this subgraph with @external)."#,
+                r#"[S] On field "T.g", for @requires(fields: "f b"): Cannot query field "b" on type "T"."#,
```
duckki, line 494: I think we should annotate this as a failure (`should_panic`), until we add the `(if the field is defined in another subgraph, you need to add it to this subgraph with @external)` part.

tninesling: I get what you mean. This is what comes out of apollo-rs with that `unstable_compat_message` translation, and I was thinking we should just use what it gives us. I'm hesitant to run through that compatibility translation and then also apply some regex search to tack on this suffix.

Maybe we just use the original apollo-compiler message? That looks like
```
[S] On field "T.g", for @requires(fields: "f b"): Error: type `T` does not have a field `b`
   ╭─[ field_set.graphql:1:3 ]
   │
 1 │ f b
   │   ┬
   │   ╰── field `b` selected here
   │
   │ Note: path to the field: `query → b`
───╯
   ```

**Comment by duckki** [comment]
As to the synthetic file name in`╭─[ field_set.graphql:1:1 ]`, maybe we can just give a more abstract file name like "original/subgraph-path.graphql:<line>:<column>;offset:", which would look like:

```
[S] On field "T.g", for @requires(fields: "f b"): Error: type `T` does not have a field `b`
   ╭─[ original/subgraph-path.graphql:31:12;offset:1:3 ]
   │
 1 │ f b
   │   ┬
   │   ╰── field `b` selected here
```

apollo-federation/tests/subgraph/subgraph_validation_tests.rs
tninesling: I've added this back in with a slight punctuation change to make the formatting a bit simpler to write.

apollo-federation/tests/subgraph/subgraph_validation_tests.rs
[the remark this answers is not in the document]
tninesling: I think we should strip this for now, since it's potentially misleading and we already have the coordinates available in the message. I've updated these to just include the first line, which is closer to the original text. Long-term, I would love to replace the inconsistent coordinates with apollo-compiler-style formatting for all the errors though.

duckki (approved):
Looks good!

PR #7486, opened 2025-05-14
chore: simplify yaml config in traffic shaping tests

goto-bus-stop:
I was reading this for unrelated reasons and was a bit horrified that we're using string concatenation to merge config in a language with significant whitespace. I think there's not much benefit to factoring out this config, and it makes the tests harder to read, so I've just copy-pasted it into each usage site.

### DIFF
--- a/apollo-router/tests/integration/traffic_shaping.rs
+++ b/apollo-router/tests/integration/traffic_shaping.rs
@@ -11,7 +11,11 @@ use crate::integration::common::Query;
 use crate::integration::common::Telemetry;
 use crate::integration::common::graph_os_enabled;
 
-const PROMETHEUS_CONFIG: &str = r#"
+#[tokio::test(flavor = "multi_thread")]
+async fn test_router_timeout() -> Result<(), BoxError> {
+    let mut router = IntegrationTest::builder()
+        .config(format!(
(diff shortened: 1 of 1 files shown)


apollo-router/tests/integration/traffic_shaping.rs
```diff
@@ -180,7 +188,13 @@ async fn test_router_custom_metric() -> Result<(), BoxError> {
...
             r#"
-            {PROMETHEUS_CONFIG}
+            telemetry:
+                exporters:
+                    metrics:
+                        prometheus:
+                            listen: 127.0.0.1:4000
+                            enabled: true
+                            path: /metrics
                 instrumentation:
```
goto-bus-stop, line 198: This one was especially horrifying:
```yaml
{PROMETHEUS_CONFIG}
    instrumentation:
```
it relies on the indentation inside the `PROMETHEUS_CONFIG` const to be aligned with this string, and you can't know what `instrumentation:` refers to without reading the `PROMETHEUS_CONFIG` value.

carodewig (approved):
I think it would be nice to figure out how to template this more generally going forward! But for now, I agree that it would be better to not rely on the indentation to stay consistent.

PR #7496, opened 2025-05-14
Add new guide for EKS

smyrick:
Branched from the big docs rework, add more pages for AWS including a basic page to start for EKS


smyrick:
This might make more sense as a separate doc. I wrote this intially with details we have internally but how to setup eksctl is really not important for our docs

PR #7498, opened 2025-05-14
Send Enabled Features to Apollo Protobuf

rregitsky:
Passes through the enabled features from #7429 to the apollo protobuf.

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [x] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/plugins/telemetry/mod.rs
```diff
@@ -1722,7 +1750,17 @@ impl Telemetry {
...
+            // The distributed APQ cache is only considered enabled if the redis config is also set.
+            distributed_apq_cache: {
+                let enabled = full_config["apq"]["enabled"].as_bool().unwrap_or(true);
+                let redis_cache_config_set =
+                    full_config["apq"]["router"]["cache"]["redis"].is_object();
+                enabled && redis_cache_config_set
+            },
+            // Entity cache's top-level enabled flag defaults to false. If the top-level flag is
+            // enabled, the feature is considered enabled regardless of the subgraph-level enabled
+            // settings.
```
bonnici, line 1763: Is it possible to have the top-level flag off but a subgraph-level flag on, and should that be considered "enabled"?

bonnici (approved):
Looks good to me pending router-core's approval. Just one question.

apollo-router/src/plugins/telemetry/metrics/apollo/snapshots/apollo_router__plugins__telemetry__metrics__apollo__studio__test__aggregation.snap
```diff
@@ -1465,5 +1466,9 @@ expression: aggregated_metrics
       "subtype": null,
       "licensed_operation_count": 2
     }
-  }
+  },
+  "router_features_enabled": [
+    "apq",
```
timbotnik, line 1471: Shouldn't this be `distributed_apq_cache`?

apollo-router/src/plugins/telemetry/metrics/apollo/studio.rs
```diff
@@ -666,6 +667,7 @@ mod test {
                     query_metadata: None,
                 },
             )]),
+            router_features_enabled: vec!["apq".to_string(), "entity_cache".to_string()],
```
timbotnik, line 670: Probably related to previous comment.

apollo-router/src/plugins/telemetry/mod.rs
timbotnik: Agree that this is likely possible and should represent an "enabled" state.

timbotnik (dismissed):
I'm pretty sure we need to handle the "partially enabled" state for individual subgraph configs overriding the overall config, otherwise looking pretty good!

apollo-router/src/plugins/telemetry/mod.rs
BrynCooke: It's not possible. The global enabled flag is required to be enabled for entity caching to be enabled at all.

apollo-router/src/plugins/telemetry/metrics/apollo/snapshots/apollo_router__plugins__telemetry__metrics__apollo__studio__test__aggregation.snap
BrynCooke: Yes, and the path into the config is:

```
apq: 
  router: 
    cache: 
      redis: 
```

apollo-router/src/plugins/telemetry/mod.rs
rregitsky: Yeah, we [ensure the enabled flag is on](https://github.com/apollographql/router/blob/bd4c551ce3940d69d5c256f9a7a9219f538dd886/apollo-router/src/plugins/cache/entity.rs#L524) before checking if a subgraph is enabled.

PR #7499, opened 2025-05-14
Fix bug where headers were sometimes being sent twice during header propagation

andrewmcgivery:
Fix bug where headers were sometimes being sent twice during header propagation

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [x] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/plugins/connectors/tests/mod.rs
```diff
@@ -731,23 +731,159 @@ async fn test_override_headers_with_config() {
...
+                    }
+                }
+            },
+            "headers": {
+              "connector": {
+                "all": {
+                  "request": [
+                  {
+                    "propagate": {
+                      "named": "x-forward",
```
lennyburdette, line 780: should we also have a insert rule test, and a propagate matching test? mostly as a regression check

lennyburdette: what happens if there's `headers: [{ name: "x-foo", value: "bar" }]` in SDL, and `remove: { named: "x-foo" }` in yaml?

apollo-router/src/plugins/headers/mod.rs
```diff
@@ -539,46 +539,29 @@ impl Propagate {
...
-                let mut previous_name = None;
-
                 supergraph_headers
                     .iter()
                     .filter(|(name, _)| {
                         !RESERVED_HEADERS.contains(*name) && matching.is_match(name.as_str())
                     })
                     .for_each(|(name, value)| {
                         if !already_propagated.contains(name.as_str()) {
-                            headers_mut.append(name, value.clone());
```
lennyburdette, line 563: i *think* what this is trying to do:

* `supergraph_headers` could include the same header multiple times.
* there's an assumption that headers of the same name will appear together (`x-foo, x-foo, x-bar`, not `x-foo, x-bar, x-foo`)
* the loop handles one header at a time. when we see the first x-foo, we track its name. if the next header is also x-foo, append it. if the next header is x-bar, we must be done with x-foo and so we mark it as propagated.

this doesn't make much sense to me. it seems like we should group_by name first, then handle a group all at once, and mark that group's header name as already_propagated. that would have a subtle change in supporting `x-foo, x-bar, x-foo`, but is otherwise the same.

but if you grouped first, you could implement something that acted like `insert_multi`. this would preserve the ability to propagate the same header multiple times, while ensuring that this rule overwrites any existing headers with that name. something like

```
if let Some(first) = group.next() {
  headers_mut.insert(name, first.clone());
}
for rest in &group {
  headers_mut.append(name, rest.clone());
}
```

that code might be nonsense, but i think the idea is sound. we'll definitely want careful review on this though.

lennyburdette (changes requested):

apollo-router/src/plugins/headers/mod.rs
```diff
@@ -551,34 +574,39 @@ impl Propagate {
...
-                                None => previous_name = Some(name),
-                                Some(previous) => {
-                                    if previous != name {
-                                        already_propagated.insert(previous.to_string());
-                                        previous_name = Some(name);
-                                    }
+                            // If the header was already added previously by some other method (E.g Connectors), remove it since we're going to override it
+                            if existing_headers.contains_key(name) {
+                                // Use `remove_entry_mult` since `remove` only removes the first instance of a header
+                                if let Entry::Occupied(entry) = headers_mut.entry(name) {
```
lennyburdette, line 589: these are redundant right?

apollo-router/src/plugins/headers/mod.rs
```diff
@@ -551,34 +574,39 @@ impl Propagate {
...
+                                    entry.remove_entry_mult();
                                 }
                             }
+                            headers.for_each(|(_, value)| {
+                                headers_mut.append(name, value.clone());
+                                already_propagated.insert(name.to_string());
```
lennyburdette, line 595: you can do this outside the loop

apollo-router/src/plugins/headers/mod.rs
lennyburdette: you're totally right, i misread them as the same map.

lennyburdette:
looks good to me but we should run this by the router team before merging!

BrynCooke (changes requested):
Please can you add some new tests directly to the headers module? There is a material change to the logic in this module, but no test that shows this change.

In addition, are there any docs that need clarification?

.changesets/fix_am_bug_duplicatedropogateheader.md
```diff
@@ -0,0 +1,5 @@
+### Fix bug where headers were sometimes being sent twice during header propagation ([PR #7499](https://github.com/apollographql/router/pull/7499))
+
+Fix bug where headers were sometimes being sent twice during header propagation when propagation was configured in both a connector and yaml config.
```
lennyburdette, line 3: I guess this needs additional clarification about removing all headers that match a rule, and not just the first. That's the only behavior change now, right? (and something that needs a test)

dylan-apollo: Should probably adjust the top line to specify Connectors?

apollo-router/src/plugins/headers/mod.rs
```diff
@@ -551,34 +574,28 @@ impl Propagate {
...
-                            // multiple times
-                            match previous_name {
-                                None => previous_name = Some(name),
-                                Some(previous) => {
-                                    if previous != name {
-                                        already_propagated.insert(previous.to_string());
-                                        previous_name = Some(name);
-                                    }
+                            // If the header was already added previously by some other method (E.g Connectors), remove it since we're going to override it
+                            if existing_headers.contains_key(name) {
```
dylan-apollo, line 587: I'm very confused both by the code and by the intended behavior. Why doesn't `already_propagated` track this? And... what is the user supposed to expect here? If you set headers to get added/forwarded in schema, then configure something about that same header in router config... router config overrides schema?

andrewmcgivery: `already_propagated` tracks what heads have been propagated _by this plugin_. Schema headers are set before this plugin so effectively what this code does is if it sees an "existing header" (one that was set before the plugin ran), it will remove it before setting it's own.

lennyburdette: i expanded the comment, hopefully it helps?

.changesets/fix_am_bug_duplicatedropogateheader.md
lennyburdette: my suggestion is not relevant, the remove behavior was always the same. and i updated the changelog to talk about connectors

lennyburdette:
@BrynCooke turns out we didn't need some of the changes (the behavior of `remove` with multiple headers didn't actually change, and i added a test that confirms this), so i believe the new tests are sufficient. this fix now aligns with our docs: https://www.apollographql.com/docs/graphos/connectors/deployment/overriding-headers

lennyburdette (approved):
approving to remove my "request changes" but still waiting for @BrynCooke 's approval

BrynCooke (approved):

PR #7503, opened 2025-05-15
fix: don't carryover unused input definitions

lennyburdette:
it's possible that carrying over unused definitions will result in disagreements in associations between types and and subgraphs. a truly unused type will never be involved in query planning, so this change drops those definitions from the expanded supergraph entirely.

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-federation/src/schema/referencer.rs
```diff
@@ -127,6 +139,17 @@ pub(crate) struct EnumTypeReferencers {
     pub(crate) directive_arguments: IndexSet<DirectiveArgumentDefinitionPosition>,
 }
 
+impl EnumTypeReferencers {
+    pub(crate) fn len(&self) -> usize {
```
Hunter-apollo, line 143: Should this be expressed as a common Trait?

lennyburdette: that's a good question for the fed-core team, i'll ask!

lennyburdette: sounds like a YAGNI situation

naomijub (approved):

apollo-federation/src/sources/connect/expand/carryover/inputs.rs
```diff
@@ -54,6 +54,16 @@ pub(super) fn copy_input_types(
         }
         match ty {
             ExtendedType::Scalar(node) => {
+                let references = from.referencers().scalar_types.get(name).ok_or_else(|| {
+                    FederationError::internal(format!(
+                        "Cannot find enum type referencers for {name}"
```
dylan-apollo, line 59: Might just exit early here too? Maybe `is_none_or` with that length check?

dylan-apollo (approved):

apollo-federation/src/sources/connect/expand/carryover/inputs.rs
lennyburdette: i like it!

PR #7504, opened 2025-05-15
Port specs for inaccessible and tag

tninesling:
* Ports directive specifications for `@inaccessible` and `@tag`.
* Ensures that injected `@link` to the link spec is always first
* Fixes missing composes/supergraph spec config for `@context`

Had some issues with insta snapshot file names being too long, so I updated a few tests comparing long strings to output a text diff via `similar` on failure.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [X] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-federation/tests/subgraph/subgraph_validation_tests.rs
+++ b/apollo-federation/tests/subgraph/subgraph_validation_tests.rs
@@ -936,73 +936,67 @@ Did you mean "@shareable"?{}"#,
 // PORT_NOTE: Corresponds to '@core/@link handling' tests in JS
 #[cfg(test)]
 mod link_handling_tests {
+    use similar::TextDiff;
+
     use super::*;
 
-    // TODO(FED-543): Remaining directive definitions should be added to the schema
     #[allow(dead_code)]
-    const EXPECTED_FULL_SCHEMA: &str = r#"
-    schema
-      @link(url: "https://specs.apollo.dev/link/v1.0")
-      @link(url: "https://specs.apollo.dev/federation/v2.0", import: ["@key"])
-    {
-      query: Query
-    }
+    const EXPECTED_FULL_SCHEMA: &str = r#"schema {
+  query: Query
+}
 
-    directive @link(url: String, as: String, for: link__Purpose, import: [link__Import]) repeatable on SCHEMA
+extend schema @link(url: "https://specs.apollo.dev/link/v1.0") @link(url: "https://specs.apollo.dev/federation/v2.0", import: ["@key"])
 
-    directive @key(fields: federation__FieldSet!, resolvable: Boolean = true) repeatable on OBJECT | INTERFACE
+directive @link(url: String, as: String, for: link__Purpose, import: [link__Import]) repeatable on SCHEMA
 
-    directive @federation__requires(fields: federation__FieldSet!) on FIELD_DEFINITION
+directive @key(fields: federation__FieldSet!, resolvable: Boolean = true) repeatable on OBJECT | INTERFACE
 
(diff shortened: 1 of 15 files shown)


apollo-federation/tests/subgraph/subgraph_validation_tests.rs
```diff
@@ -953,73 +953,67 @@ Did you mean "@shareable"?{}"#,
...
-    directive @federation__provides(fields: federation__FieldSet!) on FIELD_DEFINITION
+directive @key(fields: federation__FieldSet!, resolvable: Boolean = true) repeatable on OBJECT | INTERFACE
 
-    directive @federation__external(reason: String) on OBJECT | FIELD_DEFINITION
+directive @federation__requires(fields: federation__FieldSet!) on FIELD_DEFINITION
 
-    directive @federation__tag(name: String!) repeatable on FIELD_DEFINITION | OBJECT | INTERFACE | UNION | ARGUMENT_DEFINITION | SCALAR | ENUM | ENUM_VALUE | INPUT_OBJECT | INPUT_FIELD_DEFINITION
+directive @federation__provides(fields: federation__FieldSet!) on FIELD_DEFINITION
 
-    directive @federation__extends on OBJECT | INTERFACE
```
duckki, line 980: `@federation__extends` is missing?

apollo-federation/tests/subgraph/subgraph_validation_tests.rs
```diff
@@ -953,73 +953,67 @@ Did you mean "@shareable"?{}"#,
...
-    const EXPECTED_FULL_SCHEMA: &str = r#"
-    schema
-      @link(url: "https://specs.apollo.dev/link/v1.0")
-      @link(url: "https://specs.apollo.dev/federation/v2.0", import: ["@key"])
-    {
-      query: Query
-    }
-
-    directive @link(url: String, as: String, for: link__Purpose, import: [link__Import]) repeatable on SCHEMA
+    const EXPECTED_FULL_SCHEMA: &str = r#"schema {
```
duckki, line 961: We can drop `#[allow(dead_code)]`.

duckki (changes requested):
Looks good except for missing `@extends`. It may be easy to fix or we can just annotate the test and deal with it later.

apollo-federation/tests/subgraph/subgraph_validation_tests.rs
tninesling: I thought `@link` was fed 2 and `@extends` was fed 1. Are we supposed to have both of those in the same schema?

duckki: At least, that was what the original test checked.
It seems still available in 2.0: https://www.apollographql.com/docs/graphos/schema-design/federated-schemas/reference/directives#extends

dariuszkuc: It is not used for anything in fed v2 we only kept it for backwards compatibility reasons.

tninesling: Ah, we were only adding it for fed 1. Fixed to also include `@extends` for fed 2 and the ordering matches the original JS assertion. The primary diff is that we make a new `extend schema ...` for the links at the beginning, but I think addressing that is out of scope for this ticket

duckki (approved):
Awesome. We now match the `EXPECTED_FULL_SCHEMA`!

PR #7513, opened 2025-05-16
Fix validating pre-encoded symbols before the first expression in an absolute URL

dylan-apollo:
Motivating case is spaces, but this also handles other symbols we will automatically encode at runtime.

[CNN-804]: https://apollographql.atlassian.net/browse/CNN-804?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apollo-federation/src/sources/connect/validation/connect/http.rs
+++ b/apollo-federation/src/sources/connect/validation/connect/http.rs
@@ -384,6 +384,10 @@ fn validate_absolute_connect_url(
     let mut is_relative = true;
     let mut dynamic_in_domain = None;
 
+    // Consume each part of the string template that *should* result in a static, valid base URI (scheme+host+port).
+    // - if we don't encounter a scheme, it's not a valid base URL
+    // - once we encounter a character not allowed in scheme+host+port, stop consuming
+    // - if we encounter a dynamic part before we break, we have an illegal dynamic component
     for part in &url.parts {
         match part {
             Part::Constant(constant) => {
@@ -432,6 +436,8 @@ fn validate_absolute_connect_url(
         });
     }
 
+    // Evaluate the template, replacing all dynamic expressions with empty strings. This should result in a valid
+    // URL because of the URL building logic in `interpolate_uri`, even if the result is illogical with missing values.
     let url = url
         .interpolate_uri(&Default::default())
         .map_err(|err| Message {
--- a/apollo-federation/src/sources/connect/validation/connect/http.rs
+++ b/apollo-federation/src/sources/connect/validation/connect/http.rs
@@ -6,7 +6,6 @@ use std::str::FromStr;
 use apollo_compiler::Name;
 use apollo_compiler::Node;
 use apollo_compiler::ast::Value;
-use http::Uri;
 use multi_try::MultiTry;
 use shape::Shape;
 
@@ -16,7 +15,6 @@ use crate::sources::connect::spec::schema::CONNECT_BODY_ARGUMENT_NAME;
 use crate::sources::connect::spec::schema::CONNECT_SOURCE_ARGUMENT_NAME;
 use crate::sources::connect::spec::schema::HTTP_ARGUMENT_NAME;
 use crate::sources::connect::string_template;
-use crate::sources::connect::string_template::Constant;
 use crate::sources::connect::string_template::Part;
 use crate::sources::connect::string_template::StringTemplate;
 use crate::sources::connect::validation::Code;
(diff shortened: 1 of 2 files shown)


apollo-federation/src/sources/connect/validation/connect/http.rs
```diff
@@ -383,65 +381,69 @@ fn validate_absolute_connect_url(
...
-        return Err(relative_url_error());
-    };
-
-    let Some((_scheme, without_scheme)) = first.value.split_once("://") else {
-        return Err(relative_url_error());
-    };
+    let mut is_relative = true;
+    let mut dynamic_in_domain = None;
+
+    for part in &url.parts {
```
lennyburdette, line 387:
```suggestion
    // Consume each part of the string template that *should* result in a static, valid base URI (scheme+host+port).
    // - if we don't encounter a scheme, it's not a valid base URL
    // - once we encounter a character not allowed in scheme+host+port, stop consuming
    // - if we encounter a dynamic part before we break, we have an illegal dynamic component
    for part in &url.parts {
```

lennyburdette (approved):

apollo-federation/src/sources/connect/validation/connect/http.rs
```diff
@@ -383,65 +381,75 @@ fn validate_absolute_connect_url(
...
-        return Err(relative_url_error());
-    };
+    let mut is_relative = true;
+    let mut dynamic_in_domain = None;
+
+    // Consume each part of the string template that *should* result in a static, valid base URI (scheme+host+port).
+    // - if we don't encounter a scheme, it's not a valid base URL
+    // - once we encounter a character not allowed in scheme+host+port, stop consuming
+    // - if we encounter a dynamic part before we break, we have an illegal dynamic component
+    for part in &url.parts {
```
naomijub, line 391: How about doing it without mutability:

```rust
let (is_relative, dynamic_in_domain) = url.parts.iter()
        .try_fold((true, None), |(is_relative, dynamic_in_domain), part| {
            match part {
                Part::Constant(constant) => {
                    let (relative, value) = match constant.value.split_once("://") {
                        Some((_schema, rest)) => (false, rest),
                        None => (is_relative, constant.value.as_str()),
                    };
                    if value.contains('/') || value.contains('?') || value.contains('#') {
                        return Err(());
                    }
                    Ok((relative && is_relative, dynamic_in_domain))
                },
                Part::Expression(dynamic) => Ok((is_relative, Some(dynamic))),
            }
        }).map_err(|err| ...)?;
```

naomijub: In this case, you can also move the 
```rust
 if let Some(dynamic) = dynamic_in_domain {
        return Err(Message {
            code: Code::InvalidUrl,
            message: format!(
                "{coordinate} must not contain dynamic pieces in the domain section (before the first `/` or `?`).",
            ),
            locations: str_value
                .line_col_for_subslice(dynamic.location.clone(), schema)
                .into_iter()
                .collect(),
        });
    }
```
To be an early exit in `Part::Expression` pattern, early exiting

dylan-apollo: I tried riffing on this a few ways to get the code clear enough to avoid the comments, but the different phases are so intertwined because the start of path `/` is a sub-pattern of the scheme indicator `://`. So I can't think of a way to break them up enough that the logic looks clear.

I'm also realizing that this won't be an issue once we start allowing dynamic expressions in domains—or at least the issue will be different, because the interpolate step will still fail with an unhelpful message.

So I'm going to leave as-is knowing that we have to come back to it in a future PR.

naomijub: Maybe this?
```rust
    let (is_relative, dynamic_in_domain) = url.parts.iter()
        // Consume each part of the string template that *should* result in a static, valid base URI (scheme+host+port).
        .map(|part| match part {
                Part::Constant(constant) => {
                    let (relative, value) = match constant.value.split_once("://") {
                        Some((_schema, rest)) => (false, rest),
                        // if we don't encounter a scheme, it's not a valid base URL
                        None => (true, constant.value.as_str()),
                    };
                    (relative, value,  None)
                },
                // if we encounter a dynamic part before we break, we have an illegal dynamic component
                Part::Expression(dynamic) => (true, "", Some(dynamic))
            })
            // once we encounter a character not allowed in scheme+host+port, stop consuming
            .take_while(|(_is_relative, value, _dynamic_in_domains)| 
                !value.contains('/') && !value.contains('?') && !value.contains('#'))
            .map(|(is_relative, _, dynamic_in_domain)| (is_relative, dynamic_in_domain))
            .find(|(is_relative, dynamic_in_domain)| *is_relative || dynamic_in_domain.is_some())
            .unwrap_or((false, None));
```

PR #7514, opened 2025-05-16
chore: update composition stubs

dariuszkuc:
* Update `Supergraph<Merged>` state to expose and hold `Valid<Schema>`
* Adds pre_/post_ merge validation steps

### DIFF
--- a/apollo-federation/src/composition/mod.rs
+++ b/apollo-federation/src/composition/mod.rs
@@ -77,7 +77,7 @@ pub fn merge_subgraphs(
 }
 
 pub fn post_merge_validations(
-    _supergraph: &Supergraph<Merged>
+    _supergraph: &Supergraph<Merged>,
 ) -> Result<(), Vec<FederationError>> {
     panic!("post_merge_validations is not implemented yet")
 }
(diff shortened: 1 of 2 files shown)


apollo-federation/src/composition/mod.rs
```diff
@@ -57,8 +63,21 @@ pub fn validate_subgraphs(
...
+    panic!("pre_merge_validations is not implemented yet")
+}
+
 pub fn merge_subgraphs(
     _subgraphs: Vec<Subgraph<Validated>>,
 ) -> Result<Supergraph<Merged>, Vec<FederationError>> {
     panic!("merge_subgraphs is not implemented yet")
 }
+
+pub fn post_merge_validations(
```
clenfest, line 79: Is this necessary? I think Sachin already did this with his PR this week: https://github.com/apollographql/router/pull/7451

I think we just need to call `validate_merged_schema` instead of creating this function.

dariuszkuc: Those are just stubs and yes they might be unnecessary. Once we start filling in those stubs we'll determine whether we need it or not.

Sachin's PR does not use the new typestate approach and we might want to refactor it to use it.

clenfest (approved):

PR #7523, opened 2025-05-19
Update Router request termination docs

gwardwell:
The docs currently state:

> Your Rhai script can terminate the associated client request that triggered it. To do so, it must throw an exception from the supergraph service.

This is misleading as they can _also_ be terminated from the request service, which can be ideal for some cases since the router service request stage executes before the body is parsed of the operation is validated. So things like rejecting based on a missing header are likely better to be done here.

This PR attempts to update the docs to make it clear that the request can be terminated in the router service.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/docs/source/routing/customization/rhai/reference.mdx
+++ b/docs/source/routing/customization/rhai/reference.mdx
@@ -59,10 +59,10 @@ log_trace("trace-level log message");
 
 Your Rhai script can terminate the associated client request that triggered it. To do so, it must throw an exception from either the router service or supergraph service. This returns an `Internal Server Error` to the client with a `500` response code.
 
-When choosing between terminating the request from the router service versus the supergraph service, consider that:
+When choosing between router service versus supergraph service termination:
 
-* The request/response body is not available in the router service. If you need to examine the body when determining whether or not to terminate the request, you will need to do so in the supergraph service.
-* The router service request stage happens before the supergraph service request stage and before the request body is parsed and the operation is validated. So it may be best to perform some actions, like enforcing a required header, in the router service request stage so they can be consistently applied even when an invalid operation is provided by the client.
+- **Router service**: Cannot access request/response body and executes before parsing and validation; ideal for checks that should be consistently applied, like checking required headers, even when invalid operations are provided by the client.
+- **Supergraph service**: Use whenever you need to examine request bodies to enforce termination.
 
 For example:
 ```rhai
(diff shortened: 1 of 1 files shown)


Meschreiber (approved):
Thanks for this update @gwardwell ! Do we have any guidance on when you should use one service vs. the other? If so, would be great to include here, but nothing blocking on my end!

gwardwell:
@Meschreiber I'll try adding that. I think the guidance is:
- `router_service` doesn't have access to the request body, so if you need to examine that, it must be done in the `supergraph_service`
- `router_service` request stage happens before `supergraph_service` request stage and before the request is even parsed/validated, so some actions like enforcing the presence of a header are appropriate to do there because it will happen before something like an invalid operation is caught by request validation

I'll try to put that in more concise/clearer wording in the docs.

gwardwell:
@Meschreiber I added some additional info on when to choose router service over supergraph service.

docs/source/routing/customization/rhai/reference.mdx
```diff
@@ -57,11 +57,28 @@ log_trace("trace-level log message");
...
 
-Your Rhai script can terminate the associated client request that triggered it. To do so, it must throw an exception from the supergraph service. This returns an `Internal Server Error` to the client with a `500` response code.
+Your Rhai script can terminate the associated client request that triggered it. To do so, it must throw an exception from either the router service or supergraph service. This returns an `Internal Server Error` to the client with a `500` response code.
+
+When choosing between terminating the request from the router service versus the supergraph service, consider that:
+
+* The request/response body is not available in the router service. If you need to examine the body when determining whether or not to terminate the request, you will need to do so in the supergraph service.
+* The router service request stage happens before the supergraph service request stage and before the request body is parsed and the operation is validated. So it may be best to perform some actions, like enforcing a required header, in the router service request stage so they can be consistently applied even when an invalid operation is provided by the client.
```
Meschreiber, line 65: Suggestion to make this more concise:

```suggestion
When choosing between router service versus supergraph service termination:

- **Router service**: Cannot access request/response body and executes before parsing and validation; ideal for checks that should be consistently applied, like checking required headers, even when invalid operations are provided by the client.
- **Supergraph service**: Use whenever you need to examine request bodies to enforce termination.
```

Meschreiber (approved):
Amazing, tysm @gwardwell , left one copyedit to make this concise, but otherwise lgtm!

PR #7532, opened 2025-05-20
Port auth specs

tninesling:
Ports three remaining specs from JS:

- `@authenticated` ([source](https://github.com/apollographql/federation/blob/main/internals-js/src/specs/authenticatedSpec.ts))
- `@policy` and `Policy`: ([source](https://github.com/apollographql/federation/blob/main/internals-js/src/specs/policySpec.ts))
- `@requiresScopes` and `Scope`: ([source](https://github.com/apollographql/federation/blob/main/internals-js/src/specs/requiresScopesSpec.ts))

These are registered in the federation spec for the appropriate versions (v2.5+ for auth/policy and v2.6+ for requiresScopes). When imported directly from their own specs, they are expanded via `FederationBlueprint::expand_known_features`.

---

**Checklist**

- Tests added and passing[^3]
    - [X] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-federation/src/link/federation_spec_definition.rs
```diff
@@ -875,15 +879,35 @@ impl SpecDefinition for FederationSpecDefinition {
             ));
         }
 
+        if self.version().satisfies(&Version { major: 2, minor: 5 }) {
```
duckki, line 882: Just a thought: Maybe we should automate this method using the information in `AUTHENTICATED_VERSIONS`, `REQUIRES_SCOPES_VERSIONS`, etc, since they have `minimum_federation_version` data.

apollo-federation/src/schema/blueprint.rs
```diff
@@ -337,22 +342,103 @@ impl FederationBlueprint {
...
+
         let Some(links_metadata) = schema.metadata() else {
             return Ok(());
         };
 
         for link in links_metadata.links.clone() {
-            // TODO: Pick out known features by link identity and call `add_elements_to_schema`.
-            // JS calls coreFeatureDefinitionIfKnown here, but we don't have a feature registry yet.
-
+            if link.url.identity == Identity::authenticated_identity() {
```
duckki, line 366: nit: Could these repeating if-statements be factored out as a function?

duckki (commented):
Looks good. I just have a few questions and suggestions.

apollo-federation/src/link/authenticated_spec_definition.rs
```diff
@@ -0,0 +1,151 @@
...
+            false, // not repeatable
+            &[
+                DirectiveLocation::FieldDefinition,
+                DirectiveLocation::Object,
+                DirectiveLocation::Interface,
+                DirectiveLocation::Scalar,
+                DirectiveLocation::Enum,
+            ],
+            true, // composes
+            Some(&|v| AUTHENTICATED_VERSIONS.get_minimum_required_version(v)),
```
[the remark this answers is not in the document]
tninesling: I think the JS logic is incorrect but the bug hasn't surfaced because there is only one version of the spec. If we added a new incompatible version of this spec, we'd start including that for older federation versions which don't support it.

apollo-federation/src/link/federation_spec_definition.rs
tninesling: We could potentially change this to something like `if let Some(spec) = AUTH_VERSION.get_minimum_required_version(self.version()) ...`. I was mostly trying to stay consistent with JS. I think what I have now more clearly shows the mapping of fed version -> feature version.

apollo-federation/src/link/authenticated_spec_definition.rs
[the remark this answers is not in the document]
duckki: Makes sense. `get_minimum_required_version` sounds a bit misleading, since it returns the latest compatible version 🤯

PR #7533, opened 2025-05-20
conwuegb/fed 606

conwuegb:
*Description here*

Fixes #**issue_number**

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-federation/src/utils/definitions/fed2_0.graphqls
+++ b/apollo-federation/src/utils/definitions/fed2_0.graphqls
@@ -1,0 +1,52 @@
+#
+# https://specs.apollo.dev/federation/v2.0/federation-v2.0.graphql
+#
+
+directive @key(fields: FieldSet!, resolvable: Boolean = true) repeatable on OBJECT | INTERFACE
+directive @requires(fields: FieldSet!) on FIELD_DEFINITION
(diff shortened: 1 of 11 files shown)


duckki:
`FederationSpecDefinition` struct and `FEDERATION_VERSIONS` constant holds all the information you need. We don't need to duplicate the data into static files, when we could compute from the existing code.

I think it's just a matter of printing the data. But, we need to make sure exactly what format the information needs to be printed.

duckki:
Or, you could generate an empty subgraph schema with a particular federation version like this:

```rust
    fn new_empty_federation_subgraph_schema(version: &Version) -> Result<FederationSchema, FederationError> {
        let mut inner_schema = Schema::new();
        add_federation_link_to_schema(&mut inner_schema, version)?;
        expand_schema(inner_schema)
    }
```

The result should print pretty close to what was asked for (ignoring comments and ordering of elements).
I see a couple discrepancies that we can fix with a new option on the `add_federation_link_to_schema` (where we want to import types in addition to directives).

PR #7540, opened 2025-05-21
switch connectors feature gate from 0.2 to 0.3.

lennyburdette:
connect/v0.3 doesn't exist yet, but i'm optimistically adding support for it so that this code is still exercised

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/uplink/testdata/connectv0_3.router.yaml
+++ b/apollo-router/src/uplink/testdata/connectv0_3.router.yaml
@@ -5,4 +5,4 @@ homepage:
 limits:
   parser_max_recursion: 1000
 connectors:
-  preview_connect_v0_2: true
+  preview_connect_v0_3: true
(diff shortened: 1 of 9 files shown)


Hunter-apollo (approved):

apollo-router/src/uplink/testdata/connectv0_3.router.yaml
```diff
@@ -5,4 +5,4 @@ homepage:
 limits:
   parser_max_recursion: 1000
 connectors:
-  preview_connect_v0_2: true
+  preview_connect_v0_3: true
```
benjamn, line 8: So could we totally omit this `connectors` section now, and we'd get `connect/v0.2` support?

benjamn (approved):

apollo-router/src/uplink/testdata/connectv0_3.router.yaml
lennyburdette: correct, this was just a gate we added so we could merge code to dev/release router versions without allowing users to accidentally use 0.2 features. we're removing the gate when we're officially releasing 0.2.

andrewmcgivery (approved):

PR #7541, opened 2025-05-22
feat: add `cost` to query planner statistics 

briannafugate408:
We want to expose the query planning cost in router so that we can use it in harness to measure query planning cost as part of our SLO metrics. 

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-federation/src/query_plan/mod.rs
+++ b/apollo-federation/src/query_plan/mod.rs
@@ -23,7 +23,6 @@ pub type QueryPlanCost = f64;
 #[derive(Debug, Default, PartialEq, Serialize, Deserialize)]
 pub struct QueryPlan {
     pub node: Option<TopLevelPlanNode>,
-    /// `cost` can be NaN, if the cost is not computed or irrelevant.
     pub statistics: QueryPlanningStatistics,
 }
 
--- a/apollo-federation/src/query_plan/query_planner.rs
+++ b/apollo-federation/src/query_plan/query_planner.rs
@@ -172,6 +172,7 @@ impl Default for QueryPlannerDebugConfig {
 pub struct QueryPlanningStatistics {
     pub evaluated_plan_count: Cell<usize>,
     pub evaluated_plan_paths: Cell<usize>,
+    /// `best_plan_cost` can be NaN, if the cost is not computed or irrelevant.
     pub best_plan_cost: f64,
 }
 


apollo-federation/src/query_plan/mod.rs
```diff
@@ -37,6 +38,7 @@ pub enum TopLevelPlanNode {
     Defer(DeferNode),
     #[from(ConditionNode, Box<ConditionNode>)]
     Condition(Box<ConditionNode>),
+    Statistics(QueryPlanningStatistics),
```
duckki, line 41: Can you explain why we need this node variant?

apollo-federation/src/query_plan/query_planner.rs
```diff
@@ -863,25 +862,31 @@ fn compute_plan_for_defer_conditionals(
...
-                    on_final_operation,
-                )?
-                .map(Box::new),
+                if_clause: if_node.map(Box::new),
+                else_clause: else_node.map(Box::new),
             };
-            Ok(Some(PlanNode::Condition(Box::new(node))))
+            Ok((
+                Some(PlanNode::Condition(Box::new(node))),
+                if_cost + else_cost,
```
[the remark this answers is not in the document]
duckki: I reviewed this part of code. And taking the max of them should work.
In fact, we could simply take the `if_cost `, since the `else_cost` should always be smaller (after reducing defers). Having said that, let's not take the shortcut and just use the max for clarity. And it doesn't cost much computationally.

apollo-federation/cli/src/main.rs
```diff
@@ -306,6 +306,8 @@ fn cmd_plan(
         println!("{}", serde_json::to_string_pretty(&query_plan).unwrap());
     } else {
         println!("{query_plan}");
+        dbg!("query_plan cost: ", &query_plan.cost);
```
duckki, line 309: also, don't forget to remove them. It's safer not to add in the first place.

duckki: Use `git add -p <target>` to add blocks of code selectively.

duckki (approved):
Looks good!

PR #2223, opened 2022-12-06
Redeploy router if mounted supergraph schema changes 

toneill818:
Building off of @garypen's PR #2119, if the supergraphFile is set, add a label to the pods with a hash of its contents. This will force new pods to be spun up if the supergraphFile is changed so the new file will be loaded.

PR #7542, opened 2025-05-22
Helm: Use the supergraph file provided

### LINKED ISSUE #2491: Unable to load local supergraph schema file using the Helm chart
**Describe the bug**
I am attempting to load a local supergraph schema GraphQL file using the Helm chart value, however, the file is not mounting or passing any configuration to the router executable within the container.

**To Reproduce**
1. `helm install apollo-router --set-file supergraphFile=supergraph-schema.graphql ./path-to-chart`.
2. Wait for deployment and service to come up in Kubernetes.
3. Inspect the Apollo Router logs and see the error: `The Apollo Router requires a composed supergraph schema at startup.`

It appears that a ConfigMap called `apollo-router-supergraph` is created with the `supergraph-schema.graphql` file in it. However, the ConfigMap is never mounted to the container like the regular configuration one is. I do see a checksum annotation at the top of `deployment.yaml`, but the `supergraphFile` value nor it's ConfigMap is ever referenced again.

**Expected behavior**
When I supply a `supergraphFile`, I expect the file to be added to a ConfigMap and mounted to the container as a volume. I also expect Apollo Router to automatically use this supergraph schema file if not managed federation options are provided. Currently, it appears that there are no facilities to mount the schema file or automatically modify the args to use it.

**Output**
```
Apollo Router v1.9.0 // (c) Apollo Graph, Inc. // Licensed as ELv2 (https://go.apollo.dev/elv2)

⚠️  The Apollo Router requires a composed supergraph schema at startup. ⚠️

👉 DO ONE:

  * Pass a local schema file with the '--supergraph' option:

      $ ./router --supergraph <file_path>

  * Fetch a registered schema from Apollo Studio by setting
    these environment variables:

      $ APOLLO_KEY="..." APOLLO_GRAPH_REF="..." ./router

      For details, see the Apollo docs:
      https://www.apollographql.com/docs/router/managed-federation/setup

🔬 TESTING THINGS OUT?

  1. Download an example supergraph schema with Apollo-hosted subgraphs:

    $ curl -L https://supergraph.demo.starstuff.dev/ > starstuff.graphql

  2. Run the Apollo Router in development mode with the supergraph schema:

    $ ./router --dev --supergraph starstuff.graphql

    

Process finished with exit code 0
```

 - OS: macOS 13.2 (arm64 and amd64)
 - Version 1.9.0

abernix:
I saw @lambertjosh stumble on this at one point and I wrote this quick fix but never opened the PR. 

I'm not sure what the impact of using a single configmap versus multiple, but thought I'd get the conversation started.

Fixes #2491 

---

**Checklist**

- [x] Documentation[^2]
- Tests added and passing[^3]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [pending] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


helm/chart/router/templates/configmap.yaml
```diff
@@ -21,4 +21,8 @@ metadata:
 data:
   configuration.yaml: |
     {{- toYaml $configuration | nindent 4 }}
+  {{- if .Values.supergraphFile }}
```
theJC, line 24: Is there any way to test the size of the file to ensure it is small enough to fit in a configmap ? (1 MiB)

abernix: That's a great question/consideration and the kind of feedback I was hoping for. 😄  My personal simple answer is — i don't know.  Perhaps that limitation is the reason that it was never fully implemented correctly, and perhaps this just needs to be refactored into a different approach.

garypen: I don't know of any way to check for a ConfigMap > 1MB from helm, apart from if you try and deploy it you will get an error.

garypen:
I think this has never been approached because:
 - This is an incompatible change and could break someone.
 - It's not an issue for GraphOS in general, since the supergraph is retrieved from the network.
I'm not

garypen:
I think there'd be some kind of impact on this: https://github.com/apollographql/router/blob/605ca17b2479a1bdab08f807e5cd259ef6a26c63/helm/chart/router/templates/deployment.yaml#L35

helm/chart/router/templates/configmap.yaml
abernix: An error when you try deploying might be acceptable.  Is there a pattern that wouldn't break that I'm missing?

abernix:
> * This is an incompatible change and could break someone.

Not sure I can understand the break.  How would it be incompatible under which circumstance? My understanding is it doesn't even work right now. Are we saying that someone might have figured out some other way to pass in `--supergraph` by extending the chart and we'd end up with two?

> I think there'd be some kind of impact on this:

what even _is_ that?  Just a checksum that we make available?

garypen:
> > * This is an incompatible change and could break someone.
> 
> Not sure I can understand the break. How would it be incompatible under which circumstance? My understanding is it doesn't even work right now. Are we saying that someone might have figured out some other way to pass in `--supergraph` by extending the chart and we'd end up with two?

It's incompatible because you are deleting the `supergraph-cm.yaml` template. That means anyone using the existing chart to mount that config map manually will find their helm deploys are broken.

I know I use that mechanism in a variety of different POC/Demo code examples.

> 
> > I think there'd be some kind of impact on this:
> 
> what even _is_ that? Just a checksum that we make available?

Yes. The motivation is described here: https://github.com/apollographql/router/pull/2223

I don't know if anyone is using/relying on this mechanism, but that's also a potential source of incompatibility.

garypen:
> ok, sounds like this would be more of a breaking change than i'd thought.
> 
> > It's incompatible because you are deleting the `supergraph-cm.yaml` template. That means anyone using the existing chart to mount that config map manually will find their helm deploys are broken.
> 
> is it possible to use Helm’s templating to support both mechanisms:
> 
> * If supergraphFile is provided, use a new mechanism.
> * If not specified, fall back to the existing ConfigMap mounting approach.
> 
> Does that fix the broken user experience without waiting for the next major breaking change of the router?

No, because that's how the existing mechanism works. e.g.:
```
{{- if .Values.supergraphFile }}
{{- $routerFullName := include "router.fullname" .  -}}
```
etc...

We could work-around that by adding something to trigger the new behaviour, but we'd have to retain the default as the current behaviour to preserve compatibility.

PR #7546, opened 2025-05-22
feat(federation/composition): port `advancePathWithDirectTransition()` 

sachindshinde:
This PR ports `advancePathWithDirectTransition()` from the JS codebase, along with any currently un-ported  downstream functions. It is stacked on [the refactor PR](https://github.com/apollographql/router/pull/7547).

### DIFF
--- a/apollo-federation/src/query_graph/graph_path/transition.rs
+++ b/apollo-federation/src/query_graph/graph_path/transition.rs
@@ -140,7 +140,7 @@ impl TransitionGraphPath {
 
     // PORT_NOTE: Named `findOverriddingSourcesIfOverridden()` in the JS codebase, but we've changed
     // sources to subgraphs here for clarity that we don't need to handle the federated root source.
-    fn find_overridding_subgraphs_if_overridden(
+    fn find_overriding_subgraphs_if_overridden(
         field_pos_in_subgraph: &FieldDefinitionPosition,
         field_subgraph: &Arc<str>,
         subgraphs: &IndexMap<Arc<str>, ValidFederationSchema>,
@@ -211,7 +211,7 @@ impl TransitionGraphPath {
         &self,
         transition: &QueryGraphEdgeTransition,
         condition_resolver: &mut impl ConditionResolver,
-        override_conditions: &EnabledOverrideConditions,
+        override_conditions: &Arc<EnabledOverrideConditions>,
     ) -> Result<Either<Vec<TransitionGraphPath>, UnadvanceableClosures>, FederationError> {
         ensure!(
             transition.collect_operation_elements(),
@@ -225,9 +225,8 @@ impl TransitionGraphPath {
         } = transition
         {
             let tail_weight = self.graph.node_weight(self.tail)?;
-            if let Ok(tail_type) = <QueryGraphNodeType as TryInto<
-                CompositeTypeDefinitionPosition,
-            >>::try_into(tail_weight.type_.clone())
+            if let Ok(tail_type) =
+                CompositeTypeDefinitionPosition::try_from(tail_weight.type_.clone())
             {
                 if field_definition_position.parent().type_name() != tail_type.type_name()
                     && !self.tail_is_interface_object()?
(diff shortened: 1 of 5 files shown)


apollo-federation/src/query_graph/graph_path/transition.rs
```diff
@@ -26,3 +55,508 @@ impl GraphPathTriggerVariant for QueryGraphEdgeTransition {
...
+        if let QueryGraphEdgeTransition::FieldCollection {
+            source,
+            field_definition_position,
+            ..
+        } = transition
+        {
+            let tail_weight = self.graph.node_weight(self.tail)?;
+            if let Ok(tail_type) = <QueryGraphNodeType as TryInto<
+                CompositeTypeDefinitionPosition,
+            >>::try_into(tail_weight.type_.clone())
```
duckki, line 230: How about `try_from`?
```suggestion
            if let Ok(tail_type) = CompositeTypeDefinitionPosition::try_from(tail_weight.type_.clone())
```

apollo-federation/src/query_graph/graph_path/transition.rs
```diff
@@ -26,3 +55,508 @@ impl GraphPathTriggerVariant for QueryGraphEdgeTransition {
...
+                .matches_supergraph_transition(transition)?
+            {
+                continue;
+            }
+
+            if let Some(override_condition) = &edge_weight.override_condition {
+                if !edge_weight.satisfies_override_conditions(override_conditions) {
+                    let graph = to_advance.graph.clone();
+                    let override_condition = override_condition.clone();
+                    let override_conditions = override_conditions.clone();
```
duckki, line 294: Looks like we will need to wrap `override_conditions` with `Rc` (or `Arc`) when we optimize the code.
Also, we can capture the `edge` here, instead of `override_conditions`. But, these can be done later when we optimize this code. I think it's worth creating a separate Jira ticket for optimization.

apollo-federation/src/query_graph/graph_path/transition.rs
```diff
@@ -26,3 +55,508 @@ impl GraphPathTriggerVariant for QueryGraphEdgeTransition {
...
+        );
+        Ok(format!(
+            " (please ensure that this is not due to key {} being accidentally marked @external)",
+            fields_list
+        ))
+    }
+
+    // PORT_NOTE: Named `findOverriddingSourcesIfOverridden()` in the JS codebase, but we've changed
+    // sources to subgraphs here for clarity that we don't need to handle the federated root source.
+    fn find_overridding_subgraphs_if_overridden(
```
duckki, line 143: spelling
```suggestion
    fn find_overriding_subgraphs_if_overridden(
```

apollo-federation/src/query_graph/graph_path/transition.rs
```diff
@@ -26,3 +55,508 @@ impl GraphPathTriggerVariant for QueryGraphEdgeTransition {
...
+                            .directives
+                            .get(external_directive_definition_name)
+                        else {
+                            bail!(
+                                "{} in {} is not external but there is no corresponding edge",
+                                field_pos_in_subgraph,
+                                subgraph,
+                            );
+                        };
+                        let external_reason = metadata
```
duckki, line 518: A comment block in JS code is not ported here.
```typescript
            // but the field is external in the "subgraph-extracted-from-the-supergraph", but it might have been forced to an external
            // due to being a used-overriden field, in which case we want to amend the message to avoid confusing the user.
            // Note that the subgraph extraction marks such "forced external due to being overriden" by setting the "reason" to "[overridden]".
```
It seems worth porting. But, we need to correct some typos "overriden" => "overridden".

duckki (approved):
Looks good. Approved with just a few minor suggestions.

apollo-federation/src/query_graph/graph_path/transition.rs
sachindshinde: Good callout, it was simple enough that I just did it now (the closure re-reads the `edge`'s `override_condition`, and `override_conditions` is in an `Arc`).

apollo-federation/src/query_graph/graph_path/transition.rs
sachindshinde: Good catch, ported the comment over with corrections.

PR #7547, opened 2025-05-22
chore(federation/composition): Split trigger-specific code from `graph_path.rs`

sachindshinde:
The upcoming satisfiability PRs are going to add a lot of code to `graph_path.rs` for transition-trigger path logic, so this PR splits out trigger-specific code into `graph_path/operation.rs` and `graph_path/transition.rs` (note this was done via IDE, so while the LOC is high, it shouldn't contain bugs/behavioral changes hopefully).


duckki (approved):
Looks good!

PR #7549, opened 2025-05-22
Port tag and shareable validations

tninesling:
# Validations for `@shareable`

Ports validation rules for `@shareable` and enables relevant tests for subgraph validation. The rules enforced are:

- `@shareable` cannot be applied twice on the same object type definition, but applying `@shareable` both to a type definition and an extension of the same type is fine.
- `@shareable` cannot be applied twice on a field definition. Like the previous rule, this is also validated via [validateShareableNotRepeatedOnSameDeclaration](https://github.com/apollographql/federation/blob/main/internals-js/src/federation.ts#L1808) in the JS version, but our Rust implementation doesn't provide extension ids for directives on field definitions. I've interpreted this to mean that we just care about not repeating `@shareable` on the given definition, so I updated this to use a simple `len()` check instead.
- `@shareable` cannot be applied to interface fields

# Validation for `@tag`

This change creates tests for validating user-defined `@tag` definitions which may conflict with our federation-provided definition. I originally implemented this in the validation phase but realized that the errors were being produced during the `expand_links` phase. The actual validation occurs inside `check_or_add` when we expand links to various specs, so I haven't bothered adding the redundant check in the validation phase, as I'm unaware of flows which would skip the link expansion call.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [X] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


duckki (approved):
Looks great!

PR #7557, opened 2025-05-23
Remove `tracing::event!` calls inside tracing layers

goto-bus-stop:
Using `tracing::error!()` or other event macros inside a
tracing-subscriber layer callback may just drop the event.

Elsewhere, we've solved this by manually building an event and using
`context.event()`.
https://github.com/apollographql/router/blob/678dee7e1a85d4520d84e90ef0913b892a55ee65/apollo-router/src/plugins/telemetry/reload.rs#L175-L187

I think this is not appropriate to do in this case. If an error
occurred inside the tracing-subscriber layer, we shouldn't expect to be
able to record new events with `tracing`. So I simply changed these
calls to `eprintln!()`. The messages will at least go _somewhere_.

I think there's an argument that these could be panics, but I don't
want to risk breaking the world.

[ROUTER-1016]: https://apollographql.atlassian.net/browse/ROUTER-1016?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apollo-router/src/plugins/telemetry/fmt_layer.rs
+++ b/apollo-router/src/plugins/telemetry/fmt_layer.rs
@@ -126,7 +126,7 @@ where
                 extensions.insert(fields);
             }
         } else {
-            eprintln!("Span not found, this is a bug");
+            eprintln!("FmtLayer::on_new_span: Span not found, this is a bug");
         }
     }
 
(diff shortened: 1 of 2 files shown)


carodewig (commented):
Would it be worth updating the printed line to include the function name? It would be difficult to debug which span was missing without context.

Perhaps something like this:
```rust
eprintln!("Span not found within fmt_layer::on_record");
```

goto-bus-stop:
For posterity, it's intentional that there is no changeset here, due to how internal this change is

PR #7559, opened 2025-05-23
Take advantage of improvement in the http crate for cloning

garypen:
In an older version of the crate, `http::Extensions` were not `Clone`. This had the consequence that `http::Request` and `http::Response` were also not clone. To work-around this limitation, we had functions in our http_ext module which would "clone" a `Request` or a `Response`, but they'd omit `Extensions`.

Generally speaking, this wasn't a problem for most use cases. However, over time we started to add functionality, such as file uploads, to the router which relied on the correct behaviour of cloning and expected Extensions to be preserved across a clone.

This finally resulted in an issue for a customer who could not understand why modifying a request in a Rhai script would cause a File Upload to fail.

It transpires that at version 1.0.0 of the http crate, an incompatible change was made which forces `http::Extensions` to be `Clone` which means both `http::Request` and `http::Response` are now `Clone`.

I've replaced our `http::Extensions` ignoring implementation with the provided `Clone` behaviour and this means that Request and Response Extensions are now correctly cloned. I've added tests for this as well.

Note: This can't be easily back-ported to 1.x. `http` is used all over the place and a test compile results in:
```
error: could not compile `apollo-router` (lib) due to 114 previous errors
```
I think that updating the code would be very intrusive.

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-1259]: https://apollographql.atlassian.net/browse/ROUTER-1259?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apollo-router/src/http_ext.rs
```diff
@@ -216,27 +216,22 @@ pub(crate) fn header_map(
...
+    request.clone()
 }
 
-/// Ignores `http::Extensions`
+// In an earlier version of the http crate, `Response` didn't implement `Clone`, so we
+// implemented partial support (excluding Extensions) for `Clone`. `Response` does now implement
+// `Clone`, so this function does now clone Responses fully.
+//  - https://github.com/hyperium/http/pull/634
+//  - https://github.com/hyperium/http/releases/tag/v1.0.0
 pub(crate) fn clone_http_response<B: Clone>(response: &http::Response<B>) -> http::Response<B> {
```
goto-bus-stop, line 233: I think we should remove these functions and replace the callsites. Primarily because the comment in this callsite is now incorrect https://github.com/apollographql/router/blob/678dee7e1a85d4520d84e90ef0913b892a55ee65/apollo-router/src/services/router/service.rs#L709-L710

goto-bus-stop: Oh, I didn't see you updated the comment, then it doesn't matter :)

.changesets/fix_garypen_http_ext_clone_works_properly.md
```diff
@@ -0,0 +1,9 @@
+### Take advantage of improvement in the http crate for cloning ([PR #7559](https://github.com/apollographql/router/pull/7559))
```
goto-bus-stop, line 1: I think this changeset contains too much technical depth for a typical user. Maybe we could focus just on the known case that was fixed here?

goto-bus-stop (commented):
I'd like to see an integration test that uses file upload + rhai...

goto-bus-stop (approved):
lovely 😁

PR #7562, opened 2025-05-23
Fix remaining subgraph validation tests

tninesling:
# Overview

Fixes the three remaining tests we had disabled for subgraph validation, with a few logic updates to make them work.

## Interface object message duplication

Fixes `key_on_interfaces_require_key_on_all_implementations`

We had a slight logic error in the interface object validation which aims to collect all implementing types which are missing keys for a given interface. The fix here is a slight indentation change to match the [JS source](https://github.com/apollographql/federation/blob/main/internals-js/src/federation.ts#L975-L999) (notice the if block for reporting errors is level with the loop above instead of inside).

## Interface object requires resolvable key

Fixes `key_on_interfaces_with_key_on_some_implementation_non_resolvable`

Here we were wrongly allowing a non-resolvable key in an implementing type because the validation asserts that the implementing type has the same value for `resolvable` as the interface it implements. The `@key(resolvable:)` argument should default to `true`, per the definition, but we were using `.unwrap_or(false)` inside `FederationSpecDefinition::key_directive_arguments`.

## Compatible directive redefinitions

Fixes the test formerly known as `allows_known_directives_with_incomplete_but_compatible_definitions`

This test had a bunch of test cases wrapped up into one, so I've split it into a few individual tests to help pinpoint the actual issue. There was a small logic error inside `ensure_same_arguments` where we cloned and mutated the type we were looking at to make it nullable, but a check below relied on using the original value [here](https://github.com/apollographql/federation/blob/main/internals-js/src/directiveAndTypeSpecification.ts#L417). The upshot of this change is that we now properly allow redefining an optional directive argument as a required one. 

Additionally,  there was one small issue where the test rewrote the `link__Purpose` enum to a custom scalar. The JS test allows this, but Rust rejects it. From the validation code, I think this should be rejected, so I've updated the test to use the typical enum for `Purpose`. That particular rewrite is not the point of the test anyway, so I think it's a valid change.

---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [X] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/.changesets/docs_baby_fellow_hard_historian.md
+++ b/.changesets/docs_baby_fellow_hard_historian.md
@@ -1,0 +1,5 @@
+### Rhai script examples for returning Demand Control metrics as response headers ([PR #7564](https://github.com/apollographql/router/pull/7564))
+
+Added a new section to the [demand control documentation](https://www.apollographql.com/docs/graphos/routing/security/demand-control#accessing-programmatically) showing how to use Rhai scripts to expose cost estimation data in response headers. This allows clients to see the estimated cost, actual cost, and other demand control metrics directly in HTTP responses, which is useful for debugging and client-side optimization.
+
+By [@abernix](https://github.com/abernix) in https://github.com/apollographql/router/pull/7564
--- a/.changesets/feat_bnjjj_changeset_7070.md
+++ b/.changesets/feat_bnjjj_changeset_7070.md
@@ -1,0 +1,19 @@
+### Allow ignoring irrelevant headers during subscriptions de-duplication ([PR #7070](https://github.com/apollographql/router/pull/7070))
+
+The router now supports ignoring specific headers when deduplicating requests to subgraphs providing subscription events.. Previously, any differing headers which didn't actually affect the subscription response (e.g., `user-agent`) would prevent or limit the potential of deduplication.
+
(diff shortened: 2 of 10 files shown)


apollo-federation/tests/subgraph/subgraph_validation_tests.rs
```diff
@@ -1268,7 +1282,10 @@ type Query {
                 ) repeatable on SCHEMA
 
                 scalar link__Import
-                scalar link__Purpose
+                enum link__Purpose {
```
duckki, line 1285: Wait. `scalar link__Purpose` itself should be fine. The test was modified from the JS version.

Original test schema:
```graphql
        extend schema
          @link(
            url: "https://specs.apollo.dev/federation/v2.0"
            import: ["@key"]
          )

        type T @key(fields: "k") {
          k: ID!
        }

        directive @link(
          url: String!
          as: String
          for: link__Purpose
          import: [link__Import]
        ) repeatable on SCHEMA

        scalar link__Import
        scalar link__Purpose
```
The `@link(url: "https://specs.apollo.dev/link/v1.0")` line is missing here. Without that, both JS and Rust compositions fail.

duckki (changes requested):
A minor test bug. Otherwise, it looks good!

apollo-federation/tests/subgraph/subgraph_validation_tests.rs
tninesling: Oh, nice. I didn't realize that was failing because we were missing the link spec. Thanks!

PR #7576, opened 2025-05-27
`urlSafe` method for connectors

naomijub:
Implements `urlSafe` method for connectors

Fixes [#**CNN-730**](https://apollographql.atlassian.net/browse/CNN-730)

---

**Checklist**

- Tests added and passing[^3]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-federation/src/sources/connect/json_selection/methods/future/and.rs
```diff
@@ -48,7 +48,11 @@ fn and_method(
             None,
             vec![ApplyToError::new(
                 format!("Method ->{} requires arguments", method_name.as_ref()),
-                input_path.to_vec(),
+                input_path
+                    .to_vec()
+                    .into_iter()
+                    .map(|safe_json| safe_json.into())
+                    .collect(),
```
benjamn, line 55: I don't think we use `InputPath::to_vec` for anything other than this, so we might be able to change its return type to avoid most of the noisy changes in this PR (i.e., lines like this could stay `input_path.to_vec(),`).

naomijub: a68308e12

PR #7580, opened 2025-05-27
fix(federation/qp): suppress root type conditions in subgraph fetch operations

duckki:
---

**Checklist**

- Tests added and passing[^3]
    - [ ] Unit Tests
    - [x] Integration Tests
    - [x] Manual Tests

### DIFF
--- a/apollo-federation/src/operation/mod.rs
+++ b/apollo-federation/src/operation/mod.rs
@@ -1699,8 +1699,9 @@ impl SelectionSet {
                 let Some(sub_selection_type) = element.sub_selection_type_position()? else {
                     return Err(FederationError::internal("unexpected error: add_at_path encountered a field that is not of a composite type".to_string()));
                 };
+                let element_key = element.key().to_owned_key();
                 let mut selection = Arc::make_mut(&mut self.selections)
-                    .entry(ele.key())
+                    .entry(element_key.as_borrowed_key())
                     .or_insert(|| {
                         Selection::from_element(
                             element,
--- a/apollo-federation/src/operation/rebase.rs
+++ b/apollo-federation/src/operation/rebase.rs
@@ -398,6 +398,12 @@ impl InlineFragment {
                 CompositeTypeDefinitionPosition::try_from(rebased_condition_position)
             })
         {
+            // Root types can always be rebased and the type condition is unnecessary.
+            // Moreover, the actual subgraph might have renamed the root types, but the
+            // supergraph schema does not contain that information.
+            if parent_schema.is_root_type(rebased_condition.type_name()) {
+                return (true, None);
+            }
             // chained if let chains are not yet supported
             // see https://github.com/rust-lang/rust/issues/53667
             if runtime_types_intersect(parent_type, &rebased_condition, parent_schema) {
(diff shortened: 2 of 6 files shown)


apollo-federation/src/operation/rebase.rs
```diff
@@ -398,6 +398,12 @@ impl InlineFragment {
                 CompositeTypeDefinitionPosition::try_from(rebased_condition_position)
             })
         {
+            // Root types can always be rebased and the type condition is unnecessary.
+            // Moreover, the actual subgraph might have renamed the root types, but the
+            // supergraph schema does not contain that information.
+            if parent_schema.is_root_type(rebased_condition.type_name()) {
```
sachindshinde, line 404: Note this is only safe if the rebased type condition equals the parent type, which you should check for here.

If it doesn't equal the parent type, for now, we can just continue with the rest of the logic. (This means the PR only fixes the bug in some cases, but fixing the bug in those other cases is non-trivial, and those other cases are pretty rare to begin with. The full fix probably won't happen until after the source-awareness initiative is finished.)

sachindshinde (changes requested):

apollo-federation/src/schema/mod.rs
```diff
@@ -168,6 +168,13 @@ impl FederationSchema {
         self.get_type(type_name).ok()
     }
 
+    pub(crate) fn is_root_type(&self, type_name: &Name) -> bool {
```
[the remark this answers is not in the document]
goto-bus-stop: default-named root types should be handled by apollo-rs already

apollo-federation/src/operation/mod.rs
```diff
@@ -1699,8 +1699,9 @@ impl SelectionSet {
                 let Some(sub_selection_type) = element.sub_selection_type_position()? else {
                     return Err(FederationError::internal("unexpected error: add_at_path encountered a field that is not of a composite type".to_string()));
                 };
+                let element_key = element.key().to_owned_key();
                 let mut selection = Arc::make_mut(&mut self.selections)
-                    .entry(ele.key())
+                    .entry(element_key.as_borrowed_key())
```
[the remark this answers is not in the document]
duckki: It was a necessary change. After `can_rebase_on` change, some tests failed at `.or_insert` call below because `ele.key()` and `element.key()` became different in some cases (`... on Query` vs `...`).

In general, `ele` and `element` may be different and it makes sense to look up `element.key()`, instead of `ele.key()`. It appears that resulting in a different key via `rebase` here never happened before.

BTW, `to_owned_key`/`as_borrowed_key` calls had to be added, since `element` itself is going to move into the closure below. So, the key had to be cloned using `to_owned_key` (the normal `clone()` would only do a shallow cloning).

apollo-federation/src/operation/rebase.rs
duckki: Thanks. I added a comment about that.

sachindshinde: I don't see the updated condition in this method, it still says
```rust
if parent_schema.is_root_type(rebased_condition.type_name()) {
    return (true, None);
}
```
You'll want to check that, but also that `parent_type` equals `rebased_condition`.

duckki: Oh, I forgot to push the commit. The check is in place now.

PR #7584, opened 2025-05-28
port enum_type code from JS

clenfest:
## Summary
- Implement comprehensive enum merging functionality for Apollo Federation schema composition
- Add support for union vs intersection merging strategies based on enum usage (input vs output)
- Integrate with federation spec definitions for proper `@join__enumValue` directive handling

## Key Features

- **Enum Merging Logic**: Handles union strategy for output enums and intersection strategy for input enums
- **Federation Spec Integration**: Uses `JoinSpecDefinition` instead of hardcoded directive names
- **Comprehensive Error/Hint Reporting**: Detailed reporting for inconsistent enum values with subgraph information
- **Test Coverage**: 11 comprehensive tests covering all edge cases and scenarios

## Test plan

All 11 new tests pass covering enum merging scenarios, builds successfully, and code is properly formatted.


clenfest:
I talked about refactoring into a new file `merge_enum.rs` file. Will do that in a separate PR afterwards.

apollo-federation/src/merger/merger.rs
```diff
@@ -199,6 +242,371 @@ impl Merger {
...
+
+    fn record_applied_directives_to_merge<T>(
+        &mut self,
+        _sources: &Sources<Option<T>>,
+        _dest: &mut T,
+    ) {
+        todo!("Implement record_applied_directives_to_merge")
+    }
+
+    fn add_join_enum_value(
```
tninesling, line 441: Would it be better if this lived inside the `JoinSpecDefinition`? I'm thinking anything that switches logic based on the spec version or the spec name prefix might be better encapsulated in the specs themselves, and logic like this can just use dispatch without worrying about the details.

apollo-federation/src/merger/merger.rs
```diff
@@ -199,6 +242,371 @@ impl Merger {
...
+        // Check if this join spec version supports @join__enumValue (added in v0.3)
+        if *self.join_spec_definition.version()
+            < (crate::link::spec::Version { major: 0, minor: 3 })
+        {
+            return Ok(());
+        }
+
+        // Get the directive name from the join spec definition
+        // In the merged schema, join spec directives are prefixed with "join__"
+        // TODO: We are somewhat limited by the fact that the merged schema is a Schema rather than a FederationSchema.
```
tninesling, line 455: Should we wrap it in a `FederationSchema`? I believe `FederationSchema` is primarily meant to handle schema mutations, which we'll be doing a lot of in the merge phase.

apollo-federation/src/merger/merger.rs
```diff
@@ -199,6 +242,371 @@ impl Merger {
...
+            if !enum_type.values.contains_key(value_name) {
+                self.report_mismatch_hint(
+                    HintCode::InconsistentEnumValueForOutputEnum,
+                    format!(
+                        "Value \"{}\" of enum type \"{}\" has been added to the supergraph but is only defined in a subset of the subgraphs defining \"{}\": ",
+                        value_name, dest_name, dest_name
+                    ),
+                    sources,
+                    |source| {
+                        if let Some(enum_type) = source {
```
tninesling, line 599: Feel free to ignore, but in these situations I like to use `enum_type.is_some_and(|x| ...)` to reduce the nesting

apollo-federation/src/link/join_spec_definition.rs
```diff
@@ -776,6 +779,29 @@ impl JoinSpecDefinition {
...
+        &self,
+        target: &mut Component<EnumValueDefinition>,
+        _subgraph_name: &str,
+    ) -> Result<(), FederationError> {
+        // Check if this join spec version supports @join__enumValue (added in v0.3)
+        if *self.version() < (Version { major: 0, minor: 3 }) {
+            return Ok(()); // Should this be an error?
+        }
+
+        let _join_str = format!("join__{}", JOIN_ENUMVALUE_DIRECTIVE_NAME_IN_SPEC);
```
TylerBloom, line 792: nit: Variables that are used shouldn't start with an `_`.

apollo-federation/src/link/join_spec_definition.rs
```diff
@@ -776,6 +779,29 @@ impl JoinSpecDefinition {
         ))
     }
 
+    pub(crate) fn add_join_enum_value(
+        &self,
+        target: &mut Component<EnumValueDefinition>,
+        _subgraph_name: &str,
```
TylerBloom, line 785: nit: Variables that are used shouldn't start with an _.

apollo-federation/src/merger/merger.rs
```diff
@@ -199,13 +239,677 @@ impl Merger {
...
         todo!("Implement post-merge validations")
     }
+
+    /// Merge enum type from multiple subgraphs
+    pub(crate) fn merge_enum(
+        &mut self,
+        sources: Sources<&EnumType>,
+        dest: &mut EnumType,
+    ) -> Result<(), FederationError> {
+        let usage = self.enum_usages.get(&dest.name.to_string()).cloned().unwrap_or_else(|| {
```
TylerBloom, line 249: You don't need to create a new string here:
```suggestion
        let usage = self.enum_usages.get(dest.name.as_str()).cloned().unwrap_or_else(|| {
```

apollo-federation/src/link/join_spec_definition.rs
```diff
@@ -776,6 +779,29 @@ impl JoinSpecDefinition {
...
+    ) -> Result<(), FederationError> {
+        // Check if this join spec version supports @join__enumValue (added in v0.3)
+        if *self.version() < (Version { major: 0, minor: 3 }) {
```
conwuegb, line 788: Should this comparison be done with [satisfies()](https://github.com/apollographql/federation/blob/02ebb9ed09f157c4bcc00dbe27de6b57422b98cc/internals-js/src/specs/coreSpec.ts#L693-L701) instead?

apollo-federation/src/merger/merger.rs
```diff
@@ -199,13 +239,677 @@ impl Merger {
...
+            if let Some(source) = source {
+                for value in source.values.values() {
+                    // Note that we add all the values we see as a simple way to know which values there is to consider. But some of those value may
+                    // be removed later in `merge_enum_value`
+                    if !dest.values.contains_key(&value.node.value) {
+                        dest.values.insert(value.node.value.clone(), value.clone());
+                    }
+                }
+            }
+        }
```
TylerBloom, line 280: This is a broader style question: This can fairly cleanly be transformed into a series of chained combinators. What are folks' preference here? This would require 4 combinators (`filter_map`, `flat_map`, `filter`, `map`) + `.extend`. Is there a cap on the length of a chain we're ok with?

apollo-federation/src/link/inaccessible_spec_definition.rs
```diff
@@ -547,7 +547,7 @@ fn validate_inaccessible_in_fields(
 }
 
 /// Generic way to check for @inaccessible directives on a position or its parents.
-trait IsInaccessibleExt {
+pub(crate) trait IsInaccessibleExt {
```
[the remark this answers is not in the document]
clenfest: I'm calling is_inaccessible() in merger.rs. If it's not public(crate) it's only usable within this file.

apollo-federation/src/merger/merger.rs
```diff
@@ -199,13 +239,677 @@ impl Merger {
...
+
+        // Merge each enum value
+        let value_names: Vec<Name> = dest.values.keys().cloned().collect();
```
TylerBloom, line 283: We don't need to `collect` here. It just gets consumed by the for loop.

apollo-federation/src/merger/merger.rs
```diff
@@ -3,25 +3,53 @@ use std::collections::HashSet;
...
 use crate::merger::error_reporter::ErrorReporter;
+use crate::merger::hints::HintCode;
+use crate::schema::FederationSchema;
+use crate::schema::position::EnumValueDefinitionPosition;
 use crate::subgraph::typestate::Subgraph;
 use crate::subgraph::typestate::Validated;
 use crate::supergraph::CompositionHint;
 
+/// Type alias for Sources mapping - maps subgraph indices to optional values
+type Sources<T> = HashMap<usize, Option<T>>;
```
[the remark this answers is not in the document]
clenfest: I didn't realize that the ordering was preserved in JS. Will do.

apollo-federation/src/link/join_spec_definition.rs
clenfest: As discussed, we're just removing this code.

apollo-federation/src/merger/merger.rs
```diff
@@ -199,13 +250,617 @@ impl Merger {
     fn post_merge_validations(&mut self) {
         todo!("Implement post-merge validations")
     }
+
+    /// Merge enum type from multiple subgraphs
+    pub(crate) fn merge_enum(
+        &mut self,
+        sources: Sources<&EnumType>,
```
sachindshinde: When the caller runs this code, it's probably going to run into the problem where `EnumType`'s owner is `self`, so you can't pass in an `&mut` for `self` while also passing in an `&` for `EnumType`.

You'll probably want `Sources<Node<EnumType>>` instead; note that `Node` is basically `Arc`/cheap to clone. (Could technicaly use `Sources<EnumTypeDefinitionPosition>`, but its more cumbersome and unnecessary since subgraphs won't be mutated here.)

apollo-federation/src/merger/merger.rs
```diff
@@ -199,13 +250,617 @@ impl Merger {
...
+        dest: &EnumType,
```
sachindshinde: Similar story as above with this argument. You'll probably want `EnumTypeDefinitionPosition` here instead, since you'll be mutating the `EnumType` in question.

apollo-federation/src/merger/merger.rs
```diff
@@ -199,13 +250,617 @@ impl Merger {
...
+            sources
+                .iter()
+                .filter_map(|(_, source)| source.as_ref())
+                .flat_map(|source| source.values.values())
+                .map(|value| value.node.value.clone()),
+        );
+
+        // Merge each enum value
+        for value_name in enum_values {
+            let value_pos = EnumValueDefinitionPosition {
```
sachindshinde: You can use `EnumTypeDefinitionPosition::value()` here, with `dest`'s `EnumTypeDefinitionPosition`

apollo-federation/src/merger/merger.rs
```diff
@@ -199,13 +250,617 @@ impl Merger {
...
+                        name: spec.name.clone(),
+                        arguments: vec![Node::new(Argument {
+                            name: name!(JOIN_GRAPH_DIRECTIVE_NAME_IN_SPEC),
+                            value: Node::new(Value::Enum(name!(join_spec_name))),
+                        })],
+                    });
+                    let value_pos = EnumValueDefinitionPosition {
+                        type_name: value_pos.type_name.clone(),
+                        value_name: value_pos.value_name.clone(),
+                    };
```
sachindshinde: nit: For encapsulation sake, this kind of logic would typically go in `join_spec_definition.rs`, under a function named e.g. `enum_value_directive()`. [Here](https://github.com/apollographql/router/blob/b17521b49473af7df66426f52262cddd4c3d47ba/apollo-federation/src/link/federation_spec_definition.rs#L184-L208)'s an analogous example in `federation_spec_definition.rs`

apollo-federation/src/merger/merger.rs
```diff
@@ -199,13 +250,617 @@ impl Merger {
...
+            let dest = value_pos.get(self.merged.schema())?;
+
+            for (&idx, source) in sources.iter() {
+                if source.is_some() {
+                    // Get the join spec name for this subgraph
+                    let subgraph_name = &self.names[idx];
+                    let join_spec_name = self
+                        .subgraph_names_to_join_spec_name
+                        .get(subgraph_name)
+                        .ok_or_else(|| SingleFederationError::Internal {
```
sachindshinde: You can use the `bail!()` macro here.

apollo-federation/src/merger/merger.rs
```diff
@@ -199,13 +250,617 @@ impl Merger {
...
+    ) -> Result<(), FederationError> {
+        if let Some(spec) = self.join_spec_definition.enum_value_directive_spec() {
+            let dest = value_pos.get(self.merged.schema())?;
+
+            for (&idx, source) in sources.iter() {
+                if source.is_some() {
+                    // Get the join spec name for this subgraph
+                    let subgraph_name = &self.names[idx];
+                    let join_spec_name = self
+                        .subgraph_names_to_join_spec_name
```
sachindshinde: For the map `subgraph_names_to_join_spec_name`, the type of the value should be `Name` instead of `String`, as join spec names are always valid GraphQL names (and using `Name` in the type means you won't have to convert to `Name` later).

apollo-federation/src/merger/merger.rs
```diff
@@ -199,13 +250,617 @@ impl Merger {
...
+        // The merging strategy depends on the enum type usage:
+        //  - if it is _only_ used in position of Input type, we merge it with an "intersection" strategy (like other input types/things).
+        //  - if it is _only_ used in position of Output type, we merge it with an "union" strategy (like other output types/things).
+        //  - otherwise, it's used as both input and output and we can only merge it if it has the same values in all subgraphs.
+        // So in particular, the value will be in the supergraph only if it is either an "output only" enum, or if the value is in all subgraphs.
+        // Note that (like for input object fields), manually marking the value as @inaccessible let's use skips any check and add the value
+        // regardless of inconsistencies.
+        if !is_inaccessible
+            && !matches!(usage, EnumTypeUsage::Output { .. })
+            && !matches!(usage, EnumTypeUsage::Unused)
```
sachindshinde: This can be
```suggestion
            && !matches!(usage, EnumTypeUsage::Output { .. } | EnumTypeUsage::Unused)
```

apollo-federation/src/merger/merger.rs
```diff
@@ -199,13 +250,617 @@ impl Merger {
...
+                        sources,
+                        |source| {
+                            source.map_or("no", |enum_type| {
+                                if enum_type.values.contains_key(&value_pos.value_name) { "yes" } else { "no" }
+                            })
+                        },
+                    );
+                    value_pos.remove(&mut self.merged)?;
+                }
+                _ => todo!(),
```
sachindshinde: You can `bail!()` in this case, since the `if` clause above should make other cases impossible.

apollo-federation/src/merger/merger.rs
```diff
@@ -199,13 +250,617 @@ impl Merger {
...
+            }
+        } else if matches!(usage, EnumTypeUsage::Output { .. })
+            || matches!(usage, EnumTypeUsage::Unused)
+        {
```
sachindshinde: Similar to above, this can be
```suggestion
        } else if matches!(usage, EnumTypeUsage::Output { .. } | EnumTypeUsage::Unused) {
```

apollo-federation/src/merger/merger.rs
TylerBloom: You can instead use if-guards on the other branches for this branch to just do nothing.
(Of course, this would change the look of the code, but, IMO, if we can remove an unnecessary error, we should)
This would look something like this:
```rust
let branch = !is_inaccessible && sources.values().any(...);
match usage {
  EnumTypeUsage::Both if branch => { ... },
  EnumTypeUsage::Input if branch => { ... },
  _ => { }
}
```

apollo-federation/src/merger/merger.rs
```diff
@@ -199,13 +250,617 @@ impl Merger {
...
+            })
+        {
+            // We have a source (subgraph) that _has_ the enum type but not that particular enum value. If we're in the "both input and output usages",
+            // that's where we have to fail. But if we're in the "only input" case, we simply don't merge that particular value and hint about it.
+            match usage {
+                EnumTypeUsage::Both {
+                    input_example,
+                    output_example,
+                } => {
+                    self.report_mismatch_error_with_specifics(
```
sachindshinde: Note that this function takes more arguments in JS, which you've shifted into `report_mismatch_error_with_specifics()`. This might be fine for now, but it'll become an issue later when another caller wants to call `report_mismatch_error_with_specifics()` with different arguments (the other such caller is the JS method `Merger.validateExternalFields()`).

clenfest: I added a couple of TODOs in that code and I'll make a ticket to come back to it.

tninesling (approved):

sachindshinde (approved):

PR #7588, opened 2025-05-29
Standardize language within changesets

carodewig:
General updates:
* Headers start with a verb
* Add context to descriptions
* Capitalize `Router`


lrlna:
@carodewig this should go into `prep-2.3.0` branch somehow. That branch should have been opened last week so you could make these edits there directly. I am sorry that stepped was missed and you had to do this manually like this.

PR #7589, opened 2025-05-29
feat(federation/composition): port `advancePathWithTransition()`

sachindshinde:
This PR ports `advancePathWithTransition()` from the JS codebase, along with any currently un-ported  downstream functions.


duckki (approved):
Looks great! I had a question on memory management (`Arc<LazyLock<...>>`), which isn't necessary, yet. But, Sachin said it will be necessary for the next task.

PR #7590, opened 2025-05-29
feat(federation/composition):  SingleCompositionError 

TylerBloom:
This PR splits out the `SubgraphError` variant of `SingleFederationError` into a new enum, `SingleCompositionError`. This includes a conversion from `SubgraphError` to `SingleCompositionError` but *not* from `FederationError`/`SingleFederationError`. This is to ensure all federation errors that are intended to be `SubgraphError`s do. Lastly, any function that was returning a vec of (single) federation errors is now returning a `Vec<SingleCompositionError>`

---

**Checklist**

- Tests added and passing[^4]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-federation/src/error/mod.rs
```diff
@@ -114,6 +115,48 @@ pub enum UnsupportedFeatureKind {
     Alias,
 }
 
+#[derive(Debug, Clone, thiserror::Error)]
+pub enum SingleCompositionError {
```
dariuszkuc, line 119: I was under impression that we don't want to preserve the existing `Single`/`Multi`/`Aggregate` federation error enum. I thought we just want a `CompositionError` and return a vec if we need multiple errors?

TylerBloom: You are correct. That is our long term goal, but removing the multiple fed error variants is a lot of work. At one point in `build_query_plan`, we could get multiple errors (it's very early on). Fixing that is outside the scope of this ticket, so removing the two multiple fed errors variants is too.

We are using `MultipleFederationErrors` in composition code. I can replace those with a Vec of composition errors, but those are currently being converted into a SubgraphError and then into a composition error. This largely happens in the upgrade and validation code.

I think the final goal is to identify the subset of errors that might get turned into SubgraphError and use that error type in the locations I mentioned above. Then, merging and satisfiability can add to `SingleCompositionError` and the errors that need to be SubgraphErrors can be. I think this should be a follow up PR, though.

dariuszkuc: Should we just start with `CompositionError` and tackle the `MultipleFederationErrors` later on?

TylerBloom: We should, and that's my goal here. What I don't think we can do (yet) is have upgrade and validate use composition errors inside of `MultipleFederationErrors`. Otherwise, you end up with the same recursive error problem with SubgraphError I got rid of here.

clenfest:
Probably should make changes to the `src/merger/*` code as well.

apollo-federation/src/schema/validators/merged.rs
```diff
@@ -361,7 +361,6 @@ fn add_requires_error(
             application: requires_application.to_string(),
             message,
         }
-        .add_subgraph(subgraph_name.to_owned())
```
sachindshinde, line 364: We should create a ticket to add this subgraph info back in later (I'm guessing it'll happen when the `errors` argument changes to a vector of composition errors).

duckki:
There is a lint error.

apollo-federation/src/schema/validators/merged.rs
TylerBloom: I left two notes (one on the function above and one here) stating that both functions will likely need to change their signatures once actually used.

PR #7591, opened 2025-05-29
Connectors Debugger V2 to pass through all types of mapping problems

andrewmcgivery:
Connectors Debugger V2 to pass through all types of mapping problems

Open questions....
1. Do I actually need all the `clone` and `cloned` that I currently have? Rust seems to think so but open to suggestions
2. What is the impact of some of this code at run time when debug is _not_ enabled? Is it worth passing `debug` all over to avoid some of the mapping of errors to problems? (thinking no since it's pretty trivial code but open to opinions)

Relies on https://github.com/mdg-private/studio-ui/pull/12112 (that needs to be in prod before this change lands in a Router release!)

---

**Checklist**

- Tests added and passing[^4]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/plugins/connectors/plugin/debug.rs
+++ b/apollo-router/src/plugins/connectors/plugin/debug.rs
@@ -37,12 +37,12 @@ impl ConnectorContext {
         if let Some(request) = request {
             self.items.push(ConnectorContextItem {
                 request: *request,
-                response: ConnectorDebugHttpResponse::from((
+                response: ConnectorDebugHttpResponse::new(
                     parts,
                     json_body,
                     selection_data,
                     error_settings,
-                )),
+                ),
                 problems,
             });
         } else {
@@ -191,23 +191,13 @@ pub(crate) struct ConnectorDebugHttpRequest {
     connect_headers: Option<Vec<(String, String)>>,
 }
 
-impl
-    From<(
-        &http::Request<String>,
-        String,
-        Option<&serde_json_bytes::Value>,
-        Option<SelectionData>,
-        &HttpJsonTransport,
-    )> for ConnectorDebugHttpRequest
-{
-    fn from(
-        (req, kind, json_body, selection_data, transport): (
-            &http::Request<String>,
-            String,
-            Option<&serde_json_bytes::Value>,
-            Option<SelectionData>,
-            &HttpJsonTransport,
-        ),
+impl ConnectorDebugHttpRequest {
+    pub(crate) fn new(
+        req: &http::Request<String>,
+        kind: String,
+        json_body: Option<&serde_json_bytes::Value>,
+        selection_data: Option<SelectionData>,
+        transport: &HttpJsonTransport,
     ) -> Self {
         let headers = transport.headers.iter().fold(
             HashMap::new(),
@@ -277,21 +267,12 @@ struct ConnectorDebugHttpResponse {
     errors: Option<ConnectorDebugErrors>,
 }
 
-impl
-    From<(
-        &http::response::Parts,
-        &serde_json_bytes::Value,
-        Option<SelectionData>,
-        &ConnectorErrorsSettings,
-    )> for ConnectorDebugHttpResponse
-{
-    fn from(
-        (parts, json_body, selection_data, error_settings): (
-            &http::response::Parts,
-            &serde_json_bytes::Value,
-            Option<SelectionData>,
-            &ConnectorErrorsSettings,
-        ),
+impl ConnectorDebugHttpResponse {
+    pub(crate) fn new(
+        parts: &http::response::Parts,
+        json_body: &serde_json_bytes::Value,
+        selection_data: Option<SelectionData>,
+        error_settings: &ConnectorErrorsSettings,
     ) -> Self {
         ConnectorDebugHttpResponse {
             status: parts.status.as_u16(),
--- a/apollo-router/src/plugins/connectors/http_json_transport.rs
+++ b/apollo-router/src/plugins/connectors/http_json_transport.rs
@@ -105,7 +105,7 @@ pub(crate) fn make_request(
 
     let debug_request = debug.as_ref().map(|_| {
         if is_form_urlencoded {
-            Box::new(ConnectorDebugHttpRequest::from((
+            Box::new(ConnectorDebugHttpRequest::new(
                 &request,
                 "form-urlencoded".to_string(),
                 form_body
@@ -117,9 +117,9 @@ pub(crate) fn make_request(
                     result: json_body,
                 }),
                 transport,
-            )))
+            ))
         } else {
-            Box::new(ConnectorDebugHttpRequest::from((
+            Box::new(ConnectorDebugHttpRequest::new(
                 &request,
                 "json".to_string(),
                 json_body.as_ref(),
@@ -129,7 +129,7 @@ pub(crate) fn make_request(
                     result: json_body.clone(),
                 }),
                 transport,
-            )))
+            ))
         }
     });
 


apollo-federation/src/connectors/models/problem_location.rs
```diff
@@ -0,0 +1,19 @@
...
+    SourceUri,
+    SourcePath,
+    SourceQueryParams,
+    ConnectUri,
+    ConnectPath,
+    ConnectQueryParams,
+    SourceHeaders,
+    ConnectHeaders,
+    Selection,
+    ErrorsMessage,
```
dylan-apollo, line 16: Do we not need to differentiate between source & connect message? Might help users know which one triggered and went wrong (if they're not realizing it's being overridden)

apollo-federation/src/connectors/models/problem_location.rs
```diff
@@ -0,0 +1,19 @@
+use serde::Deserialize;
+use serde::Serialize;
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Hash, PartialOrd, Ord)]
```
dylan-apollo:
```suggestion
#[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq, Eq, Hash, PartialOrd, Ord)]
```
`Copy` should be used for tiny types (like this one, which will be like 1 byte) because it's actually more efficient to copy the value around than to take pointers to it.

apollo-federation/src/connectors/models/http_json_transport.rs
```diff
@@ -353,6 +385,12 @@ impl HeaderSource {
     }
 }
 
+#[derive(Clone, Debug, Hash, Eq, PartialEq)]
```
dylan-apollo:
```suggestion
#[derive(Clone, Copy, Debug, Hash, Eq, PartialEq)]
```

apollo-router/src/plugins/connectors/plugin/debug.rs
```diff
@@ -108,50 +156,117 @@ struct ConnectorDebugBody {
...
+#[serde(rename_all = "camelCase")]
 pub(crate) struct ConnectorDebugHttpRequest {
     url: String,
     method: String,
     headers: Vec<(String, String)>,
     body: Option<ConnectorDebugBody>,
+    source_uri: Option<ConnectorDebugUri>,
+    connect_uri: ConnectorDebugUri,
+    source_headers: Option<Vec<(String, String)>>,
+    connect_headers: Option<Vec<(String, String)>>,
```
dylan-apollo, line 191: So this is just repeating `headers`, but splitting them by source? Should we instead make `headers` a list of `Header` where each has a `name`, `value`, and `source`?

What about headers that come from router config, are we tracking those separately as well?

apollo-router/src/plugins/connectors/plugin/debug.rs
```diff
@@ -108,50 +156,117 @@ struct ConnectorDebugBody {
...
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
 pub(crate) struct ConnectorDebugHttpRequest {
     url: String,
     method: String,
     headers: Vec<(String, String)>,
     body: Option<ConnectorDebugBody>,
+    source_uri: Option<ConnectorDebugUri>,
+    connect_uri: ConnectorDebugUri,
```
dylan-apollo: We should probably use `url` everywhere... it's my fault that we have a mixture, but no reason to propagate that

apollo-router/src/plugins/connectors/plugin/debug.rs
```diff
@@ -108,50 +156,117 @@ struct ConnectorDebugBody {
...
-    errors: Vec<Problem>,
-}
+impl
+    From<(
+        &http::Request<String>,
+        String,
+        Option<&serde_json_bytes::Value>,
+        Option<SelectionData>,
+        &HttpJsonTransport,
+    )> for ConnectorDebugHttpRequest
```
dylan-apollo: I think big tuples are so ugly 😓. I'd stick to an old fashion method instead of `From`

apollo-router/src/plugins/connectors/plugin/debug.rs
```diff
@@ -7,8 +14,14 @@ use crate::plugins::connectors::mapping::Problem;
...
 #[derive(Debug, Clone, Serialize, Deserialize, Default)]
 pub(crate) struct ConnectorContext {
-    requests: Vec<ConnectorDebugHttpRequest>,
-    responses: Vec<ConnectorDebugHttpResponse>,
+    items: Vec<ConnectorContextItem>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub(crate) struct ConnectorContextItem {
+    problems: Vec<(ProblemLocation, Problem)>,
```
dylan-apollo, line 22: Can we put the location in the `Problem` instead of a vec of tuple?

apollo-federation/src/connectors/models/http_json_transport.rs
```diff
@@ -45,7 +46,7 @@ pub struct HttpJsonTransport {
     pub source_url: Option<Uri>,
     pub connect_template: StringTemplate,
     pub method: HTTPMethod,
-    pub headers: IndexMap<HeaderName, HeaderSource>,
+    pub headers: IndexMap<HeaderName, (HeaderSource, OriginatingDirective)>,
```
dylan-apollo: I don't like this tuple either, but adding the directive the the `HeaderSource` looks like it'd be a pain so... 🤷 can't win them all

apollo-federation/src/connectors/models/http_json_transport.rs
```diff
@@ -125,12 +138,21 @@ impl HttpJsonTransport {
             })
     }
 
-    pub fn make_uri(&self, inputs: &IndexMap<String, Value>) -> Result<Uri, MakeUriError> {
+    pub fn make_uri(
+        &self,
+        inputs: &IndexMap<String, Value>,
+    ) -> Result<(Uri, Vec<(ProblemLocation, ApplyToError)>), MakeUriError> {
```
dylan-apollo, line 121: Thinking we should define `Problem` here (instead of up in router) to clean up this type a bit? Not sure how easy that'd be with how aggregation works...

apollo-federation/src/connectors/models/http_json_transport.rs
```diff
@@ -125,12 +138,21 @@ impl HttpJsonTransport {
...
-        // TODO: Return these warnings for both Sandbox debugging and mapping playground
         let mut warnings = Vec::new();
 
-        let connect_uri = self.connect_template.interpolate_uri(inputs)?;
+        let (connect_uri, connect_template_warnings) =
+            self.connect_template.interpolate_uri(inputs)?;
+        warnings.extend(
+            connect_template_warnings
+                .iter()
+                .cloned()
```
dylan-apollo:
```suggestion
                .into_iter()
```

apollo-federation/src/connectors/models/http_json_transport.rs
```diff
@@ -146,7 +168,12 @@ impl HttpJsonTransport {
             path.write_without_encoding(source_uri_path)?;
         }
         if let Some(source_path) = self.source_path.as_ref() {
-            warnings.extend(extend_path_from_expression(&mut path, source_path, inputs)?);
+            warnings.extend(
+                extend_path_from_expression(&mut path, source_path, inputs)?
+                    .iter()
+                    .cloned()
```
dylan-apollo:
```suggestion
                    .into_iter()
```

apollo-federation/src/connectors/models/http_json_transport.rs
```diff
@@ -160,11 +187,12 @@ impl HttpJsonTransport {
...
         if let Some(connect_path) = self.connect_path.as_ref() {
-            warnings.extend(extend_path_from_expression(
-                &mut path,
-                connect_path,
-                inputs,
-            )?);
+            warnings.extend(
+                extend_path_from_expression(&mut path, connect_path, inputs)?
+                    .iter()
+                    .cloned()
```
dylan-apollo:
```suggestion
                    .into_iter()
```
Avoids some of the clones you were worried about 😁

apollo-federation/src/connectors/models/http_json_transport.rs
```diff
@@ -177,11 +205,12 @@ impl HttpJsonTransport {
...
         if let Some(source_query) = self.source_query_params.as_ref() {
-            warnings.extend(extend_query_from_expression(
-                &mut query,
-                source_query,
-                inputs,
-            )?);
+            warnings.extend(
+                extend_query_from_expression(&mut query, source_query, inputs)?
+                    .iter()
+                    .cloned()
```
dylan-apollo:
```suggestion
                   .into_iter()
```

apollo-federation/src/connectors/models/http_json_transport.rs
```diff
@@ -191,11 +220,12 @@ impl HttpJsonTransport {
...
         if let Some(connect_query) = self.connect_query_params.as_ref() {
-            warnings.extend(extend_query_from_expression(
-                &mut query,
-                connect_query,
-                inputs,
-            )?);
+            warnings.extend(
+                extend_query_from_expression(&mut query, connect_query, inputs)?
+                    .iter()
+                    .cloned()
```
dylan-apollo:
```suggestion
                    .into_iter()
```

dylan-apollo (commented):
First batch of comments, will review more after my next meeting ☺️

apollo-router/src/plugins/connectors/plugin/debug.rs
andrewmcgivery: One of my goals was to not introduce any breaking changes as part of this, so everything was additive.

But also because `headers` is the final calculated values whereas `source_headers` and `connect_headers` are the raw mappings... so not quite the same thing. 😅

apollo-federation/src/connectors/models/http_json_transport.rs
andrewmcgivery: Are you thinking move both `Problem` and it's mapping function `aggregate_apply_to_errors` to `apollo-federation`?

apollo-federation/src/connectors/models/problem_location.rs
andrewmcgivery: This is a weird one. I had thought about this too but I'd have to go back and re-do a bunch of code earlier in the lifecycle that decides which one to use and then move that decision later in the code 😅 

It didn't feel worth doing at the time but maybe it is?

apollo-router/src/plugins/connectors/plugin/debug.rs
andrewmcgivery: I can change this. This was something I started doing after @naomijub mentioned on another PR that we should use `From` instead of "from_x" methods. So I'd be curious on her thoughts here before I change this!

apollo-router/src/plugins/connectors/plugin/debug.rs
andrewmcgivery: Maybe? But it would be a pretty big refactor cause we have Problem all over the place. Not against it, but want to make sure it's worth doing before I undertake it lol

apollo-federation/src/connectors/models/http_json_transport.rs
andrewmcgivery: Yea, this one I think is probably not a whole lot I can do about it 😓

apollo-federation/src/connectors/models/http_json_transport.rs
andrewmcgivery: omg ty

apollo-router/src/plugins/connectors/plugin/debug.rs
andrewmcgivery: Done!

apollo-federation/src/connectors/models/http_json_transport.rs
lennyburdette: I'll be moving `Problem` into a-f after this lands, so we can resolve this ask

apollo-federation/src/connectors/models/problem_location.rs
lennyburdette: Does it change the API between the router and sandbox? If so, let's discuss this now. If not, we can change this later.

apollo-router/src/plugins/connectors/plugin/debug.rs
lennyburdette: please make a ticket tracking this ask

apollo-router/src/plugins/connectors/plugin/debug.rs
lennyburdette: 👍🏻 to more `From`s, but also 👍🏻 to not using `From` with huge tuples here

apollo-federation/src/connectors/models/problem_location.rs
andrewmcgivery: These enums are used as identifiers in the sandbox to categorize the problems. So if we wanted to show it differently in the sandbox for `source` vs `connect` then it would make sense to split it apart in the payload.

I'm just a bit on the fence if it's worth differentiating.

apollo-router/src/plugins/connectors/plugin/debug.rs
andrewmcgivery: Done: https://apollographql.atlassian.net/browse/CNN-874

apollo-federation/src/connectors/models/http_json_transport.rs
andrewmcgivery: This is actually gone now :)

apollo-router/src/plugins/connectors/plugin/debug.rs
andrewmcgivery: Done!

PR #7596, opened 2025-05-31
Expose endpoint to get string representing federation spec definition at version number

conwuegb:
This change 
- exposes `get_federation_spec_definition_string_from_version(...)` to allow clients to get a `String` representation of `FederationSpecDefinition` at a specified version number
- updates `schema_as_fed2_subgraph(...)` to take a version number, in order to expand the schema accordingly per version
- updates `validate_federation_imports(...)` to dynamically generate imports instead of relying on hard-coded arrays, which were out of date
- makes a small fix to `@fromContext` directive's argument spec (arg name was incorrect, and arg was marked as non-nullable)

> [!NOTE]
>
> the `@link` directive is not included in this change, given federation's current implementation so the following substring will not appear in the output of `get_federation_spec_definition_string_from_version(...)`:
> ```graphql
> directive @link(url: String!, as: String, import: [Import], for: Purpose) repeatable on SCHEMA
> scalar Import
> enum Purpose {
>     SECURITY
>     EXECUTION
> }
> ```

---

**Checklist**

- Tests added and passing[^4]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

[^1]: It may be appropriate to bring upcoming changes to the attention of other (impacted) groups. Please endeavor to do this before seeking PR approval. The mechanism for doing this will vary considerably, so use your judgement as to how and when to do this.


apollo-federation/src/subgraph/typestate.rs
```diff
@@ -528,10 +528,16 @@ fn schema_as_fed2_subgraph(
     // we want to maintain backward compatibility for those who have already upgraded and we had
     // been upgrading the url to latest, but we never automatically import directives that exist
     // past 2.4
-    let imports: Vec<_> = FederationSpecDefinition::auto_expanded_federation_spec()
+    let imports: Vec<_> = fed_spec
```
duckki, line 549: It's potentially a nice idea. But, I'm afraid we are deviating from JS code too much. For example, `new_empty_federation_2_subgraph_schema` will be broken after this change, since that function get to use the latest federation spec, but only imports 2.4 directives.

apollo-federation/src/subgraph/typestate.rs
```diff
@@ -483,7 +483,7 @@ pub(crate) fn add_fed1_link_to_schema(
 //            the input and returns the updated inner Schema.
 fn schema_as_fed2_subgraph(
     mut schema: FederationSchema,
-    use_latest: bool,
+    version: &Version,
```
duckki, line 504: Why don't we leave `schema_as_fed2_subgraph`'s interface as it was before to be compatible with JS composition? We can factor out the functionality (which can set the federation version and import version separately) and use it to implement a new function that supports arbitrary versions.

apollo-federation/src/lib.rs
```diff
@@ -56,9 +56,11 @@ use crate::error::FederationError;
 use crate::error::SingleFederationError;
 use crate::link::context_spec_definition::CONTEXT_VERSIONS;
 use crate::link::context_spec_definition::ContextSpecDefinition;
+pub use crate::link::federation_spec_definition::get_federation_spec_definition_string_from_version;
```
duckki: We may want to make it clear that those pub functions are our internal API.
`apollo-router` crate has a `pub` but private API in the `_private` module: https://github.com/apollographql/router/blob/c42fffb1447d177bc03a33b8117957a2e54c3b26/apollo-router/src/lib.rs#L113

If we do this, I'd like to give the mod a proper API name, instead of `_private`. Something like `internal_lsp_api`.

apollo-federation/src/link/federation_spec_definition.rs
```diff
@@ -114,6 +115,35 @@ impl FederationSpecDefinition {
...
+        let directive_definitions = self.directive_specs();
+        for dir_def in directive_definitions {
+            if let Some(dir_def) = subgraph
+                .schema()
+                .schema()
+                .directive_definitions
+                .get(dir_def.name())
+            {
+                result.push_str(&dir_def.to_string());
+                result.push('\n');
```
TylerBloom, line 136: If you import `std::fmt::Write`, this can be simplified and you don't need to allocate a temporary string.
```suggestion
                _ = writeln!(result, "{dir_def}")
```
Note: `writeln!` returns a `Result`, but write a string should never fail (barring extreme situations), so we can just ignore it.

apollo-federation/src/link/federation_spec_definition.rs
```diff
@@ -114,6 +115,35 @@ impl FederationSpecDefinition {
...
+            }
+        }
+
+        let type_definitions = self.type_specs();
+        for type_def in type_definitions {
+            if let Some(type_def) = subgraph.schema().schema().types.get(type_def.name()) {
+                result.push_str(&type_def.to_string());
```
TylerBloom, line 143: I similar `write!` change can be made here.

apollo-federation/src/link/federation_spec_definition.rs
```diff
@@ -1065,3 +1088,249 @@ pub(crate) fn fed1_link_imports() -> Vec<Arc<link::Import>> {
...
 }
+
+pub fn get_federation_spec_definition_string_from_version(
+    version: Version,
+) -> Result<String, FederationError> {
+    let fed_spec_definition = FEDERATION_VERSIONS.find(&version).ok_or_else(|| {
+        SingleFederationError::UnknownFederationLinkVersion {
+            message: format!("Unknown federation spec version: {}", version),
+        }
+    })?;
```
TylerBloom, line 1109: You can use `let-else` here:
```suggestion
    let fed_spec_definition = FEDERATION_VERSIONS.find(&version) else {
        return SingleFederationError::UnknownFederationLinkVersion {
            message: format!("Unknown federation spec version: {}", version),
        }.into()
    };
```
In most of the codebase, this pattern of `Option -> Result` via `ok_or_else + ?` was replaced with let-else (and `bail!` where useable).

conwuegb: image

I'm good with changing this to match the codebase's pattern, but I do get an error about the else clause being superfluous. The suggested fix for that brings me back to `ok_or_else()`.

For context, this was the full change I made:
```suggestion
    let fed_spec_definition = FEDERATION_VERSIONS.find(&version) else {
        return Err(SingleFederationError::UnknownFederationLinkVersion {
            message: format!("Unknown federation spec version: {}", version),
        }.into());
    };

    fed_spec_definition.expect("Federation spec definition should exist after version check").federation_spec_to_string()
```

apollo-federation/src/subgraph/typestate.rs
conwuegb: Are you saying that `new_empty_federation_2_subgraph_schema` *is correct* in using the latest federation spec while only importing directives up to 2.4? And that this change would end up incorrectly importing all directives from the latest version? If so, why were we using the latest specs but not importing all of its directives?

apollo-federation/src/subgraph/typestate.rs
conwuegb: I'm open to this. Thoughts, @tninesling?

apollo-federation/src/subgraph/typestate.rs
duckki: I don't know why it's the way it is. But, for the initial Rust port, we are porting and preserving [the JS behavior](https://github.com/apollographql/federation/blob/02ebb9ed09f157c4bcc00dbe27de6b57422b98cc/internals-js/src/federation.ts#L1917).

conwuegb: @tninesling, did you have an opinion about diverging from the JS behavior here?

apollo-federation/src/link/federation_spec_definition.rs
duckki: The pattern was missing in the original suggestion. Tyler meant `let Some(fed_spec_definition) = ...`.

apollo-federation/src/link/federation_spec_definition.rs
```diff
@@ -114,6 +115,35 @@ impl FederationSpecDefinition {
         }
     }
 
+    pub(crate) fn federation_spec_to_string(&self) -> Result<String, FederationError> {
```
duckki, line 118: In this method, several calls returns Result with SubgraphError. I'd propagate SubgraphError here, so we don't need to convert it to FederationError in multiple places.

Then, we can add a `.map_err(|e| e.into_inner())` at the end of `get_federation_spec_definition_string_from_version` function once.

apollo-federation/src/link/federation_spec_definition.rs
```diff
@@ -114,6 +115,35 @@ impl FederationSpecDefinition {
...
+        // Parse a simple schema to get an expanded subgraph
+        let subgraph = Subgraph::parse("S", "http://S", "type Query { hello: String }")?
+            .into_fed2_subgraph(self.version())?
```
duckki, line 122: According to your document, we will be provided a schema document.
I think we can call `expand_subgraphs` and `upgrade_subgraphs_if_necessary` (following the usual composition steps in `composition::compose` function).

duckki: So, we don't need to change `into_fed2_subgraph`, I think.

apollo-federation/src/subgraph/typestate.rs
```diff
@@ -1025,6 +1068,183 @@ mod tests {
         );
     }
 
+    #[test]
+    fn returns_correct_schema_diff_for_fed_1_0() {
+        let schema_string = String::from_str(
```
duckki, line 1073: Instead of `String::from_str(r#"xxx"#).unwrap()`, you can just do `r#xxx"#.to_string()`.

apollo-federation/src/lib.rs
duckki: As we discussed off-line, the whole apollo_federation is a more or less private library. So, I don't think we need to make a new private module to hold this interface.

apollo-federation/src/lib.rs
```diff
@@ -83,6 +82,10 @@ use crate::subgraph::ValidSubgraph;
 pub use crate::supergraph::ValidFederationSubgraph;
 pub use crate::supergraph::ValidFederationSubgraphs;
 
+pub mod internal_lsp_api {
+    pub use crate::subgraph::typestate::schema_diff_expanded_from_initial;
```
duckki, line 86: I know I went back and forth. I like this approach, too.

apollo-federation/src/link/federation_spec_definition.rs
duckki: I was confused. There is `schema_diff_expanded_from_initial` function that does it. Please ignore.

PR #7600, opened 2025-06-02
Fill out most Merger constructor logic (minus prepare_supergraph)

tninesling:
# Overview

Ports bulk of `Merger` constructor, primarily consisting of logic for implicit version upgrades based on imported specs. The JS source is located [here](https://github.com/apollographql/federation/blob/main/composition-js/src/merging/merge.ts#L384-L470).

There are some slight differences in the implementation where we'll need to pass arguments into the helper functions where JS references `this`. In JS the various `this` bindings can occur prior to the constructor logic, but we don't have access to the equivalent `self` yet in Rust.

## Small refactorings to make this cleaner

1. I moved the logic to enumerate all known linked features from `FederationBlueprint::expand_known_features` to `Subgraph::all_features` for reuse. This is now also used in the implicit version upgrade functionality.
2. Introduced `DirectiveReferencers::extend` so we can roll up referencers from various subgraphs. This means `fields_with_from_context` and `fields_with_override` are now `DirectiveReferencers` instead of the JS `Set<string>`, which I think better follows our Rust idioms for checking if a particular coordinate has a directive applied.

Note: there are still a couple `todo!` calls that prevent testing this constructor. We'll implement those in upcoming PRs.

---

**Checklist**

- Tests added and passing[^4]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-federation/src/schema/mod.rs
```diff
@@ -703,6 +711,117 @@ impl FederationSchema {
...
+        version: &impl std::fmt::Display,
+        supported_versions: impl Iterator<Item = impl std::fmt::Display>,
+    ) -> SingleFederationError {
+        SingleFederationError::UnknownLinkVersion {
+            message: format!(
+                "Detected unsupported {spec_name} specification version {version}. Please upgrade to a composition version which supports that version, or select one of the following supported versions: {}.",
+                supported_versions
+                    .map(|v| v.to_string())
+                    .collect::<Vec<_>>()
+                    .join(", ")
```
TylerBloom: A general note: If `Itertools` is imported, you don't need to map into `String`s nor collect into a `Vec` as `Itertools` provides [a `join` method](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.join) that does this. Though, this is in an error path, so it won't really affect performance

tninesling: We aren't yet importing `Itertools` in this file, which is the main reason I haven't used it. I've used it occasionally for convenience, but admittedly I haven't looked into potential performance implications until now. I'm on board with preferring iterators over loops where possible, and `Itertools` looks to empower that to a greater extent.

Is there a way we could encode that preference in a linter rule? I have a strong preference that style choices be encoded in the linter so code reviews can largely focus on correctness and overall architecture.

I'll update this occurrence in the meantime, but there are several other locations where we just use the `collect` + `join` from the standard library.

apollo-federation/src/link/spec_definition.rs
```diff
@@ -204,17 +204,31 @@ impl<T: SpecDefinition> SpecDefinitions<T> {
...
         &'static self,
         federation_version: &Version,
-    ) -> Option<&'static dyn SpecDefinition> {
+    ) -> Option<&'static T> {
         self.definitions
             .values()
             .find(|spec| federation_version.satisfies(spec.minimum_federation_version()))
+    }
+
+    pub(crate) fn get_dyn_minimum_required_version(
```
clenfest, line 227: This is probably just a case where I need to be educated, but why is the dynamic dispatch required here? I can't see why it's needed.

clenfest:
I can't comment on the individual files, but I assume it was a mistake to include `schema/validators/tag.rs` and `tag_validation_template.graphqls` as they're both empty.

apollo-federation/src/schema/referencer.rs
```diff
@@ -198,4 +198,31 @@ impl DirectiveReferencers {
                     .map(|pos| ObjectOrInterfaceFieldDefinitionPosition::Interface(pos.clone())),
             )
     }
+
+    pub(crate) fn extend(&mut self, other: &Self) {
+        if let Some(schema) = other.schema.clone() {
```
clenfest: I'm a little worried about the `schema` variable here as it doesn't quite fit in with the rest. If you extend in a loop as you do when you call this, we'll be overwriting this value silently every time, no?

apollo-federation/src/merger/merger.rs
```diff
@@ -84,29 +95,141 @@ impl Merger {
...
+            .unwrap_or_else(|| FEDERATION_VERSIONS.latest().version())
+    }
+
+    fn get_latest_federation_version_used_in_subgraph<'a>(
+        subgraph: &'a Subgraph<Validated>,
+        error_reporter: &mut ErrorReporter,
+    ) -> &'a Version {
+        let linked_federation_version = subgraph.metadata().federation_spec_definition().version();
+
+        let linked_features = subgraph.schema().all_features().unwrap_or_default();
```
clenfest, line 161: What is the default here?

clenfest (approved):
A couple of questions but otherwise LGTM!

apollo-federation/src/link/spec_definition.rs
tninesling: This is used when defining directive specifications, which need a function which can get their relevant spec versions based on a federation version. The way we've defined that is `Fn(&Version) -> Option<&dyn SpecDefinition>`. Unfortunately, Rust doesn't recognize an `Option<&JoinSpecDefinition>` as an `Option<&dyn SpecDefinition>`, so I added this utility to automatically cast to the trait.

apollo-federation/src/merger/merger.rs
tninesling: The default for `Vec` is just an empty vec

apollo-federation/src/schema/referencer.rs
tninesling: This overwrites, but when we start with the default value, it will be set to `None`. So after we loop over all the subgraphs, this effectively acts as a flag indicating that at least one subgraph referenced the directive at the schema root if this is `Some`. The `SchemaDefinitionPosition` doesn't carry any data, it just represents the root of the schema, so there isn't anything to overwrite.

This won't be used for now, since this implementation is to support fields referencing `context` and `fromContext`, but I do think it's the most valid implementation for rolling up that particular field in the struct.

PR #7604, opened 2025-06-03
Cooperative Cancellation

Velfi:
*Description here*

Fixes #**issue_number**

---

**Checklist**

- Tests added and passing[^4]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/configuration/mod.rs
+++ b/apollo-router/src/configuration/mod.rs
@@ -900,6 +900,41 @@ pub(crate) struct QueryPlanning {
     /// If cache warm up is configured, this will allow the router to keep a query plan created with
     /// the old schema, if it determines that the schema update does not affect the corresponding query
     pub(crate) experimental_reuse_query_plans: bool,
+
+    /// A timeout for query planning. This option requires that `experimental_enable_cooperative_cancellation` is set to true.
+    /// Otherwise, it will be ignored.
+    pub(crate) experimental_timeout_in_seconds: Option<f64>,
+
+    /// Enable cooperative cancellation of query planning
+    ///
+    /// When enabled, query planning will be cancelled if the client waiting on the QP closes their connection.
+    pub(crate) experimental_enable_cooperative_cancellation: bool,
+}
+
+#[buildstructor::buildstructor]
+impl QueryPlanning {
+    #[builder]
+    #[allow(dead_code)]
+    pub(crate) fn new(
+        cache: Option<QueryPlanCache>,
+        warmed_up_queries: Option<usize>,
+        experimental_plans_limit: Option<u32>,
+        experimental_paths_limit: Option<u32>,
+        experimental_reuse_query_plans: Option<bool>,
+        experimental_timeout_in_seconds: Option<f64>,
+        experimental_enable_cooperative_cancellation: Option<bool>,
+    ) -> Self {
+        Self {
+            cache: cache.unwrap_or_default(),
+            warmed_up_queries,
+            experimental_plans_limit,
+            experimental_paths_limit,
+            experimental_reuse_query_plans: experimental_reuse_query_plans.unwrap_or_default(),
+            experimental_timeout_in_seconds,
+            experimental_enable_cooperative_cancellation: experimental_enable_cooperative_cancellation
+                .unwrap_or_default(),
+        }
+    }
 }
 
 /// Cache configuration
--- a/apollo-router/src/query_planner/caching_query_planner.rs
+++ b/apollo-router/src/query_planner/caching_query_planner.rs
@@ -10,6 +10,7 @@ use rand::seq::SliceRandom;
 use rand::thread_rng;
 use sha2::Digest;
 use sha2::Sha256;
+use tokio_util::time::FutureExt;
 use tower::BoxError;
 use tower::ServiceBuilder;
 use tower::ServiceExt;
@@ -93,6 +94,8 @@ pub(crate) struct CachingQueryPlanner<T: Clone> {
     plugins: Arc<Plugins>,
     enable_authorization_directives: bool,
     config_mode_hash: Arc<ConfigModeHash>,
+    enable_cooperative_cancellation: bool,
+    cooperative_cancellation_timeout: Option<core::time::Duration>,
 }
 
 fn init_query_plan_from_redis(
@@ -140,6 +143,15 @@ where
         let mut hasher = StructHasher::new();
         configuration.rust_query_planner_config().hash(&mut hasher);
         let config_mode_hash = Arc::new(ConfigModeHash(hasher.finalize()));
+        let enable_cooperative_cancellation = configuration
+            .supergraph
+            .query_planning
+            .experimental_enable_cooperative_cancellation;
+        let cooperative_cancellation_timeout = configuration
+            .supergraph
+            .query_planning
+            .experimental_timeout_in_seconds
+            .map(|s| core::time::Duration::from_secs_f64(s));
 
         Ok(Self {
             cache,
@@ -148,6 +160,8 @@ where
             subgraph_schemas,
             plugins: Arc::new(plugins),
             enable_authorization_directives,
+            enable_cooperative_cancellation,
+            cooperative_cancellation_timeout,
             config_mode_hash,
         })
     }
@@ -388,15 +402,15 @@ impl CachingQueryPlanner<QueryPlannerService> {
     }
 }
 
-impl<T: Clone + Send + 'static> tower::Service<query_planner::CachingRequest>
+impl<T: Clone + Send + 'static> Service<query_planner::CachingRequest>
     for CachingQueryPlanner<T>
 where
-    T: tower::Service<
+    T: Service<
             QueryPlannerRequest,
             Response = QueryPlannerResponse,
             Error = MaybeBackPressureError<QueryPlannerError>,
         >,
-    <T as tower::Service<QueryPlannerRequest>>::Future: Send,
+    <T as Service<QueryPlannerRequest>>::Future: Send,
 {
     type Response = QueryPlannerResponse;
     type Error = CacheResolverError;
@@ -428,19 +442,19 @@ where
 
 impl<T> CachingQueryPlanner<T>
 where
-    T: tower::Service<
+    T: Service<
             QueryPlannerRequest,
             Response = QueryPlannerResponse,
             Error = MaybeBackPressureError<QueryPlannerError>,
         > + Clone
         + Send
         + 'static,
-    <T as tower::Service<QueryPlannerRequest>>::Future: Send,
+    <T as Service<QueryPlannerRequest>>::Future: Send,
 {
     async fn plan(
         mut self,
         request: query_planner::CachingRequest,
-    ) -> Result<<T as tower::Service<QueryPlannerRequest>>::Response, CacheResolverError> {
+    ) -> Result<<T as Service<QueryPlannerRequest>>::Response, CacheResolverError> {
         if self.enable_authorization_directives {
             AuthorizationPlugin::update_cache_key(&request.context);
         }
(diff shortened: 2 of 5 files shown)


apollo-router/src/configuration/mod.rs
```diff
@@ -900,6 +900,41 @@ pub(crate) struct QueryPlanning {
...
+            warmed_up_queries,
+            experimental_plans_limit,
+            experimental_paths_limit,
+            experimental_reuse_query_plans: experimental_reuse_query_plans.unwrap_or_default(),
+            experimental_timeout_in_seconds,
+            experimental_enable_cooperative_cancellation: experimental_enable_cooperative_cancellation
+                .unwrap_or_default(),
+        }
+    }
 }
```
Velfi: Added for use in tests.

apollo-router/src/configuration/mod.rs
```diff
@@ -900,6 +900,41 @@ pub(crate) struct QueryPlanning {
...
     pub(crate) experimental_reuse_query_plans: bool,
+
+    /// A timeout for query planning. This option requires that `experimental_enable_cooperative_cancellation` is set to true.
+    /// Otherwise, it will be ignored.
+    pub(crate) experimental_timeout_in_seconds: Option<f64>,
+
+    /// Enable cooperative cancellation of query planning
+    ///
+    /// When enabled, query planning will be cancelled if the client waiting on the QP closes their connection.
+    pub(crate) experimental_enable_cooperative_cancellation: bool,
```
Velfi: These two options should be an enum instead IMO. Something like this:

```rust
enum CooperativeCancellation {
    Disabled,
    Enabled,
    EnabledWithTimeoutInSeconds(f64)
}
```

apollo-router/src/query_planner/caching_query_planner.rs
```diff
@@ -508,96 +522,121 @@ where
...
+                    None => planning_task.await,
+                }
+            } else {
+                // some clients might timeout and cancel the request before query planning is finished,
+                // so we execute it in a task that can continue even after the request was canceled and
+                // the join handle was dropped. That way, the next similar query will use the cache instead
+                // of restarting the query planner until another timeout
+                tokio::task::spawn(planning_task).await
+            }
+            .map_err(convert_join_error)?
```
Velfi: This is the core of it.

apollo-router/src/query_planner/caching_query_planner.rs
```diff
@@ -166,9 +177,10 @@ where
         experimental_pql_prewarm: &PersistedQueriesPrewarmQueryPlanCache,
     ) {
         let _timer = Timer::new(|duration| {
-            f64_histogram!(
+            f64_histogram_with_unit!(
```
bnjjj: It will create a breaking change because it will change the name of this metric in prometheus :( It adds a suffix with the unit on the metric for prometheus

apollo-router/src/query_planner/caching_query_planner.rs
```diff
@@ -508,96 +519,168 @@ where
...
+                            // Abort is a no-op if the task has already completed.
+                            abort_handle.abort();
+                        } else if self.cooperative_cancellation.is_measure_mode() {
+                            // This is always going to get dropped so we have to make sure we're
+                            // only measuring when the task would be cancelled.
+                            if !abort_handle.is_finished() {
+                                u64_counter_with_unit!(
+                                    "apollo.router.operations.cooperative_cancellations",
+                                    "The number of times cooperative cancellation was triggered but not enforced",
+                                    "cancellations",
```
bnjjj: https://opentelemetry.io/docs/specs/semconv/general/metrics/#instrument-units make sure it follows these conventions

apollo-router/src/query_planner/caching_query_planner.rs
```diff
@@ -508,96 +519,168 @@ where
...
+                            // When in measure mode, we want to spawn a timeout task, but we don't
+                            // want to race it against the planning task. If it happens to finish
+                            // first, we emit a log. Otherwise, the scopeguard will ensure that it
+                            // gets cancelled and no log will be emitted.
+                            let timeout_task = tokio::task::spawn(async move {
+                                tokio::time::sleep(timeout).await;
+                                u64_counter_with_unit!(
+                                    "apollo.router.operations.cooperative_cancellations",
+                                    "The number of times cooperative cancellation was triggered but not enforced",
+                                    "cancellations",
```
bnjjj: https://opentelemetry.io/docs/specs/semconv/general/metrics/#instrument-units

apollo-router/src/query_planner/caching_query_planner.rs
```diff
@@ -507,96 +526,207 @@ where
...
+                                Outcome::Cancelled as u8,
+                                Ordering::SeqCst,
+                                Ordering::SeqCst,
+                            )
+                            .is_ok()
+                        {
+                            if enforce_mode {
+                                abort_handle.abort();
+                            }
+                            tracing::Span::current().record("outcome", "cancelled");
```
goto-bus-stop: Nit: attribute name and values could be constants

apollo-router/src/query_planner/caching_query_planner.rs
```diff
@@ -507,96 +526,207 @@ where
...
+                }
+            } else {
+                // some clients might timeout and cancel the request before query planning is finished,
+                // so we execute it in a task that can continue even after the request was canceled and
+                // the join handle was dropped. That way, the next similar query will use the cache instead
+                // of restarting the query planner until another timeout
+                tokio::task::spawn(planning_task).await
+            }
+            .inspect(|res| {
+                match res {
```
goto-bus-stop, line 715: I think it's worth adding a comment that we won't reach this `match` if the future was already cancelled, and that it won't overwrite the `outcome = cancelled` attribute that was added above.

apollo-router/src/query_planner/caching_query_planner.rs
```diff
@@ -507,96 +526,207 @@ where
...
+                                        Ordering::SeqCst,
+                                        Ordering::SeqCst,
+                                    )
+                                    .is_ok()
+                            {
+                                tracing::Span::current().record("outcome", "timeout");
+                            }
+                            res.map_err(convert_timeout_error)?
+                        } else if measure_mode {
+                            // In measure mode, spawn a timeout task that only records outcome
```
goto-bus-stop, line 676: I'm not a huge fan of the code duplication between enforce mode and measure mode, but it does seem justified 🙈; otherwise we would not be able to use tokio's `.timeout`, and instead would have to manually abort the task after the sleep.

apollo-router/src/query_planner/caching_query_planner.rs
```diff
@@ -847,6 +1038,225 @@ mod tests {
...
+            .extensions()
+            .with_lock(|lock| lock.insert::<ParsedDocument>(doc));
+
+        // Create a span with the outcome field declared
+        let span = tracing::info_span!("test_span", outcome = tracing::field::Empty);
+
+        // Keep the span alive and ensure it's the current span during the entire operation
+        let _span_guard = span.enter();
+
+        // Instrument the call to ensure span context is propagated
```
goto-bus-stop: wrong/outdated comment?

apollo-router/src/configuration/cooperative_cancellation.rs
```diff
@@ -0,0 +1,57 @@
+use schemars::JsonSchema;
+use serde::Deserialize;
+use serde::Serialize;
+
+use crate::configuration::mode::Mode;
+
+pub(crate) type CooperativeCancellation = Mode<Config>;
```
goto-bus-stop: I do prefer if we use the less-strict format from the design doc:
```yaml
      enabled: true # or false
      mode: measure # or enforce
      timeout: 5s # or nothing
```
mostly because it matches demand control, and we also already allow setting `enabled: false` together with other options in almost all our plugins (I take your point that this allows for useless/ineffective configs, but imho consistency is more important).

.changesets/feat_zelda_feature_cooperative_cancellation.md
```diff
@@ -0,0 +1,65 @@
+## Cooperative Cancellation for Query Planning
```
goto-bus-stop, line 1:
```suggestion
## Cooperative Cancellation for Query Planning ([PR #7604](https://github.com/apollographql/router/pull/7604))
```

I suggest:
- keeping the changeset text more concise, maybe just give one config example
- mentioning the metric and trace attribute name

apollo-router/src/query_planner/caching_query_planner.rs
```diff
@@ -167,7 +187,7 @@ where
         let _timer = Timer::new(|duration| {
             f64_histogram!(
                 "apollo.router.query_planning.warmup.duration",
-                "Time spent warming up the query planner queries in seconds",
+                "Time spent warming up the query planner queries",
```
goto-bus-stop:
```suggestion
                "Time spent warming up the query planner queries in seconds",
```
let's keep it because we can't introduce the unit thing yet 🥲

apollo-router/src/query_planner/caching_query_planner.rs
```diff
@@ -507,96 +526,207 @@ where
...
+                    scopeguard::guard(planning_task.abort_handle(), move |abort_handle| {
+                        // Client drop handler
+                        // Only record outcome if not already recorded, and only if timeout hasn't already been recorded
+                        if outcome_recorded_for_abort
+                            .compare_exchange(
+                                Outcome::None as u8,
+                                Outcome::Cancelled as u8,
+                                Ordering::SeqCst,
+                                Ordering::SeqCst,
+                            )
```
goto-bus-stop, line 637: smart, i like it

goto-bus-stop (commented):
Mostly minor comments, the actual feature is a little more involved than I expected and I overall like the solution :) My only real objection is the config shape.

goto-bus-stop (approved):
Pre-approval assuming comments are addressed in a manner considered adequate by the PR author 🫡

BrynCooke (changes requested):
Locking as reviewing

.changesets/feat_zelda_feature_cooperative_cancellation.md
```diff
@@ -0,0 +1,22 @@
...
+### Example
+
+Configuring a timeout in Measure Mode:
+```yaml
+supergraph:
+  query_planning:
+    experimental_cooperative_cancellation:
+      enabled: true
+      mode: measure
+      timeout_in_seconds: 1.0
```
bnjjj: We often use human time for timeout or duration. Something like `1s` Do you think we could replicate this for this config ?

bnjjj: It would then be called `timeout`.

apollo-router/src/query_planner/caching_query_planner.rs
```diff
@@ -507,96 +534,209 @@ where
...
                 CacheResolverError::RetrievalError(Arc::new(QueryPlannerError::JoinError(
                     e.to_string(),
                 )))
-            })?
+            }
+
+            // When cooperative cancellation is enabled, we want to cancel the query planner
+            // task if the request is canceled.
+            if self.cooperative_cancellation.is_enabled() {
+                let outcome_recorded = Arc::new(AtomicU8::new(Outcome::None as u8));
```
BrynCooke, line 622: Outcome needs to be documented in the docs.

apollo-router/src/query_planner/query_planner_service.rs
```diff
@@ -184,7 +184,16 @@ impl QueryPlannerService {
             let result = result.map_err(FederationErrorBridge::from);
 
             let elapsed = start.elapsed().as_secs_f64();
-            metric_query_planning_plan_duration(RUST_QP_MODE, elapsed);
+            match &result {
+                Ok(_) => metric_query_planning_plan_duration(RUST_QP_MODE, elapsed, "success"),
+                Err(FederationErrorBridge::Cancellation(e)) if e.contains("timeout") => {
```
bnjjj, line 189: I'm wondering if there's no better way to handle this error ? With a type instead of trying to check if the error message contains `timeout` ?

apollo-router/src/query_planner/caching_query_planner.rs
goto-bus-stop: I think docs will be a follow up, because merging docs for unreleased features into dev goes straight to the public docs website...

apollo-router/src/query_planner/query_planner_service.rs
Velfi: Sadly, FederationErrorBridge is stringly-typed:

```rust
pub(crate) enum FederationErrorBridge {
    /// {0}
    UnknownOperation(String),
    /// {0}
    OperationNameNotProvided(String),
    /// {0}
    Other(String),
    /// {0}
    Cancellation(String),
}
```

Velfi: I could try adding a flag to it that gets set at creation if you like.

Velfi: We could add a new variant to `SingleFederationError` or update the existing cancelled variant but the function responsible for checking for cancellation isn't aware of the reason for cancellation and I'm not sure how to communicate it.

Velfi: Another option: I could remove the timeout heuristic for now and we could just say that it was cancelled.

apollo-router/src/query_planner/caching_query_planner.rs
carodewig: Is there a way to include comments in our MDX doc files? That way releasing the feature would just involve uncommenting the docs that were added as part of this PR.

mabuyo: Popping in here from Edu team! I'm not as familiar with the router side of docs, but I did come across [this PR on the MCP server](https://github.com/apollographql/apollo-mcp-server/pull/132/files) that uncomments out the part of the docs so it seems like it should be possible!

apollo-router/src/query_planner/query_planner_service.rs
bnjjj: If there's no other viable option let's keep it like this

apollo-router/src/query_planner/caching_query_planner.rs
Velfi: I added docs that explain outcome to the metric docs.

apollo-router/src/query_planner/caching_query_planner.rs
```diff
@@ -507,96 +533,205 @@ where
...
+                let planning_task = tokio::task::spawn(planning_task);
```
BrynCooke, line 623: This puts the actual planning back on the tokio main thread pool, and bypasses the compute job pool.
I think this is probably not what we want.

BrynCooke (changes requested):

apollo-router/src/query_planner/caching_query_planner.rs
BrynCooke: Actually I'm talking garbage. This is in the caching query planner.

PR #7605, opened 2025-06-03
Stop using `tracing_test::traced_test` macro

carodewig:
`tracing_test::traced_test` installs a global tracing subscriber via a `OnceLock`; this can cause other tests to fail.

Instead, make use of some of the internals of `tracing_test` to capture log output. 

---

**Checklist**

- Tests added and passing[^4]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/plugins/authentication/jwks.rs
+++ b/apollo-router/src/plugins/authentication/jwks.rs
@@ -722,6 +722,7 @@ mod test {
     use super::APOLLO_AUTHENTICATION_JWT_CLAIMS;
     use super::Context;
     use super::jwt_expires_in;
+    use crate::test_harness::tracing_test_subscriber;
 
     #[test]
     fn test_exp_defaults_to_max_when_no_jwt_claims_present() {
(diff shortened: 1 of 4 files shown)


apollo-router/src/plugins/authentication/jwks.rs
```diff
@@ -731,20 +732,27 @@ mod test {
...
+        let _guard = tracing::dispatcher::set_default(&subscriber);
+
         let context = Context::new();
         context.insert_json_value(APOLLO_AUTHENTICATION_JWT_CLAIMS, json!("not an object"));
 
         let expiry = jwt_expires_in(&context);
         assert_eq!(expiry, Duration::MAX);
 
-        assert!(logs_contain("expected JWT claims to be an object"));
+        assert!(tracing_test::internal::logs_with_scope_contain(
```
goto-bus-stop: It might be good to wrap this up in a function in test_harness so that all access to tracing-test internals is in one place and can be updated together if necessary down the line.

goto-bus-stop (commented):
should we pin the exact version of tracing-test in cargo.toml, as the `internal` module has no stability guarantees?

PR #7610, opened 2025-06-05
Move the ParsedDocument hack

goto-bus-stop:
We've been looking at refactoring parts of the pipeline for a long time.
This change should solve a big blocker to making tweaks to the pipeline,
and refactoring functionality into layers.

Whenever I try to move something to a more sensible place, all the tests
start failing, because the supergraph service must have a
`ParsedDocument` in context (it's inserted by the router service), but
none of the tests put it in context. Actually, tests usually **cannot**
put it in the context, because to do so, they need to be able to parse
the query against the schema. When using `TestHarness`, you do not have
access to the schema.

Solving that completely will be difficult. But we can make things a bit
better. The hack is currently applied _right before_ query planning.
We can instead apply it inside the `TestHarness`, which more closely
mimicks the real-world behaviour: the ParsedDocument is inserted before
we reach the supergraph service.

Now, we can freely move things around inside the supergraph service
without breaking a load of tests that rely on this hack.

Some of our tests were testing that the supergraph service
raised an error on invalid GraphQL. The supergraph service does not
handle invalid GraphQL, so this was testing a nonsensical behaviour. I
moved those tests to use the router service.


apollo-router/src/test_harness.rs
```diff
@@ -297,32 +297,59 @@ impl<'a> TestHarness<'a> {
...
 
-        Ok(tower::service_fn(move |request| {
+        Ok(tower::service_fn(move |request: supergraph::Request| {
             let router = supergraph_creator.make();
 
+            // FIXME: we have about 80 tests creating a supergraph service and crafting a supergraph request for it
+            // none of those tests create an executable document to put it in the context, and the document cannot be created
+            // from inside the supergraph request fake builder, because it needs a schema matching the query.
+            // So while we are updating the tests to create a document manually, this here will make sure current
+            // tests will pass.
```
goto-bus-stop: I'd actually suggest that we _shouldn't_ create a document manually in tests (because we don't have the canned schema readily available in most cases), and instead either rely on this hack or have a larger change to the test harness to make that ergonomic.

apollo-router/tests/integration/introspection.rs
```diff
@@ -107,51 +107,6 @@ async fn two_operations() {
     "###);
 }
 
-#[tokio::test]
-async fn operation_name_error() {
```
SimonSapin, line 111: Instead of removing these two tests can we change them to use the router service?

goto-bus-stop: Oh i should've explained, the same tests already exist in operation_name.rs, and they don't test anything related to introspection

SimonSapin (approved):

PR #7611, opened 2025-06-05
Move operation limits enforcement into a layer

goto-bus-stop:
An example of the refactoring unlocked by #7610.

Operation limits enforcement can be done in a layer at the supergraph service. This way it's a composable piece of functionality that we can add or remove in a single-line code change. This also means it's done _a bit earlier_ in the pipeline. And that we don't need to thread the generated query metrics through the query planner service just to be able to put it in context. Because we don't need a bunch of hardcoded bits in the query planner service to do this, we can actually move all operation limits code into the limits plugin.

### DIFF
--- a/apollo-router/src/plugins/limits/enforce_operation_limits_layer.rs
+++ b/apollo-router/src/plugins/limits/enforce_operation_limits_layer.rs
@@ -131,8 +131,6 @@ where
 
                 return std::future::ready(response).boxed();
             }
-        } else {
-            panic!("No document?");
         }
 
         self.inner.call(req).boxed()
--- a/apollo-router/src/plugins/limits/enforce_operation_limits_layer.rs
+++ b/apollo-router/src/plugins/limits/enforce_operation_limits_layer.rs
@@ -3,20 +3,20 @@ use futures::StreamExt as _;
 use futures::future::BoxFuture;
 use tower::Service;
 
+use super::operation_limits::OperationLimits;
 use crate::graphql;
 use crate::services::layers::query_analysis::ParsedDocument;
 use crate::services::supergraph;
-use crate::spec::operation_limits::OperationLimits;
 
 /// Layer that enforces operation limits and rejects GraphQL requests that exceed the limits.
 ///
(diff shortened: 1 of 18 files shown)


apollo-router/src/plugins/limits/enforce_operation_limits_layer.rs
```diff
@@ -0,0 +1,140 @@
...
+                    .map_err(Self::Error::from)
+                    .map(|http_response| supergraph::Response {
+                        response: http_response,
+                        context: req.context,
+                    });
+
+                return std::future::ready(response).boxed();
+            }
+        } else {
+            panic!("No document?");
```
goto-bus-stop: This was for debugging, before merging we would simply delete this panic (and not enforce limits when there is no parsed document, because that doesn't make sense)

apollo-router/src/plugins/limits/enforce_operation_limits_layer.rs
```diff
@@ -0,0 +1,140 @@
...
+            config: config.clone(),
+        }
+    }
+}
+
+impl<S> tower::Layer<S> for EnforceOperationLimitsLayer {
+    type Service = EnforceOperationLimits<S>;
+
+    fn layer(&self, inner: S) -> Self::Service {
+        EnforceOperationLimits {
```
goto-bus-stop, line 30: We could also use one of the `.map_future`-style layers here to implement the service. I think it's a pretty good exercise to get into writing our own services though, once you understand them it's not that hard... and we all need to understand them to work with tower effectively.

goto-bus-stop:
I would still like to do this but should likely hardcode where the operation limits are applied, instead of using a plugin hook, since it risks changing user-observable behaviour.
The layer could maybe go on the QueryPlannerService which would be a nice demonstration of what we intend to do more of in the future.

aaronArinder:
still necessary or closeable? going to close next weds if I don't hear back, but we can always reopen it if it turns out to be needed

goto-bus-stop:
Still nice to have but closeable. We'd eventually naturally pick this up as we tower more in the router.

PR #7613, opened 2025-06-05
Implement validation for tag

tninesling:
Implements validation for the `@tag` directive, which asserts the `name` argument is under 129 characters and matches a specified regex pattern. The validation itself is simple, but the implementation involved adding all the remaining locations for the `@tag` directive to our `tag_directive_applications` iterator. I have added some subgraph validation tests to assert the validation hits all the relevant schema locations.

---

**Checklist**

- Tests added and passing[^4]
    - [ ] Unit Tests
    - [X] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-federation/src/error/mod.rs
```diff
@@ -482,6 +482,8 @@ pub enum SingleFederationError {
     ListSizeInvalidSlicingArgument { message: String },
     #[error("{message}")]
     ListSizeInvalidSizedField { message: String },
+    #[error("{message}")]
```
clenfest, line 485: Do we want to start adding these to `CompositionError` rather than `SingleFederationError`?

apollo-federation/src/schema/validators/tag.rs
```diff
@@ -0,0 +1,53 @@
...
+    schema: &FederationSchema,
+    errors: &mut MultipleFederationErrors,
+) -> Result<(), FederationError> {
+    let tag_applications = schema.tag_directive_applications()?;
+
+    for tag_directive_result in tag_applications {
+        let tag_directive = match tag_directive_result {
+            Ok(directive) => directive,
+            Err(error) => {
+                errors.push(error);
```
clenfest, line 28: What type of errors can occur here? Should we push these errors onto the error stack or bail?

clenfest (approved):
Two questions ,but otherwise looks good.

apollo-federation/src/error/mod.rs
clenfest: Actually, I think this is good for now, we should do that in a separate refactoring PR for all the validators.

apollo-federation/src/schema/validators/tag.rs
tninesling: The errors produced by the iterator in this case are issues with parsing arguments. For example, if you're missing a required argument, that will appear here as an error. I've been pushing these onto the error collector for all application iterators, as I think these roughly align with a validation error.

apollo-federation/src/error/mod.rs
tninesling: Ah, totally forgot we split those out 😅 I'll make a ticket to move these over

PR #7623, opened 2025-06-06
[JSONSelection] Reenable `->` method shape checking and fully implement `math_shape`

benjamn:
Although we have not made them public yet, one precondition for shipping the arithmetic methods (`->add`, `->sub`, `->mul`, `->div`, and `->mod`) is fully implementing the code responsible for statically determining the output "shape" of each method (given input/argument shapes), which in this case is always `Int`, `Float`, or possibly a `Error` if the input/argument shapes are not acceptable.

This PR provides this implementation for `math_shape`, but in order to make these changes meaningful/testable, I've also included a commit that reenables method shape checking as part of `PathList::compute_output_shape`, aligning it better with the runtime changes to `PathList::apply_to_path` from PR #7144, and resolving [this open JIRA task](https://apollographql.atlassian.net/browse/CNN-722) (see commit messages for more details).

### DIFF
--- a/apollo-federation/src/connectors/json_selection/apply_to.rs
+++ b/apollo-federation/src/connectors/json_selection/apply_to.rs
@@ -2863,6 +2863,113 @@ mod tests {
     }
 
     #[test]
+    fn test_left_associative_output_shapes() {
+        assert_eq!(selection!("$batch.id").shape().pretty_print(), "$batch.id");
+
+        assert_eq!(
+            selection!("$batch.id->first").shape().pretty_print(),
+            "$batch.id.0",
+        );
+
+        assert_eq!(
+            selection!("$batch.id->last").shape().pretty_print(),
+            "$batch.id.*",
+        );
+
+        let mut named_shapes = IndexMap::default();
+        named_shapes.insert(
+            "$batch",
+            Shape::list(
+                Shape::record(
+                    {
+                        let mut map = Shape::empty_map();
+                        map.insert("id".to_string(), Shape::int([]));
+                        map
+                    },
+                    [],
+                ),
+                [],
+            ),
+        );
+
+        let root_shape = Shape::name("$root", []);
+
+        let computed_batch_id = selection!("$batch.id").compute_output_shape(
+            root_shape.clone(),
+            &named_shapes,
+            &SourceId::Other("JSONSelection".into()),
+        );
+        assert_eq!(computed_batch_id.pretty_print(), "List<Int>");
+
+        let computed_first = selection!("$batch.id->first").compute_output_shape(
+            root_shape.clone(),
+            &named_shapes,
+            &SourceId::Other("JSONSelection".into()),
+        );
+        assert_eq!(computed_first.pretty_print(), "One<Int, None>");
+
+        let computed_last = selection!("$batch.id->last").compute_output_shape(
+            root_shape.clone(),
+            &named_shapes,
+            &SourceId::Other("JSONSelection".into()),
+        );
+        assert_eq!(computed_last.pretty_print(), "One<Int, None>");
+
+        assert_eq!(
+            selection!("$batch.id->jsonStringify")
+                .shape()
+                .pretty_print(),
+            "String",
+        );
+
+        assert_eq!(
+            selection!("$batch.id->map([@])->echo([@])->jsonStringify")
+                .shape()
+                .pretty_print(),
+            "String",
+        );
+
+        assert_eq!(
+            selection!("$batch.id->map(@)->echo(@)")
+                .shape()
+                .pretty_print(),
+            "List<$batch.id.*>",
+        );
+
+        assert_eq!(
+            selection!("$batch.id->map(@)->echo([@])")
+                .shape()
+                .pretty_print(),
+            "[List<$batch.id.*>]",
+        );
+
+        assert_eq!(
+            selection!("$batch.id->map([@])->echo(@)")
+                .shape()
+                .pretty_print(),
+            "List<[$batch.id.*]>",
+        );
+
+        assert_eq!(
+            selection!("$batch.id->map([@])->echo([@])")
+                .shape()
+                .pretty_print(),
+            "[List<[$batch.id.*]>]",
+        );
+
+        assert_eq!(
+            selection!("$batch.id->map([@])->echo([@])")
+                .compute_output_shape(
+                    root_shape,
+                    &named_shapes,
+                    &SourceId::Other("JSONSelection".into()),
+                )
+                .pretty_print(),
+            "[List<[Int]>]",
+        );
+    }
+
+    #[test]
     fn test_lit_paths() {
         let data = json!({
             "value": {
--- a/apollo-federation/src/connectors/json_selection/apply_to.rs
+++ b/apollo-federation/src/connectors/json_selection/apply_to.rs
@@ -649,7 +649,64 @@ impl ApplyToInternal for WithRange<PathList> {
         named_var_shapes: &IndexMap<&str, Shape>,
         source_id: &SourceId,
     ) -> Shape {
-        match self.as_ref() {
+        if input_shape.is_none() {
+            // If the previous path prefix evaluated to None, path evaluation
+            // must terminate because there is no JSON value to pass as the
+            // input_shape to the rest of the path, so the output shape of the
+            // whole path must be None. Any errors that might explain an
+            // unexpected None value
+            return input_shape;
+        }
+
+        match input_shape.case() {
+            ShapeCase::One(shapes) => {
+                return Shape::one(
+                    shapes.iter().map(|shape| {
+                        self.compute_output_shape(
+                            shape.clone(),
+                            dollar_shape.clone(),
+                            named_var_shapes,
+                            source_id,
+                        )
+                    }),
+                    [], // TODO Better locations?
+                );
+            }
+            ShapeCase::All(shapes) => {
+                return Shape::all(
+                    shapes.iter().map(|shape| {
+                        self.compute_output_shape(
+                            shape.clone(),
+                            dollar_shape.clone(),
+                            named_var_shapes,
+                            source_id,
+                        )
+                    }),
+                    [], // TODO Better locations?
+                );
+            }
+            ShapeCase::Error(error) => {
+                return match error.partial.as_ref() {
+                    Some(partial) => Shape::error_with_partial(
+                        error.message.clone(),
+                        self.compute_output_shape(
+                            partial.clone(),
+                            dollar_shape,
+                            named_var_shapes,
+                            source_id,
+                        ),
+                        input_shape.locations.clone(),
+                    ),
+                    None => input_shape.clone(),
+                };
+            }
+            _ => {}
+        };
+
+        // Given the base cases above, we can assume below that input_shape is
+        // neither ::One, ::All, nor ::Error.
+
+        let (current_shape, tail) = match self.as_ref() {
             PathList::Var(ranged_var_name, tail) => {
                 let var_name = ranged_var_name.as_ref();
                 let var_shape = if var_name == &KnownVariable::AtSign {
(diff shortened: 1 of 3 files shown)


apollo-federation/src/connectors/validation/snapshots/validation_tests@headers__expressions_that_evaluate_to_invalid_types.graphql.snap
```diff
@@ -4,13 +4,27 @@ expression: "format!(\"{:#?}\", result.errors)"
...
             12:45..12:47,
         ],
     },
+    Message {
+        code: InvalidHeader,
+        message: "In `@source(http.headers:)`: object values aren't valid here",
+        locations: [
+            13:73..13:75,
+        ],
+    },
```
benjamn, line 27: These additions are legit, if you check the schema in question:
```graphql
extend schema
@link(
    url: "https://specs.apollo.dev/connect/v0.2"
    import: ["@connect", "@source"]
)
@source(
    name: "v1"
    http: {
        baseURL: "http://127.0.0.1"
        headers: [
            { name: "an-array", value: "{$->echo([])}" }, // line 11
            { name: "an-object", value: "{$({})}"},
            { name: "maybe-object", value: "{$config->match([1, 1], [2, {}])}" }, // line 13
        ]
    }
)

type Query {
    blah(anArray: [String]): String @connect(
        source: "v1",
        http: {GET: "/blah", headers: {name: "an-array", value: "{$args.anArray}"}},
        selection: "$"
    )
}
```

apollo-federation/src/connectors/json_selection/methods/future/arithmetic.rs
```diff
@@ -139,13 +140,165 @@ infix_math_op!(rem_op, %);
...
+    }
+
+    match check_result {
+        CheckNumericResult::IntForSure => {
+            // TODO If we wanted to climb Mount Cleverest, we could perform
+            // static integer math when all the inputs are statically known, and
+            // return a specific Shape::int_value when that math succeeds. That
+            // might require using different shape functions for each of the
+            // math operations, rather than a single math_shape function.
+            Shape::int(method_name.shape_location(source_id))
```
benjamn, line 202: > [!CAUTION]
> Many person-lifetimes have been lost on the treacherous slopes of Mount Cleverest.

apollo-federation/src/connectors/json_selection/apply_to.rs
```diff
@@ -2795,6 +2862,113 @@ mod tests {
         );
     }
 
+    #[test]
+    fn test_left_associative_output_shapes() {
```
andrewmcgivery, line 2866: Small nit but we're trying to avoid having several test cases in a single test (general rule of thumb being when you hit an "assert", that should be the end of the test). 

Can you break this function up into several tests instead of one long one? :)

apollo-federation/src/connectors/json_selection/methods/future/arithmetic.rs
```diff
@@ -139,13 +140,165 @@ infix_math_op!(rem_op, %);
 #[allow(dead_code)] // method type-checking disabled until we add name resolution
```
andrewmcgivery: Is this truly still dead code if this PR is re-enabling the type checking?

apollo-federation/src/connectors/json_selection/methods/future/arithmetic.rs
```diff
@@ -139,13 +140,165 @@ infix_math_op!(rem_op, %);
...
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+enum CheckNumericResult {
+    IntForSure,
+    FloatPossible,
+    Neither,
+}
+
+fn check_numeric_shape(arg_shape: &Shape) -> CheckNumericResult {
```
andrewmcgivery: Just thinking out loud here... if we broke the math methods out into separate methods instead of methods created by the macro, would that simplify this code at all or is this just a reality of the shape of math functions? 🤔 

I'm cool if the answer is "this is just the way it is" but wanted to check cause I was expecting the shape code for Math to be much simpler but that may just be me being naive and not yet familiar enough with shape 😅

apollo-federation/src/connectors/json_selection/methods/future/arithmetic.rs
```diff
@@ -349,4 +505,171 @@ mod tests {
             (Some(json!(0)), vec![]),
         );
     }
+
+    #[test]
+    fn add_shape_can_return_int() {
```
andrewmcgivery, line 586: Same request as the previous file... can you break these function up into several tests instead of one long one so that each test only hits one test case? :)

apollo-federation/src/connectors/validation/snapshots/validation_tests@headers__expressions_that_evaluate_to_invalid_types.graphql.snap
andrewmcgivery: Is it because re-enabling the method shapes made these errors actually work properly? 😅

apollo-federation/src/connectors/json_selection/methods/future/arithmetic.rs
```diff
@@ -349,4 +505,171 @@ mod tests {
...
+
+        let missing_catchall_case = selection!("$->add(@->eq(0)->match([true, 1], [false, 2]))");
+        let missing_catchall_case_shape = missing_catchall_case.compute_output_shape(
+            Shape::int([]),
+            &IndexMap::default(),
+            &SourceId::Other("JSONSelection".into()),
+        );
+        assert_eq!(
+            missing_catchall_case_shape.pretty_print(),
+            "Error<\"Method ->add received non-numeric argument 0\">"
```
lennyburdette, line 727: what does `argument 0` refer to? the argument at index 0?

apollo-federation/src/connectors/json_selection/apply_to.rs
benjamn: Sure, I can break this up, but I'm still going to assert multiple things in each test function. Restricting ourselves to one assert per test function doesn't seem right.

lennyburdette (commented):
what's our strategy for enabling the new errors only for v0.3?

apollo-federation/src/connectors/json_selection/methods/future/arithmetic.rs
benjamn: Yeah, should that be 1-indexed (ordinal)?

apollo-federation/src/connectors/json_selection/methods/future/arithmetic.rs
benjamn: Thanks for calling out the complexity here!

Bad news: yes, this is the kind of logic that needs to happen somehow, in a type system where unions and intersections and placeholder named types exist.

Good news: I believe we can use the `Shape::accepts(&self, other: &Shape) -> bool` method to hide a lot of that complexity, since it's not complexity that's specific to the `math_shape` logic.

I've added some regression tests in https://github.com/apollographql/router/pull/7623/commits/533daa0dda5e6ca63ea82a3de9979ab4d221b6de to capture the current behavior of `check_numeric_shape`, and I pushed that commit by itself to demonstrate all CI tests are passing.

Following that, I'll reimplement `check_numeric_shape` to be simpler, ideally without failing (or changing) any of the tests.

benjamn: Here's the simplification: https://github.com/apollographql/router/pull/7623/commits/0c8c8f9ce843cb8da83d3d0178a15e2a48d28aec

apollo-federation/src/connectors/json_selection/methods/future/arithmetic.rs
benjamn: https://github.com/apollographql/router/pull/7623/commits/10f9db1863cf88bb42c81263e3f8e40d5ab14842

apollo-federation/src/connectors/validation/snapshots/validation_tests@headers__expressions_that_evaluate_to_invalid_types.graphql.snap
benjamn: Yeah, previously we had to give up and say every method (including `->echo`) returns `Unknown`, and now we can tell exactly what each of these cases returns.

Comment pointing to the line where `Unknown` used to be returned unconditionally: https://github.com/apollographql/router/pull/7623/files#r2132602377

apollo-federation/src/connectors/json_selection/apply_to.rs
```diff
@@ -661,95 +718,105 @@ impl ApplyToInternal for WithRange<PathList> {
...
+                        Some(tail),
+                    )
+                } else {
+                    (
                         Shape::error(
                             format!("Method ->{} not found", method_name.as_str()),
                             method_name.shape_location(source_id),
-                        )
-                    },
-                    |_method| Shape::unknown(method_name.shape_location(source_id)),
```
benjamn, line 742: Here's where we were previously returning the `Unknown` shape for all `->` methods.

apollo-federation/src/connectors/json_selection/apply_to.rs
andrewmcgivery: Sorry, the "general rule of thumb" isn't the end-all-be-all. If the asserts are all related to the same test case, that's totally cool.

I think more what I'm saying is we don't want this...

```
... run code...

... assert something ...

.. run more code...

... assert something else...
```

But this is totally fine:

```
... run code...

...assert something...
...assert something else about the code than ran above...
```

apollo-federation/src/connectors/json_selection/apply_to.rs
```diff
@@ -649,7 +649,64 @@ impl ApplyToInternal for WithRange<PathList> {
         named_var_shapes: &IndexMap<&str, Shape>,
         source_id: &SourceId,
     ) -> Shape {
-        match self.as_ref() {
+        if input_shape.is_none() {
+            // If the previous path prefix evaluated to None, path evaluation
+            // must terminate because there is no JSON value to pass as the
+            // input_shape to the rest of the path, so the output shape of the
+            // whole path must be None. Any errors that might explain an
+            // unexpected None value
```
dylan-apollo, line 657: This is missing the end of the sentence, I think?

apollo-federation/src/connectors/json_selection/apply_to.rs
```diff
@@ -661,95 +718,105 @@ impl ApplyToInternal for WithRange<PathList> {
...
-                named_var_shapes,
-                source_id,
+                Some(tail),
             ),
 
-            PathList::Method(method_name, _method_args, _tail) => ArrowMethod::lookup(method_name)
-                .map_or_else(
-                    || {
+            PathList::Method(method_name, method_args, tail) => {
+                if let Some(method) = ArrowMethod::lookup(method_name) {
```
dylan-apollo, line 766: This needs to be gated behind a check for connect spec >= 0.3

apollo-federation/src/connectors/json_selection/apply_to.rs
```diff
@@ -661,95 +718,105 @@ impl ApplyToInternal for WithRange<PathList> {
...
+            PathList::Selection(selection) => {
+                if input_shape.is_none() {
+                    return Shape::error(
+                        format!(
+                            "Selection cannot be applied to {}",
+                            input_shape.pretty_print()
+                        ),
+                        self.shape_location(source_id),
+                    );
+                }
```
dylan-apollo, line 798: Isn't this unreachable because of the check first thing in the function?

apollo-federation/src/connectors/json_selection/methods/future/arithmetic.rs
```diff
@@ -136,16 +136,111 @@ infix_math_op!(mul_op, *);
...
+            // might require using different shape functions for each of the
+            // math operations, rather than a single math_shape function.
+            Shape::int(method_name.shape_location(source_id))
+        }
+        CheckNumericResult::FloatPossible => Shape::float(method_name.shape_location(source_id)),
+        CheckNumericResult::Neither => unreachable!("handled above"),
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
```
dylan-apollo, line 209:
```suggestion
#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
```

apollo-federation/src/connectors/json_selection/methods/future/arithmetic.rs
```diff
@@ -136,16 +136,111 @@ infix_math_op!(mul_op, *);
...
+        return Shape::error(
+            format!(
+                "Method ->{} received non-numeric input",
+                method_name.as_ref()
+            ),
+            input_shape.locations.iter().cloned(),
+        );
+    }
+
+    if method_name.as_ref() == "div" {
```
dylan-apollo, line 159: This really should be comparing to an enum, not an inline string. I still think we should be passing around `ArrowMethod` instead of strings 😁

apollo-federation/src/connectors/json_selection/methods/future/arithmetic.rs
```diff
@@ -136,16 +136,111 @@ infix_math_op!(mul_op, *);
...
+        // The ->div method stays safe by always returning Float, so
+        // check_result starts off false in that case.
+        check_result = CheckNumericResult::FloatPossible;
+    }
+
+    for (i, arg) in method_args
+        .iter()
+        .flat_map(|args| args.args.iter())
+        .enumerate()
+    {
```
dylan-apollo, line 169: We're not checking that the number of arguments to the function is correct? Is it always one arg? If so, let's just verify that. If not, we need per-method validation I think.

apollo-federation/src/connectors/json_selection/methods/future/arithmetic.rs
```diff
@@ -136,16 +136,111 @@ infix_math_op!(mul_op, *);
...
+        CheckNumericResult::IntForSure => {
+            // TODO If we wanted to climb Mount Cleverest, we could perform
+            // static integer math when all the inputs are statically known, and
+            // return a specific Shape::int_value when that math succeeds. That
+            // might require using different shape functions for each of the
+            // math operations, rather than a single math_shape function.
+            Shape::int(method_name.shape_location(source_id))
+        }
+        CheckNumericResult::FloatPossible => Shape::float(method_name.shape_location(source_id)),
+        CheckNumericResult::Neither => unreachable!("handled above"),
```
dylan-apollo, line 205: I think we can actually make the Rust code represent this rather than risk a potential panic in a future refactoring.

We're already making a `Shape::int` and `Shape::float` in `check_numeric_shape`. Why don't we return the `Shape` from that and just add the location at the end here rather than translating the enum? Something like (pseudo):

```rust
let Some(first_arg_shape) = check_numeric_shape(&input_shape) else {
  // ... return error
}

let Some(second_arg_shape) = ...

if method == Method::Div { return Shape::float() }
if first_arg_shape.is_int() && second_arg_shape.is_int() { return Shape::int() }

return Shape::float()
```

dylan-apollo (changes requested):
A couple of comments in the review. Additionally:

1. I found one false positive while just messing around: `$(["a", 1])->slice(0)->first->slice(0)` is an error. Obviously this is a contrived example, but still something we need to allow. I'll want to spend more time looking deeper for other false positives.
2. This expression gave me a stack overflow when compiled to a `supergraph` binary 😓: `$args.anObject->entries`. Let me know if you want the full schema to help diagnose. We definitely have to root-cause-analysis that as well.
3. I think I was conflating two issues with my "wasn't there a reason we couldn't enable this?" question. The trouble is that enabling this is one breaking change, and enabling full type resolution is another. So we should have a plan to ship full type resolution with this in 0.3 rather than later need an additional switch. IIRC resolution already works for non-methods with the `expression` wrapper, so only methods should be affected?
4. Reminder to re-enable [these tests](https://github.com/apollographql/router/blob/f834eba96d654a3b1e1f87afdb45c3b3747f0cba/apollo-federation/src/connectors/validation/expression.rs#L735)

I also have a vague memory of there being a reason we couldn't just return an error when `.child()` is none, but I can't for the life of me figure out what it was!

benjamn:
Closing in favor of a forthcoming smaller PR with just the `math_shape`-related changes.

PR #7627, opened 2025-06-06
feat(federation/composition): port remaining functionality in `TransitionPathWithLazyIndirectPaths`

sachindshinde:
This PR ports the remaining functionality of the class `TransitionPathWithLazyIndirectPaths` from the JS codebase (this also updates the Rust `GraphPath::advance_with_non_collecting_and_type_preserving_transitions()` to provide unadvanceable metadata).


apollo-federation/src/query_graph/graph_path.rs
```diff
@@ -1656,240 +1699,373 @@ where
...
+                    edge_weight.transition,
+                    QueryGraphEdgeTransition::KeyResolution
+                ) {
+                    let edge_head_weight = self.graph.node_weight(edge_head)?;
+                    if edge_head_weight.source != edge_tail_weight.source {
+                        heap.push(HeapElement(updated_path));
                     }
+                }
+            }
+            if does_non_collecting_next_edge_exist {
```
duckki: Did you mean its negation?
```suggestion
            if !does_non_collecting_next_edge_exist {
```

The condition corresponds to `nextEdges.length === 0`, but `does_non_collecting_next_edge_exist` is true only if we have at least one next edges.

apollo-federation/src/query_graph/graph_path/transition.rs
```diff
@@ -282,13 +310,14 @@ impl TransitionGraphPath {
                             num_options,
                         );
                     };
-                    to_advance = Cow::Owned(updated_path);
+                    to_advance = Either::Right(updated_path);
                     // We can now continue on dealing with the actual field.
                 }
             }
         }
+        let to_advance: &Self = either::for_both!(&to_advance, p => p);
```
duckki: Is it different from below?
```suggestion
        let to_advance = &to_advance;
```

apollo-federation/src/query_graph/graph_path.rs
sachindshinde: Good catch!

apollo-federation/src/query_graph/graph_path/transition.rs
sachindshinde: I double-checked and it looks like we don't need this line at all (I think while I was writing the code, I needed it to compile, but at some point I changed the usage sites enough that it became unnecessary).

duckki: I thought you wanted to make it clear that it won't be mutable afterwards, which sounded good too :)

PR #7634, opened 2025-06-06
feat: add BASE_VERSION argument to Dockerfile and build commands

jasonbarnett667:
This pull request updates the build process and Dockerfile configuration to include a new `BASE_VERSION` argument, ensuring better version tracking and consistency across debug and release images.

### Build process updates:
* [`.circleci/config.yml`](diffhunk://#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47L742-R745): Added `BASE_VERSION` as a build argument in both debug and release image builds to enhance version management.

### Dockerfile configuration updates:
* [`dockerfiles/Dockerfile.router`](diffhunk://#diff-c3ec2f1d6640fe019f5473c8d95b385a50ff3eb78d293d92b6f3103453404380R26): Introduced the `BASE_VERSION` argument to the Dockerfile for use in the build process.
* [`dockerfiles/Dockerfile.router`](diffhunk://#diff-c3ec2f1d6640fe019f5473c8d95b385a50ff3eb78d293d92b6f3103453404380R60): Added a new label `org.opencontainers.image.version` to include the `BASE_VERSION` in the image metadata.

---

**Checklist**

- Tests added and passing[^4]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


garypen (commented):
Do you want BASE_VERSION or VERSION?

BASE_VERSION is the unadorned version from the Cargo.toml file and will look something like:
```
2.3.0
```

VERSION is modified to transform illegal docker name characters (such as `+` to `-`) and prepends a `v`. It will look like this:
```
v2.3.0
```
which matches the names we assign to our images in our registry: https://github.com/apollographql/router/pkgs/container/router

jasonbarnett667:
> Do you want BASE_VERSION or VERSION?
> 
> BASE_VERSION is the unadorned version from the Cargo.toml file and will look something like:
> 
> ```
> 2.3.0
> ```
> 
> VERSION is modified to transform illegal docker name characters (such as `+` to `-`) and prepends a `v`. It will look like this:
> 
> ```
> v2.3.0
> ```
> 
> which matches the names we assign to our images in our registry: https://github.com/apollographql/router/pkgs/container/router

I was wondering the same thing, I believe it has to be this way since this is the format we use in our security advisories and for them to match, we need to leave out the prefix.

garypen (approved):

goto-bus-stop:
@jasonbarnett667 Is this good to go? I assume yes but i'm not sure about the context for this change. if you think it's ready i can approve & merge it.

dockerfiles/Dockerfile.router
```diff
@@ -56,6 +57,7 @@ COPY dockerfiles/router.yaml config
 
 LABEL org.opencontainers.image.authors="Apollo Graph, Inc. ${REPO_URL}"
 LABEL org.opencontainers.image.source="${REPO_URL}"
+LABEL org.opencontainers.image.version="${BASE_VERSION}"
```
abernix, line 60: i think this probably isn't a breaking change because this particular file is meant for us to build our images, but I do wonder if just defaulting to blank would be better than requiring this argument?

I also wonder how many folks are actually using our `diy/Dockerfile`s to build the Router (we have it in our docs, I think) and if we'd benefit from hard-coding the base version into that `Dockerfile` so we could provide some semblance of similar functionality in that image.  Our `xtask release` script does some regex voodoo specifically to commit things like that, and we could extend the same mechanism. Just a thought, but this does seem like a reasonable starting point.

jasonbarnett667:
> @jasonbarnett667 Is this good to go? I assume yes but i'm not sure about the context for this change. if you think it's ready i can approve & merge it.

Yep - Sorry I missed this comment @goto-bus-stop 
@abernix I agree, I can do some testing and see if we can port this over to the DIY dockerfiles.

dockerfiles/Dockerfile.router
jasonbarnett667: I **_think_** not passing an arg defaults to blank but I will have to test to make sure. I think we want to ensure we are including this in the build though.

PR #7635, opened 2025-06-06
feat(federation/composition): ported ValidationContext class

duckki:
ported `ValidationContext` class.

work in progress: unit tests.

---

**Checklist**

- Tests added and passing[^4]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

Incremental work towards a new feature.

### DIFF
--- a/apollo-federation/src/composition/satisfiability.rs
+++ b/apollo-federation/src/composition/satisfiability.rs
@@ -1,12 +1,159 @@
 mod satisfiability_error;
 
+use apollo_compiler::Name;
+use apollo_compiler::Node;
+use apollo_compiler::ast;
+use apollo_compiler::collections::IndexMap;
+use apollo_compiler::collections::IndexSet;
+
+use crate::CONTEXT_VERSIONS;
+use crate::JoinSpecDefinition;
+use crate::bail;
 use crate::error::CompositionError;
+use crate::error::FederationError;
+use crate::link::spec::Identity;
+use crate::schema::FederationSchema;
+use crate::schema::position::CompositeTypeDefinitionPosition;
+use crate::schema::position::FieldDefinitionPosition;
+use crate::schema::position::TypeDefinitionPosition;
 use crate::supergraph::Merged;
 use crate::supergraph::Satisfiable;
 use crate::supergraph::Supergraph;
+use crate::validate_supergraph_for_query_planning;
 
 pub fn validate_satisfiability(
     _supergraph: Supergraph<Merged>,
 ) -> Result<Supergraph<Satisfiable>, Vec<CompositionError>> {
     panic!("validate_satisfiability is not implemented yet")
+}
+
+struct ValidationContext {
+    supergraph_schema: FederationSchema,
+    join_spec: &'static JoinSpecDefinition,
+    join_type_directive: Node<ast::DirectiveDefinition>,
+    join_field_directive: Node<ast::DirectiveDefinition>,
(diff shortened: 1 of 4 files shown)


apollo-federation/src/composition/satisfiability.rs
```diff
@@ -1,12 +1,306 @@
...
+            .clone();
+        let join_field_directive = join_spec
+            .field_directive_definition(&supergraph_schema)?
+            .clone();
+
+        let mut types_to_contexts = IndexMap::default();
+        let Some(links_metadata) = supergraph_schema.metadata() else {
+            bail!("Supergraph schema should have links metadata");
+        };
+        if let Some(context_link) = links_metadata.for_identity(&Identity::context_identity()) {
```
tninesling: I know the JS doesn't have this, but it looks like `validate_supergraph_for_query_planning` is already returning an `Option<&'static ContextSpecDefinition>` (third item in the tuple)

apollo-federation/src/schema/mod.rs
```diff
@@ -411,6 +413,32 @@ impl FederationSchema {
         Ok(applications)
     }
 
+    /// For supergraph schemas where the `@context` directive is a "context" spec directive.
+    pub(crate) fn context_directive_applications_in_supergraph(
```
tninesling, line 417: This is a great find, and it's actually a bug in the current implementation of `context_directive_applications`. Subgraphs can still import `@context` directly from the context spec, so we need to handle both. I think the right way to go about that is to replace
```
let federation_spec = get_federation_spec_definition_from_subgraph(self)?;
let context_directive_definition = federation_spec.context_directive_definition(self)?;
let context_directive_referencers = self.referencers().get_directive(&context_directive_definition.name)?;
```
with
```
let Some(context_directive_name) = ContextSpecDefinition::context_directive_name(&self)? else {
  return Ok(Vec::new());
}
let context_directive_referencers = self.referencers().get_directive(context_directive_name)?;
```

The `ContextDirectiveSpec` first checks for direct `context` spec imports and falls back to `federation` imports.

sachindshinde: The context spec is purely a supergraph schema spec. If the subgraph schema tries to import it, it should be ignored (IIRC that's the behavior today in JS composition). Subgraphs today use `@context` by importing it from the federation spec (or using the prefixed version, `@federation__context`).

duckki: I just confirmed too that, if the context spec is used in subgraph, it is ignored. So, the existing `context_directive_applications` function should be correct.

tninesling: Regardless of whether we disallow importing directly from the context spec or not, I think:
1. We can update the code to handle both and reuse the iterator for subgraphs and supergraphs. This makes it easier to maintain in the future if we add new locations to the context directives
2. We will need to support this for other specs (tag, cost, etc) which can be imported from their base spec or the federation spec, and I think it would be good to provide breadcrumbs for that pattern if someone needs to define one of these iterators for another directive in the future

tninesling (approved):
Added one last argument on combining the context iterators, but I don't think that decision should block this PR. Otherwise, looks great

apollo-federation/src/schema/mod.rs
duckki: I'll think more about the refactoring with `tag`/`cost` in mind and create a separate PR.

PR #7637, opened 2025-06-06
Optional chaining for selection syntax

andrewmcgivery:
Optional chaining for selection syntax

---

**Checklist**

- Tests added and passing[^4]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-federation/src/connectors/json_selection/apply_to.rs
```diff
@@ -749,6 +757,18 @@ impl ApplyToInternal for WithRange<PathList> {
...
+            PathList::Question(continuation) => {
+                // Optional operation always produces nullable output
+                let result_shape = continuation.compute_output_shape(
+                    input_shape,
+                    dollar_shape,
+                    named_var_shapes,
+                    source_id,
+                );
+                // Make result nullable since optional chaining can produce null
+                Shape::one([result_shape, Shape::none()], vec![])
```
andrewmcgivery: Difference between this approach versus:

```
                if input_shape.is_null() {
                    // TODO Somehow copy names from input_shape to this new
                    // Shape::none() shape?
                    (Shape::none(), None)
                } else {
                    // TODO This should return input_shape.question() once we
                    // support that new child shape method.
                    // (input_shape.question(), Some(tail))
                    (input_shape, Some(tail))
                }
```

apollo-federation/src/connectors/json_selection/apply_to.rs
```diff
@@ -634,6 +634,14 @@ impl ApplyToInternal for WithRange<PathList> {
                 )
             }
             PathList::Selection(selection) => selection.apply_to_path(data, vars, input_path),
+            PathList::Question(continuation) => {
+                // Universal null check for any operation after ?
+                if data.is_null() {
+                    (Some(JSON::Null), vec![])
```
andrewmcgivery: Should this actually be `None`?

apollo-federation/src/connectors/json_selection/parser.rs
```diff
@@ -553,6 +553,10 @@ pub(super) enum PathList {
     // middle/tail (not the beginning) of a PathSelection.
     Method(WithRange<String>, Option<MethodArgs>, WithRange<PathList>),
 
+    // Universal null guard that can wrap any path continuation.
+    // If data is null, returns null instead of continuing with the wrapped operation.
+    Question(WithRange<PathList>),
```
andrewmcgivery, line 830: Ben's prototype had another param where `The firstAdd commentMore actions parameter wraps the range of the ? token.` but I am unclear as to why.

It's probably related to why Ben's version below also had `depth + 1`.

andrewmcgivery:
EDIT: This has been resolved after discussion with Ben. :) 

In `apply_to.rs`, Ben's code has a change to the `Property {} not found in {}` logic. I think this is making it not return errors on not found properties if they are immediately followed by a `?` since `?` applies that whatever comes immediately before it is infact nullable and so not an error if not found at all.

apollo-federation/src/connectors/json_selection/parser.rs
andrewmcgivery: But also looking at his code it seems like the first param is never actually used. 🤔

apollo-federation/src/connectors/json_selection/apply_to.rs
```diff
@@ -634,6 +634,14 @@ impl ApplyToInternal for WithRange<PathList> {
                 )
             }
             PathList::Selection(selection) => selection.apply_to_path(data, vars, input_path),
+            PathList::Question(continuation) => {
```
andrewmcgivery: I should rename `continuation` to `tail` to match everywhere else.

apollo-federation/src/connectors/json_selection/apply_to.rs
```diff
@@ -561,35 +561,30 @@ impl ApplyToInternal for WithRange<PathList> {
...
+                        vec![ApplyToError::new(
+                            format!(
+                                "Property {} not found in {}",
+                                key.dotted(),
+                                json_type_name(data),
+                            ),
+                            input_path_with_key.to_vec(),
+                            key.range(),
+                        )],
+                    );
```
dylan-apollo: Making this error is not free, and we don't want to allocate the two vecs and two strings for every path step even if there wasn't an error.

Sooo, if you want to deduplicate the code, you could make this a closure?

```suggestion
                    let not_found = || (
                        None,
                        vec![ApplyToError::new(
                            format!(
                                "Property {} not found in {}",
                                key.dotted(),
                                json_type_name(data),
                            ),
                            input_path_with_key.to_vec(),
                            key.range(),
                        )],
                    );
```

And then just add the calling parens after each invocation down below, like `not_found()`

apollo-federation/src/connectors/json_selection/apply_to.rs
```diff
@@ -561,35 +561,30 @@ impl ApplyToInternal for WithRange<PathList> {
...
-                                input_path_with_key.to_vec(),
-                                key.range(),
-                            )],
-                        );
+                        return not_found;
+                    }
+
+                    if let Some(child) = data.get(key.as_str()) {
+                        tail.apply_to_path(child, vars, &input_path_with_key)
+                    } else if tail.is_question() {
```
dylan-apollo, line 615: This behavior isn't quite the same as "coerce `null` to `None`". Instead what we have is:

1. If a value is `None` (no question), emit an `ApplyToError`
2. If the value is `null` (no question), and the tail doesn't like `null`, tail emits error
3. If the value is `None` and there is a question, return `None` with no `ApplyToError`
4. If the value is `null` and there is a question, return `None` with no `ApplyToError`

If this is our intention, that `?` coerces `null` to `None` _and_ suppresses any errors about missing keys, then that's fine, we should just be super clear about it in the docs and make sure everyone is bought in. Personally I like it as a way to suppress warnings you don't care about, `null` or not, but it means that `?` is a little more complex than the current explanations.

apollo-federation/src/connectors/json_selection/apply_to.rs
```diff
@@ -749,6 +752,18 @@ impl ApplyToInternal for WithRange<PathList> {
...
+            PathList::Question(tail) => {
+                // Optional operation always produces nullable output
+                let result_shape = tail.compute_output_shape(
+                    input_shape,
+                    dollar_shape,
+                    named_var_shapes,
+                    source_id,
+                );
+                // Make result nullable since optional chaining can produce null
+                Shape::one([result_shape, Shape::none()], vec![])
```
dylan-apollo:
```suggestion
                Shape::one(
                    [
                        result_shape,
                        Shape::none().with_locations(self.shape_location(source_id)),
                    ],
                    self.shape_location(source_id),
                )
```

We'll want to highlight this in some cases—and some day will want a validation test like:

```graphql
type MyThing {
  aRequiredField: String!
      @connect(selection: "$?")
}
```

The question mark is an error here, that field is not allowed to be `None`. However, we're not using shapes for `selection` yet and I can't think of another scenario when we'd know `?` is invalid.

apollo-federation/src/connectors/json_selection/apply_to.rs
```diff
@@ -2966,4 +2982,465 @@ mod tests {
...
+
+    #[test]
+    fn test_optional_selection_set_with_null_data() {
+        use serde_json_bytes::json;
+
+        let data = json!({
+            "user": null
+        });
+
+        let (result, errors) = JSONSelection::parse("$.user ?{ id name }")
```
dylan-apollo, line 3581: Are spaces like that allowed everywhere? Wouldn't `$.user .name` be invalid?

I feel like we need to require there not be a space between the key and `?` to disambiguate for ` ?? `

apollo-federation/src/connectors/json_selection/apply_to.rs
```diff
@@ -2966,4 +2982,465 @@ mod tests {
...
+
+    #[test]
+    fn test_optional_method_on_null() {
+        use serde_json_bytes::json;
+
+        let data = json!({
+            "items": null
+        });
+
+        let (result, errors) = JSONSelection::parse("$.items?->first")
```
dylan-apollo, line 3429: What about `?` in arguments? Does `$->map(@?)` filter out nulls?

If we want `?` to suppress errors also, this might be challenging with method args, because I think most of them will error if a param is none.

apollo-federation/src/connectors/json_selection/parser.rs
```diff
@@ -3154,4 +3175,187 @@ mod tests {
         assert_debug_snapshot!(selection_object_path_v0_2);
         assert_debug_snapshot!(selection_array_path_v0_2);
     }
+
+    #[test]
+    fn test_optional_key_access() {
```
dylan-apollo, line 3532: Are/should these be valid?

1. `subSelection { .. }?` I don't think the `?` could ever do anything, so that _shouldn't_ be legacy, can we make sure?
2. `$("a literal")?` should maybe be legal? I don't know why you'd want `$(null)?` but there's probably some wacky edge case
3. `$?` and `$args?` probably shouldn't be special and will work fine, yeah? What if some day we let you use `$args?` in `@source` for sharable cross-connector logic?
4. Quoted properties, like `$config."funky name"?` work fine, right?
5. `@?`, also mentioned above re: function params

apollo-federation/src/connectors/json_selection/selection_set.rs
```diff
@@ -262,6 +262,10 @@ impl PathList {
                     path.range(),
                 ),
             ),
+            Self::Question(tail) => Self::Question(WithRange::new(
+                tail.apply_selection_set(document, selection_set),
+                tail.range(),
```
dylan-apollo, line 259: We should probably have a location test up in validations, yeah? So that if we want to highlight something before/after the `?` it highlights the right range and isn't off-by-one for some reason. Can't think of another way to be confident that the location parsing is correct.

apollo-federation/src/connectors/json_selection/apply_to.rs
andrewmcgivery: Yea, this is expected. And I agree... we should document this.

I originally didn't have this in my PR and saw it in @benjamn's POC code and it seemed like a reasonable addition.

apollo-federation/src/connectors/json_selection/apply_to.rs
andrewmcgivery: Curious on @benjamn's thoughts on this one.

From my understanding based on the grammar document, spaces are allowed in most places including between field identifiers and path steps.

Doing a quick test to verify this, `$.user .name` is valid... works fine!

apollo-federation/src/connectors/json_selection/apply_to.rs
andrewmcgivery: Yea, I suspect this would give weird results and you'd probably need to use `$(a ?? null)` pattern.

apollo-federation/src/connectors/json_selection/parser.rs
andrewmcgivery: 1. Adding a `?` after a subselection isn't valid, throws an error
2. Yep, `$(null)?` is valid because a literal can be followed by a path step.
3. Yep, both should work fine.
4. Yep
5. Yep

apollo-federation/src/connectors/json_selection/selection_set.rs
andrewmcgivery: Added a validate test that generates an error so we can see locations.. it also seems to generate a weird error message with internal stuff that we should review 🤔

apollo-federation/src/connectors/json_selection/apply_to.rs
benjamn: Spaces and/or `#` comments are allowed almost everywhere (except where indicated by `NO_SPACE`), as discussed in [this meta section of the README](https://github.com/apollographql/router/blob/dev/apollo-federation/src/connectors/json_selection/README.md#whitespace-comments-and-no_space).

That said, if/when we have an infix `??` operator, expressions like `a???b` definitely become challenging to parse. Although there might be multiple solutions, the only strategy I'm currently aware of / happy with involves the following changes:
* We need to ensure the `?` optional chaining token cannot be repeated consecutively, because it's logically idempotent, so `a??` or `a???` both mean the same thing as `a?`. If we never add the `??` operator, this repetition might be harmless, but we lose nothing by enforcing only one `?` at a time, and then it's easier to ensure `a???b` parses the same as `a? ?? b` or `a ? ??b` or `a ? ?? b`.
* In terms of precedence, we should always attempt to parse a single `?` first, because if we parse the `??` first, we might end up with the invalid `a ?? ? b` interpretation.
* The proposed `??` operator (and all multi-character operators, like `?!` or `==`) can never be split apart with spaces in between.

Fortunately, I believe we can enforce the no-consecutive-`?`s rule in the formal grammar, though it means the original definition of `PathStep ::= "." Key | "->" Identifier MethodArgs? | "?"` is not enough. Here's what I've worked out as a better set of grammar changes:
```diff
diff --git a/apollo-federation/src/connectors/json_selection/README.md b/apollo-federation/src/connectors/json_selection/README.md
index d9fffff0c..b07051b88 100644
--- a/apollo-federation/src/connectors/json_selection/README.md
+++ b/apollo-federation/src/connectors/json_selection/README.md
@@ -87,10 +87,11 @@ NamedSelection       ::= (Alias | "...")? PathSelection
 Alias                ::= Key ":"
 Path                 ::= VarPath | KeyPath | AtPath | ExprPath
 PathSelection        ::= Path SubSelection?
-VarPath              ::= "$" (NO_SPACE Identifier)? PathStep*
-KeyPath              ::= Key PathStep+
-AtPath               ::= "@" PathStep*
-ExprPath             ::= "$(" LitExpr ")" PathStep*
+VarPath              ::= "$" (NO_SPACE Identifier)? PathTail
+KeyPath              ::= Key PathTail
+AtPath               ::= "@" PathTail
+ExprPath             ::= "$(" LitExpr ")" PathTail
+PathTail             ::= "?"? (PathStep "?"?)*
 PathStep             ::= "." Key | "->" Identifier MethodArgs?
 Key                  ::= Identifier | LitString
 Identifier           ::= [a-zA-Z_] NO_SPACE [0-9a-zA-Z_]*
```
The key change is that we now have a `PathTail ::= "?"? (PathStep "?"?)*` production, and the `PathStep` production remains unchanged (without adding `| "?"`). I suggest taking some time to work through some examples to see how this enforces the rule.

If/when we add the infix `??` and/or `?!` operators, they will be part of the `LitExpr` family of grammar rules, so we shouldn't need any additional changes to the `PathTail`/`PathStep` rules.

A very subtle point here (which keeps the grammar simpler) is that `KeyPath` no longer enforces that it has at least one `PathStep` (previously, unlike `VarPath`, `AtPath`, and `ExprPath`, which all used `PathStep*`, the `KeyPath` production was `::= Key PathStep+`). To make this (desirable!) change safe, I need to do some nontrivial grammar/parser work to eliminate the distinction between a `NamedFieldSelection` and a single-`Key` `KeyPath`. I can do that in a separate PR, but it probably needs to land before this one.

apollo-federation/src/connectors/json_selection/parser.rs
```diff
@@ -703,6 +707,14 @@ impl PathList {
             ));
         }
 
+        // Universal optional operator: ? (note: we parse this before other operators to avoid conflicts)
+        if let Ok((suffix, question)) = ranged_span("?")(input) {
+            // Parse whatever comes after the ? normally
+            let (remainder, tail) = Self::parse_with_depth(suffix, depth)?;
+            let full_range = merge_ranges(question.range(), tail.range());
+            return Ok((remainder, WithRange::new(Self::Question(tail), full_range)));
+        }
```
benjamn: This implementation of `?` parsing will need to be a bit more involved to account for the new `PathTail` production I described in https://github.com/apollographql/router/pull/7637#discussion_r2190638385.

@andrewmcgivery I would request you let me make these changes, rather than doing it yourself. I'm sure you could figure it out, but I already know what to do.

apollo-federation/src/connectors/json_selection/parser.rs
benjamn: One pretty subtle/tricky case that should work: a single field selection followed by a `?`, as in `foo { bar? baz }`.

Historically, `foo` and `bar` and `baz` were/are parsed as `NamedFieldSelection`s, which means the path-related grammar rules don't automatically apply to them. I have a plan for unifying all the `NamedSelection` productions into a single struct that always contains a `PathSelection`, so everything can be treated as a path.

This will be a nontrivial refactoring, but it should also allow us to use single key path expressions in more places, like `a->and(b, c)`, rather than having to disambiguate paths from fields with `a->and($.b, $.c)`. It'll make everything more intuitive and cleaner-looking for developers, but it's work I need to do in a separate PR, before this lands.

benjamn (commented):
Almost everything in this PR is good and will not need to change based on the comments I left above, but I do think I need to "grab the lock" to work through some grammar/parser refactorings that will make it easier to get this `?` syntax exactly right (and prepare us for future operators like `??`).

I'll let you know when you can rebase and review the parser changes, but let's make sure we don't merge this PR before then.

apollo-federation/src/connectors/json_selection/apply_to.rs
benjamn: In case the paragraph beginning "A very subtle point…" was too much detail, here's a potential simplification: https://gist.github.com/benjamn/0d7068c7fe14b0e2f23eed46922ecea9 (note the new `NonEmptyPathTail` rule, which is also useful for `LitPath`).

apollo-federation/src/connectors/json_selection/apply_to.rs
```diff
@@ -3346,6 +3689,143 @@ mod tests {
...
+                },
+                null,
+                {
+                    "id": 3,
+                    "name": "Charlie"
+                }
+            ]
+        });
+
+        let (result, errors) = JSONSelection::parse("$.users ?{ id name }")
```
benjamn, line 3798: In my follow-up PR #7993, I will most likely change these tests to use `#[rstest]` and `JSONSelection::parse_with_spec`, and version-gate the `?` syntax to `connect/v0.3` only (open to discussion/objections, but we're also version-gating `...` to v0.3, fwiw).

No need to address that here, but if we cut a preview release with just your current PR, then `?` will temporarily be available regardless of `ConnectSpec` version (🙀).

apollo-federation/src/connectors/json_selection/mod.rs
```diff
@@ -21,3 +21,65 @@ pub(crate) use pretty::*;
...
+    #[case::arrow_method_optional           ("results?->first { name }",            Some(json!({"name": "Alice"})),                                               "0.3" )]
+    #[case::arrow_method_null_optional("nullField?->first { name }", None, "0.3")]
+    #[case::arrow_method_missing_optional("missingField?->first { name }", None, "0.3")]
+    #[case::optional_subselection           ("user: user? { firstName lastName }",  Some(json!({"user": { "firstName": "Alice", "lastName": "InChains" }})),      "0.3" )]
+    #[case::optional_subselection_short     ("user? { firstName lastName }",        Some(json!({"user": { "firstName": "Alice", "lastName": "InChains" }})),      "0.3" )]
+    fn kitchen_sink(
+        #[case] selection: &str,
+        #[case] expected: Option<Value>,
+        #[case] minimum_version: &str,
+        #[values(ConnectSpec::V0_2, ConnectSpec::V0_3)] version: ConnectSpec,
```
benjamn, line 54: This is so cool 😎

apollo-federation/src/connectors/json_selection/parser.rs
```diff
@@ -990,6 +994,14 @@ impl PathList {
             ));
         }
 
+        // Universal optional operator: ? (note: we parse this before other operators to avoid conflicts)
+        if let Ok((suffix, question)) = ranged_span("?")(input.clone()) {
```
benjamn, line 998: Here's where I'll be checking the `ConnectSpec` in #7993, probably.

apollo-federation/src/connectors/json_selection/README.md
```diff
@@ -91,7 +91,7 @@ VarPath              ::= "$" (NO_SPACE Identifier)? PathStep*
 KeyPath              ::= Key PathStep*
 AtPath               ::= "@" PathStep*
 ExprPath             ::= "$(" LitExpr ")" PathStep*
-PathStep             ::= "." Key | "->" Identifier MethodArgs?
+PathStep             ::= "." Key | "->" Identifier MethodArgs? | "?"
```
benjamn, line 94: The grammar will change slightly in #7993, but this version is not wrong (that is, it won't forbid any usages we want to support, though it might allow some we'd like to forbid, like multiple `missing???` operators).

benjamn (approved):
LGTM, with some comments about what I plan to do in #7993. Let's get this in!

Probably want to merge `dev` into this PR, wait for CI, then squash-and-merge, I guess?

PR #7640, opened 2025-06-06
Remove a few clones/duplication from Connectors startup

dylan-apollo:
Found when working on unifying header parsing 😁


lennyburdette:
seems reasonable, but the snapshot diffs are little sus (Some to None)

dylan-apollo:
> seems reasonable, but the snapshot diffs are little sus (Some to None)

@lennyburdette that's because `Connectors::batch_settings` was being forced to `Some` always, even if there was no `batch` declared in the `@connect`.

The only usage of it I've found is `connector.batch_settings.as_ref().and_then(|bs| bs.max_size)` which should have the same effect whether the whole thing is `None` or just `max_size` is `None` (as was the case before).

PR #7641, opened 2025-06-07
Add experimental "response cache" plugin

SimonSapin:
Add a new "response cache" plugin that starts as a copy of the "entity cache" plugin, but:

* With a more accurate name
* Marked (for now) experimental instead of preview
* Hidden from schema config (until it’s ready for use)
* Intended to change significantly (in later PRs) while "entity cache" stays unchanged, to avoid disrupting existing users and decouple the eventual migration from Router version upgrades

---

**Checklist**

- Tests added and passing[^4]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


SimonSapin:
I’ve split the commits to try and make review easier, so I suggest reviewing by commit.

At the moment the `configuration::tests::schema_generation` test fails because https://github.com/apollographql/router/commit/7117b2432cbe7040ba4828a2ee4e225c613831cc effectively broke `HIDDEN_FROM_CONFIG_JSON_SCHEMA`. I don’t know what `pattern_properties` is supposed to do, but stuff intended to be hidden ends up being visible in the config schema. @rregitsky, what was the motivation for that commit?

apollo-router/tests/integration/mod.rs
```diff
@@ -15,6 +15,8 @@ mod lifecycle;
 mod operation_limits;
 mod operation_name;
 mod query_planner;
+#[cfg(any(not(feature = "ci"), all(target_arch = "x86_64", target_os = "linux")))]
```
bnjjj, line 21: Maybe add a comment to explain this condition

PR #7645, opened 2025-06-09
Share `SourceName` model/parsing with runtime

dylan-apollo:
This PR is the first step toward sharing more parsing/model code between validations & runtime. It comes with a few key improvements:

1. There are two fewer places where argument definitions can drift, since `@connect(source:)` and `@source(name:)` parsing is now shared instead of redefined.
2. Using the dedicated `SourceName` type instead of `String` everywhere will make it easier to find usages and prevent potential bugs (putting the wrong string in the wrong place)
3. `SourceName` uses `Arc<str>` instead of `String` which will reduce of ton of little clones all over the place. This isn't expected to meaningfully impact performance, though.
4. While walking through the code, I was able to rethink/simplify a few things, like getting rid of the map of connectors.

This first step also reveals a few insights about this refactor effort:

1. We are not going to be able to use the same lifetime/copy-based approach we have been in validation, since the runtime can't keep track of those lifetimes. Instead, we should be using reference counting (which has to be `Arc` since runtime is asynchronous) where things are shared. This makes for a bit more effort in rewriting and _may_ have a performance implication for validations of very large schemas.
2. There's probably a lot of other tech debt lurking that we're going to uncover while unifying parsing.
3. It's going to take a lot of work just to figure out the right _order_ to tackle these things in. This PR is actually the third branch as I followed a rabbit hole, realized something else had to be ported first, then continued.

### DIFF
--- a/apollo-federation/src/connectors/mod.rs
+++ b/apollo-federation/src/connectors/mod.rs
@@ -57,6 +57,7 @@ pub use self::models::HTTPMethod;
 pub use self::models::HeaderSource;
 pub use self::models::HttpJsonTransport;
 pub use self::models::MakeUriError;
+pub use self::models::SourceName;
 pub use self::spec::schema::ConnectBatchArguments;
 use crate::schema::position::ObjectFieldDefinitionPosition;
 use crate::schema::position::ObjectOrInterfaceFieldDefinitionPosition;
@@ -66,7 +67,7 @@ use crate::schema::position::ObjectOrInterfaceFieldDirectivePosition;
 pub struct ConnectId {
     pub label: String,
     pub subgraph_name: String,
-    pub source_name: Option<String>,
+    pub source_name: Option<SourceName>,
     pub(crate) directive: ConnectorPosition,
 }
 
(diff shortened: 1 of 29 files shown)


apollo-federation/src/connectors/mod.rs
```diff
@@ -81,8 +82,12 @@ impl ConnectId {
     }
 
     pub fn subgraph_source(&self) -> String {
-        let source = format!(".{}", self.source_name.as_deref().unwrap_or(""));
-        format!("{}{}", self.subgraph_name, source)
+        let source = self
+            .source_name
+            .as_ref()
+            .map(SourceName::as_str)
+            .unwrap_or("");
```
naomijub: `unwrap_or_default`

apollo-federation/src/connectors/models/source.rs
```diff
@@ -0,0 +1,238 @@
...
+    node: Option<Arc<Node<Value>>>,
+}
+
+impl SourceName {
+    /// Create a `SourceName`, but without checking most validations.
+    ///
+    /// Useful for speeding up parsing at runtime & tests.
+    ///
+    /// For enhanced validity checks, use [`SourceName::from_directive`]
+    pub(crate) fn from_directive_permissive(
```
lennyburdette, line 38: Is this an opportunity for a typestate-style pattern, like Valid<SourceName> or SourceName<Initial | Validated>? I put no thought into this idea, just pattern matching with apollo-compiler and the new composition work

apollo-federation/src/connectors/mod.rs
dylan-apollo: Clippy pedantic would have caught that. Guess my internal Clippy isn't up to the task 😓

apollo-federation/src/connectors/models/source.rs
dylan-apollo: I worry it would just be annoying since we'd want the router to accept _either_ type, and wouldn't want to carry generics everywhere 🤔. Something to keep in mind though as we get further into code sharing, maybe there's a natural pattern that wouldn't take too much boilerplate.

PR #7646, opened 2025-06-09
Connectors `filter` method

andrewmcgivery:
Connectors `filter` method

---

**Checklist**

- Tests added and passing[^4]
    - [x] Unit Tests
    - [x] Integration Tests
    - [x] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-federation/src/connectors/json_selection/methods/public/filter.rs
```diff
@@ -0,0 +1,293 @@
...
+            (Some(JSON::Array(output)), errors)
+        }
+    } else {
+        // For non-array inputs, treat as single-element array
+        // Apply filter condition and return either [value] or []
+        let (condition_result, mut condition_errors) =
+            first_arg.apply_to_path(data, vars, input_path);
+
+        match condition_result {
+            Some(JSON::Bool(true)) => (Some(JSON::Array(vec![data.clone()])), condition_errors),
```
naomijub, line 92: if this is not an array, why treat it as an array?

Languages like Rust, OCaml, Clojure, Haskell have pretty interesting ways of dealing with non-array filtering. Maybe we should consider a broader approach to this case and return the value itself or none?

https://doc.rust-lang.org/std/option/enum.Option.html#method.filter

andrewmcgivery: We do the same thing in map:  https://github.com/apollographql/router/blob/dev/apollo-federation/src/connectors/json_selection/methods/public/map.rs#L71

So I largely did this for consistency. Happy to re-consider this if @lennyburdette and/or @benjamn have some opinions one way or another.

naomijub: `map` seems even better to support transforming a data-type into itself

benjamn: How about `->flatMap` then?

apollo-federation/src/connectors/json_selection/methods/public/filter.rs
```diff
@@ -0,0 +1,293 @@
...
+        return (
+            None,
+            vec![ApplyToError::new(
+                format!("Method ->{} requires one argument", method_name.as_ref()),
+                input_path.to_vec(),
+                method_name.range(),
+            )],
+        );
+    };
+    let Some(first_arg) = args.args.first() else {
```
benjamn:
```suggestion
    let Some(first_arg) = method_args.and_then(|args| args.args.first()) else {
```
I think it's worth combining these two cases, since `args` isn't used later, and the errors are the same.

apollo-federation/src/connectors/json_selection/methods/public/filter.rs
```diff
@@ -0,0 +1,293 @@
...
+        }
+
+        if has_non_boolean_error {
+            (None, errors)
+        } else {
+            (Some(JSON::Array(output)), errors)
+        }
+    } else {
+        // For non-array inputs, treat as single-element array
+        // Apply filter condition and return either [value] or []
```
benjamn, line 87: This policy for non-array inputs makes intuitive sense to me, and should be straightforward to reflect at the shape level.

Another method I've been considering in this design space is `condition->then(expression)`, which would return `expression` when `condition` is true and `None` when `condition` is false. With the proposed `->filter` behavior for non-array inputs, that's equivalent to `expression->filter(condition)->first`, which has the advantage that it continues to work if you replace `expression` with an array, and you can use the expression as `@` within the `condition` expression.

apollo-federation/src/connectors/json_selection/methods/public/filter.rs
```diff
@@ -0,0 +1,293 @@
...
+                    output.push(element.clone());
+                }
+                Some(JSON::Bool(false)) | None => {
+                    // Condition is false or errored, exclude the element
+                }
+                Some(_) => {
+                    // Condition returned a non-boolean value, this is an error
+                    has_non_boolean_error = true;
+                    errors.push(ApplyToError::new(
+                        "Filter condition must return a boolean value".to_string(),
```
benjamn:
```suggestion
                        format!(
                            "->{} condition must return a boolean value",
                            method_name.as_ref()
                        ),
```
Let's keep the method name configurable/flexible.

apollo-federation/src/connectors/json_selection/methods/public/filter.rs
```diff
@@ -0,0 +1,293 @@
...
+        let (condition_result, mut condition_errors) =
+            first_arg.apply_to_path(data, vars, input_path);
+
+        match condition_result {
+            Some(JSON::Bool(true)) => (Some(JSON::Array(vec![data.clone()])), condition_errors),
+            Some(JSON::Bool(false)) => (Some(JSON::Array(vec![])), condition_errors),
+            Some(_) => {
+                // Condition returned a non-boolean value, this is an error
+                condition_errors.push(ApplyToError::new(
+                    "Filter condition must return a boolean value".to_string(),
```
benjamn:
```suggestion
                    format!(
                        "->{} condition must return a boolean value",
                        method_name.as_ref()
                    ),
```

apollo-federation/src/connectors/json_selection/methods/public/filter.rs
```diff
@@ -0,0 +1,293 @@
...
+        ShapeCase::Array { prefix: _, tail } => {
+            // Filter preserves the element types but may reduce the count
+            // We can't know statically how many elements will pass the filter,
+            // so we return an array with the same element type but no fixed prefix
+            Shape::list(tail.clone(), input_shape.locations)
+        }
+        _ => {
+            // For non-array inputs, we return a list that may contain 0 or 1 elements
+            // of the same type as the input
+            Shape::list(input_shape.any_item([]), input_shape.locations)
```
benjamn: Possibly worth noting: when we know `input_shape` is not an array, `input_shape.any_item([])` simplifies to just `input_shape`, which I think aligns nicely with the `->filter` behavior for non-array inputs.

apollo-federation/src/connectors/json_selection/methods/public/filter.rs
```diff
@@ -0,0 +1,293 @@
...
+    // Validate that the condition evaluates to a boolean
+    if !matches!(condition_shape.case(), ShapeCase::Bool(_)) {
+        return Shape::error(
+            "Filter condition must return a boolean value".to_string(),
+            method_name.shape_location(source_id),
+        );
+    }
+
+    match input_shape.case() {
+        ShapeCase::Array { prefix: _, tail } => {
```
benjamn: We are going to want to filter the `prefix` as well, I think! Then we can filter tuples with known length and element shapes the same way we filter arbitrary-length lists.

benjamn (changes requested):
This looks very close to complete to my eyes. Since you asked, I like the policy about non-array inputs.

A few specific thoughts / suggested changes below…

apollo-federation/src/connectors/json_selection/methods/public/filter.rs
andrewmcgivery: To clarify, is this a change suggestion or just a note? 😅

apollo-federation/src/connectors/json_selection/methods/public/filter.rs
andrewmcgivery: Can you expand on this? I'm not understanding 😅

apollo-federation/src/connectors/json_selection/methods/public/filter.rs
benjamn: Just a note!

apollo-federation/src/connectors/json_selection/methods/public/filter.rs
benjamn: The `prefix` is a `Vec<Shape>` that represents known shapes of some prefix of the elements of the array.

Some `ShapeCase::Array` shapes are just a `prefix` (that is, `prefix.len() > 0` and `tail` is `None`), and we sometimes call those arrays _tuples_ for clarity. Some are just a `tail` shape, with no fixed `prefix` (that is, `prefix.is_empty()`), and those are your typical `List<T>` lists of any length and only one element shape, repeated. Some arrays have both a `prefix` and a `tail` shape, so the first few elements have statically known shapes but there's a dynamic tail shape for the rest of the array, however long it is. The empty array has `prefix.is_empty()` and `tail.is_none()`.

I suppose there's a distant future where we might also want to model the `suffix` of the array, with a dynamic middle between the `prefix` and `suffix`, but I'm waiting to see a case where that would really help. Pretty-printing such a shape would look something like `[Int, ...List<String>, null]`, where `vec![Shape::int()]` is the `prefix`, `String` is the `tail`/`rest`/`middle`, and `vec![Shape::null()]` is the `suffix`.

**Edit:** Rethinking this next part:

The goal here (in the shape processing code) is to imagine what the `->filter` operation would do to all parts of a given array representation, then produce a new array representation that accurately describes the shape of the filtered result at runtime. ~So you basically filter the `prefix` vector by itself and then map the `tail` (like you're already doing) and make a new `Shape::array(prefix, tail, locations)` out of that.~

benjamn: Ok, I think I was overestimating our ability to filter the `prefix` without runtime information. I think the safest result shape we can report is the one you use for non-array inputs:
```rust
Shape::list(input_shape.any_item([]), input_shape.locations)
```
The `.any_item` bit will ensure all the shapes from the `prefix` and the `tail` are combined into one union shape, which then becomes the element shape of the resulting list. This handles your array case as well, because `array.any_item([])` is just the `tail` shape of the array if the array has only a `tail`.

It's not great that we go from `[P1, P2, ...List<T>]` to `List<One<P1, P2, T>>` (losing ordering), but I think that may be the best `->filter` can promise, in terms of shapes.

andrewmcgivery: Is your suggestion to always return `Shape::list(input_shape.any_item([]), input_shape.locations)` regardless of the  `input_shape`?

benjamn: If we wanted to get fancy, when `input_shape` is definitely a non-array, we could return
```rust
// One<[], [InputShape]> rather than List<InputShape>:
Shape::one([
  Shape::tuple([], input_shape.locations),
  Shape::tuple([input_shape], input_shape.locations),
], [])
```
However, I think `List<InputShape>` is technically accurate (not a lie), just less precise as a description of what will happen at runtime, so we could safely report `Shape::list(input_shape.any_item([]), input_shape.locations)` in all cases, though we would need to be on the lookout for DX surprises (common idioms that should work but don't shape-check in our system, due to the loss of shape precision with `List<InputShape>`).

andrewmcgivery: Done!

apollo-federation/src/connectors/json_selection/methods/public/filter.rs
```diff
@@ -0,0 +1,280 @@
...
+    // Compute the shape of the filter condition argument
+    let condition_shape = first_arg.compute_output_shape(
+        input_shape.clone(),
+        dollar_shape,
+        named_var_shapes,
+        source_id,
+    );
+
+    // Validate that the condition evaluates to a boolean
+    if !matches!(condition_shape.case(), ShapeCase::Bool(_)) {
```
benjamn:
```suggestion
    // Validate that the condition evaluates to a boolean or
    // something that could become a boolean
    if !(
        matches!(condition_shape.case(), ShapeCase::Bool(_)) ||
        // This allows Unknown and Name shapes, which can produce boolean
        // values at runtime, without any runtime errors.
        condition_shape.accepts(&Shape::unknown([]) {
```
I worry this may be slightly too restrictive, as the condition might produce a placeholder-like shape that _could_ evaluate to a boolean at runtime, like `Unknown` or a `Name`.

benjamn (commented):
Thanks for addressing my feedback. I'm removing "changes requested," but I have one more idea about the `condition` boolean check (below).

apollo-federation/src/connectors/json_selection/methods/public/filter.rs
```diff
@@ -0,0 +1,285 @@
...
+
+            match applied_opt {
+                Some(JSON::Bool(true)) => {
+                    output.push(element.clone());
+                }
+                Some(JSON::Bool(false)) | None => {
+                    // Condition is false or errored, exclude the element
+                }
+                Some(_) => {
+                    // Condition returned a non-boolean value, this is an error
```
benjamn, line 66: Just want to point out that this works at runtime because these are JSON values and there's no `Unknown` or `Name` type in JSON, only in the shape system.

benjamn (approved):

PR #7649, opened 2025-06-10
Add gte method support for JSON selection in connectors

andrewmcgivery:
Add gte method support for JSON selection in connectors

---

**Checklist**

- Tests added and passing[^4]
    - [x] Unit Tests
    - [x] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-federation/src/connectors/json_selection/methods/public/gte.rs
+++ b/apollo-federation/src/connectors/json_selection/methods/public/gte.rs
@@ -30,6 +30,7 @@ fn gte_method(
 ) -> (Option<JSON>, Vec<ApplyToError>) {
     if let Some(first_arg) = method_args.and_then(|args| args.args.first()) {
         let (value_opt, arg_errors) = first_arg.apply_to_path(data, vars, input_path);
+        let mut apply_to_errors = arg_errors;
         // We have to do this because Value doesn't implement PartialOrd
         let matches = value_opt.is_some_and(|value| {
             match (data, &value) {
@@ -44,11 +45,24 @@ fn gte_method(
                 // Null comparisons (null == null)
                 (JSON::Null, JSON::Null) => true,
                 // Mixed types or uncomparable types (including arrays and objects) return false
-                _ => false,
+                _ => {
+                    apply_to_errors.push(ApplyToError::new(
+                        format!(
+                            "Method ->{} can directly compare numbers, strings, booleans, and null. Either a mix of these was provided or something else such as an array or object. Found: {} >= {}",
+                            method_name.as_ref(),
+                            data,
+                            value
+                        ),
+                        input_path.to_vec(),
+                        method_name.range(),
+                    ));
+
+                    false
+                }
             }
         });
 
-        return (Some(JSON::Bool(matches)), arg_errors);
+        return (Some(JSON::Bool(matches)), apply_to_errors);
     }
     (
         None,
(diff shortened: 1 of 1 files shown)


naomijub (approved):
LGTM

apollo-federation/src/connectors/json_selection/methods/public/gte.rs
```diff
@@ -0,0 +1,300 @@
...
+
+impl_arrow_method!(GteMethod, gte_method, gte_shape);
+/// Returns true if the applied to value is greater than or equal to the argument value.
+/// Simple examples:
+///
+/// $->echo(3)->gte(3)       results in true
+/// $->echo(4)->gte(3)       results in true
+/// $->echo(2)->gte(3)       results in false
+/// $->echo("a")->gte("b")   results in false
+/// $->echo("c")->gte("b")   results in true
```
benjamn:
```suggestion
/// $(3)->gte(3)       results in true
/// $(4)->gte(3)       results in true
/// $(2)->gte(3)       results in false
/// $("a")->gte("b")   results in false
/// $("c")->gte("b")   results in true
```
Can we use less `->echo` in this/future examples?

Here's another syntax that works now (#6784):
```suggestion
/// $(3->gte(3))       results in true
/// $(4->gte(3))       results in true
/// $(2->gte(3))       results in false
/// $("a"->gte("b"))   results in false
/// $("c"->gte("b"))   results in true
```

apollo-federation/src/connectors/json_selection/methods/public/gte.rs
```diff
@@ -0,0 +1,300 @@
...
+fn gte_method(
+    method_name: &WithRange<String>,
+    method_args: Option<&MethodArgs>,
+    data: &JSON,
+    vars: &VarsWithPathsMap,
+    input_path: &InputPath<JSON>,
+) -> (Option<JSON>, Vec<ApplyToError>) {
+    if let Some(MethodArgs { args, .. }) = method_args {
+        if let [arg] = args.as_slice() {
+            let (value_opt, arg_errors) = arg.apply_to_path(data, vars, input_path);
```
benjamn:
```suggestion
    if let Some(first_arg) = method_args.and_then(|args| args.args.first()) {
        let (value_opt, arg_errors) = first_arg.apply_to_path(data, vars, input_path);
```
We do this style elsewhere, and I like it.

apollo-federation/src/connectors/json_selection/methods/public/gte.rs
```diff
@@ -0,0 +1,300 @@
...
+                r#"
+                    result: value->gte({"a": 1})
+                "#
+            )
+            .apply_to(&json!({ "value": {"a": 1, "b": 2} })),
+            (
+                Some(json!({
+                    "result": false,
+                })),
+                vec![],
```
benjamn: Should there be a runtime `ApplyToError` in this `vec![]`, since the comparison did not exactly succeed?

benjamn (commented):
Looking good!

I presume you also have (future PR) plans for `->lte` and `->gt` and `->lt`?

andrewmcgivery:
> I presume you also have (future PR) plans for ->lte and ->gt and ->lt?

Yep, I am doing separate PRs per method to keep them small, focused, and easy to review/merge. :)

apollo-federation/src/connectors/json_selection/methods/public/gte.rs
andrewmcgivery: Good question.

Currently I wrote it to just return `false` if it hits non string/bool/null (aka objects and arrays).

Perhaps we could return false _and_ return an `ApplyToError` basically saying that comparisons should not be done against objects or arrays?

andrewmcgivery: Updated to return an ApplyToError in this situation!

apollo-federation/src/connectors/json_selection/methods/public/gte.rs
```diff
@@ -44,11 +45,24 @@ fn gte_method(
                 // Null comparisons (null == null)
                 (JSON::Null, JSON::Null) => true,
                 // Mixed types or uncomparable types (including arrays and objects) return false
-                _ => false,
+                _ => {
+                    apply_to_errors.push(ApplyToError::new(
+                        format!(
+                            "Method ->{} can directly compare numbers, strings, booleans, and null. Either a mix of these was provided or something else such as an array or object. Found: {} >= {}",
```
benjamn, line 51: We may at some point want to devise a rule for truncating very long string representations of JSON data in error messages like these.

apollo-federation/src/connectors/json_selection/methods/public/gte.rs
```diff
@@ -243,56 +257,71 @@ mod tests {
...
+            result.0,
+            Some(json!({
+                "result": false,
+            })),
+        );
+        assert!(!result.1.is_empty());
+        assert!(
+            result.1[0]
+                .message()
+                .contains("Method ->gte can directly compare numbers, strings, booleans, and null. Either a mix of these was provided or something else such as an array or object. Found: [1,2,3] >= [1,2]")
```
benjamn, line 278: This test case actually makes me want to support array comparison (using lexicographic ordering), as long as all the array elements meet the other requirements.

However, we can always go from forbidding arrays to allowing them, in the future.

PR #7650, opened 2025-06-10
Remove unenforceable test

garypen:
The relative priority test is testing behaviour that is better tested at a lower level. The behaviour it was intended to test is already tested in ageing_priority_queue::test_priorties.


carodewig:
LGTM - sorry I missed this yesterday!

PR #7665, opened 2025-06-11
Batching is not supported for GET requests

goto-bus-stop:
When sending an operation as a GET request, like `/graphql?query={ user { id } }`, you cannot send a batch request.

We use `serde_urlencoded` to parse such requests, and the behaviour is like this:
```rust
use serde_json::Value;
use serde_json::json;

assert_eq!(
    serde_urlencoded::from_str::<Value>("query=graphql").unwrap(),
    json!({ "query": "graphql" }),
);
assert_eq!(
    serde_urlencoded::from_str::<Value>("query=graphql&query=other").unwrap(),
    json!({ "query": "other" }),
);
assert_eq!(
    serde_urlencoded::from_str::<Value>("query[]=graphql&query[]=other").unwrap(),
    json!({ "query[]": "other" }),
);
```

That is, there is no way for `batch_from_urlencoded_query` to ever parse a `serde_json::Value::Array` from the query string.

That means that parsing a batch from a query string is essentially a no-op, and I suggest we remove the code for it.

---

**Checklist**

- Tests added and passing[^4]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

There were already no tests.

**Notes**

### DIFF
--- a/apollo-router/src/graphql/request.rs
+++ b/apollo-router/src/graphql/request.rs
@@ -161,20 +161,6 @@ impl Request {
         seed.deserialize(&mut de)
     }
 
-    /// Convert encoded URL query string parameters (also known as "search
-    /// params") into a GraphQL [`Request`].
-    ///
-    /// An error will be produced in the event that the query string parameters
-    /// cannot be turned into a valid GraphQL `Request`.
-    pub(crate) fn batch_from_urlencoded_query(
-        url_encoded_query: String,
-    ) -> Result<Vec<Request>, serde_json::Error> {
-        let value: Value = serde_urlencoded::from_bytes(url_encoded_query.as_bytes())
-            .map_err(serde_json::Error::custom)?;
-
-        Request::process_query_values(value)
-    }
-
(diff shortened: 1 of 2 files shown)


garypen:
It seems like no-one is trying to use this (since they can't), but now I wonder if this is something we should be supporting. Even though we couldn't parse out an `Array`, we could parse out a `Vec` of queries...

So my question becomes: is this a bug or a feature that nobody wants? If nobody wants it, then we should just remove it (as per this PR).

goto-bus-stop:
It's not documented as a feature anywhere that I can find. Supporting this would probably require inventing a new format (eg: `?batch[]={JSON_HERE}&batch[]={JSON_HERE}`. I believe nobody wants it.

garypen:
> It's not documented as a feature anywhere that I can find. Supporting this would probably require inventing a new format (eg: `?batch[]={JSON_HERE}&batch[]={JSON_HERE}`. I believe nobody wants it.

I'm happy to see the code removed since I agree it's unlikely that anyone wants this.

PR #7670, opened 2025-06-11
Port shallow type addition for composition merge step

tninesling:
# Overview

Ports `Merger.addTypesShallow` from JS federation ([primary source](https://github.com/apollographql/federation/blob/31104c6fc5319fb8140d697776249e9ee0ca9285/composition-js/src/merging/merge.ts#L780-L828)). This involves adding a new check for determining whether a given type is an interface object type; some new utilities in `TypeDefinitionPosition` for adding empty, named positions to the merged schema; and some rather complicated updates to the error reporter (more details below).

## Mismatch reporter updates

The logic to report unexpected type mismatches calls [MismatchReporter.reportMismatchError](https://github.com/apollographql/federation/blob/31104c6fc5319fb8140d697776249e9ee0ca9285/composition-js/src/merging/reporter.ts#L15-L36), which has been ported along with the underlying [MismatchReporter.reportMismatch](https://github.com/apollographql/federation/blob/31104c6fc5319fb8140d697776249e9ee0ca9285/composition-js/src/merging/reporter.ts#L145-L197). These functions assemble error messages in a fairly convoluted way. I would love to refactor them, but I'm too scared to do so while we don't have merge tests running. So, I've kept these as close to the JS source as possible for now.

## Future directions

1. In the short-term, we'll need to decide on an implementation of `ASTNode`, which is collected and attached to errors in the MismatchReporter. The new mismatch reporting utilities have a generic parameter which is a placeholder for whatever that struct ends up being, and I have marked the relevant codepoints with TODO comments so we can fill those in at a later date.
2. Once we get merge tests ported over, we should seriously consider refactoring this error construction logic. It shouldn't take this many nested closures to create error strings with some subgraphs in them.

---

**Checklist**

- Tests added and passing[^4]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-federation/src/merger/merger.rs
```diff
@@ -335,7 +348,125 @@ impl Merger {
...
+                    if expects_interface
+                        && !matches!(previous, TypeDefinitionPosition::Interface(_))
+                    {
+                        mismatched_types.insert(pos.clone());
+                    }
+                    if !expects_interface && previous != pos {
+                        mismatched_types.insert(pos.clone());
+                    }
+                } else {
+                    pos.pre_insert(&mut self.merged);
```
duckki, line 382: [The JS code](https://github.com/apollographql/federation/blob/31104c6fc5319fb8140d697776249e9ee0ca9285/composition-js/src/merging/merge.ts#L800) seems to convert the interface-object's kind to interface as it is added. So, don't we need to change `pos` to interface variant?

apollo-federation/src/merger/merger.rs
```diff
@@ -335,7 +348,125 @@ impl Merger {
...
+                    pos.insert_empty(&mut self.merged);
+                }
+            }
+        }
+
+        for mismatched_type in mismatched_types.iter() {
+            self.report_mismatched_type_definitions(mismatched_type, &types_with_interface_object);
+        }
+
+        for type_ in types_with_interface_object.iter() {
```
duckki, line 395: JS code has some friendly code comments worth porting over here:
```js
    // Most invalid use of @interfaceObject are reported as a mismatch above, but one exception is the
    // case where a type is used only with @interfaceObject, but there is no corresponding interface
    // definition in any subgraph.
```

duckki (approved):
Thanks for implementing`report_mismatch_error` :)
Looks good overall. I just a few minor suggestions.

apollo-federation/src/merger/merger.rs
tninesling: Good catch, fixed in [1d11f87](https://github.com/apollographql/router/pull/7670/commits/1d11f87a7f31d71d19df88958dc8f1f1006fb90e)

apollo-federation/src/merger/merger.rs
tninesling: Ported over in [0999bed](https://github.com/apollographql/router/pull/7670/commits/0999bed38e4e3ff30648265964461c4a43f63365)

PR #7671, opened 2025-06-11
add support for $env in connectors

lennyburdette:
---

**Checklist**

- Tests added and passing[^4]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-federation/src/connectors/models.rs
+++ b/apollo-federation/src/connectors/models.rs
@@ -61,6 +61,8 @@ pub struct Connector {
     pub request_headers: HashSet<String>,
     /// The request or response headers referenced in the connectors response mapping
     pub response_headers: HashSet<String>,
+    /// The environment variables referenced in the connector
+    pub env: HashSet<String>,
 
     pub batch_settings: Option<ConnectBatchArguments>,
 
@@ -196,7 +198,7 @@ impl Connector {
             .iter()
             .map(|var_ref| var_ref.namespace.namespace)
             .collect();
-        let request_headers = extract_header_references(request_references);
+        let request_headers = extract_header_references(&request_references);
 
         // Calculate which variables and headers are in use in the response (including errors.message and errors.extensions)
         let response_references: HashSet<VariableReference<Namespace>> = connect
@@ -208,7 +210,9 @@ impl Connector {
             .iter()
             .map(|var_ref| var_ref.namespace.namespace)
             .collect();
-        let response_headers = extract_header_references(response_references);
+        let response_headers = extract_header_references(&response_references);
+
+        let env = extract_env_references(&request_references, &response_references);
 
         // Last couple of items here!
         let entity_resolver = determine_entity_resolver(
@@ -241,6 +245,7 @@ impl Connector {
             response_variables,
             request_headers,
             response_headers,
+            env,
             batch_settings,
             error_settings,
         })
@@ -370,7 +375,7 @@ fn determine_entity_resolver(
 
 /// Get any headers referenced in the variable references by looking at both Request and Response namespaces.
 fn extract_header_references(
-    variable_references: HashSet<VariableReference<Namespace>>,
+    variable_references: &HashSet<VariableReference<Namespace>>,
 ) -> HashSet<String> {
     variable_references
         .iter()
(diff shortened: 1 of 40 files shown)


apollo-federation/src/connectors/models.rs
```diff
@@ -61,6 +61,8 @@ pub struct Connector {
     pub request_headers: HashSet<String>,
     /// The request or response headers referenced in the connectors response mapping
     pub response_headers: HashSet<String>,
+    /// The environment variables referenced in the connector
+    pub env: HashSet<String>,
```
dylan-apollo, line 65: I don't like how big/flat this struct is getting. Feels like maybe `request_variables` and `response_variables` should just contain all of the variables (not just namespaces) and we should have a resolver function which collects all required variables.

apollo-federation/src/connectors/models.rs
```diff
@@ -390,6 +395,24 @@ fn extract_header_references(
...
+    request_references
+        .iter()
+        .chain(response_references.iter())
+        .flat_map(|var_ref| {
+            if var_ref.namespace.namespace == Namespace::Env {
+                var_ref.selection.keys().cloned().collect()
+            } else {
+                Vec::new()
+            }
+        })
```
dylan-apollo:
```suggestion
        .filter(|var_ref| var_ref.namespace.namespace == Namespace::Env)
        .flat_map(|var_ref| var_ref.selection.keys())
        .cloned()
```

Saves allocating the intermediate `Vec` which we don't need

apollo-federation/src/connectors/validation/expression.rs
```diff
@@ -52,6 +52,8 @@ static RESPONSE_SHAPE: LazyLock<Shape> = LazyLock::new(|| {
     )
 });
 
+static ENV_SHAPE: LazyLock<Shape> = LazyLock::new(|| Shape::dict(Shape::string([]), []));
```
dylan-apollo:
```suggestion
fn env_shape() -> Shape {
    Shape::dict(Shape::string([]), [])
}
```

Since we're just cloning anyway, this reads as simpler to me then a `LazyLock`. Would be nice if we had `const` shapes, though

apollo-router/src/plugins/connectors/make_requests/request_merger.rs
```diff
@@ -168,4 +172,18 @@ impl MappingContextMerger<'_> {
         }
         self
     }
+
+    pub(crate) fn env(mut self, env_vars_used: &HashSet<String>) -> Self {
+        if self.variables_used.contains(&Namespace::Env) {
+            let env_vars: Map<ByteString, Value> = std::env::vars()
+                .filter_map(|(key, value)| {
+                    env_vars_used
+                        .contains(&key)
```
dylan-apollo: Are environment variables case-sensitive on every OS? If not, we probably need to look them up one-by-one and let the OS deal with it.

I also wonder if it's more efficient to pull individual vars anyway vs loading everything into a new map and filtering down 🤔.

apollo-router/src/plugins/connectors/tests/variables.rs
```diff
@@ -0,0 +1,41 @@
...
+#[tokio::test]
+async fn test_env_var() {
+    let mock_server = MockServer::start().await;
+    Mock::given(method("GET"))
+        .and(path("/"))
+        .respond_with(ResponseTemplate::new(200).set_body_json(serde_json::json!("hi")))
+        .mount(&mock_server)
+        .await;
+
+    unsafe { std::env::set_var("CONNECTORS_ENV_VAR_TESTING", "environment variable value") };
```
dylan-apollo: I feel like setting env vars in tests can often cause flakey issues—but that's probably only when multiple tests depend on the same var? Only alternative is to give us some sort of mocking interface  which sounds like a lot of work, so we should just keep an eye out for this test (or future ones with `$env`) flaking.

dylan-apollo (approved):
Design looks good, a few questions/comments but I _think_ no blockers

apollo-router/src/plugins/connectors/make_requests/request_merger.rs
lennyburdette: good call, i don't know the answer but we should definitely look up vars by name.

apollo-router/src/plugins/connectors/tests/variables.rs
lennyburdette: i made the var name extra long to hopefully avoid collisions with future tests, i could make it even longer/more specific

apollo-federation/src/connectors/models.rs
lennyburdette: https://apollographql.atlassian.net/browse/CNN-859

PR #7674, opened 2025-06-12
Refactor the router service to be a tower of layers

goto-bus-stop:
This PR is just for information for now. I think we could/should already apply this on dev, but it can also just serve as an example of the things we can gain by moving towards a more towery world.

The router service all but disappears: it's now just a ServiceBuilder with about 6 layers.

The biggest gain is that functionality is separated: previously, PQs and APQs and Parsing were intertwined because of the manual error propagation (now this is done by short-circuiting requests in each layer). Batching was mixed in with a lot of other stuff when parsing requests and formatting respones, now all (ingress-side) batching is done in its own layer in its own module. The router service used to juggle multiple request and response types (router request, graphql request, and supergraph request, router response, supergraph response), including sometimes misaligned ones (eg. part of the service operated on supergraph requests but returned router responses). Now there is a single conversion point for each request and response type. Subjectively, I think this makes the router service _much_ easier to understand and modify.


apollo-router/src/services/router/batching.rs
```diff
@@ -0,0 +1,421 @@
...
+                status: StatusCode::BAD_REQUEST,
+                extension_code: "BATCHING_NOT_ENABLED".to_string(),
+                extension_details,
+            })
+        }
+    }
+
+    /// Turns a parsed batch of queries into multiple separate requests that can flow through the
+    /// rest of the pipeline.
+    fn build_batch_requests(
```
goto-bus-stop, line 296: This contains the batching parts of what used to be `RouterService::translate_requests`.

apollo-router/src/services/router/batching.rs
```diff
@@ -0,0 +1,421 @@
...
+            })
+        })
+    }
+}
+
+impl<S> BatchingService<S>
+where
+    S: Service<RouterRequest, Response = RouterResponse, Error = BoxError> + Clone,
+{
+    fn parse_batch_request(
```
goto-bus-stop, line 233: This contains the batching parts of what used to be `RouterService::get_graphql_requests`.

apollo-router/src/services/router/service.rs
```diff
@@ -88,38 +86,54 @@ static ACCEL_BUFFERING_HEADER_NAME: HeaderName = HeaderName::from_static("x-acce
...
     ) -> Self {
-        let supergraph_service: supergraph::BoxCloneService =
-            ServiceBuilder::new().buffered().service(sgb).boxed_clone();
+        // Some of the layers in the stack are wrapping previous implementations that are called
+        // layers, but are not tower layers at all.
+        let apq_layer = Arc::new(apq_layer);
+        let query_analysis_layer = Arc::new(query_analysis_layer);
+
+        let service = ServiceBuilder::new()
+            .layer(DisplayRouterRequestLayer)
```
goto-bus-stop, line 118: This layer is re-ordered: previously, it happened in the middle of other batching/router service code. It is not possible to split up batching from the router service without changing the ordering of this feature.

It must be before the batching layer, because it has to log the entire batch request as one, not as multiple single-query requests.

It reads the body, but only if it actually needs to.

apollo-router/src/services/router/service.rs
```diff
@@ -88,38 +86,54 @@ static ACCEL_BUFFERING_HEADER_NAME: HeaderName = HeaderName::from_static("x-acce
...
+    type Future = BoxFuture<'static, Result<Self::Response, Self::Error>>;
+
+    fn poll_ready(&mut self, cx: &mut std::task::Context<'_>) -> Poll<Result<(), Self::Error>> {
+        self.service.poll_ready(cx)
+    }
+
+    fn call(&mut self, req: RouterRequest) -> Self::Future {
+        self.service.call(req)
     }
 }
```
goto-bus-stop, line 145: This was the goal: the router service _disappears_ and now only exists as a hook-point for plugins. A follow-up PR could literally delete the `RouterService` structure.

apollo-router/src/services/router/service.rs
```diff
@@ -225,58 +392,48 @@ impl Service<RouterRequest> for RouterService {
...
-
-        if let Ok(supergraph_request) = request_res {
-            request_res = self.apq_layer.supergraph_request(supergraph_request).await;
-        }
+impl<S> RouterToSupergraphRequestService<S>
+where
+    S: Service<SupergraphRequest, Response = SupergraphResponse, Error = BoxError> + Send + 'static,
+    S::Future: Send + 'static,
+{
+    async fn call_inner(&mut self, req: RouterRequest) -> Result<RouterResponse, BoxError> {
```
goto-bus-stop, line 400: This layer is still a bit nasty on the response side: it does some content negotiation, error counting, sets the status code for cancelled requests, and serializes defer and subscription responses. This is fixable but I would rather do it separately, because it is trickier than the stuff I've done here, and it will complicate the PR even more.

apollo-router/src/services/router/batching.rs
```diff
@@ -0,0 +1,421 @@
...
+        //
+        // (Technically we could clone extensions, since it is held under an `Arc`, but that
+        // would mean all the requests in a batch shared the same set of extensions and review
+        // comments expressed the sentiment that this may be a bad thing...)
+        //
+        // Note: If we enter this loop, then we must be processing a batch.
+        for (index, graphql_request) in ok_results_it.enumerate() {
+            let body = router::body::from_bytes(serde_json::to_vec(&graphql_request).unwrap());
+            let new = http::Request::from_parts(parts.clone(), body);
+            // XXX Lose some private entries, is that ok?
```
rohan-b99, line 348: Not sure if we covered this when looking previously, could this cause an issue in future?

apollo-router/src/services/router/batching.rs
```diff
@@ -0,0 +1,421 @@
...
+            let first = results_it
+                .next()
+                .expect("we should have at least one response");
+            let (parts, body) = first.response.into_parts();
+            let context = first.context;
+            let mut bytes = BytesMut::new();
+            bytes.put_u8(b'[');
+            bytes.extend_from_slice(&router::body::into_bytes(body).await?);
+            for result in results_it {
+                bytes.put(&b", "[..]);
```
rohan-b99: Nitpick, can we omit the space here?
```suggestion
                bytes.put_u8(b',');
```

apollo-router/src/services/router/batching.rs
goto-bus-stop: Oh, we didn't cover it, but it is just a copy-paste of the previous code, not a new problem. It seems like it might not be an issue because we haven't heard anything from customers as far as I know?

apollo-router/src/services/router/batching.rs
goto-bus-stop: might as well!

PR #7676, opened 2025-06-12
Align `on_graphql_error` selector return values with `subgraph_on_graphql_error`

carodewig:
Currently the `subgraph_on_graphql_error` selector will return `true` or `false` while the `on_graphql_error` selector will return `true` or `None`. This updates the `on_graphql_error` selector to return `true` or `false` to align with `subgraph_on_graphql_error`.

---

**Checklist**

- Tests added and passing[^4]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


goto-bus-stop:
I don't quite understand, is this a user-visible change or an internal improvement? The new test seems to pass either way

carodewig:
> I don't quite understand, is this a user-visible change or an internal improvement? The new test seems to pass either way

@goto-bus-stop This is user visible! The test I added in 50a90d52af10573 is a bit clearer. The current behavior of the router is to not include the attribute on the metric when the selector returns `None` (ie no attributes would be present on the new datapoints); this change makes the attribute visible as `false` instead.

PR #7684, opened 2025-06-13
Remove `@` from error paths

carodewig:
When a subgraph returns an unexpected response (ie not a body with at least one of `errors` or `data`), the errors surfaced by the router include an `@` in the path which indicates an error applied to all elements in the array. This is not a behavior defined in the GraphQL spec and is not easily parsed.

This PR expands the `@` symbol to reflect all paths that the error should apply to.

The tests reflect the following federated graph with 4 subgraphs:
```
products ----> inventory
         \---> reviews
                     \---> accounts
```

If all subgraphs respond successfully, the response might look like this:
```json
{
    "data": {
        "topProducts": [
            {
                "name": "Table",
                "inStock": true,
                "reviews": [
                    {"id": "1", "author": {"username": "@ada", "name": "Ada"}},
                    {"id": "2", "author": {"username": "@alan", "name": "Alan"}}
                ]
            },
            {
                "name": "Chair",
                "inStock": false,
                "reviews": [
                    {"id": "3", "author": {"username": "@alan", "name": "Alan"}}
                ]
            }
        ]
    }
}
```

Currently, if the `inventory` subgraph returns a malformed response, the router response will look like:
```json
{
    "data": {"topProducts": [{"name": "Table", "inStock": null, "reviews": [{"id": "1", "author": {"username": "@ada", "name": "Ada"}}, {"id": "2", "author": {"username": "@alan", "name": "Alan"}}]}, {"name": "Chair", "inStock": null, "reviews": [{"id": "3", "author": {"username": "@alan", "name": "Alan"}}]}]}, 
    "errors": [
        {
            "message": "service 'inventory' response was malformed: graphql response without data must contain at least one error", 
            "path": ["topProducts", "@"], 
            "extensions": {"service": "inventory", "reason": "graphql response without data must contain at least one error", "code": "SUBREQUEST_MALFORMED_RESPONSE"}
        }
    ]
}
```

With this change, the response will look like:
```json
{
    "data": {"topProducts": [{"name": "Table", "inStock": null, "reviews": [{"id": "1", "author": {"username": "@ada", "name": "Ada"}}, {"id": "2", "author": {"username": "@alan", "name": "Alan"}}]}, {"name": "Chair", "inStock": null, "reviews": [{"id": "3", "author": {"username": "@alan", "name": "Alan"}}]}]}, 
    "errors": [
        {
            "message": "service 'inventory' response was malformed: graphql response without data must contain at least one error",
            "path":["topProducts",0],
            "extensions":{"service":"inventory","reason":"graphql response without data must contain at least one error","code":"SUBREQUEST_MALFORMED_RESPONSE"}
        },
        {
            "message":"service 'inventory' response was malformed: graphql response without data must contain at least one error",
            "path":["topProducts",1],
            "extensions":{"service":"inventory","reason":"graphql response without data must contain at least one error","code":"SUBREQUEST_MALFORMED_RESPONSE"}
        }
    ]
}
```

If `include_subgraph_errors` is not enabled, the response will be:
```json
{
    "data": {"topProducts": [{"name": "Table", "inStock": null, "reviews": [{"id": "1", "author": {"username": "@ada", "name": "Ada"}}, {"id": "2", "author": {"username": "@alan", "name": "Alan"}}]}, {"name": "Chair", "inStock": null, "reviews": [{"id": "3", "author": {"username": "@alan", "name": "Alan"}}]}]}, 
    "errors": [
        {
            "message": "Subgraph errors redacted",
            "path": ["topProducts", 0]
        },
        {
            "message": "Subgraph errors redacted",
            "path": ["topProducts", 1]
        }
    ]
}
``` 

TODO:
- [x] Would it be better to just specify `0` as the element rather than expanding all the paths?
- [ ] Is `inverted_paths` the best way to handle this?
- [ ] Should this be applied in the `Fetch` step of the query plan, or should it take place in the `Flatten` step? I put it in the `Fetch` step because that's where the `inverted_paths` variable is accessible, so I think that's where we know all the paths the error applies to.
- [x] Document the reasoning behind `interim_errors` inline
- [x] Use snapshots to capture which errors are actually present rather than just counting the number of errors (in the tests)

---

**Checklist**

- Tests added and passing[^4]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


goto-bus-stop:
> Would it be better to just specify 0 as the element rather than expanding all the paths?

I think probably that expanding everything is better.

It's worth checking with the GraphQL spec what it means to return both `data` and `errors` at the same path--I think normally when there is a field error, you would expect to get `null` at that path. So perhaps we should even extend the paths to reference the specific fields that would be returned by that subgraph. But hopefully it's just allowed to do what this PR does :)

I'm not familiar with this part of the code yet so I don't immediately have a reaction to the other open questions, hopefully later.

SimonSapin:
The most relevant bit of spec I find is in https://spec.graphql.org/draft/#sec-Handling-Execution-Errors
> If an execution error is raised while resolving a field (either directly or nested inside any lists), it is handled as though the [response position](https://spec.graphql.org/draft/#response-position) at which the error occurred resolved to null, and the error must be added to the "errors" list in the [execution result](https://spec.graphql.org/draft/#execution-result).

I believe this text is written with the assumption that an execution error can only happen *instead of* a data result. It says when we don’t have data to use `null` instead, somewhere in `data`.

The possibility of having *both* an execution error and a data result for the same field is not discussed but it’s not clear to me that that makes it non-conforming.

apollo-router/src/axum_factory/snapshots/apollo_router__axum_factory__tests__defer_is_not_buffered.snap
```diff
@@ -23,7 +23,18 @@ expression: parts
...
-          "@"
+          0
+        ],
+        "extensions": {
+          "code": "FETCH_ERROR",
+          "service": "reviews"
+        }
+      },
+      {
+        "message": "couldn't find mock for query {\"query\":\"query($representations: [_Any!]!) { _entities(representations: $representations) { ... on Product { reviews { __typename id product { __typename upc } } } } }\",\"variables\":{\"representations\":[{\"__typename\":\"Product\",\"upc\":\"1\"},{\"__typename\":\"Product\",\"upc\":\"2\"}]}}",
```
goto-bus-stop: this test now no longer produces error responses after https://github.com/apollographql/router/pull/7711

apollo-router/src/query_planner/execution.rs
```diff
@@ -298,14 +300,40 @@ impl PlanNode {
...
+                                // include an @ in the path. This indicates the error should be applied
+                                // to all elements in the array.
+                                errors = Vec::default();
+                                for err in raw_errors {
+                                    if let Some(err_path) = err.path.as_ref() {
+                                        if err_path
+                                            .iter()
+                                            .any(|elem| matches!(elem, PathElement::Flatten(_)))
+                                        {
+                                            for path in paths.iter().flatten() {
```
duckki, line 324: I think `paths` should be filtered here.
Suppose the `raw_errors` has multiple patterns like

- `x.@`
- `y.z.@`

Each pattern will capture a group of paths, not necessarily all `paths`. For example, `x.@` will match `x.1` and `x.2`, etc., but not `y.z.1`. So, I think we need to `match` each path agains the given `err_path`'s "pattern" in a similar fashion as regex matching.

carodewig: Great point, thank you! Addressed in bba7317

abernix:
@carodewig Can you update the PR description on this (or just comment) with an example of what the response will look like after this change if [`include_subgraph_errors`](https://www.apollographql.com/docs/graphos/routing/observability/subgraph-error-inclusion) is **not** enabled?

carodewig:
> @carodewig Can you update the PR description on this (or just comment) with an example of what the response will look like after this change if [`include_subgraph_errors`](https://www.apollographql.com/docs/graphos/routing/observability/subgraph-error-inclusion) is **not** enabled?

@abernix Description and changeset updated! 2b6eed8

BrynCooke:
@carodewig What is the potential max number of errors that users will get. Could it potentially blow up?

Is there a way of setting the max errors to a sensible default?

carodewig:
@BrynCooke It is unlimited as it corresponds to the number of entities requested. Limiting the maximum number of errors is a good idea but we will consider it separate from this PR.

PR #7688, opened 2025-06-13
fix(composition): remove unnecessary metadata recomputation

dariuszkuc:
While upgrading subgraphs we don't need to recompute this metadata.

### DIFF
--- a/apollo-federation/src/schema/schema_upgrader.rs
+++ b/apollo-federation/src/schema/schema_upgrader.rs
@@ -12,7 +12,6 @@ use apollo_compiler::validation::Valid;
 
 use super::FederationSchema;
 use super::TypeDefinitionPosition;
-use super::compute_subgraph_metadata;
 use super::field_set::collect_target_fields_from_field_set;
 use super::position::DirectiveDefinitionPosition;
 use super::position::FieldDefinitionPosition;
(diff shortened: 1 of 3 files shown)


tninesling (approved):

apollo-federation/src/schema/schema_upgrader.rs
```diff
@@ -702,44 +701,32 @@ impl SchemaUpgrader {
...
                 for field in pos.fields(schema.schema())? {
                     has_fields = true;
                     let field_def = FieldDefinitionPosition::from(field.clone());
-                    let metadata = compute_subgraph_metadata(schema)?.ok_or_else(|| {
-                        internal_error!(
-                            "Unable to detect federation version used in subgraph '{}'",
-                            upgrade_metadata.subgraph_name
-                        )
-                    })?;
+                    let metadata = &upgrade_metadata.metadata;
```
duckki, line 713: Isn't it necessary to compute the metadata at the top of this method? Some stuffs are removed between the `UpgradeMetadata`'s construction and the call to this function.

dariuszkuc: We need metadata to quickly check if field is `@external` AND it is not used in any of our directives (`@key`, `@provides`, etc).

AFAIK this is a catch all for fields that are marked `@external` but are not used by any directives. It is true that by the time we hit this, some fields (e.g. `@key`) will have `@external` removed but they will still be "used" so the condition check will still be valid.

```graphql
type Foo @key(fields: "id") {
  id: ID! @external # doesn't make sense to have keys as external, so we'll remove the directive
  required: String @external # used by @requires below, we'll keep it
  computed: String @requires(fields: "required")
  extra: String @external # not used by any directive, will be dropped by this method
}
```

We also only need to process objects as for interfaces its "nonsensical" (:)) to have `@external` fields. We just remove those directive applications and there is nothing to post process here.

PR #7694, opened 2025-06-15
fix(http): correct Tower service usage and preserve backpressure

BrynCooke:
## Summary

Fixed incorrect Tower service usage in the HTTP client that was breaking backpressure flow control. The service was consuming itself on each call instead of being reusable, which violates Tower's service patterns and could lead to connection issues.

## Problem

The HTTP client service was incorrectly using Tower's `call()` method which consumes the service, making it unusable for subsequent requests. This breaks Tower's backpressure mechanism that services rely on for proper flow control. Without proper service reuse, each request created unnecessary overhead and could lead to connection problems when making requests to subgraph services.

## Solution

Implemented proper Tower service patterns with concurrent caching:

- Added service caching in `HttpClientServiceFactory` using `DashMap` for lock-free concurrent access
- Used `ServiceBuilder` with buffered cloning to maintain proper backpressure while enabling service reuse
- Changed subgraph service from `call()` to `oneshot()` for correct Tower service consumption
- Added tests that cover cache hits, misses, and service isolation by name

## Technical Details

The implementation ensures Tower services follow their intended patterns:

- **Backpressure Preservation**: Services are cached and reused rather than consumed, maintaining flow control through the service chain
- **Concurrent Access**: `DashMap` provides thread-safe caching without blocking or lock contention
- **Service Lifecycle**: Buffered cloning allows multiple concurrent requests while preserving proper Tower service semantics

## Review Strategy

- **Service Lifecycle**: Examine `HttpClientServiceFactory::create()` method for proper caching and service creation logic
- **Tower Patterns**: Verify the change from `call()` to `oneshot()` in subgraph service interaction follows Tower conventions
- **Backpressure Flow**: Check that buffered services properly propagate `poll_ready` signals through the service chain
- **API Compatibility**: Confirm constructor changes don't break existing `HttpClientServiceFactory` usage patterns
- **Test Coverage**: Review `test_service_caching()` for adequate validation of caching behavior

---

**Checklist**

- Tests added and passing[^4]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


.changesets/maint_refactor_http_service_clone.md
```diff
@@ -0,0 +1,7 @@
+### HTTP client service now allows backpressure ([PR #7694](https://github.com/apollographql/router/pull/7694))
+
+This internal refactoring caches http_client services rather than recreated them on every call.
```
garypen:
```suggestion
This internal refactoring caches http_client services rather than recreating them on every call. 
```

apollo-router/src/services/http.rs
```diff
@@ -61,17 +76,29 @@ impl HttpClientServiceFactory {
...
-        let service = self.service.clone();
-        self.plugins
-            .iter()
-            .rev()
-            .fold(service.boxed(), |acc, (_, e)| {
-                e.http_client_service(name, acc)
-            })
+        // Check if we already have a memoized service for this name
+        if let Some(service) = self.cache.get(name) {
+            service.deref().clone().boxed()
```
garypen:
```suggestion
            service.clone().boxed()
```

apollo-router/src/services/http.rs
```diff
@@ -1,8 +1,14 @@
 #![allow(dead_code)]
+
+use std::ops::Deref;
```
garypen:
```suggestion
```

garypen (commented):
Looks good. The `deref()` is redundant.

.changesets/maint_refactor_http_service_clone.md
```diff
@@ -0,0 +1,7 @@
+### HTTP client service now allows backpressure ([PR #7694](https://github.com/apollographql/router/pull/7694))
+
+This internal refactoring caches http_client services rather than recreated them on every call. 
+This has no impact on user functionality but is retained in the changelog so that other teams
+can be alerted to this change in behavior.
```
goto-bus-stop: It will mean the `http_client_service` hook in plugins will be called only once per subgraph, instead of for every request, right? that seems worth calling out.

.changesets/maint_refactor_http_service_clone.md
```diff
@@ -0,0 +1,7 @@
+### HTTP client service now allows backpressure ([PR #7694](https://github.com/apollographql/router/pull/7694))
+
+This internal refactoring caches http_client services rather than recreating them on every call. 
+This has no impact on user functionality but is retained in the changelog so that other teams
+can be alerted to this change in behavior.
```
BrynCooke:
```suggestion
This internal refactoring caches http_client services rather than recreating them on every call. 
This has no impact on user functionality but is retained in the changelog so that other teams
can be alerted to this change in behavior. 

Note that plugins that implement PluginPrivate will now only receive one call to http_client_service for each distict client rather than many.
```

.changesets/maint_refactor_http_service_clone.md
BrynCooke: [c03052f](https://github.com/apollographql/router/pull/7694/commits/c03052fc19500727f8ac3dad9c08e75e3cb87f75)

goto-bus-stop:
Oh, I didn't realise http_client_service is only for private plugins, maybe we don't need to mention it then 🙈

garypen (approved):

goto-bus-stop (approved):

BrynCooke:
Before merging we need to check any plugin that uses http_client_service is not expecting the old way of working.

PR #7698, opened 2025-06-16
New respsonse caching implementation

bnjjj:
This PR contains new response caching implementation based on 

Fixes #**issue_number**

---

**Checklist**

- Tests added and passing[^4]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/tests/integration/mod.rs
```diff
@@ -27,6 +27,8 @@ mod typename;
 
 // In the CI environment we only install Redis on x86_64 Linux
 #[cfg(any(not(feature = "ci"), all(target_arch = "x86_64", target_os = "linux")))]
+mod postgres;
+#[cfg(any(not(feature = "ci"), all(target_arch = "x86_64", target_os = "linux")))]
```
SimonSapin:
```suggestion
// In the CI environment we only install PostgreSQL on x86_64 Linux
#[cfg(any(not(feature = "ci"), all(target_arch = "x86_64", target_os = "linux")))]
mod postgres;
// In the CI environment we only install Redis on x86_64 Linux
#[cfg(any(not(feature = "ci"), all(target_arch = "x86_64", target_os = "linux")))]
```

apollo-router/src/query_planner/bridge_query_planner_pool.rs
goto-bus-stop, line 1: this file shouldn't exist anymore, maybe a merge artifact?

apollo-federation/src/connectors/string_template.rs
```diff
@@ -109,7 +109,7 @@ impl StringTemplate {
     /// Interpolate the expressions in the template into a basic string.
     ///
     /// For URIs, use [`Self::interpolate_uri`] instead.
-    pub(crate) fn interpolate(
+    pub fn interpolate(
```
bnjjj, line 112: @lennyburdette any concerns ?

lennyburdette: not on our end!

lennyburdette (commented):
connectors related changes look good to me!

.circleci/config.yml
```diff
@@ -27,6 +27,10 @@ executors:
     docker:
       - image: ghcr.io/apollographql/ci-utility-docker-images/apollo-rust-builder:0.11.2
       - image: cimg/redis:7.4.4
+      - image: cimg/postgres:17.0
+        environment:
+          POSTGRES_USER: root
+          POSTGRES_DB: root
```
SimonSapin, line 33: This is the only non-whitespace change in this file, right?

apollo-router/src/plugins/response_cache/invalidation_endpoint.rs
```diff
@@ -269,115 +263,115 @@ fn valid_shared_key(
             .unwrap_or_default()
 }
 
-#[cfg(test)]
-mod tests {
-    use std::collections::HashMap;
+// #[cfg(test)]
+// mod tests {
```
SimonSapin: Let’s either uncomment/fix or remove this module

apollo-router/src/plugins/mock_subgraphs/execution/input_coercion.rs
```diff
@@ -83,7 +83,7 @@ fn graphql_value_to_json(
 pub(crate) fn coerce_argument_values(
     schema: &Schema,
     document: &Valid<ExecutableDocument>,
-    variable_values: &Valid<JsonMap>,
```
SimonSapin: Please revert this change and at the call site use either `coerce_variable_values` or `Valid::assume_valid_ref`. If the latter, add a comment here to explain that something else (maybe `Query::validate_variables`?) fills the same role.

apollo-router/src/plugins/mock_subgraphs/execution/input_coercion.rs
```diff
@@ -160,7 +160,7 @@ pub(crate) fn coerce_argument_values(
 fn coerce_argument_value(
     schema: &Schema,
     document: &Valid<ExecutableDocument>,
-    variable_values: &Valid<JsonMap>,
```
SimonSapin: Same as above

apollo-federation/src/cache_tag/spec/mod.rs
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    pub(crate) fn url(&self) -> Url {
+        Url {
+            identity: Self::identity(),
+            version: (*self).into(),
+        }
+    }
+
+    // pub(crate) fn get_from_schema(schema: &Schema) -> Option<(Self, Link)> {
```
SimonSapin: Let’s remove this (and another commented out method below) if it isn’t needed

apollo-router/src/plugins/response_cache/invalidation.rs
```diff
@@ -50,148 +49,153 @@ impl std::fmt::Display for InvalidationErrors {
...
-            InvalidationOrigin::Endpoint => "endpoint",
-            InvalidationOrigin::Extensions => "extensions",
-        };
         u64_counter!(
-            "apollo.router.operations.response_cache.invalidation.event",
-            "Response cache received a batch of invalidation requests",
-            1u64,
-            "origin" = origin
+            "apollo.router.operations.entity.invalidation.event",
+            "Entity cache received a batch of invalidation requests",
```
SimonSapin: Looks like this accidentally reverts the rename from entity cache to response cache. This revert should be reverted.

apollo-router/src/plugins/response_cache/invalidation.rs
```diff
@@ -50,148 +49,153 @@ impl std::fmt::Display for InvalidationErrors {
...
-                    scan_res.next();
-                }
+        let count = match request {
+            InvalidationRequest::Subgraph { subgraph } => {
+                let count = pg_storage
+                    .invalidate_by_subgraphs(vec![subgraph.clone()])
+                    .await?;
+                u64_counter!(
+                    "apollo.router.operations.entity.invalidation.entry",
+                    "Entity cache counter for invalidated entries",
```
SimonSapin: Same as above re "entity cache" v.s. "response cache", in both the metric ID and description

apollo-router/src/plugins/response_cache/invalidation.rs
```diff
@@ -50,148 +49,153 @@ impl std::fmt::Display for InvalidationErrors {
...
-        u64_counter!(
-            "apollo.router.operations.response_cache.invalidation.entry",
-            "response_cache counter for invalidated entries",
-            count,
-            "origin" = origin,
-            "subgraph.name" = subgraph.clone()
-        );
+                u64_counter!(
+                    "apollo.router.operations.entity.invalidation.entry",
+                    "Entity cache counter for invalidated entries",
```
SimonSapin: Same again

apollo-router/src/plugins/response_cache/invalidation.rs
```diff
@@ -50,148 +49,153 @@ impl std::fmt::Display for InvalidationErrors {
...
+                    )
+                    .await?
+                // TODO: fixme
+                // u64_counter!(
+                //     "apollo.router.operations.entity.invalidation.entry",
+                //     "Entity cache counter for invalidated entries",
+                //     count,
+                //     "origin" = origin,
+                //     "subgraph.name" = subgraphs.clone()
+                // );
```
SimonSapin: What fix does this need?

apollo-router/src/plugins/response_cache/invalidation.rs
```diff
@@ -50,148 +49,153 @@ impl std::fmt::Display for InvalidationErrors {
...
+                InvalidationRequest::CacheTag { subgraphs, .. } => {
+                    let mut storages = Vec::new();
+                    for subgraph in subgraphs {
+                        match self.storage.get(subgraph) {
+                            Some(s) => storages.push(s),
+                            None => continue,
+                        }
+                    }
+
+                    storages
```
SimonSapin: Nit: (likely needs fmt)
```suggestion
                    subgraphs.iter().filter_map(|subgraph| self.storage.get(subgraph)).collect()
```

apollo-router/src/plugins/response_cache/invalidation.rs
```diff
@@ -50,148 +49,153 @@ impl std::fmt::Display for InvalidationErrors {
...
-            let semaphore = self.semaphore.clone();
-            let f = async move {
-                // limit the number of invalidation requests executing at any point in time
-                let _ = semaphore.acquire().await;
+            for pg_storage in storages {
+                let semaphore = self.semaphore.clone();
+                let mut request = request.clone();
+                let f = async move {
+                    // limit the number of invalidation requests executing at any point in time
+                    let _ = semaphore.acquire().await;
```
SimonSapin: Is this is relevant as it was with Redis’ `SCAN` full-database linear search?

apollo-router/src/plugins/response_cache/invalidation.rs
```diff
@@ -50,148 +49,153 @@ impl std::fmt::Display for InvalidationErrors {
...
-                f64_histogram!(
-                    "apollo.router.cache.invalidation.duration",
-                    "Duration of the invalidation event execution, in seconds.",
-                    start.elapsed().as_secs_f64()
-                );
-                res
-            };
-            futures.push(f);
+                    f64_histogram!(
+                        "apollo.router.cache.invalidation.duration",
```
SimonSapin: Is this ID consistent with others we’ve settled on?

apollo-router/src/plugins/response_cache/invalidation.rs
```diff
@@ -226,17 +230,31 @@ pub(crate) enum InvalidationRequest {
         r#type: String,
         key: serde_json_bytes::Map<ByteString, Value>,
     },
+    CacheTag {
+        subgraphs: HashSet<String>,
```
SimonSapin, line 215: (Since this `InvalidationRequest` enum is already in a `Vec` I’m not convinced it’s that useful to allow multiple subgraphs here, but we can revisit that in a future PR.)

apollo-router/src/plugins/response_cache/invalidation.rs
```diff
@@ -226,17 +230,31 @@ pub(crate) enum InvalidationRequest {
...
+            InvalidationRequest::Subgraph { subgraph }
+            | InvalidationRequest::Type { subgraph, .. }
+            | InvalidationRequest::Entity { subgraph, .. } => vec![subgraph.clone()],
+            InvalidationRequest::CacheTag { subgraphs, .. } => {
+                subgraphs.clone().into_iter().collect()
+            }
+        }
+    }
     /// Compute a cache key prefix. For entity keys, this destructively sorts all objects.
     fn key_prefix(&mut self) -> String {
```
SimonSapin: We’re not still doing a prefix search, are we? Should this be renamed to something like `invalidation_key`?

apollo-router/src/plugins/response_cache/invalidation_endpoint.rs
```diff
@@ -153,11 +145,13 @@ impl Service<router::Request> for InvalidationService {
...
                                 );
-                                let valid_shared_key =
-                                    body.iter().map(|b| b.subgraph_name()).any(|subgraph_name| {
-                                        valid_shared_key(&config, shared_key, subgraph_name)
+                                let invalid_shared_key = body
+                                    .iter()
+                                    .flat_map(|b| b.subgraph_names())
+                                    .any(|subgraph_name| {
+                                        !valid_shared_key(&config, shared_key, &subgraph_name)
                                     });
```
SimonSapin: This changes the logic from "valid if any is valid" to "invalid if any is invalid". The latter is equivalent to "valid if all are valid".

I’m honestly not sure which logic is correct. It’s weird in the first place that config allows a different expected key per subgraph, but one endpoint request can only have one auth header but can affect multiple subgraphs. Should we change config to a single expected key per router? (And while we’re at it, maybe rename it to passphrase or something because we have others things called keys)

apollo-router/src/plugins/response_cache/metrics.rs
```diff
@@ -13,16 +13,16 @@ use tower::ServiceBuilder;
...
-use crate::plugins::response_cache::plugin::Ttl;
-use crate::plugins::response_cache::plugin::hash_query;
-use crate::plugins::response_cache::plugin::hash_vary_headers;
 use crate::services::subgraph;
 use crate::spec::TYPENAME;
 
-pub(crate) const CACHE_INFO_SUBGRAPH_NAME_CONTEXT_KEY: &str =
-    "apollo::router::response_cache_info_subgraph_name";
+pub(crate) const CACHE_INFO_SUBGRAPH_CONTEXT_KEY: &str =
+    "apollo::router::plugin_cache_info_subgraph";
```
SimonSapin, line 25: Renaming `response_cache` to `plugin_cache` looks maybe wrong? This could use a comment to explain what it is for. (I found code using `CacheMetricContextKey` but it’s not easy to reverse-engineer.)

apollo-router/src/plugins/response_cache/metrics.rs
```diff
@@ -249,7 +249,7 @@ impl CacheCounter {
         for (typename, (cache_hit, total_entities)) in seen.into_iter() {
             if separate_metrics_per_type {
                 f64_histogram!(
-                    "apollo.router.operations.response_cache.cache_hit",
+                    "apollo.router.operations.entity.cache_hit",
```
SimonSapin: Revert?

apollo-router/src/plugins/response_cache/metrics.rs
```diff
@@ -259,7 +259,7 @@ impl CacheCounter {
                 );
             } else {
                 f64_histogram!(
-                    "apollo.router.operations.response_cache.cache_hit",
+                    "apollo.router.operations.entity.cache_hit",
```
SimonSapin: Revert?

apollo-router/src/plugins/response_cache/plugin.rs
```diff
@@ -91,45 +94,55 @@ pub(crate) struct ResponseCache {
...
+            all.migrate().await?;
+        }
+        futures::future::try_join_all(self.subgraphs.values().map(|s| s.migrate())).await?;
+
+        Ok(())
+    }
 }
 
-/// Configuration for entity caching
+/// Configuration for subgraph caching
```
SimonSapin:
```suggestion
/// Configuration for response caching
```

apollo-router/src/plugins/response_cache/plugin.rs
```diff
@@ -91,45 +94,55 @@ pub(crate) struct ResponseCache {
...
 #[derive(Clone, Debug, JsonSchema, Deserialize)]
 #[serde(rename_all = "snake_case", deny_unknown_fields)]
 pub(crate) struct Config {
-    /// Enable or disable the entity caching feature
+    /// Enable or disable the subgraph caching feature
```
SimonSapin:
```suggestion
    /// Enable or disable the response caching feature
```

apollo-router/src/plugins/response_cache/plugin.rs
```diff
@@ -192,7 +205,6 @@ pub(crate) struct CacheHitMiss {
 
 #[async_trait::async_trait]
 impl PluginPrivate for ResponseCache {
-    const HIDDEN_FROM_CONFIG_JSON_SCHEMA: bool = true;
```
SimonSapin: Isn’t it still a bit early for experimental users?

apollo-router/src/plugins/response_cache/plugin.rs
```diff
@@ -208,43 +220,44 @@ impl PluginPrivate for ResponseCache {
...
+            let required_to_start = postgres_config.required_to_start;
             // we need to explicitly disable TTL reset because it is managed directly by this plugin
-            redis_config.reset_ttl = false;
-            all = match RedisCacheStorage::new(redis_config, "entity").await {
+            // postgres_config.reset_ttl = false;
+            all = match PostgresCacheStorage::new(&postgres_config).await {
                 Ok(storage) => Some(storage),
                 Err(e) => {
                     tracing::error!(
                         cache = "entity",
```
SimonSapin:
```suggestion
                        cache = "response",
```

apollo-router/src/plugins/response_cache/plugin.rs
```diff
@@ -208,43 +220,44 @@ impl PluginPrivate for ResponseCache {
...
-        if let Some(redis) = &init.config.subgraph.all.redis {
-            let mut redis_config = redis.clone();
-            let required_to_start = redis_config.required_to_start;
+        if let Some(postgres) = &init.config.subgraph.all.postgres {
+            let postgres_config = postgres.clone();
+            let required_to_start = postgres_config.required_to_start;
             // we need to explicitly disable TTL reset because it is managed directly by this plugin
-            redis_config.reset_ttl = false;
-            all = match RedisCacheStorage::new(redis_config, "entity").await {
+            // postgres_config.reset_ttl = false;
```
SimonSapin: Should this be removed? (Also below)

apollo-router/src/plugins/response_cache/plugin.rs
```diff
@@ -664,22 +682,70 @@ impl CacheService {
...
                             CacheSubgraph(cache_hit),
                         );
+                        let mut root_operation_fields: Vec<String> = Vec::new();
+                        let mut debug_subgraph_request = None;
+                        if self.debug {
+                            root_operation_fields = request
+                                .executable_document
+                                .as_ref()
+                                .and_then(|executable_document| {
+                                    Some(
```
SimonSapin:
```suggestion
                                .map(|executable_document| {
```
(and remove the matching `)` below)

apollo-router/src/plugins/response_cache/plugin.rs
```diff
@@ -664,22 +682,70 @@ impl CacheService {
...
+                                        executable_document
+                                            .operations
+                                            .iter()
+                                            .next()?
```
SimonSapin: Even though it shouldn’t make a difference since the planner only generates one operation, I prefer this:
```suggestion
                                            .get(request.subgraph_request.body().operation_name)?
```

apollo-router/src/plugins/response_cache/plugin.rs
```diff
@@ -894,34 +922,38 @@ impl CacheService {
...
-                tracing::error!(error = %e,
-                   message = "could not invalidate response cache entries",
-                );
-            }
-        }
-    }
 }
 
+#[derive(Default, Clone)]
+pub(crate) struct InvalidationKeysPropagation(HashSet<String>);
```
SimonSapin: Is this for `@cacheTag` "cascade"?

apollo-router/src/plugins/response_cache/plugin.rs
```diff
@@ -931,100 +963,235 @@ async fn cache_lookup_root(
...
+        })?;
+    let query_object_type = supergraph_schema
+        .get_object(root_query_type.as_str())
+        .ok_or_else(|| FetchError::MalformedRequest {
+            reason: "cannot get the root query type from supergraph schema".to_string(),
+        })?;
+
+    let cache_keys = root_operation_fields
+        .map(|field| {
+            let field_def = query_object_type.fields.get(&field.name).ok_or_else(|| {
```
SimonSapin, line 1041: Is this not `&field.definition`? If not let’s add a code comment to explain

apollo-router/src/plugins/response_cache/plugin.rs
```diff
@@ -931,100 +963,235 @@ async fn cache_lookup_root(
...
+                                    cascade,
+                                    value
+                                        .as_str()
+                                        .and_then(|v| v.parse::<StringTemplate>().ok()),
+                                )
+                            } else if field_name.as_str() == "cascade" {
+                                let cascade = match &**value {
+                                    apollo_compiler::ast::Value::Boolean(cascade) => Some(*cascade),
+                                    _ => None,
+                                };
```
SimonSapin:
```suggestion
                                let cascade = value.to_bool();
```

apollo-router/src/plugins/response_cache/plugin.rs
```diff
@@ -931,100 +963,235 @@ async fn cache_lookup_root(
...
+                        .ok()?
+                        .as_object()?
+                        .iter()
+                        .fold((None, None), |(cascade, format), (field_name, value)| {
+                            if field_name.as_str() == "format" {
+                                (
+                                    cascade,
+                                    value
+                                        .as_str()
+                                        .and_then(|v| v.parse::<StringTemplate>().ok()),
```
SimonSapin: Is this parsing expensive? I think it and much of the directive wrangling here could be moved to plugin init rather than in the request hot path, but that would be a future PR.

apollo-router/src/plugins/response_cache/plugin.rs
```diff
@@ -1131,93 +1289,73 @@ fn update_cache_control(context: &Context, cache_control: &CacheControl) {
...
-                                );
-                            }
-                        }
-
-                        val
-                    })?;
-            }
 
+            let subgraph_name = response.subgraph_name.clone();
             tokio::spawn(async move {
```
SimonSapin, line 1288: As mentioned by Gary, let’s add a comment here to explain that deliberately don’t wait for cache insertion to finish before moving on with the request/response process

apollo-router/src/plugins/response_cache/plugin.rs
```diff
@@ -1910,125 +2242,125 @@ impl Ord for CacheKeyStatus {
...
-            Some(Duration::from_secs(42))
-        );
-        assert_eq!(
-            response_cache.subgraph_ttl("archive", &redis_cache),
-            Some(Duration::from_millis(5000))
-        );
-    }
-}
+// #[cfg(test)]
+// mod tests {
```
SimonSapin: Here as well, let’s uncomment or remove this module.

apollo-router/src/plugins/response_cache/plugin.rs
```diff
@@ -1910,125 +2242,125 @@ impl Ord for CacheKeyStatus {
...
+//                 "private_id": "sub",
+//                 "enabled": true
+//             },
+//             "archive": {
+//                 "private_id": "sub",
+//                 "enabled": false
+//             }
+//         });
+
+//         let mut entity_cache = SubgraphCache::with_mocks(
```
SimonSapin: If/when uncommenting, revert the revert of rename from entity cache to subgraph cache

apollo-router/src/plugins/response_cache/postgres.rs
```diff
@@ -0,0 +1,444 @@
...
+        }
+    }
+
+    pub(crate) async fn migrate(&self) -> anyhow::Result<()> {
+        sqlx::migrate!().run(&self.pg_pool).await?;
+        Ok(())
+    }
+
+    #[allow(dead_code)]
+    pub(crate) async fn truncate(&self) -> anyhow::Result<()> {
```
SimonSapin: If it’s unused, remove this?

apollo-router/src/plugins/response_cache/postgres.rs
```diff
@@ -0,0 +1,444 @@
...
+#[derive(sqlx::FromRow, Debug, Clone)]
+pub(crate) struct CacheEntryRow {
+    pub(crate) id: i64,
+    pub(crate) cache_key: String,
+    pub(crate) data: String,
+    pub(crate) expires_at: DateTime<Utc>,
+    pub(crate) control: String,
+}
+
+#[allow(dead_code)]
```
SimonSapin:
```suggestion
```

apollo-router/src/plugins/response_cache/invalidation.rs
bnjjj: The count per subgraph should be returned and not only the count

.circleci/config.yml
bnjjj: Yes

apollo-router/src/plugins/response_cache/invalidation_endpoint.rs
bnjjj: Fixed

apollo-router/src/plugins/response_cache/invalidation.rs
bnjjj: But it can be done in a follow up

apollo-router/src/plugins/response_cache/invalidation.rs
bnjjj: I'm not sure TBH. What do you think ?

apollo-router/src/plugins/response_cache/invalidation.rs
bnjjj: That's something we can discuss, but it will be more efficient at the execution level, or we could try to make a process to group the invalidation requests together to only execute it once. It's also pretty useful as you can specify cacheTag on different subgraphs. I think we can revisit this but we need more opinion/feedback

apollo-router/src/plugins/response_cache/invalidation_endpoint.rs
bnjjj: Passphrase sounds great ! I still think having several passphrases for each subgraph is relevant. Maybe we could provide a way to specify several authorization passphrase ?

apollo-router/src/plugins/response_cache/metrics.rs
bnjjj: Feel free to rename it properly. But it's basically used to create metrics based on cached data IIRC

apollo-router/src/plugins/response_cache/plugin.rs
bnjjj: Not intentional, just a bad merge I think

apollo-router/src/plugins/response_cache/plugin.rs
bnjjj: We don't use field.definition because we need the directive set in supergraph schema not in the executable document

bnjjj: I added a comment

apollo-router/src/plugins/response_cache/plugin.rs
bnjjj: yes I think in a follow up PR we should move this logic when getting the directive data directly in the QP codebase and give the list of cache tags in `FetchNode` if we can.

apollo-router/src/plugins/response_cache/plugin.rs
bnjjj: Fixed, it's uncommented

apollo-router/src/plugins/response_cache/invalidation.rs
SimonSapin: Let’s remove it

apollo-router/src/plugins/response_cache/invalidation_endpoint.rs
SimonSapin: If config has different passwords per subgraph, a single auth header can never match all of them so I’ll change the logic to "any". We can revisit in a future PR (e.g. finding a way for the request to specify auth per subgraph)

SimonSapin (approved):

apollo-bot2:

PR #7711, opened 2025-06-17
Use new subgraph mocking functionality for canned test supergraph

goto-bus-stop:
Many of our tests use a simple "canned" default schema. This PR replaces the hardcoded query/response mocks with the new subgraph mocking functionality that uses a GraphQL execution engine for that schema.

This is more future-proof in case query plans change in benign ways.

Additionally:
- `axum_factory::tests::defer_is_not_buffered` was using a query that didn't have existing mocks, but now it works
- Some studio metrics tests were using a query that didn't have existing mocks, so they were always reporting `has_errors: true`, now it works


lrlna (approved):
🙏

PR #7712, opened 2025-06-17
[Error Counting] Add ID to graphql::Error

rregitsky:
This makes two major changes to `graphql::Error`:
 1. Adds an `apollo_id` UUID to errors to allow us to uniquely identify Errors. These IDs are randomly generated on construction unless one is specified by the input. Its setter is private so that Error construction is forced through `new()` where we randomly generate the ID. The getter is public through `apollo_id()`. This value is _not_ serialized as we'd prefer not to expose the value the user. This is not for security reasons, but more because it has no value outside of the router context.
  2.  Converts all `Error` construction into builders. The `apollo_id` setter is private so we are forced to avoid using the constructor.

Other minor, but notable changes:
 - `apollo.router.operations.error` measurements with  the `_entity` path will now emit using `_entity` path rather than being replaced with the entity's name. This is to prevent double counting when the error counting migration is introduced in part 3. We will attempt to fix this behavior in a fast follow.
 - Removes `Default` trait from `Error` to force a random `apollo_id` to be generated
 - Makes `extension_code` no longer required when building an `Error` as many places weren't setting one with the constructor. We can revisit making it required once all of those places have proper error codes.

Part 2 of a split from https://github.com/apollographql/router/pull/7357. Part 1 can be found here: https://github.com/apollographql/router/pull/7699

---

**Checklist**

- Tests added and passing[^4]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/axum_factory/tests.rs
```diff
@@ -1066,15 +1067,17 @@ async fn response_failure() -> Result<(), ApolloRouterError> {
...
-        }
-        .to_response()
-    );
+    let mut expected_response = crate::error::FetchError::SubrequestHttpError {
+        status_code: Some(200),
+        service: "Mock service".to_string(),
+        reason: "Mock error".to_string(),
+    }
+    .to_response();
+    // Overwrite error IDs to avoid random Uuid mismatch
```
timbotnik: This overwriting feels a little messy - alternatives come to mind:
- Mock out the ID property before comparing (I don't know if this is "a thing" in Rust)
- Implement a compare function that ignores the IDs for testing
- Allow passing in the ID to the constructor (this might not be a good idea b/c Rust - there are no default arguments so you'd have to change every call site to pass `None` 🙄 )

apollo-router/src/graphql/mod.rs
```diff
@@ -196,13 +227,47 @@ impl Error {
...
+            message, locations, path, None, extensions,
+            None, // apollo_id is not serialized, so it will never exist in a serialized vc error
+        ))
+    }
+
+    /// Extract the error code from [`Error::extensions`] as a String if it is set.
+    pub fn extension_code(&self) -> Option<String> {
+        self.extensions.get("code").and_then(|c| match c {
+            Value::String(s) => Some(s.as_str().to_owned()),
+            Value::Bool(b) => Some(format!("{b}")),
```
timbotnik: Seems overly generous to provide a boolean implementation here, but maybe it will help someone to find a bug at some point (vs. just putting bools below along with the `None` case)

apollo-router/src/plugins/authentication/tests.rs
```diff
@@ -233,6 +233,8 @@ async fn it_rejects_when_there_is_no_auth_header() {
     let expected_error = graphql::Error::builder()
         .message("The request is not authenticated")
         .extension_code("AUTH_ERROR")
+        // Overwrite ID to avoid random Uuid mismatch
+        .apollo_id(response.errors[0].apollo_id())
```
timbotnik: Would be nice if you could mock this out somehow rather than having to override it everywhere.

apollo-router/src/services/layers/apq.rs
```diff
@@ -629,7 +609,10 @@ mod apq_tests {
     }
 
     fn assert_error_matches(expected_error: &Error, res: Response) {
```
timbotnik: Maybe we could elevate this method to a shared location for use in other tests?

timbotnik (commented):
Behavioral changes seem good, just wondering if we can get abstract away some of the boilerplate overwriting of null IDs in tests, which seems right now a bit heavy-handed and could be more DRY.

apollo-router/src/axum_factory/tests.rs
rregitsky: I had tried looking into alternatives, to no avail
 > Mock out the ID property before comparing (I don't know if this is "a thing" in Rust)

I'm probably not understanding, but that's essentially what `with_null_id()` is, no? I looked a bit into mocking out the Uuid generator, but I think we'd have to pass a generator in to the builder/new method which feels rough

> Implement a compare function that ignores the IDs for testing

Probably the simplest solution here, although I'm not sure the best place to store it in Rust. Test util, or a test-only fn on the struct. 
I looked into a test-only partial equality, but there are times where we want the ID (e.g. testing error counting) so to me this is too heavy handed.

> Allow passing in the ID to the constructor (this might not be a good idea b/c Rust - there are no default arguments so you'd have to change every call site to pass None 🙄 )

It can already be passed to the builder and this PR does convert every constructor to a builder method. But I'm not sure what the benefit of that is? The problem is that most of these tests we don't have control over Error construction so we need to either copy the ID to expected or null both IDs out before comparison.

rregitsky: I decided to pull it all into 3 macros. Wasn't exactly sure the right place for them, so they're in `graphql/mod.rs` where Error is defined. Lemme know what you think.

There are a few places where we still have to do a bit of ugliness because of either a snapshot test or integration tests where we aren't in the `test` config so macros aren't available.

apollo-router/src/query_planner/fetch.rs
```diff
@@ -362,16 +362,21 @@ impl FetchNode {
...
-                                        path: Some(Path::from_iter(
-                                            values_path.0.iter().chain(&path.0[2..]).cloned(),
-                                        )),
-                                        message: error.message.clone(),
-                                        extensions: error.extensions.clone(),
-                                    })
+                                    errors.push(
+                                        Error::builder()
+                                            .locations(error.locations.clone())
+                                            // append to the entitiy's path the error's path without
```
timbotnik:
```suggestion
                                            // append to the entity's path the error's path without
```

goto-bus-stop:
Is there some background documentation/discussion on what these IDs are for, and why `graphql::Error` is the right place to keep them?

apollo-router/src/graphql/mod.rs
```diff
@@ -52,7 +54,7 @@ pub struct Location {
 /// as may be found in the `errors` field of a GraphQL [`Response`].
 ///
 /// Converted to (or from) JSON with serde.
-#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize, Default)]
```
BrynCooke, line 55: Removing Default is a public API change. We should not do this.

apollo-router/src/plugins/connectors/handle_responses.rs
```diff
@@ -1428,6 +1428,10 @@ mod tests {
         ])
         .unwrap();
 
+        // Overwrite error IDs to avoid random Uuid mismatch
```
BrynCooke: Can you explain more about why this is needed?

BrynCooke (changes requested):
Only a couple of comments!

timbotnik:
> Is there some background documentation/discussion on what these IDs are for, and why `graphql::Error` is the right place to keep them?

We will need a “strong identity” with which errors to count / not / count / redact in each router execution layer that is maintained in the response context.  This means we’ll need to mark each error with some kind of internal error id which can then be added to a list of “already counted” errors.

apollo-router/src/plugins/connectors/handle_responses.rs
timbotnik: Fleshed out the comments.

apollo-router/src/graphql/mod.rs
timbotnik: Let me know if you are happy with the way I changed this.

BrynCooke (approved):

PR #7715, opened 2025-06-17
Add ->ne connectors method

andrewmcgivery:
Add ->ne connectors method

---

**Checklist**

- Tests added and passing[^4]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-federation/src/connectors/json_selection/methods/public/ne.rs
```diff
@@ -0,0 +1,101 @@
...
+    use crate::selection;
+
+    #[test]
+    fn ne_should_return_false_when_applied_to_equals_argument() {
+        assert_eq!(
+            selection!(
+                r#"
+                    result: value->ne(123)
+                "#
+            )
```
dylan-apollo, line 132: Can you run some tests with different types? I'm specifically worried about the numbers, since there are 3 different types of numbers within `serde_json_bytes` and I'm not sure if eq is defined correctly.

Like, is 1.0 == 1? And -1.0 == -1?

andrewmcgivery: Fixed!

apollo-federation/src/connectors/json_selection/methods/public/eq.rs
```diff
@@ -48,12 +55,19 @@ fn eq_method(
...
-    _method_args: Option<&MethodArgs>,
+    method_args: Option<&MethodArgs>,
     _input_shape: Shape,
     _dollar_shape: Shape,
     _named_var_shapes: &IndexMap<&str, Shape>,
     source_id: &SourceId,
 ) -> Shape {
+    if method_args.and_then(|args| args.args.first()).is_none() {
+        return Shape::error(
+            format!("Method ->{} requires one argument", method_name.as_ref()),
```
dylan-apollo, line 94: Should error on >1 arguments as well.

dylan-apollo: Should we also error any time two types are not the same type? Maybe we can reuse the "is_comparable" or whatever logic from the other methods? Just seems like an obvious error to compare an object to a string that we should catch at compile time.

Speaking of, I don't see a test for comparing two different types 😁

andrewmcgivery: So With `eq` and `ne`, the logic is much simpler because we can actually just compare the `Value` type directly against each other when doing `==` or `!=`. With anything else, we had to be more explicit because `Value` doesn't implement `PartialOrd`.

So it doesn't have the same type contraints. But maybe it should? 🤔

benjamn: We can use `Shape::error_with_partial`, where the partial value is `false`, to provide an error that essentially amounts to a warning. Regardless what you think about whether we should error, the logical answer to comparing values with different known shapes is `false`.

apollo-federation/src/connectors/json_selection/methods/public/ne.rs
benjamn: The `->eq` and `->ne` methods need to work for all JSON values (including objects and arrays!), so please add a bunch more tests here exercising values other than numbers.

benjamn:
Since my review comments may have gotten buried:
* https://github.com/apollographql/router/pull/7715#discussion_r2159228087
* https://github.com/apollographql/router/pull/7715#discussion_r2159220139

apollo-federation/src/connectors/json_selection/methods/public/eq.rs
```diff
@@ -48,17 +72,53 @@ fn eq_method(
...
+        dollar_shape,
+        named_var_shapes,
+        source_id,
+    );
+
+    // Ensures that the arguments are of the same type... this includes covering cases like int/float and unknown/name
+    if !(input_shape.accepts(&arg_shape) || arg_shape.accepts(&input_shape)) {
+        return Shape::error_with_partial(
+            format!(
+                "Method ->{} requires the applied to value and argument to be the same type to be comparable.",
```
dylan-apollo:
```suggestion
                "Method ->{} can only compare values of the same type.",
```

"applied to" is very internal terminology

apollo-federation/src/connectors/json_selection/methods/public/ne.rs
```diff
@@ -0,0 +1,504 @@
...
+        dollar_shape,
+        named_var_shapes,
+        source_id,
+    );
+
+    // Ensures that the arguments are of the same type... this includes covering cases like int/float and unknown/name
+    if !(input_shape.accepts(&arg_shape) || arg_shape.accepts(&input_shape)) {
+        return Shape::error_with_partial(
+            format!(
+                "Method ->{} requires the applied to value and argument to be the same type to be comparable.",
```
dylan-apollo:
```suggestion
                "Method ->{} can only compare values of the same type.",
```

apollo-federation/src/connectors/json_selection/methods/public/ne.rs
```diff
@@ -0,0 +1,504 @@
...
+    method_name: &WithRange<String>,
+    method_args: Option<&MethodArgs>,
+    data: &JSON,
+    vars: &VarsWithPathsMap,
+    input_path: &InputPath<JSON>,
+) -> (Option<JSON>, Vec<ApplyToError>) {
+    if let Some(MethodArgs { args, .. }) = method_args {
+        if let [arg] = args.as_slice() {
+            let (value_opt, arg_errors) = arg.apply_to_path(data, vars, input_path);
+            let mut apply_to_errors = arg_errors;
```
dylan-apollo:
```suggestion
            let (value_opt, mut apply_to_errors) = arg.apply_to_path(data, vars, input_path);
```

Not a suggestion, really, just an option. I don't feel strongly one way or the other.

apollo-federation/src/connectors/json_selection/methods/public/ne.rs
```diff
@@ -0,0 +1,504 @@
...
+                "Method ->{} requires exactly one argument",
+                method_name.as_ref()
+            ),
+            input_path.to_vec(),
+            method_name.range(),
+        )],
+    )
+}
+#[allow(dead_code)] // method type-checking disabled until we add name resolution
+fn ne_shape(
```
dylan-apollo, line 72: This is _identical_ to `eq_shape` with the exception of the bug I'm about to point out below. So I feel like they should be the same function so they're guaranteed to always agree on what is our equivalent of implementing `PartialEq`.

apollo-federation/src/connectors/json_selection/methods/public/ne.rs
```diff
@@ -0,0 +1,504 @@
...
+    );
+
+    // Ensures that the arguments are of the same type... this includes covering cases like int/float and unknown/name
+    if !(input_shape.accepts(&arg_shape) || arg_shape.accepts(&input_shape)) {
+        return Shape::error_with_partial(
+            format!(
+                "Method ->{} requires the applied to value and argument to be the same type to be comparable.",
+                method_name.as_ref()
+            ),
+            Shape::bool_value(false, method_name.shape_location(source_id)),
```
dylan-apollo: This is actually always `true` in this case, right? Two different types will never be equal?


```suggestion
            Shape::bool_value(true, method_name.shape_location(source_id)),
```

dylan-apollo (changes requested):
Spotted one baby bug and added a few style options/suggestions.

I'm also a little worried that our type checking will be too aggressive with no way to opt out. Imagine a type can be `{a: String, b: String}` _or_ `{a: int, b: int}`, but you're unable to represent that.. like with GraphQL input types. Instead you'd have to have something like `{a: String | int, b: String | int}`. Then when you type check `a->eq(b)`, you're going to get an error, because you _could_ be comparing `String` and `int`. You'd have no way to silence this error and unblock your build.

So our options are:

1. Remove this check for now, and in the future if we can be smarter about types, we add it back in as a breaking change.
2. Leave it in and if a user encounters it, we'll finally give them a way to `# type: ignore` or whatever
3. Find a way to make _some_ type errors, like this one, warnings so that while they may annoy users, at least they won't block a build.

apollo-federation/src/connectors/json_selection/methods/public/ne.rs
andrewmcgivery: My preference would be to keep these separate. The one line that is different feels like it would be more complicated to account for than to just have the two different functions. Though I guess this may change depending on the larger worry comment you left above.

dylan-apollo: I'm not sure how worthwhile it is to have the partial literal bool? If it's one of the other, I'd prefer we have some more code consistency than add a bit of data we may or may not use in checks. Odds are, we'll just show the error and stop processing.

andrewmcgivery: This would make it inconsistent with the other comparison methods which all return a `error_with_partial` which @benjamn suggested we should do on these in this PR.

So... if y'all collectively decide we don't want these partials after all, cool, but I'd like to avoid re-factoring circles on these PRs. 😅

apollo-federation/src/connectors/json_selection/methods/public/eq.rs
```diff
@@ -48,17 +72,53 @@ fn eq_method(
...
+    if !(input_shape.accepts(&arg_shape) || arg_shape.accepts(&input_shape)) {
+        return Shape::error_with_partial(
+            format!(
+                "Method ->{} can only compare values of the same type.",
+                method_name.as_ref()
+            ),
+            Shape::bool_value(false, method_name.shape_location(source_id)),
+            method_name.shape_location(source_id),
+        );
+    }
```
benjamn:
```suggestion
    if !(input_shape.accepts(&arg_shape) || arg_shape.accepts(&input_shape)) {
        return Shape::error_with_partial(
            format!(
                "Method ->{} can only compare values of the same type. Got {input_shape} == {arg_shape}.",
                method_name.as_ref(),
            ),
            Shape::bool_value(false, method_name.shape_location(source_id)),
            method_name.shape_location(source_id),
        );
    }
```
A little more information for debugging?

Although we could probably get away with just returning the `false` shape with no error here, I think this level of comparability checking makes sense, since the error is likely going to end up in a union like
```
One<Bool, Error<"Method ->eq can only compare values of the same type...", false>>
```
which is a pretty plausible description of what can happen at runtime. Of course we'll have to make some decisions about how we report unions that involve errors, but that's a general issue we have to figure out.

That said, if we find that developers are confused by these errors (even if they're just warnings), we can remove the comparability check (at least for methods like `->eq` and `->ne` that can reasonably return a `Bool` or `false` when the input/arg have totally different shapes).

apollo-federation/src/connectors/json_selection/methods/public/eq.rs
```diff
@@ -28,9 +29,32 @@ fn eq_method(
...
+                            apply_to_errors.push(err);
+                            return None;
+                        }
+                    };
 
-            return (Some(JSON::Bool(matches)), arg_errors);
+                    Some(JSON::Bool(left == right))
+                }
+                // Everything else
+                _ => Some(JSON::Bool(&value == data)),
```
benjamn, line 54: Since we're not erroring/warning at runtime, in some sense there's no error/warning to anticipate at shape processing time. I still think it could be helpful to warn in the `{ne,eq}_shape` functions, but we can go either way, depending on how confusing this ends up being for developers.

benjamn (approved):
I'm happy with the current state (no blocking comments below), though there's a possibility the comparability checking will be more confusing than helpful (compared to always reporting `Bool`). On balance, it's easier to remove an error/warning than to add it later, so let's see how it goes.

apollo-federation/src/connectors/json_selection/methods/public/eq.rs
andrewmcgivery: Done!

apollo-federation/src/connectors/json_selection/methods/public/ne.rs
dylan-apollo: We'll keep @benjamn's suggestion on this one since he's the one whose gonna be enabling this type-checking and ultimately surfacing the errors. We also have several months of previews to change it if we want to 😁.

PR #7717, opened 2025-06-17
gt connectors method

andrewmcgivery:
gt connectors method

---

**Checklist**

- Tests added and passing[^4]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-federation/src/connectors/json_selection/methods/public/gt.rs
```diff
@@ -0,0 +1,327 @@
...
+                "Method ->{} requires exactly one argument",
+                method_name.as_ref()
+            ),
+            input_path.to_vec(),
+            method_name.range(),
+        )],
+    )
+}
+
+#[allow(dead_code)] // method type-checking disabled until we add name resolution
```
dylan-apollo, line 92: Are we adding an AI on the method type checking code to fill this in? Or should we fill it in here to not risk missing it there?

apollo-federation/src/connectors/json_selection/methods/public/gt.rs
```diff
@@ -0,0 +1,327 @@
...
+/// $("a")->gt("b")   results in false
+/// $("c")->gt("b")   results in true
+fn gt_method(
+    method_name: &WithRange<String>,
+    method_args: Option<&MethodArgs>,
+    data: &JSON,
+    vars: &VarsWithPathsMap,
+    input_path: &InputPath<JSON>,
+) -> (Option<JSON>, Vec<ApplyToError>) {
+    if let Some(first_arg) = method_args.and_then(|args| args.args.first()) {
```
dylan-apollo: I prefer `let Some(..) = ...  else { return }` to denest, personally.

apollo-federation/src/connectors/json_selection/methods/public/gt.rs
```diff
@@ -0,0 +1,327 @@
...
+                (JSON::Number(left), JSON::Number(right)) => {
+                    left.as_f64().unwrap_or(0.0) > right.as_f64().unwrap_or(0.0)
+                }
+                // String comparisons
+                (JSON::String(left), JSON::String(right)) => left > right,
+                // Boolean comparisons
+                (JSON::Bool(left), JSON::Bool(right)) => left > right,
+                // Null comparisons (null == null)
+                (JSON::Null, JSON::Null) => false,
+                // Mixed types or uncomparable types (including arrays and objects) return false
```
dylan-apollo: I would error on these instead, doesn't make sense to compare them in this way.

apollo-federation/src/connectors/json_selection/methods/public/gt.rs
andrewmcgivery: I don't follow. Are you suggesting that the shape isn't just `bool`?

apollo-federation/src/connectors/json_selection/methods/public/gt.rs
andrewmcgivery: To be clear, on both bool and null?

apollo-federation/src/connectors/json_selection/methods/public/gt.rs
dylan-apollo: These methods don't _just_ produce the output shape, they're supposed to validate the inputs as well. Maybe a better name for them all would be `validate_<method_name>`—but `Error` is a `Shape` so...

Anyway, this should also check that there's only 1 argument, and that argument is comparable with the input shape.

apollo-federation/src/connectors/json_selection/methods/public/gt.rs
dylan-apollo: Yes. I think only strings & numbers have a clear < > relationship

andrewmcgivery: Done

apollo-federation/src/connectors/json_selection/methods/public/gt.rs
andrewmcgivery: Done!

apollo-federation/src/connectors/json_selection/methods/public/gt.rs
andrewmcgivery: Done!

apollo-federation/src/connectors/json_selection/methods/public/gt.rs
```diff
@@ -0,0 +1,348 @@
...
+                (JSON::Number(left), JSON::Number(right)) => {
+                    left.as_f64().unwrap_or(0.0) > right.as_f64().unwrap_or(0.0)
+                }
+                // String comparisons
+                (JSON::String(left), JSON::String(right)) => left > right,
+                // Mixed types or uncomparable types (including arrays and objects) return false
+                _ => {
+                    apply_to_errors.push(ApplyToError::new(
+                        format!(
+                            "Method ->{} can directly compare numbers and strings. Either a mix of these was provided or something else such as an array, object, null, or bool. Found: {} > {}",
```
dylan-apollo:
```suggestion
                            "Method ->{method_name} can only compare numbers and strings. Found: {data} > {value}",
```

dylan-apollo: Trying to make more succinct and inlining variables (not sure if `method_name` can be)

apollo-federation/src/connectors/json_selection/methods/public/gt.rs
```diff
@@ -0,0 +1,348 @@
...
+                            method_name.as_ref(),
+                            data,
+                            value
+                        ),
+                        input_path.to_vec(),
+                        method_name.range(),
+                    ));
+
+                    false
```
dylan-apollo: Shouldn't this return `None` to match other methods so that we don't continue processing?

apollo-federation/src/connectors/json_selection/methods/public/gt.rs
```diff
@@ -0,0 +1,348 @@
...
+    dollar_shape: Shape,
+    named_var_shapes: &IndexMap<&str, Shape>,
+    source_id: &SourceId,
+) -> Shape {
+    let Some(first_arg) = method_args.and_then(|args| args.args.first()) else {
+        return Shape::error(
+            format!("Method ->{} requires one argument", method_name.as_ref()),
+            method_name.shape_location(source_id),
+        );
+    };
```
dylan-apollo, line 117: Also add a check on `.len()` in case multiple arguments were passed. Alternatively you can pattern match on `&[first_arg]` potentially to both get the first _and_ check there are no others at the same time.

We do this so often in methods we probably want a helper at some point.

apollo-federation/src/connectors/json_selection/methods/public/gt.rs
```diff
@@ -0,0 +1,348 @@
...
+        input_shape.clone(),
+        dollar_shape,
+        named_var_shapes,
+        source_id,
+    );
+
+    match (arg_shape.case(), input_shape.case()) {
+        (ShapeCase::Unknown, ShapeCase::Unknown)
+        | (ShapeCase::Int(_), ShapeCase::Int(_))
+        | (ShapeCase::Float, ShapeCase::Float)
```
dylan-apollo: (Int, Float) and (Float, Int) are okay too, yeah?

apollo-federation/src/connectors/json_selection/methods/public/gt.rs
```diff
@@ -0,0 +1,348 @@
...
+        | (ShapeCase::String(_), ShapeCase::String(_)) => {
+            Shape::bool(method_name.shape_location(source_id))
+        }
+        _ => Shape::error(
+            format!(
+                "Method ->{} can directly compare numbers and strings. Either a mix of these was provided or something else such as an array, object, null, or bool. Found: {} > {}",
+                method_name.as_ref(),
+                input_shape,
+                arg_shape
```
dylan-apollo:
```suggestion
                "Method ->{} can only compare two numbers or two strings. Found {input_shape} > {arg_shape}",
                method_name.as_ref(),
```

dylan-apollo (changes requested):

apollo-federation/src/connectors/json_selection/methods/public/gt.rs
andrewmcgivery: `method_name` can't be inlined but the other two can. :)

andrewmcgivery: Made the adjustment to the wording and inlining the values!

apollo-federation/src/connectors/json_selection/methods/public/gt.rs
andrewmcgivery: Good call

apollo-federation/src/connectors/json_selection/methods/public/gt.rs
andrewmcgivery: Done

apollo-federation/src/connectors/json_selection/methods/public/gt.rs
andrewmcgivery: Good catch!

apollo-federation/src/connectors/json_selection/methods/public/gt.rs
andrewmcgivery: Done

apollo-federation/src/connectors/json_selection/methods/public/gt.rs
dylan-apollo: Not saying to change it, but just for completeness wanted to show what the pattern matching version was I was talking about:

```rust
let Some([arg]) = method_args.map(|args| args.args.as_slice()) else {
        return Shape::error(
            format!(
                "Method ->{} requires only one argument, but {arg_count} were provided",
                method_name.as_ref(),
                arg_count = method_args.map(|args| args.args.len()).unwrap_or_default()
            ),
            vec![],
        );
    };
```

So all at once this says method_args should be `Some` and should contain exactly 1 element.

apollo-federation/src/connectors/json_selection/methods/public/gt.rs
```diff
@@ -0,0 +1,330 @@
...
+
+    let arg_shape = first_arg.compute_output_shape(
+        input_shape.clone(),
+        dollar_shape,
+        named_var_shapes,
+        source_id,
+    );
+
+    match (arg_shape.case(), input_shape.case()) {
+        (ShapeCase::Unknown, ShapeCase::Unknown)
```
dylan-apollo: Oh this is gonna get yucky—so we definitely will want a shared function to use with the other comparable methods. Unknown is also safe to compare against anything else that is comparable.

So maybe the whole match looks something like:

```rust
(ShapeCase::Unknown, ShapeCase::Unknown | ShapeCase::Int(_) | ShapeCase::Float | ShapeCase::String(_))
        (ShapeCase::Int(_), ShapeCase::Int(_) | ShapeCase::Float | ShapeCase::Unknown)
        | (ShapeCase::Float, ShapeCase::Int(_) | ShapeCase::Float | ShapeCase::Unknown)
        | (ShapeCase::String(_), ShapeCase::String(_) | ShapeCase::Unknown) => {
```

If we don't know a shape case (Unknown) on either side, we will let you compare it to anything else as long as that other thing is a valid type for this function.

Probably want some exhaustive tests for this shape function to confirm that.

apollo-federation/src/connectors/json_selection/methods/public/gt.rs
andrewmcgivery: Thats interesting... I can understand `[arg]` grabbing the first argument my brain isn't comprehending how that ensures only one. Is it cause the `let Some([arg])` is specifying that it's an array with only 1 element?

apollo-federation/src/connectors/json_selection/methods/public/gt.rs
andrewmcgivery: Ok, I implemented this as a shared function... when we're happy with it in this PR (and get it merged), I'll utilize it in my other PRs. :)

apollo-federation/src/connectors/json_selection/methods/common.rs
```diff
@@ -0,0 +1,71 @@
+use shape::ShapeCase;
+
+pub(crate) fn is_comparable_shape_combination(shape1: &ShapeCase, shape2: &ShapeCase) -> bool {
+    matches!(
+        (shape1, shape2),
+        // Unknown is compatible with comparable types
```
benjamn: `ShapeCase::Name` is also compatible with comparable values.

apollo-federation/src/connectors/json_selection/methods/common.rs
```diff
@@ -0,0 +1,71 @@
...
+    #[case(&ShapeCase::Unknown, &ShapeCase::Int(Some(1)), true)]
+    #[case(&ShapeCase::Float, &ShapeCase::Float, true)]
+    #[case(&ShapeCase::Float, &ShapeCase::Unknown, true)]
+    #[case(&ShapeCase::Unknown, &ShapeCase::Float, true)]
+    #[case(&ShapeCase::Int(Some(1)), &ShapeCase::Float, true)]
+    #[case(&ShapeCase::Float, &ShapeCase::Int(Some(1)), true)]
+    #[case(&ShapeCase::String(Some("a".to_string())), &ShapeCase::String(Some("a".to_string())), true)]
+    #[case(&ShapeCase::String(Some("a".to_string())), &ShapeCase::Unknown, true)]
+    #[case(&ShapeCase::Unknown, &ShapeCase::String(Some("a".to_string())), true)]
+    fn test_is_comparable_shape_combination_positive_cases(
```
benjamn, line 73: Let's add some `ShapeCase::Name` tests here.

apollo-federation/src/connectors/json_selection/methods/public/gt.rs
```diff
@@ -0,0 +1,324 @@
...
+    };
+
+    let (value_opt, arg_errors) = first_arg.apply_to_path(data, vars, input_path);
+    let mut apply_to_errors = arg_errors;
+    // We have to do this because Value doesn't implement PartialOrd
+    let matches = value_opt.and_then(|value| {
+        match (data, &value) {
+            // Number comparisons
+            (JSON::Number(left), JSON::Number(right)) => Some(JSON::Bool(
+                left.as_f64().unwrap_or(0.0) > right.as_f64().unwrap_or(0.0),
```
benjamn: Let's return `false` and error instead of defaulting to `0.0`. See also https://github.com/apollographql/router/pull/7722#discussion_r2159145125.

apollo-federation/src/connectors/json_selection/methods/common.rs
```diff
@@ -0,0 +1,71 @@
...
+    matches!(
+        (shape1, shape2),
+        // Unknown is compatible with comparable types
+        (ShapeCase::Unknown, ShapeCase::Int(_) | ShapeCase::Float | ShapeCase::String(_) | ShapeCase::Unknown) |
+          (ShapeCase::Int(_) | ShapeCase::Float | ShapeCase::String(_), ShapeCase::Unknown) |
+          // Numeric types are cross-compatible
+          (ShapeCase::Int(_), ShapeCase::Int(_) | ShapeCase::Float) |
+          (ShapeCase::Float, ShapeCase::Int(_) | ShapeCase::Float) |
+          // String types
+          (ShapeCase::String(_), ShapeCase::String(_))
```
benjamn: I find this style hard to read and not convincingly/obviously correct.

Besides `ShapeCase::Name`, one thing that should work but does not: `ShapeCase::One` unions of comparable number shapes are not handled. That's why I would recommend using the `Shape::accepts` method instead of matching on cases directly.

Here's my take:
```rust
if Shape::float([]).accepts(shape1) {
    // Note that Float accepts any number, so you don't have to spell out the other cases
    Shape::float([]).accepts(shape2) || shape2.accepts(Shape::unknown([]))
} else if Shape::string([]).accepts(shape1) {
    Shape::string([]).accepts(shape2) || shape2.accepts(Shape::unknown([]))
} else if shape1.accepts(Shape::unknown([])) {
    Shape::float([]).accepts(shape2)
    || Shape::string([]).accepts(shape2)
    || shape2.accepts(Shape::unknown([]))
} else {
    false
}
```
The `shape.accepts(Shape::unknown([]))` pattern is an idiomatic way to test whether `shape` is `Unknown` or an unresolved `Name`.

benjamn (changes requested):
I'm probably going to leave similar comments on each of the comparison method PRs, so feel free to adapt this feedback to the others if you see an opportunity!

apollo-federation/src/connectors/json_selection/methods/public/gt.rs
benjamn: Less yucky IMO: https://github.com/apollographql/router/pull/7717#discussion_r2159194424

apollo-federation/src/connectors/json_selection/methods/public/gt.rs
andrewmcgivery: `false` or `None`? I got the impression from feedback from @dylan-apollo that we should return `None` when we hit an error like this.

andrewmcgivery: I added this in (with `None`) but I wasn't able to create any tests that actually trigger a failing 😅 It seems like the `as_f64()` method never actually returns `None`. So pretty big caveat there.

EDIT: Or there is some crazy edge case that I just wasn't able to find... I tried very large numbers and it still converted successfully. LLM also could not find a case where it returns `None` 😅

apollo-federation/src/connectors/json_selection/methods/common.rs
dylan-apollo: I feel like we might want some `is_number()`, `is_string()` and `is_unknown()` helpers at some point to make this easier to read. Not immediately obvious to most what `.accepts()` means probably. Especially for unknown.

BTW what _should_ happen when the input is `None`? That generates an error, right? I'm worried about the static check ending up in an error because a property is optional.

apollo-federation/src/connectors/json_selection/methods/public/gt.rs
dylan-apollo: Yeah, it's pattern matching against the whole slice: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=6262e0277b8522e5fa6e6498b0589a0c

apollo-federation/src/connectors/json_selection/methods/common.rs
benjamn: I agree about adding helpers, and I agree this particular trick is not obvious, but I still think it's worth understanding how `Shape::accepts` works, since it solves a number of problems.

Intuitively, `s1.accepts(s2)` means if you had a variable with shape/type `s1`, you could safely assign a value with shape/type `s2` to it. So it's like `s2 extends s1` in TypeScript. The only variable shapes that allow assigning `Unknown` to them are `Unknown` and any `Name`. So `s1.accepts(Shape::unknown([]))` could be turned into a helper method like `s1.could_be_anything()` (which is true of both `Unknown` and `Name` shapes).

As for what should happen when input is `None`, I believe I can say that will never happen in practice, because we implement the `None` short-circuiting outside of any particular `->` method. So while `None` exists in the shape system, and might appear in the input or result shapes of a method, `->` methods should not have to worry about receiving it as an input to their `*_shape` function. If they did, they'd probably immediately return `(None, vec![])` anyway.

andrewmcgivery: I've got this integrated now with a bunch of test cases!

apollo-federation/src/connectors/json_selection/methods/common.rs
```diff
@@ -0,0 +1,87 @@
...
+    #[case(Shape::name("test", []), Shape::name("test", []), true)]
+    #[case(Shape::int([]), Shape::float([]), true)]
+    #[case(Shape::int([]), Shape::int([]), true)]
+    #[case(Shape::int([]), Shape::name("test", []), true)]
+    #[case(Shape::int([]), Shape::unknown([]), true)]
+    fn test_is_comparable_shape_combination_2_positive_cases(
+        #[case] shape1: Shape,
+        #[case] shape2: Shape,
+        #[case] expected: bool,
+    ) {
```
dylan-apollo:
```suggestion
    #[case(Shape::float([]), Shape::float([]))]
    #[case(Shape::float([]), Shape::unknown([]))]
    #[case(Shape::float([]), Shape::name("test", []))]
    #[case(Shape::float([]), Shape::int([]))]
    #[case(Shape::string([]), Shape::string([]))]
    #[case(Shape::string([]), Shape::unknown([]))]
    #[case(Shape::string([]), Shape::name("test", []))]
    #[case(Shape::unknown([]), Shape::float([]))]
    #[case(Shape::unknown([]), Shape::int([]))]
    #[case(Shape::unknown([]), Shape::string([]))]
    #[case(Shape::unknown([]), Shape::unknown([]))]
    #[case(Shape::unknown([]), Shape::name("test", []))]
    #[case(Shape::name("test", []), Shape::float([]))]
    #[case(Shape::name("test", []), Shape::string([]))]
    #[case(Shape::name("test", []), Shape::unknown([]))]
    #[case(Shape::name("test", []), Shape::int([]))]
    #[case(Shape::name("test", []), Shape::name("test", []))]
    #[case(Shape::int([]), Shape::float([]))]
    #[case(Shape::int([]), Shape::int([]))]
    #[case(Shape::int([]), Shape::name("test", []))]
    #[case(Shape::int([]), Shape::unknown([]))]
    fn test_is_comparable_shape_combination_2_positive_cases(
        #[case] shape1: Shape,
        #[case] shape2: Shape,
    ) {
```

apollo-federation/src/connectors/json_selection/methods/common.rs
```diff
@@ -0,0 +1,87 @@
...
+        let actual = is_comparable_shape_combination(&shape1, &shape2);
+        assert_eq!(expected, actual);
```
dylan-apollo:
```suggestion
        assert!(is_comparable_shape_combination(&shape1, &shape2));
        assert!(is_comparable_shape_combination(&shape2, &shape1));
```
The opposite should always be true, yeah? Might mean we can eliminate some `rstest` up above, too, so it just lists the two shapes which should be comparable.

apollo-federation/src/connectors/json_selection/methods/common.rs
```diff
@@ -0,0 +1,87 @@
...
+    #[case(Shape::unknown([]), Shape::null([]), false)]
+    #[case(Shape::bool([]), Shape::float([]), false)]
+    #[case(Shape::bool([]), Shape::string([]), false)]
+    #[case(Shape::bool([]), Shape::int([]), false)]
+    #[case(Shape::bool([]), Shape::unknown([]), false)]
+    #[case(Shape::bool([]), Shape::name("test", []), false)]
+    fn test_is_comparable_shape_combination_2_negative_cases(
+        #[case] shape1: Shape,
+        #[case] shape2: Shape,
+        #[case] expected: bool,
```
dylan-apollo: Same suggestions here, since they're all `false` we can get rid of that param and use `assert!(!is_comparable_shape_combination())` for both combinations (eliminating duplicates.

apollo-federation/src/connectors/json_selection/methods/public/gt.rs
```diff
@@ -0,0 +1,356 @@
...
+    // We have to do this because Value doesn't implement PartialOrd
+    let matches = value_opt.and_then(|value| {
+        match (data, &value) {
+            // Number comparisons
+            (JSON::Number(left), JSON::Number(right)) => {
+                let left = match left.as_f64() {
+                    Some(val) => val,
+                    None => {
+                        // Note that we don't have tests for these `None` cases because I can't actually find a case where this ever actually fails
+                        // It seems that the current implementation in serde_json always returns a value
```
dylan-apollo: It only fails on `NaN` I think, so the test will be something like `$(1)->gt(2->div(0))`? Unless of course dividing by zero doesn't result in NaN, but instead returns `None`. Then we don't need this check at all...

apollo-federation/src/connectors/json_selection/methods/common.rs
```diff
@@ -0,0 +1,87 @@
...
+    #[case(Shape::null([]), Shape::name("test", []), false)]
+    #[case(Shape::name("test", []), Shape::bool([]), false)]
+    #[case(Shape::name("test", []), Shape::null([]), false)]
+    #[case(Shape::unknown([]), Shape::bool([]), false)]
+    #[case(Shape::unknown([]), Shape::null([]), false)]
+    #[case(Shape::bool([]), Shape::float([]), false)]
+    #[case(Shape::bool([]), Shape::string([]), false)]
+    #[case(Shape::bool([]), Shape::int([]), false)]
+    #[case(Shape::bool([]), Shape::unknown([]), false)]
+    #[case(Shape::bool([]), Shape::name("test", []), false)]
```
dylan-apollo: I think we also need tests for arrays, objects, `One`, and `None`. That _last_ one shouldn't raise an error, though I think? Unclear. If someone has an optional/nullable int they want to compare, they should be able to do `thing->gt(optionalThing)` or maybe `thing->gt(optionalThing?)`. I don't think there's another way to work around comparing optional numbers?

Unless we want to force them to do `thing->gt(optionalThing ?? 0)`?.

Whatever we choose, we need to create a follow-up action item in the nullability PR or the method shape-checking PR to make the error message helpful.

@benjamn what do you think?

apollo-federation/src/connectors/json_selection/methods/public/gt.rs
andrewmcgivery: Oh I can't even do that cause `div` has a check for 0 😅 

```
Method ->div failed on argument 0
```

apollo-federation/src/connectors/json_selection/methods/common.rs
andrewmcgivery: I had thought @benjamn had said that methods will never actually receive a `None` value 🤔

andrewmcgivery: Got test cases in there for arrays, objects, and one.

apollo-federation/src/connectors/json_selection/methods/public/gt.rs
```diff
@@ -0,0 +1,356 @@
...
+                        apply_to_errors.push(ApplyToError::new(
+                            format!(
+                                "Method ->{} fail to convert applied to value to float.",
+                                method_name.as_ref(),
+                            ),
+                            input_path.to_vec(),
+                            method_name.range(),
+                        ));
+                        return None;
```
dylan-apollo, line 58: This doesn't actually return the apply to errors. Same as below.

Also, we should make a single `value_as_float -> Result<f64, ApplyToError>` helper (in `common`?) we can use for both of these and all the other similar methods coming up.

apollo-federation/src/connectors/json_selection/methods/public/gt.rs
```diff
@@ -0,0 +1,356 @@
...
+                method_name.as_ref()
+            ),
+            Shape::bool(method_name.shape_location(source_id)),
+            method_name.shape_location(source_id),
+        )
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
dylan-apollo: Add some tests for the error cases of shape as well. Don't need to cover all the cases in `is_comparable`, but checks for `->gt` (no args), `->gt()` (empty args) `->gt(1, 2)` (too many args), `->gt({})` (invalid arg)

dylan-apollo (changes requested):

apollo-federation/src/connectors/json_selection/methods/public/gt.rs
andrewmcgivery: It does cause this code is in a `and_then` so it's returning on that, not on the function. The `apply_to_errors` is declared outside of this code block.

This is definitely working cause I'm doing the same thing in the `_` match arm and the error is reflected in the tests.

andrewmcgivery: Refactored the float logic to be a shared function!

apollo-federation/src/connectors/json_selection/methods/public/gt.rs
andrewmcgivery: Added a handful of tests for the shape method!

apollo-federation/src/connectors/json_selection/methods/public/gt.rs
```diff
@@ -0,0 +1,453 @@
...
+                Some(json!({
+                    "result": false,
+                })),
+                vec![],
+            ),
+        );
+    }
+
+    #[test]
+    fn gt_should_return_false_with_error_for_null_values() {
```
dylan-apollo:
```suggestion
    fn gt_should_error_for_null_values() {
```

apollo-federation/src/connectors/json_selection/methods/public/gt.rs
```diff
@@ -0,0 +1,453 @@
...
+        assert!(!result.1.is_empty());
+        assert!(
+            result.1[0]
+                .message()
+                .contains("Method ->gt can only compare numbers and strings. Found: null > null")
+        );
+    }
+
+    #[test]
+    fn gt_should_return_false_with_error_for_boolean_values() {
```
dylan-apollo:
```suggestion
    fn gt_should_error_for_boolean_values() {
```

apollo-federation/src/connectors/json_selection/methods/public/gt.rs
```diff
@@ -0,0 +1,453 @@
...
+        assert!(!result.1.is_empty());
+        assert!(
+            result.1[0]
+                .message()
+                .contains("Method ->gt can only compare numbers and strings. Found: true > false")
+        );
+    }
+
+    #[test]
+    fn gt_should_return_false_with_error_for_arrays() {
```
dylan-apollo:
```suggestion
    fn gt_should_error_for_arrays() {
```

apollo-federation/src/connectors/json_selection/methods/public/gt.rs
```diff
@@ -0,0 +1,453 @@
...
+        assert!(!result.1.is_empty());
+        assert!(
+            result.1[0].message().contains(
+                "Method ->gt can only compare numbers and strings. Found: [1,2,3] > [1,2]"
+            )
+        );
+    }
+
+    #[test]
+    fn gt_should_return_false_with_error_for_objects() {
```
dylan-apollo:
```suggestion
    fn gt_should_error_for_objects() {
```

apollo-federation/src/connectors/json_selection/methods/public/gt.rs
```diff
@@ -0,0 +1,453 @@
...
+
+        assert_eq!(result.0, Some(json!({})),);
+        assert!(!result.1.is_empty());
+        assert!(result.1[0].message().contains(
+            "Method ->gt can only compare numbers and strings. Found: {\"a\":1,\"b\":2} > {\"a\":1}"
+        ));
+    }
+
+    #[test]
+    fn gt_should_return_false_and_error_for_mixed_types() {
```
dylan-apollo:
```suggestion
    fn gt_should_error_for_mixed_types() {
```

apollo-federation/src/connectors/json_selection/methods/public/gt.rs
```diff
@@ -0,0 +1,453 @@
...
+                Shape::bool([get_location()]),
+                [get_location()]
+            )
+        );
+    }
+
+    #[test]
+    fn gt_shape_should_error_on_no_args() {
+        assert_eq!(
+            get_shape(vec![], Shape::string([])),
```
dylan-apollo, line 429: There's a slightly different case, too, where args is `None` which I think is how `->gt` (no parens) would appear?

apollo-federation/src/connectors/json_selection/methods/common.rs
```diff
@@ -0,0 +1,143 @@
...
+pub(crate) fn number_value_as_float(
+    number: &Number,
+    method_name: &WithRange<String>,
+    input_path: &InputPath<JSON>,
+) -> Result<f64, ApplyToError> {
+    match number.as_f64() {
+        Some(val) => Ok(val),
+        None => {
+            // Note that we don't have tests for these `None` cases because I can't actually find a case where this ever actually fails
+            // It seems that the current implementation in serde_json always returns a value
```
benjamn, line 33: Yeah, even if the input integer is very large, the float representation can still represent huge exponential values:
```rust
pub const MAX: f64 = 1.7976931348623157e+308;
pub const MAX: i64 = 9223372036854775807;
```
I'm willing to believe this conversion is essentially infallible, but thanks for adding the `number_value_as_float` helper.

apollo-federation/src/connectors/json_selection/methods/public/gt.rs
benjamn: LGTM

PR #7720, opened 2025-06-17
Promote ->or connectors method to public

andrewmcgivery:
Promote ->or connectors method to public

---

**Checklist**

- Tests added and passing[^4]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-federation/src/connectors/json_selection/methods/public/or.rs
```diff
@@ -67,14 +100,17 @@ fn or_shape(
...
         }
-        ShapeCase::Int(Some(value)) if *value != 0 => {
-            return Shape::bool_value(true, method_name.shape_location(source_id));
-        }
-        ShapeCase::String(Some(value)) if !value.is_empty() => {
-            return Shape::bool_value(true, method_name.shape_location(source_id));
-        }
-        ShapeCase::Array { .. } | ShapeCase::Object { .. } => {
-            return Shape::bool_value(true, method_name.shape_location(source_id));
+        ShapeCase::Array { .. }
```
dylan-apollo: I think we should list the okay shapes and use _ to make an error so we don't forget any (like float)

apollo-federation/src/connectors/json_selection/methods/public/or.rs
```diff
@@ -91,14 +127,17 @@ fn or_shape(
...
                 }
-                ShapeCase::Int(Some(value)) if *value != 0 => {
-                    return Shape::bool_value(true, method_name.shape_location(source_id));
-                }
-                ShapeCase::String(Some(value)) if !value.is_empty() => {
-                    return Shape::bool_value(true, method_name.shape_location(source_id));
-                }
-                ShapeCase::Array { .. } | ShapeCase::Object { .. } => {
-                    return Shape::bool_value(true, method_name.shape_location(source_id));
+                ShapeCase::Array { .. }
```
dylan-apollo: We should also make the check on `@` be the same logic as each arg so they don't ever get out of sync

apollo-federation/src/connectors/json_selection/methods/public/or.rs
andrewmcgivery: Yea, I think you are right. I was trying to preserve the way that this was originally written but I agree that "safelisting" is a better approach here.

andrewmcgivery: Done!

apollo-federation/src/connectors/json_selection/methods/public/or.rs
andrewmcgivery: I'm not sure if I actually addressed this in my latest commit so please let me know if I did not... and if I didn't, I'd appreciate a suggestion on what you're looking for. :D

apollo-federation/src/connectors/json_selection/methods/public/or.rs
```diff
@@ -39,7 +55,23 @@ fn or_method(
...
+                                "Method ->{} can only accept boolean arguments.",
+                                method_name.as_ref()
+                            ),
+                            input_path.to_vec(),
+                            method_name.range(),
+                        )]);
+                        false
+                    }
+                })
+                .unwrap_or(false);
```
dylan-apollo: Maybe we should make it standard practice to start adding `#![warn(clippy::pedantic)]` to new modules since I think I mostly just repeat what it says.

1. `unwrap_or(false)` after a `map` usually means you wanted `.is_some_and()` instead
2. Using `match` on a single pattern is also a smell, indicates you probably are over-nesting _or_ should use `if let`

In this case, I think the solution is to do this:

```suggestion
            match value_opt {
                Some(JSON::Bool(value)) => result = value,
                Some(_) => {
                    errors.extend(vec![ApplyToError::new(
                        format!(
                            "Method ->{} can only accept boolean arguments.",
                            method_name.as_ref()
                        ),
                        input_path.to_vec(),
                        method_name.range(),
                    )]);
                }
                None => {}
            }
        }
```

The code read's a little easier to me, at least. If we got a bool value, update it. If we got any other value, that's an error. If we got `None`, move on.

Updating to `false` never makes sense because we only `break` on true, so the value was already false.

Personally I would probably `return` as soon as I saw true rather than waiting for the next loop or to fall out the bottom 🤷

apollo-federation/src/connectors/json_selection/methods/public/or.rs
```diff
@@ -67,17 +100,19 @@ fn or_shape(
...
-        ShapeCase::Array { .. } | ShapeCase::Object { .. } => {
-            return Shape::bool_value(true, method_name.shape_location(source_id));
+        _ => {
+            return Shape::error(
+                format!(
+                    "Method ->{} can only be applied to boolean values.",
+                    method_name.as_ref()
+                ),
+                method_name.shape_location(source_id),
+            );
```
dylan-apollo: Interestingly with this and many other of these methods, nullable values (incredibly common) are all going to fail composition now (once type checking is turned on) unless you add `?` to them 🤔. Are we good with that?

Seems fine to me, but we should probably make an explicit error message for `null` that indicates the user should put `?` to ignore `null` values (once that syntax exists).

apollo-federation/src/connectors/json_selection/methods/public/or.rs
```diff
@@ -91,33 +126,25 @@ fn or_shape(
                 ShapeCase::Bool(Some(true)) => {
                     return Shape::bool_value(true, method_name.shape_location(source_id));
                 }
-                ShapeCase::Int(Some(value)) if *value != 0 => {
-                    return Shape::bool_value(true, method_name.shape_location(source_id));
-                }
-                ShapeCase::String(Some(value)) if !value.is_empty() => {
-                    return Shape::bool_value(true, method_name.shape_location(source_id));
+                ShapeCase::Unknown | ShapeCase::Bool(Some(false)) => {
```
dylan-apollo: ah fudge... I also just realized we need to handle `ShapeCase::Name` and `ShapeCase::None` in all of these method shapes 😠.

I think we need to spell out every single case rather than `_`, unfortunately, so we can be sure we're not missing any others.

apollo-federation/src/connectors/json_selection/methods/public/or.rs
andrewmcgivery: Feels like something we should maybe look at addressing as part of the `?` epic perhaps?

apollo-federation/src/connectors/json_selection/methods/public/or.rs
andrewmcgivery: After chatting with @benjamn, ended up redoing the shape method here to use `accepts` instead of `ShapeCase` which will give us more holistic (and simpler!) coverage. :)

We also discussed that we could always add in the "figure out if it is true if we can" later on as a non-breaking change whereas adding it now and removing it later would be breaking. So we're going to go with "add it later if we think it adds value" and keep the validation simple for now. :)

apollo-federation/src/connectors/json_selection/methods/public/or.rs
```diff
@@ -0,0 +1,203 @@
...
+        for arg in args {
+            let arg_shape = arg.compute_output_shape(
+                input_shape.clone(),
+                dollar_shape.clone(),
+                named_var_shapes,
+                source_id,
+            );
+
+            // We will accept anything bool-like OR unknown/named
+            if !(Shape::bool([]).accepts(&arg_shape) || arg_shape.accepts(&Shape::unknown([]))) {
```
dylan-apollo, line 123: I think the question about `None` or `One<Bool, None>` is gonna be the same as for `->gt`

apollo-federation/src/connectors/json_selection/methods/public/or.rs
```diff
@@ -0,0 +1,317 @@
...
+use crate::connectors::json_selection::ApplyToInternal;
+use crate::connectors::json_selection::MethodArgs;
+use crate::connectors::json_selection::VarsWithPathsMap;
+use crate::connectors::json_selection::immutable::InputPath;
+use crate::connectors::json_selection::location::Ranged;
+use crate::connectors::json_selection::location::WithRange;
+use crate::impl_arrow_method;
+
+impl_arrow_method!(OrMethod, or_method, or_shape);
+/// Given 2 or more values to compare, returns true if any of the values are truthy or false if none of them are truthy.
```
dylan-apollo:
```suggestion
/// Given 2 or more values to compare, returns true if any of the values are true
```

apollo-federation/src/connectors/json_selection/methods/public/or.rs
```diff
@@ -0,0 +1,317 @@
...
+                    format!(
+                        "Method ->{} can only be applied to boolean values.",
+                        method_name.as_ref()
+                    ),
+                    input_path.to_vec(),
+                    method_name.range(),
+                )],
+            );
+        }
+    };
```
dylan-apollo:
```suggestion
   let Some(mut result) = data.as_bool() else {
        return (
            None,
            vec![ApplyToError::new(
                format!(
                    "Method ->{} can only be applied to boolean values.",
                    method_name.as_ref()
                ),
                input_path.to_vec(),
                method_name.range(),
            )],
        );
    };
```

apollo-federation/src/connectors/json_selection/methods/public/or.rs
```diff
@@ -0,0 +1,317 @@
...
+    if method_args.and_then(|args| args.args.first()).is_none() {
+        return (
+            None,
+            vec![ApplyToError::new(
+                format!("Method ->{} requires arguments", method_name.as_ref()),
+                input_path.to_vec(),
+                method_name.range(),
+            )],
+        );
+    };
```
dylan-apollo:
```suggestion
   let Some(MethodArgs { args, .. }) = method_args else {
        return (
            None,
            vec![ApplyToError::new(
                format!("Method ->{} requires arguments", method_name.as_ref()),
                input_path.to_vec(),
                method_name.range(),
            )],
        );
    };
```

Lets us unnest the loop later. Doesn't check that there's at least one argument, but do we really care at runtime? We'll start statically checking this in this release anyway, so we could save some work/code 🤷

dylan-apollo (approved):
Looks good, only stylistic suggestions

PR #7721, opened 2025-06-17
move request input and variable merging into apollo-federation

lennyburdette:
passing context to a-f code is handled by a trait

---

**Checklist**

- Tests added and passing[^4]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-federation/src/connectors/runtime/inputs.rs
+++ b/apollo-federation/src/connectors/runtime/inputs.rs
@@ -3,7 +3,6 @@ use std::sync::Arc;
 
 use apollo_compiler::collections::HashSet;
 use apollo_compiler::collections::IndexMap;
-use apollo_federation::connectors::Namespace;
 use http::HeaderMap;
 use http::HeaderValue;
 use http::response::Parts;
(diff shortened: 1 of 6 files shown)


apollo-federation/src/connectors/runtime/inputs.rs
```diff
@@ -13,23 +12,75 @@ use serde_json_bytes::Map;
...
-    pub(super) response: Option<Value>,
-    pub(super) env: Option<Value>,
+use crate::connectors::Namespace;
+
+pub trait ContextReader {
+    fn to_json_map(&self) -> Map<ByteString, Value>;
+}
+
+/// Convert a HeaderMap into a HashMap
+pub(crate) fn externalize_header_map(
```
dylan-apollo, line 22: Annoying to have to duplicate, but we might come up with something more optimized for our use-case anyway 🤷

apollo-federation/src/connectors/runtime/inputs.rs
```diff
@@ -82,21 +133,17 @@ impl MappingContextMerger<'_> {
         map
     }
 
-    pub(crate) fn context(mut self, context: &Context) -> Self {
+    pub fn context<'a>(mut self, context: impl ContextReader + 'a) -> Self {
```
dylan-apollo, line 136: We're probably going to change this all again for click-to-run to calculate inputs and then set them all—but I don't know what that's going to look like yet.

Thinking, though, there's no need to clone _every_ context value, for example, we can _just_ pull in the values asked for by the Connector, and that list can be calculated once per connector.

lennyburdette: I looked into not copying everything, but because it could be an arbitrarily nested value it’s more work than I was willing to do at this stage. But now that I think about it, even only copying the root keys would be a massive improvement.  I’ll update the plan to do that. 

I don’t think you’d use requestmerger for click to run. You just need a map of variables to values. The merger is just a way to construct that map.

PR #7722, opened 2025-06-17
feat: add lte method for connectors mapping language

andrewmcgivery:
Add lte method for connectors mapping language)

---

**Checklist**

- Tests added and passing[^4]
    - [x] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-federation/src/connectors/json_selection/methods/public/lte.rs
```diff
@@ -0,0 +1,372 @@
...
+    };
+
+    let (value_opt, arg_errors) = first_arg.apply_to_path(data, vars, input_path);
+    let mut apply_to_errors = arg_errors;
+    // We have to do this because Value doesn't implement PartialOrd
+    let matches = value_opt.is_some_and(|value| {
+        match (data, &value) {
+            // Number comparisons
+            (JSON::Number(left), JSON::Number(right)) => {
+                left.as_f64().unwrap_or(0.0) <= right.as_f64().unwrap_or(0.0)
```
benjamn: I don't think substituting `0.0` here is a good idea, especially since it returns `true` if both unwrappings fail. Let's return `false` with an error if either conversion fails?

apollo-federation/src/connectors/json_selection/methods/public/lte.rs
```diff
@@ -0,0 +1,372 @@
...
+        | (ShapeCase::Int(_), ShapeCase::Int(_))
+        | (ShapeCase::Float, ShapeCase::Float)
+        | (ShapeCase::Int(_), ShapeCase::Float)
+        | (ShapeCase::Float, ShapeCase::Int(_))
+        | (ShapeCase::Bool(_), ShapeCase::Bool(_))
+        | (ShapeCase::Null, ShapeCase::Null)
+        | (ShapeCase::String(_), ShapeCase::String(_)) => {
+            Shape::bool(method_name.shape_location(source_id))
+        }
+        _ => Shape::error(
```
benjamn: This makes it an error if the input shape is `ShapeCase::Name`, which could become/represent a number at runtime. Instead of lumping that case into the `_ => Shape::error` branch, let's make it explicit.

apollo-federation/src/connectors/json_selection/methods/public/lte.rs
```diff
@@ -0,0 +1,372 @@
...
+
+    let arg_shape = first_arg.compute_output_shape(
+        input_shape.clone(),
+        dollar_shape,
+        named_var_shapes,
+        source_id,
+    );
+
+    match (arg_shape.case(), input_shape.case()) {
+        (ShapeCase::Unknown, ShapeCase::Unknown)
```
benjamn: I don't think we should require both shapes to be `Unknown` here. As long as one of the shapes is among the allowed cases below, the other shape can be `Unknown`. Of course both `Unknown` is also allowed, just not required.

benjamn (changes requested):
Looking good with a few last details to tie up!

apollo-federation/src/connectors/json_selection/methods/public/lte.rs
benjamn: Possibly similar logic: https://github.com/apollographql/router/pull/7717#discussion_r2159194424

apollo-federation/src/connectors/json_selection/methods/public/gte.rs
```diff
@@ -28,69 +30,115 @@ fn gte_method(
...
+                        return None;
+                    }
+                };
+
+                Some(JSON::Bool(left >= right))
             }
-        });
+            // String comparisons
+            (JSON::String(left), JSON::String(right)) => Some(JSON::Bool(left >= right)),
+            // Mixed types or incomparable types (including arrays and objects) return false
```
dylan-apollo, line 73: This does change the semantics of the original, because we're not always returning true if `left->eq(right)`. I think that's okay, though, personally, because we could always lift this restriction in the future if we really needed to.

Effectively, `->gte` is now `$->or(left->gt(right), left->eq(right))` instead of `$->or(left->eq(right), left->gt(right))`. The latter short-circuits today and would not produce None/error, but `true` the second it saw it.

However I _also_ think our decision about universal `None` handling was such that eventually even `->or` will error if _any_ condition is an error (rather than short circuiting), so then this behavior matches either pattern.

Just want to ping @benjamn to make sure that's right, since he's going to be implementing the consistent `None` handling stuff I think.

dylan-apollo (commented):
LGTM after we double-check my comment is okay

apollo-federation/src/connectors/json_selection/methods/public/gte.rs
```diff
@@ -28,69 +30,115 @@ fn gte_method(
...
+            _ => {
+                apply_to_errors.push(ApplyToError::new(
+                    format!(
+                        "Method ->{} can only compare numbers and strings. Found: {data} >= {value}",
```
benjamn, line 77: There's a `json_type_name` helper in `helpers.rs` that might be helpful for limiting the length of this error message. Not blocking, just spreading awareness.

apollo-federation/src/connectors/json_selection/methods/public/gte.rs
benjamn: > However I also think our decision about universal None handling was such that eventually even ->or will error if any condition is an error (rather than short circuiting), so then this behavior matches either pattern.

Yes, I agree that `or_shape` should compute the shape of every argument (without short-circuiting), to prevent runtime surprises. If we do that, we should be fine here, as you said.

apollo-federation/src/connectors/json_selection/methods/public/gte.rs
```diff
@@ -204,124 +252,217 @@ mod tests {
...
-        );
+        assert_eq!(result.0, Some(json!({})),);
         assert!(!result.1.is_empty());
-        assert!(
-            result.1[0]
-                .message()
-                .contains("Method ->gte can directly compare numbers, strings, booleans, and null. Either a mix of these was provided or something else such as an array or object. Found: [1,2,3] >= [1,2]")
-        );
+        assert!(result.1[0].message().contains(
+            "Method ->gte can only compare numbers and strings. Found: [1,2,3] >= [1,2]"
```
benjamn, line 302: Just reiterating that I do eventually want inequality comparisons to work for arrays (lexicographically, somewhat like strings), but that can wait for a later PR.

apollo-federation/src/connectors/json_selection/methods/public/lte.rs
```diff
@@ -0,0 +1,468 @@
...
+
+    if is_comparable_shape_combination(&arg_shape, &input_shape) {
+        Shape::bool(method_name.shape_location(source_id))
+    } else {
+        Shape::error_with_partial(
+            format!(
+                "Method ->{} can only compare two numbers or two strings. Found {input_shape} <= {arg_shape}",
+                method_name.as_ref()
+            ),
+            Shape::bool(method_name.shape_location(source_id)),
```
benjamn:
```suggestion
        Shape::error_with_partial(
            format!(
                "Method ->{} can only compare two numbers or two strings. Found {input_shape} <= {arg_shape}",
                method_name.as_ref()
            ),
            Shape::bool_value(false, method_name.shape_location(source_id)),
```
Returning `false` instead of just `Bool` seems accurate and would be consistent with JavaScript (to pick a programming language where inequality operators are relatively tolerant). We're doing "better" than JS by warning, but `false` also makes sense as the `partial` shape, I think.

apollo-federation/src/connectors/json_selection/methods/public/gte.rs
andrewmcgivery: I tried that function and IMO the message is actually more user friendly without it.

```
# without that function
[1,2,3] >= [1,2]

# with it
array >= array
```

So I think in this case, I'm going to go without it... but really helpful to know that function exists!

apollo-federation/src/connectors/json_selection/methods/public/lte.rs
andrewmcgivery: Done!

PR #7724, opened 2025-06-18
Add support of cache tags in extensions

bnjjj:


apollo-router/src/plugins/response_cache/plugin.rs
bnjjj, line 1: I can't see where you remove the support of `Surrogate-Key` header. I think we should remove this and only keeps the extensions part. cc @SimonSapin

PR #7728, opened 2025-06-18
Revert "coprocessor: fix parsing of graphql responses with null as `data` (backport #7141)"

abernix:
Reverts apollographql/router#7315


abernix:
This was a defensive attempt to get out ahead of a git-bisect, but we found the actual root problem elsewhere, so I can close this.

PR #7731, opened 2025-06-19
fix(coprocessor): improve handling of invalid GraphQL responses with conditional validation

BrynCooke:
## Task

Fix handling of invalid GraphQL responses returned from coprocessors, particularly when used with subscriptions. Added conditional response validation and improved testing to ensure correctness.

## Why this change is needed

The router was creating invalid GraphQL responses internally, especially when subscriptions terminate. When a coprocessor is configured, it validates all responses for correctness, causing errors to be logged when the router generates invalid internal responses. This affects the reliability of subscription workflows with coprocessors.

### Coprocessor response validation improvements
- Added conditional response validation that only validates coprocessor responses when the incoming payload was valid
- Added `response_validation` configuration option (defaults to true) for backward compatibility
- Improved handling of invalid responses to prevent validation failures on already-invalid data

### Subscription integration tests
- **WebSocket passthrough tests**: Mock server using axum with GraphQL-WS protocol, tests connection lifecycle and streaming behavior
- **HTTP callback tests**: Mock callback and subgraph servers, tests subscription lifecycle through HTTP callbacks with error scenarios
- Tests cover normal operation, error payloads, pure error responses, and coprocessor integration
- Added consistent event verification and error logging detection

### Documentation updates
- Added comprehensive documentation for the `response_validation` configuration option
- Documented use cases, behavior differences, and migration guidance

## Technical details

- Conditional validation prevents failures on responses that were already invalid before coprocessor processing
- `ENABLE_CONDITIONAL_VALIDATION` const controls the new behavior (set to true)
- `was_incoming_payload_valid()` function checks if original response was minimally valid
- Added comprehensive test coverage for all coprocessor stages with validation enabled/disabled

## Suggested review strategy

- Examine conditional validation logic in `handle_graphql_response()` function in coprocessor/mod.rs
- Review `was_incoming_payload_valid()` implementation for payload validity checking
- Check callback server HTTP protocol implementation in subscription test files
- Validate test coverage for coprocessor validation scenarios across all pipeline stages
- Review documentation accuracy for the new `response_validation` configuration option

<\!-- start metadata -->
EOF < /dev/null


apollo-router/tests/integration/subscriptions/ws_passthrough.rs
```diff
@@ -0,0 +1,483 @@
...
+            "extensions": {
+                "code": "INTERNAL_ERROR"
+            }
+        }]
+    })
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn test_subscription_ws_passthrough() -> Result<(), BoxError> {
+    if std::env::var("TEST_APOLLO_KEY").is_ok() && std::env::var("TEST_APOLLO_GRAPH_REF").is_ok() {
```
bnjjj: We already have a function for this

apollo-router/src/graphql/response.rs
```diff
@@ -78,6 +80,11 @@ impl Response {
         incremental: Vec<IncrementalResponse>,
         created_at: Option<Instant>,
     ) -> Self {
+        // It's not possible to have a graphql response that is not present and has empty errors.
+        // In this situation we coerce into a valid response by setting a data of null.
+        if data.is_none() && errors.is_empty() {
```
bnjjj: Should we also add this in graphql::Response::default() ?

apollo-router/src/plugins/coprocessor/mod.rs
```diff
@@ -716,18 +734,36 @@ where
...
+                if response_validation {
+                    graphql::Response::from_value(body_as_value).unwrap_or_else(|error| {
+                        graphql::Response::builder()
+                            .errors(vec![
+                                Error::builder()
+                                    .message(format!(
+                                        "couldn't deserialize coprocessor output body: {error}"
+                                    ))
+                                    .extension_code(COPROCESSOR_DESERIALIZATION_ERROR_EXTENSION)
+                                    .build(),
```
carodewig: Nit: this error build call is duplicated a few times - might be worth creating a function for it?

apollo-router/tests/integration/subscriptions/ws_passthrough.rs
BrynCooke: The new tests are more extensive.

apollo-router/src/graphql/response.rs
BrynCooke: This was reverted as we discussed a way forward as a group. We realised that it was too risky a change and that there is conflicting guidance in grapqhl subscriptions/streaming/defer specs.

apollo-router/src/plugins/coprocessor/mod.rs
BrynCooke: Good point: [9334d72](https://github.com/apollographql/router/pull/7731/commits/9334d7237527c3191f4954b521c05ab30a77f220)

apollo-router/tests/integration/subscriptions/ws_passthrough.rs
bnjjj: I was talking about this
```
if !graph_os_enabled() {
        return;
    }
```

BrynCooke: Ahh, OK. I'll fix .

apollo-router/tests/common.rs
```diff
@@ -171,6 +208,85 @@ impl IntegrationTest {
             .expect("lock poisoned")
             .expect("no bind address set, router must be started first.")
     }
+
+    /// Reserve a port for use in the test and return it
+    /// If port is provided, use that port; otherwise allocate a new one
```
carodewig:
```suggestion
```

There's no option to provide a port in this function

apollo-router/tests/common.rs
```diff
@@ -944,6 +1063,103 @@ impl IntegrationTest {
...
+                    && !line.contains("error_handling")
+                    && !line.contains("no_error")
+                    && !line.contains("ERROR:")
+                // Actual error logs usually have ERROR: prefix
+                {
+                    continue;
+                }
+                if line.contains("ERROR:") {
+                    error_logs.push(line.clone());
+                }
```
carodewig: Aren't these checks redundant?
```suggestion
                // Actual error logs usually have ERROR: prefix
                if line.contains("ERROR:") {
                    error_logs.push(line.clone());
                }
```

apollo-router/tests/common.rs
```diff
@@ -944,6 +1063,103 @@ impl IntegrationTest {
...
+            panic!(
+                "Found {} unexpected error(s) in router logs (excluding {} exceptions):\n\n{}\n\nFull log dump:\n\n{}",
+                error_logs.len(),
+                exceptions.len(),
+                error_logs.join("\n"),
+                self.logs.join("\n")
+            );
+        }
+    }
+
```
carodewig: Nit: this is more a style suggestion, but what would you think about breaking these functions up a bit?
```suggestion
    #[allow(dead_code)]
    pub fn error_logs(&mut self) -> Vec<String> {
        // Read any remaining logs from buffer
        self.read_logs();
        
        const JSON_ERROR_INDICATORS: [&str; 3] = ["\"level\":\"ERROR\"", "panic", "PANIC"];
        
        let mut error_logs = Vec::new();
        for line in &self.logs {
            if JSON_ERROR_INDICATORS.iter().any(|err| line.contains(err)) {
                error_logs.push(line.clone());
            } else if line.contains("ERROR") && !line.contains("level") {
                error_logs.push(line.clone());
            }
        }
        error_logs
    }

    #[allow(dead_code)]
    pub fn assert_no_error_logs(&mut self) {
        let error_logs = self.error_logs();
        if !error_logs.is_empty() {
            panic!(
                "Found {} unexpected error(s) in router logs:\n\n{}\n\nFull log dump:\n\n{}",
                error_logs.len(),
                error_logs.join("\n"),
                self.logs.join("\n")
            );
        }
    }

    #[allow(dead_code)]
    pub fn assert_no_error_logs_with_exceptions(&mut self, exceptions: &[&str]) {
        let mut error_logs = self.error_logs();
        
        // remove any logs that contain our exceptions
        error_logs.retain(|line| !exceptions.iter().any(|exception| line.contains(exception)));

        if !error_logs.is_empty() {
            panic!(
                "Found {} unexpected error(s) in router logs (excluding {} exceptions):\n\n{}\n\nFull log dump:\n\n{}",
                error_logs.len(),
                exceptions.len(),
                error_logs.join("\n"),
                self.logs.join("\n")
            );
        }
    }
```

apollo-router/src/plugins/coprocessor/mod.rs
```diff
@@ -326,6 +330,9 @@ struct Conf {
     #[schemars(with = "String", default = "default_timeout")]
     #[serde(default = "default_timeout")]
     timeout: Duration,
+    /// Response validation defaults to true
+    #[serde(default = "default_response_validation")]
+    response_validation: bool,
```
carodewig, line 335: Is there any benefit to having the response validation be conditional per stage? Or having validation on the request pipeline side but not the response?

For example, I noticed a comment that router responses will never undergo response validation - that's something that would be more obvious if the `response_validation` config option was not available within `conf.router`.

apollo-router/src/plugins/coprocessor/execution.rs
```diff
@@ -388,11 +383,20 @@ where
 
     validate_coprocessor_output(&co_processor_output, PipelineStep::ExecutionResponse)?;
 
+    // Check if the incoming GraphQL response was valid according to GraphQL spec
+    let incoming_payload_was_valid =
+        crate::plugins::coprocessor::was_incoming_payload_valid(&first, response_config.body);
+
```
carodewig, line 389: Question: it looks like this incoming payload validity check is only present on response stages. Is it assumed that the payload will always be valid on request stages, or is there some other reason it isn't checked there?

Apologies if this is documented somewhere already and I missed it.

apollo-router/src/plugins/coprocessor/execution.rs
```diff
@@ -1008,4 +1027,411 @@ mod tests {
             json!({ "data": { "test": 3, "has_next": false }, "hasNext": false }),
         );
     }
+
+    // Helper function to create execution stage for validation tests
+    fn create_execution_stage_for_validation_test() -> ExecutionStage {
```
carodewig: Nit: perhaps rename to be more clear between this and `create_execution_stage_for_request_validation_test `?
```suggestion
    fn create_execution_stage_for_response_validation_test() -> ExecutionStage {
```

apollo-router/src/plugins/coprocessor/supergraph.rs
```diff
@@ -1224,4 +1245,412 @@ mod tests {
             json!({ "data": { "test": 3 }, "hasNext": false }),
         );
     }
+
+    // Helper function to create supergraph stage for validation tests
+    fn create_supergraph_stage_for_validation_test() -> SupergraphStage {
```
carodewig:
```suggestion
    fn create_supergraph_stage_for_response_validation_test() -> SupergraphStage {
```

abernix:
@mergifyio backport dev

apollo-router/src/plugins/coprocessor/execution.rs
bnjjj: We could this add a check for request later but right now it's not something we want to validate, especially in 1.x

apollo-router/src/plugins/coprocessor/mod.rs
bnjjj: This configuration is basically to be able to rollback to the previous behavior before we had this validation in place. This option will only exist in 1.x and 2.x to make sure we won't break anything and we will implement it differently in 3.x probably.

apollo-router/src/plugins/rhai/tests.rs
```diff
@@ -237,67 +237,39 @@ fn new_rhai_test_engine() -> Engine {
...
-    ));
-    assert!(tracing_test::internal::logs_with_scope_contain(
-        "apollo_router",
-        "error log"
-    ));
+    use tracing::subscriber;
+
+    use crate::assert_snapshot_subscriber;
+
+    subscriber::with_default(assert_snapshot_subscriber!(), || {
```
goto-bus-stop: This test changed. Previously it was asserting that _all_ message levels were logged, but the new snapshot only asserts that info/warn/error logs are present.

docs/source/routing/customization/coprocessor.mdx
```diff
@@ -152,6 +152,38 @@ coprocessor:
...
+```yaml title="router.yaml"
+coprocessor:
+  url: http://127.0.0.1:8081
+  response_validation: true # Default: true. Set to false to disable GraphQL response validation.
+```
+
+You can disable response validation by setting `response_validation: false`. This is useful in scenarios where:
+
+- Your coprocessor intentionally modifies responses in ways that may not conform to strict GraphQL validation rules
+- You want to maximize performance by skipping validation overhead
```
goto-bus-stop: I do not think we should include this point.

docs/source/routing/customization/coprocessor.mdx
```diff
@@ -152,6 +152,38 @@ coprocessor:
...
+<Caution>
+
+Disabling response validation can result in invalid GraphQL responses being sent to clients, which may cause client-side errors or unexpected behavior. Only disable validation if you're certain your coprocessor produces valid responses.
+
+</Caution>
+
+When response validation is enabled (the default), the router:
+- Validates that coprocessor-modified responses conform to the GraphQL specification
+- Only applies validation when the original response to the coprocessor was valid (conditional validation)
+- Returns appropriate errors to clients if validation fails
```
goto-bus-stop: Should clarify if conditional validation also falls back to the original response, like when validation is disabled.

goto-bus-stop (changes requested):
A few things that jumped out, I haven't reviewed the code in depth yet

goto-bus-stop:
I'm not really happy with the docs yet, and as they would be available immediately after landing this PR, I've just deleted them. I'll follow up with a PR that re-introduces the docs changes and apply some edits at that point.

apollo-router/tests/integration/subscriptions/callback.rs
```diff
@@ -0,0 +1,517 @@
...
+            }]
+        }),
+    ];
+    verify_callback_events(&callback_state, expected_user_events).await?;
+
+    // Check for errors in router logs
+    router.assert_no_error_logs();
+
+    tracing::info!("✅ Callback mode subscription test completed successfully");
+
```
goto-bus-stop: already reported by `cargo test`
```suggestion
```

apollo-router/tests/integration/subscriptions/callback.rs
```diff
@@ -0,0 +1,517 @@
...
+    assert_eq!(
+        actual_user_events, expected_user_events,
+        "Callback user events do not match expected events"
+    );
+
+    tracing::info!(
+        "✅ Successfully verified {} callback events",
+        callbacks.len()
+    );
+
```
goto-bus-stop:
```suggestion
```

apollo-router/tests/integration/subscriptions/callback.rs
```diff
@@ -0,0 +1,517 @@
...
+    let response = client
+        .post(&callback_url)
+        .json(&complete_payload)
+        .send()
+        .await?;
+
+    assert_eq!(response.status(), 202, "Valid completion should return 202");
+
+    tracing::info!("✅ All callback error scenarios tested successfully");
+
```
goto-bus-stop:
```suggestion
```

apollo-router/tests/integration/subscriptions/callback.rs
```diff
@@ -0,0 +1,517 @@
...
+            // Missing reviews field to test error handling
+        }),
+    ];
+    verify_callback_events(&callback_state, expected_user_events).await?;
+
+    // Check for errors in router logs
+    router.assert_no_error_logs();
+
+    tracing::info!("✅ Callback mode subscription test with error payload completed successfully");
+
```
goto-bus-stop:
```suggestion
```

apollo-router/tests/integration/subscriptions/callback.rs
```diff
@@ -0,0 +1,517 @@
...
+    ];
+    verify_callback_events(&callback_state, expected_user_events).await?;
+
+    // Check for errors in router logs
+    router.assert_no_error_logs();
+
+    tracing::info!(
+        "✅ Callback mode subscription test with pure error payload completed successfully"
+    );
+
```
goto-bus-stop:
```suggestion
```

apollo-router/tests/integration/subscriptions/ws_passthrough.rs
```diff
@@ -0,0 +1,497 @@
...
+        .map_err(|e| format!("Event verification failed: {}", e))?;
+
+    // Check for errors in router logs
+    router.assert_no_error_logs();
+
+    info!(
+        "✅ Passthrough subscription mode test completed successfully with {} events",
+        custom_payloads.len()
+    );
+
```
goto-bus-stop:
```suggestion
```

apollo-router/tests/integration/subscriptions/ws_passthrough.rs
```diff
@@ -0,0 +1,497 @@
...
+    // Check for errors in router logs (allow expected coprocessor error)
+    router.assert_no_error_logs();
+
+    info!(
+        "✅ Passthrough subscription mode with coprocessor test completed successfully with {} events",
+        custom_payloads.len()
+    );
+    info!("✅ Coprocessor successfully processed subscription requests and responses");
+
```
goto-bus-stop:
```suggestion
```

apollo-router/tests/integration/subscriptions/ws_passthrough.rs
```diff
@@ -0,0 +1,497 @@
...
+        .map_err(|e| format!("Event verification failed: {}", e))?;
+
+    // Check for errors in router logs
+    router.assert_no_error_logs();
+
+    info!(
+        "✅ WebSocket passthrough with error payload test completed successfully with {} events",
+        custom_payloads.len()
+    );
+
```
goto-bus-stop:
```suggestion
```

apollo-router/tests/integration/subscriptions/ws_passthrough.rs
```diff
@@ -0,0 +1,497 @@
...
+        .map_err(|e| format!("Event verification failed: {}", e))?;
+
+    // Check for errors in router logs
+    router.assert_no_error_logs();
+
+    info!(
+        "✅ WebSocket passthrough with pure error payload test completed successfully with {} events",
+        custom_payloads.len()
+    );
+
```
goto-bus-stop:
```suggestion
```

apollo-router/tests/integration/subscriptions/ws_passthrough.rs
```diff
@@ -0,0 +1,497 @@
...
+        .await
+        .map_err(|e| format!("Event verification failed: {}", e))?;
+
+    // Check for errors in router logs
+    router.assert_no_error_logs();
+
+    info!(
+        "✅ WebSocket passthrough with pure error payload and coprocessor test completed successfully"
+    );
+
```
goto-bus-stop:
```suggestion
```

apollo-router/tests/integration/subscriptions/ws_passthrough.rs
```diff
@@ -0,0 +1,472 @@
...
+/// Creates an empty payload (sent to mock server)
+fn create_empty_data_payload() -> serde_json::Value {
+    serde_json::json!({
+        // No data attribute at all
+    })
+}
+
+/// Creates an expected error response payload (when coprocessor detects issues)
+#[allow(dead_code)]
```
goto-bus-stop: this seems like it should be tested. maybe this is what the "pure error" tests below are supposed to use? And then we need another set of tests for the "empty payload" case?

apollo-router/tests/integration/subscriptions/ws_passthrough.rs
```diff
@@ -0,0 +1,472 @@
...
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn test_subscription_ws_passthrough_pure_error_payload() -> Result<(), BoxError> {
+    if !graph_os_enabled() {
+        eprintln!("test skipped");
+        return Ok(());
+    }
+    // Create custom payloads: one normal event, one pure error event (no data, only errors)
+    let custom_payloads = vec![create_user_data_payload(1), create_empty_data_payload()];
```
goto-bus-stop: The comment is out of sync with the code. The "pure error" tests in this file are using an empty payload, without errors

goto-bus-stop (dismissed):
I think this is the last thing that jumps out at me. Let's resolve it tomorrow and then I think we can merge.

abernix:
@mergifyio backport dev

abernix:
https://github.com/Mergifyio backport dev

abernix:
@mergifyio backport dev

abernix:
@mergifyio backport dev

abernix:
@mergifyio backport dev

PR #7735, opened 2025-06-20
jemalloc metrics

Velfi:
---

**Checklist**

- Tests added and passing[^4]
    - [ ] Unit tests
    - [ ] Integration tests

**Exceptions**

*Note any exceptions here*

**Notes**


.changesets/feat_zelda_jemalloc_metrics.md
```diff
@@ -0,0 +1,12 @@
+### jemalloc metrics ([PR #7735](https://github.com/apollographql/router/pull/7735))
+
+This PR adds the following new metrics when running Linux with the `global-allocator` feature:
```
garypen:
```suggestion
This PR adds the following new metrics when running the router on Linux with its default `global-allocator` feature:
```

apollo-router/src/axum_factory/axum_http_server_factory.rs
```diff
@@ -481,11 +499,23 @@ where
...
+    let jemalloc_metrics_instrument = jemalloc_metrics_instruments();
     router = router.layer(layer_fn(move |service| {
-        let _ = &instrument;
+        let _ = &session_count_instrument;
+        #[cfg(all(
+            feature = "global-allocator",
+            not(feature = "dhat-heap"),
+            target_os = "linux"
+        ))]
+        let _ = &jemalloc_metrics_instrument;
```
garypen: Couldn't both these things just be created inside the `layer_fn`? (I know the existing `session_count_instrument()` isn't your choice, but it was less messy due to the lack of `cfg` statements). Now we have them, it's a bit messy and you could clean things up with:
```
    // Tie the lifetime of the various instruments to the lifetime of the router
    // by creating them in a no-op layer.
    router = router.layer(layer_fn(|service| {
        let _session_count_instrument = session_count_instrument();
        #[cfg(all(
            feature = "global-allocator",
            not(feature = "dhat-heap"),
            target_os = "linux"
        ))]
        let _je_malloc_instrument = jemalloc_metrics_instruments();
        service
    }));
```

apollo-router/src/axum_factory/metrics.rs
```diff
@@ -0,0 +1,137 @@
...
+
+    pub(crate) fn create_active_gauge() -> ObservableGauge<u64> {
+        meter_provider()
+            .meter("apollo/router")
+            .u64_observable_gauge("apollo.router.jemalloc.active")
+            .with_description("Total active bytes in jemalloc")
+            .with_unit("bytes")
+            .with_callback(|gauge| {
+                if tikv_jemalloc_ctl::epoch::advance().is_err() {
+                    tracing::error!("failed to read jemalloc active stats");
```
garypen: This is not a strong opinion, but I wouldn't consider missing a stats read an error, but a warn.
```suggestion
                    tracing::warn!("failed to read jemalloc active stats");
```

apollo-router/src/axum_factory/metrics.rs
garypen, line 1: I think that advancing the epoch each time these are read may lead to too many advances and we'd lose the benefits of `jemalloc` stats caching. In an ideal (or more ideal) world, we'd advance the epoch once for all of these stats when we tried to read any of them to make best use of the `jemalloc` cache.

Spawn a task in the router layer (where you create the instruments) and have that task advance the epoch every "clock tick" (say 500ms).

The idea being to avoid excessive re-freshing of `jemalloc` stats which is probably expensive since it's a global stats refresh and to also ensure that, in general, stats were read from the same epoch (Note, this second point isn't actually enforced by my proposal, since a read could occur over an epoch boundary, but in general it's close enough).

We could also have a common function for creating gauges to avoid repetition. Something like:

```
    fn create_gauge(metric: &str) -> ObservableGauge<u64> {
       <implement common functionality)
    }

    pub(crate) fn create_active_gauge() -> ObservableGauge<u64> {
        create_gauge("active")
    }

    etc...
```

apollo-router/src/axum_factory/metrics.rs
```diff
@@ -0,0 +1,66 @@
...
+))]
+pub(crate) mod jemalloc {
+    use std::time::Duration;
+
+    use opentelemetry::metrics::MeterProvider;
+    use opentelemetry::metrics::ObservableGauge;
+
+    use crate::metrics::meter_provider;
+
+    fn start_epoch_advance_loop() -> tokio::task::JoinHandle<()> {
```
garypen:
```suggestion
    pub(crate) fn start_epoch_advance_loop() -> tokio::task::JoinHandle<()> {
```

BrynCooke:
@mergify.io backport 1.x

PR #7736, opened 2025-06-20
fix(federation): For backwards compatibility, permit field sets where enum values are mistakenly strings

sachindshinde:
For historical reasons, some supergraph schemas may have field sets where enum values are mistakenly written as strings. This PR updates subgraph extraction to automatically fix this invalidity when discovered, correcting them back to enum values.


apollo-federation/src/schema/field_set.rs
```diff
@@ -92,21 +93,204 @@ pub(crate) fn parse_field_set(
...
+}
+
+fn fix_string_enum_values_in_directives(
+    schema: &Valid<Schema>,
+    directives: &mut executable::DirectiveList,
+) -> bool {
+    let mut is_modified = false;
+    for directive in directives.iter_mut() {
+        let Some(directive_definition) = schema.directive_definitions.get(&directive.name) else {
+            return is_modified;
```
duckki: Shouldn't we `continue` here?

duckki (approved):
I have just a minor suggestion. Otherwise, it looks good!

apollo-federation/src/schema/field_set.rs
sachindshinde: Yep we should use `continue` (it better aligns with how the rest of the code treats GraphQL invalidities), will fix.

PR #7737, opened 2025-06-20
Naomijub/connectors test support

naomijub:
---

**Checklist**

- Tests added and passing[^4]
    - [ ] Unit tests
    - [ ] Integration tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-federation/src/connectors/runtime/errors.rs
```diff
@@ -15,6 +16,20 @@ pub struct RuntimeError {
...
 }
 
+impl Serialize for RuntimeError {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let mut state = serializer.serialize_struct("RuntimeError", 3)?;
+        state.serialize_field("message", &self.message)?;
+        state.serialize_field("path", &self.path)?;
```
lennyburdette: this path isn't correct (the full path is a combination of information only known by the router and a transformation of this value) so we should probably leave it out

apollo-router/src/plugins/connectors/handle_responses.rs
```diff
@@ -50,27 +49,12 @@ impl From<RuntimeError> for graphql::Error {
     fn from(error: RuntimeError) -> Self {
         let path: Path = (&error.path).into();
 
-        let mut err = graphql::Error::builder()
+        let err = graphql::Error::builder()
```
lennyburdette, line 52: 🙌🏻

Cargo.lock
lennyburdette, line 1: should we expect any version bumps when you're only added a feature?

apollo-federation/src/connectors/mod.rs
```diff
@@ -82,10 +83,23 @@ impl ConnectId {
     /// Until we have a source-aware query planner, we'll need to split up connectors into
     /// their own subgraphs when doing planning. Each subgraph will need a name, so we
     /// synthesize one using metadata present on the directive.
-    pub(crate) fn synthetic_name(&self) -> String {
+    pub fn synthetic_name(&self) -> String {
         format!("{}_{}", self.subgraph_name, self.directive.synthetic_name())
     }
 
+    #[cfg(feature = "test_support")]
```
dylan-apollo: Why are we putting things behind a feature flag? Ultimately I think the only consumer of `apollo-federation` that _won't_ want this is router, do we think the new code will have a noticeable impact on compile times or deps?

apollo-federation/src/connectors/spec/mod.rs
```diff
@@ -173,3 +174,33 @@ impl From<ConnectSpec> for Version {
         }
     }
 }
+
+impl TryFrom<Version> for ConnectSpec {
```
dylan-apollo: There's already `TryFrom<&Version>` above, can we use that?

apollo-federation/src/connectors/mod.rs
naomijub: I don't think it will have any noticeable compile times differences, I made it this way to avoid people calling the functions on this feature from the router, as they are more focused on test usage. I'm Ok to remove it

Cargo.lock
naomijub: I don't think we need a new version bump for this feature, probably a simple patch works

apollo-federation/src/connectors/runtime/errors.rs
naomijub: I forgot to push, 10e0a2633

Cargo.lock
naomijub: I did remove it as it was only used by one function with a clear name on it

apollo-federation/src/connectors/mod.rs
naomijub: 10e0a2633

apollo-federation/src/connectors/spec/mod.rs
naomijub: 10e0a2633

Cargo.lock
```diff
@@ -1882,9 +1882,9 @@ dependencies = [
 
 [[package]]
 name = "crossbeam-deque"
-version = "0.8.5"
+version = "0.8.6"
```
dylan-apollo: Should probably revert these lock changes if we don't need them.

apollo-federation/src/connectors/runtime/responses.rs
```diff
@@ -56,7 +56,7 @@ pub fn handle_raw_response(
 }
 
 // --- RAW RESPONSE ------------------------------------------------------------
-
+#[derive(Debug)] // REMOVE
```
dylan-apollo: // REMOVE?

apollo-federation/src/connectors/spec/mod.rs
```diff
@@ -62,7 +62,29 @@ impl ConnectSpec {
         }
     }
 
-    pub(crate) fn identity() -> Identity {
+    const IDENTITY_NAME: Name = name!("connect");
+
+    #[expect(dead_code)]
```
dylan-apollo: Can we delete this? We have so much dead code in this repo that never gets cleaned up 😭

apollo-federation/src/connectors/spec/mod.rs
```diff
@@ -149,6 +171,14 @@ impl From<ConnectSpec> for Version {
     }
 }
 
+#[derive(Debug, thiserror::Error)]
+#[error("connect spec version {major}.{minor}, latest available is: {latest}")]
+pub struct Error {
```
dylan-apollo: Did we end up using this?

apollo-federation/src/connectors/runtime/errors.rs
```diff
@@ -27,6 +27,30 @@ impl RuntimeError {
         }
     }
 
+    pub fn extensions(&self) -> Map<ByteString, Value> {
```
dylan-apollo, line 30: Building and checking these extensions for testing feels like the wrong level of abstraction to me. `RuntimeError` is very much a for-the-router construct.

Could we instead have the functions return more structured errors (like ApplyToError) and let the router do the coercing into GraphQLy things like this? Then we should be able to give better feedback to users long-term (like in-IDE hints)

apollo-federation/src/connectors/spec/mod.rs
naomijub: I was not the one that added it and I added it to check if it would fail somewhere, weirdly it is appearing in my PR. Should only appear the ` #[expect(dead_code)]`

apollo-federation/src/connectors/runtime/errors.rs
naomijub: That was actually @lennyburdette suggestion for returning extensions. However, in response to the `ApplyToError`, Yes, I can definitively try to work with that on the testing framework.

Cargo.lock
naomijub: e1998c49f

apollo-federation/src/connectors/spec/mod.rs
naomijub: e1998c49f

apollo-federation/src/connectors/runtime/responses.rs
naomijub: e1998c49f

apollo-federation/src/connectors/runtime/errors.rs
lennyburdette: there's an interesting challenge we should discuss in a design doc about what output shapes are appropriate for tests. 
- do we use the data/errors/extensions output keys even though they don't quite make sense?
- should we expose `_entities`? root fields?
- should errors look like graphql errors (message/extensions)?
- are batch results sorted appropriately?
- what should apply to errors look like?

i was thinking that top-level errors should look like this
```yaml
response:
  errors: 
    - message: request error
      extensions:
        code: CONNECTOR_FETCH
        service: subgraph_name
        connector:
          coordinate: Query.foo[0]
        http:
          status: 400
```
which means we should do all the extensions creation at this point

naomijub: I was personally expecting that Data that I got would look like what I get in a graphql file. I feel that exposing `_entities` or root fields doesn't make a lot of sense from user's perspective. However, I feel that the errors as in your example feel more granular and testable.

apollo-federation/src/connectors/spec/mod.rs
lennyburdette: this looks like a merge conflict, i believe i removed this in favor of similar functionality further down the file

apollo-federation/src/connectors/mod.rs
lennyburdette, line 1: i'm fine with these changes as temporary public API for the exploration, but let's definitely prioritize design work around connector identification

apollo-federation/src/connectors/spec/mod.rs
naomijub: Updated 3e4e55cb9

dylan-apollo (approved):

PR #7739, opened 2025-06-23
move response key

lennyburdette:
prerequisite for refactoring errors/path for connector errors

---

**Checklist**

- Tests added and passing[^4]
    - [ ] Unit tests
    - [ ] Integration tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/plugins/connectors/make_requests.rs
```diff
@@ -80,8 +28,8 @@ impl ResponseKey {
 /// - parent path: `["posts", @, "user"]
 /// - path from key: `["_entities", 0, "user", "profile"]`
 /// - result: `["posts", 1, "user", "profile"]`
-impl From<&ResponseKey> for Path {
-    fn from(key: &ResponseKey) -> Self {
+impl Path {
+    pub(crate) fn from_response_key(key: &ResponseKey) -> Self {
```
carodewig: NIt: if possible, it would be nice to keep the ergonomics of `impl From` - could that be moved into `apollo-federation/src/connectors/runtime/key.rs` rather than implementing a separate method here?

If not, the doc comment should be moved to just above the `from_response_key` function rather than applying to the whole `impl Path`

lennyburdette: good catch on the doc comment location, thanks!

i can't keep this an `impl From` because `Path` is a apollo-router type and I'm moving `ResponseKey` into apollo-federation, which doesn't know about Path. this is an intermediary step in a larger migration, so while I'm not crazy about the API, I have plans to improve things as the migration progresses.

apollo-federation/src/connectors/runtime/key.rs
```diff
@@ -0,0 +1,108 @@
+use std::sync::Arc;
+
+use apollo_compiler::Name;
+use apollo_compiler::executable::FieldSet;
+use apollo_compiler::validation::Valid;
+
+use crate::connectors::JSONSelection;
+use crate::connectors::runtime::inputs::RequestInputs;
+
+#[derive(Clone)]
```
naomijub, line 10: Debug trait is usually good to have as well

apollo-federation/src/connectors/runtime/key.rs
```diff
@@ -0,0 +1,108 @@
...
+                .field("inputs", inputs)
+                .finish(),
+            Self::BatchEntity {
+                selection,
+                keys,
+                inputs,
+            } => f
+                .debug_struct("BatchEntity")
+                .field("selection", &selection.to_string())
+                .field("key_selection", &keys.serialize().no_indent().to_string())
```
naomijub: Why this one changed names?

lennyburdette: no good reason, i'll change it back

apollo-router/src/plugins/connectors/make_requests.rs
lennyburdette: turns out i can delete this function entirely in the next PR!

PR #7740, opened 2025-06-23
feat(federation/composition): port `ValidationState`, except `validateTransition()`

sachindshinde:
This PR ports `ValidationState()` from the JS codebase, except `validateTransition()`. Note that:
1. We've skipped `currentSubgraphs()`, as Rust composition doesn't currently at least give node information.
2. The subgraph context keys were strings in JS, but that was just because they needed to be used in JS `Map`s efficiently. We don't have that restriction in Rust, so we use actual data structures for the keys.
    - As part of this, we don't need actually need a map from subgraph names to subgraph enum values (that was used in JS to avoid issues with separator characters in the keys potentially being in the subgraph names). 
3. We did not port any fixes from https://github.com/apollographql/federation/pull/3277 yet to this code.

### DIFF
--- a/apollo-federation/src/composition/satisfiability.rs
+++ b/apollo-federation/src/composition/satisfiability.rs
@@ -1,4 +1,5 @@
 mod satisfiability_error;
+mod validation_state;
 
 use apollo_compiler::Name;
 use apollo_compiler::Node;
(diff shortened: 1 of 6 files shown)


duckki (approved):
Looks good!

PR #7741, opened 2025-06-23
new error container for connector errors

lennyburdette:
instead of using the more generic `graphql::Error` type, that we don't want to bring into apollo-federation, this is a more specific type that can be converted to `graphql::Error` at the end of mapping.

this is not the ideal API, but it unblocked moving the mapping code into apollo-federation.

---

**Checklist**

- Tests added and passing[^4]
    - [ ] Unit tests
    - [ ] Integration tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-federation/src/connectors/runtime/errors.rs
+++ b/apollo-federation/src/connectors/runtime/errors.rs
@@ -1,0 +1,103 @@
+use serde_json_bytes::ByteString;
+use serde_json_bytes::Map;
+use serde_json_bytes::Value;
+
+use crate::connectors::Connector;
+use crate::connectors::runtime::key::ResponseKey;
+
+#[derive(Debug)]
+pub struct RuntimeError {
+    pub message: String,
+    code: Option<String>,
+    pub coordinate: Option<String>,
+    pub subgraph_name: Option<String>,
+    pub response_key: ResponseKey,
+    pub extensions: Map<ByteString, Value>,
+}
+
+impl RuntimeError {
+    pub fn new(message: impl Into<String>, response_key: &ResponseKey) -> Self {
+        Self {
+            message: message.into(),
+            code: None,
+            coordinate: None,
+            subgraph_name: None,
+            response_key: response_key.clone(),
+            extensions: Default::default(),
+        }
+    }
+
+    pub fn extension<K, V>(mut self, key: K, value: V) -> Self
+    where
+        K: Into<ByteString>,
+        V: Into<Value>,
+    {
+        self.extensions.insert(key.into(), value.into());
+        self
+    }
+
+    pub fn with_code(mut self, code: impl Into<String>) -> Self {
+        self.code = Some(code.into());
+        self
+    }
+
+    pub fn code(&self) -> String {
+        self.code
+            .clone()
+            .unwrap_or_else(|| "CONNECTORS_FETCH".to_string())
+    }
+}
+
+/// An error sending a connector request. This represents a problem with sending the request
+/// to the connector, rather than an error returned from the connector itself.
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[error("Request limit exceeded")]
+    RequestLimitExceeded,
+
+    #[error("Rate limit exceeded")]
+    RateLimited,
+
+    /// Timeout
+    #[error("Request timed out")]
+    GatewayTimeout,
+
+    /// {0}
+    #[error("Connector error: {0}")]
+    TransportFailure(String),
+}
+
+impl Error {
+    pub fn to_runtime_error(
+        &self,
+        connector: &Connector,
+        response_key: ResponseKey,
+    ) -> RuntimeError {
+        RuntimeError {
+            message: self.to_string(),
+            code: Some(self.code()),
+            coordinate: Some(connector.id.coordinate()),
+            subgraph_name: Some(connector.id.subgraph_name.clone()),
+            response_key,
+            extensions: Default::default(),
+        }
+    }
+
+    pub fn code(&self) -> String {
+        match self {
+            Self::RequestLimitExceeded => "REQUEST_LIMIT_EXCEEDED".to_string(),
+            Self::RateLimited => "RATE_LIMIT_EXCEEDED".to_string(),
+            Self::GatewayTimeout => "GATEWAY_TIMEOUT".to_string(),
+            Self::TransportFailure(_) => "HTTP_CLIENT_ERROR".to_string(),
+        }
+    }
+}
+
+impl Clone for Error {
+    fn clone(&self) -> Self {
+        match self {
+            Self::TransportFailure(err) => Self::TransportFailure(err.to_string()),
+            err => err.clone(),
+        }
+    }
+}
--- a/apollo-router/src/plugins/connectors/handle_responses.rs
+++ b/apollo-router/src/plugins/connectors/handle_responses.rs
@@ -8,6 +8,8 @@ use apollo_federation::connectors::ProblemLocation;
 use apollo_federation::connectors::runtime::debug::ConnectorContext;
 use apollo_federation::connectors::runtime::debug::ConnectorDebugHttpRequest;
 use apollo_federation::connectors::runtime::debug::SelectionData;
+use apollo_federation::connectors::runtime::errors::Error;
+use apollo_federation::connectors::runtime::errors::RuntimeError;
 use apollo_federation::connectors::runtime::http_json_transport::HttpResponse;
 use apollo_federation::connectors::runtime::http_json_transport::TransportResponse;
 use apollo_federation::connectors::runtime::key::ResponseKey;
@@ -41,7 +43,6 @@ use crate::plugins::telemetry::consts::OTEL_STATUS_CODE_OK;
 use crate::plugins::telemetry::tracing::apollo_telemetry::emit_error_event;
 use crate::services::connect::Response;
 use crate::services::connector;
-use crate::services::connector::request_service::Error;
 use crate::services::fetch::AddSubgraphNameExt;
 use crate::services::router;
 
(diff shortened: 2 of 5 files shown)


apollo-federation/src/connectors/runtime/errors.rs
```diff
@@ -0,0 +1,103 @@
...
+        self.extensions.insert(key.into(), value.into());
+        self
+    }
+
+    pub fn with_code(mut self, code: impl Into<String>) -> Self {
+        self.code = Some(code.into());
+        self
+    }
+
+    pub fn code(&self) -> String {
```
naomijub: maybe use `as_str` and return a `&str` here, let the caller allocate a string if they need

apollo-federation/src/connectors/runtime/errors.rs
```diff
@@ -0,0 +1,103 @@
...
+    pub message: String,
+    code: Option<String>,
+    pub coordinate: Option<String>,
+    pub subgraph_name: Option<String>,
+    pub response_key: ResponseKey,
+    pub extensions: Map<ByteString, Value>,
+}
+
+impl RuntimeError {
+    pub fn new(message: impl Into<String>, response_key: &ResponseKey) -> Self {
```
naomijub, line 19: I think it is better to be explicit when cloning than passing a reference to clone hiddenly.

apollo-federation/src/connectors/runtime/errors.rs
```diff
@@ -0,0 +1,103 @@
...
+use crate::connectors::Connector;
+use crate::connectors::runtime::key::ResponseKey;
+
+#[derive(Debug)]
+pub struct RuntimeError {
+    pub message: String,
+    code: Option<String>,
+    pub coordinate: Option<String>,
+    pub subgraph_name: Option<String>,
+    pub response_key: ResponseKey,
```
naomijub: to avoid further cloning here, you could do something like:

```rust
#[derive(Debug)]
pub struct RuntimeError<'error> {
    pub message: String,
    code: Option<String>,
    pub coordinate: Option<String>,
    pub subgraph_name: Option<String>,
    pub response_key: &'error ResponseKey,
    pub extensions: Map<ByteString, Value>,
}

impl<'error> RuntimeError<'error> {
    pub fn new(message: impl Into<String>, response_key: &'error ResponseKey) -> Self {
        Self {
            message: message.into(),
            code: None,
            coordinate: None,
            subgraph_name: None,
            response_key,
            extensions: Default::default(),
        }
    }
```

and 
```rust
impl Error {
    pub fn to_runtime_error<'error>(
        &self,
        connector: &Connector,
        response_key: &'error ResponseKey,
    ) -> RuntimeError<'error> {
        RuntimeError::<'error> {
            message: self.to_string(),
            code: Some(self.code()),
            coordinate: Some(connector.id.coordinate()),
            subgraph_name: Some(connector.id.subgraph_name.clone()),
            response_key,
            extensions: Default::default(),
        }
    }
```

apollo-router/src/plugins/connectors/handle_responses.rs
```diff
@@ -56,14 +57,27 @@ pub(crate) enum HandleResponseError {
     MergeError(String),
 }
 
+impl graphql::Error {
+    fn from_connectors_runtime_error(error: &RuntimeError) -> Self {
```
naomijub: As there are multiple clonings going on here, it might be worth checking if cloning the whole object is cheaper? Also, unless we cant change `graphql::Error::builder()` functions to receive references, I would consider using only references

apollo-router/src/plugins/connectors/handle_responses.rs
```diff
@@ -556,13 +538,28 @@ pub(crate) async fn process_response<T: HttpBody>(
...
+    if let MappedResponse::Error { ref error, .. } = mapped_response {
+        emit_error_event(
+            &error.code(),
+            &error.message,
+            Some(Path::from_response_key(&error.response_key)),
+        );
+    }
+
+    connector::request_service::Response {
+        context: context.clone(),
```
naomijub, line 569: Can `Response` receive a reference of `context`? not really sure about the implications of this clone, but as far as I remember it is a map and Response doesnt need ownership of it

apollo-federation/src/connectors/runtime/errors.rs
lennyburdette: no more cloning!

apollo-federation/src/connectors/runtime/errors.rs
lennyburdette: no more cloning!

apollo-router/src/plugins/connectors/handle_responses.rs
lennyburdette: let's sequence this work in the next phase

apollo-federation/src/connectors/runtime/errors.rs
naomijub: This should work:

```rust
 pub fn code(&self) -> &str {
        self.code
            .as_ref()
            .map_or("CONNECTORS_FETCH", |v| v)
    }
```

apollo-router/src/plugins/connectors/handle_responses.rs
lennyburdette: i was able to remove a few clones, the builder does take references. there must be a way to remove the clone of the extensions, but because `error.code()` now returns a reference, i'm getting an "already borrowed" error.

naomijub (approved):

apollo-federation/src/connectors/runtime/errors.rs
```diff
@@ -0,0 +1,101 @@
...
+    /// {0}
+    #[error("Connector error: {0}")]
+    TransportFailure(String),
+}
+
+impl Error {
+    pub fn to_runtime_error(
+        &self,
+        connector: &Connector,
+        response_key: ResponseKey,
```
dylan-apollo:
```suggestion
        response_key: &ResponseKey,
```

apollo-federation/src/connectors/runtime/errors.rs
```diff
@@ -0,0 +1,90 @@
...
+        self
+    }
+
+    pub fn with_code(mut self, code: impl Into<String>) -> Self {
+        self.code = Some(code.into());
+        self
+    }
+
+    pub fn code(&self) -> &str {
+        self.code.as_ref().map_or("CONNECTORS_FETCH", |v| v)
```
dylan-apollo:
```suggestion
        self.code.as_deref().unwrap_or("CONNECTORS_FETCH")
```

apollo-federation/src/connectors/runtime/errors.rs
dylan-apollo: I've taken to putting `#![warn(clippy::pedantic)]` on new files so it will show me optimizations like this

apollo-router/src/services/connector/request_service.rs
```diff
@@ -162,70 +161,9 @@ impl Response {
...
-        match self {
-            Self::RequestLimitExceeded => "REQUEST_LIMIT_EXCEEDED",
-            Self::TransportFailure(_) => "HTTP_CLIENT_ERROR",
-            Self::RateLimited => "REQUEST_RATE_LIMITED",
-            Self::GatewayTimeout => "GATEWAY_TIMEOUT",
-        }
-        .to_string()
+        self.code().to_string()
     }
 }
```
dylan-apollo: Kinda looks like we don't even need this to implement the trait?

apollo-router/src/plugins/connectors/handle_responses.rs
```diff
@@ -488,20 +490,17 @@ pub(crate) async fn process_response<T: HttpBody>(
...
-    match result {
+    let (mapped_response, result) = match result {
         // This occurs when we short-circuit the request when over the limit
         Err(error) => {
             let raw = RawResponse::Error {
-                error: error.to_graphql_error(
-                    connector.clone(),
-                    Some(Path::from_response_key(&response_key)),
-                ),
+                error: error.to_runtime_error(&connector, response_key.clone()),
```
dylan-apollo:
```suggestion
                error: error.to_runtime_error(&connector, &response_key),
```
With another change in another comment

apollo-router/src/plugins/connectors/handle_responses.rs
```diff
@@ -56,14 +58,44 @@ pub(crate) enum HandleResponseError {
     MergeError(String),
 }
 
+impl graphql::Error {
+    fn from_connectors_runtime_error(error: RuntimeError) -> Self {
```
dylan-apollo:
```suggestion
impl From<RuntimeError> for graphql::Error {
    fn from(error: RuntimeError) -> Self {
```

apollo-router/src/plugins/connectors/handle_responses.rs
```diff
@@ -207,9 +234,8 @@ impl RawResponse {
                 };
 
                 // Now we can create the error object using either the default message or the message calculated by the JSONSelection
-                let mut error = graphql::Error::builder()
-                    .message(message)
-                    .path(Path::from_response_key(&key));
+                let mut error = RuntimeError::new(message, &key);
+                error.subgraph_name = Some(connector.id.subgraph_name.clone());
```
dylan-apollo, line 242:
```suggestion
                error.subgraph_name = Some(connector.id.subgraph_name.clone());
                error.coordinate = Some(connector.id.coordinate().into());
```
Or maybe we don't want it in this case for a specific reason?

Wondering if `::new` should take in an optional Connector? Or store an `Arc<Connector>` and spit out subgraph name / coordinate on demand if we're worried about runtime cost?

dylan-apollo (approved):
Kind of a thorny refactor, but looks okay to me. I think my only comments are about performance/style.

apollo-router/src/plugins/connectors/handle_responses.rs
lennyburdette: added a note to come back to this after finishing the main refactoring work

PR #7742, opened 2025-06-23
Create schema-differ function for LSP

conwuegb:
Implement a function which accepts an initial schema as a `String`, expands that schema, and returns the diff as a `String`.

This is the first of two changes toward exposing an endpoint for this purpose in the CLI tool.

Context for this change including example input and output can be found here: [Federation LSP Schema-Diffing Endpoint Overview](https://apollographql.atlassian.net/wiki/x/gYC0Y).

---

**Checklist**

- Tests added and passing[^2]
    - [x] Unit tests

**Exceptions**

*Note any exceptions here*

**Notes**

[^2]: Tick whichever testing boxes are applicable. If you are adding Manual Tests, please document the manual testing (extensively) in the Exceptions.

### DIFF
--- a/apollo-federation/src/link/context_spec_definition.rs
+++ b/apollo-federation/src/link/context_spec_definition.rs
@@ -16,6 +16,7 @@ use crate::error::FederationError;
 use crate::internal_error;
 use crate::link::argument::directive_required_string_argument;
 use crate::link::federation_spec_definition::FEDERATION_CONTEXT_DIRECTIVE_NAME_IN_SPEC;
+use crate::link::federation_spec_definition::FEDERATION_FIELD_ARGUMENT_NAME;
 use crate::link::federation_spec_definition::FEDERATION_FROM_CONTEXT_DIRECTIVE_NAME_IN_SPEC;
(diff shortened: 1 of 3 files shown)


apollo-federation/src/link/context_spec_definition.rs
```diff
@@ -75,10 +76,10 @@ impl ContextSpecDefinition {
     fn field_argument_specification() -> DirectiveArgumentSpecification {
         DirectiveArgumentSpecification {
             base_spec: ArgumentSpecification {
-                name: CONTEXTFIELDVALUE_SCALAR_NAME,
+                name: FEDERATION_FIELD_ARGUMENT_NAME,
                 get_type: |schema, _| {
                     Self::context_field_value_type(schema)
-                        .map(|pos| Type::non_null(Type::Named(pos.type_name)))
+                        .map(|pos| Type::nullable(Type::Named(pos.type_name)))
```
duckki, line 82: Yeah, thanks for the fix. We should be consistent with the doc and JS code for now.
However, it might be a bug in doc and JS code ([FED-675](https://apollographql.atlassian.net/browse/FED-675)), which can be addressed separately.

PR #7745, opened 2025-06-24
[BaseUrl Variables] Allow baseUrl templates to use `$config` and `$env` variables

Hunter-apollo:
### Description
Allow `baseUrl` property of sources to use `$config` and `$env` environment variables in URL templates.

Note: Additional validation is added to ensure no other variables (i.e. `$args`, `$request`, etc) are used in baseUrl templates.

### Manual Testing
In addition to unit and integ tests added in this PR, the following source definition was manually tested with a local mock-RestAPI:
```
  @source(
    name: "user-api"
    http: {
      baseURL: "http://{$config.host}:{$env.PORT}"
    }
  )
```

---

**Checklist**

- Tests added and passing[^4]
    - [X] Unit tests
    - [X] Integration tests
    - [X] Manual tests, as necessary

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/.changesets/docs_lambertjosh_readd_graph_docs.md
+++ b/.changesets/docs_lambertjosh_readd_graph_docs.md
@@ -1,0 +1,5 @@
+### Add back the graph artifact documentation for the containers [PR #7752](hhttps://github.com/apollographql/router/pull/7752)
+
+Adds back accidentally overwritten docs which occurred in PR 7734. The missing commit added graph artifact usage information.
+
+By [lambertjosh](https://github.com/lambertjosh) in https://github.com/apollographql/router/pull/7752
--- a/.changesets/docs_lambertjosh_update_docker_docs_runtime.md
+++ b/.changesets/docs_lambertjosh_update_docker_docs_runtime.md
@@ -1,0 +1,5 @@
+### Updates the Docker deployment instructions to use the Apollo Runtime container [PR #7734](https://github.com/apollographql/router/pull/7734)
+
+With the release of the MCP Server, a method to easily deploy Apollo's runtime services was needed. The Apollo Runtime container was designed to address this need, which includes both the Router and MCP Servers in a single Docker container. This PR updates the Router deployment Docker instructions to use this new container. 
+
+By [lambertjosh](https://github.com/lambertjosh) in https://github.com/apollographql/router/pull/7734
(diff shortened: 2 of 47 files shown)


apollo-federation/src/connectors/spec/source.rs
```diff
@@ -203,15 +202,36 @@ impl BaseUrl {
...
+            Message {
+                code: Code::InvalidUrl,
+                message: format!(
+                    "`@{directive_name}({BASE_URL:})` value {str_value} is not a valid URL Template: {inner}."
+                ),
+                locations: value.line_column_range(sources).into_iter().collect(),
+            }
         })?;
+
+        // ONLY environment vars and config values are permitted in source URL Templates.
```
lennyburdette: should this use a new ["expression context"](https://github.com/apollographql/router/blob/dev/apollo-federation/src/connectors/validation/expression.rs#L74) instead of separate code? the expression validation in that file would also validate that the types of expressions are strings (when known)

Hunter-apollo: Ill take a look at that. My initial read of the `Context` was that it was for Connect directives, but might be useful here.

Hunter-apollo: Updated to use expression context for validation.

apollo-federation/src/connectors/spec/source.rs
```diff
@@ -203,15 +201,18 @@ impl BaseUrl {
...
-            code: Code::InvalidUrl,
-            message: format!(
-                "`@{directive_name}({BASE_URL:})` value {str_value} is not a valid URL: {inner}."
-            ),
-            locations: value.line_column_range(sources).into_iter().collect(),
+        let template: StringTemplate = str_value.parse().map_err(|inner: string_template::Error| {
+            Message {
+                code: Code::InvalidUrl,
+                message: format!(
+                    "`@{directive_name}({BASE_URL:})` value {str_value} is not a valid URL Template: {inner}."
```
lennyburdette: this is not your code, but it looks funny to me and i don't see any tests that result in this message — should it be `{BASE_URL}:`?

apollo-federation/src/connectors/validation/source.rs
```diff
@@ -178,8 +190,24 @@ impl<'schema> SourceDirective<'schema> {
...
             messages.extend(errors.type_check(self.schema).err().into_iter().flatten());
         }
-        if let Some(url) = self.url {
-            messages.extend(url.type_check(self.schema));
+        if let Some(url) = self.base_url {
+            let expression_context =
+                Context::for_source_url(self.schema, &url.node, Code::InvalidUrl);
+            for expression in url.template.expressions() {
+                messages.extend(
+                    expression::validate(expression, &expression_context, &scalars())
```
lennyburdette, line 198: maybe an extra `@source` in invalid_source_url_template.graphql that tries an expression like `{$({})}` to exercise the type checking code?

apollo-federation/src/connectors/spec/source.rs
Hunter-apollo: I think that will just runs default with no formatting args. I removed.

PR #7746, opened 2025-06-24
Fix several hot reload issues with subscriptions

goto-bus-stop:
When a hot reload is triggered by a configuration change, the router attempted to apply updated configuration to open subscriptions. But this could cause excessive logging.

When a hot reload is triggered by a schema change, the router closed subscriptions with a `SUBSCRIPTION_SCHEMA_RELOAD` error. But this happened *before* the new schema was fully active and warmed up, so clients could reconnect tothe _old_ schema.

To fix these issues, a configuration and a schema change now have the same behavior. The router waits for the new configuration and schema to be active, and then closes all subscriptions with a `SUBSCRIPTION_SCHEMA_RELOAD` error, so clients can reconnect.

Drafted pending a test that verifies that the old behaviour is wrong and the new behaviour is correct. No strong indication yet that this is causing serious problems today, but it should be an improvement.

---

**Checklist**

- Tests added and passing[^4]
    - [ ] Unit tests
    - [ ] Integration tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/services/supergraph/service.rs
+++ b/apollo-router/src/services/supergraph/service.rs
@@ -88,6 +88,7 @@ use crate::spec::operation_limits::OperationLimits;
 use crate::uplink::license_enforcement::LicenseState;
 
 pub(crate) const FIRST_EVENT_CONTEXT_KEY: &str = "apollo::supergraph::first_event";
+const SUBSCRIPTION_RELOAD_EXTENSION_CODE: &str = "SUBSCRIPTION_SCHEMA_RELOAD";
 
 /// An [`IndexMap`] of available plugins.
 pub(crate) type Plugins = IndexMap<String, Box<dyn DynPlugin>>;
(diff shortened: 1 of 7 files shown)


apollo-router/src/services/supergraph/service.rs
```diff
@@ -620,101 +603,27 @@ async fn subscription_task(
...
-
-
-                    execution_service_factory = ExecutionServiceFactory {
-                        schema: execution_service_factory.schema.clone(),
-                        subgraph_schemas: execution_service_factory.subgraph_schemas.clone(),
-                        plugins: plugins.clone(),
-                        fetch_service_factory,
-                    };
-                }
+            Some(_new_configuration) = configuration_updated_rx.next() => {
```
goto-bus-stop, line 607: @bnjjj  It seems that we're now handling these in the same way. It's not useful to clients to know that a subscription was closed due to a schema or a configuration change (and maybe even desireable that they don't know). Should we just have a single update channel for both updates, and use the `SUBSCRIPTION_SCHEMA_RELOAD` extension code for both?

bnjjj: Yes I think we should be let's do this in a follow up PR if you agree, I just want to minimize the impact if possible

goto-bus-stop: Sure, but clients might treat `SUBSCRIPTION_SCHEMA_RELOAD` specially, so i think we should at least use the same extension code for both to get the correct behaviour

goto-bus-stop: I've made that change, I'm fine with keeping the notification system intact and just having that duplicate code for the time being.

goto-bus-stop:
Closing in favour of upcoming forward-port of https://github.com/apollographql/router/pull/7777.

PR #7749, opened 2025-06-24
simplify variable reference collection

lennyburdette:
1. collect variable references and *first level* of keys in one pass
2. use that map instead of the redundant Set of namespaces

I had to leave the headers maps alone because they use the *second* level of keys like `$request.headers.x-foo`.

---

**Checklist**

- Tests added and passing[^4]
    - [ ] Unit tests
    - [ ] Integration tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-federation/src/connectors/models.rs
+++ b/apollo-federation/src/connectors/models.rs
@@ -57,15 +57,13 @@ pub struct Connector {
     /// Which version of the connect spec is this connector using?
     pub spec: ConnectSpec,
 
-    pub request_variables: HashSet<Namespace>,
-    pub response_variables: HashSet<Namespace>,
-
     /// The request headers referenced in the connectors request mapping
     pub request_headers: HashSet<String>,
     /// The request or response headers referenced in the connectors response mapping
     pub response_headers: HashSet<String>,
-    /// The environment variables referenced in the connector
-    pub env: HashSet<String>,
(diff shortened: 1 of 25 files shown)


apollo-federation/src/connectors/models.rs
```diff
@@ -195,34 +195,29 @@ impl Connector {
...
-        let response_variables: HashSet<Namespace> = response_references
-            .iter()
-            .map(|var_ref| var_ref.namespace.namespace)
-            .collect();
         let response_headers = extract_header_references(&response_references);
 
-        let env = extract_env_references(&request_references, &response_references);
+        let request_variable_keys = extract_variable_key_references(request_references.iter());
+        let response_variable_keys = extract_variable_key_references(
+            request_references.iter().chain(response_references.iter()),
```
dylan-apollo: Why are we including request references in response variables now? Doesn't that mean that response is always a superset of request?

dylan-apollo (approved):
Seems good to me, just one question below about whether we need all of request also in response. If so, probably a good opportunity for re-using those variables in the future?

I think this moves us toward a world where we can clone/recalculate less per-request, which is great!

apollo-federation/src/connectors/models.rs
lennyburdette: oh you're right, i copied something wrong!

PR #7750, opened 2025-06-24
fix(composition): return error instead of panic for unimplemented methods

dariuszkuc:


duckki (approved):

duckki:
Duh, flaky test failed.

PR #7771, opened 2025-06-25
pq: allow plugins to override PQ operation ID

glasser:
This lets you use alternate clients that put the operation ID in a header, path name, or query parameter instead. This is especially helpful for onboarding existing clients from other PQ systems, and for path-based logging.

---

**Checklist**

- Tests added and passing[^4]
    - [x] Unit tests
    - [ ] Integration tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/docs/source/routing/security/persisted-queries.mdx
+++ b/docs/source/routing/security/persisted-queries.mdx
@@ -184,6 +184,29 @@ To enable safelisting, you _must_ turn off [automatic persisted queries](/router
 
 GraphOS Router can be [customized](/graphos/routing/customization/overview) via several mechanisms such as [Rhai scripts](/graphos/routing/customization/rhai) and [coprocessors](/graphos/routing/customization/coprocessor). These plugins can affect your router's persistent query processing by writing to the request context.
 
+#### `apollo_persisted_queries::operation_id`
+
+By default, Apollo's protocol for sending PQ IDs from clients to GraphOS Router's protocol for sending is to put the ID in a GraphQL request extension named `persistedQuery`, wrapped in an object under the field `sha256Hash`. (This is the same protocol used for [automatic persisted queries](/router/configuration/in-memory-caching#caching-automatic-persisted-queries-apq) (APQ).) That is, the POST body of the request looks something like
+
+```
+{
+  "extensions": {
+    "persistedQuery": {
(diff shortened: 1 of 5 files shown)


docs/source/routing/security/persisted-queries.mdx
```diff
@@ -184,6 +184,29 @@ To enable safelisting, you _must_ turn off [automatic persisted queries](/router
...
+    }
+  }
+}
+```
+
+However, you can design your own protocol for specifying the PQ ID, such as including it in the URL path or a query parameter. (You are responsible for creating a client that sends PQ IDs in this way.)
+
+Your customization (Rhai script, coprocessor, etc) can examine a request during the [Router Service stage](/graphos/routing/customization/overview#request-path) of the request path and set the `apollo_persisted_queries::operation_id` value in the request context to the request's operation ID. If this value is set, GraphOS Router will not look in the `persistedQuery` extension for the ID.
+
+This means that your GraphQL request does not need to contain any of the usual pieces: `query`, `operationName`, `variables`, or `extensions`. If your request is a POST, you should still include a JSON body containing at least `{}`; if your request is a GET, you should include some query string at the end of the URL (even `?q`).
```
goto-bus-stop, line 206: I think this is quite an awkward limitation. Should we rather have a recommendation to replace the body/query of a request entirely? If you have a custom protocol, do `request.uri = '?extensions={"persistedQuery":etc}'` for example (not real syntax because i don't know Rhai well enough offhand 😄 ).

This wouldn't work yet with POSTs because we don't have a hook-point for arbitrary JSON, but hopefully in some months we will. If this isn't super urgent, we could avoid having yet another way to populate the query built in to the router by waiting for the JSON service to exist.

glasser: I like the idea of URL rewriting but it seems like it might be kinda error prone to get all the escaping correct — ie you might have to parse and reserialize and existing extensions thing?

It's been a minute since I sent this so I don't remember if the error in question shows up before the extension runs or not... if the extension runs first then maybe we just skip the error if the PQ ID is in the context?

aaronArinder:
still necessary or closeable? going to close next weds if I don't hear back, but we can always reopen it if it turns out to be needed

PR #7777, opened 2025-06-27
Fix hot reload with subscriptions

bnjjj:
When a hot reload is triggered by a configuration change, the router attempted to apply updated configuration to open subscriptions. But this could cause excessive logging.

When a hot reload is triggered by a schema change, the router closed subscriptions with a `SUBSCRIPTION_SCHEMA_RELOAD` error. But this happened *before* the new schema was fully active and warmed up, so clients could reconnect tothe _old_ schema.

To fix these issues, a configuration and a schema change now have the same behavior. The router waits for the new configuration and schema to be active, and then closes all subscriptions with a `SUBSCRIPTION_SCHEMA_RELOAD` error, so clients can reconnect.

Drafted pending a test that verifies that the old behaviour is wrong and the new behaviour is correct. No strong indication yet that this is causing serious problems today, but it should be an improvement.

---

**Checklist**

- Tests added and passing[^4]
    - [ ] Unit tests
    - [ ] Integration tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/tests/integration/subscriptions/mod.rs
```diff
@@ -175,60 +178,52 @@ pub async fn start_coprocessor_server() -> wiremock::MockServer {
...
-                            );
-                        }
+            chunk_string = format!("{chunk_string}{}", String::from_utf8_lossy(&chunk));
+        }
+        let events = chunk_string
+            .split("\r\n--graphql\r\ncontent-type: application/json\r\n\r\n")
+            .filter_map(|s| {
+                let parsed = serde_json::from_str::<serde_json::Value>(
+                    s.trim_end_matches("\r\n--graphql--\r\n"), // If it's the last event
+                )
```
lrlna: we only grab the graphql chunks? is that what this is?

apollo-router/tests/integration/subscriptions/mod.rs
```diff
@@ -175,60 +178,52 @@ pub async fn start_coprocessor_server() -> wiremock::MockServer {
...
+                    return None;
+                }
+                if parsed == serde_json::json!({}) {
+                    if include_heartbeats {
+                        Some(parsed)
+                    } else {
+                        None
                     }
+                } else {
+                    Some(parsed)
```
lrlna, line 212: to double check this is so we can use heartbeats in testing to verify subscriptions?

apollo-router/tests/integration/subscriptions/ws_passthrough.rs
```diff
@@ -544,3 +560,208 @@ async fn test_subscription_ws_passthrough_pure_error_payload_with_coprocessor()
...
+    let stream = response.bytes_stream();
+    let expected_events = vec![
+        create_initial_empty_response(),
+        create_expected_user_payload(1),
+        create_expected_user_payload(2),
+        create_expected_reload_payload(),
+    ];
+
+    // try to reload the config file
+    router.replace_config_string("replaceable", "replaced");
```
lrlna, line 634: should we also do a config change related to subscriptions?

apollo-router/tests/integration/subscriptions/ws_passthrough.rs
```diff
@@ -544,3 +560,208 @@ async fn test_subscription_ws_passthrough_pure_error_payload_with_coprocessor()
...
+    let stream = response.bytes_stream();
+    let expected_events = vec![
+        create_initial_empty_response(),
+        create_expected_user_payload(1),
+        create_expected_user_payload(2),
+        create_expected_reload_payload(),
+    ];
+
+    // try to reload the config file
+    router.replace_schema_string("createdAt", "created");
```
lrlna, line 736: is this still a valid schema?

apollo-router/tests/integration/subscriptions/mod.rs
bnjjj: Yes

apollo-router/tests/integration/subscriptions/mod.rs
bnjjj: Yes

bnjjj: It's basically in the way we want to write tests if we want to include heartbeats or not

apollo-router/tests/integration/subscriptions/ws_passthrough.rs
bnjjj: No, it's not related to subscriptions, because we're re-using the execution pipeline we can't only reload when it's only subscription config, because it includes all the other plugins so we have to reload everytime the config changes

apollo-router/tests/integration/subscriptions/ws_passthrough.rs
bnjjj: Yes and we don't interact with this part of the supergraph so we're safe

lrlna: should we also check that this works as expected if it's the part of the supergraph we do interact with, or that doesn't matter?

apollo-router/tests/integration/subscriptions/ws_passthrough.rs
lrlna: gotcha!

.changesets/fix_renee_resubscribe_after_reload.md
```diff
@@ -0,0 +1,9 @@
+### Fix several hot reload issues with subscriptions ([PR #7746](https://github.com/apollographql/router/pull/7777))
+
+When a hot reload is triggered by a configuration change, the router attempted to apply updated configuration to open subscriptions. But this could cause excessive logging.
```
garypen:
```suggestion
When a hot reload is triggered by a configuration change, the router attempted to apply updated configuration to open subscriptions. This could cause excessive logging.
```

.changesets/fix_renee_resubscribe_after_reload.md
```diff
@@ -0,0 +1,9 @@
...
+
+When a hot reload is triggered by a schema change, the router closed subscriptions with a `SUBSCRIPTION_RELOAD` error. But this happened *before* the new schema was fully active and warmed up, so clients could reconnect to the _old_ schema.
```
garypen:
```suggestion
When a hot reload was triggered by a schema change, the router closed subscriptions with a `SUBSCRIPTION_RELOAD` error.  This happened *before* the new schema was fully active and warmed up, so clients could reconnect to the _old_ schema, which should not happen.
```

apollo-router/src/notification.rs
```diff
@@ -986,8 +986,8 @@ pub(crate) struct RouterBroadcasts {
 impl RouterBroadcasts {
     pub(crate) fn new() -> Self {
         Self {
-            configuration: broadcast::channel(1),
-            schema: broadcast::channel(1),
+            configuration: broadcast::channel(2),
+            schema: broadcast::channel(2),
```
garypen, line 991: Could we add a comment that explains why 2 slots are needed in the broadcast channel?

apollo-router/tests/integration/subscriptions/ws_passthrough.rs
bnjjj: It doesn't matter because we close the subscription and we make sure in our tests it's not using the old pipeline

apollo-router/src/services/supergraph/service.rs
```diff
@@ -563,54 +562,37 @@ async fn subscription_task(
...
-                    };
-                }
+            Some(_new_configuration) = configuration_updated_rx.next() => {
+                let _ = sender
+                    .send(
+                        Response::builder()
+                            .subscribed(false)
+                            .error(
+                                graphql::Error::builder()
+                                    .message("subscription has been closed due to a router reload")
```
garypen:
```suggestion
                                    .message("subscription has been closed due to a configuration reload")
```

apollo-router/src/services/supergraph/service.rs
```diff
@@ -563,54 +562,37 @@ async fn subscription_task(
...
-                        )
-                        .await;
+            Some(_new_schema) = schema_updated_rx.next() => {
+                let _ = sender
+                    .send(
+                        Response::builder()
+                            .subscribed(false)
+                            .error(
+                                graphql::Error::builder()
+                                    .message("subscription has been closed due to a router reload")
```
garypen:
```suggestion
                                    .message("subscription has been closed due to a schema reload")
```

apollo-router/src/services/supergraph/snapshots/apollo_router__services__supergraph__tests__subscription_callback_schema_reload-3.snap
```diff
@@ -6,9 +6,9 @@ expression: stream.next_response().await.unwrap()
   "data": null,
   "errors": [
     {
-      "message": "subscription has been closed due to a schema reload",
+      "message": "subscription has been closed due to a router reload",
       "extensions": {
-        "code": "SUBSCRIPTION_SCHEMA_RELOAD"
+        "code": "SUBSCRIPTION_RELOAD"
```
garypen: Will it confuse any clients receiving these errors that we have changed the code? Could anything be explicitly depending on the code being "SUBSCRIPTION_SCHEMA_RELOAD"?

apollo-router/tests/integration/subscriptions/mod.rs
```diff
@@ -175,60 +178,52 @@ pub async fn start_coprocessor_server() -> wiremock::MockServer {
...
-                        if let Ok(parsed) = serde_json::from_str::<serde_json::Value>(json_str) {
-                            // Store the raw parsed response without any transformation
-                            subscription_events.push(parsed.clone());
-                            info!(
-                                "Received subscription event {}: {}",
-                                subscription_events.len(),
-                                parsed
-                            );
-                        }
+            chunk_string = format!("{chunk_string}{}", String::from_utf8_lossy(&chunk));
```
garypen: Using `format!()` to concat strings seems clunky. Something like
```suggestion
            chunk_string += &String::from_utf8_lossy(&chunk);
```
seems clearer to me.

apollo-router/tests/integration/subscriptions/mod.rs
```diff
@@ -175,60 +178,52 @@ pub async fn start_coprocessor_server() -> wiremock::MockServer {
...
+        }
+        let events = chunk_string
+            .split("\r\n--graphql\r\ncontent-type: application/json\r\n\r\n")
```
garypen: Maybe make a constant or multiple constants?

apollo-router/src/services/supergraph/snapshots/apollo_router__services__supergraph__tests__subscription_callback_schema_reload-3.snap
bnjjj: No I thought our clients were using this error code but not, they're just using a catch all on all errors and retry the connection if there's an error so we're good

apollo-router/src/services/supergraph/service.rs
bnjjj: I intentionally make this message less obvious for the clients because I think it's not something interesting for the clients and the router already has metrics to describe different kind of events triggered in the state machine

apollo-router/src/services/supergraph/service.rs
bnjjj: I intentionally make this message less obvious for the clients because I think it's not something interesting for the clients and the router already has metrics to describe different kind of events triggered in the state machine

apollo-router/src/services/supergraph/snapshots/apollo_router__services__supergraph__tests__subscription_callback_schema_reload-3.snap
abernix: is this is better as a follow-up? I definitely think the `code` might be more sensibly named with the new token you've selected which more closely reflects that it might be the supergraph **or** the router configuration which have changed, which definitely doesn't matter to the client which.  This change just very much exceeds the scope of this PR which is about fixing hot-reloading.

garypen: If we do change this, and I'm not convinced that we should because there may be clients we don't know about, then we need to clearly highlight it in the CHANGELOG entry and possibly in the docs.

bnjjj: I rollbacked

SimonSapin (approved):

goto-bus-stop:
@mergifyio backport dev

PR #7781, opened 2025-06-29
Error counting: move to telemetry

timbotnik:
---

**Checklist**

- Tests added and passing[^4]
    - [ ] Unit tests
    - [ ] Integration tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/tests/integration/telemetry/apollo_otel_metrics.rs
```diff
@@ -39,93 +49,494 @@ async fn test_validation_error() {
...
+            .attribute("apollo.router.error.service", expected_service)
+            .value(1)
+            .build(),
+    );
+    router.graceful_shutdown().await;
+}
+
+/// Assert that the given metric exists in the list of Otel requests. This is a crude attempt at
+/// replicating _some_ assert_counter!() functionality since that test util can't be accessed here.
+fn assert_metrics_contain(actual_metrics: &[ExportMetricsServiceRequest], expected_metric: Metric) {
```
rregitsky, line 498: This is a quick and dirty replication of `assert_counter!()`. It could certainly be improved and more generalized, but at some point I had to move on and we can fix it iteratively.

apollo-router/tests/common.rs
```diff
@@ -995,20 +996,28 @@ impl IntegrationTest {
...
-        )
-        .await;
+
+        while Instant::now() < deadline {
+            if let Some(msg) = self.apollo_otlp_metrics_rx.recv().await {
+                // Only break once we see a batch with metrics in it
+                if msg.resource_metrics.iter().any(|rm| !rm.scope_metrics.is_empty()) {
+                    metrics.push(msg);
+                    break;
+                }
```
timbotnik, line 1017: Should we have a sleep here?  I suppose the `await` will block on receiving data so maybe we don't need it.

apollo-router/src/plugins/license_enforcement/mod.rs
```diff
@@ -183,28 +183,69 @@ mod test {
...
+
+            // Put the error response in an arc and mutex so we can share it with telemetry threads
+            let slot = Arc::new(Mutex::new(Some(license_plugin_error_response)));
+            // We have to do a weird thing where we take the response from the license plugin and feed
+            // it as the mock response of the telemetry plugin so that telemetry plugin will count
+            // the errors. Ideally this would be done using a TestHarness, but using a "full"
+            // router with the Telemetry plugin will hit reload_metrics() on activation thus
+            // breaking async(){}.with_metrics() by shutting down its metrics provider.
+            // Ultimately this is the best way anyone could think of to simulate this scenario.
+            let _telemetry_service = PluginTestHarness::<Telemetry>::builder()
```
timbotnik, line 215: I wonder if this is now better covered as an integration test?  I see later on that we use the same pattern in other unit tests, so perhaps this is just the best we've got right now.

apollo-router/src/plugins/telemetry/error_counter.rs
```diff
@@ -0,0 +1,954 @@
...
+        id: response.id,
+        response: response.response,
+    }
+}
+
+pub(crate) async fn count_supergraph_errors(
+    response: SupergraphResponse,
+    errors_config: &ErrorsConfiguration,
+) -> SupergraphResponse {
+    // TODO streaming subscriptions?
```
timbotnik: Some TODOs here - we should remove them if they are done / tested or otherwise leave more informative comments.

apollo-router/src/plugins/telemetry/error_counter.rs
```diff
@@ -0,0 +1,954 @@
...
+    // TODO multiple responses in the stream?
+
+    let context = response.context.clone();
+    let errors_config = errors_config.clone();
+
+    let (parts, stream) = response.response.into_parts();
+
+    let stream = stream.inspect(move |response_body| {
+        // TODO ensure free plan is captured
```
timbotnik: TODO here

apollo-router/src/plugins/telemetry/error_counter.rs
```diff
@@ -0,0 +1,954 @@
...
+            operation_id = usage_reporting.get_operation_id();
+        }
+
+        // Also try to get operation name from usage reporting if it's not in context
+        if operation_name.is_empty() {
+            operation_name = usage_reporting.get_operation_name();
+        }
+    }
+
+    // TODO how do we account for redacted errors when comparing? Likely skip them completely (they will have been counted with correct codes in subgraph layer)
```
timbotnik: TODOs here

apollo-router/src/plugins/telemetry/mod.rs
```diff
@@ -528,7 +534,7 @@ impl PluginPrivate for Telemetry {
 
                         let span = Span::current();
                         span.set_span_dyn_attributes(custom_attributes);
-                        let response: Result<router::Response, BoxError> = fut.await;
+                        let mut response: Result<router::Response, BoxError> = fut.await;
```
timbotnik: Could we explain why this needs to be mutable in a comment?

apollo-router/src/plugins/telemetry/mod.rs
```diff
@@ -892,7 +918,7 @@ impl PluginPrivate for Telemetry {
                     async move {
                         let span = Span::current();
                         span.set_span_dyn_attributes(custom_attributes);
-                        let result: Result<SubgraphResponse, BoxError> = f.await;
+                        let mut result: Result<SubgraphResponse, BoxError> = f.await;
```
timbotnik: As above

apollo-router/tests/integration/telemetry/apollo_otel_metrics.rs
```diff
@@ -39,93 +49,494 @@ async fn test_validation_error() {
...
-                    experimental_otlp_metrics_protocol: http
-                    batch_processor:
-                        scheduled_delay: 1s # lowering this seems to make the test flaky
-                    errors:
-                        preview_extended_error_metrics: enabled
-            "#,
-        )
-        .responder(ResponseTemplate::new(500).append_header("Content-Type", "application/json"))
+        .config(include_str!(
+            "fixtures/apollo_otel_metrics_include_subgraph_errors_enabled.router.yaml"
```
timbotnik: Personally I prefer the inline config since it makes the test much more self-contained, especially if there are very_long_names_of_files_with_specific_test_cases.

apollo-router/tests/integration/telemetry/apollo_otel_metrics.rs
timbotnik: Having looked at this quickly I think it is a great trade-off.

apollo-router/tests/integration/telemetry/fixtures/apollo_otel_metrics_csrf_required_headers.router.yaml
```diff
@@ -0,0 +1,10 @@
+telemetry:
```
timbotnik: IMO these files are not big enough to warrant existence vs. inline config.

apollo-router/src/plugins/license_enforcement/mod.rs
rregitsky: I had that thought as well. I'm not sure if we can set the license config properly to trigger the test conditions in an integration test. I'll give it a look over today and see if I can figure anything out.

apollo-router/tests/common.rs
rregitsky: Yeah it should block until `recv()` gets a hit

apollo-router/tests/integration/telemetry/apollo_otel_metrics.rs
rregitsky: I wasn't loving it as I was doing it. My preferred method would be pulling out the telemetry part of the config so tests could be less verbose and it'd make really clear which part of the config "mattered" for the test. But doing that in Rust was nearly as verbose as putting the whole config in there. 🤷 

I moved it inline since I don't have a strong opinion

apollo-router/src/metrics/mod.rs
```diff
@@ -389,8 +376,15 @@ pub(crate) mod test_utils {
...
-            attrs1.iter().zip(attrs2.iter()).all(|((k, v), kv)| {
+        fn equal_attributes(expected: &AttributeSet, actual: &[KeyValue]) -> bool {
+            // If lengths are different, we can short circuit. This also accounts for a bug where
+            // an empty attributes list would always be considered "equal" due to zip capping at
+            // the shortest iter's length
+            if expected.iter().count() != actual.len() {
+                return false;
+            }
+            // This works because the attributes are always sorted
+            expected.iter().zip(actual.iter()).all(|((k, v), kv)| {
```
rregitsky, line 387: Noting that I fixed a bug here where these would have always passed regardless of the actual number of attributes in a matching emitted metric due to how `zip` works.:
```rust
// This used to pass
u64_counter!('my.other.metric', 1, attr = "val");
assert_counter!('my.metric.count', 1, &[])
assert_counter!('my.metric.count', 1)

// This also used to pass
u64_counter!('my.other.metric', 1);
assert_counter!('my.other.metric, 1, attr = "val");
```

Didn't want it to get lost in the massive deletions in this file.

rregitsky: E.g. this miswritten existing test that should have been failing: https://github.com/apollographql/router/blob/52ce4daa23d65a0cf72d43a51bc224bd150989b7/apollo-router/src/metrics/mod.rs#L1772-L1773

apollo-router/src/plugins/license_enforcement/mod.rs
rregitsky: I dug into this a bit -- I think we'd need a JWT and to mock out the private key to encode it that the jwks library is expecting. Guessing that's not gonna be easy. I plan to ask about it in `#team-runtime-readiness` on Monday in case they have any ideas. At this point I'd prefer to do a fast-follow if it's involved.

timbotnik: Seems fine not to fix in this PR 👍

apollo-router/src/plugins/telemetry/mod.rs
```diff
@@ -607,7 +613,11 @@ impl PluginPrivate for Telemetry {
                             custom_events.on_error(err, &ctx);
                         }
 
-                        response
+                        if let Ok(resp) = response {
+                            Ok(count_router_errors(resp, &config.apollo.errors).await)
```
timbotnik, line 617: Not sure how hard this is, but wondering if instead of rebuilding the response we could use a mutator like `resp.count_router_errors`.  Might save some clones.

timbotnik (commented):
This is looking pretty good to me.  One question but I'd be pushing for a final Router team review.

apollo-router/src/plugins/telemetry/mod.rs
rregitsky: Hmm... I'm not sure there's a good way to avoid it. I have to split the `response` (technically the `response.response`) into parts to be able to pull out the errors. That action consumes the original response which I believe means we are forced to rebuild it. Happy to be told otherwise though.

The only exception is the router layer, in which the errors are sitting on the context. I'd prefer to keep a similar pattern between all layers though instead of mutating the response in the router layer only.

BrynCooke (changes requested):
Hopefully fairly simple, but the telemetry mod.rs is now 4000 lines. Please can you split the tests for the new functionality into a new module so that we can start breaking stuff up.

apollo-router/src/lib.rs
```diff
@@ -43,7 +43,7 @@ macro_rules! failfast_error {
 }
 
 #[macro_use]
-mod json_ext;
+pub mod json_ext;
```
BrynCooke: Why was this made public? Can it be pub (crate).

apollo-router/src/context/mod.rs
```diff
@@ -29,6 +29,10 @@ pub(crate) const OPERATION_NAME: &str = "apollo::supergraph::operation_name";
 pub(crate) const OPERATION_KIND: &str = "apollo::supergraph::operation_kind";
 /// The key to know if the response body contains at least 1 GraphQL error
 pub(crate) const CONTAINS_GRAPHQL_ERROR: &str = "apollo::telemetry::contains_graphql_error";
+/// The key to a map of errors that were already counted in a previous layer
```
BrynCooke: Document these to say they may change and are not supported for user access.

BrynCooke (changes requested):

apollo-router/src/lib.rs
rregitsky: It's necessary to allow us to use `Path` in integration tests. Unfortunately, `pub(crate)` visibility isn't enough. https://github.com/apollographql/router/blob/81f6be9654bcb9918766c08587ca47bcd966a52f/apollo-router/tests/integration/telemetry/apollo_otel_metrics.rs#L8

If you have any other ideas of how to use it please let me know.

rregitsky: Usage ex: https://github.com/apollographql/router/blob/81f6be9654bcb9918766c08587ca47bcd966a52f/apollo-router/tests/integration/telemetry/apollo_otel_metrics.rs#L181

rregitsky: Hmm, actually it seems like the path fn can just take in a &str. I'll change it.

PR #7787, opened 2025-07-01
Add SQL query metadata update script

SimonSapin:


goto-bus-stop:
This feels like it should be an xtask(?)

abernix:
@mergifyio copy dev

abernix:
I'll take over the copy.

PR #7788, opened 2025-07-01
Add support of `@cacheTag` directive for interfaces and add `$key.__typename` in format

bnjjj:
This adds support for `@cacheTag` to be set on GraphQL `interface` (only if there's no cacheTag already set on the type itself). 

Example of supergraph:
```graphql
# ...
interface Siret
    @join__type(graph: ORGA, key: "id")
    @join__type(graph: USER, key: "id")
    @join__directive(
        graphs: [USER, ORGA]
        name: "federation__cacheTag"
        args: { format: "organization" }
    )
    @join__directive(
        graphs: [USER, ORGA]
        name: "federation__cacheTag"
        args: { format: "organization-{$key.__typename}-{$key.id}" }
    )
    @join__directive(
        graphs: [USER, ORGA]
        name: "federation__cacheTag"
        args: { format: "organization-{$key.id}" }
    ) {
    id: ID!
    siret: String @join__field(graph: ORGA)
}

type Organization implements Siret
    @join__type(graph: ORGA, key: "id")
    @join__type(graph: USER, key: "id")
    @join__implements(graph: ORGA, interface: "Siret")
    @join__implements(graph: USER, interface: "Siret") {
    id: ID
    creatorUser: User @join__field(graph: ORGA)
    name: String @join__field(graph: ORGA)
    nonNullId: ID! @join__field(graph: ORGA)
    suborga: [Organization] @join__field(graph: ORGA)
    siret: String @join__field(graph: ORGA)
}
# ...
```

This example will generate cache tags like: 
```
"organization",
"organization-1",
"organization-Organization-1"
```

---

**Checklist**

- Tests added and passing[^4]
    - [x] Unit tests
    - [x] Integration tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-1351]: https://apollographql.atlassian.net/browse/ROUTER-1351?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apollo-router/src/plugins/response_cache/plugin.rs
+++ b/apollo-router/src/plugins/response_cache/plugin.rs
@@ -1620,49 +1620,43 @@ fn get_invalidation_entity_keys_from_schema(
     typename: &str,
     entity_keys: &serde_json_bytes::Map<ByteString, Value>,
 ) -> Result<HashSet<String>, anyhow::Error> {
-    let field_def =
+    let type_def =
         supergraph_schema
             .get_object(typename)
             .ok_or_else(|| FetchError::MalformedRequest {
                 reason: "can't find corresponding type for __typename {typename:?}".to_string(),
             })?;
-    let cache_keys = field_def
+    let implements_cache_keys = type_def
+        .implements_interfaces
+        .iter()
+        .filter_map(|interface_comp| {
+            let interface = supergraph_schema.get_interface(interface_comp.name.as_str())?;
+            let cache_keys =
+                interface
+                    .directives
+                    .get_all("join__directive")
+                    .filter_map(|directive| {
+                        get_format_from_cache_tag_directive(
+                            supergraph_schema,
+                            subgraph_name,
+                            subgraph_enums,
+                            directive,
+                        )
+                    });
+
+            Some(cache_keys)
+        })
+        .flatten();
+    let cache_keys = type_def
         .directives
         .get_all("join__directive")
-        .filter_map(|dir| {
-            let name = dir.argument_by_name("name", supergraph_schema).ok()?;
-            if name.as_str()? != CACHE_TAG_DIRECTIVE_NAME {
-                return None;
-            }
-            let is_current_subgraph = dir
-                .argument_by_name("graphs", supergraph_schema)
-                .ok()
-                .and_then(|f| {
-                    Some(
-                        f.as_list()?
-                            .iter()
-                            .filter_map(|graph| graph.as_enum())
-                            .any(|g| {
-                                subgraph_enums.get(g.as_str()).map(|s| s.as_str())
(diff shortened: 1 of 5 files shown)


lennyburdette:
are y'all still open to discussing `$this` instead of `$key` for alignment with connectors?

bnjjj:
@lennyburdette yes

apollo-router/src/plugins/response_cache/plugin.rs
```diff
@@ -1620,58 +1620,102 @@ fn get_invalidation_entity_keys_from_schema(
...
 ) -> Result<HashSet<String>, anyhow::Error> {
-    let field_def =
+    let type_def =
         supergraph_schema
             .get_object(typename)
             .ok_or_else(|| FetchError::MalformedRequest {
                 reason: "can't find corresponding type for __typename {typename:?}".to_string(),
             })?;
-    let cache_keys = field_def
+    let implements_cache_keys = type_def
```
duckki, line 1629: The validation rule will ensure all implementor types inherit all cache tags from their interfaces. So, we won't need to check cache tags on interfaces.

apollo-router/src/plugins/response_cache/plugin.rs
```diff
@@ -1620,58 +1620,102 @@ fn get_invalidation_entity_keys_from_schema(
...
         });
     let mut vars = IndexMap::default();
-    vars.insert("$key".to_string(), Value::Object(entity_keys.clone()));
-    let invalidation_cache_keys = cache_keys
+    let mut key_vars = entity_keys.clone();
+    key_vars.insert(
+        ByteString::from(TYPENAME.to_string()),
+        Value::String(typename.to_string().into()),
+    );
+    vars.insert("$key".to_string(), Value::Object(key_vars.clone()));
```
duckki, line 1667: For initial GA release, I propose not to support `$key.__typename` in format values, since `__typename` can be too dynamic for interface cache tags.

We may allow object types to have additional cache tags (in addition to the inherited cache tags from their interfaces) that are more specific for the types, which I think safer than `__typename`.

duckki (changes requested):

bnjjj:
@duckki Regarding your comments I think there's nothing to do with interface. I think we only support `@cacheTag` for consistency right now but there's no real advantage of setting this `@cacheTag` directive on interface instead of types directly with the constraints we want to have. We will probably change this behavior in future version to have real use cases using interface but for now it's mainly to be consistent. I'm closing this PR because it's not needed now

duckki:
Ok. We can add interface support later.

PR #7790, opened 2025-07-01
Port merge utility for merging directive applications

tninesling:
# Directive Application Merging

This change ports [mergeAppliedDirectives](https://github.com/apollographql/federation/blob/main/composition-js/src/merging/merge.ts#L2968-L3062) and the related [reportMismatchHint](https://github.com/apollographql/federation/blob/main/composition-js/src/merging/reporter.ts#L100-L142).

## mergeAppliedDirectives

Given a directive name and a set of sources, this will find all applications of that directive and merge them together. Directives are bucketed by their arguments, and directives with matching arguments are considered to be the same application. For repeatable directives, all unique applications are propagated to the supergraph. For directives that aren't repeatable, we merge the arguments with the supplied argument merger, if one exists. Otherwise, we simply use the unique application which appears most frequently across all subgraphs.

The deduplication works slightly differently than JS, which loops over the applications many times to filter out duplicates. In this version, we rely on the `Hash` implementation to deduplicate applications when we add them to the `HashMap` tracking their counts.

## reportMismatchHint

This is a similar utility to `reportMismatchError`, and it delegates to the more generic `reportMismatch`. I made a slight update to `reportMismatch` so the closure passed in actually adds the error or hint to the correct collection. Previously it was automatically adding to the errors list, which is not valid for hints. This required adding a new parameter of type `&mut Self` to the closure, since we cannot capture `self` at the time the closure is defined. Instead, the `self` reference is passed into the closure at the time it is called.

---

**Checklist**

- Tests added and passing[^4]
    - [ ] Unit tests
    - [ ] Integration tests

**Exceptions**

*Note any exceptions here*

**Notes**

[FED-553]: https://apollographql.atlassian.net/browse/FED-553?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apollo-federation/src/merger/merger.rs
```diff
@@ -564,6 +580,149 @@ impl Merger {
...
+                        vec![]
+                    }
+                })
+                .counts();
+
+            if directive_in_supergraph.definition.repeatable {
+                for directive in directive_counts.keys() {
+                    pos.insert_directive(dest, (*directive).clone())?;
+                }
+            } else if directive_counts.len() == 1 {
```
duckki, line 635: Why don't we use `crate::utils::iter_into_single_item` here, instead of `unwrap`?

duckki (approved):
Just a nit. Otherwise, it looks good!

apollo-federation/src/merger/merger.rs
tninesling: `iter_into_single_item` is still going to give me an `Option`, so I would have to unwrap anyway. I usually try to avoid it, but we check immediately above that the `len` is 1. I think that's local enough that folks can see the invariant holds

duckki: I meant to change the condition to match with the `Option` result, so you can avoid `.unwrap()`.

PR #7791, opened 2025-07-01
Re-implement and promote ->and connectors method

andrewmcgivery:
Re-implement and promote ->and connectors method
---

**Checklist**

- Tests added and passing[^4]
    - [x] Unit tests
    - [ ] Integration tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/plugins/response_cache/plugin.rs
```diff
@@ -470,9 +470,11 @@ impl PluginPrivate for ResponseCache {
 }
 
 #[cfg(test)]
+#[allow(dead_code)]
```
dylan-apollo: Are these from some conflict upstream? Maybe better to make them their own PR to get the responsible party to review?

apollo-federation/src/connectors/json_selection/methods/public/and.rs
```diff
@@ -0,0 +1,332 @@
...
+                method_name.range(),
+            )],
+        );
+    };
+
+    let mut errors = Vec::new();
+    for arg in args {
+        if !result {
+            break;
+        }
```
dylan-apollo, line 60: I think we've decided that we should evaluate every argument in this and `->or` and generate errors if any of them error instead of short circuit? cc @benjamn

dylan-apollo (commented):
Just want to double-check on the desired error/none behavior vs short-circuit before we merge this and forget about it.

apollo-router/src/plugins/response_cache/plugin.rs
andrewmcgivery: Yea, I accidentally left these in. Removing!

apollo-federation/src/connectors/json_selection/methods/public/and.rs
andrewmcgivery: So @dylan-apollo are you suggesting that both `and`/`or` should be updated to remove this early `break` and evaluate every argument?

benjamn: My position is that we should keep the short-circuiting at runtime (that is, here in `and_method`), but that `and_shape` should compute the shapes of all arguments at shape processing time (to prevent runtime surprises). I believe developers will expect the runtime short-circuiting, and also expect that runtime errors from the unevaluated arguments will be omitted/ignored.

apollo-federation/src/connectors/json_selection/methods/public/and.rs
```diff
@@ -0,0 +1,332 @@
...
+        );
+    }
+
+    #[test]
+    fn and_shape_should_error_on_none_args() {
+        let location = get_location();
+        assert_eq!(
+            and_shape(
+                &WithRange::new("and".to_string(), Some(location.span)),
+                None,
```
benjamn, line 324: What if an argument's shape computes as `None`? Can we write a test for that? I think it will already produce an error, as desired, but I'm not 100% sure.

apollo-federation/src/connectors/json_selection/methods/public/and.rs
```diff
@@ -0,0 +1,332 @@
...
+            ),
+            method_name.shape_location(source_id),
+        );
+    };
+
+    // We will accept anything bool-like OR unknown/named
+    if !(Shape::bool([]).accepts(&input_shape) || input_shape.accepts(&Shape::unknown([]))) {
+        return Shape::error(
+            format!(
+                "Method ->{} can only be applied to boolean values.",
```
benjamn:
```suggestion
                "Method ->{} can only be applied to boolean values. Got {input_shape}.",
```
More information for debugging. Feel free to tweak wording.

apollo-federation/src/connectors/json_selection/methods/public/and.rs
```diff
@@ -0,0 +1,332 @@
...
+                dollar_shape.clone(),
+                named_var_shapes,
+                source_id,
+            );
+
+            // We will accept anything bool-like OR unknown/named
+            if !(Shape::bool([]).accepts(&arg_shape) || arg_shape.accepts(&Shape::unknown([]))) {
+                return Shape::error(
+                    format!(
+                        "Method ->{} can only accept boolean arguments.",
```
benjamn:
```suggestion
                        "Method ->{} can only accept boolean arguments. Got {arg_shape} at position {i}",
```
Note the `i` variable doesn't exist yet, so don't merge this suggestion as-is. Probably something like `for (i, arg) in args.enumerate()` above.

benjamn (approved):
LGTM, with just a few suggestions.

apollo-federation/src/connectors/json_selection/methods/public/and.rs
andrewmcgivery: So... no changes to my code then? The shape methods don't have the short circuit. Only the runtime does.

PR #7793, opened 2025-07-01
Add ->find connectors method, add shape tests to ->filter

andrewmcgivery:
Add ->find connectors method, add shape tests to ->filter

---

**Checklist**

- Tests added and passing[^4]
    - [x] Unit tests
    - [ ] Integration tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-federation/src/connectors/json_selection/methods/public/find.rs
+++ b/apollo-federation/src/connectors/json_selection/methods/public/find.rs
@@ -13,21 +13,23 @@ use crate::connectors::json_selection::location::Ranged;
 use crate::connectors::json_selection::location::WithRange;
 use crate::impl_arrow_method;
 
-impl_arrow_method!(FilterMethod, filter_method, filter_shape);
-/// "Filter" is an array method that returns a new array containing only items that match the criteria.
-/// You can use it to filter an array of values based on a boolean condition.
+impl_arrow_method!(FindMethod, find_method, find_shape);
+/// "Find" is an array method that returns the first item that matches the criteria.
+/// You can use it to find the first value in an array based on a boolean condition.
+/// If no matching item is found, it returns None.
 ///
 /// For example, given a selection of [1, 2, 3, 4, 5]:
 ///
-/// $->filter(@->eq(3))      result is [3]
-/// $->filter(@->gt(3))      result is [4, 5]
+/// $->find(@->eq(3))      result is 3
+/// $->find(@->gt(3))      result is 4
+/// $->find(@->eq(10))     result is None (no match)
 ///
-/// We are taking each value passed into filter via @ and running the condition function against that value.
-/// Only values where the condition returns true will be included in the result array.
+/// We are taking each value passed into find via @ and running the condition function against that value.
+/// The first value where the condition returns true will be returned.
 ///
 /// Example with objects:
-/// users->filter(@.active->eq(true))    returns only users where active is true
-fn filter_method(
+/// users->find(@.active->eq(true))    returns the first user where active is true
+fn find_method(
     method_name: &WithRange<String>,
     method_args: Option<&MethodArgs>,
     data: &JSON,
@@ -46,9 +48,7 @@ fn filter_method(
     };
 
     if let JSON::Array(array) = data {
-        let mut output = Vec::new();
         let mut errors = Vec::new();
-        let mut has_non_boolean_error = false;
 
         for (i, element) in array.iter().enumerate() {
             let input_path = input_path.append(JSON::Number(i.into()));
@@ -57,14 +57,14 @@ fn filter_method(
 
             match applied_opt {
                 Some(JSON::Bool(true)) => {
-                    output.push(element.clone());
+                    // Found the first matching element, return it
+                    return (Some(element.clone()), errors);
                 }
                 Some(JSON::Bool(false)) | None => {
-                    // Condition is false or errored, exclude the element
+                    // Condition is false or errored, continue searching
                 }
                 Some(_) => {
                     // Condition returned a non-boolean value, this is an error
-                    has_non_boolean_error = true;
                     errors.push(ApplyToError::new(
                         format!(
                             "->{} condition must return a boolean value",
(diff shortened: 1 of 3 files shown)


apollo-federation/src/connectors/json_selection/methods/public/filter.rs
```diff
@@ -283,3 +283,125 @@ mod tests {
...
+            )
+        );
+    }
+
+    #[test]
+    fn filter_shape_should_handle_unknown_condition_shape() {
+        let input_shape = Shape::list(Shape::int([]), []);
+        // Unknown shapes should be accepted as they could produce boolean values at runtime
+        let result = get_shape(
+            vec![WithRange::new(LitExpr::Bool(true), None)], // This will create a bool shape
```
dylan-apollo: Wouldn't this end up being a bool shape, so it's not testing unknown? Maybe we should stick a `Shape::name` here instead?

apollo-federation/src/connectors/json_selection/methods/public/find.rs
```diff
@@ -0,0 +1,407 @@
...
+    dollar_shape: Shape,
+    named_var_shapes: &IndexMap<&str, Shape>,
+    source_id: &SourceId,
+) -> Shape {
+    let Some(first_arg) = method_args.and_then(|args| args.args.first()) else {
+        return Shape::error(
+            format!("Method ->{} requires one argument", method_name.as_ref()),
+            method_name.shape_location(source_id),
+        );
+    };
```
dylan-apollo, line 137: Need to also error on multiple args

apollo-federation/src/connectors/json_selection/methods/public/find.rs
```diff
@@ -0,0 +1,407 @@
...
+            format!(
+                "->{} condition must return a boolean value",
+                method_name.as_ref()
+            ),
+            method_name.shape_location(source_id),
+        );
+    }
+
+    // Find returns a single item (or None), so we return the item type of the input shape
+    input_shape.any_item([])
```
dylan-apollo: Does `.any_item` return a union of the item or None? If not, we need to make that union ourselves, since it's always possible for `->find` to return None, right?

dylan-apollo: Although this is another case where the user might have more information than us, and if we end up erroring when they try to assign to a non-nullable field, that might be problematic 🤔

andrewmcgivery: I did this to match `filter` which @benjamn suggested the `any_item` shape :)

https://github.com/apollographql/router/pull/7646#discussion_r2145219251

apollo-federation/src/connectors/json_selection/methods/public/filter.rs
andrewmcgivery: You're 100% right. Good catch.

I swapped it to a `LitExpr::path` instead.

apollo-federation/src/connectors/json_selection/methods/public/find.rs
dylan-apollo: Well `filter` returns an array where each item is one of the elements of the previous array. `find` is different, because it returns a single item, which is _either_ one of the elements of the previous array _or_ None. So I think we need to make the union here.

dylan-apollo: But we need to test that that's okay in real schemas...

andrewmcgivery: Oh 🤦 You're totally right about `None`. I think I didn't quite register what your comment was saying 😅 I'll fix this!

andrewmcgivery: Updated!

apollo-federation/src/connectors/json_selection/methods/public/filter.rs
```diff
@@ -120,6 +120,17 @@ fn filter_shape(
     named_var_shapes: &IndexMap<&str, Shape>,
     source_id: &SourceId,
 ) -> Shape {
+    let arg_count = method_args.map(|args| args.args.len()).unwrap_or_default();
+    if arg_count > 1 {
```
naomijub, line 124: I think this block can be combined with the following using this:
```rust
if method_args.as_ref().is_none_or(|args| args.len() != 1) {
        return format!(
            "Method ->{} requires only one argument, but {} were provided",
            method_name,
            method_args.map_or(0, |args| args.len())
        )
    }
```

apollo-federation/src/connectors/json_selection/methods/public/find.rs
```diff
@@ -0,0 +1,445 @@
...
+fn find_shape(
+    method_name: &WithRange<String>,
+    method_args: Option<&MethodArgs>,
+    input_shape: Shape,
+    dollar_shape: Shape,
+    named_var_shapes: &IndexMap<&str, Shape>,
+    source_id: &SourceId,
+) -> Shape {
+    let arg_count = method_args.map(|args| args.args.len()).unwrap_or_default();
+    if arg_count > 1 {
```
naomijub, line 122: `is_none_or can be applied here as well`

apollo-federation/src/connectors/json_selection/methods/public/filter.rs
andrewmcgivery: I almost like this but I don't like that we have to do the additional `map_or` to get the count a second time in order to return the error. Also this ends up being inconsistent with the other methods which are doin the `arg_count` as I currently have it.

apollo-federation/src/connectors/json_selection/methods/public/filter.rs
```diff
@@ -283,3 +294,149 @@ mod tests {
...
+
+    #[test]
+    fn filter_shape_should_return_list_on_valid_boolean_condition() {
+        let input_shape = Shape::list(Shape::int([]), []);
+        assert_eq!(
+            get_shape(
+                vec![WithRange::new(LitExpr::Bool(true), None)],
+                input_shape.clone()
+            ),
+            Shape::list(input_shape.any_item([]), [])
```
dylan-apollo: Can we update these tests to explicitly have the expected shape instead of `.any_item([])`? So it's more clear what the expectation actually is:


```suggestion
            input_shape
```

In this case, `->filter` should not change the shape, right?

apollo-federation/src/connectors/json_selection/methods/public/find.rs
```diff
@@ -0,0 +1,445 @@
...
+            match applied_opt {
+                Some(JSON::Bool(true)) => {
+                    // Found the first matching element, return it
+                    return (Some(element.clone()), errors);
+                }
+                Some(JSON::Bool(false)) | None => {
+                    // Condition is false or errored, continue searching
+                }
+                Some(_) => {
+                    // Condition returned a non-boolean value, this is an error
```
dylan-apollo, line 67: Why would this error result in an immediate return but _other_ errors (`None` above) continue searching? We should pick one or the other, I think, and be consistent.

dylan-apollo (changes requested):

apollo-federation/src/connectors/json_selection/methods/public/find.rs
andrewmcgivery: Good point. `None` should also short circuit.

dylan-apollo (approved):

PR #7795, opened 2025-07-01
expand connector links in federation blueprint

lennyburdette:
uses `add_elements_to_schema` from SpecDefinition trait instead of our own code

---

**Checklist**

- Tests added and passing[^4]
    - [ ] Unit tests
    - [ ] Integration tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-federation/src/subgraph/typestate.rs
+++ b/apollo-federation/src/subgraph/typestate.rs
@@ -1045,6 +1045,107 @@ mod tests {
     }
 
     #[test]
+    fn injects_missing_directive_definitions_connect_v0_1() {
+        let subgraph = Subgraph::parse(
+            "S",
+            "",
+            r#"
+                extend schema @link(url: "https://specs.apollo.dev/federation/v2.10") @link(url: "https://specs.apollo.dev/connect/v0.1")
+
+                type Query {
+                    s: String
+                }"#,
+        )
+        .expect("valid schema")
+        .expand_links()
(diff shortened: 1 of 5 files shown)


apollo-federation/src/subgraph/typestate.rs
```diff
@@ -1044,6 +1044,107 @@ mod tests {
         );
     }
 
+    #[test]
+    fn injects_missing_directive_definitions_connect_v0_1() {
+        let subgraph = Subgraph::parse(
+            "S",
+            "",
+            r#"
+                extend schema @link(url: "https://specs.apollo.dev/federation/v2.10") @link(url: "https://specs.apollo.dev/connect/v0.1")
```
tninesling, line 1053: I'm surprised this doesn't return an error for an unknown federation version. I think we only have up to 2.9 defined at the moment.

Do we allow users to import connect spec elements through the federation import link, or do they have to use the connect spec directly? Either way, we may need to register versions 2.10 - 2.12 in `federation_spec_definition`, even if they don't necessarily have directive changes from 2.9.

dariuszkuc: @tninesling this was merged yesterday -> https://github.com/apollographql/router/pull/7785

tninesling: Nevermind, looks like you already added those versions. LGTM!

PR #7796, opened 2025-07-02
response_cache: Enable debug mode only when configuration is enabled and header is passed

bnjjj:
Currently, if the response cache configuration has debug set to true, the cache debugger data is sent for every request. To align this with how it is handled for connectors, we should introduce a specific header. This way, the cache debugger data in extensions will only be sent if both the configuration is enabled and the request header is correctly set.

---

**Checklist**

- Tests added and passing[^4]
    - [ ] Unit tests
    - [x] Integration tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-1352]: https://apollographql.atlassian.net/browse/ROUTER-1352?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apollo-router/src/plugins/response_cache/plugin.rs
```diff
@@ -618,6 +620,12 @@ impl CacheService {
         mut self,
         request: subgraph::Request,
     ) -> Result<subgraph::Response, BoxError> {
+        self.debug = self.debug
+            && (request
+                .supergraph_request
+                .headers()
+                .get(CACHE_DEBUG_HEADER_NAME)
+                == Some(&HeaderValue::from_static("true")));
```
SimonSapin, line 628: Nit: do we want to only accept the exact string `true` or should the presence of the header with any value be enough?

apollo-router/src/plugins/response_cache/tests.rs
```diff
@@ -267,6 +277,213 @@ async fn insert() {
...
+        .unwrap();
+
+    let service = TestHarness::builder()
+        .configuration_json(serde_json::json!({
+            "include_subgraph_errors": { "all": true },
+            "experimental_mock_subgraphs": subgraphs,
+        }))
+        .unwrap()
+        .schema(SCHEMA)
+        .extra_private_plugin(response_cache.clone())
```
SimonSapin, line 359: Does this test need to go through the trouble of manually creating the plugin data structure instead of having the normal router pipeline create it based on configuration?

apollo-router/src/plugins/response_cache/plugin.rs
bnjjj: For now I'm mainly consistent with how it works for the connectors debugger

apollo-router/src/plugins/response_cache/tests.rs
bnjjj: If we change this then it could be done as a refactor because we're using this everywhere in our tests

PR #7797, opened 2025-07-02
response_cache: add support of `cleanup_interval` configuration

bnjjj:
You can now configure the `cleanup_interval` on response caching to configure the interval at which the cleanup cron task is executed inside pg to remove expired cached data.

Example of configuration:
```yaml
experimental_response_cache:
  enabled: true
  subgraph:
    all:
      enabled: true
      postgres:
        url: "postgres://localhost:5432"
        cleanup_interval: 10mins
```

---

**Checklist**

- Tests added and passing[^4]
    - [x] Unit tests
    - [x] Integration tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-1353]: https://apollographql.atlassian.net/browse/ROUTER-1353?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apollo-router/src/plugins/response_cache/postgres.rs
```diff
@@ -72,6 +73,11 @@ pub(crate) struct PostgresCacheConfig {
     /// Useful when running tests in parallel to avoid conflicts
     #[serde(default)]
     pub(crate) namespace: Option<String>,
+
+    #[serde(deserialize_with = "humantime_serde::deserialize", default)]
+    #[schemars(with = "Option<String>", default)]
+    /// Specifies the interval between cache cleanup operations (e.g., "2 hours", "30mins"). Default: 30mins
+    pub(crate) cleanup_interval: Option<Duration>,
```
carodewig: Won't this default to `None`? Where is the 30min coming from?

I assume we'll always want some cleanup interval, so would it be easier to have this be a `Duration` and use the serde default fn to specify 30 minutes?

carodewig: Also `humantime` doesn't seem to support `mins`, just `min`
```suggestion
    #[serde(deserialize_with = "humantime_serde::deserialize", default)]
    #[schemars(with = "Option<String>", default)]
    /// Specifies the interval between cache cleanup operations (e.g., "2 hours", "30min"). Default: 30min
    pub(crate) cleanup_interval: Option<Duration>,
```

apollo-router/src/plugins/response_cache/postgres.rs
```diff
@@ -115,6 +122,10 @@ pub(crate) enum PostgresCacheStorageError {
     BadConfiguration(String),
     #[error("postgres error: {0}")]
     PgError(#[from] sqlx::Error),
+    #[error("cleanup_interval configuration is out of range: {0}")]
+    OutOfRangeError(#[from] chrono::OutOfRangeError),
+    #[error("cleanup_interval configuration is incorrect: {0}")]
+    WrongCleanupInterval(String),
```
carodewig: Nit: this is very minor semantics, but I think incorrect/wrong imply that there's one 'correct' interval rather than a huge range of options

```suggestion
    #[error("cleanup_interval configuration is invalid: {0}")]
    InvalidCleanupInterval(String),
```

apollo-router/src/plugins/response_cache/postgres.rs
```diff
@@ -418,4 +429,246 @@ impl PostgresCacheStorage {
...
     }
+
+    pub(crate) async fn update_cron(&self) -> anyhow::Result<()> {
+        if let Some(cleanup_interval) = &self.cleanup_interval {
+            let cron = Cron::try_from(cleanup_interval)
+                .map_err(PostgresCacheStorageError::WrongCleanupInterval)?;
+            sqlx::query!("SELECT cron.alter_job((SELECT jobid FROM cron.job WHERE jobname = 'delete-old-cache-entries'), $1)", &cron.0)
+                .execute(&self.pg_pool)
+                .await?;
+        }
```
carodewig: Would it be helpful to add some trace-level 'success' logging here for the sake of debugging later?

```suggestion
            log::trace!("Configured `delete-old-cache-entries` cron to have interval = `{}`", &cron.0);
        }
```

apollo-router/src/plugins/response_cache/postgres.rs
bnjjj: The default is coming from the migration itself. But actually it's set to 1h by default so I was wrong :p

apollo-router/src/plugins/response_cache/postgres.rs
```diff
@@ -72,6 +73,11 @@ pub(crate) struct PostgresCacheConfig {
     /// Useful when running tests in parallel to avoid conflicts
     #[serde(default)]
     pub(crate) namespace: Option<String>,
+
+    #[serde(deserialize_with = "humantime_serde::deserialize", default)]
+    #[schemars(with = "Option<String>", default)]
+    /// Specifies the interval between cache cleanup operations (e.g., "2 hours", "30min"). Default: 1 hour
+    pub(crate) cleanup_interval: Option<Duration>,
```
SimonSapin: Why is this an option? I think we shouldn’t use `Option` unless `None` means something different than any `Some(…)` value, such as disabling cleanup entirely.

apollo-router/src/plugins/response_cache/postgres.rs
```diff
@@ -421,4 +432,167 @@ impl PostgresCacheStorage {
 
         Ok(rec.count.unwrap_or_default() as u64)
     }
+
+    pub(crate) async fn update_cron(&self) -> anyhow::Result<()> {
+        if let Some(cleanup_interval) = &self.cleanup_interval {
```
SimonSapin: Since there is no `else` clause here, rather than one hour it looks like the default is "whatever previous configuration the database already has" which doesn’t sound desirable. I think making this config non-optional is the way to go.

apollo-router/src/plugins/response_cache/postgres.rs
bnjjj: The main reason I'm using this option here and yes it's not the same pattern than usually because here the default value is directly set in ou sqlx migration files. So in order to avoid executing a sql query every time the router starts if it's the default value I chose to use an Option

apollo-router/src/plugins/response_cache/postgres.rs
bnjjj: Ah yes ! You're right I didn't think about that ! I'll fix it

SimonSapin (approved):

PR #7800, opened 2025-07-02
Reintegrate xtask into workspace

goto-bus-stop:
In the past xtask was separated into its own root crate due to a dependency compatibility collision or so:

> "Note that xtask is not in the workspace member because it relies on dependencies that are incompatible with the router."

This is not true anymore, and there are some downsides to the separation:
- Separate Cargo.lock files and target directories increase disk space usage and compilation time
  - This means we need special treatment to make xtask build times acceptable on CI
- Linting, formatting, testing, etc do not apply to xtask, so it's possible for CI to accept things that don't work
- Sometimes renovate gets upset (eg: https://github.com/apollographql/router/pull/7799#issuecomment-3028212046)

This PR makes xtask a normal workspace member again.

Fixes https://github.com/apollographql/router/issues/7769

---

**Checklist**

- <strike>Changeset is included for user-facing changes</strike>
- <strike>Documentation[^2] completed</strike>
- Tests added and passing[^4]
    - <strike>Unit tests</strike>
    - <strike>Integration tests</strike>

**Exceptions**

*Note any exceptions here*

**Notes**

This might save some compile time(?) because dependencies can be shared,
but we'll see.This might save some compile time(?) because dependencies can be shared,
but we'll see.This might save some compile time(?) because dependencies can be shared,
but we'll see.This might save some compile time(?) because dependencies can be shared,
but we'll see.


dylan-apollo:
Can we update xtask stuff to use the same deps (e.g., hyper) as router? Otherwise this makes it hard to tell when we accidentally have duplicate deps in router (which hopefully we can avoid most of the time).

goto-bus-stop:
@dylan-apollo I was going to follow up with that later (also renovate might do some of it for us after this change)

lrlna (approved):

glasser:
Nice, thanks!

PR #7801, opened 2025-07-02
Fix error when removing non-UTF-8 headers with Rhai plugin

Velfi:
When trying to remove non-UTF-8 headers from a Rhai plugin, users were faced with an unhelpful error. Now, non-UTF-8 values will be lossy converted to UTF-8 when accessed from Rhai. This change affects `get`, `get_all`, and `remove` operations.

---

**Checklist**

- Tests added and passing[^4]
    - [x] Unit tests
    - [x] Integration tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-1291]: https://apollographql.atlassian.net/browse/ROUTER-1291?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changesets/fix_fix_removing_non_utf_8_headers_with_rhai.md
+++ b/.changesets/fix_fix_removing_non_utf_8_headers_with_rhai.md
@@ -1,0 +1,7 @@
+### Fix error when removing non-UTF-8 headers with Rhai plugin ([PR #7801](https://github.com/apollographql/router/pull/7801))
+
+When trying to remove non-UTF-8 headers from a Rhai plugin, users were faced with an unhelpful error. Now, non-UTF-8 values will be lossy converted to UTF-8 when accessed from Rhai. This change affects `get`, `get_all`, and `remove` operations.
+
+[ROUTER-1291]: https://apollographql.atlassian.net/browse/ROUTER-1291?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
+
+By [@Velfi](https://github.com/Velfi) in https://github.com/apollographql/router/pull/7801
--- a/.changesets/fix_fix_removing_non_utf_8_headers_with_rhai.md
+++ b/.changesets/fix_fix_removing_non_utf_8_headers_with_rhai.md
@@ -2,6 +2,4 @@
 
 When trying to remove non-UTF-8 headers from a Rhai plugin, users were faced with an unhelpful error. Now, non-UTF-8 values will be lossy converted to UTF-8 when accessed from Rhai. This change affects `get`, `get_all`, and `remove` operations.
 
-[ROUTER-1291]: https://apollographql.atlassian.net/browse/ROUTER-1291?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
-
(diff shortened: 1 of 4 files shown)


.changesets/fix_fix_removing_non_utf_8_headers_with_rhai.md
```diff
@@ -0,0 +1,7 @@
+### Fix error when removing non-UTF-8 headers with Rhai plugin ([PR #7801](https://github.com/apollographql/router/pull/7801))
+
+When trying to remove non-UTF-8 headers from a Rhai plugin, users were faced with an unhelpful error. Now, non-UTF-8 values will be lossy converted to UTF-8 when accessed from Rhai. This change affects `get`, `get_all`, and `remove` operations.
+
+[ROUTER-1291]: https://apollographql.atlassian.net/browse/ROUTER-1291?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
+
```
goto-bus-stop:
```suggestion
```

apollo-router/src/plugins/rhai/tests.rs
```diff
@@ -848,3 +848,64 @@ async fn it_can_access_demand_control_context() -> Result<(), BoxError> {
 
     Ok(())
 }
+
+#[tokio::test]
+async fn test_rhai_header_removal_with_non_utf8_header() -> Result<(), BoxError> {
+    let bytes = b"\x80";
+    // Prove that the bytes are not valid UTF-8
+    assert!(String::from_utf8(bytes.to_vec()).is_err());
```
goto-bus-stop, line 856:
```suggestion
    assert!(str::from_utf8(bytes).is_err());
```
nitty nit

apollo-router/src/plugins/rhai/tests.rs
```diff
@@ -848,3 +848,64 @@ async fn it_can_access_demand_control_context() -> Result<(), BoxError> {
...
+    // Removing a non-UTF-8 header should be OK
+    let body = service_response.next_response().await.unwrap();
+    if body.errors.is_empty() {
+        // yay, no errors
+    } else {
+        let rhai_error = body
+            .errors
+            .iter()
+            .find(|e| e.message.contains("rhai execution error"))
+            .expect("must have a rhai error");
```
goto-bus-stop:
```suggestion
            .expect("unexpected non-rhai error");
```
 having a rhai error is also a failure, just a different kind, so i think it may be confusing if the test that asserts there's no errors fails with "must have a rhai error"

apollo-router/src/plugins/rhai/tests.rs
Velfi: If I use that then I get:
```
❯ cargo check --tests
    Blocking waiting for file lock on build directory
    Checking apollo-router v2.4.0 (/Users/zeldahessler/Documents/router/apollo-router)
warning: calls to `str::from_utf8` with an invalid literal always return an error
   --> apollo-router/src/plugins/rhai/tests.rs:857:13
    |
855 |     let bytes = b"\x80";
    |                 ------- the literal was valid UTF-8 up to the 0 bytes
856 |     // Prove that the bytes are not valid UTF-8
857 |     assert!(str::from_utf8(bytes).is_err());
    |             ^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `#[warn(invalid_from_utf8)]` on by default
```

and the `allow` attribute doesn't work on `assert!`. Therefore, I'm going to revert this.

goto-bus-stop (approved):

PR #7803, opened 2025-07-02
Implementation of merge_field 

clenfest:
Implemenation of merge_field(). . Note that there are no tests yet since several dependent functions are not yet implemented.

### DIFF
--- a/apollo-federation/src/merger/merger.rs
+++ b/apollo-federation/src/merger/merger.rs
@@ -1,5 +1,6 @@
 use std::collections::HashMap;
 use std::collections::HashSet;
+use std::rc::Rc;
 use std::sync::LazyLock;
 
 use apollo_compiler::Name;
@@ -15,6 +16,7 @@ use apollo_compiler::schema::Type;
 use apollo_compiler::validation::Valid;
 use itertools::Itertools;
 
+use crate::LinkSpecDefinition;
 use crate::bail;
 use crate::error::CompositionError;
 use crate::error::FederationError;
@@ -80,9 +82,9 @@ pub(crate) struct MergeResult {
 }
 
 pub(in crate::merger) struct MergedDirectiveInfo {
-    definition: DirectiveDefinition,
-    arguments_merger: Option<ArgumentMerger>,
-    static_argument_transform: Option<Box<StaticArgumentsTransform>>,
+    pub(in crate::merger) definition: DirectiveDefinition,
+    pub(in crate::merger) arguments_merger: Option<ArgumentMerger>,
+    pub(in crate::merger) static_argument_transform: Option<Rc<StaticArgumentsTransform>>,
 }
 
 #[derive(Debug, Default)]
@@ -109,8 +111,10 @@ pub(crate) struct Merger {
     pub(in crate::merger) fields_with_override: DirectiveReferencers,
     pub(in crate::merger) inaccessible_directive_name_in_supergraph: Option<Name>,
     pub(in crate::merger) schema_to_import_to_feature_url: HashMap<String, HashMap<String, Url>>,
+    pub(in crate::merger) link_spec_definition: &'static LinkSpecDefinition,
     pub(in crate::merger) join_directive_identities: HashSet<Identity>,
     pub(in crate::merger) join_spec_definition: &'static JoinSpecDefinition,
+    pub(in crate::merger) latest_federation_version_used: Version,
 }
 
 #[allow(unused)]
@@ -131,7 +135,14 @@ impl Merger {
                 latest_federation_version_used
             )
         };
-        let link_spec = LINK_VERSIONS.get_minimum_required_version(latest_federation_version_used);
+        let Some(link_spec_definition) =
+            LINK_VERSIONS.get_minimum_required_version(latest_federation_version_used)
+        else {
+            bail!(
+                "No link spec version found for federation version {}",
+                latest_federation_version_used
+            )
+        };
         let fields_with_from_context = Self::get_fields_with_from_context_directive(&subgraphs);
         let fields_with_override = Self::get_fields_with_override_directive(&subgraphs);
 
@@ -164,9 +175,11 @@ impl Merger {
             fields_with_from_context,
             fields_with_override,
             schema_to_import_to_feature_url,
+            link_spec_definition,
             join_directive_identities,
             inaccessible_directive_name_in_supergraph: todo!(),
             join_spec_definition: join_spec,
+            latest_federation_version_used: latest_federation_version_used.clone(),
         })
     }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6372,9 +6372,9 @@ dependencies = [
 
 [[package]]
 name = "serde_json"
-version = "1.0.140"
+version = "1.0.141"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "20068b6e96dc6c9bd23e01df8827e6c7e1f2fddd43c21810382803c136b99373"
+checksum = "30b9eff21ebe718216c6ec64e1d9ac57087aad11efc64e32002bce4a0d4c03d3"
 dependencies = [
  "indexmap 2.10.0",
  "itoa",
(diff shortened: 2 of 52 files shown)


apollo-federation/src/merger/merge_field.rs
```diff
@@ -0,0 +1,423 @@
...
+            self.error_reporter.add_error(error);
+
+            return Ok(());
+        }
+
+        let without_external = self.validate_and_filter_external(sources);
+        self.merge_description(&without_external, dest);
+        self.record_applied_directives_to_merge(&without_external, dest);
+        self.add_arguments_shallow(&without_external, dest);
+        // todo!("iterate over arguments of destination and merge the arguments");
```
tninesling: Looks like JS has a section here to deep merge the arguments 
```
for (const destArg of dest.arguments()) {
  const subgraphArgs = mapSources(withoutExternal, f => f?.argument(destArg.name));
  this.mergeArgument(subgraphArgs, destArg);
}
```
Is that happening in a different PR, or should we add it in this one?

Also, JS has a very descriptive comment here which might be good to carry over

apollo-federation/src/merger/merge_field.rs
```diff
@@ -0,0 +1,423 @@
...
+        // Check each directive for violations
+        for directive in &field_def.directives {
+            if self.is_merged_directive(&self.names[source_idx], directive) {
+                // Contrarily to most of the errors during merging that "merge" errors for related elements, we're logging one
+                // error for every application here. But this is because the error is somewhat subgraph specific and is
+                // unlikely to span multiple subgraphs. In fact, we could almost have thrown this error during subgraph validation
+                // if this wasn't for the fact that it is only thrown for directives being merged and so is more logical to
+                // be thrown only when merging.
+
+                let error = CompositionError::SubgraphError {
```
tninesling: I'm not sure how feasible this is, but I think in an ideal world we remove this `SubgraphError` variant and instead add the subgraph name in the message itself. It looks like we should be able to create it the same way as JS here since we already have the subgraph name available.

apollo-federation/src/merger/merge_field.rs
```diff
@@ -0,0 +1,423 @@
...
+            if !self.is_field_external(*source_idx, source_field_pos) {
+                continue;
+            }
+
+            let source_field =
+                source_field_pos.get(self.subgraphs[*source_idx].validated_schema().schema())?;
+            let source_args = source_field.arguments.to_vec();
+
+            if dest_field_ty != source_field.ty
+                && (all_types_equal || !Self::is_strict_subtype(&dest_field_ty, &source_field.ty))
```
tninesling: I don't think this is equivalent to the JS version since only the right side distributed the negation
```
if (!(sameType(dest.type!, source.type!) || (!allTypesEqual && this.isStrictSubtype(dest.type!, source.type!))))
```

apollo-federation/src/merger/merge_field.rs
```diff
@@ -0,0 +1,423 @@
...
+                        ),
+                    },
+                    dest,
+                    sources,
+                    |source, _| Some(format!("type \"{}\"", source)),
+                );
+                continue;
+            }
+
+            // 2. Validate argument presence and types
```
tninesling: Did I miss a `1.`, or is this comment stale?

apollo-federation/src/merger/merge_field.rs
```diff
@@ -0,0 +1,423 @@
...
+                            // Build description for interface object abstraction
+                            let interface_types: Vec<String> = itf_object_fields
+                                .iter()
+                                .map(|f| f.type_name().to_string())
+                                .collect();
+
+                            Some(format!(
+                                "{} (through @interfaceObject {})",
+                                self.names[*i],
+                                interface_types.join(", ")
```
duckki: We have `human_readable_list` - a port of `printHumanReadableList`. `printTypes` can be ported with it.

apollo-federation/src/merger/merge_field.rs
```diff
@@ -0,0 +1,423 @@
...
+                ),
+            };
+
+            self.error_reporter.add_error(error);
+
+            return Ok(());
+        }
+
+        let without_external = self.validate_and_filter_external(sources);
+        self.merge_description(&without_external, dest);
```
duckki, line 87: I think we should port [comments](https://github.com/apollographql/federation/blob/4653320016ed4202a229d9ab5933ad3f13e5b6c0/composition-js/src/merging/merge.ts#L1734), too.

apollo-federation/src/merger/merge_field.rs
duckki: +1

apollo-federation/src/merger/merge_field.rs
```diff
@@ -0,0 +1,423 @@
...
+            } else {
+                sources
+            };
+
+        let all_types_equal = self.merge_type_reference(sources_for_type_merge, dest, false);
+        if self.has_external(sources) {
+            self.validate_external_fields(sources, dest, all_types_equal)?;
+        }
+        self.add_join_field(sources, dest);
+        self.add_join_directive_directives(sources, dest);
```
duckki, line 151: JS version passes `allTypesEqual, mergeContext`, too. Is that also in the todo list?

apollo-federation/src/merger/merge_field.rs
duckki: I agree this is decided to be a merging error (composition error) than a subgraph error, according to the comment. So, it makes sense to be a variant of CompositionError.

apollo-federation/src/merger/merge_field.rs
duckki: +1

apollo-federation/src/merger/merge_field.rs
duckki: +1 :)
Also, please port [the comment](https://github.com/apollographql/federation/blob/4653320016ed4202a229d9ab5933ad3f13e5b6c0/composition-js/src/merging/merge.ts#L1864).

apollo-federation/src/merger/merge_field.rs
```diff
@@ -0,0 +1,423 @@
...
+
+            // 2. Validate argument presence and types
+            for dest_arg in &dest_args {
+                // Check if external field has this argument
+                let Some(source_arg) = source_args.iter().find(|arg| arg.name == dest_arg.name)
+                else {
+                    self.report_mismatch_error_with_specifics(
+                        CompositionError::ExternalArgumentMissing {
+                            message: format!(
+                                "External field \"{}\" is missing argument \"{}\" from supergraph",
```
duckki: The error message is different than [JS version](https://github.com/apollographql/federation/blob/4653320016ed4202a229d9ab5933ad3f13e5b6c0/composition-js/src/merging/merge.ts#L1896).

duckki: BTW, having one loop here means, the order of error messages will be different. I think JS version wanted to group the same kind of errors together.

apollo-federation/src/merger/merge_field.rs
```diff
@@ -0,0 +1,423 @@
...
+                            ),
+                        },
+                        &argument_pos,
+                        &self.argument_sources(sources, &dest_arg.name)?,
+                        |source, _| Some(format!("type \"{}\"", source)),
+                    );
+                    continue;
+                }
+
+                if dest_arg.default_value != source_arg.default_value {
```
duckki, line 377: I think we need to port `valueEquals`, since that is different from `Eq`. Surprisingly, I couldn't find an existing port. I think we missed this in several different places. I'll file a follow-up Jira for this.

apollo-federation/src/merger/merge_field.rs
clenfest: Yeah, I'm just establishing the call site, that can get fixed when it gets implemented.

apollo-federation/src/merger/merger.rs
```diff
@@ -880,6 +880,29 @@ impl Merger {
         };
         self.error_reporter.add_hint(hint);
     }
+
+    /// Map over sources, applying a function to each element
+    pub(in crate::merger) fn map_sources<T, U, F>(&self, sources: &Sources<T>, f: F) -> Sources<U>
```
duckki: nit: Since `self` is not used here, it could be moved out of `Merger`, or be a method on `Sources`.

duckki (approved):
Looks good!
I have a nit. But, that's not a big deal.

PR #7807, opened 2025-07-03
fix(federation/composition): fixed `validate_federation_imports` to use the actual federation spec

duckki:
This PR fixes a bug where Rust composition does not accept newer federation directives like `@authenticated`.

* Fixed validate_federation_imports to use the actual federation spec, instead of hardcoded names.
* Also fixed subgraph tests to auto-import all latest federation directives and not to use supergraph spec links.
  * `into_fed2_test_subgraph` was updated to have a `bool` argument that allow auto-importing all federation directives into test subgraph schemas.
  * With that, subgraph tests don't need to have spec `@links` any more.

---

**Checklist**

- Tests added and passing[^4]
    - [x] Unit tests
    - [ ] Integration tests

### DIFF
--- a/apollo-federation/src/link/database.rs
+++ b/apollo-federation/src/link/database.rs
@@ -9,19 +9,44 @@ use apollo_compiler::collections::IndexMap;
 use apollo_compiler::schema::DirectiveDefinition;
 use apollo_compiler::ty;
 
+use crate::SpecDefinition;
 use crate::link::DEFAULT_LINK_NAME;
 use crate::link::Link;
 use crate::link::LinkError;
 use crate::link::LinksMetadata;
+use crate::link::federation_spec_definition::FED_1;
+use crate::link::federation_spec_definition::FEDERATION_VERSIONS;
+use crate::link::federation_spec_definition::FederationSpecDefinition;
 use crate::link::federation_spec_definition::fed1_link_imports;
 use crate::link::spec::Identity;
 use crate::link::spec::Url;
-use crate::subgraph::spec::FEDERATION_V2_DIRECTIVE_NAMES;
-use crate::subgraph::spec::FEDERATION_V2_ELEMENT_NAMES;
+use crate::link::spec::Version;
+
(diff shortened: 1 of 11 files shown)


apollo-federation/src/merger/merge_enum.rs
```diff
@@ -295,14 +294,31 @@ pub(crate) mod tests {
...
+            enum link__Purpose {
+                SECURITY
+                EXECUTION
+            }
             "#,
                 "",
             )
             .build()?;
-        let mut schema = expand_schema(schema)?;
+        let mut schema = FederationSchema::new(schema)?;
```
duckki, line 321: `expand_schema` is not meant to work on supergraph schemas. I just added missing definitions in the template schema, instead of calling `expand_schema`.

apollo-federation/src/link/database.rs
```diff
@@ -9,27 +9,52 @@ use apollo_compiler::collections::IndexMap;
...
+use crate::link::federation_spec_definition::FEDERATION_VERSIONS;
+use crate::link::federation_spec_definition::FederationSpecDefinition;
 use crate::link::federation_spec_definition::fed1_link_imports;
 use crate::link::spec::Identity;
 use crate::link::spec::Url;
-use crate::subgraph::spec::FEDERATION_V2_DIRECTIVE_NAMES;
-use crate::subgraph::spec::FEDERATION_V2_ELEMENT_NAMES;
+use crate::link::spec::Version;
+
+fn find_federation_spec_for_version<'a>(version: &Version) -> Option<&'a FederationSpecDefinition> {
```
dariuszkuc, line 25: are we still using `link/database.rs` stuff? this was created almost two years ago when we kicked off the federation rewrite project when `apollo-rs` was forcing some of those patterns. AFAIK a lot of it should have been rewritten - it might be safer to just drop this file altogether

duckki: Yeah, I agree that we should review the code.

tninesling (approved):

PR #7817, opened 2025-07-03
chore: use multer to test subscription responses

goto-bus-stop:
Follow-up to #7731.

Use the multer crate for parsing subscription `multipart/mixed` streams
instead of a hand-rolled implementation. multer technically is for
`multipart/form-data`, which is sort of the successor format to
`multipart/mixed` if i understand the RFCs correctly. But the format is
essentially identical.

This implementation is mostly stricter than before. These things now
cause test failures:
- Final `--graphql--` boundaries in the middle of the stream
- Omitting the final `--graphql--` boundary
- Not draining the entire subscription
- Non-JSON response chunks

But these benign things will not:
- Having multiple headers on a chunk (we don't do this, but it sounds
  valid)
- A chunk boundary occurring at the edge of an HTTP frame from hyper

### DIFF
--- a/apollo-router/tests/integration/subscriptions/mod.rs
+++ b/apollo-router/tests/integration/subscriptions/mod.rs
@@ -219,7 +219,7 @@ pub async fn verify_subscription_events(
         );
     });
 
-    _ = timeout.await.expect("Subscription test timed out");
+    timeout.await.expect("Subscription test timed out");
 
     assert!(
         subscription_events.len() == expected_events.len(),
--- a/apollo-router/tests/integration/subscriptions/mod.rs
+++ b/apollo-router/tests/integration/subscriptions/mod.rs
@@ -184,7 +184,7 @@ pub async fn verify_subscription_events(
     stream: impl futures::Stream<Item = Result<bytes::Bytes, reqwest::Error>> + Send,
     expected_events: Vec<serde_json::Value>,
     include_heartbeats: bool,
-) -> Result<Vec<serde_json::Value>, String> {
+) -> Vec<serde_json::Value> {
     use pretty_assertions::assert_eq;
 
     // Use `multipart/form-data` parsing. The router actually responds with `multipart/mixed`, but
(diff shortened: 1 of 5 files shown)


apollo-router/tests/integration/subscriptions/mod.rs
```diff
@@ -177,42 +174,40 @@ pub async fn start_coprocessor_server() -> wiremock::MockServer {
...
     // Set a longer timeout for receiving all events
     let timeout = tokio::time::timeout(tokio::time::Duration::from_secs(60), async {
-        let mut chunk_string = String::new();
-        while let Some(chunk) = stream.next().await {
-            let chunk = chunk.map_err(|e| format!("Stream error: {}", e))?;
-            chunk_string += &String::from_utf8_lossy(&chunk);
+        while let Some(field) = multipart
+            .next_field()
+            .await
+            .expect("could not read next chunk")
```
goto-bus-stop, line 200: I should actually change all these expects to error returns.

goto-bus-stop: Or maybe not, but at least be consistent

apollo-router/tests/integration/subscriptions/ws_passthrough.rs
```diff
@@ -653,16 +643,7 @@ async fn test_subscription_ws_passthrough_on_config_reload() -> Result<(), BoxEr
     assert_eq!(total_active, 1);
     assert_eq!(total_active + total_terminating, 1);
 
-    match verify_subscription_events(stream, expected_events, true)
-        .await
-        .map_err(|e| format!("Event verification failed: {}", e))
-    {
-        Ok(_events) => {}
-        Err(err) if err.contains("unexpected EOF") => {
```
goto-bus-stop, line 661: @carodewig I wasn't able to produce these errors. I think ideally they wouldn't occur, but I assume you put this in because they did 😅  Do you still see them when you run tests?

carodewig: I think you can get rid of that!

I added that logic to explore what happens when the subscription server abruptly shuts down rather than cleanly closing the connection. That's how I was initially exploring the critical/non-critical error behavior.

goto-bus-stop: great, that makes my life easier 😁

Velfi (approved):

goto-bus-stop:
@mergifyio backport dev

PR #7833, opened 2025-07-07
Revert "Revert "Add _dd.measured to spans when agent sampling is enabled""

BrynCooke:
Reverts apollographql/router#7403

Reopening this PR so that we can see if there are conflicts while we get clarification from datadog.

The current situation where the APM view is not functional is unsustainable and adding yet another config to allow users to set measured on a per span basis is complicated. 

Given that the behaviour in Router 1.x is that all spans are measured this feels like a regression rather than a nice to have.

### DIFF
--- a/apollo-router/src/plugins/telemetry/tracing/datadog/span_processor.rs
+++ b/apollo-router/src/plugins/telemetry/tracing/datadog/span_processor.rs
@@ -1,4 +1,5 @@
 use opentelemetry::Context;
+use opentelemetry::KeyValue;
 use opentelemetry::trace::SpanContext;
 use opentelemetry::trace::TraceResult;
 use opentelemetry_sdk::Resource;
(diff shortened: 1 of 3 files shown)


apollo-router/src/plugins/telemetry/tracing/datadog/span_processor.rs
```diff
@@ -36,6 +37,7 @@ impl<T: SpanProcessor> SpanProcessor for DatadogSpanProcessor<T> {
             span.span_context.is_remote(),
             span.span_context.trace_state().clone(),
         );
+        span.attributes.push(KeyValue::new("_dd.measured", 1));
```
goto-bus-stop, line 40: This seems to happen for both the datadog exporter, and the OTLP exporter. The datadog exporter already configures measuring separately, so is it correct to do this here? I think it is duplicate for the DD exporter and could cause confusing interference in the future, and should probably only happen in OTLP.

abernix:
Marking this as a draft until it can get to the top of the stack.

Issue #7702, opened 2025-06-17
Support CORS Access-Control-Request-Private-Network capability


theJC:
## Describe the solution you'd like
Improve router CORS feature to support `Access-Control-Request-Private-Network` capability.  
This was added to the tower CORS support in version [0.4.2](https://github.com/tower-rs/tower-http/releases/tag/tower-http-0.4.2) back in 2023.
https://docs.rs/tower-http/latest/tower_http/cors/struct.AllowPrivateNetwork.html

Chrome's Private Network Access (PNA) for non-secure contexts deprecation trial is ending soon.  see: https://developer.chrome.com/blog/pna-permission-prompt-ot-end

When complete:
> all private network requests will only work if the endpoint opts in by responding to a preflight request.

## Describe alternatives you've considered
There is none.

## Additional context
This change will allow additional options for developer experience when running Router locally.  For local development serving a /graphql endpoint, and using a website that allows to point to make requests against that local graphql endpoint - either the customers, or simply Apollo explorer or router -- allowing your customers using this this will allow hitting a local endpoint without encountering CORS issues.

<img width="1695" alt="Image" src="https://github.com/user-attachments/assets/456b7ac8-9edc-4c18-ac0c-5621ebfac18e" />

https://developer.chrome.com/blog/private-network-access-update?utm_source=devtools&utm_campaign=stable

theJC:
Addressed by https://github.com/apollographql/router/pull/8279

PR #7853, opened 2025-07-08
add trusted_origins support for selective credential handling

Velfi:
---

**Checklist**

- Tests added and passing[^4]
    - [ ] Unit tests
    - [ ] Integration tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/configuration/cors.rs
+++ b/apollo-router/src/configuration/cors.rs
@@ -53,6 +53,13 @@ pub(crate) struct Cors {
     /// Note that `origins` will be evaluated before `match_origins`
     pub(crate) match_origins: Option<Vec<String>>,
 
+    /// List of trusted origins that are allowed to send credentials.
+    /// When specified, CORS will be enabled for all origins, but only these
+    /// trusted origins will receive `Access-Control-Allow-Credentials: true`.
+    /// All other origins will receive `Access-Control-Allow-Origin: *` without credentials.
+    /// Defaults to an empty list.
+    pub(crate) trusted_origins: Option<Vec<String>>,
+
     /// Allowed request methods. Defaults to GET, POST, OPTIONS.
     pub(crate) methods: Vec<String>,
 
@@ -86,12 +93,14 @@ impl Cors {
         expose_headers: Option<Vec<String>>,
         origins: Option<Vec<String>>,
         match_origins: Option<Vec<String>>,
+        trusted_origins: Option<Vec<String>>,
         methods: Option<Vec<String>>,
         max_age: Option<Duration>,
     ) -> Self {
         Self {
             expose_headers,
             match_origins,
+            trusted_origins,
             max_age,
             origins: origins.unwrap_or_else(default_origins),
             methods: methods.unwrap_or_else(default_cors_methods),
@@ -103,9 +112,19 @@ impl Cors {
 }
 
 impl Cors {
+    /// Check if trusted origins mode is enabled
+    pub(crate) fn has_trusted_origins(&self) -> bool {
+        self.trusted_origins.is_some() && !self.trusted_origins.as_ref().unwrap().is_empty()
+    }
+
     pub(crate) fn into_layer(self) -> Result<CorsLayer, String> {
         // Ensure configuration is valid before creating CorsLayer
         self.ensure_usable_cors_rules()?;
+
+        // If trusted_origins is specified, we need custom handling
+        if self.has_trusted_origins() {
+            return Err("Trusted origins mode requires custom CORS handling".to_string());
+        }
 
         let allow_headers = if self.allow_headers.is_empty() {
             cors::AllowHeaders::mirror_request()
@@ -168,7 +187,28 @@ impl Cors {
                 "Invalid CORS configuration: use `allow_any_origin: true` to set `Access-Control-Allow-Origin: *`",
             );
         }
-        if self.allow_credentials {
+
+        // When trusted_origins is specified, we handle credentials dynamically
+        // so we can be more permissive with the validation
+        if self.has_trusted_origins() {
+            // With trusted origins, we can't use allow_any_origin since we need to
+            // differentiate between trusted and untrusted origins
+            if self.allow_any_origin {
+                return Err(
+                    "Invalid CORS configuration: Cannot combine `trusted_origins` with `allow_any_origin: true`",
+                );
+            }
+
+            // Validate trusted origins list
+            if let Some(trusted_origins) = &self.trusted_origins {
+                if trusted_origins.iter().any(|x| x == "*") {
+                    return Err(
+                        "Invalid CORS configuration: Cannot use '*' in `trusted_origins` list",
+                    );
+                }
+            }
+        } else if self.allow_credentials {
+            // Standard CORS validation when trusted_origins is not used
             if self.allow_headers.iter().any(|x| x == "*") {
                 return Err(
                     "Invalid CORS configuration: Cannot combine `Access-Control-Allow-Credentials: true` \
@@ -296,5 +336,45 @@ mod tests {
             .build();
         let layer = cors.into_layer();
         assert!(layer.is_ok());
+    }
+
+    #[test]
+    fn test_trusted_origins_requires_custom_handling() {
+        let cors = Cors::builder()
+            .trusted_origins(vec![String::from("https://trusted.com")])
+            .build();
+        let layer = cors.into_layer();
+        assert!(layer.is_err());
+        assert_eq!(
+            layer.unwrap_err(),
+            "Trusted origins mode requires custom CORS handling"
+        );
+    }
+
+    #[test]
+    fn test_trusted_origins_cannot_use_wildcard() {
+        let cors = Cors::builder()
+            .trusted_origins(vec![String::from("*")])
+            .build();
+        let layer = cors.into_layer();
+        assert!(layer.is_err());
+        assert_eq!(
+            layer.unwrap_err(),
+            "Invalid CORS configuration: Cannot use '*' in `trusted_origins` list"
+        );
+    }
+
+    #[test]
+    fn test_trusted_origins_cannot_use_allow_any_origin() {
+        let cors = Cors::builder()
+            .trusted_origins(vec![String::from("https://trusted.com")])
+            .allow_any_origin(true)
+            .build();
+        let layer = cors.into_layer();
+        assert!(layer.is_err());
+        assert_eq!(
+            layer.unwrap_err(),
+            "Invalid CORS configuration: Cannot combine `trusted_origins` with `allow_any_origin: true`"
+        );
     }
 }
--- a/apollo-router/src/plugins/cors.rs
+++ b/apollo-router/src/plugins/cors.rs
@@ -1,0 +1,702 @@
+//! CORS (Cross-Origin Resource Sharing) plugin for Apollo Router
+//!
+//! This plugin adds CORS headers to HTTP responses and handles preflight OPTIONS requests.
+//! It operates at the router service level to ensure CORS is applied to all requests.
+
+use std::future::Future;
+use std::pin::Pin;
+use std::sync::Arc;
+use std::task::Context;
+use std::task::Poll;
+
+use http::HeaderMap;
+use http::HeaderValue;
+use http::Method;
+use http::StatusCode;
+use tokio::sync::Mutex;
+use tower::BoxError;
+use tower::Layer;
+use tower::Service;
+use tower::ServiceExt;
+use tower_http::cors::CorsLayer;
+
+use crate::configuration::cors::Cors;
+use crate::plugin::Plugin;
+use crate::plugin::PluginInit;
+use crate::register_plugin;
+use crate::services::router;
+
+/// CORS plugin
+#[derive(Debug, Clone)]
+pub(crate) struct CorsPlugin {
+    config: Cors,
+}
+
+#[async_trait::async_trait]
+impl Plugin for CorsPlugin {
+    type Config = Cors;
+
+    async fn new(init: PluginInit<Self::Config>) -> Result<Self, BoxError> {
+        Ok(CorsPlugin {
+            config: init.config,
+        })
+    }
+
+    fn router_service(&self, service: router::BoxService) -> router::BoxService {
+        if self.config.has_trusted_origins() {
+            // Use custom CORS handling for trusted origins
+            let trusted_cors_service = TrustedOriginsCorsService::new(self.config.clone(), service);
+            ServiceExt::boxed(trusted_cors_service)
+        } else {
+            // Use standard tower-http CORS layer
+            match self.config.clone().into_layer() {
+                Ok(cors_layer) => {
+                    let cors_service = CorsService::new(cors_layer, service);
+                    ServiceExt::boxed(cors_service)
+                }
+                Err(err) => {
+                    tracing::error!("Failed to create CORS layer: {}", err);
+                    service
+                }
+            }
+        }
+    }
+}
+
+/// Service that applies CORS to router requests by converting types
+#[derive(Clone)]
+struct CorsService<S> {
+    cors_layer: CorsLayer,
+    inner: Arc<Mutex<S>>,
+}
+
+impl<S> CorsService<S> {
+    fn new(cors_layer: CorsLayer, inner: S) -> Self {
+        Self {
+            cors_layer,
+            inner: Arc::new(Mutex::new(inner)),
+        }
+    }
+}
+
+impl<S> Service<router::Request> for CorsService<S>
+where
+    S: Service<router::Request, Response = router::Response, Error = BoxError> + Send + 'static,
+    S::Future: Send + 'static,
+{
+    type Response = router::Response;
+    type Error = BoxError;
+    type Future =
+        Pin<Box<dyn Future<Output = Result<Self::Response, Self::Error>> + Send + 'static>>;
+
+    fn poll_ready(&mut self, _cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        // For tokio::sync::Mutex, we can't poll_ready synchronously,
+        // so we just return Ready and handle the actual readiness in call()
+        Poll::Ready(Ok(()))
+    }
+
+    fn call(&mut self, req: router::Request) -> Self::Future {
+        // Convert router request to HTTP request
+        let http_req: http::Request<router::Body> = req.into();
+
+        // Create HTTP service that wraps the router service
+        let inner = self.inner.clone();
+        let http_service = tower::service_fn(move |http_req: http::Request<router::Body>| {
+            let inner = inner.clone();
+            async move {
+                let router_req: router::Request = http_req.into();
+                let mut inner = inner.lock().await;
+                let router_res = inner.ready().await?.call(router_req).await?;
+                let http_res: http::Response<router::Body> = router_res.into();
+                Ok::<_, BoxError>(http_res)
+            }
+        });
+
+        // Apply CORS layer to the HTTP service
+        let cors_service = self.cors_layer.layer(http_service);
+
+        // Call the CORS service and convert back
+        let mut cors_service = cors_service;
+        Box::pin(async move {
+            let http_res = cors_service.ready().await?.call(http_req).await?;
+            let router_res: router::Response = http_res.into();
+            Ok(router_res)
+        })
+    }
+}
+
+/// Service that handles trusted origins CORS logic
+#[derive(Clone)]
+struct TrustedOriginsCorsService<S> {
+    config: Cors,
+    inner: Arc<Mutex<S>>,
+}
+
+impl<S> TrustedOriginsCorsService<S> {
+    fn new(config: Cors, inner: S) -> Self {
+        Self {
+            config,
+            inner: Arc::new(Mutex::new(inner)),
+        }
+    }
+
+    /// Check if the origin is in the trusted list
+    fn is_trusted_origin(&self, origin: &str) -> bool {
+        if let Some(trusted_origins) = &self.config.trusted_origins {
+            trusted_origins.iter().any(|trusted| trusted == origin)
+        } else {
+            false
+        }
+    }
+
+    /// Handle preflight OPTIONS request
+    fn handle_preflight(&self, req: &router::Request) -> Option<router::Response> {
+        let headers = req.router_request.headers();
+
+        // Check if this is a preflight request
+        if req.router_request.method() == Method::OPTIONS {
+            if let Some(origin) = headers.get("origin") {
+                if let Ok(origin_str) = origin.to_str() {
+                    let is_trusted = self.is_trusted_origin(origin_str);
+
+                    // Build preflight response
+                    let mut response_headers = HeaderMap::new();
+
+                    // Set origin header
+                    if is_trusted {
+                        response_headers.insert("access-control-allow-origin", origin.clone());
+                        response_headers.insert(
+                            "access-control-allow-credentials",
+                            HeaderValue::from_static("true"),
+                        );
+                    } else {
+                        response_headers
+                            .insert("access-control-allow-origin", HeaderValue::from_static("*"));
+                    }
+
+                    // Set other CORS headers
+                    response_headers.insert(
+                        "access-control-allow-methods",
+                        HeaderValue::from_str(&self.config.methods.join(", "))
+                            .unwrap_or_else(|_| HeaderValue::from_static("GET, POST, OPTIONS")),
+                    );
+
+                    if !self.config.allow_headers.is_empty() {
+                        response_headers.insert(
+                            "access-control-allow-headers",
+                            HeaderValue::from_str(&self.config.allow_headers.join(", "))
+                                .unwrap_or_else(|_| HeaderValue::from_static("*")),
+                        );
+                    } else if let Some(requested_headers) =
+                        headers.get("access-control-request-headers")
+                    {
+                        response_headers
+                            .insert("access-control-allow-headers", requested_headers.clone());
+                    }
+
+                    if let Some(max_age) = &self.config.max_age {
+                        response_headers.insert(
+                            "access-control-max-age",
+                            HeaderValue::from_str(&max_age.as_secs().to_string())
+                                .unwrap_or_else(|_| HeaderValue::from_static("86400")),
+                        );
+                    }
+
+                    // Create response
+                    let response = http::Response::builder()
+                        .status(StatusCode::OK)
+                        .body(router::body::empty())
+                        .map_err(|_| "Failed to build preflight response")
+                        .ok()?;
+
+                    if let Ok(mut resp) = router::Response::http_response_builder()
+                        .response(response)
+                        .context(req.context.clone())
+                        .build()
+                    {
+                        // Copy headers to the response
+                        for (key, value) in response_headers {
+                            if let Some(key) = key {
+                                resp.response.headers_mut().insert(key, value);
+                            }
+                        }
+                        return Some(resp);
+                    }
+                }
+            }
+        }
+        None
+    }
+}
+
+impl<S> Service<router::Request> for TrustedOriginsCorsService<S>
+where
+    S: Service<router::Request, Response = router::Response, Error = BoxError> + Send + 'static,
+    S::Future: Send + 'static,
+{
+    type Response = router::Response;
+    type Error = BoxError;
+    type Future =
+        Pin<Box<dyn Future<Output = Result<Self::Response, Self::Error>> + Send + 'static>>;
+
+    fn poll_ready(&mut self, _cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        Poll::Ready(Ok(()))
+    }
+
+    fn call(&mut self, req: router::Request) -> Self::Future {
+        // Handle preflight request
+        if let Some(preflight_response) = self.handle_preflight(&req) {
+            return Box::pin(async move { Ok(preflight_response) });
+        }
+
+        let inner = self.inner.clone();
+        let config = self.config.clone();
+
+        // Extract origin from request before moving it
+        let origin = req
+            .router_request
+            .headers()
+            .get("origin")
+            .and_then(|h| h.to_str().ok())
+            .map(|s| s.to_string());
+
+        Box::pin(async move {
+            let mut inner = inner.lock().await;
+            let mut response = inner.ready().await?.call(req).await?;
+
+            // Add CORS headers to the response based on the extracted origin
+            let response_headers = response.response.headers_mut();
+
+            if let Some(origin_str) = origin {
+                let is_trusted = if let Some(trusted_origins) = &config.trusted_origins {
+                    trusted_origins.iter().any(|trusted| trusted == &origin_str)
+                } else {
+                    false
+                };
+
+                if is_trusted {
+                    response_headers.insert(
+                        "access-control-allow-origin",
+                        HeaderValue::from_str(&origin_str)
+                            .unwrap_or_else(|_| HeaderValue::from_static("*")),
+                    );
+                    response_headers.insert(
+                        "access-control-allow-credentials",
+                        HeaderValue::from_static("true"),
+                    );
+                } else {
+                    response_headers
+                        .insert("access-control-allow-origin", HeaderValue::from_static("*"));
+                }
+
+                // Add expose headers if configured
+                if let Some(expose_headers) = &config.expose_headers {
+                    if !expose_headers.is_empty() {
+                        if let Ok(expose_value) = HeaderValue::from_str(&expose_headers.join(", "))
+                        {
+                            response_headers.insert("access-control-expose-headers", expose_value);
+                        }
+                    }
+                }
+            } else {
+                // No origin header, allow all
+                response_headers
+                    .insert("access-control-allow-origin", HeaderValue::from_static("*"));
+            }
+
+            Ok(response)
+        })
+    }
+}
+
+register_plugin!("apollo", "cors", CorsPlugin);
+
+#[cfg(test)]
+mod tests {
+    use std::time::Duration;
+
+    use http::Method;
+    use http::StatusCode;
+    use tower::ServiceExt;
+
+    use super::*;
+    use crate::plugin::test::MockRouterService;
+    use crate::services::router::Request;
+    use crate::services::router::Response;
+
+    #[tokio::test]
+    async fn test_cors_plugin_default_config() {
+        let mut mock_service = MockRouterService::new();
+        mock_service.expect_call().times(1).returning(|_| {
+            Ok(Response::fake_builder()
+                .status_code(StatusCode::OK)
+                .build()
+                .unwrap())
+        });
+
+        let plugin = CorsPlugin::new(PluginInit::fake_new(Cors::default(), Default::default()))
+            .await
+            .expect("Failed to create CORS plugin");
+
+        let mut service = plugin.router_service(mock_service.boxed());
+
+        let request = Request::fake_builder()
+            .method(Method::GET)
+            .header("origin", "https://studio.apollographql.com")
+            .build()
+            .unwrap();
+
+        let response = service.ready().await.unwrap().call(request).await.unwrap();
+
+        // Check that CORS headers are present
+        let headers = response.response.headers();
+        assert!(headers.contains_key("access-control-allow-origin"));
+        assert_eq!(
+            headers.get("access-control-allow-origin").unwrap(),
+            "https://studio.apollographql.com"
+        );
+    }
+
+    #[tokio::test]
+    async fn test_cors_plugin_with_credentials() {
+        let mut mock_service = MockRouterService::new();
+        mock_service.expect_call().times(1).returning(|_| {
+            Ok(Response::fake_builder()
+                .status_code(StatusCode::OK)
+                .build()
+                .unwrap())
+        });
+
+        let config = Cors::builder()
+            .allow_credentials(true)
+            .origins(vec!["https://example.com".to_string()])
+            .build();
+
+        let plugin = CorsPlugin::new(PluginInit::fake_new(config, Default::default()))
+            .await
+            .expect("Failed to create CORS plugin");
+
+        let mut service = plugin.router_service(mock_service.boxed());
+
+        let request = Request::fake_builder()
+            .method(Method::GET)
+            .header("origin", "https://example.com")
+            .build()
+            .unwrap();
+
+        let response = service.ready().await.unwrap().call(request).await.unwrap();
+
+        let headers = response.response.headers();
+        assert!(headers.contains_key("access-control-allow-credentials"));
+        assert_eq!(
+            headers.get("access-control-allow-credentials").unwrap(),
+            "true"
+        );
+    }
+
+    #[tokio::test]
+    async fn test_cors_plugin_trusted_origins_trusted_domain() {
+        let mut mock_service = MockRouterService::new();
+        mock_service.expect_call().times(1).returning(|_| {
+            Ok(Response::fake_builder()
+                .status_code(StatusCode::OK)
+                .build()
+                .unwrap())
+        });
+
+        let config = Cors::builder()
+            .trusted_origins(vec!["https://trusted.com".to_string()])
+            .build();
+
+        let plugin = CorsPlugin::new(PluginInit::fake_new(config, Default::default()))
+            .await
+            .expect("Failed to create CORS plugin");
+
+        let mut service = plugin.router_service(mock_service.boxed());
+
+        let request = Request::fake_builder()
+            .method(Method::GET)
+            .header("origin", "https://trusted.com")
+            .build()
+            .unwrap();
+
+        let response = service.ready().await.unwrap().call(request).await.unwrap();
+
+        let headers = response.response.headers();
+        assert_eq!(
+            headers.get("access-control-allow-origin").unwrap(),
+            "https://trusted.com"
+        );
+        assert_eq!(
+            headers.get("access-control-allow-credentials").unwrap(),
+            "true"
+        );
+    }
+
+    #[tokio::test]
+    async fn test_cors_plugin_trusted_origins_untrusted_domain() {
+        let mut mock_service = MockRouterService::new();
+        mock_service.expect_call().times(1).returning(|_| {
+            Ok(Response::fake_builder()
+                .status_code(StatusCode::OK)
+                .build()
+                .unwrap())
+        });
+
+        let config = Cors::builder()
+            .trusted_origins(vec!["https://trusted.com".to_string()])
+            .build();
+
+        let plugin = CorsPlugin::new(PluginInit::fake_new(config, Default::default()))
+            .await
+            .expect("Failed to create CORS plugin");
+
+        let mut service = plugin.router_service(mock_service.boxed());
+
+        let request = Request::fake_builder()
+            .method(Method::GET)
+            .header("origin", "https://untrusted.com")
+            .build()
+            .unwrap();
+
+        let response = service.ready().await.unwrap().call(request).await.unwrap();
+
+        let headers = response.response.headers();
+        assert_eq!(headers.get("access-control-allow-origin").unwrap(), "*");
+        assert!(!headers.contains_key("access-control-allow-credentials"));
+    }
+
+    #[tokio::test]
+    async fn test_cors_plugin_trusted_origins_preflight_trusted() {
+        let mut mock_service = MockRouterService::new();
+        mock_service.expect_call().times(0); // Preflight should not reach the service
+
+        let config = Cors::builder()
+            .trusted_origins(vec!["https://trusted.com".to_string()])
+            .methods(vec![
+                "GET".to_string(),
+                "POST".to_string(),
+                "OPTIONS".to_string(),
+            ])
+            .build();
+
+        let plugin = CorsPlugin::new(PluginInit::fake_new(config, Default::default()))
+            .await
+            .expect("Failed to create CORS plugin");
+
+        let mut service = plugin.router_service(mock_service.boxed());
+
+        let request = Request::fake_builder()
+            .method(Method::OPTIONS)
+            .header("origin", "https://trusted.com")
+            .header("access-control-request-method", "POST")
+            .build()
+            .unwrap();
+
+        let response = service.ready().await.unwrap().call(request).await.unwrap();
+
+        let headers = response.response.headers();
+        assert_eq!(
+            headers.get("access-control-allow-origin").unwrap(),
+            "https://trusted.com"
+        );
+        assert_eq!(
+            headers.get("access-control-allow-credentials").unwrap(),
+            "true"
+        );
+        assert!(headers.contains_key("access-control-allow-methods"));
+    }
+
+    #[tokio::test]
+    async fn test_cors_plugin_trusted_origins_preflight_untrusted() {
+        let mut mock_service = MockRouterService::new();
+        mock_service.expect_call().times(0); // Preflight should not reach the service
+
+        let config = Cors::builder()
+            .trusted_origins(vec!["https://trusted.com".to_string()])
+            .methods(vec![
+                "GET".to_string(),
+                "POST".to_string(),
+                "OPTIONS".to_string(),
+            ])
+            .build();
+
+        let plugin = CorsPlugin::new(PluginInit::fake_new(config, Default::default()))
+            .await
+            .expect("Failed to create CORS plugin");
+
+        let mut service = plugin.router_service(mock_service.boxed());
+
+        let request = Request::fake_builder()
+            .method(Method::OPTIONS)
+            .header("origin", "https://untrusted.com")
+            .header("access-control-request-method", "POST")
+            .build()
+            .unwrap();
+
+        let response = service.ready().await.unwrap().call(request).await.unwrap();
+
+        let headers = response.response.headers();
+        assert_eq!(headers.get("access-control-allow-origin").unwrap(), "*");
+        assert!(!headers.contains_key("access-control-allow-credentials"));
+        assert!(headers.contains_key("access-control-allow-methods"));
+    }
+
+    #[tokio::test]
+    async fn test_cors_plugin_preflight_request() {
+        let mut mock_service = MockRouterService::new();
+        // For preflight requests, the service might not be called as CORS layer handles it
+        mock_service.expect_call().times(0..=1).returning(|_| {
+            Ok(Response::fake_builder()
+                .status_code(StatusCode::OK)
+                .build()
+                .unwrap())
+        });
+
+        let config = Cors::builder()
+            .allow_headers(vec![
+                "content-type".to_string(),
+                "authorization".to_string(),
+            ])
+            .methods(vec![
+                "GET".to_string(),
+                "POST".to_string(),
+                "OPTIONS".to_string(),
+            ])
+            .max_age(Duration::from_secs(3600))
+            .build();
+
+        let plugin = CorsPlugin::new(PluginInit::fake_new(config, Default::default()))
+            .await
+            .expect("Failed to create CORS plugin");
+
+        let mut service = plugin.router_service(mock_service.boxed());
+
+        let request = Request::fake_builder()
+            .method(Method::OPTIONS)
+            .header("origin", "https://studio.apollographql.com")
+            .header("access-control-request-method", "POST")
+            .header("access-control-request-headers", "content-type")
+            .build()
+            .unwrap();
+
+        let response = service.ready().await.unwrap().call(request).await.unwrap();
+
+        let headers = response.response.headers();
+        assert!(headers.contains_key("access-control-allow-origin"));
+        assert!(headers.contains_key("access-control-allow-methods"));
+        assert!(headers.contains_key("access-control-max-age"));
+    }
+
+    #[tokio::test]
+    async fn test_cors_plugin_with_match_origins() {
+        let mut mock_service = MockRouterService::new();
+        mock_service.expect_call().times(1).returning(|_| {
+            Ok(Response::fake_builder()
+                .status_code(StatusCode::OK)
+                .build()
+                .unwrap())
+        });
+
+        let config = Cors::builder()
+            .origins(vec!["https://example.com".to_string()])
+            .match_origins(vec![r"https://.*\.example\.com".to_string()])
+            .build();
+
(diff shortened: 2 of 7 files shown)


theJC:
This is awesome that the CORS support in Router is finally being evolved. 

We at Indeed absolutely need more flexibility along these lines in order to have parity in what we had with Apollo Gateway (and the CORS support koa's library allowed us to configure) before we migrate our web traffic to Router.

What we see in this early draft pull request is moving towards what we need, however in order for this to fully support our needs, we also require:

- "trusted_match_origins" - We need regex match support for a given CORS policy, which identifies our "owned origins"
    -   both regex matches and a list of origins should be supported
- Ability to control for `Access-Control-Expose-Headers` header to be controllable for a given CORS policy that is for our "owned origins"


ie:
```
cors:
  methods: [POST, OPTIONS]  # Global default
  origins:
    # Trusted origins get credentials
    - matched_origins:
       - "^https://.*?.mycompany.(?:com|jp|co.uk)(?::[1-9][0-9]*)?$"
    - origins:
        - "https://studio.apollographql.com"
        - "https://app.siblingcompany.com"
        - "https://jobs.trustedthirdpartycompany.com"
      allow_credentials: true
      allow_headers: ["content-type", "authorization", "bespokeheader1", "bespokeheader2"]
      expose_headers: ["header1",  "header2"]
    
    # Catch-all for untrusted origins
    - origins: ["*"]
      allow_credentials: false
      allow_headers: ["content-type", "authorization","bespokeheader1", "bespokeheader2"]
      # Note: not exposing any headers to client-side javascript for untrusted origins
```

Please refer to TSH-18129 for more detailed/specific examples of the regex we currently use for our trusted origins

cc: @abernix

apollo-router/src/axum_factory/tests.rs
```diff
@@ -1126,7 +1127,11 @@ async fn cors_preflight() -> Result<(), ApolloRouterError> {
     assert_header_contains!(
         &response,
         ACCESS_CONTROL_ALLOW_HEADERS,
-        &["Content-type, x-an-other-test-header, apollo-require-preflight"],
+        &[
+            "content-type",
```
Velfi, line 1130: We're doing a case-sensitive test here since these are header values so I changed `Content-type` to the format we're using elsewhere in the codebase: `content-type`.

Velfi:
@theJC Could I get your review on this?

docs/source/routing/security/cors.mdx
```diff
@@ -67,8 +68,9 @@ To allow browsers to pass credentials to the router, set `allow_credentials` to
 ```yaml {6} title="router.yaml"
 cors:
   origins:
-    - https://www.your-app.example.com
-    - https://studio.apollographql.com
+    - origins:
```
theJC, line 71: Hardest part of comp sci is naming things.  I think customers may benefit if a different name was come up with for one of the two nested "origin" names at the different levels.  It's caused me confusion in reading the code and config.

Spitballing ideas.... perhaps the upper one can be called `policies` ?

apollo-router/src/configuration/cors.rs
```diff
@@ -84,90 +154,338 @@ impl Cors {
...
+            .and_then(|oc| {
+                if oc.methods.is_empty() {
+                    None
+                } else {
+                    Some(&oc.methods)
+                }
+            })
+            .unwrap_or(&config.methods);
+
+        let max_age = config.max_age;
```
theJC: We dont have a neeed for this, but could other customers have a use case to have max_age be able to be set both globally and differently in each originconfig?

apollo-router/src/configuration/cors.rs
```diff
@@ -84,90 +154,338 @@ impl Cors {
...
+
+        None
+    }
+
+    /// Add CORS headers to the response
+    fn add_cors_headers<ResBody>(
+        response: &mut Response<ResBody>,
+        config: &Cors,
+        origin_config: &Option<&OriginConfig>,
+        origin: &Option<http::HeaderValue>,
```
theJC: nit: maybe rename to request_origin to make it more clear via its name?

apollo-router/src/configuration/cors.rs
```diff
@@ -84,90 +154,338 @@ impl Cors {
...
+                ACCESS_CONTROL_ALLOW_CREDENTIALS,
+                http::HeaderValue::from_static("true"),
+            );
+        }
+
+        // Set Access-Control-Allow-Headers
+        if !allow_headers.is_empty() {
+            for header in allow_headers {
+                // Use the header value as-is to match test expectations
+                response.headers_mut().append(
```
theJC: My Rust is weak as well as I dont know these basic http classes well.  

Is there an optimization available here on precomputing the headervalue and doing an insert once instead of append?  less lookups/memory allocations

apollo-router/src/configuration/cors.rs
```diff
@@ -84,90 +154,338 @@ impl Cors {
...
+        for origin_config in &self.origins {
+            if origin_config.origins.iter().any(|x| x == "*") {
+                return Err(
+                    "Invalid CORS configuration: use `allow_any_origin: true` to set `Access-Control-Allow-Origin: *`",
+                );
+            }
         }
+
         if self.allow_credentials {
             if self.allow_headers.iter().any(|x| x == "*") {
```
theJC, line 280: do some of these checks also need to be done for the origin_configs as well?

apollo-router/src/configuration/cors.rs
```diff
@@ -84,90 +154,338 @@ impl Cors {
...
+    type Response = S::Response;
+    type Error = S::Error;
+    type Future = Pin<Box<dyn Future<Output = Result<Self::Response, Self::Error>> + Send>>;
+
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        self.inner.poll_ready(cx)
+    }
+
+    fn call(&mut self, req: Request<ReqBody>) -> Self::Future {
+        let origin = req.headers().get(ORIGIN).cloned();
```
theJC: nit: rename to request_origin

apollo-router/src/configuration/cors.rs
```diff
@@ -1,18 +1,101 @@
...
+    /// Defaults to an empty list.
+    /// Note that `origins` will be evaluated before `match_origins`
+    #[serde(with = "serde_regex")]
+    #[schemars(with = "Vec<String>")]
+    pub(crate) match_origins: Vec<Regex>,
+
+    /// Allowed request methods for these origins
+    pub(crate) methods: Vec<String>,
+
+    /// The origins to allow requests from
```
theJC: document default value

apollo-router/src/configuration/cors.rs
```diff
@@ -1,18 +1,101 @@
...
+    pub(crate) expose_headers: Vec<String>,
+
+    /// `Regex`es you want to match the origins against to determine if they're allowed.
+    /// Defaults to an empty list.
+    /// Note that `origins` will be evaluated before `match_origins`
+    #[serde(with = "serde_regex")]
+    #[schemars(with = "Vec<String>")]
+    pub(crate) match_origins: Vec<Regex>,
+
+    /// Allowed request methods for these origins
```
theJC: document default value

apollo-router/src/configuration/cors.rs
```diff
@@ -84,90 +154,338 @@ impl Cors {
...
+            for method in methods {
+                response.headers_mut().append(
+                    ACCESS_CONTROL_ALLOW_METHODS,
+                    http::HeaderValue::from_str(method)
+                        .unwrap_or_else(|_| http::HeaderValue::from_static("")),
+                );
+            }
         }
+
+        // Set Access-Control-Max-Age
```
theJC: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Access-Control-Max-Age indicates this header is only on preflight requests.

Also:
https://github.com/tower-rs/tower-http/blob/35a6cb83242b2004352a8a45f97c0c2cb5475254/tower-http/src/cors/mod.rs#L683

apollo-router/src/configuration/cors.rs
```diff
@@ -84,90 +154,338 @@ impl Cors {
...
+        // Set Access-Control-Allow-Credentials
+        if allow_credentials {
+            response.headers_mut().insert(
+                ACCESS_CONTROL_ALLOW_CREDENTIALS,
+                http::HeaderValue::from_static("true"),
+            );
+        }
+
+        // Set Access-Control-Allow-Headers
+        if !allow_headers.is_empty() {
```
theJC: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Access-Control-Allow-Headers indicates this header is only for preflight requests.

Also: https://github.com/tower-rs/tower-http/blob/35a6cb83242b2004352a8a45f97c0c2cb5475254/tower-http/src/cors/mod.rs#L682C39-L682C52

apollo-router/src/configuration/cors.rs
```diff
@@ -84,90 +154,338 @@ impl Cors {
...
+                        ACCESS_CONTROL_ALLOW_HEADERS,
+                        http::HeaderValue::from_str(headers_str)
+                            .unwrap_or_else(|_| http::HeaderValue::from_static("")),
+                    );
+                }
+            }
+        }
+
+        // Set Access-Control-Expose-Headers
+        if let Some(headers) = expose_headers {
```
theJC: This header is only for NON preflight requests

https://github.com/tower-rs/tower-http/blob/35a6cb83242b2004352a8a45f97c0c2cb5475254/tower-http/src/cors/mod.rs#L693

theJC:
Note - 
Please know this issue exists as well: https://github.com/apollographql/router/issues/7702 which is another CORS header control that will need to be added into Router to support developers running router locally (and support for this header exists in the tower library today).

Im unsure if its better to do this now in this PR while we got the hood open on CORS or, wait until the dust settles after this change.

BrynCooke (changes requested):
It looks like there are a bunch of new tests for config but we don't actually send a request through the layer to see if the responses match our expectations. 

I ran a coverage test and it looks like allow credentials and expose headers in add_cors_headers is not exercised.

` cargo llvm-cov nextest --package apollo-router --html -- cors`

apollo-router/src/configuration/cors.rs
```diff
@@ -84,90 +154,338 @@ impl Cors {
...
+impl<S> CorsService<S> {
+    /// Find the matching origin configuration for a given origin
+    fn find_matching_origin_config<'a>(
+        config: &'a Cors,
+        origin: &'a Option<http::HeaderValue>,
+    ) -> Option<&'a OriginConfig> {
+        let origin_str = origin.as_ref()?.to_str().ok()?;
+
+        // Check for exact origin matches first
+        for origin_config in &config.origins {
```
BrynCooke: This code will check each origin in order, does it make sense to check all exact matches first before any regex matches? I could see users adding an exact match below a match origin and then not realising that they their rule is not being hit.

apollo-router/src/configuration/cors.rs
```diff
@@ -84,90 +154,338 @@ impl Cors {
...
+
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        self.inner.poll_ready(cx)
+    }
+
+    fn call(&mut self, req: Request<ReqBody>) -> Self::Future {
+        let origin = req.headers().get(ORIGIN).cloned();
+        let is_preflight = req.method() == http::Method::OPTIONS;
+        let config = self.config.clone();
+        let request_headers = req.headers().get("access-control-request-headers").cloned();
```
BrynCooke: There is a constant for this:
```suggestion
        let request_headers = req.headers().get(ACCESS_CONTROL_REQUEST_HEADERS).cloned();
```

apollo-router/src/configuration/cors.rs
BrynCooke: Do we need to do anything with `Access-Control-Request-Method` ?

apollo-router/src/configuration/cors.rs
Velfi: That could be confusing for users but I did my best to explain it in docs. I'm open to suggestions.

apollo-router/src/configuration/cors.rs
Velfi: **No, we don't need to do anything with `Access-Control-Request-Method`.**

Here's why:

1. **`Access-Control-Request-Method` is a request header**: According to the [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Access-Control-Request-Method), this header is sent by browsers during preflight requests to tell the server which HTTP method will be used in the actual request.

2. **Our current implementation is correct**: We don't need to read or validate this header because:
   - We detect preflight requests by checking if `req.method() == http::Method::OPTIONS`
   - We respond with `Access-Control-Allow-Methods` containing all the methods we're willing to allow
   - The browser will then decide whether to proceed with the actual request based on whether the requested method is in our allowed list

3. **This follows the CORS specification**: The `Access-Control-Request-Method` header is informational - it tells us what the browser wants to do, but we respond with our own policy via `Access-Control-Allow-Methods`. The browser then enforces whether the actual request can proceed.

4. **Our tests confirm this behavior**: The test shows that we correctly handle preflight requests with `Access-Control-Request-Method: POST` and respond with the appropriate `Access-Control-Allow-Methods` header.

The current implementation is already handling CORS preflight requests correctly according to the specification.

Velfi:
> It looks like there are a bunch of new tests for config but we don't actually send a request through the layer to see if the responses match our expectations.
> 
> I ran a coverage test and it looks like allow credentials and expose headers in add_cors_headers is not exercised.
> 
> ` cargo llvm-cov nextest --package apollo-router --html -- cors`

Thanks. I added several new tests because of this.

apollo-router/src/configuration/cors.rs
Velfi: You're right. I'll update that.

apollo-router/src/configuration/cors.rs
theJC: As a customer, I would expect the policies to be searched in order.  
- First the first listed policy, check exact matches then regexes (if both were configured)
- only if not matched then move to the next policy
- my personal opinion is doing it differently would be confusing and not expected

apollo-router/src/configuration/cors.rs
```diff
@@ -1,18 +1,173 @@
...
+    ///     - origins: []
+    ///
+    /// # Within a policy - custom origins
+    /// cors:
+    ///   policies:
+    ///     - origins:
+    ///         - https://myapp.com
+    ///         - https://studio.apollographql.com
+    /// ```
+    pub(crate) origins: Vec<String>,
```
theJC, line 100: ~Right now what Im seeing when trying to craft a new Router config using a yaml schema based off this this branch is that I am required to specify `origins`, otherwise it fails schema validation~

~Shouldnt this be Option (or does something else need to happen) to allow either origin and/or match_origins or allow_any_origin be specified in a policy, but not require `origins` always be present?~

theJC: Actually, let me try again, I think I was using another binary/schema.... sorry....

theJC: Ok, looks like the config I was trying get setup is validating fine once I used the right binary.

Please ignore this original comment, again sorry.

apollo-router/src/configuration/cors.rs
BrynCooke: SGTM

apollo-router/src/configuration/cors.rs
BrynCooke: Great

apollo-router/Cargo.toml
```diff
@@ -208,7 +208,7 @@ serde.workspace = true
 serde_derive_default = "0.1"
 serde_json_bytes.workspace = true
 serde_json.workspace = true
-serde_regex = { version = "1.1.0", optional = true }
+serde_regex = { version = "1.1.0" }
```
goto-bus-stop, line 211: We can remove the corresponding entry in `dev-dependencies` now that it's unconditionally used here

apollo-router/src/configuration/cors.rs
```diff
@@ -44,22 +199,62 @@ pub(crate) struct Cors {
...
-    /// Defaults to an empty list.
-    /// Note that `origins` will be evaluated before `match_origins`
-    pub(crate) match_origins: Option<Vec<String>>,
-
     /// Allowed request methods. Defaults to GET, POST, OPTIONS.
+    ///
+    /// - **Omit this field (do nothing):** Defaults to `["GET", "POST", "OPTIONS"]`.
+    /// - `methods: []` (empty list): No methods are allowed.
+    ///
+    /// This field uses [`#[serde(default = "default_cors_methods")]`](https://serde.rs/attributes.html#default).
```
goto-bus-stop: does it?

apollo-router/src/configuration/cors.rs
```diff
@@ -198,6 +665,32 @@ impl Cors {
                     );
                 }
             }
+
+            // Check per-policy fields for wildcards when credentials are enabled
+            if let Some(policies) = &self.policies {
```
goto-bus-stop: it looks like this is currently only checked for the "global" allow_credentials, not the origin-specific one, but this should also fail: 
```yaml
cors:
  allow_credentials: false
  policies:
    - { allow_any_origin: true, allow_credentials: true }
```

docs/shared/config/cors.mdx
```diff
@@ -12,6 +12,6 @@ cors:
   - GET
   - POST
   - OPTIONS
-  origins:
+  policies:
```
goto-bus-stop, line 14: this doesnt seem correct now? same for most of the config changes in the docs. `policies` should have a list of policy objects, not URLs

apollo-router/src/configuration/cors.rs
```diff
@@ -84,91 +271,371 @@ impl Cors {
...
+            policies: policies.or_else(|| Some(vec![Policy::default()])),
         }
     }
 }
 
-impl Cors {
-    pub(crate) fn into_layer(self) -> Result<CorsLayer, String> {
+/// Our custom CORS layer that supports per-origin configuration
+#[derive(Clone, Debug)]
+pub(crate) struct CorsLayer {
```
goto-bus-stop: We should move the `CorsLayer` / `CorsService` implementations out of the `configuration` module

apollo-router/src/configuration/cors.rs
```diff
@@ -1,18 +1,173 @@
...
+    ///     - methods: [GET, POST]
+    /// ```
+    pub(crate) methods: Vec<String>,
+
+    /// The origins to allow requests from.
+    ///
+    /// **Note:** These defaults only apply when the entire `policies` field is omitted from the CORS config.
+    /// When specifying individual policies, this field defaults to an empty list unless explicitly set.
+    ///
+    /// - **When `policies` is omitted:** Defaults to `["https://studio.apollographql.com"]` (uses default Policy).
```
goto-bus-stop: It seems confusing to document this case on a config object that you can only construct within a `policies` key. I think this documentation should exist only on the top-level cors configuration.

apollo-router/src/configuration/cors.rs
```diff
@@ -44,22 +199,62 @@ pub(crate) struct Cors {
...
+    /// in order, first by exact match, then by regex. The first policy that matches the request
+    /// origin will be used to determine the CORS headers to return.
+    ///
+    /// - **Omit this field (do nothing):** Use the default Apollo Studio policy (`https://studio.apollographql.com`)
+    /// - `policies: []` (empty list): Disable all origins (no CORS headers will be set)
+    /// - `policies: [...]`: Use custom policies
+    ///
+    /// This field uses [`#[serde(default)]`](https://serde.rs/attributes.html#default) and
+    /// [`Option<T>` handling](https://serde.rs/attr-default.html), so omitting the field in YAML
+    /// will use the default, while specifying an empty list disables CORS.
```
goto-bus-stop: Users will use YAML, so we shouldn't mention serde or Options in the docs

apollo-router/src/configuration/cors.rs
```diff
@@ -84,91 +271,371 @@ impl Cors {
...
+
+impl<S> CorsService<S> {
+    /// Find the matching policy for a given origin
+    fn find_matching_policy<'a>(
+        config: &'a Cors,
+        origin: &'a Option<http::HeaderValue>,
+    ) -> Option<&'a Policy> {
+        let origin_str = origin.as_ref()?.to_str().ok()?;
+
+        // Check for exact origin matches first
```
goto-bus-stop: It doesn't seem like it actually does that, it checks each policy in order and uses the first match.

IMO it's better to use the first match than to try to implicitly reorder it, but I know not everyone agrees.

apollo-router/src/configuration/cors.rs
```diff
@@ -84,91 +271,371 @@ impl Cors {
...
+        if let Some(policies) = &config.policies {
+            for policy in policies.iter() {
```
goto-bus-stop: Do we also need to check `allow_any_origin` here?

apollo-router/src/configuration/cors.rs
goto-bus-stop: Oh, or is `allow_any_origin` only allowed on the top level config? I guess then this works fine

apollo-router/src/configuration/cors.rs
goto-bus-stop: (Probably not: https://github.com/apollographql/router/pull/7853/files#r2215958170)

theJC:
Here is an example config we are using to test this branch as of the evening of July 17th.

This may help pressure-test by providing an example based upon our company's specific needs to point to while trying to arrive to consensus for some threads.  I'd be glad to create some other hypothetical config use cases that aren't what we need but help drive discussions.

```yaml
cors:
  allow_headers:
    - authorization
    - content-type
    - some-other-headers
  methods: [ POST ]      # Global default
  max_age: 24h
  policies:
    - # Trusted Origins Policy
      match_origins:      
        - "^https://.*?.company.(?:com|uk)$"
        # Additional
        - "^https://.*?.partner1.com$"
        - "^https://.*?.partner2.com$"        
      origins: ["https://studio.apollographql.com"]
      allow_credentials: true
      expose_headers: ["some-response-header"]
    - # Non-Trusted Origins Policy for all other origins
      match_origins: [".*"]
      allow_credentials: false
```

Just making sure using the regex pattern `.*` for everything is the intent/expected and only way to have a final "catch-all" bucket for any origin other than those that match the earlier policies containing specific origins/match_origins.

apollo-router/src/configuration/cors.rs
```diff
@@ -44,22 +151,17 @@ pub(crate) struct Cors {
...
-    pub(crate) origins: Vec<String>,
-
-    /// `Regex`es you want to match the origins against to determine if they're allowed.
-    /// Defaults to an empty list.
-    /// Note that `origins` will be evaluated before `match_origins`
-    pub(crate) match_origins: Option<Vec<String>>,
-
-    /// Allowed request methods. Defaults to GET, POST, OPTIONS.
+    /// Allowed request methods. See module documentation for default behavior.
     pub(crate) methods: Vec<String>,
```
theJC, line 179: As a consumer of this feature, this is not currently working as one would expect.

When (see my example config up [above](https://github.com/apollographql/router/pull/7853#issuecomment-3089944763)) :
- this global default "methods" is set to POST in the top level config
- there exists policy/policies that do not specify "methods"
- at runtime origins that match the policies are currently getting response of GET, POST, OPTIONS (the Default for Policy defines a default of default_cors_methods )
- this defeats the whole purpose of having a global default


The expected behavior is if someone does not specify an override for `methods` in a `policy`, the global default `methods` should apply.

apollo-router/src/plugins/cors.rs
```diff
@@ -0,0 +1,572 @@
...
+            } else {
+                Some(&policy.expose_headers)
+            }
+        } else {
+            config.expose_headers.as_ref()
+        };
+
+        let methods = policy
+            .and_then(|p| {
+                if p.methods.is_empty() {
```
theJC: Per my comment in config... if a customer doesnt specify methods in a policy, the code currently defaults it to GET, POST,  OPTIONS -- therefore this wont return None, and then the unwrap below to default to the global will not happen.

This effectively makes the global default ineffective.

apollo-router/src/plugins/cors.rs
```diff
@@ -0,0 +1,572 @@
...
+                    .map(|header| {
+                        http::HeaderValue::from_str(header)
+                            .unwrap_or_else(|_| http::HeaderValue::from_static(""))
+                    })
+                    .collect();
+
+                for header_value in header_values {
+                    response
+                        .headers_mut()
+                        .append(ACCESS_CONTROL_ALLOW_HEADERS, header_value);
```
theJC: this results in different cors response header behavior than tower http's cors and also the javascript koa framework's cors that we use Apollo Gateway on. 

This code currently results in many different access-control-allow-headers headers returned, each one with one of the allowed headers.  Although these are supposedly equivalent per [IETF rfc](https://www.rfc-editor.org/rfc/rfc9110.html#name-field-lines-and-combined-fi), in my experience the comma separated mechanism is much more prevalent in practice.  Additionally I would recommend trying to make the resulting responses imperceptible from the original tower cors.

towercors: 
towercors

apollocors (behavior of this MR at this time):
apollocors

theJC: Similarly for the access-control-allow-methods response header, the recommendation is to make that be one header with comma separated values when more than one is present.

theJC: Further reason to collapse to only one header: browser parsing inconsistencies:  https://claude.ai/share/75f9cc30-1c45-444c-9b6a-caf439a51489

Velfi: fixed

apollo-router/src/plugins/cors.rs
Velfi: should be fixed now.

Velfi:
@theJC Can I get your review of the latest changes when you have time?

Re:
> Just making sure using the regex pattern .* for everything is the intent/expected and only way to have a final "catch-all" bucket for any origin other than those that match the earlier policies containing specific origins/match_origins.

How would you like it to work?

theJC:
> Re:
> 
> > Just making sure using the regex pattern .* for everything is the intent/expected and only way to have a final "catch-all" bucket for any origin other than those that match the earlier policies containing specific origins/match_origins.
> 
> How would you like it to work?

I think this is the only mechanism we currently have to accomplish this unless Apollo added a specialized policy indicator that indicates a policy is a "catch all" (except for `null` origin).    I dont know how "special case" this is that we do something like this, so Im not inclined to request/push for formal support of something like that, and am fine using the regex, just wanted to make sure I wasnt missing another way to be able to accomplish that.

theJC:
> @theJC Can I get your review of the latest changes when you have time?

I've got a series of system tests written in our codebase for the preflight requests but not the main graphql POST requests yet (note we dont use GET at all, just POST).

By the end of the business day CDT I should be able to do a re-review and also re-run those preflight tests to provide you feedback.   I should have the non-preflight requests system tests done sometime tomorrow and test this branch again at that time.

apollo-router/src/plugins/cors.rs
theJC: Verified that our preflight requests system tests for this scenario now pass with this configuration now.

theJC:
An issue I believe I've identified is:

- The documented intent is to allow "null" origin only via setting allow_any_origin which is at the global settings.  https://github.com/apollographql/router/pull/7853/files#diff-ea34c7fa4e3f6984aae7f7e0d20d22930d5386763c6be9d9fb2f017adf16f746R2347
 > "allow_any_origin": {
          "default": false,
          "description": "Set to true to allow any origin. Defaults to false. This is the only way to allow Origin: null.", 

- Right now, null origin is able to be allowed if I use our fallback / catch-all final policy that matches ".*" origin
```
    - # Non-Trusted Origins Policy for all other origins
      match_origins: [".*"]
      allow_credentials: false
```     
- null origins are highly [not-recommended ](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Access-Control-Allow-Origin), so if this was left as possible, I think this could be something Apollo's customers could easily inadvertently allow and be a security issue.  If and when Apollo would want to allow customers do support any corner cases of allowing null other than setting the top level allow_any_origin, that should be a separate ticket/new configuration item that be exposed in a policy.  

- I think your best option may be to guard against the null value when executing on a request unless allow_any_origin is true

- seems like you probably need to add a test case to verify and prevent this isn't allowed/regressed in the future.

apollo-router/src/configuration/cors.rs
theJC: Verified this is now working as I would expect.

apollo-router/src/configuration/cors.rs
```diff
@@ -1,28 +1,141 @@
...
+//!
+//! When specifying individual policies within the `policies` array:
+//! - **Origins:** Defaults to an empty list (no origins allowed) unless explicitly set
+//! - **Methods:** Has three possible states:
+//!   - `null` (not specified): Use the global default methods
+//!   - `[]` (empty array): No methods allowed for this policy
+//!   - `[values]` (with values): Use these specific methods
+//! - **Allow headers:** Defaults to an empty list (mirrors client headers) unless explicitly set
+//! - **Expose headers:** Defaults to an empty list unless explicitly set
+//! - **Match origins:** Defaults to an empty list (no regex matching) unless explicitly set
```
theJC: recommend to move Match origins up and immediately after Origins

apollo-router/src/plugins/cors.rs
```diff
@@ -0,0 +1,694 @@
...
+                request_headers,
+            );
+            return Box::pin(async move { Ok(response) });
+        }
+
+        let fut = self.inner.call(req);
+        Box::pin(async move {
+            let mut response = fut.await?;
+            // Find matching origin configuration
+            let policy = Self::find_matching_policy(&config, &request_origin);
```
theJC, line 149: These two lines are repeated both here and inside the preflight if block 18 lines up.

Should they be moved up and pulled out of the if block and done just once before the preflight `if` to DRY this up?

apollo-router/src/plugins/cors.rs
```diff
@@ -0,0 +1,694 @@
...
+                        .unwrap_or_else(|_| http::HeaderValue::from_static("")),
+                );
+            }
+        }
+
+        // Set Vary header
+        response
+            .headers_mut()
+            .insert(VARY, http::HeaderValue::from_static("Origin"));
+    }
```
theJC, line 367: I dont believe we we want this code to wipe out and remove any existing values for the Vary header, especially for non-preflight requests.

https://www.apollographql.com/docs/graphos/routing/cloud/configuration#response-vary-header

Consider a regular GraphQL request (not preflight) case -- a customers plugin (or Apollos code or plugin) could have set several values already in this header to ensure proper caching behavior on the client or proxies between the client and Router.  

This statement would wipe whatever value this header might have after Router completed processing the request and set it to the incoming Origin value.  I dont think this is what we want.

theJC: Additionally (but likely less important) it seems like preflight requests Vary response header might also contain additional headers other than Origin as well:

Ref:
- https://claude.ai/chat/70b6f669-cc2c-43c4-92ce-c23a7b5e3bd1
- https://github.com/tower-rs/tower-http/blob/main/tower-http/src/cors/mod.rs#L812-L820

Ive not yet been able to eyeball this behavior in the wild yet even with the released Router 2.4.0, and I would have expected to see examples of these values in the Vary header in the examples listed in https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CORS  so perhaps this can be kept out of scope of this ticket.

apollo-router/src/plugins/cors.rs
```diff
@@ -0,0 +1,1083 @@
...
+impl<S> CorsService<S> {
+    /// Find the matching policy for a given origin
+    fn find_matching_policy<'a>(
+        config: &'a Cors,
+        origin: &'a Option<http::HeaderValue>,
+    ) -> Option<&'a Policy> {
+        let origin_str = origin.as_ref()?.to_str().ok()?;
+
+        // Security guard: null origins are only allowed when allow_any_origin is true
+        if origin_str == "null" && !config.allow_any_origin {
```
goto-bus-stop, line 173: I've been trying to think of why this might be a thing. `Origin: null` as a literal value doesn't make sense, it's populated by the browser and I don't think that a browser would ever set it to the string value "null". But I see we had pre-existing comments about `Origin: null` in the `allow_any_origin` configuration.

I think when the comments on `allow_any_origin` talk about `Origin: null`, it's probably referring to there not being an Origin header at all? Is there any other interpretation that could make sense?

goto-bus-stop: Apparently, there is also this, but I don't think that's what we meant? https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Access-Control-Allow-Origin#null

apollo-router/src/plugins/cors.rs
```diff
@@ -0,0 +1,1083 @@
...
+        if let Some(existing_vary) = headers.get(VARY) {
+            // Get existing value and append new value
+            if let Ok(existing_str) = existing_vary.to_str() {
+                // Check if the value is already present to avoid duplicates
+                let existing_values: Vec<&str> =
+                    existing_str.split(',').map(|v| v.trim()).collect();
+
+                if !existing_values.contains(&value) {
+                    let new_vary = format!("{}, {}", existing_str, value);
+                    if let Ok(new_header_value) = http::HeaderValue::from_str(&new_vary) {
```
goto-bus-stop: I think we should `.expect()` this and the other from_str below or log an error, since there is no reason it should ever fail.

apollo-router/src/plugins/cors.rs
goto-bus-stop: After thinking about it more I just don't know what we meant. If this maintains existing behaviour for `Origin: null` then we should keep it I guess.

goto-bus-stop (changes requested):
We should make sure all the docs configs are in a format we'd recommend.

Is that:
```yaml
cors:
  policies:
  - origins: [...]
    allow_credentials: true
```
or
```yaml
cors:
  policies:
  - origins: [...]
  allow_credentials: true
```
or
```yaml
cors:
  origins: [...]
  allow_credentials: true
```
?

I feel like the first one is best. The second one is what we have in some places now in the PR and it seems weird to me but please disagree if it's not 🤷🏻‍♀️

apollo-router/src/plugins/cors.rs
theJC: When I read null origins, that actual value of the string`null` was how I had interpreted.  I reserve the right to have interpreted it wrong, and as we tidy up this MR, it seems like we may want to be even more explicit so that there is one way and only one way to interpret.

theJC:
Ok, from a system test it appears that with the changes you made earlier today, you've nailed it!  This PR provides exactly the capabilities we require.  Thank you for making this happen!


| Before (attempting our best with a single policy in Router 2.4.0): | After (with this MR): |
|--------|--------|
| image | image |

apollo-router/src/plugins/cors.rs
```diff
@@ -0,0 +1,1083 @@
...
+                response.headers_mut().insert(
+                    ACCESS_CONTROL_MAX_AGE,
+                    http::HeaderValue::from_str(&max_age_secs.to_string())
+                        .unwrap_or_else(|_| http::HeaderValue::from_static("")),
+                );
+            }
+        }
+
+        // Set Vary header - append to existing values instead of overwriting
+        Self::append_vary_header(response, "Origin");
```
theJC: I just realized I didnt have a `vary` header system test and added a couple.  I ended up getting responses with:  ```Vary: origin, Origin```  when executing against the lastest version of this branch.  

It appears append_vary_header needs to do a case insensitive check for the value already existing

theJC (changes requested):

apollo-router/src/configuration/cors.rs
```diff
@@ -1,28 +1,141 @@
 //! Cross Origin Resource Sharing (CORS configuration)
+//!
+//! This module provides configuration structures for CORS (Cross-Origin Resource Sharing) settings.
+//!
+//! # Default Behavior
```
goto-bus-stop, line 5: Comments here won't show up in our JSON schema. But let's keep it for now, as our schema generation isn't that good with doc comments anyways. I might try to improve it a bit after https://github.com/apollographql/router/pull/7893.

apollo-router/src/plugins/cors.rs
goto-bus-stop: Ah, I missed your earlier comment about `null`! I agree with your interpretation there.

Velfi: Here's what I could dig up with Claude. I included sources wherever possible.

# Apollo Router CORS Handling of Origin: null

## Overview

This document explains how the Apollo Router's custom CORS implementation handles `Origin: null` requests and how this behavior is tested.

## How Origin: null is Handled

### Security Guard Logic

The Apollo Router implements a security-first approach to null origins through a dedicated guard in the policy matching logic (`src/plugins/cors.rs:172-175`):

```rust
// Security guard: null origins are only allowed when allow_any_origin is true
if origin_str == "null" && !config.allow_any_origin {
    return None;
}
```

### Key Behavior

- **Default Behavior**: `Origin: null` is **explicitly rejected** unless `allow_any_origin: true` is configured
- **Explicit Allowance**: When `allow_any_origin: true` is set, null origins are allowed and respond with `Access-Control-Allow-Origin: *`
- **Request Types**: This security guard applies to both preflight (OPTIONS) and actual requests
- **No Selective Support**: You cannot allow null origins while restricting other origins - it's all-or-nothing

### Configuration

The configuration documentation clearly states (`src/configuration/cors.rs:138`):

```rust
/// Set to true to allow any origin. Defaults to false. This is the only way to allow Origin: null.
pub(crate) allow_any_origin: bool,
```

## Security Rationale

### When Null Origins Occur

According to the [MDN Origin header documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Origin), null origins occur in several security-sensitive scenarios:

1. **Non-hierarchical schemes**: Origins whose scheme is not one of `http`, `https`, `ftp`, `ws`, `wss`, or `gopher` 
2. **Local files**: `file://` protocol URLs
3. **Sandboxed documents**: Documents with Content-Security-Policy "sandbox" directive without "allow-same-origin"
4. **Sandboxed iframes**: iframes with sandbox attributes that don't include `allow-same-origin`
5. **Data URLs**: `data:` scheme documents  
6. **Cross-origin redirects**: Documents created through [redirect-tainted origins](https://fetch.spec.whatwg.org/) 
7. **Programmatically created documents**: Documents without a creator browsing context
8. **Network errors**: Responses that are network errors

### Security Risks and Attack Vectors

The [MDN Access-Control-Allow-Origin documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin) explicitly warns against allowing null origins:

> **The value `null` should not be used**: "Any origin can create a hostile document with a `null` origin. Many browsers will grant such documents access to a response with an `Access-Control-Allow-Origin: null` header."

#### Common Attack Pattern (2024)

Security researchers have documented a prevalent attack vector using **sandboxed iframes** to exploit CORS misconfigurations that trust null origins:

```html
<iframe sandbox="allow-scripts allow-top-navigation allow-forms" 
        src='data:text/html,<script>
          // Malicious CORS exploit code here
          // This request will have Origin: null
        </script>'>
</iframe>
```

When an iframe has the `sandbox` attribute without `allow-same-origin`, the browser sets the Origin header to `null` for any requests made from within that iframe. Attackers can exploit this to:

1. **Bypass origin restrictions** if the server accepts `Origin: null`
2. **Steal sensitive data** including API keys, credentials, and PII
3. **Perform authenticated requests** if `Access-Control-Allow-Credentials` is enabled

This attack vector remains a significant concern in 2024, with multiple security researchers documenting real-world cases where applications incorrectly trust null origins.

### Apollo Router's Defense

The Apollo Router's approach prevents these attacks by:

- **Rejecting null origins by default** (secure default)
- **Requiring explicit global allowance** (`allow_any_origin: true`) to accept null origins
- **Preventing accidental null origin trust** through configuration design

## Test Coverage

The implementation includes comprehensive test coverage for null origin handling:

### Core Test Cases

1. **`test_null_origin_rejected_with_catch_all_regex`** (`src/plugins/cors.rs:736`)
   - Ensures null origins are rejected even when there's a catch-all `.*` regex pattern
   - Verifies that regex matching doesn't bypass the null origin security guard

2. **`test_null_origin_rejected_with_specific_regex`** (`src/plugins/cors.rs:759`)
   - Ensures null origins are rejected even with a regex that specifically matches "null" (e.g., `n.ll`)
   - Confirms the security guard takes precedence over regex matching

3. **`test_null_origin_allowed_with_allow_any_origin`** (`src/plugins/cors.rs:782`)
   - Verifies null origins are allowed when `allow_any_origin: true` is configured
   - Confirms the response includes `Access-Control-Allow-Origin: *`

### Preflight Request Tests

4. **`test_null_origin_preflight_request_rejected`** (`src/plugins/cors.rs:848`)
   - Tests that null origins are rejected in preflight (OPTIONS) requests
   - Ensures consistent behavior across request types

5. **`test_null_origin_preflight_allowed_with_allow_any_origin`** (`src/plugins/cors.rs:876`)
   - Tests that null origins are allowed in preflight requests when `allow_any_origin: true`
   - Verifies preflight responses include proper CORS headers

### Integration Tests

6. **`test_regular_origins_still_work_with_null_guard`** (`src/plugins/cors.rs:796`)
   - Ensures normal origin matching (both exact and regex) continues to work after adding the null guard
   - Verifies the security guard doesn't break legitimate origin handling

## Implementation Details

### Policy Matching Process

The null origin check occurs in the `find_matching_policy` function before any other origin matching logic:

1. Extract origin string from request header
2. **Security Guard**: Check if origin is "null" and `allow_any_origin` is false → reject immediately
3. If not rejected, proceed with normal policy matching (exact origins, then regex patterns)

### Response Behavior

- **Rejected**: No `Access-Control-Allow-Origin` header is set (request fails CORS check)
- **Allowed**: `Access-Control-Allow-Origin: *` header is set (along with other configured CORS headers)

### Configuration Examples

#### Default (Secure) Configuration
```yaml
cors:
  allow_any_origin: false  # null origins rejected
  policies:
    - origins: ["https://trusted.com"]
```

#### Allow All Origins (Including Null)
```yaml
cors:
  allow_any_origin: true   # null origins allowed
```

## Comparison with Tower-HTTP

The Apollo Router's custom CORS implementation differs significantly from the standard `tower-http` CORS layer in how it handles null origins.

### Tower-HTTP CORS Implementation

Tower-HTTP provides a flexible `AllowOrigin` configuration through several methods ([source](https://github.com/tower-rs/tower-http/blob/main/tower-http/src/cors/allow_origin.rs)):

```rust
// Single exact origin
let layer = CorsLayer::new().allow_origin(
    "http://example.com".parse::<HeaderValue>().unwrap(),
);

// Multiple origins in a list  
let origins = [
    "http://example.com".parse().unwrap(),
    "null".parse().unwrap(),  // ⚠️ Technically possible but dangerous
];
let layer = CorsLayer::new().allow_origin(origins);

// Wildcard (allows all origins including null)
let layer = CorsLayer::new().allow_origin(AllowOrigin::any());

// Custom predicate function
let layer = CorsLayer::new().allow_origin(AllowOrigin::predicate(
    |origin: &HeaderValue, _request_parts| {
        origin == "null" || origin == "https://trusted.com"
    }
));
```

### Key Differences

| Aspect | Tower-HTTP | Apollo Router |
|--------|------------|---------------|
| **Null Origin Support** | Can explicitly include `"null"` in origin lists via `HeaderValue` | Only via `allow_any_origin: true` |
| **Configuration Flexibility** | Multiple methods: exact, list, predicate, mirror | Policy-based with explicit security guard |
| **Default Security** | Relies on developer awareness | Security-first with null origin guard |
| **Accidental Allowance** | Possible via `AllowOrigin::list()` or predicates | Prevented by design |

### Tower-HTTP Null Origin Vulnerabilities

Tower-HTTP's flexible approach can lead to security issues:

1. **Explicit Null Allowance**: Developers can accidentally include `"null"` in origin lists:
   ```rust
   // ⚠️ DANGEROUS: Explicitly allowing null origins
   let origins = vec![
       "https://trusted.com".parse().unwrap(),
       "null".parse().unwrap(),  // Security vulnerability
   ];
   ```

2. **Predicate-Based Risks**: Custom predicates might inadvertently allow null origins:
   ```rust
   // ⚠️ RISKY: Regex that could match null
   AllowOrigin::predicate(|origin, _| {
       origin.to_str().unwrap().contains("example") || 
       origin == "null"  // Accidental null allowance
   })
   ```

3. **Mirror Request Vulnerability**: `AllowOrigin::mirror_request()` reflects any origin, including null:
   ```rust
   // ⚠️ DANGEROUS: Mirrors null origins back to client
   let layer = CorsLayer::new().allow_origin(AllowOrigin::mirror_request());
   ```

### Apollo Router's Security-First Approach

The Apollo Router prevents these vulnerabilities through:

1. **Explicit Security Guard** (`src/plugins/cors.rs:172-175`):
   ```rust
   // Security guard: null origins are only allowed when allow_any_origin is true
   if origin_str == "null" && !config.allow_any_origin {
       return None;
   }
   ```

2. **No Direct Null Configuration**: You cannot accidentally include `"null"` in origin lists
3. **All-or-Nothing Design**: Null origins are only allowed with global wildcard (`allow_any_origin: true`)
4. **Configuration Validation**: Pre-flight validation prevents dangerous configurations

### Security Trade-offs

**Tower-HTTP Advantages:**
- Maximum flexibility for complex use cases
- Support for custom origin validation logic
- Fine-grained control over individual origins

**Tower-HTTP Disadvantages:**
- Requires security-conscious developers
- Easy to accidentally allow null origins
- No built-in protection against common CORS vulnerabilities

**Apollo Router Advantages:**
- Security by default
- Prevents accidental null origin allowance
- Clear, explicit configuration required for null origins
- Built-in validation and warnings

**Apollo Router Disadvantages:**
- Less flexibility for edge cases
- Cannot selectively allow null origins with specific other origins

## Test Coverage Assessment

The current test suite provides **comprehensive coverage** for null origin handling:

- **Security Guard Logic**: Tests rejection with various policy configurations  
- **Allowance Logic**: Tests acceptance with `allow_any_origin: true`  
- **Request Types**: Covers both preflight and regular requests  
- **Integration**: Ensures null guard doesn't break normal functionality  
- **Edge Cases**: Tests interaction with regex patterns and multiple policies  

The implementation appears to have full test coverage for the security-focused approach to null origin handling.

## Conclusion

The Apollo Router implements a security-first approach to `Origin: null` handling that:

- **Blocks null origins by default** (secure default behavior)
- **Only allows null origins with explicit wildcard configuration** (`allow_any_origin: true`)
- **Prevents common CORS bypass attacks** using null origins
- **Is thoroughly tested** with comprehensive test coverage

This design prioritizes security while still providing a clear path for applications that legitimately need to support null origins.

## References

### Web Standards and Documentation
- [MDN: Origin Header Documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Origin)
- [MDN: Access-Control-Allow-Origin Header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin)
- [MDN: Cross-Origin Resource Sharing (CORS)](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS)
- [WHATWG Fetch Specification](https://fetch.spec.whatwg.org/)

### Apollo Router Source Code
- [Apollo Router CORS Implementation](apollo-router/src/plugins/cors.rs)
- [Apollo Router CORS Configuration](apollo-router/src/configuration/cors.rs)

### Tower-HTTP Source Code
- [Tower-HTTP AllowOrigin Implementation](https://github.com/tower-rs/tower-http/blob/main/tower-http/src/cors/allow_origin.rs)
- [Tower-HTTP CORS Documentation](https://docs.rs/tower-http/latest/tower_http/cors/index.html)
- [Tower-HTTP AllowOrigin API](https://docs.rs/tower-http/latest/tower_http/cors/struct.AllowOrigin.html)

apollo-router/src/plugins/cors.rs
theJC: Verified fixed with latest updates to this PR.  All of our system tests are green, including the our newly added Vary tests.


image

apollo-router/src/configuration/cors.rs
```diff
@@ -1,28 +1,141 @@
...
+//!
+//! When specifying individual policies within the `policies` array:
+//! - **Origins:** Defaults to an empty list (no origins allowed) unless explicitly set
+//! - **Match origins:** Defaults to an empty list (no regex matching) unless explicitly set
+//! - **Methods:** Has three possible states:
+//!   - `null` (not specified): Use the global default methods
+//!   - `[]` (empty array): No methods allowed for this policy
+//!   - `[values]` (with values): Use these specific methods
+//! - **Allow headers:** Defaults to an empty list (mirrors client headers) unless explicitly set
+//! - **Expose headers:** Defaults to an empty list unless explicitly set
```
theJC: If a global value is set for either of these two, then they will default to that global value, right?  If so, this should be elaborated on here.

apollo-router/src/plugins/cors.rs
theJC: I believe both of these concerns have now been addressed by updates to this PR.  Thank you.

.changesets/feat_zelda_cors_configuration_updates.md
```diff
@@ -0,0 +1,69 @@
...
+      allow_headers: ["content-type", "authorization"]
+```
+
+#### Default Configuration
+
+The default CORS configuration remains unchanged for backward compatibility:
+- Default origins: `https://studio.apollographql.com`
+- Default methods: `GET`, `POST`, `OPTIONS`
+- Default `allow_credentials`: `false`
+- Default `allow_headers`: empty (mirrors client's `Access-Control-Request-Headers`)
```
goto-bus-stop: Narrowed down to user impact.

```suggestion
### Introduce per-origin CORS policies ([PR #7853](https://github.com/apollographql/router/pull/7853))

Configuration can now specify different Cross-Origin Resource Sharing (CORS) rules for different origins using the `cors.policies` key. See the [CORS documentation](https://www.apollographql.com/docs/graphos/routing/security/cors) for details.

```yaml
cors:
  policies:
    # The default CORS options work for Studio.
    - origins: ["https://studio.apollographql.com"]
    # Specific config for trusted origins
    - match_origins: ["^https://(dev|staging|www)?\\.my-app\\.(com|fr|tn)$"]
      allow_credentials: true
      allow_headers: ["content-type", "authorization", "x-web-version"]
    # Catch-all for untrusted origins
    - origins: ["*"]
      allow_credentials: false
      allow_headers: ["content-type"]
```

apollo-bot2:

apollo-router/src/configuration/cors.rs
BrynCooke: [ce968cb](https://github.com/apollographql/router/pull/7853/commits/ce968cb524da0d7e059af7b23914747c3cd6074a)

BrynCooke (approved):

apollo-router/src/configuration/cors.rs
abernix: @bryncooke addressed this in https://github.com/apollographql/router/pull/7853/commits/ce968cb524da0d7e059af7b23914747c3cd6074a

apollo-router/src/configuration/cors.rs
abernix: this was addressed! marking comment as resolved.

docs/source/routing/cloud/configuration.mdx
```diff
@@ -234,43 +234,45 @@ For a more general introduction to CORS and common considerations, see the follo
...
+        - https://www.your-app.example.com
+        - https://studio.apollographql.com # Keep this so GraphOS Studio can run queries against your router
+    - origins:
+        - https://api.example.com
+      match_origins:
+        - 'https://([a-z0-9]+[.])*api[.]example[.]com' # any host that uses https and ends with .api.example.com
 ```
 
-You can also turn off CORS entirely by setting `origins` to an empty list:
+You can also disable CORS entirely by setting `policies` to an empty list:
```
phryneas, line 271: I would suggest a different wording here - both "turn off" and "disable" might cause inexperienced devs to think "if I do this, I don't have to deal with CORS anymore", where in reality, CORS itself cannot be disabled as it's a browser security feature.


```suggestion
You can also forbid CORS requests entirely by setting `policies` to an empty list:
```

apollo-router/src/configuration/cors.rs
```diff
@@ -1,28 +1,159 @@
...
+//!   - not specified: Use the global default allow_credentials
+//!   - `true`: Enable credentials for this policy
+//!   - `false`: Disable credentials for this policy
+//! - **Allow headers:** Inherits global headers if empty, otherwise uses policy-specific headers
+//! - **Expose headers:** Inherits global headers if empty, otherwise uses policy-specific headers
+//! - **Methods:** Has three possible states:
+//!   - not specified: Use the global default methods
+//!   - `[]` (empty array): No methods allowed for this policy
+//!   - `[values]` (with values): Use these specific methods
+//! - **Max age:** Has three possible states:
```
theJC, line 29: there appears to be only `two` possible states listed here (which appear to be the correct number of states)

apollo-router/src/configuration/cors.rs
```diff
@@ -1,28 +1,159 @@
...
+//! When specifying individual policies within the `policies` array:
+//! - **Origins:** Defaults to `["https://studio.apollographql.com"]` when no policies are specified at all.
+//!   If you specify any policies, you must explicitly set origins for each policy.
+//! - **Match origins:** Defaults to an empty list (no regex matching) unless explicitly set
+//! - **Allow credentials:** Has three possible states:
+//!   - not specified: Use the global default allow_credentials
+//!   - `true`: Enable credentials for this policy
+//!   - `false`: Disable credentials for this policy
+//! - **Allow headers:** Inherits global headers if empty, otherwise uses policy-specific headers
+//! - **Expose headers:** Inherits global headers if empty, otherwise uses policy-specific headers
```
theJC, line 24: For these two (allow headers and expose headers), from the perspective of a customer crafting their config.yaml.... I think there are the same 3 states that can be done in the yaml file.  
- not specified - use the global setting
- empty array - use the global setting
- an array with policy specific headers - use these policy-specific behavior

PR #7858, opened 2025-07-09
chore: use multer to test subscription responses (backport #7817)

goto-bus-stop:
Follow-up to #7731.

Use the multer crate for parsing subscription `multipart/mixed` streams
instead of a hand-rolled implementation. multer technically is for
`multipart/form-data`, which is sort of the successor format to
`multipart/mixed` if i understand the RFCs correctly. But the format is
essentially identical.

This implementation is mostly stricter than before. These things now
cause test failures:
- Final `--graphql--` boundaries in the middle of the stream
- Omitting the final `--graphql--` boundary
- Not draining the entire subscription
- Non-JSON response chunks

But these benign things will not:
- Having multiple headers on a chunk (we don't do this, but it sounds
  valid)
- A chunk boundary occurring at the edge of an HTTP frame from hyper<hr>This is an automatic backport of pull request #7817 done by [Mergify](https://mergify.com).


goto-bus-stop:
Will redo after https://github.com/apollographql/router/pull/7836

PR #7859, opened 2025-07-09
feat(response_cache): add support for cache-control header in request

bnjjj:
If a client doesn't want to interact with the cache they can set the header `cache-control` to `no-store` and it won't use response cache plugin. This is an opt-in feature from the configuration because the router admin will have to propagate the header from supergraph request to subgraph request. Without this configuration nothing will happen if the cache-control is set to anything.

---

**Checklist**

- Tests added and passing[^4]
    - [ ] Unit tests
    - [ ] Integration tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/plugins/response_cache/plugin.rs
+++ b/apollo-router/src/plugins/response_cache/plugin.rs
@@ -651,6 +651,37 @@ impl CacheService {
         {
             return self.service.call(request).await;
         }
+        // Don't use cache at all if no-store is set in cache-control header
+        if request
+            .subgraph_request
+            .headers()
+            .get_all(&CACHE_CONTROL)
+            .iter()
+            .next()
+            .is_some()
+        {
+            let cache_control = match CacheControl::new(request.subgraph_request.headers(), None) {
+                Ok(cache_control) => cache_control,
+                Err(err) => {
+                    return Ok(subgraph::Response::builder()
+                        .subgraph_name(request.subgraph_name)
+                        .context(request.context)
+                        .error(
+                            graphql::Error::builder()
+                                .message(format!("cannot get cache-control header: {err}"))
+                                .extension_code("INVALID_CACHE_CONTROL_HEADER")
+                                .build(),
+                        )
+                        .extensions(Object::default())
+                        .build());
+                }
+            };
+            if cache_control.no_store {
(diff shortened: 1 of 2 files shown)


apollo-router/src/plugins/response_cache/tests.rs
```diff
@@ -1152,6 +1152,172 @@ async fn no_cache_control() {
...
+        password: None,
+        timeout: Some(std::time::Duration::from_secs(5)),
+        required_to_start: true,
+        pool_size: default_pool_size(),
+        batch_size: default_batch_size(),
+        namespace: Some(String::from("test_no_store_from_client")),
+    })
+    .await
+    .unwrap();
+    let response_cache = ResponseCache::for_test(
```
carodewig, line 1198: Would it be possible to test that the response cache is empty after the first request? As far as I can tell, this test is just verifying that the cache-control no-store directive is propagated, not that the cache wasn't used.

carodewig (commented):
Is there a plan to add support for cache-control header in the subgraph response in a separate PR? I suspect there will be times a subgraph should set no-store in the response regardless of of the request

apollo-router/src/plugins/response_cache/plugin.rs
```diff
@@ -651,6 +651,37 @@ impl CacheService {
...
             return self.service.call(request).await;
         }
+        // Don't use cache at all if no-store is set in cache-control header
+        if request
+            .subgraph_request
+            .headers()
+            .get_all(&CACHE_CONTROL)
+            .iter()
+            .next()
+            .is_some()
```
SimonSapin:
```suggestion
            .headers()
            .contains_key(&CACHE_CONTROL)
```

SimonSapin (approved):
Approved with one change

bnjjj:
@carodewig I'm not sure to understand but it's already supported for subgraph response. We merge different cache control together and we always take cache-control header from subgraph response

apollo-router/src/plugins/response_cache/tests.rs
bnjjj: Fixed

carodewig (approved):

PR #7870, opened 2025-07-10
chore(deps): configure Renovate to handle exact version pins separately

abernix:
Cargo dependencies pinned with an exact version (using '=' prefix) will now:
- Be handled in individual PRs (not grouped with other updates)
- Require dashboard approval before a PR is opened
- Never be automerged

This fixes an issue where pinned dependencies like 'rhai' were still being
included in dependency update PRs despite being explicitly pinned for stability
reasons (example: https://github.com/apollographql/router/pull/7863).

### DIFF
--- a/renovate.json5
+++ b/renovate.json5
@@ -112,5 +112,16 @@
       matchManagers: ['cargo'],
       matchPackageNames: ['/^axum$/', '/^axum-extra$/'],
     },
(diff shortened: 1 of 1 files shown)


goto-bus-stop (approved):
seems good

abernix:
### After

image
image

PR #7879, opened 2025-07-11
Fix the subscription duplication problem when the client terminates the original subscription.

bnjjj:
Address a regression in subscription deduplication introduced in #5505. The client connection close event does not need handling in this task because it is managed elsewhere. Additionally, part of #5505 resolved the issue related to closing the websocket connection. This PR also includes tests to ensure no future regressions related to subscription deduplication and websocket close events.

---

**Checklist**

- Tests added and passing[^4]
    - [x] Unit tests
    - [x] Integration tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-1371]: https://apollographql.atlassian.net/browse/ROUTER-1371?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apollo-router/src/services/subgraph_service.rs
+++ b/apollo-router/src/services/subgraph_service.rs
@@ -686,11 +686,17 @@ async fn call_websocket(
 
     let (handle_sink, handle_stream) = handle.split();
 
+    // Forward GraphQL subscription stream to WebSocket handle
+    // Connection lifecycle is managed by the WebSocket infrastructure,
+    // so we don't need to handle connection_closed_signal here
     tokio::task::spawn(async move {
-        let _ = gql_stream
+        if let Err(e) = gql_stream
             .map(Ok::<_, graphql::Error>)
             .forward(handle_sink)
-            .await;
+            .await
+        {
+            tracing::debug!("WebSocket subscription stream ended: {}", e);
+        }
     });
 
     subscription_stream_tx.send(Box::pin(handle_stream)).await?;


apollo-router/tests/integration/subscriptions/mod.rs
```diff
@@ -397,6 +408,7 @@ async fn handle_websocket(mut socket: WebSocket, config: SubscriptionServerConfi
                 _ => {}
             }
         }
+        is_closed.store(true, std::sync::atomic::Ordering::Relaxed);
```
BrynCooke: There is a return statement in the code above which may cause this line not to be hit. Is this OK? 
```
if socket
    .send(axum::extract::ws::Message::text(
        serde_json::to_string(&complete).unwrap(),
    ))
    .await
    .is_err()
{
    return;
}
```

bnjjj: Good catch. It's fixed

.changesets/fix_bnjjj_rh_924.md
```diff
@@ -0,0 +1,5 @@
+### Fix the subscription duplication problem when the client terminates the original subscription. ([PR #7879](https://github.com/apollographql/router/pull/7879))
+
+Address a regression in subscription deduplication introduced in #5505. The client connection close event does not need handling in this task because it is managed elsewhere. Additionally, part of #5505 resolved the issue related to closing the websocket connection. This PR also includes tests to ensure no future regressions related to subscription deduplication and websocket close events.
```
goto-bus-stop: User impact, please make sure i represented it fairly though
```suggestion
### Fix deduplicated subscriptions hanging when one subscription closes ([PR #7879](https://github.com/apollographql/router/pull/7879))

Fixes a regression introduced in v1.50.0. When multiple client subscriptions are deduped onto a single subgraph subscription in WebSocket passthrough mode, and the first client subscription closes, the Router would close the subgraph subscription. The other deduplicated subscriptions would then silently stop receiving events.

Now outgoing subscriptions to subgraphs are kept open as long as _any_ client subscription uses them.
```

apollo-router/src/services/subgraph_service.rs
```diff
@@ -689,26 +687,10 @@ async fn call_websocket(
     let (handle_sink, handle_stream) = handle.split();
 
     tokio::task::spawn(async move {
```
BrynCooke, line 689: This block could be something like:

```
 // Forward GraphQL subscription stream to WebSocket handle
    // Connection lifecycle is managed by the WebSocket infrastructure,
    // so we don't need to handle connection_closed_signal here
    let task_name = format!("ws-subscription-{}", service_name);
    if let Err(e) = tokio::task::Builder::new()
        .name(&task_name)
        .spawn(async move {
            if let Err(e) = gql_stream
                .map(Ok::<_, graphql::Error>)
                .forward(handle_sink)
                .await
            {
                tracing::debug!("WebSocket subscription stream ended for {}: {}", service_name, e);
            }
        })
        // If tokio can't spawn tasks we should kill the router.
        .expect("Failed to spawn WebSocket subscription forwarding task");
```

PR #7882, opened 2025-07-11
add ResponseErrors selector to router response

Aguilarjaf:
Introducing the `ResponseErrors` selector in telemetry configurations to capture router response errors, allowing users to capture and log errors encountered at the router service layer. This new selector enhances logging for the router service, as it allows users the option to only log router errors instead of the entire router response body to reduce noise.

``` yaml
telemetry:
  instrumentation:
    events:
      router:
         router.error:
            attributes:
               "my_attribute":
                   response_errors: "$.[0]"
                   # Examples: "$.[0].message", "$.[0].locations", "$.[0].extensions", etc.
```

---
**Checklist**

- Tests added and passing[^4]
    - [x] Unit tests
    - [ ] Integration tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/plugins/telemetry/config_new/router/selectors.rs
```diff
@@ -216,7 +226,7 @@ impl Selector for RouterSelector {
             } => get_baggage(baggage).or_else(|| default.maybe_to_otel_value()),
             RouterSelector::Static(val) => Some(val.clone().into()),
             RouterSelector::StaticField { r#static } => Some(r#static.clone().into()),
-            RouterSelector::ResponseBody { response_body } if *response_body => {
+            RouterSelector::ResponseBody { response_body, truncate: _truncate } if *response_body => {
```
Velfi: Since we don't use `truncate` here, this should be fine:
```suggestion
            RouterSelector::ResponseBody { response_body, .. } if *response_body => {
```

goto-bus-stop:
Elsewhere we use JSON Path to select subsets of data from the `response_data` and `response_errors` selectors:
https://www.apollographql.com/docs/graphos/routing/observability/telemetry/instrumentation/selectors#setting-json-paths

I think it makes sense to have a feature like this in `response_body`, but I think it should also use JSON Path to match the existing functionality.
(You might also already be able to achieve what you need with those other selectors!)

Aguilarjaf:
Hi @goto-bus-stop, thanks for the feedback! 

I have updated the PR to better align with the best practice of using JSON Path to select a subset of data from the router response. I have implemented an additional selector, `ResponseErrors`, that takes a `JsonPathInst` object to extract specific information from router response errors (message, extensions, etc.). This is functionality currently does not exist in the router service response, so providing a new selector such as `ResponseErrors` can help users better capture and log those router level errors.

Velfi:
@mergifyio copy dev

abernix:
We'll close this in lieu of #8024, where you're welcome to help contribute to the conversation in getting it over the line but will no longer have branch push perms.  Thanks much for getting this opened and started originally — you'll still get contribution credit via that other PR when it merges, but we'll be able to run this with the full CI from that PR via @Velfi's account.  Thanks again to you both!

abernix:
This is shipping Router v2.7.0 — which has its release candidate up now as https://github.com/apollographql/router/releases/tag/v2.7.0-rc.0.  It should be released officially in the next couple days.

Thanks again for the original contribution!

PR #7883, opened 2025-07-11
feat(federation/composition): port `ValidationTraversal`, `validateGraphComposition()`, and `validateSatisfiability()`

sachindshinde:
This PR ports `ValidationTraversal` from the JS codebase, along with the top-level functions `validateGraphComposition()` and `validateSatisfiability()`. Note that:
1. While I was looking at the code for `ValidationContext` to call it, I noticed some bugs in the relevant functions and fixed them accordingly.
2. I moved `ValidationContext` to its own sub-module under `satisfiability/validation_context.rs`.
3. `validate_satisfiability()` does not take in `CompositionOptions`, but in the future, it should. (Looks like we haven't threaded it through yet in upstream functions.) I've left a `TODO` around this.
4. For some reason, `Supergraph<Satisfiable>` has supergraph metadata in it that isn't actually computed by the satisfiability algorithm (nor is it really metadata that only becomes available post-satisfiability). For now, I've just left this metadata blank, but at some point we should check who needs this metadata downstream, and determine how best to provide it (I've similarly left a `TODO` here). 
6. We did not port any fixes from https://github.com/apollographql/federation/pull/3277 yet to this code.


duckki:
Looks great!

Since Sachin is away, I pushed some minor changes:
* fixed `overrride_condition.check(...)` to be negated.
* renamed `TopLevelConditionResolverCache` to just `TopLevelConditionResolver`, since it's just a resolver (holding a cache internally) and aligns better with JS code as a bonus.
* improved debug log formatting.
* added a unit test for `validate_satisfiability` and a `satisfiability` CLI command for manual testing.

The unit test is currently failing. But, I don't think we should block merging this PR. We can revisit this after porting other missing pieces.

duckki (approved):

PR #7884, opened 2025-07-11
Promote connectors ->get method to public

andrewmcgivery:
Promote connectors ->get method to public.

Note that Shape method had to be completely rewritten from scratch as it was quite out of date.

The runtime and runtime tests are existing code.

---

**Checklist**

- Tests added and passing[^4]
    - [x] Unit tests
    - [ ] Integration tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-federation/src/connectors/json_selection/methods/public/get.rs
```diff
@@ -0,0 +1,1212 @@
...
+            }
+
+            (JSON::String(s), Some(i)) => {
+                let s_str = s.as_str();
+                let ilen = s_str.len() as i64;
+                // Negative indices count from the end of the array
+                let index = if i < 0 { ilen + i } else { i };
+                if index >= 0 && index < ilen {
+                    let uindex = index as usize;
+                    let single_char_string = s_str[uindex..uindex + 1].to_string();
```
dylan-apollo: We should use a similar `.get()` as arrays instead of slicing, since slicing can panic if we get something wrong up above.

apollo-federation/src/connectors/json_selection/methods/public/get.rs
```diff
@@ -0,0 +1,1212 @@
...
+                    ApplyToError::new(
+                        format!(
+                            "Method ->{} requires an integer index",
+                            method_name.as_ref()
+                        ),
+                        input_path.to_vec(),
+                        index_literal.range(),
+                    ),
+                ),
+            ),
```
dylan-apollo: This block is confusing to me... I _think_ the issue is that we got a float? Maybe we can make the error more explicit.

apollo-federation/src/connectors/json_selection/methods/public/get.rs
```diff
@@ -0,0 +1,1212 @@
...
+            _ => (
+                None,
+                vec_push(
+                    index_errors,
+                    ApplyToError::new(
+                        format!(
+                            "Method ->{} requires an array or string input, not {}",
+                            method_name.as_ref(),
+                            json_type_name(data),
```
dylan-apollo: This is also confusing, because an object is perfectly fine, it's just that an object _plus_ and integer index doesn't work. Maybe flipping the match nesting would make this flow better? First check what the input shape is, error if it's not one that can be indexed. Then check the index type and error if _that_ doesn't match the expected type for the input shape?

apollo-federation/src/connectors/json_selection/methods/public/get.rs
```diff
@@ -0,0 +1,1212 @@
...
+                        ),
+                        input_path.to_vec(),
+                        method_name.range(),
+                    ),
+                ),
+            ),
+        },
+        (Some(ref key @ JSON::String(ref s)), index_errors) => match data {
```
dylan-apollo:
```suggestion
        (Some(JSON::String(key)), index_errors) => match data {
```

No need for all the reffing, you will just need to say `key.as_str()` later in the errors I think.

apollo-federation/src/connectors/json_selection/methods/public/get.rs
```diff
@@ -0,0 +1,1212 @@
...
+                    )
+                }
+            }
+            _ => (
+                None,
+                vec_push(
+                    index_errors,
+                    ApplyToError::new(
+                        format!(
+                            "Method ->{}({}) requires an object input",
```
dylan-apollo: Again, this error isn't quite true. `->get` can take non-object inputs, it's _really_ that `->get` on an object requires a string key.

apollo-federation/src/connectors/json_selection/methods/public/get.rs
```diff
@@ -0,0 +1,1212 @@
...
+                ),
+            ),
+        },
+        (Some(value), index_errors) => (
+            None,
+            vec_push(
+                index_errors,
+                ApplyToError::new(
+                    format!(
+                        "Method ->{}({}) requires an integer or string argument",
```
dylan-apollo: We can then make this more specific, too, and only suggest the index that's appropriate for the provided input type.

apollo-federation/src/connectors/json_selection/methods/public/get.rs
```diff
@@ -0,0 +1,1212 @@
...
+                ),
+            ),
+        ),
+        (None, index_errors) => (
+            None,
+            vec_push(
+                index_errors,
+                ApplyToError::new(
+                    format!(
+                        "Method ->{} received undefined argument",
```
dylan-apollo: Doesn't this really just mean that there were `index_errors` and we should return those?

apollo-federation/src/connectors/json_selection/methods/public/get.rs
```diff
@@ -0,0 +1,1212 @@
...
+    let index_shape = index_literal.compute_output_shape(
+        input_shape.clone(),
+        dollar_shape,
+        named_var_shapes,
+        source_id,
+    );
+
+    if Shape::string([]).accepts(&input_shape) {
+        handle_string_shape(method_name, &input_shape, &index_shape, source_id)
+    } else if Shape::tuple([], []).accepts(&input_shape) {
```
dylan-apollo, line 341: I wouldn't expect this to be true, but I guess anything you can do to a completely empty list you can also do to a list containing items?

apollo-federation/src/connectors/json_selection/methods/public/get.rs
```diff
@@ -0,0 +1,1212 @@
...
+
+    let ShapeCase::Array { prefix, .. } = input_shape.case() else {
+        return Shape::one(
+            [
+                Shape::unknown(method_name.shape_location(source_id)),
+                Shape::none(),
+            ],
+            method_name.shape_location(source_id),
+        );
+    };
```
dylan-apollo: This is impossible, right? We just don't have an easy way to represent a concrete shape?

apollo-federation/src/connectors/json_selection/methods/public/get.rs
```diff
@@ -0,0 +1,1212 @@
...
+
+    #[test]
+    fn get_shape_should_error_for_string_with_negative_out_of_bounds_index() {
+        assert_eq!(
+            get_test_shape(
+                vec![WithRange::new(LitExpr::Number(Number::from(-10)), None)],
+                Shape::string_value("hello", [])
+            ),
+            Shape::error(
+                "Method ->get index -5 out of bounds in string of length 5".to_string(),
```
dylan-apollo: Should be `-10`, right?

apollo-federation/src/connectors/json_selection/methods/public/get.rs
```diff
@@ -0,0 +1,1212 @@
...
+                        .into_with_range(),
+                    }),
+                    None
+                )],
+                Shape::string([])
+            ),
+            Shape::one(
+                [Shape::string([get_location()]), Shape::none()],
+                [get_location()]
+            )
```
dylan-apollo: I don't think we should return One<> with None here, since that's really indicating that depending on the argument, it could be an error, not None.

apollo-federation/src/connectors/json_selection/methods/public/get.rs
```diff
@@ -0,0 +1,1212 @@
...
+            Shape::error(
+                "Method ->get must be provided an integer argument when applied to an array"
+                    .to_string(),
+                [get_location()]
+            )
+        );
+    }
+
+    #[test]
+    // TODO: Debug here
```
dylan-apollo: ?

apollo-federation/src/connectors/json_selection/methods/public/get.rs
```diff
@@ -0,0 +1,1212 @@
...
+                        .into_with_range(),
+                    }),
+                    None
+                )],
+                Shape::array([Shape::int([])], Shape::none(), [])
+            ),
+            Shape::one(
+                [Shape::unknown([get_location()]), Shape::none()],
+                [get_location()]
+            )
```
dylan-apollo: Same note, we don't need to One<> these to account for runtime errors, since I'm pretty sure we don't anywhere else.

apollo-federation/src/connectors/json_selection/methods/public/get.rs
```diff
@@ -0,0 +1,1212 @@
...
+        );
+    }
+
+    #[test]
+    fn get_shape_should_error_for_array_with_string_index() {
+        assert_eq!(
+            get_test_shape(
+                vec![WithRange::new(LitExpr::String("invalid".to_string()), None)],
+                Shape::array([Shape::int([])], Shape::none(), [])
```
dylan-apollo, line 1023: Rather than using `Shape::array` for these tests, they should each _either_ use `Shape::tuple` (fixed, known size array, which in practice can't happen today for our users) or `Shape::list` (dynamic array, unknown size, what we'll see today).

This is especially important since I _think_ `Shape::list` cases are gonna improperly error right now, since they don't have prefix.

apollo-federation/src/connectors/json_selection/methods/public/get.rs
```diff
@@ -0,0 +1,1212 @@
...
+    #[test]
+    fn get_shape_should_return_property_shape_for_object_with_valid_string_key() {
+        let mut fields = IndexMap::default();
+        fields.insert("key".to_string(), Shape::int([]));
+        fields.insert("other".to_string(), Shape::string([]));
+
+        assert_eq!(
+            get_test_shape(
+                vec![WithRange::new(LitExpr::String("key".to_string()), None)],
+                Shape::object(fields.into_iter().collect(), Shape::none(), [])
```
dylan-apollo:
```suggestion
                Shape::object(fields), Shape::none(), [])
```

You won't need these if you `use indexmap::IndexMap` at the top of this test module. The trouble is it's trying to use the `apollo_compiler` special version of `IndexMap`.

apollo-federation/src/connectors/json_selection/methods/public/get.rs
```diff
@@ -0,0 +1,1212 @@
...
+
+    #[test]
+    fn get_shape_should_return_none_for_object_with_missing_key() {
+        let mut fields = IndexMap::default();
+        fields.insert("existing".to_string(), Shape::int([]));
+
+        assert_eq!(
+            get_test_shape(
+                vec![WithRange::new(LitExpr::String("missing".to_string()), None)],
+                Shape::object(fields.into_iter().collect(), Shape::none(), [])
```
dylan-apollo: Similar to the `Shape::array` thing, what you're creating in all of these is a `Shape::record`, which has all known fields. These are like GraphQL types, which are the objects we'll usually have. We should _also_ add a test or two for a `Shape::dict` which is an object where we don't know the set of keys in it, but we _might_ know the shape of the values. In which case, we should return the shape of the values.

An example of that is `$request.headers` which is `Shape::dict(Shape::list(Shape::string([]), []), [])`

apollo-federation/src/connectors/json_selection/methods/public/get.rs
```diff
@@ -0,0 +1,1212 @@
...
+                            Key::field("a").into_with_range(),
+                            PathList::Empty.into_with_range(),
+                        )
+                        .into_with_range(),
+                    }),
+                    None
+                )],
+                Shape::object(fields.into_iter().collect(), Shape::none(), [])
+            ),
+            Shape::unknown([get_location()])
```
dylan-apollo: I think this should still be based on the object shape. Something like `input_shape.any_field()`? So if it's an object where all the values are strings, and we don't know which key is being selected (dynamic), we can still know the output is a string.

apollo-federation/src/connectors/json_selection/methods/public/get.rs
```diff
@@ -0,0 +1,1212 @@
...
+                vec![WithRange::new(LitExpr::Number(Number::from(0)), None)],
+                Shape::unknown([])
+            ),
+            Shape::one(
+                [
+                    Shape::string([get_location()]),
+                    Shape::unknown([get_location()]),
+                ],
+                [get_location()]
+            )
```
dylan-apollo: I'm not sure if it's worth adding `string` as a concrete case? Since `unknown` kinda envelops it.

apollo-federation/src/connectors/json_selection/methods/public/get.rs
dylan-apollo: Oh and also we still know the element type of the array, right? So it shouldn't be hard-coded to `unknown`, it should be `input_shape.any_item()`

dylan-apollo (changes requested):

apollo-federation/src/connectors/json_selection/methods/public/get.rs
andrewmcgivery: I chatted with @benjamn about this and because Shape lacks an "empty_array" Shape, this empty tuple is the next best thing.

apollo-federation/src/connectors/json_selection/methods/public/get.rs
andrewmcgivery: This should be addressed in the rewrite I just pushed!

apollo-federation/src/connectors/json_selection/methods/public/get.rs
andrewmcgivery: This should be addressed in the rewrite I just pushed!

apollo-federation/src/connectors/json_selection/methods/public/get.rs
andrewmcgivery: This should be addressed in the rewrite I just pushed!

apollo-federation/src/connectors/json_selection/methods/public/get.rs
andrewmcgivery: This should be addressed in the rewrite I just pushed!

apollo-federation/src/connectors/json_selection/methods/public/get.rs
andrewmcgivery: This should be addressed in the rewrite I just pushed!

apollo-federation/src/connectors/json_selection/methods/public/get.rs
andrewmcgivery: This should be addressed in the rewrite I just pushed!

apollo-federation/src/connectors/json_selection/methods/public/get.rs
andrewmcgivery: This should be addressed in the rewrite I just pushed!

apollo-federation/src/connectors/json_selection/methods/public/get.rs
andrewmcgivery: Yea... in theory I don't think this should ever happen but I had to use the `case()` to be able to get the value 😅

apollo-federation/src/connectors/json_selection/methods/public/get.rs
andrewmcgivery: OH yea, good catch!

apollo-federation/src/connectors/json_selection/methods/public/get.rs
andrewmcgivery: I'm not sure I understand.

The test here is that the input type is known (String) but the index is unknown... which means the index might be valid... or it might not be... if its valid, we'd return string... if it's not valid, we'd return None (with mapping errors).

apollo-federation/src/connectors/json_selection/methods/public/get.rs
andrewmcgivery: Oops 😅

apollo-federation/src/connectors/json_selection/methods/public/get.rs
andrewmcgivery: Good call on the any_item!

apollo-federation/src/connectors/json_selection/methods/public/get.rs
andrewmcgivery: Huh... that does indeed error and I'm 10/10 confused how to handle this because an array with 0 items seems to look the exact same as a list with unknown number of items? Am I missing something?

apollo-federation/src/connectors/json_selection/methods/public/get.rs
andrewmcgivery: I suspect this is going to have the same problem where we can't tell the difference between an object that legit does not have a property vs a object that we don't know what properties it has 😕

apollo-federation/src/connectors/json_selection/methods/public/get.rs
andrewmcgivery: Yea, this should just be unknown. I forgot to remove string from this test after I updated the method to only return unknown... and apparently even though I don't actually have this coming back from my code this test still passes 😅

apollo-federation/src/connectors/json_selection/methods/public/get.rs
andrewmcgivery: Oh good call, updated!

apollo-federation/src/connectors/json_selection/methods/public/get.rs
andrewmcgivery: Okay this should be fixed!

apollo-federation/src/connectors/json_selection/methods/public/get.rs
andrewmcgivery: Okay this should be fixed!

apollo-federation/src/connectors/json_selection/methods/public/get.rs
```diff
@@ -0,0 +1,1354 @@
...
+                "Method ->{} must be provided an integer argument when applied to a string",
+                method_name.as_ref()
+            )
+            .as_str(),
+            method_name.shape_location(source_id),
+        );
+    };
+
+    let ShapeCase::String(Some(input_value)) = input_shape.case() else {
+        return Shape::string(method_name.shape_location(source_id));
```
dylan-apollo, line 385: I think this one is `Shape::one()` as well, since we can't bounds check?

apollo-federation/src/connectors/json_selection/methods/public/get.rs
```diff
@@ -0,0 +1,1354 @@
...
+                input_shape.any_field(method_name.shape_location(source_id)),
+                Shape::none(),
+            ],
+            method_name.shape_location(source_id),
+        );
+    };
+
+    // If we have a rest, we cannot know for sure if the item exists at the index or not
+    // This is because a rest implies that there are 0 to many items of that type
+    if !rest.is_none() {
```
dylan-apollo: I don't think this is accurate. Pretty sure what we want is if the item is in the fields, return that, else if rest, return One<Rest, None>, else return None

apollo-federation/src/connectors/json_selection/methods/public/get.rs
andrewmcgivery: Good call

apollo-federation/src/connectors/json_selection/methods/public/get.rs
andrewmcgivery: Good call. I fixed this for the array tail logic as well.

apollo-federation/src/connectors/json_selection/methods/public/get.rs
```diff
@@ -0,0 +1,1269 @@
...
+                prefix.len()
+            )
+            .as_str(),
+            method_name.shape_location(source_id),
+        )
+    };
+
+    if let Some(item) = get_array(*index_value, prefix) {
+        item.clone()
+    } else if !tail.is_none() {
```
dylan-apollo, line 450:
```suggestion
    } else if tail.is_some() {
```

apollo-federation/src/connectors/json_selection/methods/public/get.rs
```diff
@@ -0,0 +1,1269 @@
...
+        );
+    };
+
+    let ShapeCase::Object { fields, rest } = input_shape.case() else {
+        return input_shape.any_field(method_name.shape_location(source_id));
+    };
+
+    if let Some(item) = fields.get(index_value) {
+        item.clone()
+    } else if !rest.is_none() {
```
dylan-apollo, line 490:
```suggestion
    } else if rest.is_some() {
```

apollo-federation/src/connectors/json_selection/methods/public/get.rs
andrewmcgivery: is_some doesn't actually exist on the Shape library 😅

apollo-federation/src/connectors/json_selection/methods/public/get.rs
andrewmcgivery: is_some doesn't actually exist on the Shape library 😅

PR #7888, opened 2025-07-14
Implement toString connectors method

andrewmcgivery:
---

**Checklist**

- Tests added and passing[^4]
    - [x] Unit tests
    - [ ] Integration tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-federation/src/connectors/json_selection/helpers.rs
+++ b/apollo-federation/src/connectors/json_selection/helpers.rs
@@ -79,6 +79,18 @@ pub(crate) const fn json_type_name(v: &JSON) -> &str {
     }
 }
 
+pub(crate) fn json_to_string(json: &JSON) -> Result<Option<String>, String> {
+    match json {
+        JSON::Null => Ok(None),
+        JSON::Bool(b) => Ok(Some(b.to_string())),
+        JSON::Number(n) => Ok(Some(n.to_string())),
+        JSON::String(s) => Ok(Some(s.as_str().to_string())),
+        JSON::Array(_) | JSON::Object(_) => {
+            Err("cannot convert arrays or objects to strings.".to_string())
+        }
+    }
+}
+
 pub(crate) fn vec_push<T>(mut vec: Vec<T>, item: T) -> Vec<T> {
     vec.push(item);
     vec
--- a/apollo-federation/src/connectors/json_selection/helpers.rs
+++ b/apollo-federation/src/connectors/json_selection/helpers.rs
@@ -79,15 +79,16 @@ pub(crate) const fn json_type_name(v: &JSON) -> &str {
     }
 }
 
-pub(crate) fn json_to_string(json: &JSON) -> Result<Option<String>, String> {
+/// Provides a standard method to convert JSON to string.
+/// Errors on arrays or objects because "stringigying" is not semantically the same as converting to a string.
+/// null is returned as None but commonly, it gets converted to a blank string ("")
+pub(crate) fn json_to_string(json: &JSON) -> Result<Option<String>, &'static str> {
     match json {
         JSON::Null => Ok(None),
         JSON::Bool(b) => Ok(Some(b.to_string())),
         JSON::Number(n) => Ok(Some(n.to_string())),
         JSON::String(s) => Ok(Some(s.as_str().to_string())),
-        JSON::Array(_) | JSON::Object(_) => {
-            Err("cannot convert arrays or objects to strings.".to_string())
-        }
+        JSON::Array(_) | JSON::Object(_) => Err("cannot convert arrays or objects to strings."),
     }
 }
 
--- a/apollo-federation/src/connectors/json_selection/methods/public/join_not_null.rs
+++ b/apollo-federation/src/connectors/json_selection/methods/public/join_not_null.rs
@@ -8,6 +8,7 @@ use crate::connectors::json_selection::ApplyToError;
 use crate::connectors::json_selection::ApplyToInternal;
 use crate::connectors::json_selection::MethodArgs;
 use crate::connectors::json_selection::VarsWithPathsMap;
+use crate::connectors::json_selection::helpers::json_to_string;
 use crate::connectors::json_selection::immutable::InputPath;
 use crate::connectors::json_selection::location::Ranged;
 use crate::connectors::json_selection::location::WithRange;
@@ -72,16 +73,12 @@ fn join_not_null_method(
     };
 
     fn to_string(value: &JSON, method_name: &str) -> Result<Option<String>, String> {
-        match value {
-            JSON::Bool(b) => Ok(Some(b.then_some("true").unwrap_or("false").to_string())),
-            JSON::Number(number) => Ok(Some(number.to_string())),
-            JSON::String(byte_string) => Ok(Some(byte_string.as_str().to_string())),
-            JSON::Null => Ok(None),
-            JSON::Array(_) | JSON::Object(_) => Err(format!(
+        json_to_string(value).map_err(|_| {
+            format!(
                 "Method ->{} requires an array of scalar values as input",
                 method_name
-            )),
-        }
+            )
+        })
     }
 
     let joined = match data {
--- a/apollo-federation/src/connectors/json_selection/methods/public/join_not_null.rs
+++ b/apollo-federation/src/connectors/json_selection/methods/public/join_not_null.rs
@@ -74,10 +74,7 @@ fn join_not_null_method(
 
     fn to_string(value: &JSON, method_name: &str) -> Result<Option<String>, String> {
         json_to_string(value).map_err(|_| {
-            format!(
-                "Method ->{} requires an array of scalar values as input",
-                method_name
-            )
+            format!("Method ->{method_name} requires an array of scalar values as input",)
         })
     }
 
--- a/apollo-federation/src/connectors/json_selection/methods/public/to_string.rs
+++ b/apollo-federation/src/connectors/json_selection/methods/public/to_string.rs
@@ -49,7 +49,19 @@ fn to_string_method(
         JSON::Number(n) => n.to_string().into(),
         JSON::Bool(b) => b.to_string().into(),
         JSON::Null => "null".to_string().into(),
-        JSON::Array(_) | JSON::Object(_) => data.to_string().into(),
+        JSON::Array(_) | JSON::Object(_) => {
+            return (
+                None,
+                vec![ApplyToError::new(
+                    format!(
+                        "Method ->{} cannot convert arrays or objects to strings. Use ->jsonStringify instead",
+                        method_name.as_ref()
+                    ),
+                    input_path.to_vec(),
+                    method_name.range(),
+                )],
+            );
+        }
     };
 
     (Some(JSON::String(string_value)), vec![])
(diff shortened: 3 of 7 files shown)


apollo-federation/src/connectors/json_selection/methods/public/to_string.rs
```diff
@@ -0,0 +1,388 @@
...
+use crate::impl_arrow_method;
+
+impl_arrow_method!(ToStringMethod, to_string_method, to_string_shape);
+/// Returns a string representation of the applied value.
+/// Simple examples:
+///
+/// $(42)->toString()         results in "42"
+/// $("hello")->toString()    results in "hello"
+/// $(true)->toString()       results in "true"
+/// $(null)->toString()       results in "null"
```
dylan-apollo: I think this should be empty string, actually, to match how our string templates (URL & headers) work. Ideally `"{thing}"` in a string template would be identical to `thing->toString()` in the rest of code. Then if we add in the ability to do template strings in body/selection/whatever as a way of formatting, that would _also_ be consistent.

apollo-federation/src/connectors/json_selection/methods/public/to_string.rs
```diff
@@ -0,0 +1,388 @@
...
+/// $([1,2,3])->toString()    results in "[1,2,3]"
```
dylan-apollo: This should be an error, for the same reason above. `->jsonStringify` will give the json representation, and `->joinNotNull` will give CSV, so whichever format they _wanted_, they can explicitly ask for.

apollo-federation/src/connectors/json_selection/methods/public/to_string.rs
```diff
@@ -0,0 +1,388 @@
...
+                        "Method ->{} does not accept any arguments, but {} were provided",
+                        method_name.as_ref(),
+                        args.args.len()
+                    ),
+                    input_path.to_vec(),
+                    method_name.range(),
+                )],
+            );
+        }
+    }
```
dylan-apollo, line 47: There's no real reason for us to run this check at runtime. Let's check at build time and then just ignore here for performance.

apollo-federation/src/connectors/json_selection/methods/public/to_string.rs
```diff
@@ -0,0 +1,388 @@
...
+                vec![ApplyToError::new(
+                    format!(
+                        "Method ->{} cannot convert arrays or objects to strings. Use ->jsonStringify instead",
+                        method_name.as_ref()
+                    ),
+                    input_path.to_vec(),
+                    method_name.range(),
+                )],
+            );
+        }
```
dylan-apollo: Ah, the docstring up top was just wrong about arrays, so we'll want to correct that.

apollo-federation/src/connectors/json_selection/methods/public/to_string.rs
dylan-apollo: I think this is a good case for us to have a single function which converts a JSON value to string and use that in all the places that need to be consistent. So call the same function from here and https://github.com/apollographql/router/blob/f834eba96d654a3b1e1f87afdb45c3b3747f0cba/apollo-federation/src/connectors/json_selection/methods/public/join_not_null.rs#L74 and https://github.com/apollographql/router/blob/8365edfdb3a24e195fadc64c6a2c866728c87c88/apollo-federation/src/connectors/string_template.rs#L257

apollo-federation/src/connectors/json_selection/methods/public/to_string.rs
```diff
@@ -0,0 +1,388 @@
...
+    };
+
+    (Some(JSON::String(string_value)), vec![])
+}
+
+#[allow(dead_code)] // method type-checking disabled until we add name resolution
+fn to_string_shape(
+    method_name: &WithRange<String>,
+    method_args: Option<&MethodArgs>,
+    _input_shape: Shape,
```
dylan-apollo:
```suggestion
    input_shape: Shape,
```

apollo-federation/src/connectors/json_selection/methods/public/to_string.rs
```diff
@@ -0,0 +1,388 @@
...
+            Some(&MethodArgs { args, range: None }),
+            input,
+            Shape::none(),
+            &IndexMap::default(),
+            &location.source_id,
+        )
+    }
+
+    #[test]
+    fn to_string_shape_should_return_string_for_any_input() {
```
dylan-apollo:
```suggestion
    fn to_string_shape_should_return_string_for_int_input() {
```

apollo-federation/src/connectors/json_selection/methods/public/to_string.rs
```diff
@@ -0,0 +1,388 @@
...
+    }
+
+    #[test]
+    fn to_string_shape_should_return_string_for_null_input() {
+        assert_eq!(
+            get_shape(vec![], Shape::null([])),
+            Shape::string([get_location()])
+        );
+    }
+
```
dylan-apollo, line 330: I know the code itself is real simple, but I still want to see unknown/named tests for most of these since that's going to be the most common case for users today ☺️

apollo-federation/src/connectors/json_selection/methods/public/to_string.rs
```diff
@@ -0,0 +1,388 @@
...
+            )
+        );
+    }
+
+    #[test]
+    fn to_string_shape_should_error_for_array_input() {
+        assert_eq!(
+            get_shape(vec![], Shape::tuple([], [])),
+            Shape::error_with_partial(
+                "Method ->toString cannot convert arrays or objects to strings. Use ->jsonStringify instead"
```
dylan-apollo:
```suggestion
                "Method ->toString cannot convert arrays or objects to strings. Use ->jsonStringify or ->joinNotNull instead"
```

I think in some cases, joining to like `a,b,c` is what the user might be looking for.

Complete side note, we should probably start including docs links somehow for some of these errors 🤔. Something to investigate at some point.

apollo-federation/src/connectors/json_selection/methods/public/to_string.rs
andrewmcgivery: Doesn't that make this function inconsistent with other functions that already do this check?

Should we be updating other methods to also remove this check?

apollo-federation/src/connectors/json_selection/methods/public/to_string.rs
andrewmcgivery: Done!

apollo-federation/src/connectors/json_selection/methods/public/to_string.rs
andrewmcgivery: Yea, forgot to update this! 😩

apollo-federation/src/connectors/json_selection/methods/public/to_string.rs
dylan-apollo: Hmm, actually leave it for now. When we _enable_ method type checking, users will see this statically, so it won't ever trigger at runtime. But we don't have that yet, so better to leave some indicator in for now.

apollo-federation/src/connectors/json_selection/helpers.rs
```diff
@@ -79,6 +79,18 @@ pub(crate) const fn json_type_name(v: &JSON) -> &str {
...
+        JSON::Null => Ok(None),
+        JSON::Bool(b) => Ok(Some(b.to_string())),
+        JSON::Number(n) => Ok(Some(n.to_string())),
+        JSON::String(s) => Ok(Some(s.as_str().to_string())),
+        JSON::Array(_) | JSON::Object(_) => {
+            Err("cannot convert arrays or objects to strings.".to_string())
+        }
+    }
+}
+
```
dylan-apollo:
```suggestion
pub(crate) fn json_to_string(json: &JSON) -> Result<Option<String>, &'static str> {
    match json {
        JSON::Null => Ok(None),
        JSON::Bool(b) => Ok(Some(b.to_string())),
        JSON::Number(n) => Ok(Some(n.to_string())),
        JSON::String(s) => Ok(Some(s.as_str().to_string())),
        JSON::Array(_) | JSON::Object(_) => Err("cannot convert arrays or objects to strings."),
    }
}
```

The callers don't even need a `String` it looks like, so might as well skip the allocation here.

apollo-federation/src/connectors/json_selection/helpers.rs
```diff
@@ -79,6 +79,18 @@ pub(crate) const fn json_type_name(v: &JSON) -> &str {
     }
 }
 
+pub(crate) fn json_to_string(json: &JSON) -> Result<Option<String>, String> {
```
dylan-apollo: Can you add a doc comment `///` about what this function is for & why?

apollo-federation/src/connectors/json_selection/methods/public/to_string.rs
dylan-apollo:
```suggestion
/// $(null)->toString()       results in ""
```

So this is accurate now?

apollo-federation/src/connectors/string_template.rs
```diff
@@ -258,14 +259,9 @@ pub(crate) fn write_value<Output: Write>(
...
-    match value {
-        Value::Null => Ok(()),
-        Value::Bool(b) => write!(output, "{b}"),
-        Value::Number(n) => write!(output, "{n}"),
-        Value::String(s) => output.write_str(s.as_str()),
-        Value::Array(_) | Value::Object(_) => {
-            return Err("Expression is not allowed to evaluate to arrays or objects.".into());
-        }
+    match json_to_string(value) {
+        Ok(result) => write!(output, "{}", result.unwrap_or_default()),
```
dylan-apollo, line 263: This `write!` stuff is actually a bit silly now... the whole reason I made it this way was to avoid allocating the extra `String` for each value, but `json_to_string` is doing that anyway now....

Maybe not worth refactoring all the code, though, to deal with it.

The "right" way is probably to have our _own_ `Value` type with `impl Display` on it (which would either write to buffer (efficient) or allocate a new string (🤢) as needed. I'm still assuming we want that some day, but not today.

apollo-federation/src/connectors/json_selection/methods/public/join_not_null.rs
```diff
@@ -72,16 +73,12 @@ fn join_not_null_method(
...
-        match value {
-            JSON::Bool(b) => Ok(Some(b.then_some("true").unwrap_or("false").to_string())),
-            JSON::Number(number) => Ok(Some(number.to_string())),
-            JSON::String(byte_string) => Ok(Some(byte_string.as_str().to_string())),
-            JSON::Null => Ok(None),
-            JSON::Array(_) | JSON::Object(_) => Err(format!(
+        json_to_string(value).map_err(|_| {
+            format!(
                 "Method ->{} requires an array of scalar values as input",
                 method_name
```
dylan-apollo:
```suggestion
"Method ->{method_name} requires an array of scalar values as input",
```

Since you're here 😁

dylan-apollo (approved):
Added a few more suggestions, other than those this LGTM

apollo-federation/src/connectors/string_template.rs
andrewmcgivery: Perhaps could be refactored in a future PR as needed? :D

PR #7890, opened 2025-07-15
Add support for testing Apollo OTLP + HTTP metrics

timbotnik:
---

**Checklist**

- Tests added and passing[^4]
    - [ ] Unit tests
    - [ ] Integration tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/plugins/telemetry/metrics/apollo/mod.rs
+++ b/apollo-router/src/plugins/telemetry/metrics/apollo/mod.rs
@@ -25,6 +25,7 @@ use crate::plugins::telemetry::metrics::CustomAggregationSelector;
 use crate::plugins::telemetry::metrics::MetricsBuilder;
 use crate::plugins::telemetry::metrics::MetricsConfigurator;
 use crate::plugins::telemetry::otlp::CustomTemporalitySelector;
+use crate::plugins::telemetry::otlp::Protocol;
 use crate::plugins::telemetry::tracing::BatchProcessorConfig;
 
 pub(crate) mod histogram;
@@ -52,6 +53,7 @@ impl MetricsConfigurator for Config {
             Config {
                 endpoint,
                 experimental_otlp_endpoint: otlp_endpoint,
+                experimental_otlp_metrics_protocol: otlp_metrics_protocol,
                 apollo_key: Some(key),
                 apollo_graph_ref: Some(reference),
                 schema_id,
@@ -82,6 +84,7 @@ impl MetricsConfigurator for Config {
                     builder = Self::configure_apollo_otlp_metrics(
                         builder,
                         otlp_endpoint,
+                        otlp_metrics_protocol,
                         key,
                         reference,
                         schema_id,
@@ -102,6 +105,7 @@ impl Config {
     fn configure_apollo_otlp_metrics(
         mut builder: MetricsBuilder,
         endpoint: &Url,
+        otlp_protocol: &Protocol,
         key: &str,
         reference: &str,
         schema_id: &str,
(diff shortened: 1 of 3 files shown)


apollo-router/src/plugins/telemetry/metrics/apollo/mod.rs
```diff
@@ -110,15 +114,24 @@ impl Config {
...
+                    .with_endpoint(endpoint.as_str())
+                    .with_timeout(batch_processor.max_export_timeout)
+                    .with_metadata(metadata.clone())
+                    .with_compression(opentelemetry_otlp::Compression::Gzip),
+            ),
+            Protocol::Http => MetricsExporterBuilder::Http(
+                opentelemetry_otlp::new_exporter()
+                    .http()
+                    .with_endpoint(endpoint.as_str())
+                    // .with_protocol(opentelemetry_otlp::Protocol::HttpJson)  TBD: decide if we want to use HTTP JSON vs. HTTP Protobuf
```
timbotnik: @rregitsky not sure what you will find easier here but seems fine to use either one (looks like the traces test uses `ExportTraceServiceRequest::decode` so by default it would be something similar e.g. decode as `ExportMetricsServiceRequest` before checking the contents).

rregitsky: It [looks like](https://github.com/apollographql/router/blob/c8ae92e47de4e1bb3ae56e9beed27fbc2a1e136a/apollo-router/src/plugins/telemetry/otlp.rs#L188) we do `http()` + `Protocol::Grpc` for `telemetry.exporters.metrics.otlp.protocol` so I say we follow that approach. Makes checking for both in the same test simpler.

apollo-router/tests/common.rs
```diff
@@ -538,6 +545,12 @@ impl IntegrationTest {
         let address = listener.local_addr().unwrap();
         let url = format!("http://{address}/");
 
+        let apollo_otlp_listener =
+            TcpListener::bind(SocketAddr::from(([127, 0, 0, 1], 0))).unwrap();
+        let apollo_otlp_address = apollo_otlp_listener.local_addr().unwrap();
+        // This should ideally use the process_endpoint function to add the /v1/metrics suffix.
```
timbotnik: Can remove this now (we shouldn't change the endpoint here anyway so we can eventually support traces with the same server just a different path)
```suggestion
```

.changesets/config_timbotnik_support_http_integration_tests.md
```diff
@@ -0,0 +1,5 @@
+### Brief but complete sentence that stands on its own ([PR #7890](https://github.com/apollographql/router/pull/7890))
```
goto-bus-stop: It sounds like our users should _not_ enable this. In that case, we should just not write a changeset

rregitsky: Good call -- Deleted the file.

apollo-router/src/plugins/telemetry/otlp.rs
```diff
@@ -86,7 +86,7 @@ pub(crate) enum TelemetryDataKind {
 // If you are interested in learning more about opentelemetry endpoints:
 //  https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md
 // contains the details.
-fn process_endpoint(
+pub(super) fn process_endpoint(
```
timbotnik, line 89: I'm thinking we should move this method out of the plugin.  It does feel like this is more "library related to OTel" functionality than something specific to this plugin and having a dependency on the plugin method feels a little strange.  I'll look around for other options.

timbotnik: Looking around I'm not really sure if there's somewhere better.  Open to suggestions or good to just leave it.

apollo-router/tests/integration/telemetry/apollo_otel_metrics.rs
```diff
@@ -0,0 +1,131 @@
...
+        .telemetry(Telemetry::Otlp { endpoint: None })
+        .config(
+            r#"
+            include_subgraph_errors:
+                all: true
+            telemetry:
+                apollo:
+                    experimental_otlp_metrics_protocol: http
+                    batch_processor:
+                        scheduled_delay: 1s # lowering this seems to make the test flaky
```
goto-bus-stop, line 89: Do we know how/why it makes it flaky? It's a bit worrying because 1s is a long time.

apollo-router/src/plugins/telemetry/metrics/apollo/mod.rs
```diff
@@ -102,6 +107,7 @@ impl Config {
     fn configure_apollo_otlp_metrics(
         mut builder: MetricsBuilder,
         endpoint: &Url,
+        otlp_protocol: &Protocol,
```
goto-bus-stop, line 110: Might be good to have a comment somewhere in this file also explaining that "While Apollo doesn't use the HTTP protocol, we support it in the router so we can use wiremock in tests.", here or where we create the HTTP exporters.

apollo-router/tests/integration/telemetry/apollo_otel_metrics.rs
rregitsky: Honestly, I'm not sure. I was able to reduce it to 500ms. I have  a follow-up PR that will make some significant changes to this integration test. I'll investigate further there.

goto-bus-stop: sgtm. i'm not really bothered if it's 500ms or 1s, but if we have to set it to a particular value to make tests pass then ideally we want to understand why 😇

rregitsky: Agreed -- I'd much prefer it understood. Unfortunately I had to bump back to 1s as 500ms was flaky in CI. Will definitely dig into this.

PR #7892, opened 2025-07-15
feat(response_cache): incorporate metrics for tracking expired data volume in the database

bnjjj:
This metric will be useful to configure the `cleanup_interval` properly. It will tell to our users how much expired data is still in database and will be cleaned.

---

**Checklist**

- Tests added and passing[^4]
    - [x] Unit tests
    - [x] Integration tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-1374]: https://apollographql.atlassian.net/browse/ROUTER-1374?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apollo-router/src/plugins/response_cache/metrics.rs
```diff
@@ -293,3 +301,46 @@ impl From<CacheMetricContextKey> for String {
...
+    let mut interval = IntervalStream::new(tokio::time::interval(std::time::Duration::from_secs(
+        (pg_cache.cleanup_interval.num_seconds().min(60) / 2) as u64,
+    )));
+    let expired_data_count = Arc::new(AtomicU64::new(0));
+    let expired_data_count_clone = expired_data_count.clone();
+    let meter = meter_provider().meter("apollo/router");
+    let _gauge = meter
+        .u64_observable_gauge("apollo.router.response_cache.data.expired")
+        .with_description("Amount of expired data still in database")
+        .with_unit("entry")
```
carodewig: [Non-units should use braces](https://opentelemetry.io/docs/specs/semconv/general/metrics/#instrument-units):
```suggestion
        .with_unit("{entry}")
```

apollo-router/src/plugins/response_cache/metrics.rs
```diff
@@ -293,3 +301,46 @@ impl From<CacheMetricContextKey> for String {
...
+) {
+    let mut interval = IntervalStream::new(tokio::time::interval(std::time::Duration::from_secs(
+        (pg_cache.cleanup_interval.num_seconds().min(60) / 2) as u64,
+    )));
+    let expired_data_count = Arc::new(AtomicU64::new(0));
+    let expired_data_count_clone = expired_data_count.clone();
+    let meter = meter_provider().meter("apollo/router");
+    let _gauge = meter
+        .u64_observable_gauge("apollo.router.response_cache.data.expired")
+        .with_description("Amount of expired data still in database")
```
carodewig: Nit:
```suggestion
        .with_description("Count of expired data entries still in database")
```

apollo-router/src/plugins/response_cache/metrics.rs
```diff
@@ -293,3 +301,46 @@ impl From<CacheMetricContextKey> for String {
         format!("{CACHE_INFO_SUBGRAPH_CONTEXT_KEY}_{}", val.0)
     }
 }
+
+pub(super) async fn expired_data_task(
```
carodewig, line 307: I think it would be helpful to add a comment to this function that explains that this task counts all rows in the provided `pg_cache`, and the `subgraph_name` is used when a subgraph uses a non-default cache. It took me a while to work this out when reading for the first time.

apollo-router/src/plugins/response_cache/plugin.rs
```diff
@@ -125,6 +150,15 @@ impl Storage {
...
 
+/// Call .abort on task when dropped
+#[derive(Clone)]
+struct AbortOnDrop(AbortHandle);
+impl Drop for AbortOnDrop {
+    fn drop(&mut self) {
+        self.0.abort();
+    }
+}
+
```
carodewig, line 161: This is a great idea and I could see it being useful for other things as well! Ie the abort handle in `batching::Batch`

apollo-router/src/plugins/response_cache/metrics.rs
```diff
@@ -293,3 +301,46 @@ impl From<CacheMetricContextKey> for String {
...
+    pg_cache: PostgresCacheStorage,
+    subgraph_name: Option<String>,
+) {
+    let mut interval = IntervalStream::new(tokio::time::interval(std::time::Duration::from_secs(
+        (pg_cache.cleanup_interval.num_seconds().min(60) / 2) as u64,
```
carodewig: I think it might be useful to have this interval be configurable separately, or even just defaulted to 60s. If a user sets a long `cleanup_interval`, I think it would be useful for them to see the cache growing at a more frequent rate so that they can tune / decrease their cleanup interval

bnjjj: It's not really a default it's just to not execute this more than once in a minute. I think we will document it but it's also a limit we put to not decrease the performance on the database because that could really be bad. So I would not give that as an option

BrynCooke: Should this not be `max` rather than `min`? Otherwise if the original number is 10 then 10.min(60) will be 10.

apollo-router/src/plugins/response_cache/postgres.rs
```diff
@@ -428,16 +428,39 @@ impl PostgresCacheStorage {
                 FROM cache
                 USING invalidation_key
                 WHERE invalidation_key.invalidation_key = ANY($1::text[])
-                    AND invalidation_key.cache_key_id = cache.id  AND invalidation_key.subgraph_name = ANY($2::text[]) RETURNING cache.cache_key
+                    AND invalidation_key.cache_key_id = cache.id  AND invalidation_key.subgraph_name = ANY($2::text[]) RETURNING cache.cache_key, cache.expires_at, invalidation_key.subgraph_name
             )
-        SELECT COUNT(*) AS count FROM deleted"#,
+        SELECT subgraph_name, COUNT(deleted.cache_key) AS count FROM deleted WHERE deleted.expires_at >= NOW() GROUP BY deleted.subgraph_name"#,
             &invalidation_keys,
             &subgraph_names
```
BrynCooke, line 437: Let's add some comments around this this SQL statement. I undertand that the first one is returning the number of records deleted, but don't understand the second statement without more context.

apollo-router/src/plugins/response_cache/metrics.rs
bnjjj: Done

carodewig (approved):

PR #7893, opened 2025-07-15
Various improvements to configuration JSON schema

goto-bus-stop:
Update schemars, which we use to generate the JSON schema for our configuration object. Major changes on top:
- Manually name some types, in particular generic types, to avoid having references in the schema that must be URL-encoded. Removes the need for the `RefRenameVisitor`.
- Add comments to some fields that were now failing the `all_properties_have_description` test. They started failing because the previous version of schemars was generating garbage descriptions in some cases and the new version does not.
- We now generate schemas like `type: [object, null]` instead of `type: object, nullable: true`. `nullable` was an unofficial extension to JSON schema that is no longer supported by schemars.

The rest is just adapting to breaking changes. The new schemars crate has much less strict typing but is much more concise to use.

| Before | After |
|-|-|
| image | image |

Commits are explorational and not useful for review, but I think the changes are mostly straightforward to review without going commit by commit?

---

**Checklist**

- Tests added and passing[^4]
    - [x] Unit tests
    - [ ] Integration tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-1355]: https://apollographql.atlassian.net/browse/ROUTER-1355?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apollo-router/src/configuration/schema.rs
```diff
@@ -32,48 +26,18 @@ use crate::configuration::upgrade::upgrade_configuration;
...
-        visit_schema_object(self, schema);
-    }
-}
-
 /// Generate a JSON schema for the configuration.
-pub(crate) fn generate_config_schema() -> RootSchema {
+pub(crate) fn generate_config_schema() -> Schema {
     let settings = SchemaSettings::draft07().with(|s| {
-        s.option_nullable = true;
-        s.option_add_null_type = false;
```
goto-bus-stop, line 65: This difference is worth calling out. In the new schemars i believe this option is gone. This means that instead of `nullable: true`, you get `type: [object, null]` or `anyOf: [{ type: object }, { type: null }]`. In my manual testing I did not find this a regression for the editing experience, but it does make error messages a bit worse in the latter case.

I _think_ that generating a newer version than draft-07 would enable using `nullable: true` again, as it isn't strictly part of that? But I haven't looked into the other implications of that.

apollo-router/src/configuration/schema.rs
```diff
@@ -103,18 +67,33 @@ pub(crate) fn validate_yaml_configuration(
     migration: Mode,
 ) -> Result<Configuration, ConfigurationError> {
     let defaulted_yaml = if raw_yaml.trim().is_empty() {
-        "plugins:".to_string()
+        "{}".to_string()
```
goto-bus-stop, line 70: Using an empty object seems closer to the intent of having minimally valid YAML than using `plugins: null`.

apollo-router/src/configuration/schema.rs
goto-bus-stop: Ah, `nullable: true` is just not a JSON Schema thing, but an openapi thing. This seems right then honestly.

dylan-apollo: `nullable: true` is also not an OpenAPI thing anymore, OpenAPI 3.1 adopted the more recent JSONSchema approach of `type: [T, null]` instead.

apollo-router/src/plugins/connectors/configuration.rs
```diff
@@ -92,18 +103,12 @@ pub(crate) struct SourceConfiguration {
...
-            map
-        },
-        ..Default::default()
-    }
-    .into()
+fn uri_schema(_generator: &mut schemars::SchemaGenerator) -> schemars::Schema {
+    schemars::json_schema!({
+        "type": "string",
+        "format": "uri",
+        "nullable": true,
```
dylan-apollo: Should this now be `"type": ["string", "null"]` instead of `"nullable": true`? I only added that manual attribute in the first place to match the rest of how schemars was set up 😁

apollo-router/src/plugins/connectors/configuration.rs
```diff
@@ -68,13 +70,22 @@ pub(crate) struct ConnectorsConfig {
...
     /// A map of `@source(name:)` to configuration for that source
+    #[schemars(schema_with = "sources_configuration_schema")]
     pub(crate) sources: HashMap<SourceName, SourceConfiguration>,
 
     /// Other values that can be used by connectors via `{$config.<key>}`
     #[serde(rename = "$config")]
     pub(crate) custom: CustomConfiguration,
 }
 
+// Custom schema implementation to not restrict key type
```
dylan-apollo, line 81: What did the key type schema look like? Would be nice if there was _some_ way to indicate to users that the left hand side here is a `@source(name:)`, but IDE help isn't the best with custom types _anyway_ so... 🤷

goto-bus-stop: In schemars 0.x, the key type was completely ignored, and it always generated an `additionalProperties: ValueType` schema.
In schemars 1.x, an implementation of `JsonSchema` is only provided for hashmaps with a `String` key.
So this is manually declaring the previous behaviour.

goto-bus-stop:
(fwiw, i did not see a substantial difference in compile time due to this upgrade, with `cargo build --timings`)

apollo-router/src/configuration/subgraph.rs
```diff
@@ -75,6 +75,7 @@ use serde::de::Visitor;
 
 /// Configuration options pertaining to the subgraph server component.
 #[derive(Default, Serialize, JsonSchema)]
+#[schemars(rename = "Subgraph{T}Configuration")]
```
goto-bus-stop, line 78: `{T}` syntax interpolates a type parameter's serialized name into the string. I tried to make things faaairly natural to read

Cargo.lock
```diff
@@ -327,7 +327,8 @@ dependencies = [
  "itoa",
  "jsonpath-rust",
  "jsonpath_lib",
- "jsonschema",
+ "jsonschema 0.17.1",
+ "jsonschema 0.30.0",
```
BrynCooke, line 331: Why are there two versions here now?

goto-bus-stop: 0.30.0 is used in tests to validate that the generated schema against the meta schema. It provides a trivially easy function to do so that I think doesn't exist in older versions. I would follow up with a PR to update jsonschema used in the router itself and then we wouldn't have a duplicate anymore.

PR #7894, opened 2025-07-15
Add isSuccess argument to @source and @connect directives

Hunter-apollo:
Adds a new `isSuccess` argument to the `@source` and `@connect` directive definitions. 

This parameter will allow users to override the success conditions for connector responses. For example, a user might want to include 404 response data in the data returned from the connector rather than treating the response as an error.

Note: Validation and Error handling functionality related to this argument will be added in separate PR to keep the reviews relatively small.
---

**Checklist**

- Tests added and passing[^4]
    - [ ] Unit tests
    - [ ] Integration tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-federation/src/connectors/models.rs
+++ b/apollo-federation/src/connectors/models.rs
@@ -77,12 +77,14 @@ pub struct ConnectorErrorsSettings {
     pub message: Option<JSONSelection>,
     pub source_extensions: Option<JSONSelection>,
     pub connect_extensions: Option<JSONSelection>,
+    pub connect_is_success: Option<JSONSelection>,
 }
 
 impl ConnectorErrorsSettings {
     fn from_directive(
         connect_errors: Option<&ErrorsArguments>,
         source_errors: Option<&ErrorsArguments>,
+        connect_is_success: Option<&JSONSelection>,
     ) -> Self {
         let message = connect_errors
             .and_then(|e| e.message.as_ref())
@@ -90,11 +92,12 @@ impl ConnectorErrorsSettings {
             .cloned();
         let source_extensions = source_errors.and_then(|e| e.extensions.as_ref()).cloned();
         let connect_extensions = connect_errors.and_then(|e| e.extensions.as_ref()).cloned();
-
+        let connect_is_success = connect_is_success.cloned();
         Self {
             message,
             source_extensions,
             connect_extensions,
+            connect_is_success,
         }
     }
 
@@ -111,6 +114,12 @@ impl ConnectorErrorsSettings {
             )
             .chain(
                 self.connect_extensions
+                    .as_ref()
+                    .into_iter()
+                    .flat_map(|m| m.variable_references()),
+            )
+            .chain(
+                self.connect_is_success
                     .as_ref()
                     .into_iter()
                     .flat_map(|m| m.variable_references()),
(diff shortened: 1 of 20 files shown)


apollo-federation/src/connectors/models.rs
```diff
@@ -77,24 +77,27 @@ pub struct ConnectorErrorsSettings {
...
     pub source_extensions: Option<JSONSelection>,
     pub connect_extensions: Option<JSONSelection>,
+    pub connect_is_success: Option<JSONSelection>,
 }
 
 impl ConnectorErrorsSettings {
     fn from_directive(
         connect_errors: Option<&ErrorsArguments>,
         source_errors: Option<&ErrorsArguments>,
+        connect_is_success: Option<&JSONSelection>,
```
andrewmcgivery, line 87: So something we learned doing the debugger... you probably want to pass both `connect` and `source` `is_success` through this `ConnectorErrorsSettings` and "just in time" decide which one to use so that we can make them both available to the debugger.

apollo-federation/src/connectors/spec/connect.rs
```diff
@@ -588,9 +609,25 @@ mod tests {
...
                 batch: None,
                 errors: None,
+                is_success: None,
             },
         ]
         "#
         );
     }
+
+    #[test]
```
andrewmcgivery, line 619: Do we need this test? Shouldn't the existing tests cover this?

Hunter-apollo: The existing tests do not have the `isSuccess` argument in the SDL, so no, they do not already cover this case.

apollo-federation/src/connectors/models.rs
andrewmcgivery: Nevermind, yours overrides, it doesn't merge. This is consistent with message.

apollo-federation/src/connectors/spec/source.rs
```diff
@@ -397,4 +373,38 @@ mod tests {
             .find(|var_ref| var_ref.namespace.namespace == Namespace::Config)
             .unwrap();
     }
+
+    #[test]
+    fn it_supports_is_success_in_source() {
```
andrewmcgivery, line 378: Same here... do the existing tests not cover this?

apollo-federation/src/connectors/spec/type_and_directive_specifications.rs
```diff
@@ -361,6 +362,22 @@ fn connect_directive_spec() -> DirectiveSpecification {
                 },
                 composition_strategy: None,
             },
+            DirectiveArgumentSpecification {
+                base_spec: ArgumentSpecification {
+                    name: IS_SUCCESS_ARGUMENT_NAME,
+                    get_type: |s, _| {
+                        let name = s
```
andrewmcgivery: Why is this different than the other arguments?

apollo-federation/src/connectors/spec/type_and_directive_specifications.rs
```diff
@@ -431,6 +448,22 @@ fn source_directive_spec() -> DirectiveSpecification {
...
```
andrewmcgivery: Same question... why is this different?

apollo-federation/src/connectors/tests/schemas/is-success.graphql
```diff
@@ -0,0 +1,126 @@
+schema
```
andrewmcgivery, line 1: I'm a little bit confused by your files here. You are supposed to have a yaml file with the source graphql and then have the generated .graphql file (sdl).

apollo-federation/src/connectors/spec/connect.rs
andrewmcgivery: Reason I'm asking is because we haven't added tests in this file specific to an attribute before this one. It just sticks out as an inconsistency.

Maybe this is the "right" thing to do but I guess I'm wondering if we should instead modify the existing test or if we should have been adding new tests to this file all along.

Hunter-apollo: IMO we could probably probably use more testing here.

I dont think we should update the old tests to add these checks as I want to be able to confirm they are purely addative.

apollo-federation/src/connectors/spec/type_and_directive_specifications.rs
Hunter-apollo: I guess it could be the same as the others. Ill update.

apollo-federation/src/connectors/tests/schemas/is-success.graphql
Hunter-apollo: I will add the yaml file.

andrewmcgivery (approved):

PR #7900, opened 2025-07-16
response_cache: check failure mode when pg is unavailable and improve observability

bnjjj:
If the router cannot connect to Postgres and the configuration is set to `required_to_start: false`, it will continue to serve responses correctly even if the cache is not working. Spans will be marked with errors, and counters in metrics will be incremented.

This PR introduces support for various metrics:
- `apollo.router.operations.response_cache.invalidation.error`: Counter for requests with an invalidation error.
- `apollo.router.operations.response_cache.fetch.error`: Counter for responses unable to fetch cached data from storage.
- `apollo.router.operations.response_cache.insert.error`: Counter tracking errors when inserting cached data with an error.
- `apollo.router.operations.response_cache.insert`: Histogram measuring the duration of data insertion into cache storage.

Additionally, it provides new configuration options for Postgres:
- `idle_timeout`: Maximum idle duration for individual PostgreSQL connections (default: 1min).
- `acquired_timeout`: Maximum time spent waiting for a PostgreSQL connection (default: 50ms).

---

**Checklist**

- Tests added and passing[^4]
    - [x] Unit tests
    - [x] Integration tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-1377]: https://apollographql.atlassian.net/browse/ROUTER-1377?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apollo-router/src/plugins/response_cache/invalidation.rs
```diff
@@ -173,6 +176,15 @@ impl Invalidation {
...
                     );
+                    if let Err(InvalidationError::Postgres(err)) = &res {
+                        u64_counter_with_unit!(
+                            "apollo.router.operations.response_cache.invalidation.error",
+                            "Errors when invalidating data in cache",
+                            "{error}",
+                            1,
+                            "code" = err.code()
+                        );
+                    }
```
carodewig: Should this counter also handle other types of `InvalidationError`? Or are those counted elsewhere?

apollo-router/src/plugins/response_cache/plugin.rs
```diff
@@ -713,7 +778,13 @@ impl CacheService {
                     self.supergraph_schema.clone(),
                     &self.subgraph_enums,
                 )
-                .instrument(tracing::info_span!("cache.entity.lookup"))
+                .instrument(tracing::info_span!(
+                    "response.cache.lookup",
```
carodewig: Thoughts on `response_cache.lookup` vs `response.cache.lookup`? I tend to prefer `response_cache` given that's what we use elsewhere

apollo-router/src/plugins/response_cache/postgres.rs
```diff
@@ -137,11 +157,16 @@ pub(crate) enum PostgresCacheStorageError {
 
 impl PostgresCacheStorage {
     pub(crate) async fn new(conf: &PostgresCacheConfig) -> Result<Self, PostgresCacheStorageError> {
+        // After 500ms trying to get a connection from PG pool it will return a warning in logs
+        const ACQUIRE_SLOW_THRESHOLD: std::time::Duration = std::time::Duration::from_millis(500);
```
carodewig, line 162: The `ACQUIRE_SLOW_THRESHOLD` is greater than the default acquire_timeout - that may be fine, but would it make sense to have it be smaller given our default?

apollo-router/src/plugins/response_cache/postgres.rs
```diff
@@ -440,9 +473,12 @@ impl PostgresCacheStorage {
         Ok(rec.count.unwrap_or_default() as u64)
     }
 
-    pub(crate) async fn update_cron(&self) -> anyhow::Result<()> {
-        let cron = Cron::try_from(&self.cleanup_interval)
-            .map_err(PostgresCacheStorageError::InvalidCleanupInterval)?;
+    pub(crate) async fn update_cron(&self) -> sqlx::Result<()> {
+        let cron = Cron::try_from(&self.cleanup_interval).map_err(|err| {
+            sqlx::Error::Encode(Box::new(PostgresCacheStorageError::InvalidCleanupInterval(
```
carodewig: This error variant might make more sense?
```suggestion
            sqlx::Error::Configuration(Box::new(PostgresCacheStorageError::InvalidCleanupInterval(
```

apollo-router/src/plugins/response_cache/postgres.rs
bnjjj: I want to make sure to not create useless warnings. Like if we have users setting a bigger timeout, if the time is bigger than 500ms there's definitely an issue they can't just ignore it. I don't want to take the risk of flooding our customers if we set it to a lower value

apollo-router/src/plugins/response_cache/plugin.rs
bnjjj: Makes sense, I wasn't really happy with this version so that's great

apollo-router/src/plugins/response_cache/invalidation.rs
bnjjj: fixed

apollo-router/tests/integration/response_cache.rs
```diff
@@ -204,6 +234,50 @@ async fn basic_cache_skips_subgraph_request() {
...
+    insta::assert_yaml_snapshot!(body, @r###"
+        data:
+          topProducts:
+            - reviews:
+                - id: r1a
+                - id: r1b
+            - reviews:
+                - id: r2
+    "###);
+    // Unchanged, everything is in cache so we don’t need to make more subgraph requests:
```
SimonSapin:
```suggestion
    // Would have been unchanged because both subgraph requests were cacheable,
    // but cache storage isn’t available to we fall back to calling the subgraph again
```

connect-on-interface-object.graphql
SimonSapin, line 1: Is this needed?

bnjjj: no

SimonSapin (approved):

PR #7901, opened 2025-07-16
Reliably distinguish GraphQL errors and transport errors in subscriptions

goto-bus-stop:
Using a GraphQL response extension to mark "fatal" errors that should be reported as a top-level `{errors}` key in the subscriptions chunk, while letting non-fatal errors through as `{payload: {errors}}` chunks. In the past, we tried to detect this "magically" by seeing if the error is the _final_ response in the stream, which isn't 100% accurate and can also be flaky depending on when the subgraph closes the connection.

I wanted to use a `Result` type here to enforce that fatal errors would be marked and handled correctly, but that would be a breaking change to our response types and every plugin. So instead, responses that are subscription fatal errors have a GraphQL extension `apollo::subscriptions::fatal_error`. The multipart serialisation code uses only that extension to decide how to format the response.

**Todos**
- [x] Update the test expectations
- [x] `test_subscription_ws_passthrough_pure_error_payload` appears to (still) be flaky with these changes

---

**Checklist**

- Tests added and passing[^4]
    - [x] Unit tests
    - [x] Integration tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-1343]: https://apollographql.atlassian.net/browse/ROUTER-1343?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apollo-router/src/protocols/multipart.rs
+++ b/apollo-router/src/protocols/multipart.rs
@@ -116,9 +116,15 @@ impl Stream for Multipart {
 
                     match self.mode {
                         ProtocolMode::Subscription => {
-                            let is_transport_error = response.extensions.remove(SUBSCRIPTION_ERROR_EXTENSION_KEY) == Some(true.into());
+                            let is_transport_error =
+                                response.extensions.remove(SUBSCRIPTION_ERROR_EXTENSION_KEY)
+                                    == Some(true.into());
                             // Magic empty response (that we create internally) means the connection was gracefully closed at the server side
-                            if !is_still_open && response.data.is_none() && response.errors.is_empty() && response.extensions.is_empty() {
+                            if !is_still_open
+                                && response.data.is_none()
+                                && response.errors.is_empty()
+                                && response.extensions.is_empty()
+                            {
                                 self.is_terminated = true;
                                 return Poll::Ready(Some(Ok(Bytes::from_static(&b"--\r\n"[..]))));
                             }
--- a/apollo-router/src/protocols/multipart.rs
+++ b/apollo-router/src/protocols/multipart.rs
@@ -12,6 +12,7 @@ use tokio_stream::once;
 use tokio_stream::wrappers::IntervalStream;
 
 use crate::graphql;
+use crate::services::SUBSCRIPTION_ERROR_EXTENSION_KEY;
 
 #[cfg(test)]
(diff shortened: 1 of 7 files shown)


apollo-router/src/protocols/multipart.rs
```diff
@@ -115,27 +116,33 @@ impl Stream for Multipart {
 
                     match self.mode {
                         ProtocolMode::Subscription => {
```
goto-bus-stop, line 118: This is where the actual change happens.
Now, we decide whether to send `payload` or `errors` in the response exclusively based on the `SUBSCRIPTION_ERROR_EXTENSION_KEY` being `true`.

The code is reordered a bit (special handling for graceful server-side close is moved earlier) so we aren't creating _another_ special-meaning internal message (`SubscriptionPayload` with `payload` being an empty object/null...)

apollo-router/src/protocols/multipart.rs
```diff
@@ -115,27 +116,39 @@ impl Stream for Multipart {
...
-                            // Gracefully closed at the server side
-                            if !is_still_open && resp.payload.is_none() && resp.errors.is_empty() {
+                            let is_transport_error =
+                                response.extensions.remove(SUBSCRIPTION_ERROR_EXTENSION_KEY)
+                                    == Some(true.into());
+                            // Magic empty response (that we create internally) means the connection was gracefully closed at the server side
+                            if !is_still_open
+                                && response.data.is_none()
+                                && response.errors.is_empty()
+                                && response.extensions.is_empty()
```
BrynCooke, line 126: What is the significance of extensions being empty? If something other than the subscriptions error key was in extensions then will that break graceful close?

BrynCooke: I guess if extensions is not empty then we want to return the response

goto-bus-stop: Essentially I'm trying to mimick what the previous code does. In the previous code, `resp.payload.is_none()` would only be true if both `data` and `extensions` were null/none.

I'm not sure if this is like actually required, to be honest. The subscriptions graceful close response can only have extensions if a plugin or user code added them.

PR #7875, opened 2025-07-10
Support looking up a composition spec by a directive link and name

tninesling:
This change makes some updates to the spec definitions in anticipation of adding the merge logic for links. As the links are merged, we need to look up the composition specification for each imported directive. This PR adds a static registry for all feature specs, which allows lookups equivalent to the JS function [coreFeatureDefinitionIfKnown](https://github.com/apollographql/federation/blob/main/internals-js/src/knownCoreFeatures.ts).

From there, looking up a composition specification is trickier. As currently implemented, composition specs are generated from scratch each time they are invoked. Ideally, we would place the directive and type specifications in a lookup to make them easier to find, but the use of `&dyn Fn` in the spec definitions disallows this because everything in the static registry must be `Sync`. For now, I have set it up to regenerate the whole spec, find the part corresponding to the directive we're searching for, and downcasting that to the correct type for directive specifications.

---

**Checklist**

Complete the checklist (and note appropriate exceptions) before the PR is marked ready-for-review.

- [X] PR description explains the motivation for the change and relevant context for reviewing
- [X] PR description links appropriate GitHub/Jira tickets (creating when necessary)
- [ ] Changeset is included for user-facing changes
- [X] Changes are compatible[^1]
- [ ] Documentation[^2] completed
- [ ] Performance impact assessed and acceptable
- [ ] Metrics and logs are added[^3] and documented
- Tests added and passing[^4]
    - [ ] Unit tests
    - [ ] Integration tests
    - [ ] Manual tests, as necessary

**Exceptions**

*Note any exceptions here*

**Notes**

[^1]: It may be appropriate to bring upcoming changes to the attention of other (impacted) groups. Please endeavour to do this before seeking PR approval. The mechanism for doing this will vary considerably, so use your judgement as to how and when to do this.
[^2]: Configuration is an important part of many changes. Where applicable please try to document configuration examples.
[^3]: A lot of (if not most) features benefit from built-in observability and `debug`-level logs. Please read [this guidance](https://github.com/apollographql/router/blob/dev/dev-docs/metrics.md#adding-new-metrics) on metrics best-practices.
[^4]: Tick whichever testing boxes are applicable. If you are adding Manual Tests, please document the manual testing (extensively) in the Exceptions.

[...]

PR #7906, opened 2025-07-16
Implement logic for collecting and merging in imported specs

tninesling:
This change builds on #7875 to merge specs imported by subgraphs into the resulting supergraph. The main functions are `collect_core_directives_to_compose` ([JS source](https://github.com/apollographql/federation/blob/4653320016ed4202a229d9ab5933ad3f13e5b6c0/composition-js/src/merging/coreDirectiveCollector.ts#L10)) and `validate_and_maybe_add_specs` ([JS source](https://github.com/apollographql/federation/blob/4653320016ed4202a229d9ab5933ad3f13e5b6c0/composition-js/src/merging/merge.ts#L487)).

The former iterates over the subgraphs and creates a list of specs to merge in, along with their composition specifications. The output is then passed into the latter to actually apply the specs to the supergraph. The combination of these functions happens in `prepare_supergraph` which will be implemented in a future ticket.

This also includes a Rust implementation of the JS [applyFeatureToSchema](https://github.com/apollographql/federation/blob/4653320016ed4202a229d9ab5933ad3f13e5b6c0/internals-js/src/specs/coreSpec.ts#L538).

---

**Checklist**

- Tests added and passing[^4]
    - [ ] Unit tests
    - [ ] Integration tests

**Exceptions**

*Note any exceptions here*

**Notes**

[FED-554]: https://apollographql.atlassian.net/browse/FED-554?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apollo-federation/src/merger/merge_links.rs
```diff
@@ -0,0 +1,280 @@
...
+    #[allow(dead_code)]
+    fn validate_and_maybe_add_specs(
+        &mut self,
+        directives_merge_info: &[CoreDirectiveInSubgraphs],
+    ) -> Result<(), FederationError> {
+        let mut supergraph_info_by_identity: HashMap<Identity, Vec<CoreDirectiveInSupergraph>> =
+            HashMap::new();
+
+        for subgraph_core_directive in directives_merge_info {
+            let mut name_in_supergraph: Option<&Name> = None;
```
clenfest, line 123: Sachin had some code where we would return from the function if there was no compositionSpec, but I don't really see how that would be even in the old code, so I guess it's ok.

clenfest (approved):

apollo-federation/src/merger/merge_links.rs
tninesling: There's a section down below where we skip the current link if there's no name in the supergraph. It matches the JS implementation fairly faithfully, but we can always come back and change it later

PR #7914, opened 2025-07-17
[Connectors] Add isSuccess logic to connector HTTP error handling

Hunter-apollo:
Adds the core error handling functionality for the `isSuccess` parameter. 

The `isSuccess` parameter allows users to set custom success condition that overrides the default error condition (i.e. checking that status is 2XX). This allows users to get data from an error response (for example a 404).

Note that deserialization and rate limiting errors are not mapped by this logic and will always be marked as internal connector failures.

---

**Checklist**

- Tests added and passing[^4]
    - [ ] Unit tests
    - [ ] Integration tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-federation/Cargo.toml
+++ b/apollo-federation/Cargo.toml
@@ -51,6 +51,7 @@ shape = "0.5.2"
 form_urlencoded = "1.2.1"
 parking_lot = "0.12.4"
 mime = "0.3.17"
+once_cell = "1.21.3"
 
 [dev-dependencies]
 diff = "0.1.13"
--- a/apollo-federation/src/connectors/models/problem_location.rs
+++ b/apollo-federation/src/connectors/models/problem_location.rs
@@ -12,6 +12,7 @@ pub enum ProblemLocation {
     ConnectQueryParams,
     SourceHeaders,
     ConnectHeaders,
+    IsSuccessMapping,
     Selection,
     ErrorsMessage,
     SourceErrorsExtensions,
--- a/apollo-federation/src/connectors/runtime/responses.rs
+++ b/apollo-federation/src/connectors/runtime/responses.rs
@@ -1,7 +1,8 @@
-use std::cell::LazyCell;
+use std::cell::OnceCell;
 use std::sync::Arc;
 
 use apollo_compiler::collections::HashMap;
+use apollo_compiler::collections::IndexMap;
 use http::HeaderMap;
 use http::HeaderValue;
 use itertools::Itertools;
@@ -14,7 +15,7 @@ use crate::connectors::Connector;
 use crate::connectors::JSONSelection;
 use crate::connectors::ProblemLocation;
 use crate::connectors::runtime::debug::ConnectorContext;
-use crate::connectors::runtime::debug::ConnectorDebugHttpRequest;
+use crate::connectors::runtime::debug::DebugRequest;
 use crate::connectors::runtime::debug::SelectionData;
 use crate::connectors::runtime::errors::RuntimeError;
 use crate::connectors::runtime::inputs::ContextReader;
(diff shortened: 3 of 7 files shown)


apollo-federation/src/connectors/models/problem_location.rs
```diff
@@ -12,6 +12,7 @@ pub enum ProblemLocation {
     ConnectQueryParams,
     SourceHeaders,
     ConnectHeaders,
+    IsSuccessMapping,
```
dylan-apollo:
```suggestion
    IsSuccess,
```

nit, but I think _all_ of these are mapping locations, so we don't need to suffix them

apollo-federation/Cargo.toml
```diff
@@ -51,6 +51,7 @@ shape = "0.5.2"
 form_urlencoded = "1.2.1"
 parking_lot = "0.12.4"
 mime = "0.3.17"
+once_cell = "1.21.3"
```
dylan-apollo: Looks like you're using `std::cell::OnceCell`, not `once_cell::OnceCell`.


```suggestion
```

apollo-federation/src/connectors/runtime/responses.rs
```diff
@@ -38,25 +39,87 @@ pub fn handle_raw_response(
...
+                (data.map_response(connector, debug_context, inputs), true)
+            } else {
+                (data.map_error(connector, debug_context, inputs), true)
+            }
+        }
+    }
+}
+
+// Allows us to pass a `OnceCell` for the Input parameters without worrying
+// about capturing borrowed response data in the initialization closure.
```
dylan-apollo:
```suggestion
/// Allows us to pass a [`OnceCell`] for the Input parameters without worrying
/// about capturing borrowed response data in the initialization closure.
```

apollo-federation/src/connectors/runtime/responses.rs
```diff
@@ -38,25 +39,87 @@ pub fn handle_raw_response(
...
+pub(super) struct InputCell<'ic, C: ContextReader> {
```
dylan-apollo: If I understand correctly, the only thing this cell/cache is gaining us is the ability to skip inputs calculation if the response from the underlying HTTP API is an error _and_ no `isSuccess` or `errors` are set.

The code is complex enough, that I think we should not do this optimization, especially since we don't know how frequently it'll actually be faster vs slower (there will be a negative impact on the baseline success mapping case). Maybe if we had a measure of the expected impact?

So instead, we could just do:

```rust
let inputs = data
                .key
                .inputs()
                .clone()
                .merger(&connector.response_variable_keys)
                .config(connector.config.as_ref())
                .context(context)
                .status(data.parts.status.as_u16())
                .request(&connector.response_headers, client_headers)
                .response(&connector.response_headers, Some(data.parts))
                .merge();
``` 

And then pass the `inputs: &IndexMap<String, Value>` to `is_success`, `map_response`, and `map_error`.

apollo-federation/src/connectors/runtime/responses.rs
```diff
@@ -38,25 +39,87 @@ pub fn handle_raw_response(
...
-    } else {
-        (
-            raw.map_error(connector, context, debug_context, client_headers),
-            false,
-        )
+
+    let inputs = input_cache.get(data);
+    let (res, apply_to_errors) = is_success_selection.apply_with_vars(&data.data, inputs);
+    let is_success_mapping_problems = aggregate_apply_to_errors(apply_to_errors)
+        .map(|problem| (ProblemLocation::IsSuccessMapping, problem));
```
dylan-apollo: Does the debugger need this new problem location directly supported somehow? I don't know how the client code works (cc @andrewmcgivery )

apollo-federation/src/connectors/runtime/responses.rs
```diff
@@ -38,25 +39,87 @@ pub fn handle_raw_response(
...
+    warnings.extend(is_success_mapping_problems);
+
+    // Add any warnings to debug context
+    if !warnings.is_empty() {
+        data.push_debug_warnings(connector, None, debug_context, &warnings);
+        return false;
```
dylan-apollo:
```suggestion
```

The warnings truly are warnings, so just because we warn on something doesn't mean the check failed.

For example, `$status->lt(400)->or($.isSuccess)->or($.isNotFailure)` if `isSuccess` isn't present, but `isNotFailure` is, this should still pass, but would generate warnings.

apollo-federation/src/connectors/runtime/responses.rs
```diff
@@ -67,218 +130,183 @@ pub enum RawResponse {
...
         self,
         connector: &Connector,
-        context: impl ContextReader,
         debug_context: &Option<Arc<Mutex<ConnectorContext>>>,
-        client_headers: &HeaderMap<HeaderValue>,
+        mut input_cache: InputCell<impl ContextReader>,
     ) -> MappedResponse {
-        match self {
-            RawResponse::Error { error, key } => MappedResponse::Error { error, key },
-            RawResponse::Data {
```
dylan-apollo, line 94: So this check was previously redundant, and is now only run once, right? Just want to make sure we're not subtly changing when/how mapping vs errors run.

apollo-federation/src/connectors/runtime/responses.rs
```diff
@@ -38,25 +39,87 @@ pub fn handle_raw_response(
...
+
+    let inputs = input_cache.get(data);
+    let (res, apply_to_errors) = is_success_selection.apply_with_vars(&data.data, inputs);
+    let is_success_mapping_problems = aggregate_apply_to_errors(apply_to_errors)
+        .map(|problem| (ProblemLocation::IsSuccessMapping, problem));
+    warnings.extend(is_success_mapping_problems);
+
+    // Add any warnings to debug context
+    if !warnings.is_empty() {
+        data.push_debug_warnings(connector, None, debug_context, &warnings);
```
dylan-apollo: Is this doing the right thing? Seems like it's pushing the entire request again along with these warnings? How does that work out on the debugger client-side?

I think we should be accumulating warnings on `data`, then pushing a _single_ debug response at the very end of all processing.

apollo-router/src/plugins/connectors/handle_responses.rs
```diff
@@ -1251,4 +1243,111 @@ mod tests {
...
+                &None,
+                supergraph_request.clone(),
+            )
+            .await
+            .mapped_response,
+        ])
+        .unwrap()
+        .response;
+        assert_eq!(res_expect_fail.body().data, Some(JsonValue::Null));
+        assert_eq!(res_expect_fail.body().errors.len(), 1);
```
dylan-apollo, line 1329: We should add tests for the debug stuff if we don't already have them somewhere (we might not...)

apollo-federation/src/connectors/runtime/responses.rs
Hunter-apollo: Yes. This check was redundant.
`RawResponse::Error`'s were never actually mapped and were just transparently passed through as `MappedResponse::Error{}`. That makes sense as these represent un-mappable runtime exceptions.

apollo-federation/Cargo.toml
Hunter-apollo: Ah yeah. I think this got auto-added by Intellij. Ill remove.

apollo-federation/src/connectors/runtime/responses.rs
Hunter-apollo: I suppose we dont save very much with the caching. Ill just pass through like you mentioned.

apollo-federation/src/connectors/runtime/responses.rs
andrewmcgivery: Yep. The rust debug code structs need to be updated to accept this new property and the frontend code will need to be updated with...

1. Update typescript types based on the new struct and problem location
2. Add a new view for this mapping (and sub-navigation item)
3. Couple of other small 1 liners to support the new problem location

This can all be done in a forward/backwards compatible way.

For the debug changes, up to @Hunter-apollo if you want to do those in this PR or a separate PR... but if you do go with a separate PR, please create a Jira item in your epic for it so it doesn't get lost. :)

apollo-federation/src/connectors/runtime/responses.rs
andrewmcgivery: Yea, given `is_success` is followed by `map_response` or `map_error` which also push a response, this is definitely going to result in a duplicate request.

Hunter-apollo:
Superceded by: https://github.com/apollographql/router/pull/8027 that resolves the many merge conflicts recently introduced.

PR #7917, opened 2025-07-17
Enforcing allowed features

DMallare:
- This is a poc for updates to license enforcement
- This is based on [the content in this doc](https://docs.google.com/document/d/13wh3NBmBnSPKBCH1f19eir-caWbz6Y9wKEHItXq6l1c/edit?tab=t.adap414o34k0#heading=h.1kx2ba8jc3vt)
- Jiras:
    - https://apollographql.atlassian.net/browse/BGS-437
    - https://apollographql.atlassian.net/browse/BGS-434

---

**Checklist**

- Tests added and passing[^4]
    - [x] Unit tests
    - [x] Integration tests

**Exceptions**

*Note any exceptions here*

**Notes**

(aaron while danielle is ooo) We feel confident in this work (given the integration tests and some manual e2e testing that's still underway (see [here](https://apollograph.slack.com/archives/C08M2T0QHBR/p1754659567023769) for the first round), but we want to make sure that:

- [ ] our changes to the plugin ecosystem seem judicious (we introduced a new macro, eg, but overall this file felt unwieldy--special attention here to make sure we didn't F anything up would be much appreciated)
- [ ] that the new jwks endpoint doesn't seem like a mistake (enables integration testing but might be unobvious to folks)
- [ ] that the allowed features integration tests seem sensible, comprehensive, and appropriate (ie, de-risks this work by making you/us feel confident--I feel pretty confident in them, but maybe I'm missing something)

[ROUTER-1412]: https://apollographql.atlassian.net/browse/ROUTER-1412?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apollo-router/tests/integration/allowed_features.rs
+++ b/apollo-router/tests/integration/allowed_features.rs
@@ -564,8 +564,8 @@ async fn canned_response_when_license_halted_with_valid_config_and_invalid_schem
  *  - since the license is not yet expired, the router should start with restricted features in use
  * */
 #[tokio::test(flavor = "multi_thread")]
-async fn router_starts_when_license_past_warn_at_but_not_expired_allowed_features_contains_feature()
-{
+async fn router_starts_when_license_past_warn_at_but_not_expired_allowed_features_contains_feature_subscriptions()
+ {
     let mut env = HashMap::new();
     env.insert(
         "APOLLO_TEST_INTERNAL_UPLINK_JWKS".to_string(),
@@ -584,6 +584,47 @@ async fn router_starts_when_license_past_warn_at_but_not_expired_allowed_feature
 
     router.replace_config_string("http://localhost:{{PRODUCTS_PORT}}", "localhost:4001");
     router.replace_config_string("http://localhost:{{ACCOUNTS_PORT}}", "localhost:4002");
+
+    router.start().await;
+    router.assert_started().await;
+}
+
+// In the CI environment we only install Redis on x86_64 Linux
+#[cfg(any(not(feature = "ci"), all(target_arch = "x86_64", target_os = "linux")))]
+#[tokio::test(flavor = "multi_thread")]
+async fn router_starts_when_license_past_warn_at_but_not_expired_allowed_features_contains_feature_entity_caching()
+ {
+    let mut env = HashMap::new();
+    env.insert(
+        "APOLLO_TEST_INTERNAL_UPLINK_JWKS".to_string(),
+        TEST_JWKS_ENDPOINT.as_os_str().into(),
+    );
+    let mut router = IntegrationTest::builder()
+        .config(
+            r#"
+            preview_entity_cache:
+              enabled: true
+              subgraph:
+                all:
+                  redis:
+                    urls: ["redis://127.0.0.1:6379"]
+                    ttl: "10m"
+                    required_to_start: true
+                subgraphs:
+                    connectors:
+                      enabled: true
+    "#,
+        )
+        .supergraph(PathBuf::from_iter([
+            "tests",
+            "fixtures",
+            "connectors",
+            "quickstart.graphql",
+        ]))
+        .env(env)
+        .jwt(JWT_PAST_WARN_AT_BUT_NOT_EXPIRED_WITH_COPROCESSORS_ENTITY_CACHING_TRAFFIC_SHAPING_IN_ALLOWED_FEATURES.to_string())
+        .build()
+        .await;
 
     router.start().await;
     router.assert_started().await;


apollo-router/src/axum_factory/axum_http_server_factory.rs
```diff
@@ -395,15 +395,18 @@ where
     let mut main_route = main_router::<RF>(configuration)
         .layer(decompression)
         .layer(middleware::from_fn_with_state(
-            (license, Instant::now(), Arc::new(AtomicU64::new(0))),
+            (license.clone(), Instant::now(), Arc::new(AtomicU64::new(0))),
```
DMallare, line 394: Theres a lot of new `clone()`s added in here because of the addition of the `allowed_features` field to `LicenseLimits` which is part of `LicenseState` that was previously `Copy`. One option using `Arc<LicenseState>` is found in https://github.com/apollographql/router/pull/7925

apollo-router/src/uplink/license_enforcement.rs
```diff
@@ -514,19 +515,59 @@ pub(crate) struct TpsLimit {
...
+    /// This represents a feature found in the license that the router does not recognize
+    Undefined(String),
+}
+
+impl From<&str> for AllowedFeature {
+    fn from(feature: &str) -> Self {
+        match feature {
+            "apq" => AllowedFeature::APQ,
+            "authentication" => AllowedFeature::Authentication,
+            other => AllowedFeature::Undefined(other.into()),
```
DMallare: Note: there is a TODO to finish this

apollo-router/src/uplink/license_enforcement.rs
```diff
@@ -514,19 +515,59 @@ pub(crate) struct TpsLimit {
     )]
     pub(crate) interval: Duration,
 }
+/// Allowed features for a License, representing what's available to a particular pricing tier
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize, Hash)]
+pub enum AllowedFeature {
+    /// Automated persistent queries
```
DMallare: Note: there is a TODO to finish the documentation for these

apollo-router/src/uplink/license_enforcement.rs
```diff
@@ -514,19 +515,68 @@ pub(crate) struct TpsLimit {
...
+    /// Persisted queries
+    PersistedQueries,
+    /// Rest connectors
+    RestConnectors,
+    /// Federated subscriptions
+    Subscriptions,
+    /// Traffic shaping plugin
+    TrafficShaping,
+    /// This represents a feature found in the license that the router does not recognize
+    Undefined(String),
```
lrlna: I don't think we should use the term `Undefined`. There is precedent in the standard library to use `Other`, for example in the [ErrorKind enum](https://doc.rust-lang.org/stable/std/io/enum.ErrorKind.html#variant.Other).

apollo-router/src/uplink/license_stream.rs
```diff
@@ -196,29 +196,33 @@ fn reset_checks_for_licenses(
     if halt_at > now {
         // Only add halt if it isn't immediately going to be triggered.
         checks.insert_at(
-            Event::UpdateLicense(LicenseState::LicensedHalt { limits }),
+            Event::UpdateLicense(LicenseState::LicensedHalt {
+                limits: limits.clone(),
+            }),
```
lrlna: We are not adding the allowed features when creating the `LicenseLimits` above. Should we also be adding them when we do a license reset?

apollo-router/src/router_factory.rs
```diff
@@ -631,7 +632,22 @@ pub(crate) async fn create_plugins(
...
+                    // If the license has an allowed_features claim, we know we're using a pricing
+                    // plan with a subset of allowed features
+                    if let Some(allowed_features) = $license.get_allowed_features() {
+                        if allowed_features.contains(&AllowedFeature::from($name)) {
+                            add_plugin!(name.to_string(), factory, plugin_config, full_config);
+                        } else {
+                            tracing::warn!(
+                                "{name} is a restricted feature that requires a license"
+                            );
+                        }
```
lrlna: When testing, we should check the behaviour for `Other` (`Undefined`) feature. It might be strange to send a warn for a feature that very clearly doesn't exist in the router.

apollo-router/src/axum_factory/axum_http_server_factory.rs
lrlna: Yes, using an `Arc` would be good. This is what we use for `schema` and `configuration`, for example.

apollo-router/src/uplink/license_enforcement.rs
lrlna: What are we doing with `Other` here? Is the `String` necessary for metrics? Or in other words, what's the expectation for features that router doesn't know about and how many of those do we expect from a typical response?

apollo-router/src/router_factory.rs
```diff
@@ -631,7 +632,22 @@ pub(crate) async fn create_plugins(
...
-                    add_plugin!(name.to_string(), factory, plugin_config, full_config);
+
+                    // If the license has an allowed_features claim, we know we're using a pricing
+                    // plan with a subset of allowed features
+                    if let Some(allowed_features) = $license.get_allowed_features() {
+                        if allowed_features.contains(&AllowedFeature::from($name)) {
+                            add_plugin!(name.to_string(), factory, plugin_config, full_config);
+                        } else {
+                            tracing::warn!(
+                                "{name} is a restricted feature that requires a license"
```
lrlna: So we are not initialising these plugins, right? the message should probably have something alluding to that, for example:
```suggestion
                                "{name} plugin is not registered, {name} is a restricted feature that requires a license"
```

apollo-router/src/uplink/license_enforcement.rs
```diff
@@ -514,19 +515,68 @@ pub(crate) struct TpsLimit {
     )]
     pub(crate) interval: Duration,
 }
+/// Allowed features for a License, representing what's available to a particular pricing tier
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize, Hash)]
+pub enum AllowedFeature {
```
lrlna: Parts of telemetry are a license-only feature (custom instruments). How is this going to be handled by the the allowlist?

lrlna (commented):
I had a few small comments, but overall I think the direction is good. We will definitely need a bunch of tests, especially integration tests, for this though.

apollo-router/src/router_factory.rs
DMallare: Yes that is correct we will not initialize it, thank you! Updated

apollo-router/src/uplink/license_enforcement.rs
DMallare: Thanks for calling this out, I was not aware of this. One idea is to add another item to this enum. I will explore this more to understand what parts of telemetry are license-only

apollo-router/src/uplink/license_enforcement.rs
DMallare: I saw the `String` primarily for use with logging so that, for example, if a customer tried to use a new feature with an older version of the router and didn’t see it take effect, the log message could guide them to the appropriate action - updating their router. 

newer jwt - older router
If the router encounters a feature without its own variant in `AllowedFeature` it will map to `Other`, the router will ignore it and log at the warn level that it does not recognize it and the customer should consider updating their version of the router.

older jwt - newer router
If the jwt `allowed_features`'s claim does not contain the feature they want to use with their router then that feature will show up in the license report as a restricted feature and the router either won't start up or reload.

The metrics piece & how many we expect are good call outs:

Do we want any metrics around this? 
My first thought is no - this variant will be created when a customer is using a feature unsupported by their version of the router
On the other hand, if we want to know what features customers are trying to use, albeit unsuccessfully, this might be something we want to record and keep on our radar to ensure we are communicating correctly to customers what min version of the router they need to use said feature(s).

How many `Other`'s do we expect to see? 
- This is dependent on how many new features we plan to introduce and how well we document the min version of the router they need (older router - newer jwt above)
- We may also see this if a customer has an outdated jwt (older jwt - newer router above)

apollo-router/src/axum_factory/axum_http_server_factory.rs
DMallare: Added the update to use `Arc` to this pr - will close https://github.com/apollographql/router/pull/7925

apollo-router/src/uplink/license_enforcement.rs
DMallare: Updated to `Other(String)`

apollo-router/src/uplink/license_stream.rs
DMallare: Thank you for catching that! Added it

apollo-router/src/uplink/license_enforcement.rs
lrlna: > if a customer tried to use a new feature with an older version of the router and didn’t see it take effect, the log message could guide them to the appropriate action - updating their router.

Are they manually opting in to `allowed_features` in GraphOS UI? If not, it might be a bit confusing reading a message saying "`new_feature` you're not using nor have you opted into is not available on your version of the router". We will also not be able to add a message in the router saying in which version this new feature is available, unless we literally backport to every version of the router after this change lands. So that information, that is minimum router version, will then have to live somewhere in GraphOS and make its way to the router along with the actual feature name. If they are opting into `allowed_features`, then this makes a bit more sense. 

> If the jwt allowed_features's claim does not contain the feature they want to use with their router then that feature will show up in the license report as a restricted feature and the router either won't start up or reload.

I did not yet see the logic for this in the code here. Right now it seems there is only a `warn`, and a plugin does not load (which is a nice experience - you can still use the router - but I am not sure if that's the intention). Or is that more for `configuration` based entitlements? Speaking of, while I am usually a proponent of separating PRs into smallest possible units of work, it might be easier to bring this PR and #7939 together, especially when you're testing. You are going to want to test the intersection of what happens when there isn't any config for a licensed feature (i.e. default config) but the plugin still shouldn't load because there isn't a license etc.

> On the other hand, if we want to know what features customers are trying to use, albeit unsuccessfully, this might be something we want to record and keep on our radar to ensure we are communicating correctly to customers what min version of the router they need to use said feature(s).

That sounds like we should have a metric recording the `Other` and the current router version. This could form a helpful UI in GraphOS letting users know they should update their router and would also just be helpful information for us to know.

lrlna:
@DMallare @aaronArinder I'll let you all continue working on this for the time being, but give me a shout once this is out of draft and ready for review again.

apollo-router/src/uplink/license_enforcement.rs
DMallare: > Are they manually opting in to allowed_features in GraphOS UI?

We may have this eventually but for now we don't the ability to do th is in GraphOS UI.  I am thinking we can add the warning to the jwt.

> I did not yet see the logic for this in the code here

My apologies, as you had mentioned it is in #7939. I brought the contents of #7939 to this PR to make it easier to understand everything as a whole

> That sounds like we should have a metric recording the Other and the current router version.

This sounds very interesting! I will think about how we can add this and what it would look like in terms of UI for our users

apollo-router/src/uplink/license_enforcement.rs
```diff
@@ -74,9 +75,11 @@ pub(crate) struct Claims {
     #[serde(deserialize_with = "deserialize_epoch_seconds", rename = "haltAt")]
     /// When to halt the router because of an expired license
     pub(crate) halt_at: SystemTime,
-    /// TPS limits. These may not exist in a Licnese; if not, no limits apply
+    /// TPS limits. These may not exist in a License; if not, no limits apply
     #[serde(rename = "throughputLimit")]
     pub(crate) tps: Option<TpsLimit>,
+    /// Set of allowed features. These may not exist in a License; if not, all features are enabled
+    pub(crate) allowed_features: Option<Vec<AllowedFeature>>,
```
aaronArinder: probs worth explaining why all features are enabled if there's no allowedFeatures claim (ie, middle versus end state of this work)

apollo-router/src/router_factory.rs
```diff
@@ -631,7 +632,22 @@ pub(crate) async fn create_plugins(
...
+                    // If the license has an allowed_features claim, we know we're using a pricing
+                    // plan with a subset of allowed features
+                    if let Some(allowed_features) = $license.get_allowed_features() {
+                        if allowed_features.contains(&AllowedFeature::from($name)) {
+                            add_plugin!(name.to_string(), factory, plugin_config, full_config);
+                        } else {
+                            tracing::warn!(
+                                "{name} plugin is not registered, {name} is a restricted feature that requires a license"
+                            );
+                        }
```
aaronArinder: this will come out in testing, but the allowed features are probably named differently than their plugins (some get prefixed with a `apollo.` for example); so, we'll want to make sure there's some mapping between the two that's testable

apollo-router/src/router_factory.rs
```diff
@@ -631,7 +632,22 @@ pub(crate) async fn create_plugins(
...
+                    // If the license has no allowed_features claim, we're using a pricing plan
+                    // that should have the plugin enabled regardless
+                    } else {
+                        add_plugin!(name.to_string(), factory, plugin_config, full_config);
```
aaronArinder: we'll also want to spend some time (probably through tests) in figuring out which plugins are potentially added in other ways; the telemetry plugin gets some special handling, but is that the only one?

apollo-router/src/router_factory.rs
```diff
@@ -640,21 +656,22 @@ pub(crate) async fn create_plugins(
...
                 Some(
                     apollo_plugins_config
                         .remove($name)
                         .unwrap_or(Value::Object(Map::new()))
-                )
+                ),
+                $license
             );
         };
     }
```
aaronArinder: I'm starting to worry that I led you down the wrong path with this approach; mandatory plugins should probably be _mandatory_ without the risk of not being added if they're not in the allowed features list. Part of the trouble is that the macros are fairly coupled. Not sure how you want to tackle it, but the main requirement is that mandatory plugins always be added and the main problem with our current approach is that _all_ plugins are gated on being in the allowed features list but certain plugins aren't going to be in that list (like `csrf`)

DMallare: Thank you for bringing that up! I explored a few options:
1. Using a boolean with `add_apollo_plugin` - this is less expressive but less code duplication
2. Splitting out `add_apollo_plugin` into `add_mandatory_apollo_plugin_inner` and `add_optional_apollo_plugin_inner` - more expressive, but we get code duplication
3. Using enum-like tokens with `add_apollo_plugin` - more expressive, will still have code duplication

For simplicity and improved readability I am going with option 2 for now - we can discuss further if we prefer another approach

apollo-router/src/router_factory.rs
DMallare: Are we sure they get prefixed with `apollo`? [We add the prefix in our macro here](https://github.com/apollographql/router/blob/926f6312e301e84f281eb23bd10154eabfc248b3/apollo-router/src/router_factory.rs#L617C13-L617C14) so after that step we would have `apollo.apollo.some-plugin-name`.

I agree with the mapping part. My plan was to write unit tests to ensure each plugin associated with an allowed feature was added when included in the config + `allowed_features` set and also no added if not included in the config and/ or not in the `allowed_features` set. I also want to add in the doc comments for the `AllowedFeature` enum what plugin or config etc. it maps to for documentation purposes

apollo-router/src/uplink/license_enforcement.rs
DMallare: Added a comment

apollo-router/src/plugins/connectors/tests/connect_on_type.rs
```diff
@@ -50,6 +50,7 @@ async fn basic_batch() {
         Default::default(),
         None,
         |_| {},
+        None,
```
DMallare, line 53: You will see a lot of `None`'s added so that a license could be injected into the `execute` function

apollo-router/src/router_factory.rs
DMallare, line 1: Make sure to expand this to see the changes to this file

aaronArinder:
dang it, hit `ready for review` when I was trying to find a shortcut to circleci; putting it back in draft

lrlna:
@DMallare @aaronArinder I am away next week, so if this needs to land into the next release, I'll need to start reviewing it soon.

aaronArinder:
@lrlna: goal is to have this out of draft by my EoD (your morning); we've done an initial e2e test (more this week) and I'm doing some minor things like cleaning up the git history, etc, before opening it up for review (I'll create a ticket in the backlog and let you know when it's good for review)

apollo-router/src/uplink/testdata/license.jwks.json
```diff
@@ -0,0 +1,10 @@
+{
+  "keys": [
+    {
+      "kty": "oct",
+      "alg": "HS256",
+      "k": "bWFrZV9hX2xvbmdfc2VjcmV0X2Zvcl9yZmNfNzUxOF8yNTZfYml0c19yZXF1aXJlbWVudF9ibGFo",
+      "use": "sig"
+    }
+  ]
+}
```
aaronArinder, line 10: dummy jwks file

.circleci/config.yml
```diff
@@ -365,7 +365,9 @@ commands:
           environment:
             # Use the settings from the "ci" profile in nextest configuration.
             NEXTEST_PROFILE: ci
-          command: cargo xtask test --workspace --locked --features ci,snapshot
+          command: |
+            cargo xtask test --workspace --locked --features ci,snapshot
+            cargo xtask license-enforcement-test --workspace --locked --features ci,snapshot,test-jwks
```
aaronArinder: I'd like to figure out a way to use one command for everything or to reuse the test command rather than introduce a new xtask command, but leaving this as-is for the time-being while reviews are underway; let us know what y'all think about this ^

apollo-router/src/plugins/connectors/tests/connect_on_type.rs
lrlna: I see both `authentication` and `authorization` tests use `AllowedFeature::Authorization`/`AllowedFeature::Authentication`. Connectors tests should probably mirror and use `AllowedFeature::Connectors`. In fact, I am a bit concerned this is passing given we are passing it `None`, and connectors is a license feature, no?

apollo-router/src/uplink/license_enforcement.rs
```diff
@@ -555,6 +799,19 @@ impl LicenseState {
...
     }
 
+    pub(crate) fn get_allowed_features(&self) -> Option<HashSet<AllowedFeature>> {
+        match self {
+            LicenseState::Licensed { limits }
+            | LicenseState::LicensedWarn { limits }
+            | LicenseState::LicensedHalt { limits } => match limits {
+                Some(limits) => limits.allowed_features.clone(),
+                // NB: this may change once all licenses have an `allowed_features` claim
+                None => None,
```
lrlna: Should we have a `Default` implemented on `allowed_features` that enables all of the features for this particular case for the time being?

lrlna: yeaaa the more I read through this file, the more I think you should have a wrapper around `Vec<AllowedFeature>` (`AllowedFeatures`) with a default with all of the features. You can initiliase the license state with that should the license not have the allowed_features claim, and remove that once all licenses have the allowed feature claim. That will be easier to maintain in the router, otherwise you're going to have to re-write a lot more `if`/`else` statements.

apollo-router/src/uplink/testdata/license.jwks.json
lrlna: can you explain the reasoning behind creating a separate feature flagged test? I am not sure I follow just yet.

apollo-router/src/uplink/license_stream.rs
```diff
@@ -549,7 +582,8 @@ mod test {
                     aud: OneOrMany::One(Audience::Offline),
                     warn_at: SystemTime::now(),
                     halt_at: SystemTime::now(),
-                    tps: Default::default()
+                    tps: Default::default(),
+                    allowed_features: None,
```
lrlna: following up on the comemnt before, definitely think this should also be `Default::default()` with all of the allowed features.

apollo-router/src/router_factory.rs
```diff
@@ -961,4 +1046,552 @@ mod test {
...
+                "#
+            }
+            _ => panic!("This function does not contain config for plugin: {plugin}"),
+        }
+    }
+
+    #[tokio::test]
+    #[rstest]
+    #[case::empty_allowed_feature_set(Some(HashSet::new()))]
+    #[case::allowed_features_none(None)]
```
lrlna: Again, too confusing to reason about - when I immediately read this, I think nothing should work, but it's the opposite.

apollo-router/src/router_factory.rs
```diff
@@ -961,4 +1046,552 @@ mod test {
...
+    async fn test_oss_plugins_with_allowed_features(
+        #[case] plugin: &str,
+        #[case] allowed_features: Option<HashSet<AllowedFeature>>,
+    ) {
+        /*
+         * GIVEN
+         *  - a license with an allowed features claim
+         *  - a valid config including valid config for the optional plugin
+         *  - a valid schema
+         * */
```
lrlna: I am confused by the wording here - we are testing one of the feature claims being allowed? Or all of them? and making sure the OSS plugins (i.e. plugins that don't need a license) are loaded? 

> - a license with an allowed features claim
> - a valid config including valid config for the optional plugin
         
Also - what config, what feature? And I feel like this wording get more verbose with each test, and I get more and more confused. I think we need to be a bit more specific and to the point.

apollo-router/src/router_factory.rs
```diff
@@ -961,4 +1046,552 @@ mod test {
...
+        Some(HashSet::from_iter(vec![AllowedFeature::Coprocessors, AllowedFeature::DemandControl]))
+    )]
+    #[case::conectors(
+        "connectors",
+        Some(HashSet::from_iter(vec![AllowedFeature::Coprocessors, AllowedFeature::Connectors]))
+    )]
+    async fn test_optional_plugin_with_allowed_features_containing_the_feature(
+        #[case] plugin: &str,
+        #[case] allowed_features: Option<HashSet<AllowedFeature>>,
+    ) {
```
lrlna: Do any of these tests check that when a something like DemandControl is not part of the allowed features that the plugin is not actually added? I cannot quite find that in these tests.

apollo-router/src/state_machine.rs
```diff
@@ -890,6 +1330,45 @@ mod tests {
...
+                        })
+                    })),
+                    Shutdown
+                ])
+            )
+            .await,
+            Ok(())
+        );
+        assert_eq!(shutdown_receivers.0.lock().len(), 1);
+    }
```
lrlna, line 1591: We should do a test here with a reload, i.e. you had a license that had all the features, but you now have a license with none of the features.

lrlna: Either here in the new integration test file I see you've also added. Might be easier in the integration test, in fact.

apollo-router/src/plugins/connectors/tests/connect_on_type.rs
DMallare: I ended up just having the default license used with the `TestHarness` be a license that allows everything. I can remove `AllowedFeature::Authorization`/`AllowedFeature::Authentication` from the `authentication` and `authorization` tests. I will remove those to avoid confusion

lrlna (commented):
also! this needs a changeset 😇

apollo-router/src/uplink/license_enforcement.rs
DMallare: Agreed. Making some updates now and will make an easier to understand default that defaults to allowing all features

apollo-router/src/plugins/connectors/tests/connect_on_type.rs
DMallare: Removed these from all tests - `TestHarness` now uses an unrestricted license by default

apollo-router/src/router_factory.rs
DMallare: I agree, this is not reader friendly. Updated this - we now have an `AllowedFeatures` enum with variants `Unrestricted` and `Restricted(HashSet<AllowedFeature>)`

apollo-router/src/router_factory.rs
DMallare: Updated test names to make them more succinct. 
Improved the comments within the tests to make them easier to understand.

DMallare: The goal of this test is to make sure oss plugins are added regardless of allowed features status (restricted, unrestricted, etc.)

apollo-router/src/router_factory.rs
DMallare: I do have a test that checks that optional plugins are not added, even if they are present in the configuration, if they are not contained within the allowed features set - I will make that test more more explicit that its checking this.

DMallare: Updated test name: `test_optional_plugin_not_added_with_restricted_allowed_features` and updated the comments within the test to explain more clearly what we are testing

apollo-router/src/uplink/license_enforcement.rs
DMallare: Made some updates to this: We now have an `AllowedFeatures` enum with variants `Unrestricted` and `Restricted(HashSet<AllowedFeature>)`.
Default is `Unrestricted`

apollo-router/src/uplink/license_stream.rs
DMallare: Default is now `Unrestricted` (allow everything). 
Updated the `None`s here to use the default

apollo-router/src/uplink/testdata/license.jwks.json
DMallare: We needed a new key to be able to sign our test jwts. Otherwise we were hitting errors with validating the signature on the jwts. 
We needed to use the feature flag because we don't want to use the test jwk outside of the license enforcement tests that use it.

apollo-router/src/state_machine.rs
DMallare: Adding a test

DMallare: Also adding a test for default license limits to reflect the changes made with `AllowedFeatures`

DMallare: I want to add more tests tho... specifically checking what happens when we start unlicensed then switch to a license and start using features within that license - will do integ tests for that

apollo-router/src/state_machine.rs
```diff
@@ -253,6 +253,14 @@ impl<FA: RouterSuperServiceFactory> State<FA> {
 
                             Some(new_state)
                         }
+                        Err(ApolloRouterError::LicenseViolation) => {
```
DMallare: Question: is this the behavior we want? We were not doing this previously on reload but if we do not do this then on reload we won't be stopping the router when a license violation occurs

apollo-router/src/state_machine.rs
```diff
@@ -876,18 +1246,121 @@ mod tests {
...
+                ])
+            )
+            .await,
+            Err(ApolloRouterError::LicenseViolation)
+        );
+        assert_eq!(shutdown_receivers.0.lock().len(), 1);
+    }
+
+    #[test(tokio::test)]
+    async fn restricted_licensed_with_unrestricted_allowed_features_reload_with_empty_feature_set()
```
DMallare: Added this test for reload - I want to test the opposite works as well:
If I start unlicensed and not using restricted features and then reload with a licenses with subscriptions for example and then add a config with subscriptions then I should see success - will do this as an integ test. Tried to add it here but as an integ test should be much easier

apollo-router/tests/integration/allowed_features.rs
```diff
@@ -0,0 +1,366 @@
...
+ *  - a valid license whose `allowed_features` claim is empty (contains no features)
+ *  - an empty configuration
+ *  - the license is updated, the new jwt contains the feature subscriptions in its
+ *    allowed_features claim
+ *  - the config is updated to include subscriptions
+ *
+ * THEN
+ *  - the router should reload successfully with no license violations
+ * */
+#[tokio::test(flavor = "multi_thread")]
```
DMallare, line 469: Starting to create an integ test for reload here - need to figure out how to reload with a new license. Currently these integ tests are using an env var to pass in the license jwt

apollo-router/src/state_machine.rs
aaronArinder: we do want it, yeah; this'll be key to enforcement

apollo-router/src/uplink/license_enforcement.rs
```diff
@@ -702,43 +848,55 @@ impl License {
         JWKS.get_or_init(|| {
             // Strip the comments from the top of the file.
             let re = Regex::new("(?m)^//.*$").expect("regex must be valid");
+            #[cfg(not(feature = "test-jwks"))]
             let jwks = re.replace(include_str!("license.jwks.json"), "");
+            #[cfg(feature = "test-jwks")]
+            let jwks = re.replace(include_str!("testdata/license.jwks.json"), "");
```
goto-bus-stop: Integration tests spawn a router subprocess. Could this use an environment variable at runtime instead of a separate feature? eg. `TEST_INTERNAL_APOLLO_UPLINK_JWKS=path/to/testdata/license.jwks.json`?

I think running different test suites based on feature flagging is going to be brittle to maintain going forward.

apollo-router/src/router_factory.rs
```diff
@@ -653,8 +706,14 @@ pub(crate) async fn create_plugins(
     }
 
     macro_rules! add_optional_apollo_plugin {
+        ($name: literal, $license: expr) => {
```
goto-bus-stop: optional nit: To align with other macros it might be better to just use the `license` variable from the surrounding scope here directly instead of passing it in. The difference between `add_oss_apollo_plugin` and `add_optional_apollo_plugin` is already shown in the macro names

apollo-router/src/state_machine.rs
```diff
@@ -173,8 +173,8 @@ impl<FA: RouterSuperServiceFactory> State<FA> {
             } => {
                 // When we get an unlicensed event, if we were licensed before then just carry on.
                 // This means that users can delete and then undelete their graphs in studio while having their routers continue to run.
-                if new_license == Some(LicenseState::Unlicensed)
```
goto-bus-stop, line 176: optional nit: might be a bit nicer to use `new_license.as_deref()` and `**license` here instead of `.into()`, which allocates an Arc only to compare it with a value

.changesets/feat_feature_enforcement.md
```diff
@@ -0,0 +1,12 @@
+### Enforcing allowed features ([PR #7917](https://github.com/apollographql/router/pull/7917))
+
+<!-- start metadata -->
+
+<!-- [ROUTER-1412] -->
+---
+The router now enforces feature access based on the `allowed_features` claim in the license.
```
goto-bus-stop: The changeset could use clarification on how it affects users. as a customer reading this i'm not sure what it means for me (eg. does it depend on what kind of plan i have? is this going to enforce limitations that i previously maybe relied on?).

apollo-router/src/uplink/license_enforcement.rs
goto-bus-stop: Here is where the integration testing harness can set env variables for the router subprocess. It doesn't have methods to do so right now, but you could add `env: HashMap<String, String>` to the IntegrationTest constructor, or even have a hardcoded `.with_test_jwks()` method if it's easier
https://github.com/apollographql/router/blob/8ac0bd955bcb021c0894619769f1a431ebfbeb78/apollo-router/tests/common.rs#L713

apollo-router/src/router_factory.rs
DMallare: Updated

apollo-router/src/state_machine.rs
DMallare: Good call, thank you! Updated

goto-bus-stop (approved):
as discussed with @aaronArinder, will pre-approve so you can merge it after europe EOD, after changing the feature flag thing 😇  thanksss

.circleci/config.yml
```diff
@@ -365,7 +365,12 @@ commands:
           environment:
             # Use the settings from the "ci" profile in nextest configuration.
             NEXTEST_PROFILE: ci
-          command: cargo xtask test --workspace --locked --features ci,snapshot
+          command: |
+            cargo xtask test --workspace --locked --features ci,snapshot
+
+            # run just the allowed_features integration tests with a dummy JWKS endpoint
+            APOLLO_TEST_INTERNAL_UPLINK_JWKS=true cargo xtask test --workspace --locked --features ci,snapshot,test-jwks integration_tests
```
goto-bus-stop: tell me if i'm missing something, but it should be possible to do everything in a single test run by configuring the env var only inside the integration tests that need it

.changesets/feat_feature_enforcement.md
DMallare: Updated, thank you!

apollo-router/src/uplink/license_enforcement.rs
aaronArinder: good call out; and, good thinking for isolating this to just the relevant test; pushed a commit that I think does the dirt, but let us know

PR #7920, opened 2025-07-18
Redis metrics

BrynCooke:
What was added

This PR adds some best effort metrics until we can have changes in Fred to allow better metrics collection:  https://github.com/aembke/fred.rs/issues/357

The latency metrics are marked as experimental because we will change them if we have an upstream change in fred.

**Connection and performance metrics:**
  - `apollo.router.cache.redis.connections`: Number of active Redis connections
  - `apollo.router.cache.redis.command_queue_length`: Commands waiting to be sent to Redis
  - `apollo.router.cache.redis.commands_executed`: Total number of Redis commands executed
  - `apollo.router.cache.redis.redelivery_count`: Commands retried due to connection issues
  - `apollo.router.cache.redis.errors`: Redis errors by type (auth, timeout, io, etc.)

**Experimental performance metrics:**
  - `experimental.apollo.router.cache.redis.network_latency_avg`: Average network latency to Redis
  - `experimental.apollo.router.cache.redis.latency_avg`: Average Redis command execution time  
  - `experimental.apollo.router.cache.redis.request_size_avg`: Average request payload size
  - `experimental.apollo.router.cache.redis.response_size_avg`: Average response payload size

  Technical implementation:
  - Added metrics.rs module with RedisMetricsCollector and weighted average calculation
  - Metrics collection runs on a configurable interval (default 30s via metrics_interval config)
  - Uses atomic counters for thread-safe metric updates
  - Implements proper unit conversions (seconds to microseconds for precision)
  - All metrics include kind labels for proper categorization

  Configuration:
  - Added metrics_interval configuration option to RedisCache with 30s default
  - Enabled metrics feature flag in the fred Redis client dependency

  Testing:
  - Unit tests for weighted sum calculations and unit conversions
  - Integration tests using Redis mocks to verify metrics collection
  - Tests verify proper initialization and metric update behavior

  Why this was needed

  Redis cache performance monitoring was missing from the Apollo Router observability stack. These metrics provide visibility into:
  - Cache performance and latency characteristics
  - Network connectivity health via redelivery tracking
  - Resource utilization through request/response size monitoring
  - Command queue health to detect backpressure issues

---

**Checklist**

- Tests added and passing[^4]
    - [x] Unit tests
    - [x] Integration tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-1378]: https://apollographql.atlassian.net/browse/ROUTER-1378?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.changesets/feat_bryn_redis_metrics.md
```diff
@@ -0,0 +1,46 @@
...
+  - `experimental.apollo.router.cache.redis.response_size_avg`: Average response payload size
+
+> **Note**: The experimental metrics may change in future versions as we improve the underlying Redis client integration.
+
+#### Configuration
+
+You can configure how often metrics are collected using the `metrics_interval` setting:
+
+```yaml
+preview_entity_cache:
```
carodewig: Nit: I imagine this would be more relevant to other customers if it referenced the query plan cache. Also, we probably don't want to be directing people to this entity cache preview :)

apollo-router/src/cache/redis.rs
```diff
@@ -470,6 +545,7 @@ impl RedisCacheStorage {
                     .map_err(|e| {
                         if !e.is_not_found() {
                             tracing::error!(error = %e, "redis get error");
+                            self.record_error(&e);
                         }
                         e
                     })
```
carodewig: Thoughts on this instead? Since `record_error` also logs an error and checks that the error isn't `NotFound`. This could be applied to the other cases as well
```suggestion
                    .inspect_err(|e| self.record_error(&e))
```

BrynCooke: Addressed in [0f22c8c](https://github.com/apollographql/router/pull/7920/commits/0f22c8caf1e0afb4ee57d7046b06716fa95a6ed1)

apollo-router/tests/integration/entity_cache.rs
```diff
@@ -282,3 +284,245 @@ async fn invalidate_with_endpoint() {
...
+        },
+        "reviews": {
+            "headers": {"cache-control": "public"},
+            "entities": reviews,
+        },
+    })
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn test_cache_metrics() {
```
goto-bus-stop, line 380: Actually, Caroline's point about the changeset applies here too. it'd be better to (also) have an integration test using query plan caching because response caching won't be using Redis in the future...

PR #7924, opened 2025-07-18
[Connectors] Validation for `isSuccess` for `@source` and `@connect`

Hunter-apollo:
Adds validation for the `isSuccess` argument for both `@source` and `@connect` directives.

---

**Checklist**

- Tests added and passing[^4]
    - [ ] Unit tests
    - [ ] Integration tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-federation/src/connectors/validation/connect.rs
```diff
@@ -75,6 +76,7 @@ struct Connect<'schema> {
     selection: Selection<'schema>,
     http: Http<'schema>,
     errors: Errors<'schema>,
+    is_success: Option<IsSuccessArgument<'schema>>,
```
andrewmcgivery, line 130: 2 questions here...

1. Why is this one wrapped in `Option`? `errors` is also optional but it isn't wrapped in an `Option` here
2.`IsSuccessArgument` versus `IsSuccess`... can we be consistent with how the rest of the properties here are named?

apollo-federation/src/connectors/validation/connect.rs
```diff
@@ -183,12 +185,16 @@ impl<'schema> Connect<'schema> {
                 },
                 schema,
             ))
+            .and_try(
+                IsSuccessArgument::parse_for_connector(coordinate, schema).map_err(|err| vec![err]),
```
andrewmcgivery, line 241: Consistency: Is there a reason this is called `parse_for_connector` when the other items here are just `parse`?

apollo-federation/src/connectors/validation/expression.rs
```diff
@@ -262,6 +262,31 @@ impl<'schema> Context<'schema> {
             has_response_body: false,
         }
     }
+
+    /// Create a context valid for expressions within the `isSuccess` property of
+    /// `@source` or `@connect` directives
+    pub(super) fn for_success_conditional(
```
andrewmcgivery: This is what was asking about on confluence/chat... you should be able to use `for_source_response` and `for_connect_response` here since `isSuccess` should have the exact same variables since it is acting on responses just like `errors`.

apollo-federation/src/connectors/validation/connect.rs
Hunter-apollo: There is separate handling for `for_connector` and `for_source`.

apollo-federation/src/connectors/validation/expression.rs
Hunter-apollo: `for_source_response` and `for_connect_response` include the `Request` data as well which I do not believe is required for the success conditional. The new context does not include the `request`

apollo-federation/src/connectors/validation/connect.rs
Hunter-apollo: 1. The `Errors` contains has optional nested fields that are set to none if it doesnt exist. This is a top-level optional argument, so I like to express that in the top level struct. Making it non-optional would result in needing to extract out some nested optional field which feels like added indirection to me.
2. I personally find the `Argument` suffix to be clearer as it identifies that this represents a directive argument.

apollo-federation/src/connectors/validation/expression.rs
andrewmcgivery: That doesn't answer why `isSuccess` is unique here. It should be identical to `errors` here and you get this for free. Why specifically exclude `request`?

Hunter-apollo: What `request`-bound data would you use to determine success? Maybe my imagination is just failing me, but I dont have any good use cases I can think of for that. Without a good use case I think it is best to omit.

andrewmcgivery: Right now the only thing on request is `headers`. Perhaps we don't know the exact use case now but you're introducing a limitation arbitrarily that does not need to exist where there is existing code that you can just reuse.

> Without a good use case I think it is best to omit

I might agree with this if the ask was for you to add new code, but that isn't the ask. Quite the opposite... the ask is to _not_ add new code.

The ask is use the existing code so that it is consistent with `errors` and what `selection` will eventually use. There is no reason for `isSuccess` to be different here when it is acting on the exact same data. You're actually adding unnecessary duplicative code by _not_ using the existing functions.

Hunter-apollo: It is not necessarily true that isSuccess should have access to the same vars as `selection` and/or `errors`. The duplication is minor and has the benefit of decoupling these fields. Using the same context for all could result in unintended changes to the behavior of `isSuccess` when one of the others changes. 

If you are unwilling to approve without the change I would rather change than hold up the PR, but I dont feel this duplication is unnecessary.

andrewmcgivery: > It is not necessarily true that isSuccess should have access to the same vars as selection and/or errors.

Your other PR passes the exact same inputs (https://github.com/apollographql/router/blob/c27d1efeace405b2e0b4827c12df7d35eb4fa20a/apollo-federation/src/connectors/runtime/responses.rs#L45-L64
) into `isSuccess` as `selection` and `errors`... so it is absolutely true. They use the exact same inputs at run time... why would they not use the same inputs at validation time?

andrewmcgivery: I'll also add as a clarification that the purpose of this code is not to "decide" what the user "should" have access to. The purpose is to state what is "in scope" at that point in the lifecycle.

`isSuccess`, `selection`, and `errors` are all in the same part of the lifecycle and always will be so they will always have the same variables "in scope".

apollo-federation/src/connectors/validation/errors.rs
```diff
@@ -325,3 +327,82 @@ impl Display for ErrorsExtensionsCoordinate<'_> {
...
+            coordinate,
+            node,
+        }))
+    }
+
+    /// Check that only available variables are used, and the expression results in a boolean
+    pub(crate) fn type_check(self, schema: &SchemaInfo<'_>) -> Result<(), Message> {
+        let context = match self.coordinate.coordinate {
+            ErrorsCoordinate::Source { .. } => {
+                &Context::for_source_response(schema, &self.node, Code::InvalidErrorsMessage)
```
andrewmcgivery: `InvalidErrorsMessage` is probably the wrong error code here 😅

apollo-federation/src/connectors/validation/errors.rs
```diff
@@ -325,3 +327,82 @@ impl Display for ErrorsExtensionsCoordinate<'_> {
...
+            }
+            ErrorsCoordinate::Connect { connect } => &Context::for_connect_response(
+                schema,
+                connect,
+                &self.node,
+                Code::InvalidErrorsMessage,
```
andrewmcgivery: Same here, `InvalidErrorsMessage` is probably the wrong error code here 😅

apollo-federation/src/connectors/validation/errors.rs
andrewmcgivery: `InvalidIsSuccess`?

apollo-federation/src/connectors/validation/errors.rs
andrewmcgivery: `InvalidIsSuccess`?

apollo-federation/src/connectors/validation/snapshots/validation_tests@is_success__is_success_invalid_connect.graphql.snap
```diff
@@ -0,0 +1,14 @@
+---
+source: apollo-federation/src/connectors/validation/mod.rs
+expression: "format!(\"{:#?}\", result.errors)"
+input_file: apollo-federation/src/connectors/validation/test_data/is_success/is_success_invalid_connect.graphql
+---
+[
+    Message {
+        code: InvalidIsSuccess,
+        message: "In `@connect(isSuccess:)` on `Query.users`: $request is not valid here, must be one of $status, $response, $config, $env",
```
andrewmcgivery: I assume this is an old snapshot that needs to be updated? :)

apollo-federation/src/connectors/validation/snapshots/validation_tests@is_success__is_success_invalid_sources.graphql.snap
```diff
@@ -0,0 +1,14 @@
+---
+source: apollo-federation/src/connectors/validation/mod.rs
+expression: "format!(\"{:#?}\", result.errors)"
+input_file: apollo-federation/src/connectors/validation/test_data/is_success/is_success_invalid_sources.graphql
+---
+[
+    Message {
+        code: InvalidIsSuccess,
+        message: "In `@source(name: \"using_disallowed_var\" isSuccess:)`: $request is not valid here, must be one of $status, $response, $config, $env",
```
andrewmcgivery: Old snapshot?

Hunter-apollo: Yup

PR #7926, opened 2025-07-18
Connectors prerequisites for abstract types: `ConnectSpec` version gating, `->` method shape checking, and more

benjamn:
This `benjamn/JSONSelection-abstract-types` branch will be a target for a chain of other PRs that ultimately allow connectors to handle fields with abstract GraphQL types (unions and interfaces).

The branch begins with a single commit that reworks `PrettyPrintable::pretty_print` to propagate the `inline` parameter correctly down through recursive printing, which will be useful later for some parsing tests.

Feel free to review what's here and leave comments, and I will list the various dependent PRs as I open them here:
- [x] https://github.com/apollographql/router/pull/7927
- [x] https://github.com/apollographql/router/pull/7932
  - Note: the above two PRs were factored out of my previous PR #7623, which has now been simplified to just #7933.
- [x] #7933
  - Not abstract-types-related, but demonstrates combined usage of the previous two PRs (`ConnectSpec` version gating and `->` method shape checking)

_We may merge this branch into `dev` once the above PRs land and CI checks pass_

---

_The PRs below are not expected to be fully reviewed before merging into `dev`_

- [ ] #7945
  - This internal refactoring PR is important for abstract types because it enables the `...` spread operator that allows the output of a `... type->match(…)` to be merged with other fields
- [ ] #7957
  - This PR actually simplifies the grammar and adds the `...` spread token


apollo-federation/src/connectors/json_selection/pretty.rs
```diff
@@ -207,38 +214,40 @@ impl PrettyPrintable for LitExpr {
...
 
                 let mut is_first = true;
                 for (key, value) in map {
                     if is_first {
                         is_first = false;
-                    } else if inline {
-                        result.push_str(", ");
                     } else {
-                        result.push_str(",\n");
+                        result.push(',');
```
naomijub, line 230: ehy not `enumerate()` and check if is `0`?

naomijub:
Draft LGTM

dylan-apollo:
Following up on a conversation we had—we should endeavor to remove `ConnectSpec::default` and `ConnectSpec::latest`, since they open the door to a class of bugs where the user opts in to a new version, but _some_ of their Connector doesn't get updated. Or worse, a particular piece is statically checked with one spec, but executed in the runtime with a different spec. Same thing with `StringTemplate::from_str`.

It could be okay to leave them in under a `#[cfg(test)]` to avoid a bunch of code updating for now, since most tests won't actually change between 0.2 and 0.3. That'll probably save us a bunch of updating in the short term.

We'll also want to pin the version of `StringTemplate` used in the new entity cache directives for now, and work with that team on an upgrade strategy for the future.

apollo-federation/src/connectors/spec/mod.rs
```diff
@@ -113,6 +119,20 @@ impl PartialOrd for ConnectSpec {
...
+    pub fn latest() -> Self {
+        Self::V0_2
+    }
+
+    /// Returns the next version of the [`ConnectSpec`] to be released.
+    /// Test-only!
+    #[cfg(test)]
+    pub fn next() -> Self {
+        Self::V0_3
+    }
```
benjamn: @dylan-apollo Following your https://github.com/apollographql/router/pull/7926#issuecomment-3129913018, I've configured both `ConnectSpec::latest()` and `ConnectSpec::next()` as test-only (above).

They're useful in tests (see `apollo-federation/src/schema/validators/cache_tag.rs`), because non-test code can hard-code a `DEFAULT_CONNECT_SPEC` to (say) `ConnectSpec::V0_2`, but then also have a test to make sure it's still the same as `ConnectSpec::latest()` (which is fine to use because this is just a test):
```rust
    #[test]
    fn test_latest_connect_spec() {
        // This test exists to find out when ConnectSpec::latest() changes, so
        // we can decide whether to update DEFAULT_CONNECT_SPEC.
        assert_eq!(DEFAULT_CONNECT_SPEC, ConnectSpec::latest());
    }
```
This way, there are no sudden automatic updates, but `ConnectSpec` users also find out (via harmless test failures) that they may want to consider updating `DEFAULT_CONNECT_SPEC`.

benjamn: I should add: the hard-coding of `DEFAULT_CONNECT_SPEC` only makes sense for code that needs to use `StringTemplate` or `JSONSelection` in a context (like `cache_tag.rs`) where there is no connector subgraph schema with a `connect/v0.n` version, nor any other obvious source of `ConnectSpec`.

benjamn: Update: I was able to make both `ConnectSpec::latest` and `::next` private with `pub(crate)` again (https://github.com/apollographql/router/pull/7926/commits/a9994d4a85c37a8c1ce8d1beb60894c658189bfd), since they are not actually used in any tests outside the `apollo-federation` crate.

PR #7927, opened 2025-07-18
Provide `ConnectSpec` version gating ability to all major parts of JSONSelection mapping language

benjamn:
In order to provide backwards and forwards compatibility for developers still using either `connect/v0.1` or `connect/v0.2` when they upgrade their router/federation to versions that support `connect/v0.3` or later, we need an ability to condition certain `v0.3+`-only behavior on the current `ConnectSpec`.

The first commit in this PR smuggles the `ConnectSpec` into the input `Span` type used by all the parsing functions, which allows all the `nom` combinator code to continue to work mostly seamlessly, as compared with trying to add another parameter to all the parsing functions. This was enough machinery to allow `JSONSelection` parsing behavior to take the current `ConnectSpec` into account, which is especially important so we can reliably preserve (and keep testing) old parsing behavior, even after we ship newer behaviors.

Having parsed a given `JSONSelection` using a given `ConnectSpec`, it would be convenient to store the `ConnectSpec` as part of the `JSONSelection` structure, so any further operations (e.g. `apply_with_path` and/or `compute_output_shape`) could infer the `ConnectSpec` they should use from `self.spec()`. Inconveniently, `JSONSelection` was previously an enum, so I faced the typical conundrum about whether to add a new parameter to every enum variant, or convert the enum to a struct. Long story short: converting to a struct worked well (see commit https://github.com/apollographql/router/commit/03f4baa76cc70029bcfc5ac32c296eca0e991798), despite some disruption to consuming code elsewhere in the `apollo-federation` crate. Not many consumers were actually `match`ing against the `JSONSelection` type as an enum, nor creating new `JSONSelection::Path(...)`s directly (instead of calling `::parse`).

The following commit https://github.com/apollographql/router/commit/10be67893eec6cfe9741acaae41089b2eccd3466 updates various snapshots after the `JSONSelection` enum-to-struct conversion, as expected.

Commit https://github.com/apollographql/router/commit/a4e0a058036552d020b055cb2455e161d76e4489 introduces the notion of a single `&ShapeContext` parameter for `compute_output_shape`, encapsulating (currently) both `named_shapes` and `source_id` but soon a `spec: ConnectSpec` field as well.

Commit https://github.com/apollographql/router/commit/a460a973d3b668fb189f6e7fb9c74841725eae61 ensures the current `ConnectSpec` is available throughout both `compute_output_shape` and `apply_to_path`, based solely on the spec used when the `JSONSelection` was originally parsed.

Commit https://github.com/apollographql/router/commit/c730cec57e8834d0d1afbd8bc733ee9e44f99d7a fixes some places where we were not specifying the `ConnectSpec` when parsing `StringTemplate`s, so you can now fully control the `ConnectSpec` in effect across all your `JSONSelection` strings using just the `connect/v0.n` part of the `@link` that imports the `@source` and `@connect` directives.

I will build on this PR with a few other PRs introducing new parsing/shape/runtime functionality that needs to be version-gated using the system introduced by this PR.


apollo-federation/src/connectors/json_selection/lit_expr.rs
```diff
@@ -196,7 +196,7 @@ impl LitExpr {
                     },
                 ),
             )),
-        ))(input)?;
+        ))(input.clone())?;
```
naomijub, line 199: Doesnt this defeat the parser purpose?  Cloning the input stream would cause the parser to reparse stuff

benjamn: The `input` here is a `Span` or `LocatedSpan`, which holds a string slice reference to the original source, IIUC. This `Span` type used to be `Copy`, but this PR changed some things so that `Span` is only `Clone` now (specifically because the `extra: SpanExtra` struct has an owned `errors: Vec<String>` field in it now). That means we have to add some additional `.clone()`s, but the underlying source strings should not be getting copied or re-parsed.

apollo-federation/src/connectors/json_selection/location.rs
```diff
@@ -19,10 +20,33 @@ use super::ParseResult;
...
+#[derive(Debug, PartialEq, Eq, Clone, Hash, Default)]
+pub(crate) struct SpanExtra {
+    pub(super) spec: ConnectSpec,
+    pub(super) errors: Vec<String>,
+}
 
 pub(super) fn new_span(input: &str) -> Span {
-    Span::new_extra(input, None)
+    Span::new_extra(input, Default::default())
+}
```
benjamn, line 33: Here's why we need more `.clone()`s of input `Span` values used by the parser: we went from passing around a `LocatedSpan<&'a str, Option<&'static str>>` type that was automatically `Copy` because it was just tracking `&str` references pointing into the original source (or an optional static error message string), to passing around a `LocatedSpan<&'a str, SpanExtra>`, which is still just as cheap to clone with regard to the `&'a str` part, but now requires `Clone` instead of `Copy` because `SpanExtra` has owned data.

If you see a way to make `SpanExtra` implement `Copy` again (`ConnectSpec` implements `Copy`, for what it's worth), then we could maybe undo the new clones, but I'm confident they should be relatively cheap clones. I don't think I want to go back to a singular `error: Option<&'static str>`, though I imagine that's the best option for making everything `Copy` again.

apollo-federation/src/connectors/json_selection/apply_to.rs
```diff
@@ -79,32 +88,19 @@ impl JSONSelection {
...
-                source_id,
-            ),
-            Self::Path(path_selection) => path_selection.compute_output_shape(
-                input_shape.clone(),
-                input_shape,
-                named_var_shapes,
-                source_id,
-            ),
+    pub(crate) fn compute_output_shape(&self, context: &ShapeContext, input_shape: Shape) -> Shape {
+        match &self.inner {
```
andrewmcgivery, line 95: This code acting on `self.inner` instead of `self`... I don't think I have the context for this change. 

It _looks_ like ultimately it's doing the same thing but... Can you ELI5 the what/why behind this function change? 😅

apollo-federation/src/connectors/json_selection/helpers.rs
```diff
@@ -27,7 +27,7 @@ macro_rules! selection {
 // Consumes any amount of whitespace and/or comments starting with # until the
 // end of the line.
 pub(crate) fn spaces_or_comments(input: Span) -> ParseResult<WithRange<&str>> {
-    let mut suffix = input;
+    let mut suffix = input.clone();
```
andrewmcgivery, line 30: Just want to double check that this clone is required. :)

andrewmcgivery: Maybe just as a Sanity Check, and I think you may have already addressed this on Julia's comment, but it looks like theres a bunch of new clones. Any concerns there or is the impact negligible? I'm always told to avoid clone at all costs in Rust so just want to devil's advocate here :D

andrewmcgivery: You may have already addressed this in your comment here: https://github.com/apollographql/router/pull/7927/files#r2219888472

If so, let me know and I'll refer to that comment. :)

apollo-federation/src/connectors/json_selection/location.rs
```diff
@@ -19,10 +20,33 @@ use super::ParseResult;
...
+    Span::new_extra(
+        input,
+        SpanExtra {
+            spec,
+            ..Default::default()
+        },
+    )
+}
+
+// TODO Use this helper function to get the ConnectSpec from any input Span
```
andrewmcgivery: Is this a TODO for this PR or for future work?

If it's for future work, do we have a Jira item or should there be one?

And lastly... small nit... can we add the colon so this is `TODO:`? Tooling that works with TODO comments often looks for the `:` as part of it :)

apollo-federation/src/connectors/json_selection/methods/future/get.rs
```diff
@@ -1,21 +1,21 @@
 use std::iter::empty;
```
andrewmcgivery, line 1: I just want to flag that `get` has been almost completely rewritten in https://github.com/apollographql/router/pull/7884 so we may want to get that merged and then rebase this PR otherwise we're going to have quite the mess of PR conflicts 😅

apollo-federation/src/connectors/json_selection/apply_to.rs
benjamn: In order for `JSONSelection` to retain the `spec: ConnectSpec` used to parse it (which is helpful so we have access to it later, in `apply_to_path` and `compute_output_shape`), I needed to convert `JSONSelection` from an enum to a struct (see commit https://github.com/apollographql/router/pull/7927/commits/c6918d3e97e7d9f8b215b06ac993f4c3e9e3bb25).

The `inner: TopLevelSelection` field of the new `JSONSelection` struct now holds an enum with `::Named(SubSelection)` and `::Path(PathSelection)` variants like before.

It was between this and adding the `ConnectSpec` to every variant (repetitive), or storing a `spec: ConnectSpec` field in every `PathSelection` and `SubSelection` struct (redundant since they'd be equal throughout the AST).

So, to answer your question, `match &self.inner { ... }` is like `match self { ... }`, except the enum we're matching against is now `TopLevelSelection::` instead of `Self::` (since `JSONSelection` is no longer an enum).

andrewmcgivery: Makes sense, thank you!

apollo-federation/src/connectors/json_selection/location.rs
```diff
@@ -19,10 +20,42 @@ use super::ParseResult;
...
 // we would not be cloning long String messages very often (and the rest of the
 // Span fields are cheap to clone).
-pub(crate) type Span<'a> = LocatedSpan<&'a str, Option<&'static str>>;
+pub(crate) type Span<'a> = LocatedSpan<&'a str, SpanExtra>;
+
+#[derive(Debug, PartialEq, Eq, Clone, Hash)]
+pub(crate) struct SpanExtra {
+    pub(super) spec: ConnectSpec,
+    pub(super) errors: Vec<String>,
+}
```
benjamn: Here's why we need more `.clone()`s of input `Span` values used by the parser: we went from passing around a `LocatedSpan<&'a str, Option<&'static str>>` type that was automatically `Copy`, because it was just tracking `&str` references pointing into the original source (or to static error message strings), to passing around a `LocatedSpan<&'a str, SpanExtra>`, which is still just as cheap to clone with regard to the `&'a str` part, but now requires `Clone` instead of `Copy` because `SpanExtra` has owned data.

Since `ConnectSpec` is `Copy`, I suppose we might be able to go back to something like a singular `error: &'static str` so the whole `SpanExtra` struct could be `Copy` again.

apollo-federation/src/connectors/json_selection/location.rs
benjamn: Good catch, I should remove this (in this PR)

apollo-federation/src/connectors/json_selection/methods/future/get.rs
benjamn: That's fine, I've been regularly rebasing against those `->` method-related changes, and will continue.

apollo-federation/src/connectors/json_selection/helpers.rs
benjamn: I could change the `errors: Vec<String>` field to `errors: Vec<&'static str>` to avoid some `String` cloning, but `SpanExtra` is still going to be `Clone` rather than `Copy` because of the `Vec`.

Also worth noting: the new cloning is happening mostly during a time when there are no `errors` in the `Vec` yet (errors only appear when there's been a failure and we're reporting it back up the chain), so we're usually cloning an empty vector. And cloning the `spec: ConnectSpec` field is trivial (as `ConnectSpec` is an enum that implements `Copy`).

If you're not familiar with this part of the code, I know it must seem weird that we smuggle the `ConnectSpec` into our parsing code via `LocatedSpan`, along with the actual `&str` to be parsed, and then also smuggle `errors` out using the same structure, since the output `Span` (often called `remainder` in the code) can be reported as part of a `nom` parsing error. See #6089 for previous discussion of why this is the best way to write parsing functions that work with other `nom` combinators (which unfortunately do not play well with parsing methods that have additional parameters besides the `input: Span`).

apollo-federation/src/connectors/json_selection/location.rs
```diff
@@ -51,9 +51,6 @@ pub(super) fn new_span_with_spec(input: &str, spec: ConnectSpec) -> Span {
     )
 }
 
-// TODO Use this helper function to get the ConnectSpec from any input Span
-// during parsing, so parsing behavior can be gated by version.
-#[allow(dead_code)]
```
benjamn: Commenting so I can link to this code.

apollo-federation/src/connectors/json_selection/location.rs
benjamn: Removed as part of commit https://github.com/apollographql/router/pull/7927/commits/7b07c17792a099021e9a199a21aa421abe7fe4d7 (see https://github.com/apollographql/router/pull/7927#discussion_r2223562011).

naomijub (approved):
LGTM, I didn't get what the test issue with connectspec is, I'm open to chatting about this and trying to help with a solution

PR #7930, opened 2025-07-18
support for additional deployment types with ENV

morriswchris:
This PR introduces a new way of setting the `deployment.type` attribute to the `apollo.router.instance` metric to track deployments outside of the official Apollo helm chart.

---

**Checklist**

- Tests added and passing[^4]
    - [ ] Unit tests
    - [ ] Integration tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-1398]: https://apollographql.atlassian.net/browse/ROUTER-1398?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apollo-router/src/plugins/fleet_detector.rs
```diff
@@ -502,11 +511,23 @@ fn get_otel_os() -> &'static str {
...
-    if std::env::var_os(OFFICIAL_HELM_CHART_VAR).is_some() {
+    if official_helm_chart.is_some() {
         return "official_helm_chart";
     }
+
+    // Check for a custom deployment type via APOLLO_ROUTER_DEPLOYMENT_TYPE
+    if let Some(val) = deployment_type {
+        if !val.is_empty() {
+            // Leak the string to get a 'static lifetime
+            return Box::leak(val.to_string().into_boxed_str());
```
nmoutschen: Would that risk causing a memory leak, since we'd be calling this whenever there's a change to the Router?

morriswchris: Good callout. I ended up refactoring this to get the env as a string during plugin registration. This should make it so we don't need to logic each time we emit metrics.

PR #7932, opened 2025-07-18
Refactor `PathList::compute_output_shape` to ensure left-associativity, and use `ConnectSpec` to version-gate reenablement of `->` method shape processing

benjamn:
Builds on PR #7927. Similar changes (without version gating) were originally included in PR #7623, but have been extracted into this standalone PR. Now that we have a `ConnectSpec` version gating system for the JSONSelection mapping language, we can use it when adding new features!

This PR refactors `PathList::compute_output_shape` to ensure left-associativity, and also re-enables shape processing for `->` methods, which is the specific change we want to depend on the developer explicitly choosing `connect/v0.3`, as the improved shape checking behavior could be surprising for those still using `connect/v0.2`.

I also added some regression tests to prove that https://apollographql.atlassian.net/browse/CNN-722 can be resolved.

### DIFF
--- a/apollo-federation/src/connectors/json_selection/apply_to.rs
+++ b/apollo-federation/src/connectors/json_selection/apply_to.rs
@@ -179,7 +179,7 @@ pub(crate) struct ShapeContext {
 impl ShapeContext {
     pub(crate) fn new(source_id: SourceId) -> Self {
         Self {
-            spec: ConnectSpec::latest(),
+            spec: JSONSelection::default_connect_spec(),
             named_shapes: IndexMap::default(),
             source_id,
         }
--- a/apollo-federation/src/connectors/json_selection/location.rs
+++ b/apollo-federation/src/connectors/json_selection/location.rs
@@ -22,22 +22,29 @@ use crate::connectors::ConnectSpec;
 // Span fields are cheap to clone).
 pub(crate) type Span<'a> = LocatedSpan<&'a str, SpanExtra>;
 
-#[derive(Debug, PartialEq, Eq, Clone, Hash, Default)]
+#[derive(Debug, PartialEq, Eq, Clone, Hash)]
 pub(crate) struct SpanExtra {
     pub(super) spec: ConnectSpec,
     pub(super) errors: Vec<String>,
 }
 
-pub(super) fn new_span(input: &str) -> Span {
-    Span::new_extra(input, Default::default())
+#[cfg(test)]
+pub(crate) fn new_span(input: &str) -> Span {
+    Span::new_extra(
+        input,
+        SpanExtra {
+            spec: super::JSONSelection::default_connect_spec(),
+            errors: Vec::new(),
+        },
+    )
 }
 
-pub(super) fn new_span_with_spec(input: &str, spec: ConnectSpec) -> Span {
+pub(crate) fn new_span_with_spec(input: &str, spec: ConnectSpec) -> Span {
     Span::new_extra(
         input,
         SpanExtra {
             spec,
-            ..Default::default()
+            errors: Vec::new(),
         },
     )
 }
--- a/apollo-federation/src/connectors/json_selection/mod.rs
+++ b/apollo-federation/src/connectors/json_selection/mod.rs
@@ -3,7 +3,7 @@ pub(crate) mod helpers;
 mod immutable;
 mod known_var;
 mod lit_expr;
-mod location;
+pub(crate) mod location;
 mod methods;
 mod parser;
 mod pretty;
--- a/apollo-federation/src/connectors/json_selection/parser.rs
+++ b/apollo-federation/src/connectors/json_selection/parser.rs
@@ -35,7 +35,6 @@ use super::location::ranged_span;
 use crate::connectors::ConnectSpec;
 use crate::connectors::Namespace;
 use crate::connectors::json_selection::location::get_connect_spec;
-use crate::connectors::json_selection::location::new_span;
 use crate::connectors::variable::VariableNamespace;
 use crate::connectors::variable::VariableReference;
 
@@ -141,14 +140,14 @@ impl JSONSelection {
     pub fn named(sub: SubSelection) -> Self {
         Self {
             inner: TopLevelSelection::Named(sub),
-            spec: ConnectSpec::latest(),
+            spec: Self::default_connect_spec(),
         }
     }
 
     pub fn path(path: PathSelection) -> Self {
         Self {
             inner: TopLevelSelection::Path(path),
-            spec: ConnectSpec::latest(),
+            spec: Self::default_connect_spec(),
         }
     }
 
@@ -166,7 +165,7 @@ impl JSONSelection {
     pub fn empty() -> Self {
         Self {
             inner: TopLevelSelection::Named(SubSelection::default()),
-            spec: ConnectSpec::latest(),
+            spec: Self::default_connect_spec(),
         }
     }
 
@@ -183,7 +182,11 @@ impl JSONSelection {
     // as the input type and a custom JSONSelectionParseError type as the error
     // type, rather than using Span or nom::error::Error directly.
     pub fn parse(input: &str) -> Result<Self, JSONSelectionParseError> {
-        JSONSelection::parse_with_spec(input, ConnectSpec::latest())
+        JSONSelection::parse_with_spec(input, Self::default_connect_spec())
+    }
+
+    pub(super) fn default_connect_spec() -> ConnectSpec {
+        ConnectSpec::V0_2
     }
 
     pub fn parse_with_spec(
@@ -1166,7 +1169,12 @@ impl Display for Key {
 // Identifier ::= [a-zA-Z_] NO_SPACE [0-9a-zA-Z_]*
 
 pub(super) fn is_identifier(input: &str) -> bool {
-    all_consuming(parse_identifier_no_space)(new_span(input)).is_ok()
+    // TODO Don't use the whole parser for this?
+    all_consuming(parse_identifier_no_space)(new_span_with_spec(
+        input,
+        JSONSelection::default_connect_spec(),
+    ))
+    .is_ok()
 }
 
 fn parse_identifier(input: Span) -> ParseResult<WithRange<String>> {
@@ -1295,6 +1303,15 @@ mod tests {
     use crate::connectors::json_selection::helpers::span_is_all_spaces_or_comments;
     use crate::connectors::json_selection::location::new_span;
     use crate::selection;
+
+    #[test]
+    fn test_default_connect_spec() {
+        // We don't necessarily want to update what
+        // JSONSelection::default_connect_spec() returns just because
+        // ConnectSpec::latest() changes, but we want to know when it happens,
+        // so we can consider updating.
+        assert_eq!(JSONSelection::default_connect_spec(), ConnectSpec::latest());
+    }
 
     #[test]
     fn test_identifier() {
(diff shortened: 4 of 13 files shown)


apollo-federation/src/connectors/json_selection/helpers.rs
```diff
@@ -22,6 +22,15 @@ macro_rules! selection {
             panic!("invalid selection: {:?}", $input);
         }
     };
+    ($input:expr, $spec:expr) => {
+        if let Ok(parsed) =
+            $crate::connectors::json_selection::JSONSelection::parse_with_spec($input, $spec)
```
benjamn, line 27: When using the `selection!` macro in tests, you can now specify a `ConnectSpec` as the second argument, e.g. `selection!("upc ...type->match(...)", ConnectSpec::V0_3)`, and everything else that needs to know the `ConnectSpec` can figure it out from the original `JSONSelection::parse_with_spec` call.

apollo-federation/src/connectors/json_selection/apply_to.rs
```diff
@@ -2626,10 +2701,12 @@ mod tests {
...
 
-    #[test]
-    fn test_left_associative_path_evaluation() {
+    #[rstest]
+    #[case::latest(ConnectSpec::V0_2)]
+    #[case::next(ConnectSpec::V0_3)]
+    fn test_left_associative_path_evaluation(#[case] spec: ConnectSpec) {
         assert_eq!(
-            selection!("batch.id->first").apply_to(&json!({
+            selection!("batch.id->first", spec).apply_to(&json!({
```
benjamn, line 2699: The general routine with testing different `ConnectSpec` versions is to use `#[rstest]` and pass the `spec` to the `selection!(..., spec)` macro.

Of course, not all tests will work the same way in both v0.2 and v0.3, so some test functions will be testing only one specific version.

apollo-federation/src/connectors/json_selection/apply_to.rs
```diff
@@ -346,6 +354,8 @@ impl ApplyToInternal for JSONSelection {
         input_shape: Shape,
         dollar_shape: Shape,
     ) -> Shape {
+        debug_assert_eq!(context.spec(), self.spec());
```
andrewmcgivery, line 357: Is this meant to be here?

benjamn: Yes, this was the change I ended up making after our discussion in https://github.com/apollographql/router/pull/7933#discussion_r2219478591, as the assertion seemed more appropriate for this PR.

benjamn:
Following up on @dylan-apollo's 4th suggestion from https://github.com/apollographql/router/pull/7623#pullrequestreview-2911465349: ~https://github.com/apollographql/router/pull/7932/commits/fbfb9dd12afc44f3151d6ec7d8fb6c84a071622a~ https://github.com/apollographql/router/pull/7932/commits/2670e4c267fc2dd5ac3d31f010bc0f54b254a062

apollo-federation/src/connectors/json_selection/apply_to.rs
benjamn: Recent learning: having a single `test_` function with multiple `ConnectSpec` cases does not play well with snapshot tests, since the generated filenames of the snapshots don't seem to take the cases into account, so the snapshot file writes can race against each other, leading to flaky test failures.

This particular test runs fine with either `ConnectSpec` version, in part because it doesn't have any `debug_assert_snapshot!` calls. In other cases (where snapshots conflicted), I've ended up splitting tests functions into two, one for `ConnectSpec::V0_2` and another for `ConnectSpec::V0_3`, which has often allowed simplifying the logic within the individual tests, compared to trying to write one combined test.

apollo-federation/src/connectors/json_selection/apply_to.rs
dylan-apollo: I guess there'd be lower risk of these falling out of sync if we remove `ConnectSpec::latest` (mentioned in the base draft PR)

apollo-federation/src/connectors/json_selection/apply_to.rs
```diff
@@ -708,7 +718,50 @@ impl ApplyToInternal for WithRange<PathList> {
...
+                    }),
+                    input_shape.locations.iter().cloned(),
+                );
+            }
+            ShapeCase::Error(error) => {
+                return match error.partial.as_ref() {
+                    Some(partial) => Shape::error_with_partial(
+                        error.message.clone(),
+                        self.compute_output_shape(context, partial.clone(), dollar_shape),
+                        input_shape.locations.clone(),
```
dylan-apollo:
```suggestion
                        input_shape.locations.iter().cloned(),
```

I think this form (used above as well) is more efficient, since the `Shape` constructor is going to call `into_iter()` and recollect again. So this way we reallocate the `Vec` once instead of twice.

We probably want to adjust the underlying `Shape` at some point to let us more easily move the `Vec` instead of constantly cloning it.

apollo-federation/src/connectors/json_selection/apply_to.rs
```diff
@@ -723,83 +776,106 @@ impl ApplyToInternal for WithRange<PathList> {
...
             ),
 
-            PathList::Method(method_name, _method_args, _tail) => ArrowMethod::lookup(method_name)
-                .map_or_else(
-                    || {
+            PathList::Method(method_name, method_args, tail) => {
+                if let Some(method) = ArrowMethod::lookup(method_name) {
+                    // Before connect/v0.3, we did not consult method.shape at
+                    // all, and instead returned Unknown. Since this behavior
+                    // has consequences for URI validation, the older behavior
```
dylan-apollo, line 823: Has consequences for everything except `selection` right now I think, so `errors`, `body`, `headers`, etc.

apollo-federation/src/connectors/json_selection/apply_to.rs
```diff
@@ -723,83 +776,106 @@ impl ApplyToInternal for WithRange<PathList> {
...
-                selection.compute_output_shape(context, input_shape, dollar_shape)
+                if input_shape.is_none() {
+                    return Shape::error(
+                        format!(
+                            "Selection cannot be applied to {}",
+                            input_shape.pretty_print()
+                        ),
+                        self.shape_location(context.source_id()),
+                    );
+                }
```
dylan-apollo: Isn't this unreachable because of the very first check now?

apollo-federation/src/connectors/json_selection/apply_to.rs
```diff
@@ -723,83 +776,106 @@ impl ApplyToInternal for WithRange<PathList> {
...
+            PathList::Empty => (input_shape, None),
+        };
+
+        if let (false, Some(tail)) = (current_shape.is_none(), tail) {
+            tail.compute_output_shape(context, current_shape, dollar_shape)
+        } else {
+            // If there is no tail or current_shape was None, path evaluation
+            // returns current_shape immediately.
+            current_shape
         }
```
dylan-apollo: The new short-circuit check at the top means that we don't have to double-check `none` here, right? So a simpler check would be equivalent?


```suggestion
        if let Some(tail) = tail {
            tail.compute_output_shape(context, current_shape, dollar_shape)
        } else {
            current_shape
        }
```

apollo-federation/src/connectors/json_selection/apply_to.rs
```diff
@@ -2828,17 +2908,221 @@ mod tests {
...
+                    ],
+                })
+            ),
             (Some(json!("string")), vec![]),
         );
     }
 
+    #[rstest]
+    #[case::v0_2(ConnectSpec::V0_2)]
+    fn test_left_associative_output_shapes_v0_2(#[case] spec: ConnectSpec) {
```
dylan-apollo: There are downsides to `rstest`: it makes it harder to run an individual test (no click the button in RustRover) and there's a tiny cost to the macros... which adds up. So when there's a single case, we should skip `rstest` (I gather this is an artifact of a previous approach, from your above comment about snapshots).


```suggestion
    fn test_left_associative_output_shapes_v0_2() {
        let spec = ConnectSpec::V0_2;
```

apollo-federation/src/connectors/json_selection/apply_to.rs
```diff
@@ -2828,17 +2908,221 @@ mod tests {
...
+            selection!("$batch.id->map([@])->echo([@])", spec)
+                .compute_output_shape(&shape_context, root_shape,)
+                .pretty_print(),
+            "Unknown",
+        );
+    }
+
+    #[rstest]
+    #[case::v0_3(ConnectSpec::V0_3)]
+    fn test_left_associative_output_shapes_v0_3(#[case] spec: ConnectSpec) {
```
dylan-apollo:
```suggestion
    fn test_left_associative_output_shapes_v0_3() {
        let spec = ConnectSpec::V0_3;
```

dylan-apollo (approved):
Looks good to me. I left a few minor comments, but the core implementation looks right!

apollo-federation/src/connectors/json_selection/apply_to.rs
benjamn: This will change (get better) with https://github.com/apollographql/shape-rs/pull/29, after which you'd do something like `input_shape.locations()` for a `Iterator<Item = &Location>` or `input_shape.locations().cloned()` for `Iterator<Item = Location>`. Both locations and names are stored behind the new `Shape::meta` field.

apollo-federation/src/connectors/json_selection/apply_to.rs
benjamn: Yeah! Let's lean into that base case.

apollo-federation/src/connectors/json_selection/apply_to.rs
benjamn: https://github.com/apollographql/router/pull/7932/commits/3705ee2c9a2eea3b7b7d0e08a3c8b870f28acbae

apollo-federation/src/connectors/json_selection/apply_to.rs
benjamn: https://github.com/apollographql/router/pull/7932/commits/3705ee2c9a2eea3b7b7d0e08a3c8b870f28acbae

apollo-federation/src/connectors/json_selection/apply_to.rs
benjamn: https://github.com/apollographql/router/pull/7932/commits/3705ee2c9a2eea3b7b7d0e08a3c8b870f28acbae

apollo-federation/src/connectors/json_selection/apply_to.rs
benjamn: https://github.com/apollographql/router/pull/7932/commits/3705ee2c9a2eea3b7b7d0e08a3c8b870f28acbae

apollo-federation/src/connectors/json_selection/apply_to.rs
benjamn: https://github.com/apollographql/router/pull/7932/commits/3705ee2c9a2eea3b7b7d0e08a3c8b870f28acbae

apollo-federation/src/connectors/json_selection/apply_to.rs
benjamn: https://github.com/apollographql/router/pull/7932/commits/3705ee2c9a2eea3b7b7d0e08a3c8b870f28acbae

PR #7933, opened 2025-07-18
Fully implement `math_shape` again

benjamn:
This PR builds on #7932 and supersedes #7623, including only the improvements to `math_shape` and not the other preliminary commits, which have now been moved into #7932.

The math methods (`->add`, `->sub`, `->mul`, `->div`, and `->mod`) still need to be promoted to `methods::public` rather than `methods::future`, before we can call them fully shipped. However, all `->` methods are available in tests, which is why the tests are passing before we've publicly released the new methods.

### DIFF
--- a/apollo-federation/src/connectors/json_selection/methods/future/arithmetic.rs
+++ b/apollo-federation/src/connectors/json_selection/methods/future/arithmetic.rs
@@ -1,7 +1,6 @@
 use serde_json::Number;
 use serde_json_bytes::Value as JSON;
 use shape::Shape;
-use shape::ShapeCase;
 
 use crate::connectors::json_selection::ApplyToError;
 use crate::connectors::json_selection::ApplyToInternal;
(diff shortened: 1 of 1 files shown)


apollo-federation/src/connectors/json_selection/apply_to.rs
```diff
@@ -92,6 +92,9 @@ impl JSONSelection {
     }
 
     pub(crate) fn compute_output_shape(&self, context: &ShapeContext, input_shape: Shape) -> Shape {
+        let context_clone = context.clone().with_spec(self.spec());
```
andrewmcgivery: Are the changes to this file meant to be in this PR if this PR is only for the math shape?

apollo-federation/src/connectors/json_selection/methods/future/arithmetic.rs
```diff
@@ -141,18 +141,108 @@ infix_math_op!(mul_op, *);
...
+            // static integer math when all the inputs are statically known, and
+            // return a specific Shape::int_value when that math succeeds. That
+            // might require using different shape functions for each of the
+            // math operations, rather than a single math_shape function.
+            Shape::int(method_name.shape_location(context.source_id()))
+        }
+        CheckNumericResult::FloatPossible => {
+            Shape::float(method_name.shape_location(context.source_id()))
+        }
+        CheckNumericResult::Neither => unreachable!("handled above"),
```
andrewmcgivery: To eliminate this match arm, could we instead do a `if matches` instead of the match statement?

Basically say if it's an Int for sure, return int, else return float?

apollo-federation/src/connectors/json_selection/methods/future/arithmetic.rs
```diff
@@ -357,4 +586,213 @@ mod tests {
             (Some(json!(0)), vec![]),
         );
     }
+
+    #[rstest]
+    #[case::v0_3(ConnectSpec::V0_3)]
```
andrewmcgivery, line 571: Is this new? Are we going to have to go back and update all tests that use `selection!` with a spec version?

apollo-federation/src/connectors/json_selection/methods/future/arithmetic.rs
```diff
@@ -178,10 +268,149 @@ infix_math_method!(ModMethod, mod_method, rem_op);
...
+    use shape::location::SourceId;
 
+    use super::CheckNumericResult;
+    use super::check_numeric_shape;
+    use crate::connectors::ConnectSpec;
+    use crate::connectors::json_selection::ShapeContext;
     use crate::selection;
 
+    #[test]
+    fn test_check_numeric_shape() {
```
andrewmcgivery, line 278: On other methods, I've been putting the method_tests and shape_tests into two separate mods in the same file to keep them cleanly separated (it's nice cause you can run your method tests and shape tests separately with one click).

I'm also fine with taking on doing some cleanup of the the tests in this file as part of the promotion work if that is preferred. :)

apollo-federation/src/connectors/json_selection/apply_to.rs
naomijub: I imagine that this cloning step is due to with_spec taking context as owned and not reference, right? 
If context something cheap to clone? If not would be better to have an alternative that takes it as reference, `with_borrowed_spec`?  Then we can avoid cloning or only clone what is cheap to clone

benjamn: I think this change probably belongs in a previous PR, but I noticed it while working on this PR and never moved it back. Thanks for noticing @andrewmcgivery (definitely wanted this PR to be just `math_shape` stuff), and @naomijub's comment has me thinking we can maybe just assert that `context.spec()` is the same as `self.spec()`, without any cloning.

apollo-federation/src/connectors/json_selection/methods/future/arithmetic.rs
benjamn: The general idea that some of our tests need to target a specific `ConnectSpec` version is new, though we have other tests that use `#[rstest]` in a similar way.

We do not necessarily need to go back and update all old tests to test a specific case, as they will (without changes) test using `ConnectSpec::latest()` by default. However, when we release the next update, we may want to keep testing the older latest version (likely v0.2), rather than switching over to `ConnectSpec::next()` (likely v0.3) and no longer testing the previous version.

We should discuss a better general story around testing ranges of versions, including handling version upgrades over time. The current system is enough for what I need now (e.g. testing specific older versions like v0.2 to ensure version gating works), and I hope it can evolve to become more ergonomic in the future.

apollo-federation/src/connectors/json_selection/methods/future/arithmetic.rs
benjamn: Feel free to tackle any cleanup during the promotion work!

apollo-federation/src/connectors/json_selection/methods/future/arithmetic.rs
benjamn: Yes, I think we can simplify things a bit, but it's still important to know if one of the inputs is definitely not numeric, because then `math_shape` needs to return a `Shape::error(...)` rather than `Shape::float()`.

benjamn: Ok, here's what I came up with: https://github.com/apollographql/router/pull/7933/commits/bfbd8544a5cb5a11c7eae689912a768d51c865d3

andrewmcgivery (approved):

PR #7934, opened 2025-07-18
Unify `NamedSelection` enum variants into a single struct, simplifying the EBNF grammar and parser implementation

benjamn:
Builds on PR #7933 as well as transitive dependencies including PRs #7932 and #7927, all targeting the `benjamn/JSONSelection-abstract-types` branch from PR #7926.

This PR implements a significant simplification to the JSONSelection formal EBNF grammar and parser implementation, based around the insight that `NamedFieldSelection` was essentially a single-`Key` `NamedPathSelection`, so it makes sense to represent them in a common way, rather than representing/handling them separately. Internally, the `NamedSelection` enum has become a struct that always has a `path: PathSelection` field, so we can reuse all the usual `PathSelection` and `PathList` machinery for every kind of `NamedSelection` element. This internal refactoring was possible without changing the grammar, since the new `NamedSelection` struct represents the previous grammar's AST structure in the same way it represents the new grammar's simplified structure.

Technical note: besides `path`, the other field of the new `NamedSelection` struct is `prefix: NamingPrefix`, where `NamingPrefix` is a private enum type that can represent the no-prefix case (`NamingPrefix::None`), the `Alias` case (`NamingPrefix::Alias(Alias)`), as well as a `NamingPrefix::Spread(Option<Range<usize>>)` variant, which is currently used to represent the former `PathWithSubSelection ::= Path SubSelection` case as `NamingPrefix::Spread(None)`, since existing behavior dictates the `SubSelection` gets inlined/spread into the larger selection, even without the `...` syntax that we're planning to introduce soon. Once we do introduce the explicit `...` spread operator, we can represent that with `NamingPrefix::Spread(Some(start..end))` where the range identifies the location of the `...` token in the source, when present. Of course we don't want there to be a functional difference between `NamingPrefix::Spread(None)` and `::Spread(Some(_))` (both get spread), but tracking the optional range of the `...` token seemed useful for tooling.

Building on the internal AST simplifications, it's now possible to simplify the grammar itself, down to a single
```ebnf
NamedSelection ::= (Alias | "...")? PathSelection | Path SubSelection
```
which finally unlocks the following change:
```diff
-KeyPath ::= Key PathStep+
+KeyPath ::= Key PathStep*
 PathStep ::= "." Key | "->" Identifier MethodArgs?
```
Previously, because we represented single-`Key` `KeyPath` paths differently from `NamedFieldSelection ::= Alias? Key SubSelection?` (both in the grammar and in the AST structure, with `NamedSelection::Field` and `NamedSelection::Path`), it was important for the grammar to decide whether some selection code was a `KeyPath` with just one `Key` or a `NamedSelection::Field` (which always has just one `Key`), and the presence of at least one `PathStep` in `KeyPath ::= Key PathStep+` seemed like a reasonable tie-breaking mechanism. However, this led to situations where you'd have to provide a `$.` prefix for single keys even when there was no field/path parsing ambiguity, as in `a->add($.b, $.c)`. While a developer could potentially learn why that was "necessary," I imagine almost anyone would prefer to write `a->add(b, c)`, taking advantage of the fact that `b` and `c` are being passed as arguments to the `->add` method, which is a context where they can't be mistaken for field selections, so they don't need the `$.` disambiguation. That's why `KeyPath ::= Key PathStep*` is so exciting.

In the near future, this unification paves the way for an explicit `...` spread operator (already included in the EBNF grammar to verify alignment) as well as a version of `?` for optional chaining that works even for field selections (without forcing them to become `PathSelection`s), as in `author { name age? }`. Since every `NamedSelection` ultimately has a `path: PathSelection` field now, the `?` can be handled the same way for every `PathSelection`, so field selections like `age?` will just work, with no separate grammar rules for `NamedFieldSelection`.


apollo-federation/src/connectors/json_selection/README.md
```diff
@@ -81,18 +81,14 @@ worry if it doesn't seem helpful yet, as every rule will be explained in detail
...
-Path                 ::= VarPath | KeyPath | AtPath | ExprPath
 PathSelection        ::= Path SubSelection?
-PathWithSubSelection ::= Path SubSelection
+Path                 ::= VarPath | KeyPath | AtPath | ExprPath
 VarPath              ::= "$" (NO_SPACE Identifier)? PathStep*
-KeyPath              ::= Key PathStep+
+KeyPath              ::= Key PathStep*
 AtPath               ::= "@" PathStep*
 ExprPath             ::= "$(" LitExpr ")" PathStep*
 PathStep             ::= "." Key | "->" Identifier MethodArgs?
```
benjamn, line 94: As usual with changes to the formal grammar, this `README.md` file is a good place to start your review.

benjamn:
Converting this PR to draft because I think I can/should split it into at least two separate PRs.

benjamn:
Closing in favor of #7945.

PR #7935, opened 2025-07-18
Add connectors `->parseInt` method

andrewmcgivery:
---

**Checklist**

- Tests added and passing[^4]
    - [x] Unit tests
    - [ ] Integration tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-federation/src/connectors/json_selection/methods/public/parse_int.rs
+++ b/apollo-federation/src/connectors/json_selection/methods/public/parse_int.rs
@@ -91,63 +91,60 @@ fn parse_int_method(
     }
 
     // Parse base argument or use default (10)
-    let base = match method_args.and_then(|args| args.args.first()) {
-        Some(first_arg) => {
-            let (base_opt, arg_errors) = first_arg.apply_to_path(data, vars, input_path);
+    let base = match method_args
+        .and_then(|args| args.args.first())
+        .map(|first_arg| first_arg.apply_to_path(data, vars, input_path))
+    {
+        Some((Some(JSON::Number(base_num)), _)) => {
+            let Some(base_value) = base_num.as_u64() else {
+                return (
+                    None,
+                    vec![ApplyToError::new(
+                        format!(
+                            "Method ->{} base argument must be an integer. Found: {}",
+                            method_name.as_ref(),
+                            base_num
+                        ),
+                        input_path.to_vec(),
+                        method_name.range(),
+                    )],
+                );
+            };
 
-            match base_opt {
-                Some(JSON::Number(base_num)) => {
-                    let Some(base_value) = base_num.as_u64() else {
-                        return (
-                            None,
-                            vec![ApplyToError::new(
-                                format!(
-                                    "Method ->{} base argument must be an integer. Found: {}",
-                                    method_name.as_ref(),
-                                    base_num
-                                ),
-                                input_path.to_vec(),
-                                method_name.range(),
-                            )],
-                        );
-                    };
-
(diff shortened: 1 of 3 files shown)


apollo-federation/src/connectors/json_selection/methods/public/parse_int.rs
```diff
@@ -0,0 +1,773 @@
...
+use crate::connectors::json_selection::ApplyToError;
+use crate::connectors::json_selection::ApplyToInternal;
+use crate::connectors::json_selection::MethodArgs;
+use crate::connectors::json_selection::VarsWithPathsMap;
+use crate::connectors::json_selection::immutable::InputPath;
+use crate::connectors::json_selection::location::Ranged;
+use crate::connectors::json_selection::location::WithRange;
+use crate::impl_arrow_method;
+
+const DEFAULT_BASE: u32 = 10;
```
naomijub, line 15: why not `u8`? Specially considering we only accept up to 36

apollo-federation/src/connectors/json_selection/methods/public/parse_int.rs
```diff
@@ -0,0 +1,773 @@
...
+/// $(42)->parseInt         results in error (input must be string)
+fn parse_int_method(
+    method_name: &WithRange<String>,
+    method_args: Option<&MethodArgs>,
+    data: &JSON,
+    vars: &VarsWithPathsMap,
+    input_path: &InputPath<JSON>,
+) -> (Option<JSON>, Vec<ApplyToError>) {
+    // Input must be a string
+    let JSON::String(input_str) = data else {
```
naomijub: I would expect that if I call `parseInt` in a number, I get the truncated integer of that number, so:
```
42 -> 42
123.6 -> 123
```

apollo-federation/src/connectors/json_selection/methods/public/parse_int.rs
```diff
@@ -0,0 +1,773 @@
...
+                    ),
+                    input_path.to_vec(),
+                    method_name.range(),
+                )],
+            );
+        }
+    }
+
+    // Parse base argument or use default (10)
+    let base = match method_args.and_then(|args| args.args.first()) {
```
naomijub: Nested matches of single use option are really hard to read, I know you and Dylan prefer it over map_or, but this is definitively a great use case.
`method_args.and_then(|args| args.args.first()).map_or(DEFAULT_BASE,|base| match base { })`

You could also `unwrap_or_default` at the end, but I think the first case is better

apollo-federation/src/connectors/json_selection/methods/public/parse_int.rs
andrewmcgivery: The from_str_radix function requires a u32 🤷  (Weird that they require that number range but require a u32)

apollo-federation/src/connectors/json_selection/methods/public/parse_int.rs
andrewmcgivery: Done!

apollo-federation/src/connectors/json_selection/methods/public/parse_int.rs
andrewmcgivery: Good call, I like this change!

andrewmcgivery: I'm not able to use `map_or` here because we have the early returns for errors, however, I was able to remove the nested matching using a `map`. :)

naomijub (approved):

PR #7936, opened 2025-07-18
implement prepare_supergraph and 

clenfest:


apollo-federation/src/merger/merger.rs
```diff
@@ -269,8 +274,30 @@ impl Merger {
...
+        // Add the @link specification to the merged schema
+        self.link_spec_definition
+            .add_to_schema(&mut self.merged, None)?;
+
+        // Apply the @join specification to the schema
+        self.link_spec_definition.apply_feature_to_schema(
+            &mut self.merged,
+            self.join_spec_definition,
+            None,
+            None, // purpose
```
duckki: Don't we need a purpose here? I see `for: EXECUTION` in supergraph schemas.

apollo-federation/src/merger/merge_links.rs
```diff
@@ -35,7 +35,7 @@ struct CoreDirectiveInSupergraph {
 
 impl Merger {
     #[allow(dead_code)]
-    fn collect_core_directives_to_compose(
+    pub(crate) fn collect_core_directives_to_compose(
```
duckki, line 37: We can remove `#[allow(dead_code)]` here and others.

apollo-federation/src/merger/merger.rs
duckki: I think we can use `self.join_spec_definition.purpose()` here.

duckki (approved):

PR #7937, opened 2025-07-19
test(federation): updated Display for GraphPath to show multiple conditions

duckki:
In order to match JS implementation more closely.
A follow-up of PR #7929.

---

**Checklist**

- Tests added and passing[^4]
    - [ ] Unit Tests
    - [ ] Integration Tests
    - [ ] Manual Tests

**Exceptions**

This PR is only for debug messages. No behavioral changes.


apollo-federation/src/query_graph/graph_path.rs
```diff
@@ -2338,19 +2338,12 @@ where
...
-                        if let Some(conditions) = &edge.conditions {
-                            write!(f, " --[{conditions} ⊢ {label}]--> {node}")
-                        } else if let Some(conditions) = &edge.override_condition {
-                            write!(f, " --[{conditions} ⊢ {label}]--> {node}")
-                        } else if !matches!(
-                            edge.transition,
-                            QueryGraphEdgeTransition::SubgraphEnteringTransition
-                        ) {
-                            write!(f, " --[{label}]--> {node}")
+                        let label = edge.to_string();
```
TylerBloom, line 2341: I want to make sure I understand this. The label now contains the logic for all but the final `else` statement of the old if-else chain, ya?

TylerBloom (approved):

apollo-federation/src/query_graph/graph_path.rs
duckki: I think the old code here had a bug skipping one case without printing. The `edge.to_string()` is more faithful with JS implementation.

The final else-branch was buggy, it was supposed to print --> {node} , but it skipped whole thing incorrectly. The PR fixes that, too.

PR #7939, opened 2025-07-20
Enforcing allowed features - config

DMallare:
---
- This is a poc for updates to license enforcement, specifically enforcing allowed features in config
- Enforcing allowed features as plugins is addressed in #7917 
- This is based on [the content in this doc](https://docs.google.com/document/d/13wh3NBmBnSPKBCH1f19eir-caWbz6Y9wKEHItXq6l1c/edit?tab=t.adap414o34k0#heading=h.1kx2ba8jc3vt)
- Tests have not been added yet but will be once we agree on the general direction this poc [and the doc](https://docs.google.com/document/d/13wh3NBmBnSPKBCH1f19eir-caWbz6Y9wKEHItXq6l1c/edit?tab=t.adap414o34k0#heading=h.1kx2ba8jc3vt) proposes
**Checklist**

- Tests added and passing[^4]
    - [ ] Unit tests
    - [ ] Integration tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/uplink/license_enforcement.rs
+++ b/apollo-router/src/uplink/license_enforcement.rs
@@ -284,7 +284,7 @@ impl LicenseEnforcementReport {
     }
 
     fn configuration_restrictions(license: &LicenseState) -> Vec<ConfigurationRestriction> {
-        let mut restrictions = vec![
+        let mut configuration_restrictions = vec![
             ConfigurationRestriction::builder()
                 .path("$.plugins.['experimental.restricted'].enabled")
                 .value(true)
@@ -347,10 +347,12 @@ impl LicenseEnforcementReport {
                 .build(),
         ];
 
-        // Check if the following features are in the licenses' allowed features
+        // If the license has an allowed_features claim, we know we're using a pricing
+        // plan with a subset of allowed features
+        // Check if the following features are in the licenses' allowed_features claim
         if let Some(allowed_features) = license.get_allowed_features() {
             if !allowed_features.contains(&AllowedFeature::APQ) {
-                restrictions.push(
+                configuration_restrictions.push(
                     ConfigurationRestriction::builder()
                         .path("$.apq.router.cache.redis")
                         .name("APQ caching")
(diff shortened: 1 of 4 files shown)


apollo-router/src/uplink/license_enforcement.rs
```diff
@@ -365,28 +343,141 @@ impl LicenseEnforcementReport {
...
+                configuration_restrictions.push(
+                    ConfigurationRestriction::builder()
+                        .path("$.subscription.enabled")
+                        .value(true)
+                        .name("Federated subscriptions")
+                        .build(),
+                );
+            }
+            // If the license has no allowed_features claim, we're using a pricing plan
+            // that should have the plugin enabled regardless
```
lrlna, line 432: A more philosophical question - if the pricing plan should have all plugins enabled, shouldn't their allowed features just have all the features? It's a bit of strange state as it reads for me right now:
* None - all features are allowed and all plugins should be enabled
* Some(vec![]) - (empty vec) no features are allowed (?)
* Some(vec![AllowedFeature::PersistedQueries]) - only persisted queries are allowed

lrlna (commented):
I think we need to figure out the data structure / state for `allowed_features`. It's state change is not super clear. I left a comment on it in the license_enforcement.rs file. That's probably the only thing to nail down across these two PRs. And again, this will need a bunch of tests, especially integration tests. Otherwise, the direction seems good to me.

apollo-router/src/uplink/license_enforcement.rs
DMallare: Yes, your understanding is correct and I see why it looks strange. The reasoning behind what we have currently has to do with the fact that we are only adding the `allowed_features` claim to developer/standard licenses for now. In the future, I see us adding this to all types of licenses and then an Enterprise license, for example, would have all allowed features in their claim or maybe even something like `ENABLE_ALL` - we can discuss once a design for that comes up. 

With the current work that is being done without adding the `allowed_features` claim to Enterprise licenses, having `None` mean no features enabled would disable everything for Enterprise licenses.

PR #7945, opened 2025-07-21
Unify `NamedSelection` enum variants into a single struct, without any grammar changes (yet)

benjamn:
After attempting to write the description for my PR #7934, I thought it might be helpful to break those changes up into separate PRs, as mentioned in https://github.com/apollographql/router/pull/7934#issuecomment-3097080630.

This PR contains just the internal unification of the `NamedSelection` enum into a struct of the same name. All the existing parsing logic has been updated to generate an AST in the new format, so no changes to parsing behavior are necessary, and no version gating is needed.

Of course, this change of representation triggered a whole bunch of snapshot updates, which I've kept in their own commit for ease of reviewing.

The first part of the #7934 description is still accurate/appropriate for this PR:

> Builds on PR #7933 as well as transitive dependencies including PRs #7932 and #7927, all targeting the `benjamn/JSONSelection-abstract-types` branch from PR #7926.
>
> This PR implements a significant simplification to the JSONSelection ~formal EBNF grammar and~ parser implementation, based around the insight that `NamedFieldSelection` was essentially a single-`Key` `NamedPathSelection`, so it makes sense to represent them in a common way, rather than representing/handling them separately. Internally, the `NamedSelection` enum has become a struct that always has a `path: PathSelection` field, so we can reuse all the usual `PathSelection` and `PathList` machinery for every kind of `NamedSelection` element. This internal refactoring was possible without changing the grammar, since the new `NamedSelection` struct represents the previous grammar's AST structure in the same way it represents the new grammar's simplified structure.
>
> Technical note: besides `path`, the other field of the new `NamedSelection` struct is `prefix: NamingPrefix`, where `NamingPrefix` is a private enum type that can represent the no-prefix case (`NamingPrefix::None`), the `Alias` case (`NamingPrefix::Alias(Alias)`), as well as a `NamingPrefix::Spread(Option<Range<usize>>)` variant, which is currently used to represent the former `PathWithSubSelection ::= Path SubSelection` case as `NamingPrefix::Spread(None)`, since existing behavior dictates the `SubSelection` gets inlined/spread into the larger selection, even without the `...` syntax that we're planning to introduce soon. Once we do introduce the explicit `...` spread operator, we can represent that with `NamingPrefix::Spread(Some(start..end))` where the range identifies the location of the `...` token in the source, when present. Of course we don't want there to be a functional difference between `NamingPrefix::Spread(None)` and `::Spread(Some(_))` (both get spread), but tracking the optional range of the `...` token seemed useful for tooling.

### DIFF
--- a/apollo-federation/src/connectors/expand/tests/snapshots/connectors@batch.graphql.snap
+++ b/apollo-federation/src/connectors/expand/tests/snapshots/connectors@batch.graphql.snap
@@ -54,18 +54,32 @@ input_file: apollo-federation/src/connectors/expand/tests/schemas/expand/batch.g
             inner: Named(
                 SubSelection {
                     selections: [
-                        Field(
-                            None,
-                            WithRange {
-                                node: Field(
-                                    "id",
-                                ),
-                                range: Some(
-                                    0..2,
-                                ),
+                        NamedSelection {
+                            prefix: None,
+                            path: PathSelection {
+                                path: WithRange {
+                                    node: Key(
+                                        WithRange {
+                                            node: Field(
+                                                "id",
+                                            ),
+                                            range: Some(
+                                                0..2,
+                                            ),
+                                        },
+                                        WithRange {
+                                            node: Empty,
+                                            range: Some(
+                                                2..2,
+                                            ),
+                                        },
+                                    ),
+                                    range: Some(
+                                        0..2,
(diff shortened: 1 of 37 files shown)


benjamn:
Closing in favor of #7957, which includes all commits from this PR but targets `dev` directly.

PR #7946, opened 2025-07-21
Refactor of extension elements checks

clenfest:
Move has_non_extension_elements and has_extension_elements to be on a trait since we will need them in places other than schema upgrading.


duckki (commented):
The trait applies to not just types, but also other schema elements.
See `trait SchemaElement` ([github](https://github.com/apollographql/router/blob/c49f8ccab9f6439ec5a08cb0fae3f8368c16a918/apollo-federation/src/schema/mod.rs#L1340)).
Why don't we unify the two traits into one under the `schema` module?

duckki (approved):
Thanks!

PR #7952, opened 2025-07-22
Support automatic subgraph mocking in mock subgraphs plugin

tninesling:
# Generated subgraph responses

While investigating Router issues, I often have to set up subgraphs with some mocked data so I can put the Router in front of them to see how it behaves. Usually, the actual data doesn't matter, and I just want something which matches the shape of the request that came in.

To make these debugging workflows easier, this change adds a new `auto` option to the `experimental_mock_subgraphs` plugin section of the configuration. It is off by default, but, when enabled, it will auto-generate response data for any subgraph which does not already have a static configuration. It also allows you to set an optional `seed` for the random number generators for cases where you want the same data to come back every time.

# Limitations

1. This implementation cannot generate data for custom scalars
2. The RNG is not portable, so seeded results may not be consistent across different architectures. If we want to use this for cross-platform snapshot tests in the future, we should switch to a portable generator like `rand_chacha`.

# Notes for Reviewers

The actual implementation of the plugin itself is in a3c368964d0cb0ce485aff897af2600a7eb6f74f. The following commit touches many files, but it just updates the test harness configurations to use the new format where static configurations are nested under `static_subgraphs`.

---

**Checklist**

- Tests added and passing[^4]
    - [X] Unit tests
    - [ ] Integration tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/configuration/snapshots/apollo_router__configuration__tests__schema_generation.snap
+++ b/apollo-router/src/configuration/snapshots/apollo_router__configuration__tests__schema_generation.snap
@@ -1347,8 +1347,8 @@ expression: "&schema"
       "description": "Telemetry configuration",
       "properties": {
         "apollo": {
-          "$ref": "#/definitions/Config13",
-          "description": "#/definitions/Config13"
+          "$ref": "#/definitions/Config14",
+          "description": "#/definitions/Config14"
         },
         "exporters": {
           "$ref": "#/definitions/Exporters",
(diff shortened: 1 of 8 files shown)


tninesling:
Going to reimplement with `arbitrary` and move over to `apollo-rs` to support fuzzing. I'll reopen this or an equivalent PR here when that shared crate is published. Then, we can consider pulling the shared crate into the `router-scale` project as well, since it has similar behavior.

PR #7956, opened 2025-07-23
response_cache: returns the right cache control in response header

bnjjj:
Before this fix, when having both public and private data in the same query it was returning public in cache-control header instead of private.

---

**Checklist**

- Tests added and passing[^4]
    - [ ] Unit tests
    - [ ] Integration tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/context/extensions/mod.rs
```diff
@@ -85,6 +85,18 @@ impl Extensions {
...
+        let map = self.map.get_or_insert_with(Box::default);
+        let value = map
+            .entry(TypeId::of::<T>())
+            .or_insert_with(|| Box::<T>::new(val));
+        // It should be impossible for the entry to be the wrong type as we don't allow direct access to the map.
+        value
+            .downcast_mut()
+            .expect("default value should be inserted and we should be able to downcast it")
+    }
+
```
carodewig: It looks like this new function is unused, is it still necessary?

bnjjj: Good catch, fixed.

PR #7957, opened 2025-07-23
[JSONSelection] Support `...` spread syntax for `NamedSelection` elements

benjamn:
In PR #7945, I did some internal refactoring so that `NamedSelection` could be represented by a single struct rather than multiple overlapping enum variants. One of the goals of that refactoring was to make it easier to add new `NamedSelection` grammar features, like the `...` syntax introduced by this PR. Adding this syntax is easier now because there's only one place it needs to be represented in the AST: the `NamedSelection::prefix` field, rather than multiple places (`NamedSelection::Path` and `NamedSelection::Field`).

With this refactoring, the distinction between a single-`Key` `PathSelection` and a `NamedFieldSelection` (which always has a single `Key`) has disappeared, so we can simplify/relax the EBNF grammar for `NamedSelection` in a fully backwards compatible way. The only impact of this simplification is that fewer `$.` prefixes should be needed to disambiguate single-key `PathSelection` elements, except where there is potential ambiguity with field selections. I could have put this simplification into its own PR, but it would immediately precede the `...` PR anyway. I hope combining the PRs is acceptable.


apollo-federation/src/connectors/json_selection/parser.rs
```diff
@@ -231,7 +231,15 @@ impl JSONSelection {
     }
 
     fn parse_span(input: Span) -> ParseResult<Self> {
+        match get_connect_spec(&input) {
+            ConnectSpec::V0_1 | ConnectSpec::V0_2 => Self::parse_span_v0_2(input),
+            ConnectSpec::V0_3 => Self::parse_span_v0_3(input),
+        }
+    }
+
+    fn parse_span_v0_2(input: Span) -> ParseResult<Self> {
```
benjamn, line 243: Here's how we're keeping the grammar simplification changes gated to `ConnectSpec::V0_3`.

apollo-federation/src/connectors/json_selection/apply_to.rs
```diff
@@ -3193,4 +3220,248 @@ mod tests {
...
+                    __typename: "Magazine",
+                    title: title,
+                    editor: editor.name,
+                }],
+                ["dummy", {}],
+                [@, null],
+            )
+            "#,
+            spec
+        );
```
benjamn: Here's a typical test case exercising the `...` for managing abstract/polymorphic types. As you can see, it's not just about selecting a valid `__typename`, but also about selecting the right fields to accompany each `__typename`.

Note that `...$(null)` or `...type->match([@, null])` causes the whole enclosing `SubSelection` to be `null`ified, which supports GraphQL error/`null` bubbling use cases.

apollo-federation/src/connectors/json_selection/parser.rs
```diff
@@ -597,19 +597,20 @@ impl NamedSelection {
...
+            ))(input.clone())
+            .map(|(remainder, (_spaces, spread, path))| {
+                let prefix = if let Some(spread) = spread {
+                    // An explicit ... spread token was used, so we record
+                    // NamingPrefix::Spread(Some(_)). If the path produces
+                    // something other than an object or null, we will catch
+                    // that in apply_to_path and compute_output_shape (not a
+                    // parsing concern).
+                    NamingPrefix::Spread(spread.range())
+                } else if path.has_subselection() {
```
benjamn: Besides the simplifications to the `KeyPath` grammar, the operative parsing change introduced by this PR can be found in these lines, where we enable parsing a `...` token before `PathSelection` elements (note: this change only affects the version-gated `NamedSelection::parse_v0_3` method).

apollo-federation/src/connectors/json_selection/apply_to.rs
```diff
@@ -92,19 +93,43 @@ impl JSONSelection {
...
+
+        let dollar_shape = input_shape.clone();
+
+        let handle = |computable: &dyn ApplyToInternal| -> Shape {
+            if context.named_shapes().get("$root") == Some(&input_shape) {
+                // If the $root variable is bound to the input shape, we
+                // can use it as the dollar_shape for the selection.
+                computable.compute_output_shape(context, input_shape, dollar_shape)
+            } else {
+                let cloned_context = context
```
andrewmcgivery: I like the comment on the `if`.. it would be helpful to have a similar command on the `else` with like "...otherwise" to complete the explaination.

apollo-federation/src/connectors/json_selection/apply_to.rs
```diff
@@ -92,19 +93,43 @@ impl JSONSelection {
...
-                selection.compute_output_shape(context, input_shape, dollar_shape)
-            }
-            TopLevelSelection::Path(path_selection) => {
-                let dollar_shape = input_shape.clone();
-                path_selection.compute_output_shape(context, input_shape, dollar_shape)
+        debug_assert_eq!(context.spec(), self.spec());
+
+        let dollar_shape = input_shape.clone();
+
+        let handle = |computable: &dyn ApplyToInternal| -> Shape {
```
andrewmcgivery: Is there a more descriptive name we could use than `handle`? I think I understand what this is doing but a descriptive function name here would certainly help future readers :)

apollo-federation/src/connectors/json_selection/apply_to.rs
```diff
@@ -378,91 +405,56 @@ impl ApplyToInternal for NamedSelection {
         let mut output: Option<JSON> = None;
         let mut errors = Vec::new();
 
-        match self {
-            Self::Field(alias, key, selection) => {
```
andrewmcgivery, line 382: Just so I'm understanding, with the new logic you've written, we no longer need specific logic for `Field`, `Path, and `Group`?

Almost seems too good to be true 😆

apollo-federation/src/connectors/json_selection/apply_to.rs
```diff
@@ -2518,51 +2486,9 @@ mod tests {
             ),
         );
 
-        // We have to construct this invalid selection manually because we want
-        // to test an error case requiring a PathWithSubSelection that does not
-        // actually have a SubSelection, which should not be possible to
-        // construct through normal parsing.
-        let invalid_inline_path_selection = JSONSelection::named(SubSelection {
```
andrewmcgivery, line 2525: I assume because we've "collapsed" our grammar and `PathWithSubSelection` is no longer a thing, this logic is no longer required?

apollo-federation/src/connectors/json_selection/apply_to.rs
```diff
@@ -2945,7 +2871,7 @@ mod tests {
         let root_shape = Shape::name("$root", []);
         let shape_context = ShapeContext::new(SourceId::Other("JSONSelection".into()))
             .with_spec(spec)
-            .with_named_shapes(&named_shapes);
+            .with_named_shapes(named_shapes.clone());
```
andrewmcgivery: I assume we can't avoid these `clone` and/or the overhead we've determined to be acceptable?

apollo-federation/src/connectors/json_selection/apply_to.rs
```diff
@@ -3282,4 +3208,248 @@ mod tests {
         //     "[{ k: \"wrapped\", v: $root }]",
         // );
     }
+
+    #[rstest]
+    #[case::v0_3(ConnectSpec::V0_3)]
+    fn test_spread_syntax(#[case] spec: ConnectSpec) {
```
andrewmcgivery: It looks like we're testing 3 different cases in this test (the three different `{ ... }` blocks.

Can we please split this into 3 different tests with a more descriptive name that follows the `should... when ...` naming convention? :) (As discussed in https://github.com/apollographql/rust-best-practices/blob/main/book/chapter_05.md#use-descriptive-names)

apollo-federation/src/connectors/json_selection/apply_to.rs
```diff
@@ -3282,4 +3208,248 @@ mod tests {
...
+                        spec,
+                    ),
+                ],
+            ),
+        );
+    }
+
+    #[rstest]
+    #[case::v0_3(ConnectSpec::V0_3)]
+    fn test_spread_with_match(#[case] spec: ConnectSpec) {
```
andrewmcgivery: Same here, this should ideally be multiple tests.

apollo-federation/src/connectors/json_selection/apply_to.rs
andrewmcgivery: Can we add test cases to explicitly test...

1. Spreading a `null`
2. Spreading something that cannot be spread and therefore generating an error

apollo-federation/src/connectors/json_selection/location.rs
```diff
@@ -216,25 +216,13 @@ pub(crate) mod strip_ranges {
 
     impl StripRanges for NamedSelection {
         fn strip_ranges(&self) -> Self {
-            match self {
```
andrewmcgivery, line 219: Similar question here... I assume the new simplified logic makes it so we don't need to explicitly run logic against `Field`, `Path`, and `Group`?

apollo-federation/src/connectors/json_selection/parser.rs
```diff
@@ -233,7 +234,15 @@ impl JSONSelection {
     }
 
     fn parse_span(input: Span) -> ParseResult<Self> {
+        match get_connect_spec(&input) {
+            ConnectSpec::V0_1 | ConnectSpec::V0_2 => Self::parse_span_v0_2(input),
+            ConnectSpec::V0_3 => Self::parse_span_v0_3(input),
```
andrewmcgivery, line 239: I'm not explicitly disagreeing with this approach because often the simplest answer is the best answer :D

Mostly just thinking through this and processing....

Is there any risk of repetitive code here or are the `parse` methods simple enough here that this doesn't matter? Looking at the implementation, it seems like there isn't a better way but just thinking out loud.

If `0.4` ends up being the same as `0.3`, I assume we would just do `ConnectSpec::V0_3 | ConnectSpec::V0_4`... but what if it's not the same? Do we have to introduce a `parse` function for every single connect version potentially?

Again... not suggesting changes here... just something for us to discuss/consider and that convo may be a very short "this is the best/only way". :)

apollo-federation/src/connectors/json_selection/parser.rs
```diff
@@ -779,7 +953,7 @@ impl PathList {
             // a helpful solution.
             return Err(nom_error_message(
                 input.clone(),
-                "Path selection must start with key., $variable, $, @, or $(expression)",
+                "Path selection must start with key, $variable, $, @, or $(expression)",
```
andrewmcgivery, line 956: Can they also start with `...` now or that not something we'd include here? I assume this is a user-facing error message?

apollo-federation/src/connectors/json_selection/parser.rs
```diff
@@ -3267,4 +3472,240 @@ mod tests {
         assert_debug_snapshot!(selection_object_path_v0_2);
         assert_debug_snapshot!(selection_array_path_v0_2);
     }
+
+    #[rstest]
+    #[case::v0_2(ConnectSpec::V0_2)]
+    fn test_single_key_paths_v0_2(#[case] spec: ConnectSpec) {
```
andrewmcgivery: Similar to my previous comments above, this looks like it could be 2-3 separate tests with more descriptive names. :)

apollo-federation/src/connectors/json_selection/parser.rs
```diff
@@ -3267,4 +3472,240 @@ mod tests {
...
+                fragment: "(b, c)".to_string(),
+                offset: 11,
+                spec: ConnectSpec::V0_2,
+            })
+        );
+    }
+
+    #[rstest]
+    #[case::v0_3(ConnectSpec::V0_3)]
+    fn test_single_key_paths_v0_3(#[case] spec: ConnectSpec) {
```
andrewmcgivery: Similar to my previous comments above, this looks like it could be 2-3 separate tests with more descriptive names. :)

apollo-federation/src/connectors/json_selection/parser.rs
benjamn: Yes, we implicitly need to provide a parse function for every `ConnectSpec` version, and even if the v0.3 and v0.4 behavior starts out the same (in some part of the parsing code), they might end up needing to be different as we add more features targeting v0.4, so we might start by adding `| ConnectSpec::V0_4` to an existing `match`, but we might end up needing a separate `::parse_v0_4` method.

One of the benefits of having separate `parse_v0_n` methods is that it effectively freezes/memorializes the old code/behavior. Yes, there are now multiple parsing code paths depending on the `ConnectSpec` version, but we (almost) never change or think about the older implementations, unless there's a change to the underlying representation that affects all implementations (like making `NamedSelection` a struct). When we're adding new features, we only have to update the latest parser code path (if parser changes are required).

apollo-federation/src/connectors/json_selection/parser.rs
```diff
@@ -3267,4 +3472,240 @@ mod tests {
...
+            panic!(
+                "Expected a valid selection, got error: {:?}",
+                sum_a_plus_b_plus_c
+            );
+        }
+    }
+
+    #[rstest]
+    #[case::v0_3(ConnectSpec::V0_3)]
+    fn test_basic_spread_parsing(#[case] spec: ConnectSpec) {
```
andrewmcgivery: Similar to my previous comments above, this looks like it could be 10 separate tests with more descriptive names. :)

I'd also consider possibly adding the arrays from `for input in [` as `case` params since you're already using `rstest`, but I'm not as worried about this one. :)

apollo-federation/src/connectors/json_selection/parser.rs
benjamn: Noted, but please keep commenting where you see opportunity!

I'm also going to stop using `rstest` for single-version tests like this, with `let spec = ConnectSpec::V0_3` in the function body instead.

apollo-federation/src/connectors/json_selection/parser.rs
benjamn: Perhaps also worth noting: even when we branch the parsing logic according to the `ConnectSpec` version, the code paths can still merge back together (or share significant logic), for example both `JSONSelection::parse_span_v0_2` and `::parse_span_v0_3` end up calling the version-free `SubSelection::parse_naked` method for parsing `SubSelection`s.

apollo-federation/src/connectors/json_selection/apply_to.rs
benjamn: Yes, that's the great unification!

Probably worth spelling out how field selections work now:
* Previously a `NamedSelection::Field` variant, a field selection is now a `NamedSelection` struct with `prefix == NamingPrefix::None`
* It has a single `PathList::Key` element at the beginning of its `path`, and nothing else after that (until we support the `?` operator)
* It can still have an `Alias`, in which case `prefix` is `NamingPrefix::Alias(Alias)` and the alias becomes the output key
* `NamedSelection::get_single_key` returns `Some(the_single_key)` when the above criteria are met, indicating the `NamedSelection` can serve as a field selection if called upon to do so. Anonymous paths like `$.data` return `None` from `get_single_key`.

That's why it's no longer a problem to write `$(a->mul(b, c))`: `b` and `c` might look like field selections, but they're parsed as `LitExpr::Path(PathSelection)` because they're passed as arguments to the `->mul` method. Previously, `PathSelection::parse` prevented parsing single-`Key` paths to avoid ambiguity, but this PR lifts that restriction (because it's the same `PathSelection` either way). We don't bother asking for `get_single_key` for `a` or `b` or `c` in this context, since there is no need for an output key (no larger `SubSelection`).

apollo-federation/src/connectors/json_selection/apply_to.rs
benjamn: Yes, the `invalid_inline_path_selection` selection was only invalid (before this PR) because we had to prevent single-`Key` paths to avoid ambiguity with `NamedSelection::Field`.

Also, the test previously required constructing an AST by hand that couldn't be parsed from any source string, so I don't think we're losing much by removing the test.

apollo-federation/src/connectors/json_selection/apply_to.rs
benjamn: Looks like I can pass `.with_named_shapes(named_shapes)` since `named_shapes` is not used later, and that works the same. Cool!

In general, I expect the `named_shapes` of most `ShapeContext`s will remain unchanged during `compute_output_shape`, so the `context: &ShapeContext` can be passed down as-is, avoiding cloning.

apollo-federation/src/connectors/json_selection/apply_to.rs
benjamn: They all share the common setup code (`a_b_data`, `shape_context`, `root_shape`) though. How do you handle that with separate tests?

apollo-federation/src/connectors/json_selection/apply_to.rs
benjamn: In my mind this is a multi-faceted test of a single selection string. I don't really want to duplicate the selection string across multiple tests. Any ideas for avoiding the duplication with multiple tests?

apollo-federation/src/connectors/json_selection/apply_to.rs
andrewmcgivery: The simplest answer would be just copy/paste into each test. 😅 

If this was shared amongst most or all of your tests, I'd suggest a shared function that each test can call (I do that in the shape tests of methods... for example: https://github.com/apollographql/router/blob/dev/apollo-federation/src/connectors/json_selection/methods/public/and.rs#L255) but given it's only used for these 3 tests, I'd personally probably just copy/paste.... but I'd also be okay with a shared method if you'd prefer.

apollo-federation/src/connectors/json_selection/apply_to.rs
andrewmcgivery: You could definitely used a shared function here similar to method shape tests!

E.g. https://github.com/apollographql/router/blob/dev/apollo-federation/src/connectors/json_selection/methods/public/and.rs#L255

apollo-federation/src/connectors/json_selection/apply_to.rs
andrewmcgivery: One thing to consider (totally optional) is having your `...` tests in a separate `mod` in this file so they are grouped together and your "shared functions" are local to that mod instead of being in the larger test mod.

I do this in methods with separate `method_tests` and `shape_tests` mods. :)

apollo-federation/src/connectors/json_selection/parser.rs
benjamn: We used to require the `.` after the `key` because we disallowed single-`Key` paths, but now I believe the only syntaxes that are disallowed by this error are… (1) starting a path with a `->` method, (2) empty paths, and (3) paths that begin with just a `SubSelection`. Not thrilled with the clarity of the current error, to say the least.

As for `...`, while mentioning that might be helpful, the `...` token is not parsed at this level (see `NamedSelection::parse_v0_3` instead), so might be a little stretch-y for this code to generate errors about it.

apollo-federation/src/connectors/json_selection/apply_to.rs
benjamn: https://github.com/apollographql/router/pull/7957/commits/9b9e09f7e96c797750356fe593bdadd3d5d0ac1b

apollo-federation/src/connectors/json_selection/apply_to.rs
benjamn: https://github.com/apollographql/router/pull/7957/commits/9b9e09f7e96c797750356fe593bdadd3d5d0ac1b

apollo-federation/src/connectors/json_selection/apply_to.rs
benjamn: https://github.com/apollographql/router/pull/7957/commits/1ad573152bbd5f8bd4dcfb6ea0e6b1ade667feea

apollo-federation/src/connectors/json_selection/apply_to.rs
benjamn: https://github.com/apollographql/router/pull/7957/commits/e2fdb1385546e0879e677850b2ec0c8932d19341

apollo-federation/src/connectors/json_selection/parser.rs
benjamn: https://github.com/apollographql/router/pull/7957/commits/94147e8775649a941cc58ce3e354c4999631fdf1

apollo-federation/src/connectors/json_selection/parser.rs
benjamn: https://github.com/apollographql/router/pull/7957/commits/94147e8775649a941cc58ce3e354c4999631fdf1

apollo-federation/src/connectors/json_selection/parser.rs
benjamn: https://github.com/apollographql/router/pull/7957/commits/94147e8775649a941cc58ce3e354c4999631fdf1

benjamn: And unrolling the loops for easier test failure debugging: https://github.com/apollographql/router/pull/7957/commits/9e3e14d405b4283e977afdab928f8f46ff8efd22

apollo-federation/src/connectors/json_selection/apply_to.rs
benjamn: https://github.com/apollographql/router/pull/7957/commits/acf66c0dda3a0fef9874bc640515a67dddb054d6

apollo-federation/src/connectors/json_selection/apply_to.rs
benjamn: https://github.com/apollographql/router/pull/7957/commits/dadb57c6e2fce7c03500629bb0901c86ebaa494b

PR #7959, opened 2025-07-23
response_cache: fix the way private_queries are cached in the router

bnjjj:
Remove the unbounded vec to store private queries and replace it with a LRU. Add metric `apollo.router.response_cache.private_queries.lru.size` to know the size of this LRU to be able to configure it properly if needed.

Improve the way private_queries are stored to make it possible to have some kind of polymorphic queries (either public or private depending of private_id)

---

**Checklist**

- Tests added and passing[^4]
    - [x] Unit tests
    - [x] Integration tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-1395]: https://apollographql.atlassian.net/browse/ROUTER-1395?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


goto-bus-stop:
https://github.com/apollographql/router/pull/7965 will fix ci :)

apollo-router/src/plugins/response_cache/plugin.rs
```diff
@@ -834,15 +856,19 @@ impl CacheService {
...
-            .unwrap_or_default();
-
-        let is_known_private = { self.private_queries.read().await.contains(&query) };
         let private_id = self.get_private_id(&request.context);
+        // Knowing if there's a private_id or not will differentiate the hash because for a same query it can be both public and private depending if we have private_id set or not
+        let private_query_hash = format!(
+            "hash:{}:id:{}",
+            hash_query(&request.query_hash, request.subgraph_request.body()),
+            private_id.is_some()
+        );
```
SimonSapin: The only purpose of `format!` here is to construct a string key for `LruCache<String, ()>`, right? If so, let’s replace it with `LruCache<(String, bool), ()>`, or `LruCache<QueryPrivacyKey, ()>` with:

```rust
struct QueryPrivacyKey {
    query_hash: String,
    has_private_id: bool,
}
```

apollo-router/src/plugins/response_cache/tests.rs
```diff
@@ -1320,7 +1320,272 @@ async fn no_store_from_request() {
...
+            tls: Default::default(),
+            url: "postgres://127.0.0.1".parse().unwrap(),
+            username: None,
+            password: None,
+            idle_timeout: std::time::Duration::from_secs(5),
+            acquire_timeout: std::time::Duration::from_millis(50),
+            required_to_start: true,
+            pool_size: default_pool_size(),
+            batch_size: default_batch_size(),
+            namespace: Some(String::from("private_only")),
```
SimonSapin, line 1365: Why not (here and in every test using Redis or Postgres) use a random `namespace` to isolate a test run not only from other tests but from other runs of the same test? In some integration tests I’ve used this:

```rust
pub(crate) fn namespace() -> String {
    uuid::Uuid::new_v4().simple().to_string()
}
```

apollo-router/src/plugins/response_cache/plugin.rs
bnjjj: not really, it also creates a hash of the request body (with variables and so one). Your solution might work but that would imply cloning the whole body I think

apollo-router/src/plugins/response_cache/tests.rs
bnjjj: That wouldn’t work because we also create snapshots of debugging data, which includes the key and its namespace. Redacting this information could lead to missing critical details if the key changes unexpectedly.

apollo-router/src/plugins/response_cache/plugin.rs
bnjjj: Fixed

SimonSapin (approved):

PR #7993, opened 2025-07-28
Future-proof JSONSelection grammar/parser to prevent multiple sequential `?`s

benjamn:
This is a PR-onto-a-PR targeting @andrewmcgivery's #7637.

As I foreshadowed in https://github.com/apollographql/router/pull/7637#discussion_r2190638385, I think we will be doing ourselves a favor to think ahead to other syntax the `?` operator could conflict with.

The simplicity of `PathStep ::= "." Key | "->" Identifier MethodArgs? | "?"` was nice (since `?` is just another `PathStep`), but eventually we may want infix operators like `??` and `?!`, and it should be possible to parse `a??b` or `a? ?? b?` without ambiguity. This PR should protect us by ensuring no more than one consecutive `?` is ever considered to be a `PathList::Question`, and by refusing to parse a single `?` if the text starts with `??` or `?!`.

I'm leaving this PR as a draft because it currently edits parsing code that should really be version-gated to `connect/v0.3` only. I will rebase after PR #7637 rebases to include the version gating functionality tracked in PR #7926.


apollo-federation/src/connectors/json_selection/parser.rs
```diff
@@ -707,12 +712,34 @@ impl PathList {
...
-            return Ok((remainder, WithRange::new(Self::Question(tail), full_range)));
+        // At any depth, if the next token is ? but not the PathList::Question
+        // kind, we terminate path parsing so the hypothetical ?? or ?! tokens
+        // have a chance to be parsed as infix operators.
+        if input.fragment().starts_with("??") || input.fragment().starts_with("?!") {
+            return Ok((input, WithRange::new(Self::Empty, range_if_empty)));
+        }
+
+        if let Ok((suffix, question)) = ranged_span("?")(input.clone()) {
+            let (remainder, rest) = Self::parse_with_depth(suffix, depth + 1)?;
```
benjamn: Hmm, I'm now realizing/remembering these parser changes won't cover `NamedFieldSelection`s like `age?` in `author { name age? }`, so we really do need the `NamedSelection` unification from #7945 to provide `?` syntax everywhere we want it (not just in `NamedPathSelection`).

apollo-federation/src/connectors/json_selection/parser.rs
```diff
@@ -3683,8 +3798,60 @@ mod tests {
...
+                message: "nom::error::ErrorKind::Eof".to_string(),
+                fragment: "??.bar".to_string(),
+                offset: 23,
+                spec,
+            }),
+        );
+    }
+
+    #[test]
+    fn test_invalid_infix_operator_parsing() {
```
andrewmcgivery, line 3857: I assume once we introduce `??` and `?!` in upcoming work that this test will change since we expect it to error on `0.2` but not error on `0.3`?

apollo-federation/src/connectors/json_selection/parser.rs
```diff
@@ -994,14 +999,45 @@ impl PathList {
...
+                    return match rest.as_ref() {
+                        // The ? cannot be repeated sequentially, so if rest starts with
+                        // another PathList::Question, we terminate the current path,
+                        // probably (but not necessarily) leading to a parse error for
+                        // the upcoming ?.
+                        PathList::Question(_) => {
+                            let empty_range = question.range().map(|range| range.end..range.end);
+                            let empty = WithRange::new(Self::Empty, empty_range);
+                            Ok((
+                                suffix.clone(),
```
andrewmcgivery: Is this second clone required? I would think not since this is the last usage of it.

apollo-federation/src/connectors/json_selection/parser.rs
benjamn: Good catch, I'll `spec` this test to v0.2 instead of v0.3.

apollo-federation/src/connectors/json_selection/parser.rs
benjamn: 'Twas not required after all

apollo-federation/src/connectors/json_selection/parser.rs
```diff
@@ -2768,17 +2879,34 @@ mod tests {
...
-        );
+    #[test]
+    fn test_path_expr_with_spaces_v0_2() {
+        assert_debug_snapshot!(selection!(
+            " suffix : results -> slice ( $( - 1 ) -> mul ( $args . suffixLength ) ) ",
+            // Snapshot tests can be brittle when used with (multiple) #[rstest]
+            // cases, since the filenames of the snapshots do not always take
+            // into account the differences between the cases, so we hard-code
+            // the ConnectSpec in tests like this.
+            ConnectSpec::V0_2
```
benjamn, line 2892: Here's my rule of thumb about (not) using `#[rstest]` for tests that use `assert_debug_snapshot!`.

benjamn:
Commit https://github.com/apollographql/router/pull/7993/commits/dd5fd4320268e5d1bc13999f056e196c6c229023 is noteworthy because it adds `ConnectSpec` version gating for the `?` syntax, which involves updating all tests that use `?` syntax to parse with `ConnectSpec::V0_3` (as well as other future versions, as necessary).

The `?` syntax may seem like an additive feature, but syntax changes can have ripple effects, so this gating keeps the `ConnectSpec::V0_2` behavior the same as ever, until developers choose to upgrade.

PR #8000, opened 2025-07-28
Simplify `deserialize_response`.

dylan-apollo:
Specifically pull out `debug_context`, which will make this function easier to move to and consume from `apollo-federation`


apollo-router/src/plugins/connectors/handle_responses.rs
```diff
@@ -125,13 +126,17 @@ pub(crate) async fn process_response<T: HttpBody>(
...
                 .and_then(|body| {
                     let body = body.to_bytes();
-                    let raw = deserialize_response(
-                        &body,
-                        &parts,
-                        &connector,
-                        debug_context,
-                        &debug_request,
-                    );
+                    let raw = deserialize_response(&body, &parts.headers).map_err(|_| {
```
Hunter-apollo, line 129: I really like pulling the debug setting up a level.

apollo-router/src/plugins/connectors/handle_responses.rs
```diff
@@ -125,13 +126,17 @@ pub(crate) async fn process_response<T: HttpBody>(
...
+                        if let Some(debug_context) = debug_context {
+                            debug_context.lock().push_invalid_response(
+                                debug_request.0.clone(),
```
Hunter-apollo, line 132: Out of scope for this PR, but I wonder if we should replace this tuple with a better named struct.

apollo-router/src/plugins/connectors/handle_responses.rs
```diff
@@ -253,17 +247,7 @@ fn deserialize_response(
...
-                debug_context.lock().push_invalid_response(
-                    debug_request.0.clone(),
-                    parts,
-                    body,
-                    &connector.error_settings,
-                    debug_request.1.clone(),
-                );
-            }
-        })
+        serde_json::from_slice::<Value>(body).map_err(|_| ())
```
Hunter-apollo, line 250: Why can we ignore the error here? Seems like it would indicate a serialization error?

dylan-apollo: The fact that there was _an_ error still gets reported up as `Err(())`, we were just already throwing away the error message in the previous version.

In a future PR I'll actually preserve the error so we can use it in click-to-run

PR #8001, opened 2025-07-28
Simplify `handle_raw_response`.

dylan-apollo:
Gets rid of the `RawResponse` enum and pulls `debug_context` up into the router layer.


apollo-federation/src/connectors/runtime/responses.rs
```diff
@@ -32,252 +28,166 @@ pub enum HandleResponseError {
...
+    data: &Value,
+    parts: &Parts,
+    key: ResponseKey,
+    connector: &Connector,
+    context: impl ContextReader,
+    client_headers: &HeaderMap<HeaderValue>,
+) -> MappedResponse {
+    let inputs = key
+        .inputs()
+        .clone()
```
naomijub, line 56: cant we use `&` or `as_ref` here? borrowed lifetime issue?
it seems that we only use `inputs` as a ref, then key is passed to `MappedResponse`

naomijub: same for `map_error`

apollo-federation/src/connectors/runtime/responses.rs
```diff
@@ -32,252 +28,166 @@ pub enum HandleResponseError {
...
-
-                    if let Some(code) = extensions.get("code") {
-                        extension_code = code.as_str().unwrap_or_default().to_string();
-                    }
-
-                    for (key, value) in extensions {
-                        error = error.extension(key.clone(), value.clone());
-                    }
-                }
+        for (key, value) in extensions {
```
Hunter-apollo, line 154: thought: It looks like the `extensions` var isnt used again so this might be able to be an owning iterator rather than cloning k/v.

apollo-federation/src/connectors/runtime/responses.rs
```diff
@@ -32,252 +28,166 @@ pub enum HandleResponseError {
...
+    data: &Value,
+    parts: &Parts,
+    key: ResponseKey,
+    connector: &Connector,
+    context: impl ContextReader,
+    client_headers: &HeaderMap<HeaderValue>,
+) -> MappedResponse {
+    let mut problems = Vec::new();
+
+    let inputs = LazyCell::new(|| {
```
naomijub, line 88: why this one is a `LazyCell` and the `map_data` is not?

apollo-federation/src/connectors/runtime/responses.rs
```diff
@@ -32,252 +28,166 @@ pub enum HandleResponseError {
...
+            ProblemLocation::SourceErrorsExtensions,
+        ));
+
+        // TODO: Currently this "fails silently". In the future, we probably add a warning to the debugger info.
+        let extensions = res
+            .and_then(|e| match e {
+                Value::Object(map) => Some(map),
+                _ => None,
+            })
+            .unwrap_or_default();
```
naomijub, line 148: Other ways of writing this block:

```rust
// If you want ownership
    let response = res.filter(|json| json.is_object()).map_or(Value::Null, |mut json| json.take());
    let response = res.filter(|json| json.is_object()).map_or_else(Default::default, |mut json| json.take());
    // if you are ok with references
    let response = res.filter(|json| json.is_object()).and_then(|json| json.get());
```

apollo-router/src/plugins/connectors/handle_responses.rs
```diff
@@ -145,28 +140,46 @@ pub(crate) async fn process_response<T: HttpBody>(
...
-                debug_context,
-                supergraph_request.headers(),
-            );
-
-            if is_success {
+            if let Some(debug) = debug_context {
+                let mut debug_problems: Vec<Problem> = mapped.problems().to_vec();
+                debug_problems.extend(debug_request.1);
+
+                let selection_data = if let MappedResponse::Data { key, data, .. } = &mapped {
```
naomijub, line 163: maybe this could become a function in `MappedResponse`, something like `map_data`:

```
fn map_selection_data(self) -> Option<SelectionData> {
  if let MappedResponse::Data { key, data, .. } = &mapped {
    Some(SelectionData {
        source: connector.selection.to_string(),
        transformed: key.selection().to_string(),
        result: Some(data.clone()),
    })
  } else {
    None
  }
}
```

apollo-federation/src/connectors/runtime/responses.rs
dylan-apollo: We move `key` into `MappedResponse` at the end to return it back to the caller. I don't think the inputs/context/key API overall is very good, but think we should save that part of the refactor for another time.

apollo-federation/src/connectors/runtime/responses.rs
dylan-apollo: In fact, it was already owned but cloning for some reason anyway? 😆

apollo-federation/src/connectors/runtime/responses.rs
dylan-apollo: Because `map_data` always needs this value, but `map_error` sometimes doesn't need it. However this is all gonna change with Hunter's `isSuccess` branch

apollo-federation/src/connectors/runtime/responses.rs
dylan-apollo: Those would all give back `Value`, but we specifically need the inner `Map` type.

apollo-router/src/plugins/connectors/handle_responses.rs
dylan-apollo: `SelectionData` is a router-specific structure, and `MappedResponse` is in `apollo-federation`, so I don't want to move more stuff into the shared space that only router uses. However, I'm not convinced that even `MappedResponse` belongs in `apollo-federation`, since `ResponseKey` is very router-specific, so maybe in the future we can clean this up with more refactoring.

apollo-federation/src/connectors/runtime/responses.rs
naomijub: if you need the map type you can use `.as_object()`

naomijub (approved):

PR #8013, opened 2025-07-30
[Subgraph Insights] Add Apollo Subgraph Fetch Histogram to Telemetry Plugin

rregitsky:
---
**Description**
This change adds a new, experimental histogram to capture subgraph fetch duration for Apollo Studio. The instrument,
`apollo.router.operations.fetch.duration` has the following attributes:
- `apollo.client.name`
- `apollo.client.version`
- `has_errors`
- `apollo.operation.id`
- `graphql.operation.kind`
- `graphql.operation.name`
- `subgraph.name`

This can be toggled on using a new boolean config flag:
```yaml
telemetry:
  apollo:
    experimental_subgraph_metrics: true
```
The instrument is currently only sent to GraphOS and is not available in 3rd-party OTel export targets. It is not
user customizable. For this purpose, users can take advantage of the existing customizable instrument 
`http.client.request.duration` measuring the same value.

**Implementation**
This is implemented using the custom instrument framework. These are Apollo-controlled metrics, so rather than allowing for user customization like existing use cases, we hardcode the `CustomHistogram` with specific attributes. This allows us to add more easily add apollo metrics in the future and means we can use the shared standard ways of pulling data off of the request/response/context.

Cardinality is high, so the new metric follows the `realtime_metrics` path which allows for the `scheduled_delay` config to change how often the metric is sent.

**Checklist**

- Tests added and passing[^4]
    - [x] Unit tests
    - [x] Integration tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-1406]: https://apollographql.atlassian.net/browse/ROUTER-1406?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.changesets/feat_rreg_experimental_subgraph_metrics.md
```diff
@@ -0,0 +1,27 @@
+### [Subgraph Insights] Experimental Apollo Subgraph Fetch Histogram ([PR #8013](https://github.com/apollographql/router/pull/8013))
+
+<!-- start metadata -->
+
+<!-- [PULSR-1673] -->
```
bonnici: I've never seen these tags before so i'm not sure what they're for, but we probably shouldn't leave the pulsr ticket reference here?

apollo-router/src/plugins/telemetry/config_new/apollo/instruments.rs
```diff
@@ -0,0 +1,182 @@
...
+}
+
+impl ApolloSubgraphInstruments {
+    pub(crate) fn new(
+        static_instruments: Arc<HashMap<String, StaticInstrument>>,
+        apollo_config: Config,
+    ) -> Self {
+        let selectors = Extendable {
+            attributes: SubgraphAttributes::builder()
+                .subgraph_name(StandardAttribute::Bool(true))
```
bonnici, line 53: ~I'm new to this code, but I don't understand why subgraph_name is initialised to true?~
Answered in the pairing recording.

apollo-router/src/plugins/telemetry/config_new/apollo/instruments.rs
```diff
@@ -0,0 +1,182 @@
...
+    pub(crate) fn new_builtin() -> HashMap<String, StaticInstrument> {
+        let meter = metrics::meter_provider().meter(METER_NAME);
+        let mut static_instruments = HashMap::with_capacity(1);
+
+        static_instruments.insert(
+            APOLLO_ROUTER_OPERATIONS_FETCH_DURATION.to_string(),
+            StaticInstrument::Histogram(
+                meter
+                    .f64_histogram(APOLLO_ROUTER_OPERATIONS_FETCH_DURATION)
+                    .with_unit("s")
```
bonnici, line 138: I guess seconds is the standard unit to use? Milliseconds or nanoseconds makes more sense to me but again I'm new to this code.

apollo-router/src/plugins/telemetry/config_new/apollo/instruments.rs
```diff
@@ -0,0 +1,182 @@
...
+                (
+                    "client.version".to_string(),
+                    SubgraphSelector::ResponseContext {
+                        response_context: CLIENT_VERSION.to_string(),
+                        redact: None,
+                        default: None,
+                    },
+                ),
+                (
+                    "operation.name".to_string(),
```
timbotnik: Should be `graphql.operation.name` (perhaps one of the default attributes)

apollo-router/src/plugins/telemetry/config_new/apollo/instruments.rs
```diff
@@ -0,0 +1,182 @@
...
+impl ApolloSubgraphInstruments {
+    pub(crate) fn new(
+        static_instruments: Arc<HashMap<String, StaticInstrument>>,
+        apollo_config: Config,
+    ) -> Self {
+        let selectors = Extendable {
+            attributes: SubgraphAttributes::builder()
+                .subgraph_name(StandardAttribute::Bool(true))
+                .graphql_operation_type(StandardAttribute::Aliased {
+                    alias: "operation.kind".to_string(),
```
timbotnik: We're looking for `graphql.operation.type` on the ingestion side so this probably should come from the supergraph operation type and shouldn't require an alias.  Then again, I wonder if there are any use-cases where the supergraph operation type and the subgraph operation type can be different?  Actually in the case of federation or subscriptions perhaps it can be different... in which case we might actually want to track both the supergraph type and subgraph type.  Let's ask someone from the Router team to confirm that it's possible.

apollo-router/src/plugins/telemetry/config_new/apollo/instruments.rs
```diff
@@ -0,0 +1,182 @@
...
+                (
+                    "operation.name".to_string(),
+                    SubgraphSelector::SupergraphOperationName {
+                        supergraph_operation_name: OperationName::String,
+                        redact: None,
+                        default: None,
+                    },
+                ),
+                (
+                    "operation.id".to_string(),
```
timbotnik:
```suggestion
                    "apollo.operation.id".to_string(),
```

.changesets/feat_rreg_experimental_subgraph_metrics.md
rregitsky: It's metadata so it's hidden. It's specifically for adding a `ROUTER` ticket number, so I figured we might as well use it for us.

apollo-router/src/plugins/telemetry/config_new/apollo/instruments.rs
rregitsky: We actually don't have control over this for `CustomHistogram`. 
```rust
Some(instant.elapsed().as_secs_f64())
```
In fact, all histogram refs I could find in the router were all using `f64` to measure seconds.

apollo-router/src/plugins/telemetry/config_new/apollo/instruments.rs
rregitsky: The implementation here actually notes that it should always be the same:
https://github.com/apollographql/router/blob/542c41649c6a30be6e0cda83713a8b415cc70ff4/apollo-router/src/plugins/telemetry/config_new/subgraph/attributes.rs#L99

That said, I may as well pull it from the context to make it "future proof" in case that changes in the future.

.changesets/feat_rreg_experimental_subgraph_metrics.md
rregitsky: Whoops, yeah. This pulls directly from the github description now which has the metadata included.

apollo-router/src/plugins/telemetry/config_new/apollo/instruments.rs
rregitsky: Asked about whether they can differ here: https://apollograph.slack.com/archives/C02UX05LF4K/p1753980692829289?thread_ts=1753971835.882889&cid=C02UX05LF4K

apollo-router/src/plugins/telemetry/config_new/apollo/instruments.rs
timbotnik: I suppose if it _was_ possible for this to be milliseconds maybe we would be able to use a 32 bit value and save some space over the wire.  I'd need to also handle this on the ingestion side since we don't respect the `unit` yet and just assume that everything is in seconds.  https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto#L198 - not sure that this would work since the protobuf specifies a `double` type for the bounds.

apollo-router/src/plugins/telemetry/config_new/apollo/instruments.rs
```diff
@@ -0,0 +1,184 @@
...
+                    SubgraphSelector::ResponseContext {
+                        response_context: APOLLO_OPERATION_ID.to_string(),
+                        redact: None,
+                        default: None,
+                    },
+                ),
+                (
+                    "has.errors".to_string(),
```
timbotnik: Would be good to check if there is any prior art / semantic convention for this.  We could use `error.type` as described here https://opentelemetry.io/docs/specs/semconv/general/recording-errors/#recording-errors-on-metrics, but I'm not sure we can do this easily, and we don't need to differentiate the type at this stage.

Assuming we stick with boolean, `has_errors` would be better since the `.` notation for OTel attributes is typically used as a hierarchical namespace delimiter.

apollo-router/src/plugins/telemetry/config_new/apollo/instruments.rs
goto-bus-stop: As far as I understand it is indeed the standard unit to use for all time measurements, but anyone displaying it would convert the unit to a most-appropriate scale at that point

apollo-router/tests/integration/telemetry/apollo_otel_metrics.rs
```diff
@@ -493,25 +494,93 @@ async fn test_router_layer_error_emits_metric() {
     router.graceful_shutdown().await;
 }
 
+#[tokio::test(flavor = "multi_thread")]
+async fn test_subgraph_requets_emits_histogram() {
```
goto-bus-stop:
```suggestion
async fn test_subgraph_request_emits_histogram() {
```

.changesets/feat_rreg_experimental_subgraph_metrics.md
```diff
@@ -0,0 +1,23 @@
...
+This change adds a new, experimental histogram to capture subgraph fetch duration for Apollo Studio. The instrument,
+`apollo.router.operations.fetch.duration` has the following attributes:
+- client.name
+- client.version
+- has.errors
+- operation.id
+- graphql.operation.type
+- graphql.operation.name
+- subgraph.name
+
```
goto-bus-stop: This feels all like implementation details because the instrument is not actually exposed to the user at all.

I think we should trim the changeset down to just mention:
- We now optionally send subgraph request durations to GraphOS.
- Here's what that enables for users in Studio
- Here's how you enable the experimental configuration

apollo-router/src/plugins/telemetry/config_new/apollo/instruments.rs
```diff
@@ -0,0 +1,184 @@
...
+        };
+
+        let apollo_router_operations_fetch_duration =
+            apollo_config.experimental_subgraph_metrics.then(|| {
+                CustomHistogram {
+                    inner: Mutex::new(
+                        CustomHistogramInner {
+                            increment: Increment::Duration(Instant::now()),
+                            condition: Condition::True,
+                            attributes: Vec::with_capacity(7),
```
goto-bus-stop: Must this number be kept in sync with something? Can we derive it instead of hardcoding it?

apollo-router/src/plugins/telemetry/config_new/apollo/instruments.rs
```diff
@@ -0,0 +1,184 @@
...
+                                .as_histogram()
+                                .cloned()
+                                .expect(
+                                    "cannot convert apollo instrument to histogram for subgraph; this should not happen",
+                                )
+                            ),
+                            updated: false,
+                            _phantom: PhantomData,
+                        })
+                }
```
goto-bus-stop: I think this could benefit from having a constructor function in `CustomHistogram` rather than building it up out of internal structures here.

goto-bus-stop:
Other than the above, LGTM

apollo-router/src/plugins/telemetry/config_new/apollo/instruments.rs
rregitsky: It needs to be in sync with the number of attributes defined in `selectors` above. We can _somewhat_ derive it as `custom` is in a countable data structure. Unfortunately the `attributes` struct has no good way of counting enabled attrs, so I've still hardcoded a number for now.

apollo-router/src/plugins/telemetry/config_new/apollo/instruments.rs
rregitsky: @timbotnik as I mentioned, `CustomHistogram` controls the unit recorded to the underlying histogram. We could likely add a way to specify a unit in its constructor, but if that's the route we want I'd prefer to add it as a follow-up to this PR.

apollo-router/src/plugins/telemetry/config_new/apollo/instruments.rs
rregitsky: Yeah, `error.type` doesn't seem like an easy thing to implement. Would require us to make a new selector (available publicly) and create buckets to group types of errors, which I expect isn't a trivial task.

Our instruments readme uses `graphql.errors`: https://github.com/apollographql/router/blob/814d4327d9bd3e7e1842ee02e17eb9f50840ccd9/docs/source/routing/observability/telemetry/instrumentation/instruments.mdx#L551

But otherwise not seeing any convention for the boolean flag attr naming. I'll stick with `has_errors`.

apollo-router/src/plugins/telemetry/config_new/apollo/instruments.rs
timbotnik: IMO there's not a great reason to change this now (an argument about precision would be the only reason, but the numbers we're dealing with are precise enough).  Let's make a tech debt ticket which would include updating the ingestion side to fail gracefully on different units (or support them!)

apollo-router/src/plugins/telemetry/metrics/apollo/mod.rs
```diff
@@ -187,7 +188,10 @@ impl Config {
             )),
             Box::new(
                 CustomAggregationSelector::builder()
-                    .boundaries(default_buckets())
+                    .boundaries(
+                        // Returns [~1.4ms ... ~5min]
```
timbotnik, line 194: I think we could add some additional comments about how this maps to the "Apollo" histogram format here.

> a duration x in μs is counted in the bucket of index max(0, min(ceil(ln(x)/ln(1.1)), 383)).

timbotnik (approved):

apollo-router/src/plugins/telemetry/config_new/apollo/instruments.rs
```diff
@@ -0,0 +1,170 @@
...
+        static_instruments: Arc<HashMap<String, StaticInstrument>>,
+        apollo_config: Config,
+    ) -> Self {
+        let selectors = Extendable {
+            attributes: SubgraphAttributes::builder()
+                .subgraph_name(StandardAttribute::Bool(true))
+                .build(),
+            custom: HashMap::from([
+                (
+                    "client.name".to_string(),
```
timbotnik:
```suggestion
                    "apollo.client.name".to_string(),
```

apollo-router/src/plugins/telemetry/config_new/apollo/instruments.rs
```diff
@@ -0,0 +1,170 @@
...
+                    SubgraphSelector::ResponseContext {
+                        response_context: CLIENT_NAME.to_string(),
+                        redact: None,
+                        default: None,
+                    },
+                ),
+                (
+                    "client.version".to_string(),
```
timbotnik:
```suggestion
                    "apollo.client.version".to_string(),
```

apollo-router/src/plugins/telemetry/config_new/apollo/instruments.rs
timbotnik: If possible would be nice to be using shared constants for these.

apollo-router/src/plugins/telemetry/config_new/apollo/instruments.rs
rregitsky: https://apollographql.atlassian.net/browse/PULSR-1750

PR #8015, opened 2025-07-30
add_join_type and descendents

clenfest:


clenfest:
Note to reviewers: I know that merge_type() itself is a little messed up with all the different signature types. Will fix that in a follow-up PR. Also, I'm not sure exactly about `serialize_directive_arguments()`, there is probably a better way.

apollo-federation/src/link/join_spec_definition.rs
```diff
@@ -256,6 +256,19 @@ impl JoinSpecDefinition {
         }
     }
 
+    pub(crate) fn join_type_definition<'schema>(
```
duckki, line 259: We usually use `XXX_directive_definition` or `XXX_type_definition`. In this case, shouldn't it be `type_directive_definition`?

tninesling:
Superceded by #8088

Issue #6004, opened 2024-09-15
Bug: Query Planner Fails for Nested Entity Fields Decorated With `@requires`


ryan01010111:
## Describe the bug
**Note**: This issue was first discovered in the `query-planner` package of the `federation` repo, and [an issue](https://github.com/apollographql/federation/issues/3113) has already been opened there.

When building a query plan for a query that selects fields on a nested entity type, and that are decorated with `@requires`, the router fails, resulting in error `Cannot add selection of field \"X\" to selection set of parent type \"Y\"`. This issue is blocking us from migrating to Router.

## To Reproduce
1. Fork [this CodeSandbox](https://codesandbox.io/p/devbox/apollo-router-query-planner-bug-kyh3w4)
2. Start all services (if not already running) with:
```bash
npm run dev
```
3. Run test query to reproduce the error with:
```bash
npm run test:query
```

## Expected behavior
The query plan to be successfully built and executed.

## Output
Router log:
```json
{"timestamp":"2024-09-15T04:36:25.608204335Z","level":"WARN","message":"an error happened in the worker runtime Cannot add selection of field \"Book.id\" to selection set of parent type \"Book\"\n\n","target":"router_bridge::worker","resource":{"service.name":"unknown_service:router","service.version":"1.54.0","process.executable.name":"router"}}
```

## Desktop (please complete the following information):
 - OS: Debian 12
 - Version: 1.54.0

duckki:
Note that the new native query planner has the same issue.

PR #8016, opened 2025-07-30
fix(federation/query-planning): Fix bug where subgraph jump key fields are fetched from the wrong node

sachindshinde:
When a `@requires` edge needs a subgraph jump during fetch dependency graph processing, this would sometimes cause the locally satisfiable key fields needed by that jump to be fetched from the parent node instead of the current node (which apparently was done as an optimization). This can be a problem, since `GraphPath::can_satisfy_conditions()` only really guarantees the current node has those key fields.

This PR updates the code using the parent node to now explicitly check whether the locally satisfiable key can be fetched from that node, and if not, falls back to the current node (adding it as a parent, if needed). This lets us keep the optimization and avoid unnecessarily changing query plans.

This PR mirrors its JS equivalent at https://github.com/apollographql/federation/pull/3293 , and fixes https://github.com/apollographql/router/issues/6004 .

### DIFF
--- a/.changesets/fix_fix_sachin_fed_544.md
+++ b/.changesets/fix_fix_sachin_fed_544.md
@@ -1,0 +1,5 @@
+### Fix query planning error where `@requires` subgraph jump fetches `@key` from the wrong subgraph ([PR #8016](https://github.com/apollographql/router/pull/8016))
+
+During query planning, a subgraph jump added due to a `@requires` field may sometimes try to collect the necessary `@key` fields from an upstream subgraph fetch as an optimization, but it wasn't properly checking whether that subgraph had those fields. This is now fixed, and previously could cause query planning errors with messages that look like "InternalRebaseError(CannotRebase { field_position: Object(T.id), parent_type: Object(T) })".
+
+By [@sachindshinde](https://github.com/sachindshinde) in https://github.com/apollographql/router/pull/8016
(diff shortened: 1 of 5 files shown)


.changesets/fix_fix_sachin_fed_544.md
```diff
@@ -0,0 +1,5 @@
+### Fix query planning error where `@requires` subgraph jump fetches `@key` from the wrong subgraph ([PR #8016](https://github.com/apollographql/router/pull/8016))
+
+During query planning, a subgraph jump added due to a `@requires` field may sometimes try to collect the necessary `@key` fields from an upstream subgraph fetch as an optimization, but it wasn't properly checking whether that subgraph had those fields. This is now fixed, and previously could cause query planning errors with messages that look like "InternalRebaseError(CannotRebase { field_position: Object(T.id), parent_type: Object(T) })".
```
duckki: The quote here is the debug output of the error object. But, the actual printed symptom (error message) was the same as JS version.

duckki (approved):
Looks good! Just a nit in the changelog text.

.changesets/fix_fix_sachin_fed_544.md
sachindshinde: Good catch, updated the message to be `Display` for `CannotRebase`.

duckki:
can't approve more! let's merge.

PR #8018, opened 2025-07-30
feat(redis): use replica client if in cluster

aaronArinder:
---

**Checklist**

- Tests added and passing[^4]
    - [ ] Unit tests
    - [ ] Integration tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/cache/redis.rs
+++ b/apollo-router/src/cache/redis.rs
@@ -75,6 +75,7 @@ fn record_redis_error(error: &RedisError, caller: &'static str) {
         RedisErrorKind::Sentinel => "sentinel",
         RedisErrorKind::NotFound => "not_found",
         RedisErrorKind::Backpressure => "backpressure",
+        RedisErrorKind::Replica => "replica",
     };
 
     u64_counter_with_unit!(
@@ -332,6 +333,14 @@ impl RedisCacheStorage {
             .build_pool(pool_size)?;
 
         for client in pooled_client.clients() {
+            // Use a client for the replicas if in clustered mode, otherwise we'll only connect to
+            // the writer nodes in the cluster
+            let client = if is_cluster {
+                &client.replicas().client()
+            } else {
+                client
+            };
+
             // spawn tasks that listen for connection close or reconnect events
             let mut error_rx = client.error_rx();
             let mut reconnect_rx = client.reconnect_rx();
--- a/apollo-router/Cargo.toml
+++ b/apollo-router/Cargo.toml
@@ -98,6 +98,7 @@ fred = { version = "10.1.0", features = [
     "i-cluster",
     "tcp-user-timeouts",
     "metrics",
+    "replicas"
 ] }
 futures = { version = "0.3.30", features = ["thread-pool"] }
 graphql_client = "0.14.0"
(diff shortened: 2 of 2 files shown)


apollo-router/src/cache/redis.rs
```diff
@@ -332,6 +333,14 @@ impl RedisCacheStorage {
             .build_pool(pool_size)?;
 
         for client in pooled_client.clients() {
+            // Use a client for the replicas if in clustered mode, otherwise we'll only connect to
+            // the writer nodes in the cluster
+            let client = if is_cluster {
+                &client.replicas().client()
```
goto-bus-stop: Note `.client()` returns the original non-replica client again, so it doesn't change anything this way.

We need to use `.replicas()` wherever we execute Redis commands. I _think_ it's fine to do `.replicas()` even if there are no replicas, but we might want to validate that :)

aaronArinder: oh, yeah, you're right:

```
  /// Read the underlying [RedisClient](crate::clients::Client) that interacts with primary nodes.
  pub fn client(&self) -> Client {
    Client::from(self.client.inner())
  }
```

apollo-router/src/cache/redis.rs
```diff
@@ -620,7 +638,7 @@ impl RedisCacheStorage {
             res.sort_by(|(i, _), (j, _)| i.cmp(j));
             Some(res.into_iter().map(|(_, v)| v).collect())
         } else {
-            self.inner
+            replica_client
                 .mget(
```
carodewig: Is there any chance of issues with using `self.inner.replicas()` in when redis runs in a non-clustered mode? I wonder if this function should be split into `if is_cluster { let client = self.inner.replicas(); .. } else { let client = self.inner.next(); .. }`

apollo-router/src/cache/redis.rs
```diff
@@ -526,7 +541,8 @@ impl RedisCacheStorage {
     ) -> Option<RedisValue<V>> {
         match self.ttl {
             Some(ttl) if self.reset_ttl => {
-                let pipeline: fred::clients::Pipeline<RedisClient> = self.inner.next().pipeline();
+                let pipeline = self.inner.replicas().pipeline();
+
```
carodewig: I think this branch resets the TTL of the key, so needs to be on a write node rather than a replica.

apollo-router/src/cache/redis.rs
aaronArinder: will look into it!

apollo-router/src/cache/redis.rs
aaronArinder: ack; will check it out (I spend very broken attention on this, but am keeping a list of stuff to check out and will add this to it)

apollo-router/tests/integration/redis.rs
```diff
@@ -1542,3 +1567,132 @@ async fn test_redis_connections_are_closed_on_router_reload() {
...
+            // maybe we need to trigger a config error in a different way; this test might actually
+            // test what happens with a config fred doesn't know how to use (eg, the schema isn't
+            // redis, rediss-cluster, etc)
+           r#"apollo_router_cache_redis_errors_total{error_type="unknown",kind="entity",otel_scope_name="apollo/router"}"#,
+            None,
+        )
+        .await;
+
+    router.graceful_shutdown().await;
+}
```
aaronArinder, line 1698: this is where I left off; I want to test any gaps in metrics/errors, but was running into `unknown` when trying to trigger a `configuration` error

PR #8025, opened 2025-07-31
Port `mergeDescription` to merger.rs

conwuegb:
Straight port of [mergeDescription(...)](https://github.com/apollographql/federation/blob/31104c6fc5319fb8140d697776249e9ee0ca9285/composition-js/src/merging/merge.ts#L920-L972), with additional support for reading and updating description field of schema types.

---

**Checklist**

- Tests added and passing[^4]
    - [x] Unit tests
    - [ ] Integration tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-federation/src/merger/merger.rs
+++ b/apollo-federation/src/merger/merger.rs
@@ -44,6 +44,7 @@ use crate::schema::FederationSchema;
 use crate::schema::directive_location::DirectiveLocationExt;
 use crate::schema::position::DirectiveDefinitionPosition;
 use crate::schema::position::DirectiveTargetPosition;
+use crate::schema::position::HasDescriptionPosition;
 use crate::schema::position::InterfaceTypeDefinitionPosition;
 use crate::schema::position::TypeDefinitionPosition;
 use crate::schema::referencer::DirectiveReferencers;
@@ -1186,8 +1187,79 @@ impl Merger {
         Ok(source_type.clone())
     }
 
-    pub(in crate::merger) fn merge_description<T>(&mut self, _sources: &Sources<T>, _dest: &T) {
-        todo!("Implement merge_description")
+    pub(in crate::merger) fn merge_description<T>(&mut self, sources: &Sources<T>, dest: &T)
+    where
+        T: HasDescriptionPosition + std::fmt::Display,
+    {
+        let mut descriptions: Vec<String> = Vec::new();
+        let mut counts: Vec<isize> = Vec::new();
+
+        for source in sources.values() {
+            if source.is_none() || source.as_ref().unwrap().description(&self.merged).is_none() {
+                continue;
+            }
+
+            let source_desc = source
+                .as_ref()
+                .unwrap()
+                .description(&self.merged)
+                .unwrap()
+                .as_str();
+            let idx = descriptions.iter().position(|desc| desc == source_desc);
+            if let Some(idx) = idx {
+                counts[idx] += 1;
+            } else {
+                // Very much a hack but simple enough: while we do merge 'empty-string' description if that's all we have (debatable behavior in the first place,
+                // but graphQL-js does print such description and fed 1 has historically merged them so ...), we really don't want to favor those if we
+                // have any non-empty description, even if we have more empty ones across subgraphs. So we use a super-negative base count if the description
+                // is empty so that our `indexOfMax` below never pick them if there is a choice.
+                if source_desc.is_empty() {
+                    counts.push(isize::MIN);
+                } else {
+                    counts.push(1);
+                }
+                descriptions.push(source_desc.to_string());
+            }
+        }
+
+        if !descriptions.is_empty() {
+            // we don't want to raise a hint if a description is ""
+            let non_empty_descriptions = descriptions
+                .iter()
+                .filter(|desc| !desc.is_empty())
+                .collect_vec();
+            if descriptions.len() == 1 {
+                dest.set_description(&mut self.merged, Some(Node::new_str(&descriptions[0])));
+            } else if non_empty_descriptions.len() == 1 {
+                dest.set_description(
+                    &mut self.merged,
+                    Some(Node::new_str(non_empty_descriptions[0])),
+                );
+            } else {
+                let idx = counts
+                    .into_iter()
+                    .enumerate()
+                    .max_by_key(|(_, count)| *count)
+                    .unwrap() // we know descriptions, and thus counts, isn't empty
+                    .0;
+                dest.set_description(&mut self.merged, Some(Node::new_str(&descriptions[idx])));
+                // TODO: Currently showing full descriptions in the hint messages, which is probably fine in some cases. However
+                // this might get less helpful if the description appears to differ by a very small amount (a space, a single character typo)
+                // and even more so the bigger the description is, and we could improve the experience here. For instance, we could
+                // print the supergraph description but then show other descriptions as diffs from that (using, say, https://www.npmjs.com/package/diff).
+                // And we could even switch between diff/non-diff modes based on the levenshtein distances between the description we found.
+                // That said, we should decide if we want to bother here: maybe we can leave it to studio so handle a better experience (as
+                // it can more UX wise).
+                self.error_reporter.report_mismatch_hint::<T, ()>(
+                    HintCode::InconsistentDescription,
+                    format!("{} has inconsistent descriptions across subgraphs. ", dest),
+                    dest,
+                    sources,
+                    |elt, _| elt.description(&self.merged).map(|desc| desc.to_string()),
+                    false,
+                );
+            }
+        }
     }
 
     pub(in crate::merger) fn add_join_field<T>(&mut self, _sources: &Sources<T>, _dest: &T) {
--- a/apollo-federation/src/schema/position.rs
+++ b/apollo-federation/src/schema/position.rs
@@ -152,6 +152,96 @@ macro_rules! infallible_conversions {
     }
 }
 
+/// Makes `description` field API available for use with generic types
+pub(crate) trait HasDescriptionPosition {
+    fn description<'schema>(&self, schema: &'schema FederationSchema)
+    -> Option<&'schema Node<str>>;
+    fn set_description(&self, schema: &mut FederationSchema, description: Option<Node<str>>);
+}
+
+macro_rules! impl_has_description_position_for {
+    ($struct_name:ident) => {
+        impl HasDescriptionPosition for $struct_name {
+            fn description<'schema>(
+                &self,
+                schema: &'schema FederationSchema,
+            ) -> Option<&'schema Node<str>> {
+                self.get(&schema.schema).unwrap().description.as_ref()
+            }
+
+            fn set_description(
+                &self,
+                schema: &mut FederationSchema,
+                description: Option<Node<str>>,
+            ) {
+                self.make_mut(&mut schema.schema)
+                    .unwrap()
+                    .make_mut()
+                    .description = description;
+            }
+        }
+    };
+}
+
+impl_has_description_position_for!(DirectiveDefinitionPosition);
+impl_has_description_position_for!(ScalarTypeDefinitionPosition);
+impl_has_description_position_for!(ObjectTypeDefinitionPosition);
+impl_has_description_position_for!(InterfaceTypeDefinitionPosition);
+impl_has_description_position_for!(UnionTypeDefinitionPosition);
+impl_has_description_position_for!(EnumTypeDefinitionPosition);
+impl_has_description_position_for!(InputObjectTypeDefinitionPosition);
+impl_has_description_position_for!(ObjectFieldDefinitionPosition);
+impl_has_description_position_for!(InterfaceFieldDefinitionPosition);
+impl_has_description_position_for!(EnumValueDefinitionPosition);
+
+// Irregular implementations of HasDescriptionPosition
+impl HasDescriptionPosition for SchemaDefinitionPosition {
+    fn description<'schema>(
+        &self,
+        schema: &'schema FederationSchema,
+    ) -> Option<&'schema Node<str>> {
+        self.get(&schema.schema).description.as_ref()
+    }
+
+    fn set_description(&self, schema: &mut FederationSchema, description: Option<Node<str>>) {
+        self.make_mut(&mut schema.schema).make_mut().description = description;
+    }
+}
+
+impl HasDescriptionPosition for FieldDefinitionPosition {
+    fn description<'schema>(
+        &self,
+        schema: &'schema FederationSchema,
+    ) -> Option<&'schema Node<str>> {
+        match self {
+            FieldDefinitionPosition::Object(field) => field.description(schema),
+            FieldDefinitionPosition::Interface(field) => field.description(schema),
+            FieldDefinitionPosition::Union(field) => field.description(schema),
+        }
+    }
+
+    fn set_description(&self, schema: &mut FederationSchema, description: Option<Node<str>>) {
+        match self {
+            FieldDefinitionPosition::Object(field) => field.set_description(schema, description),
+            FieldDefinitionPosition::Interface(field) => field.set_description(schema, description),
+            FieldDefinitionPosition::Union(field) => field.set_description(schema, description),
+        }
+    }
+}
+
+impl HasDescriptionPosition for UnionTypenameFieldDefinitionPosition {
+    fn description<'schema>(
+        &self,
+        schema: &'schema FederationSchema,
+    ) -> Option<&'schema Node<str>> {
+        self.get(&schema.schema).unwrap().description.as_ref()
+    }
+
+    fn set_description(&self, _schema: &mut FederationSchema, _description: Option<Node<str>>) {
+        // Description is immutable for union typename fields
+    }
+}
+
 #[derive(Clone, PartialEq, Eq, Hash, derive_more::From, derive_more::Display)]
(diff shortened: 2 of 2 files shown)


apollo-federation/src/merger/merger.rs
```diff
@@ -1186,8 +1187,79 @@ impl Merger {
...
+        let mut counts: Vec<isize> = Vec::new();
+
+        for source in sources.values() {
+            if source.is_none() || source.as_ref().unwrap().description(&self.merged).is_none() {
+                continue;
+            }
+
+            let source_desc = source
+                .as_ref()
+                .unwrap()
```
tninesling: These calls to `.unwrap()` are ok because they come right after checking if `.is_none()`, but you do have to ask the question "will this actually panic?" whenever you come across them. Might be better to just use a let binding like
```
let Some(source_desc) = source.and_then(|s| s.description(&schema)) else {
    continue;
}
```

apollo-federation/src/merger/merger.rs
```diff
@@ -1186,8 +1187,79 @@ impl Merger {
...
+                .description(&self.merged)
```
tninesling: We're merging into `&self.merged`, but think we need to read from the subgraph here to get the actual message. That will require using the `idx` from `sources` and using `self.subgraphs[idx]`

apollo-federation/src/merger/merger.rs
```diff
@@ -1186,8 +1187,79 @@ impl Merger {
...
+                .unwrap()
+                .as_str();
+            let idx = descriptions.iter().position(|desc| desc == source_desc);
```
tninesling: This is a solid port, but it is also one of those cases where the JS way is a terrible way to do things. What we really want is a `HashMap<&str, usize>` which counts the unique messages. If you extract all the descriptions into an iterator, `itertools` has a nice [counts](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.counts) utility you can use.

apollo-federation/src/merger/merger.rs
```diff
@@ -1186,8 +1187,79 @@ impl Merger {
...
+                }
+                descriptions.push(source_desc.to_string());
+            }
+        }
+
+        if !descriptions.is_empty() {
+            // we don't want to raise a hint if a description is ""
+            let non_empty_descriptions = descriptions
+                .iter()
+                .filter(|desc| !desc.is_empty())
```
tninesling: If we put these in a `HashMap`, we can just `descriptions.remove("")` before entering this loop instead of doing all the `MIN_SAFE_INTEGER` stuff.

apollo-federation/src/merger/merger.rs
```diff
@@ -1186,8 +1187,79 @@ impl Merger {
...
+                // And we could even switch between diff/non-diff modes based on the levenshtein distances between the description we found.
+                // That said, we should decide if we want to bother here: maybe we can leave it to studio so handle a better experience (as
+                // it can more UX wise).
+                self.error_reporter.report_mismatch_hint::<T, ()>(
+                    HintCode::InconsistentDescription,
+                    format!("{} has inconsistent descriptions across subgraphs. ", dest),
+                    dest,
+                    sources,
+                    |elt, _| elt.description(&self.merged).map(|desc| desc.to_string()),
+                    false,
```
tninesling, line 1318: Looks like we need to add a couple more parameters to `report_mismatch_hint`. Happy to pair with you on that if you're not sure on it

apollo-federation/src/schema/position.rs
```diff
@@ -152,6 +152,96 @@ macro_rules! infallible_conversions {
...
 }
 
+/// Makes `description` field API available for use with generic types
+pub(crate) trait HasDescriptionPosition {
+    fn description<'schema>(&self, schema: &'schema FederationSchema)
+    -> Option<&'schema Node<str>>;
+    fn set_description(&self, schema: &mut FederationSchema, description: Option<Node<str>>);
+}
+
+macro_rules! impl_has_description_position_for {
```
tninesling: Ooh, nice macro! 🔥

apollo-federation/src/schema/position.rs
```diff
@@ -152,6 +152,96 @@ macro_rules! infallible_conversions {
...
+    ($struct_name:ident) => {
+        impl HasDescriptionPosition for $struct_name {
+            fn description<'schema>(
+                &self,
+                schema: &'schema FederationSchema,
+            ) -> Option<&'schema Node<str>> {
+                self.get(&schema.schema).unwrap().description.as_ref()
```
tninesling: Instead of `.unwrap()`, it would be good to use `.ok()` to convert the `Result` to an `Option`. Otherwise, we'd have to change the signature to `Result<Option<...>>`, but plenty of our APIs return `None` if we get an error on these lookups.

apollo-federation/src/schema/position.rs
```diff
@@ -152,6 +152,96 @@ macro_rules! infallible_conversions {
...
+            }
+
+            fn set_description(
+                &self,
+                schema: &mut FederationSchema,
+                description: Option<Node<str>>,
+            ) {
+                self.make_mut(&mut schema.schema)
+                    .unwrap()
```
tninesling: Can we remove this unwrap somehow? We shouldn't panic if we can't find the element in the schema. Probably return a `Result`, or at least use a let binding

apollo-federation/src/schema/position.rs
duckki: Or, `self.try_get(...).and_then(...)`. I think `Option` makes sense since description is optional.

apollo-federation/src/schema/position.rs
duckki: +1 Lets' return `Result<(), FederationError>`.

apollo-federation/src/schema/position.rs
```diff
@@ -152,6 +152,96 @@ macro_rules! infallible_conversions {
     }
 }
 
+/// Makes `description` field API available for use with generic types
+pub(crate) trait HasDescriptionPosition {
```
duckki: How about `HasDescription`? It's short and it's already in `position` module. So...

apollo-federation/src/schema/position.rs
```diff
@@ -152,6 +152,96 @@ macro_rules! infallible_conversions {
...
+impl HasDescriptionPosition for UnionTypenameFieldDefinitionPosition {
+    fn description<'schema>(
+        &self,
+        schema: &'schema FederationSchema,
+    ) -> Option<&'schema Node<str>> {
+        self.get(&schema.schema).unwrap().description.as_ref()
+    }
+
+    fn set_description(&self, _schema: &mut FederationSchema, _description: Option<Node<str>>) {
+        // Description is immutable for union typename fields
```
duckki: Yup. GraphQL does not allow unions to have any fields. But, we add synthetic `__typename` field definition internally. So, they shouldn't have descriptions. Let's `bail!` if this is ever called.

apollo-federation/src/merger/merger.rs
duckki: +1

apollo-federation/src/merger/merger.rs
```diff
@@ -1186,8 +1187,79 @@ impl Merger {
...
+                .as_ref()
+                .unwrap()
+                .description(&self.merged)
+                .unwrap()
+                .as_str();
+            let idx = descriptions.iter().position(|desc| desc == source_desc);
+            if let Some(idx) = idx {
+                counts[idx] += 1;
+            } else {
+                // Very much a hack but simple enough: while we do merge 'empty-string' description if that's all we have (debatable behavior in the first place,
```
duckki: I like to re-wrap these long texts. I personally use Rewrap vscode extension (https://marketplace.visualstudio.com/items?itemName=stkb.rewrap).

apollo-federation/src/merger/merger.rs
```diff
@@ -1186,8 +1187,79 @@ impl Merger {
...
+            }
+        }
+
+        if !descriptions.is_empty() {
+            // we don't want to raise a hint if a description is ""
+            let non_empty_descriptions = descriptions
+                .iter()
+                .filter(|desc| !desc.is_empty())
+                .collect_vec();
+            if descriptions.len() == 1 {
```
duckki: Just a suggestion: I'd like to use `iter_into_single_item` in this case.
```suggestion
            if let Some(only_description) = iter_into_single_item(descriptions.iter()) {
```

apollo-federation/src/merger/merger.rs
```diff
@@ -1186,8 +1187,79 @@ impl Merger {
...
+                dest.set_description(
+                    &mut self.merged,
+                    Some(Node::new_str(non_empty_descriptions[0])),
+                );
+            } else {
+                let idx = counts
+                    .into_iter()
+                    .enumerate()
+                    .max_by_key(|(_, count)| *count)
+                    .unwrap() // we know descriptions, and thus counts, isn't empty
```
duckki: We would better just skip `set_description`, instead of `unwrap` here. Just not to take any chance of panic, for now and future. Or, we could `bail!` and return a `Result`.

apollo-federation/src/merger/merger.rs
```diff
@@ -1186,8 +1187,79 @@ impl Merger {
...
+                // TODO: Currently showing full descriptions in the hint messages, which is probably fine in some cases. However
+                // this might get less helpful if the description appears to differ by a very small amount (a space, a single character typo)
+                // and even more so the bigger the description is, and we could improve the experience here. For instance, we could
+                // print the supergraph description but then show other descriptions as diffs from that (using, say, https://www.npmjs.com/package/diff).
+                // And we could even switch between diff/non-diff modes based on the levenshtein distances between the description we found.
+                // That said, we should decide if we want to bother here: maybe we can leave it to studio so handle a better experience (as
+                // it can more UX wise).
+                self.error_reporter.report_mismatch_hint::<T, ()>(
+                    HintCode::InconsistentDescription,
+                    format!("{} has inconsistent descriptions across subgraphs. ", dest),
```
duckki: `dest`'s Display impl would be different from JS code below:
```typescript
dest instanceof NamedSchemaElement ? `Element "${dest.coordinate}"` : 'The schema definition';
```

apollo-federation/src/schema/position.rs
conwuegb: Done

apollo-federation/src/schema/position.rs
conwuegb: Done

apollo-federation/src/schema/position.rs
conwuegb: Done

apollo-federation/src/merger/merger.rs
conwuegb: Done

apollo-federation/src/merger/merger.rs
conwuegb: So the only problem here is that I can't figure out what `NamedSchemaElement` maps to on the Rust side. There's a [SchemaElementWithType](https://github.com/apollographql/router/blob/2539887372e88dc83d32af1a3b51f7c1c443a9b2/apollo-federation/src/merger/merger.rs#L134) trait that seems loosely connected, since it has the `coordinate()` function, but the implementation is different enough that they could also refer to distinct things.

apollo-federation/src/merger/merger.rs
conwuegb: Is `iter_into_single_item()` a predefined function somewhere or would I be creating it?

apollo-federation/src/merger/merger.rs
conwuegb: Done -- converted it to a `CountsMap`

apollo-federation/src/merger/merger.rs
conwuegb: Done

apollo-federation/src/merger/merger.rs
conwuegb: Fixed, please take a look.

apollo-federation/src/merger/merger.rs
conwuegb: Fixed, please take a look

apollo-federation/src/merger/merger.rs
conwuegb: Done

apollo-federation/src/merger/merger.rs
conwuegb: Done

apollo-federation/src/merger/merger.rs
duckki: It's under the `utils` directory.

apollo-federation/src/merger/merger.rs
duckki: `impl Display` of a position is equivalent to `.coordinate()` method. But, SchemaDefinition's Display will print nothing (or something unexpected). Porting [this test](https://github.com/apollographql/federation/blob/e17173bf9e7b3fdee42a9ee0ac4bd269de67e374/composition-js/src/__tests__/hints.test.ts#L733) will confirm it.
Since we can't instance-type check in this function, we can simply call .`to_string()` and see if it's empty or not.  Something like

```rust
let coordinate = dest.to_string();
if !coordinate.is_empty() { format!("Element {coordinate}") } else { "The schema definition" }
```

apollo-federation/src/merger/merger.rs
```diff
@@ -1005,6 +1016,17 @@ impl Merger {
...
+                |elt, subgraphs| {
+                    format!(
+                        "will use type \"{}\" (from {}) in supergraph but \"{}\" has ",
+                        elt,
+                        subgraphs.unwrap_or_else(|| "undefined".to_string()),
+                        dest.coordinate(parent_type_name)
+                    )
+                },
+                |elt, subgraphs| format!("{} \"{}\" in {}", type_class, elt, subgraphs),
+                Some(|elt: Option<&Type>| elt.is_none()),
```
duckki: `ignorePredicate` is not passed in JS code. Is there a reason for passing `Some` here?

apollo-federation/src/merger/merger.rs
```diff
@@ -817,15 +819,19 @@ impl Merger {
...
                     &directive_sources,
                     |elt, _| if elt.arguments.is_empty() {
                         Some("no arguments".to_string())
                     } else {
                         Some(format!("arguments: [{}]", elt.arguments.iter().map(|arg| format!("{}: {}", arg.name, arg.value)).join(", ")))
                     },
-                    false
+                    |application, subgraphs| format!("The supergraph will use {} (from {}), but found ", application, subgraphs.unwrap_or_else(|| "undefined".to_string())),
+                    |application, subgraphs| format!("{} in {}", application, subgraphs),
+                    Some(|elt: Option<&Directive>| elt.is_none()),
```
duckki: Same here: JS code passes nothing for `ignore_predicate`, Rust passes `Some`. Please explain why we have a difference.

apollo-federation/src/schema/position.rs
```diff
@@ -152,6 +152,111 @@ macro_rules! infallible_conversions {
...
+        }
+    }
+}
+
+impl HasDescription for UnionTypenameFieldDefinitionPosition {
+    fn description<'schema>(
+        &self,
+        schema: &'schema FederationSchema,
+    ) -> Option<&'schema Node<str>> {
+        self.get(&schema.schema).unwrap().description.as_ref()
```
duckki:
```suggestion
        self.get(&schema.schema).is_ok_and(|field| field.description.as_ref())
```

apollo-federation/src/merger/merger.rs
conwuegb: Done

apollo-federation/src/merger/merger.rs
conwuegb: Essentially, I was copying what was hardcoded in the initial implementation of [report_mismatch_hint()](https://github.com/apollographql/router/blob/a46dabafb036588d48fa16d7dd0845fbbc33b46c/apollo-federation/src/merger/error_reporter.rs#L136), but you're right -- I should be passing None instead.

apollo-federation/src/merger/merger.rs
conwuegb: Same as above: "Essentially, I was copying what was hardcoded in the initial implementation of [report_mismatch_hint()](https://github.com/apollographql/router/blob/a46dabafb036588d48fa16d7dd0845fbbc33b46c/apollo-federation/src/merger/error_reporter.rs#L136), but you're right -- I should be passing None instead."

apollo-federation/src/schema/position.rs
conwuegb: I used `map_or()` instead because it seemed like a better fit.

apollo-federation/src/merger/merger.rs
```diff
@@ -1208,8 +1231,104 @@ impl Merger {
...
+            };
+
+            descriptions.insert_or_increment(source_desc.trim().to_string());
+        }
+        descriptions.remove(&String::new());
+
+        if !descriptions.is_empty() {
+            // we don't want to raise a hint if a description is ""
+            if descriptions.len() == 1 {
+                if let Some((description, _)) = iter_into_single_item(descriptions.iter()) {
```
duckki: This condition can replace `if descriptions.len() == 1`, since the two conditions are identical.

duckki (approved):
Looks good. I have one last suggestion, which is minor.
I'll approve it, so it can be merged right away.

PR #8027, opened 2025-07-31
[Connectors] Add isSuccess logic to connector HTTP error handling

Hunter-apollo:
Adds the core error handling functionality for the `isSuccess` parameter. 

The `isSuccess` parameter allows users to set custom success condition that overrides the default error condition (i.e. checking that status is 2XX). This allows users to get data from an error response (for example a 404).

Note that deserialization and rate limiting errors are not mapped by this logic and will always be marked as internal connector failures.

---
**Checklist**

- Tests added and passing[^4]
    - [X] Unit tests
    - [ ] Integration tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-federation/src/connectors/runtime/responses.rs
```diff
@@ -122,50 +105,70 @@ pub fn map_response(
...
-    let (res, apply_to_errors) = key.selection().apply_with_vars(data, &inputs);
+// If the user has set a custom success condition selector, resolve that expression,
+// otherwise default to checking status code is 2XX
+fn is_success(
+    connector: &Connector,
+    data: &Value,
+    parts: &Parts,
+    inputs: &IndexMap<String, Value>,
+    warnings: &mut Vec<Problem>,
+) -> bool {
```
dylan-apollo: I'd prefer we return the value instead of allocating right before and passing in a mut ref. As much as I dislike tuples, I think making the info flow more explicit is worthwhile, and returning (ResultType, Warnings) is really common in this code base.


```suggestion
) -> (bool, Vec<Warning>) {
```

Hunter-apollo: Fine with me. I dont mind the mut reference, but happy to make it more consistent with other parts of the codebase. will update

andrewmcgivery (approved):

Issue #8039, opened 2025-08-02
Including the custom attributes with the telemetry in apollo router is not supported.


riginoommen:
## Is your feature request related to a problem? Please describe.
We want to include custom attributes like userId, tenantId, or JWT claims in Apollo Router’s telemetry. Right now, there’s no built-in way to do this, which limits how much we can trace or debug requests.

## Describe the solution you'd like
Allow a way to add custom attributes to spans, logs, and metrics—maybe through RHAI Scripts or the router configs

In the official documentation of the router this limitation is not documented and this one is something very much critical.

https://github.com/apollographql/router/blob/dev/docs/source/routing/customization/rhai/index.mdx#limitations

PR #8040, opened 2025-08-02
docs(router): clarify Rhai limitations on custom span creation and trace modification

riginoommen:
Added a fourth point to the list of Rhai scripting limitations in the Apollo Router
documentation. This update highlights that Rhai cannot create custom spans or 
modify the trace structure beyond the current execution context, such as injecting 
JWT claim values into trace spans.

This clarification is based on recent findings during JWT trace instrumentation
efforts and aims to set accurate expectations for teams implementing custom 
telemetry logic using Rhai.

Issue - #8039

---

**Checklist**

- [x] PR description links appropriate GitHub/Jira tickets #8039 
- Tests added and passing[^4]
    - [x] Unit tests
    - [x] Integration tests

**Exceptions**

*Note any exceptions here*

**Notes**


docs/source/routing/customization/rhai/index.mdx
```diff
@@ -468,6 +468,7 @@ Currently, Rhai scripts _cannot_ do the following:
 * Use Rust crates
 * Execute network requests
 * Read or write to disk
+* Create custom spans or modify trace structure beyond the current execution context (e.g., injecting JWT claim data into spans)
 
 If your router customization needs to do any of these, you can instead use [external co-processing](/router/customizations/coprocessor/) (this is an [Enterprise feature](/router/enterprise-features)).
```
abernix: This list is *really* far from exhaustive.  For example, you cannot add TLS certificates or configure CORS with Rhai, either — along with dozens and dozens of other things. 😄  I don't think we want to exhaustively list everything you cannot do.

Adding this bullet-point here invalidates the sentence after it because it suggests that you _can_ do what you're suggesting with coprocessors, which you cannot.  Injecting context into spans, whether it is JWT or otherwise, can _only_ be done with Router YAML configuration (though you could set _context_ in Rhai which could be then attached to spans [using `request_context`](https://www.apollographql.com/docs/graphos/routing/observability/telemetry/instrumentation/selectors#selector-configuration-reference:~:text=the%20request%20header-,request_context,-Yes) or its sister `response_context`)

We might want to just say, "Using Rhai can be quite limiting and is best used for simple modification of headers, context and payloads.  For advanced cases, including making network requests, or writing to disk, consider using coprocessors or appropriate YAML configuration".

abernix (dismissed):
Hey, huge thanks for your help in trying to improve our docs and avoid indirection — if this is a follow-up to being headed in the wrong direction, I think it's worth the time to get that right.

However, we need to do something different than this though, because the change as is results in added confusion (in my opinion).

docs/source/routing/customization/rhai/index.mdx
riginoommen: I've been migrating from Gateway to Router over the last quarter, and I’ve found it to be a tough learning experience — especially working with something like Rhai, which isn’t widely supported or documented upstream.

If you search for Rhai on Google or even in AI systems, there’s very little useful context or real-world examples available. The limitations aren’t clearly documented either, and I believe it's high time this is addressed.

As an enterprise user of the Router, I’ve seen firsthand how these gaps can slow down adoption and implementation. Clearer documentation on Rhai's limitations — and guidance on when to use YAML or coprocessors instead

abernix: I understand that, and I empathize with it.  I'm not suggesting we don't fix the documentation here, but if we add the line that you're suggesting, then the line directly below it becomes invalid — you cannot "Create custom spans or modify trace structure" with coprocessors.

abernix: What do you think about the concrete suggestion I have above to change the text to make it clear that there are limitations and to actually saying what Rhai is useful for, rather than attempting to make a list of things it **can't do**.  (The list of things it can't do is very long, as I pointed out.)

I think it's important to say that you should do as little as possible with Rhai and instead have a preference for using the YAML configuration, which is where you _can_ do what you are trying to do.

riginoommen: addressed and updated the PR as you said like

docs/source/routing/customization/rhai/index.mdx
```diff
@@ -463,13 +463,9 @@ Whenever you make changes to your scripts, check your router's log output to mak
 
 ## Limitations
 
-Currently, Rhai scripts _cannot_ do the following:
+Using Rhai can be quite limiting and is best suited for simple modifications — such as altering headers, modifying context values, or lightweight payload transformations.
```
mabuyo:
```suggestion
Rhai customization is best suited for simple modifications such as altering headers, modifying context values, or lightweight payload transformations.
```
Simplifying!

docs/source/routing/customization/rhai/index.mdx
```diff
@@ -463,13 +463,9 @@ Whenever you make changes to your scripts, check your router's log output to mak
...
 
-* Use Rust crates
-* Execute network requests
-* Read or write to disk
-
-If your router customization needs to do any of these, you can instead use [external co-processing](/router/customizations/coprocessor/) (this is an [Enterprise feature](/router/enterprise-features)).
+For more advanced functionality — including network requests, writing to disk, using RUST Crates injecting custom span data, or accessing external data — prefer using YAML configuration or [external co-processing](/router/customizations/coprocessor/) (an [Enterprise feature](/router/enterprise-features)).
```
mabuyo:
```suggestion
For more advanced functionality — including network requests, writing to disk, using RUST Crates injecting custom span data, or accessing external data — use YAML configuration or [external co-processing](/router/customizations/coprocessor/).
```
Removing the "enterprise feature" since this is no longer accurate the the linked docs page has more details on what Apollo plans support it.

docs/source/routing/customization/rhai/index.mdx
```diff
@@ -463,13 +463,9 @@ Whenever you make changes to your scripts, check your router's log output to mak
...
 
-Currently, Rhai scripts _cannot_ do the following:
+Rhai customization is best suited for simple modifications such as altering headers, modifying context values, or lightweight payload transformations.
 
-* Use Rust crates
-* Execute network requests
-* Read or write to disk
-
-If your router customization needs to do any of these, you can instead use [external co-processing](/router/customizations/coprocessor/) (this is an [Enterprise feature](/router/enterprise-features)).
+For more advanced functionality — including network requests, writing to disk, using RUST Crates injecting custom span data, or accessing external data — use YAML configuration or [external co-processing](/router/customizations/coprocessor/).
```
mabuyo:
```suggestion
For more advanced functionality — including network requests, writing to disk, using Rust crates injecting custom span data, or accessing external data — use YAML configuration or [external co-processing](/router/customizations/coprocessor/).
```
Use lowercase "Rust"

PR #8046, opened 2025-08-05
Enable jemalloc on MacOS

Velfi:
This PR enables the jemalloc allocator on MacOS by default. Previously, this was only done for Linux. We're making this change because it will make memory profiling easier than it currently is.

---

**Checklist**

- Tests added and passing[^4]
    - [ ] Unit tests
    - [ ] Integration tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-1409]: https://apollographql.atlassian.net/browse/ROUTER-1409?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BrynCooke (changes requested):
We still need to consider windows. Please can you verify that either jemalloc will work on windows or reintroduce the feature gate so that we only cover the right platforms.

Edit: I see you specified `unix`. Ignore this.

PR #8048, opened 2025-08-05
Port merge InputObject js composition over to Rust

briannafugate408:
This PR aims to port [merge InputObject](https://github.com/apollographql/federation/blob/main/composition-js/src/merging/merge.ts#L2644) logic from js composition over to Rust. 

---

**Checklist**

- Tests added and passing[^4]
    - [x] Unit tests
    - [ ] Integration tests

**Exceptions**

*Note any exceptions here*

**Notes**

[FED-407]: https://apollographql.atlassian.net/browse/FED-407?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apollo-router/tests/integration/response_cache.rs
```diff
@@ -461,6 +461,7 @@ async fn cache_control_merging_single_fetch() {
 }
 
 #[tokio::test]
+#[ignore = "flaky"]
```
TylerBloom: If you haven't already, I would ask in `#proj-router` if this change in ok. I assume it is, but best to check with the code owners.

briannafugate408: This actually might be able to be removed now. Going to check and see if the pipeline still fails

apollo-federation/src/merger/merge_input.rs
```diff
@@ -0,0 +1,605 @@
...
+                    fields_set.insert(Some(field.clone()));
+                }
+            }
+
+            // Check if this subgraph defines the input type (even if source is None)
+            if let Some(subgraph) = self.subgraphs.get(*source_index) {
+                if subgraph.schema().get_type(dest.type_name.clone()).is_ok() {
+                    extra_sources.insert(*source_index, None);
+                }
+            } else {
```
dariuszkuc: this seems like a change from JS behavior? (i.e. (in JS we only populate `extraSources` IF schema defines the type)[https://github.com/apollographql/federation/blob/main/composition-js/src/merging/merge.ts#L1290-L1298]))

apollo-federation/src/merger/merge_input.rs
```diff
@@ -0,0 +1,605 @@
...
+                }
+                // Note that we remove the element after the hint/error because we access the parent in the hint message.
+                dest_field.remove(&mut self.merged)?;
+            }
+        }
+
+        // We could be left with an input type with no fields, and that's invalid in GraphQL
+        let final_input_object = dest.get(self.merged.schema())?;
+        if final_input_object.fields.is_empty() {
+            self.error_reporter.add_error(CompositionError::InternalError {
```
dariuszkuc: I believe this should be a specific `EMPTY_MERGED_INPUT_TYPE` composition error that should include location information (`nodes` in JS)

apollo-federation/src/merger/merge_input.rs
```diff
@@ -0,0 +1,605 @@
...
+                if let Some(inaccessible_directive) =
+                    &self.inaccessible_directive_name_in_supergraph
+                {
+                    dest_field_def.directives.has(inaccessible_directive)
+                } else {
+                    false
+                }
+            } else {
+                false
+            };
```
dariuszkuc: I think we can simplify it to something like this

```suggestion
            let is_inaccessible = dest_field_def.is_some_and(|dest_field_def| {
                matches!(&self.inaccessible_directive_name_in_supergraph, Some(inaccessible_directive) if dest_field_def.directives.has(&inaccessible_directive))
            });
```

apollo-federation/src/merger/merge_input.rs
```diff
@@ -0,0 +1,605 @@
...
+                        .map(|s| s.as_str())
+                        .collect::<Vec<_>>()
+                        .join(", ");
+                    let missing_subgraphs_str = missing_subgraphs
+                        .iter()
+                        .map(|s| s.as_str())
+                        .collect::<Vec<_>>()
+                        .join(", ");
+
+                    self.error_reporter.add_error(CompositionError::InternalError {
```
dariuszkuc: this should be `REQUIRED_INPUT_FIELD_MISSING_IN_SOME_SUBGRAPH` composition error with error location info

apollo-federation/src/merger/merge_input.rs
```diff
@@ -0,0 +1,605 @@
...
+    #[allow(dead_code)]
+    pub(crate) fn merge_input(
+        &mut self,
+        sources: &Sources<Node<InputObjectType>>,
+        dest: &InputObjectTypeDefinitionPosition,
+    ) -> Result<(), FederationError> {
+        let added = self.add_fields_shallow(sources, dest);
+
+        for (dest_field, subgraph_fields) in added {
+            if let Some(field_def) = subgraph_fields.values().find_map(|f| f.as_ref()) {
```
dariuszkuc: am I missing something or we actually don't merge the input fields? it looks like we are only adding the field here (which was done in `addFieldShallow` in [JS](https://github.com/apollographql/federation/blob/main/composition-js/src/merging/merge.ts#L1305))

Is `mergeInputField` ([link](https://github.com/apollographql/federation/blob/main/composition-js/src/merging/merge.ts#L2699-L2706)) (merging descriptions, directives, default values, etc) already handled in some other part of the code base?

dariuszkuc (changes requested):
Please add the missing comments from the JS codebase

apollo-federation/src/merger/merge_input.rs
TylerBloom: IMO, function chaining is a bit cleaner here:
```rust
            let is_inaccessible = self
                .inaccessible_directive_name_in_supergraph
                .as_ref()
                .zip(dest_field.try_get(self.merged.schema()))
                .is_some_and(|(inaccessible_directive, dest_field_def)| {
                    dest_field_def.directives.has(inaccessible_directive)
                });
```

apollo-federation/src/merger/merge_input.rs
```diff
@@ -0,0 +1,605 @@
...
+            // it won't be exposed in the API, and we don't hint about it because we're just doing what the user is explicitly asking.
+            if !is_inaccessible
+                && Self::some_sources(&subgraph_fields, |field, _idx| field.is_none())
+            {
+                // If the field is optional, we remove it for the supergraph and issue a hint.
+                let mut non_optional_subgraphs = Vec::new();
+                let mut missing_subgraphs = Vec::new();
+
+                for (idx, field) in subgraph_fields.iter() {
+                    let subgraph_name = &self.names[*idx];
```
TylerBloom, line 64: This introduces a subtle panic should `self.names` is shorter than `idx`. This should be an internal error rather than a panic.

apollo-federation/src/merger/merge_input.rs
```diff
@@ -0,0 +1,605 @@
...
+                        _ => {}
+                    }
+                }
+
+                if !non_optional_subgraphs.is_empty() {
+                    let non_optional_subgraphs_str = non_optional_subgraphs
+                        .iter()
+                        .map(|s| s.as_str())
+                        .collect::<Vec<_>>()
+                        .join(", ");
```
TylerBloom: You don't need to allocate a new vector. `Itertools` provides a methods for this called `join`. Alternatively, you can skip allocating a new string altogether and use `Itertools::format`. Both of these would work
```rust
// Option A
let non_optional_subgraphs_str = non_optional_subgraphs.into_iter().join(", ");
// Option B
let non_optional_subgraphs_str = non_optional_subgraphs.into_iter().format(", ");
```

apollo-federation/src/merger/merge_input.rs
```diff
@@ -0,0 +1,605 @@
...
+                    ),
+                });
+                } else {
+                    let mut present_subgraphs = Vec::new();
+                    let mut locations = Vec::new();
+
+                    // Extract nodes and create locations for fields that exist
+                    for (idx, field) in subgraph_fields.iter() {
+                        if let Some(field_component) = field {
+                            present_subgraphs.push(&self.names[*idx]);
```
TylerBloom: There is a similar indexing issue here that could lead to panics.

apollo-federation/src/merger/merge_input.rs
```diff
@@ -0,0 +1,605 @@
...
+                                locations.extend(field_locations);
+                            }
+                        }
+                    }
+
+                    let present_subgraphs_str = present_subgraphs
+                        .iter()
+                        .map(|s| s.as_str())
+                        .collect::<Vec<_>>()
+                        .join(", ");
```
TylerBloom: Same `Vec` -> `String` -> `format!` pattern here.

apollo-router/tests/integration/response_cache.rs
duckki: For apollo-federation work, I just skip building/checking `apollo-router`. Here's my quick check script:
```
        cargo build -p apollo-federation
        cargo build -p apollo-federation-cli
        cargo test -p apollo-federation --all-features
        cargo test -p apollo-federation-cli
        cargo clippy -p apollo-federation ${CLIPPY_OPTS}
        cargo clippy -p apollo-federation-cli ${CLIPPY_OPTS}
        cargo fmt -p apollo-federation -- --check ${FMT_CONFIGS}
        cargo fmt -p apollo-federation-cli -- --check ${FMT_CONFIGS}
```

apollo-federation/src/merger/merge_input.rs
```diff
@@ -0,0 +1,685 @@
...
+        if final_input_object.fields.is_empty() {
+            let mut locations = Vec::new();
+            for (idx, input_type) in sources.iter() {
+                if let Some(input_component) = input_type {
+                    if let Some(subgraph) = self.subgraphs.get(*idx) {
+                        let type_locations =
+                            subgraph
+                                .schema()
+                                .node_locations(input_component)
+                                .map(|loc| SubgraphLocation {
```
duckki: Can we just use `Merger::source_locations` method here? Or, we can refactor it to be more widely applicable.

apollo-federation/src/merger/merger.rs
```diff
@@ -1208,6 +1208,15 @@ impl Merger {
         todo!("Implement add_arguments_shallow")
     }
 
+    pub(in crate::merger) fn merge_default_value<T>(
+        &mut self,
+        _sources: &Sources<T>,
+        _dest: &T,
+        kind: &str,
+    ) {
+        todo!("Implement merge_default_value")
```
dariuszkuc, line 1841: do we have an existing story/task to follow up on this?

apollo-federation/src/merger/merge_input.rs
```diff
@@ -0,0 +1,669 @@
...
+            Sources<Component<InputValueDefinition>>,
+        > = IndexMap::default();
+        let mut fields_to_add: IndexMap<usize, HashSet<Option<Component<InputValueDefinition>>>> =
+            IndexMap::default();
+        let mut extra_sources: Sources<Component<InputValueDefinition>> = IndexMap::default();
+
+        for (source_index, source) in sources.iter() {
+            let fields_set = fields_to_add.entry(*source_index).or_default();
+
+            // If a source is undefined, it may still have an @interfaceObject object
```
dariuszkuc: this comment should be removed as it is only applicable to output object type logic

apollo-federation/src/merger/merge_input.rs
```diff
@@ -0,0 +1,669 @@
...
+            if let Some(source_input) = source {
+                for field in source_input.fields.values() {
+                    fields_set.insert(Some(field.clone()));
+                }
+            }
+
+            if let Some(subgraph) = self.subgraphs.get(*source_index) {
+                if subgraph.schema().get_type(dest.type_name.clone()).is_ok() {
+                    // This marks the subgraph as having a relevant @interfaceObject,
+                    // even though we do not actively add the itfType.fields().
```
dariuszkuc: wrong comment? ([link](https://github.com/apollographql/federation/blob/main/composition-js/src/merging/merge.ts#L1291-L1297))

apollo-federation/src/merger/merge_input.rs
```diff
@@ -0,0 +1,669 @@
...
+            for field_opt in field_set.into_iter().flatten() {
+                let dest_field_pos = dest.field(field_opt.name.clone());
+
+                // Our needsJoinField logic adds @join__field if any subgraphs define
+                // the parent type containing the field but not the field itself. In
+                // those cases, for each field we add, we need to add undefined entries
+                // for each subgraph that defines the parent object/interface/input
+                // type. We do this by populating extraSources with undefined entries
+                // here, then create each new Sources map from that starting set (see
+                // `new Map(extraSources)` below).
```
dariuszkuc: wrong comment? (this should be added at line 187 above and there is no comment here)

apollo-federation/src/merger/merge_input.rs
```diff
@@ -0,0 +1,666 @@
...
+        added
+    }
+
+    fn merge_input_field(
+        &mut self,
+        dest_field: &InputObjectFieldDefinitionPosition,
+        sources: &Sources<Component<InputValueDefinition>>,
+    ) -> Result<(), FederationError> {
+        // NOTE: Component<InputValueDefinition> does not implement HasDescription or Display,
+        // so we cannot call merge_description here. Instead, we only record applied directives.
```
dariuszkuc: according to [graphql spec](https://spec.graphql.org/draft/#InputValueDefinition) it certainly can have a description and we probably should fix this

apollo-federation/src/merger/merge_input.rs
```diff
@@ -0,0 +1,668 @@
...
+        if let Some(dest_component) = sources.values().find_map(|c| c.as_ref()) {
+            self.record_applied_directives_to_merge(sources, dest_component);
+        }
+
```
dariuszkuc: this logic seems weird to me - we are iterating over sources to find target dest component then to pass it along together with sources? since method is not implemented it _**might**_ be ok but I'm guessing we need merge those into an input object in a target supergraph schema (i.e. `dest_field.get(self.merged.schema())` instead of picking one at random.

apollo-federation/src/merger/merge_input.rs
```diff
@@ -0,0 +1,668 @@
...
+        self.add_join_field(
+            &directive_sources,
+            &dest_directive,
+            all_types_equal,
+            &merge_context,
+        )?;
+        if let Some(dest_component) = sources.values().find_map(|c| c.as_ref()) {
+            self.merge_default_value(sources, dest_component, "Input field");
+        }
+
```
dariuszkuc: same comment as above we should be passing the component from supergraph schema

apollo-federation/src/merger/merge_input.rs
```diff
@@ -0,0 +1,668 @@
...
+        dest_field: &InputObjectFieldDefinitionPosition,
+        sources: &Sources<Component<InputValueDefinition>>,
+    ) -> Result<(), FederationError> {
+        // NOTE: Component<InputValueDefinition> does not implement HasDescription or Display,
+        // so we cannot call merge_description here. Instead, we only record applied directives.
+        if let Some(dest_component) = sources.values().find_map(|c| c.as_ref()) {
+            self.record_applied_directives_to_merge(sources, dest_component);
+        }
+
+        let type_sources: Sources<Type> = sources
```
dariuszkuc, line 186: why are we cloning sources instead of passing the inputs directly to the `merge_type_reference`?

briannafugate408: I can't call `merge_type_reference` inside the map because it needs the complete type_sources and returns a boolean, not a Type.

apollo-federation/src/error/mod.rs
```diff
@@ -195,12 +197,21 @@ pub enum CompositionError {
         message: String,
         locations: Locations,
     },
-    #[error(
-        "[{subgraph}] Type \"{dest}\" is an extension type, but there is no type definition for \"{dest}\" in any subgraph."
-    )]
-    ExtensionWithNoBase {
```
tninesling: Please restore this definition for `ExtensionWithNoBase`. This one includes the locations which we're incrementally adding to these error types.

apollo-router/tests/integration/response_cache.rs
tninesling: Can we remove this? I know the tests are flaky, but we can effectively ignore failures in apollo-router since our changes don't affect that part of the repo.

apollo-federation/src/merger/merge_input.rs
```diff
@@ -0,0 +1,688 @@
...
+            // compatibility between definitions and 2) we actually want to see if the result is marked inaccessible or not and it makes
+            // that easier.
+            if let Err(e) = self.merge_input_field(&dest_field, &subgraph_fields) {
+                self.error_reporter
+                    .add_error(CompositionError::InputFieldMergeFailed {
+                        message: format!(
+                            "Failed to merge input field \"{}.{}\": {}",
+                            dest_field.type_name, dest_field.field_name, e
+                        ),
+                        locations: vec![],
```
tninesling: Do we have a node here that we can use to get the locations?

apollo-federation/src/merger/merge_input.rs
```diff
@@ -0,0 +1,688 @@
...
+                    });
+                continue;
+            }
+            if let Some(field_def) = subgraph_fields.values().find_map(|f| f.as_ref())
+                && dest_field.try_get(self.merged.schema()).is_none()
+            {
+                dest_field.insert(&mut self.merged, field_def.clone())?;
```
tninesling: Is this redundant after `self.add_fields_shallow`? I don't see an equivalent to this in JS

apollo-federation/src/merger/merge_input.rs
```diff
@@ -0,0 +1,688 @@
...
+            }
+
+            let dest_field_def = dest_field.try_get(self.merged.schema());
+
+            let is_inaccessible = dest_field_def.is_some_and(|dest_field_def| {
```
tninesling: Instead of getting the node from the graph and then checking directives, we would usually do `dest_field.has_applied_directive(&self.merged, self.inaccessible_directive_name_in_supergraph)`

apollo-federation/src/merger/merge_input.rs
```diff
@@ -0,0 +1,688 @@
...
+                && Self::some_sources(&subgraph_fields, |field, _idx| field.is_none())
+            {
+                // One of the subgraph has the input type but not that field. If the field is optional, we remove it for the supergraph
+                // and issue a hint. But if it is required, we have to error out.
+                let mut non_optional_subgraphs: Vec<String> = Vec::new();
+                let mut missing_subgraphs: Vec<String> = Vec::new();
+
+                for (idx, field) in subgraph_fields.iter() {
+                    let Some(subgraph_name) = &self.names.get(*idx) else {
+                        return Err(internal_error!("Subgraph name not found"));
```
tninesling: nit: this is effectively what `bail!` does

apollo-federation/src/merger/merge_input.rs
```diff
@@ -0,0 +1,688 @@
...
+                        }
+                        _ => {}
+                    }
+                }
+
+                if !non_optional_subgraphs.is_empty() {
+                    let non_optional_subgraphs_str = non_optional_subgraphs.into_iter().join(",");
+                    let missing_subgraphs_str = missing_subgraphs.into_iter().join(",");
+
+                    // create new error types
```
tninesling: Is this a TODO comment?

apollo-federation/src/merger/merge_input.rs
```diff
@@ -0,0 +1,688 @@
...
+                                    });
+                                locations.extend(field_locations);
+                            }
+                        }
+                    }
+
+                    let present_subgraphs_str = present_subgraphs.into_iter().join(", ");
+                    let missing_subgraphs_str = missing_subgraphs.into_iter().join(", ");
+
+                    self.error_reporter.add_hint(CompositionHint {
```
tninesling: This is supposed to use `self.error_reporter.report_mismatch_hint` to do the string formatting. Is there anything missing from that API that's blocking this usage?

apollo-federation/src/merger/merge_input.rs
```diff
@@ -0,0 +1,688 @@
...
+                    dest.type_name
+                ),
+                locations,
+            });
+        }
+
+        Ok(())
+    }
+
+    fn add_fields_shallow(
```
tninesling: This implementation isn't quite a match of the original. Can you update the signature/implementation to include output fields as well? Some of the comments are off as well

apollo-federation/src/merger/merge_input.rs
```diff
@@ -0,0 +1,688 @@
...
+                        source_opt
+                            .as_ref()
+                            .map(|_| DirectiveTargetPosition::InputObjectField(dest_field.clone())),
+                    )
+                })
+                .collect();
+
+            self.merge_description(
+                &field_sources,
+                &DirectiveTargetPosition::InputObjectField(dest_field.clone()),
```
tninesling: Can you use `InputObjectFieldDefinitionPosition` directly here (and in the rest of these)? The whole point of uses that sources structure is that we can pass the same stuff through each of these helper methods, so this implementation feels very verbose to me.

apollo-federation/src/error/mod.rs
duckki: +1

apollo-federation/src/error/mod.rs
```diff
@@ -232,14 +243,24 @@ impl CompositionError {
                 ErrorCode::MaxValidationSubgraphPathsExceeded
             }
             Self::InternalError { .. } => ErrorCode::Internal,
+            Self::ExtensionWithNoBase { .. } => ErrorCode::ExtensionWithNoBase,
```
duckki: We seem to have a duplicate for this variant?

apollo-federation/src/merger/merge_input.rs
dariuszkuc: IMHO it might be easier to just have separate `add_input_fields_shallow` (rename this one) and `add_output_fields_shallow` methods

apollo-federation/src/merger/merge_input.rs
```diff
@@ -0,0 +1,688 @@
...
+                }
+                // Note that we remove the element after the hint/error because we access the parent in the hint message.
+                dest_field.remove(&mut self.merged)?;
+            }
+        }
+
+        // We could be left with an input type with no fields, and that's invalid in GraphQL
+        let final_input_object = dest.get(self.merged.schema())?;
+        if final_input_object.fields.is_empty() {
+            let locations = Merger::source_locations(self, sources);
```
duckki: Why not `self.source_locations(sources)`?

apollo-federation/src/merger/merge_input.rs
tninesling: I think it depends how hard it is to mimic the conditional types used in the TS version. If it's too tricky, feel free to mark this as `todo!()`, since we were planning to do this implementation in FED-549

apollo-federation/src/merger/merger.rs
briannafugate408: https://apollographql.atlassian.net/browse/FED-416

apollo-federation/src/error/mod.rs
```diff
@@ -239,7 +256,16 @@ impl CompositionError {
             }
             Self::LinkImportNameMismatch { .. } => ErrorCode::LinkImportNameMismatch,
             Self::InvalidFieldSharing { .. } => ErrorCode::InvalidFieldSharing,
+            Self::InconsistentInputObjectField { .. } => ErrorCode::Internal, // This is for hints, not errors
+            Self::RequiredInputFieldMissingInSomeSubgraph { .. } => {
+                ErrorCode::RequiredInputFieldMissingInSomeSubgraph
+            }
+            Self::EmptyMergedInputType { .. } => ErrorCode::EmptyMergedInputType,
+            Self::InputFieldMergeFailed { .. } => ErrorCode::InputFieldMergeFailed,
+            #[allow(unused)]
```
tninesling: Are these unused? I don't remember having to annotate anything in this block before

apollo-federation/src/error/mod.rs
```diff
@@ -312,21 +338,41 @@ impl CompositionError {
...
+            },
+            Self::InputFieldMergeFailed { message, locations } => Self::InputFieldMergeFailed {
+                message: format!("{message}{appendix}"),
+                locations,
+            },
             // Remaining errors do not have an obvious way to appending a message, so we just return self.
             Self::SubgraphError { .. }
             | Self::InvalidGraphQLName(..)
             | Self::FromContextParseError { .. }
-            | Self::UnsupportedSpreadDirective { .. }
```
tninesling: Is there a reason for this change?

apollo-federation/src/error/mod.rs
```diff
@@ -312,21 +338,41 @@ impl CompositionError {
...
+            | Self::UnsupportedSpreadDirective { .. } => self,
+            Self::ExtensionWithNoBase { .. } => self,
         }
     }
 
     pub fn locations(&self) -> &[SubgraphLocation] {
         match self {
-            Self::SubgraphError { locations, .. }
-            | Self::EmptyMergedEnumType { locations, .. }
-            | Self::ExtensionWithNoBase { locations, .. }
```
tninesling, line 328: Same question: is this a linter thing?

apollo-federation/src/merger/merge_input.rs
```diff
@@ -0,0 +1,227 @@
...
+            if !is_inaccessible
+                && Self::some_sources(&subgraph_fields, |field, _idx| field.is_none())
+            {
+                // One of the subgraph has the input type but not that field. If the field is optional, we remove it for the supergraph
+                // and issue a hint. But if it is required, we have to error out.
+                let mut non_optional_subgraphs: Vec<String> = Vec::new();
+                let mut missing_subgraphs: Vec<String> = Vec::new();
+
+                for (idx, field) in subgraph_fields.iter() {
+                    let Some(subgraph_name) = &self.names.get(*idx) else {
```
tninesling: I like the safety here, but if you want, I think it's ok to just index into the array like `self.names[*idx]` since we know these are all an index into our subgraph data

apollo-federation/src/merger/merge_input.rs
```diff
@@ -0,0 +1,227 @@
...
+                        }
+                        None => {
+                            missing_subgraphs.push(subgraph_name.to_string());
+                        }
+                        _ => {}
+                    }
+                }
+
+                if !non_optional_subgraphs.is_empty() {
+                    let non_optional_subgraphs_str = non_optional_subgraphs.into_iter().join(",");
```
tninesling: In JS, these use `printSubgraphNames`, so we should use the `human_readable_subgraph_names` utility here. It's slightly different than just joining with a comma.

apollo-federation/src/merger/merge_input.rs
```diff
@@ -0,0 +1,227 @@
...
+                            }
+                        }
+                    }
+                    // Create sources for the field components for the hint
+                    let field_sources: Sources<Node<InputValueDefinition>> = subgraph_fields
+                        .iter()
+                        .map(|(&idx, field_opt)| (idx, field_opt.as_ref().map(|f| f.node.clone())))
+                        .collect();
+
+                    self.report_mismatch_hint(
```
tninesling: This doesn't look like it matches the JS invocation (some of the formatters/messages are missing). Are we missing some parameters for this utility?

PR #8052, opened 2025-08-05
Implement merge logic for schema definitions and join directive utility

tninesling:
# Merging Schema Definitions

Implements the logic for merging schema definitions together, per the [JS source](https://github.com/apollographql/federation/blob/e17173bf9e7b3fdee42a9ee0ac4bd269de67e374/composition-js/src/merging/merge.ts#L3074). The logic is actually fairly simple, as it just merges any descriptions and delegates to a helper for merging any directives that need a `@join__directive`.

## Difference in validation

One important difference is that the original source performed a validation that root types were consistent across subgraphs. We rewrite those names in the link expansion step, so we should never reach a case where we get to merge without them matching. However, I think these types of validations should be pushed slightly earlier in the pipeline so we don't muddy the merge logic with unnecessary validation code. I've moved this particular validation to the `pre_merge_validations` phase, which performs validations requiring knowledge of all subgraphs.

## Implementation of `add_join_directive`

This utility is called in a few locations, so I've tried to make it as generic as possible. It is parameterized with a type `T`, which can be any position as long as that position is `Clone` and implements `TryInto<DirectiveTargetPosition>`. Once we convert to `DirectiveTargetPosition`, it uses the new `get_all_applied_directives` function added to each of the individual positions. Positions are cheap to clone, so this implementation should work well with our `From/Into` conversion setup.

---

**Checklist**

- Tests added and passing[^4]
    - [X] Unit tests
    - [ ] Integration tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-federation/src/link/join_spec_definition.rs
```diff
@@ -917,6 +917,43 @@ impl JoinSpecDefinition {
         ))
     }
 
+    /// Creates an instance of the `@join__directive` directive. Since we do not allow renaming of
+    /// join spec directives, this is infallible and always applies the directive with the standard
+    /// name.
+    pub(crate) fn directive_directive(
```
dariuszkuc, line 926: should this be named `join_directive` instead? I guess it is `directive` from `join` spec.... but naming is weird

tninesling: I think however we try to write `join__directive` directive, it's going to look weird. This follows the same naming convention of the rest of the functions in here, which is to omit the `join` prefix because we're already in the join spec. I added the doc comment here to make it clearer, but it's tough to get around the fact we created a directive called `@directive`...

apollo-federation/src/merger/merger.rs
```diff
@@ -1197,12 +1213,143 @@ impl Merger {
...
+        let Some(link_directive_name) = self
+            .link_spec_definition
+            .directive_name_in_schema(&self.merged, &DEFAULT_LINK_NAME)?
+        else {
+            bail!(
+                "Link directive must exist in the supergraph schema in order to apply join directives"
+            );
+        };
+
+        let mut latest_or_highest_link_by_identity: HashMap<Identity, (Url, Directive)> =
```
dariuszkuc, line 1305: can you add the comment from [merge.ts#L3214-L3223](https://github.com/apollographql/federation/blob/main/composition-js/src/merging/merge.ts#L3214-L3223) that explains why we do this?

tninesling: Do you know if that comment still applies? It's talking about using the latest known version, but there's nothing in this logic which does that. This part just applies the links with `@join__directive`, so maybe that block is referring to the logic in `validate_and_maybe_add_specs`?

I can add it in, it just feels misplaced here

tninesling: Added a slightly reworded version in https://github.com/apollographql/router/pull/8052/commits/8530c9e2878549e7456d3caa9fc645d276d20bf7

apollo-federation/src/link/join_spec_definition.rs
```diff
@@ -918,6 +918,43 @@ impl JoinSpecDefinition {
...
+    ) -> Directive {
+        Directive {
+            name: JOIN_DIRECTIVE_DIRECTIVE_NAME_IN_SPEC,
+            arguments: vec![
+                Node::new(Argument {
+                    name: JOIN_NAME_ARGUMENT_NAME,
+                    value: Node::new(Value::String(name.to_string())),
+                }),
+                Node::new(Argument {
+                    name: JOIN_GRAPH_ARGUMENT_NAME,
```
duckki: We don't seem to have a Rust constant for "graphs" argument name.

apollo-federation/src/link/join_spec_definition.rs
```diff
@@ -918,6 +918,43 @@ impl JoinSpecDefinition {
...
+                    value: Node::new(Value::List(
+                        graphs
+                            .into_iter()
+                            .map(|g| Node::new(Value::Enum(g)))
+                            .collect(),
+                    )),
+                }),
+                Node::new(Argument {
+                    name: JOIN_DIRECTIVE_ARGUMENTS_NAME_IN_SPEC,
```
duckki: `JOIN_DIRECTIVE_ARGUMENTS_NAME_IN_SPEC` is a type name, not an argument name. The argument name here is "args". Let's add a constant for that.

apollo-federation/src/schema/validators/root_fields.rs
```diff
@@ -0,0 +1,133 @@
+use apollo_compiler::ast::OperationType;
+
+use crate::error::CompositionError;
+use crate::subgraph::typestate::HasMetadata;
+use crate::subgraph::typestate::Subgraph;
+
+pub(crate) fn validate_consistent_root_fields<T: HasMetadata>(
```
duckki, line 11: Please port the comments from [JS code](https://github.com/apollographql/federation/blob/e17173bf9e7b3fdee42a9ee0ac4bd269de67e374/composition-js/src/merging/merge.ts#L3077).

apollo-federation/src/merger/merger.rs
```diff
@@ -1196,12 +1212,150 @@ impl Merger {
...
+    where
+        // If we implemented a `HasDirectives` trait for this bound, we could call that instead
+        // of cloning and converting to `DirectiveTargetPosition`.
+        T: Clone + TryInto<DirectiveTargetPosition>,
+        FederationError: From<<T as TryInto<DirectiveTargetPosition>>::Error>,
+    {
+        // Joins are grouped by directive name and arguments. So, a directive with the same
+        // arguments in multiple subgraphs is merged with a single `@join__directive` that
+        // specifies both graphs. If two applications have different arguments, each application
+        // gets its own `@join__directive` specifying the different arugments per graph.
```
duckki, line 1232: typo: arugments => arguments

duckki (approved):
I'll approve it preemptively assuming those corrections will be made. Thanks!

PR #8054, opened 2025-08-06
Add DockerHub References To The Documentation

jonathanrainer:
As per title, updates the documentation to add reference to our new DockerHub repo for the Runtime Container

### DIFF
--- a/docs/source/routing/self-hosted/index.mdx
+++ b/docs/source/routing/self-hosted/index.mdx
@@ -36,10 +36,10 @@ Follow our [Kubernetes quickstart](/graphos/routing/kubernetes/quickstart) to de
 
 Apollo provides the [Apollo Runtime Container](https://github.com/apollographql/apollo-runtime), which bundles all that's required to run the Apollo Runtime in one place. This includes the Apollo Router and the [Apollo MCP Server](/apollo-mcp-server).
 
-The images are downloadable from:
+You can download the images from:
 
-* [DockerHub](https://hub.docker.com/r/apollograph/apollo-runtime) (recommended)
-* [GitHub Container Registry](https://github.com/apollographql/apollo-runtime/pkgs/container/apollo-runtime)
+- [Docker Hub](https://hub.docker.com/r/apollograph/apollo-runtime) (recommended)
+- [GitHub Container Registry](https://github.com/apollographql/apollo-runtime/pkgs/container/apollo-runtime)
 
 For more information on deploying using your container environment:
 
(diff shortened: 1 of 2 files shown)


docs/source/routing/self-hosted/index.mdx
```diff
@@ -37,6 +37,7 @@ Follow our [Kubernetes quickstart](/graphos/routing/kubernetes/quickstart) to de
 Apollo provides the Apollo Runtime Container, which bundles all that's required to run the Apollo Runtime in one place. This includes the Apollo Router and the [Apollo MCP Server](/apollo-mcp-server).
 
 The images are available via GitHub, downloadable from `ghcr.io/apollographql/apollo-runtime` and the [Apollo Runtime Container repository](https://github.com/apollographql/apollo-runtime).
+Alternatively you can download the images from DockerHub at `apollograph/apollo-runtime`.
```
abernix: Is this meant to exist soon? https://hub.docker.com/r/apollograph/apollo-runtime?   (Can we link to that?, and also fix the other link.  I think it's a bit confusing to not link to the pages which have the full context of the versions, etc, particularly since we don't really want people using `latest`, I'd guess)

jonathanrainer: Yer that makes sense, I'm just getting that all spun up so will do when it's available then the docs can go out after the fact 👍

docs/source/routing/self-hosted/index.mdx
```diff
@@ -36,7 +36,8 @@ Follow our [Kubernetes quickstart](/graphos/routing/kubernetes/quickstart) to de
 
 Apollo provides the Apollo Runtime Container, which bundles all that's required to run the Apollo Runtime in one place. This includes the Apollo Router and the [Apollo MCP Server](/apollo-mcp-server).
 
-The images are available via GitHub, downloadable from `ghcr.io/apollographql/apollo-runtime` and the [Apollo Runtime Container repository](https://github.com/apollographql/apollo-runtime).
+The images are available via GitHub, downloadable from [`ghcr.io/apollographql/apollo-runtime`](https://github.com/apollographql/apollo-runtime/pkgs/container/apollo-runtime) and the [Apollo Runtime Container repository](https://github.com/apollographql/apollo-runtime).
```
mabuyo: @jonathanrainer can we simplify these two paragraphs into a list? Something like: 

> You can download the container image from any of the following locations:
> - GitHub Container Registry
> - Apollo Runtime Container repository
> - Docker Hub

If we recommend one over the other, we can also add `(recommended)` after it.

jonathanrainer: Thanks @mabuyo have done that, could you take a second look?

PR #8055, opened 2025-08-06
Always use `graph_os_enabled` helper for GraphOS tests

goto-bus-stop:
Cleanup.


apollo-router/tests/integration/telemetry/metrics.rs
```diff
@@ -70,15 +70,13 @@ async fn test_metrics_reloading() {
         .assert_metrics_does_not_contain(r#"_total_total{"#)
         .await;
 
-    if std::env::var("TEST_APOLLO_KEY").is_ok() && std::env::var("TEST_APOLLO_GRAPH_REF").is_ok() {
```
goto-bus-stop, line 73: The test already checks `graph_os_enabled()` above, so this is always true

PR #8056, opened 2025-08-06
Add support for sha256sum assertions on nightly Docker builds

abernix:
## Motivation
We recently encountered data corruption where a file in a Docker build had the correct size but wrong checksum, resulting in a broken Router image. This adds checksum validation to prevent such issues.

## Changes
Adds SHA256 validation for both nightly and release builds:

* Nightly builds: Validates local artifacts, remote artifacts in more than one way — inside CI, and within the Docker build
* Release builds: Downloads and validates against published sha256sums.txt from GitHub releases

Both paths fail with error messages if checksums don't match.

## Manual Testing

- [See this run](https://app.circleci.com/pipelines/github/apollographql/router/38179/workflows/b8012ab2-be5a-435c-8832-95b9baa2c020/jobs/286527/parallel-runs/0/steps/0-116)
- I'm also just starting another.
- I also manually tested building the Docker container (`Dockerfile.router`) with a bad `ARTIFACT_URL_SHA256` value passed as a build argument, which is how it's invoked from the CircleCI script and it correctly failed:
    ```
	docker build --progress=plain --platform linux/amd64 -t "jesse-new-router" \
	   --build-arg BASE_VERSION="0.0.0-nightly-2.5.0.20250806-6e56c771" \
	   --build-arg ROUTER_RELEASE="v0.0.0-nightly-2.5.0.20250806-6e56c771" \
	   --build-arg ARTIFACT_URL="https://github.com/apollographql/router/releases/download/v2.5.0/router-v2.5.0-x86_64-unknown-linux-gnu.tar.gz" \
	   --build-arg ARTIFACT_URL_SHA256SUM="🐑" \
	   --no-cache -f ../Dockerfile.router .
    ```
---

**Checklist**

- [x] ~Changeset is included for user-facing changes~
- [x] ~Metrics and logs are added[^3] and documented~
- Tests added and passing[^4]
    - [ ] ~Unit tests~
    - [ ] ~Integration tests~

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -718,12 +718,11 @@ jobs:
                   ARTIFACT_URL_SHA256SUM=$(sha256sum "router-artifact.tar.gz" | cut -d' ' -f1)
                   rm -f router-artifact.tar.gz
 
-                  # Additional safety net: Compare local vs remote artifact checksums
-                  REMOTE_ARTIFACT_SHA256SUM=$(curl -sSL -H "Circle-Token: ${CIRCLE_TOKEN}" "${ARTIFACT_URL}" | sha256sum | cut -d' ' -f1)
-                  if [ "${LOCAL_ARTIFACT_SHA256SUM}" != "${REMOTE_ARTIFACT_SHA256SUM}" ]; then
+                  # Compare local vs remote artifact checksums
+                  if [ "${LOCAL_ARTIFACT_SHA256SUM}" != "${ARTIFACT_URL_SHA256SUM}" ]; then
                     echo "Error: Local and remote artifact checksums don't match!"
-                    echo "Local:  ${LOCAL_ARTIFACT_SHA256SUM}"
-                    echo "Remote: ${REMOTE_ARTIFACT_SHA256SUM}"
+                    echo "Local:      ${LOCAL_ARTIFACT_SHA256SUM}"
+                    echo "Remote: ${ARTIFACT_URL_SHA256SUM}"
                     exit 1
                   fi
                   echo "Local and remote artifact checksums match: ${LOCAL_ARTIFACT_SHA256SUM}"


dockerfiles/Dockerfile.router
```diff
@@ -2,8 +2,9 @@ FROM debian:bookworm-slim AS downloader
 ARG ROUTER_RELEASE=latest
 ARG ARTIFACT_URL=
 ARG CIRCLE_TOKEN=
+ARG ARTIFACT_URL_SHA256SUM=
 
-# Install curl
+# Install curl and sha256sum
```
abernix: This comment was useless before, and I almost made it more useless by describing something I don't do anymore. :) 
```suggestion
```

dockerfiles/Dockerfile.router
```diff
@@ -2,22 +2,60 @@ FROM debian:bookworm-slim AS downloader
...
-WORKDIR /dist
+# We work in /, and the tarball produces /dist/router after extraction.
+WORKDIR /
 
 # Run the Router downloader which puts Router into current working directory
 RUN if [ -z "${ARTIFACT_URL}"]; then \
-    curl -sSL "https://router.apollo.dev/download/nix/${ROUTER_RELEASE}"/ | sh; \
+    # Download router tarball directly instead of using installer \
+    TARBALL_NAME="router-${ROUTER_RELEASE}-x86_64-unknown-linux-gnu.tar.gz"; \
+    curl -sSL "https://rover.apollo.dev/tar/router/x86_64-unknown-linux-gnu/${ROUTER_RELEASE}" -o "${TARBALL_NAME}"; \
```
abernix: I'll defend here why this URL change is reasonable/acceptable:

- There is no fixed Apollo controlled URL that lets you download a specific Router release **tarball**.
  - We currently only have the curl installer which downloads *and extracts* the tarballs.
  - This approach is acceptable and defensive since rover is guaranteed to have this in order for it to download the router, and that's not going away.  They also go through the same Orbiter endpoint.
  - It was possible to fix orbiter to also serve on the `router` domain and we could do that, but this seemed more than acceptable, and is a well-tested and monitored endpoint.
- The hard-coding of the `x86_64` bit (rather than letting the install script decide) also seems acceptable because this Docker container is built with `--platform linux/amd64` in the cirleci config (you can see in this changeset).

goto-bus-stop: I think this script is long enough that it could use its own separate file (and then that comment can be at the top of it :)).

abernix: Yeah, I wasn't sure if I should avoid making this file non-self containing or not, but I agree this is out of control with the trailing ` \`.

.circleci/config.yml
```diff
@@ -697,13 +697,43 @@ jobs:
...
+                    echo "Remote: ${REMOTE_ARTIFACT_SHA256SUM}"
+                    exit 1
+                  fi
+                  echo "Local and remote artifact checksums match: ${LOCAL_ARTIFACT_SHA256SUM}"
+
                   echo "REPO_URL: ${REPO_URL}"
                   echo "BASE_VERSION: ${BASE_VERSION}"
+                  echo "ARTIFACT_FILENAME: ${ARTIFACT_FILENAME}"
                   echo "ARTIFACT_URL: ${ARTIFACT_URL}"
+                  echo "ARTIFACT_URL_SHA256SUM: ${ARTIFACT_URL_SHA256SUM}"
```
lrlna, line 734: To triple check, we want the `ARTIFACT_URL_SHA256SUM` not the `REMOTE_ARTIFACT_SHA256SUM` here?

dockerfiles/Dockerfile.router
lrlna: and when this is a new file, maybe let's have a longer comment about why we need to validate checksums. We might forget about this reasoning 3 months from now.

dockerfiles/Dockerfile.router
```diff
@@ -2,22 +2,60 @@ FROM debian:bookworm-slim AS downloader
...
+      exit 1; \
+    fi; \
+    echo "Tarball checksum validation passed: ${ACTUAL_CHECKSUM}"; \
+    # Extract the tarball \
+    tar -xzf "${TARBALL_NAME}"; \
   else \
-    cd /; \
-    curl -sSL -H "Circle-Token: ${CIRCLE_TOKEN}" -o - "${ARTIFACT_URL}" | tar -xzf -; \
-    cd -; \
+    # Download artifact tarball \
```
lrlna: nit: bash scripts are not super readable as they are, can you add a new line before the comments to separate out the steps individual steps?

.circleci/config.yml
abernix: This is printing the literal values of the variables, so this is correct/intentional.

Are you questioning something different?

I don't think there's not a lot of value in printing `REMOTE_ARTIFACT_SHA256SUM` here because it's guaranteed to be the same as `LOCAL_ARTIFACT_SHA256SUM` below.

abernix: I could add another assertion that checks if `ARTIFACT_URL_SHA256SUM` equals `REMOTE_ARTIFACT_SHA256SUM` (in this script) prior to letting that happen in the `docker build` if that adds value?

dockerfiles/Dockerfile.router
```diff
@@ -2,22 +2,60 @@ FROM debian:bookworm-slim AS downloader
...
   apt-get update -y \
   && apt-get install -y \
     curl
 
-WORKDIR /dist
+# We work in /, and the tarball produces /dist/router after extraction.
+WORKDIR /
 
 # Run the Router downloader which puts Router into current working directory
 RUN if [ -z "${ARTIFACT_URL}"]; then \
```
BrynCooke:
```suggestion
RUN if [ -z "${ARTIFACT_URL}" ]; then \
```

Assuming bash I think that this missing space will cause issues:
```
bash-5.2$ foo=bif
bash-5.2$ [ -z "$foo"] && echo "$foo has zero length"
bash: [: missing `]'
bash-5.2$ [ -z "$foo" ] && echo "$foo has zero length"
bash-5.2$ unset foo
bash-5.2$ [ -z "$foo" ] && echo "$foo has zero length"
 has zero length
bash-5.2$ [ -z "$foo"] && echo "$foo has zero length"
 has zero length
```

I;'m not sure why we have not had this issue before because it's not part of this PR.

abernix: Heh.  Good eye.  I'll fix it, but it wouldn't affect this usage because the `if` invocation has different semantics.

dockerfiles/Dockerfile.router
abernix: See dc0050b90.

dockerfiles/Dockerfile.router
abernix: See dc0050b90.

dockerfiles/Dockerfile.router
abernix: See dc0050b90.

dockerfiles/Dockerfile.router
lrlna: my eyes thank you 🙏

.circleci/config.yml
```diff
@@ -697,13 +697,43 @@ jobs:
...
+
+                  # Additional safety net: Compare local vs remote artifact checksums
+                  REMOTE_ARTIFACT_SHA256SUM=$(curl -sSL -H "Circle-Token: ${CIRCLE_TOKEN}" "${ARTIFACT_URL}" | sha256sum | cut -d' ' -f1)
+                  if [ "${LOCAL_ARTIFACT_SHA256SUM}" != "${REMOTE_ARTIFACT_SHA256SUM}" ]; then
+                    echo "Error: Local and remote artifact checksums don't match!"
+                    echo "Local:  ${LOCAL_ARTIFACT_SHA256SUM}"
+                    echo "Remote: ${REMOTE_ARTIFACT_SHA256SUM}"
+                    exit 1
+                  fi
+                  echo "Local and remote artifact checksums match: ${LOCAL_ARTIFACT_SHA256SUM}"
```
abernix: Uhm, @lrlna's confusion made me pause and re-self-review.  This actually seems totally unnecessary duplicative (within the highlighted block) now that I read it.  I wonder if I should just keep this bottom part (because of the nice error) and get rid of the top part, and then pass `REMOTE_ARTIFACT_SHA256SUM` into the `docker build` as `ARTIFACT_URL_SHA256SUM`.

abernix: I convinced myself to remove it. 053106cb2b9593e6da334a912990147473878cba

PR #8057, opened 2025-08-06
Code clarifications in the WS subscriptions client protocols

goto-bus-stop:
This is a somewhat exploratory PR, containing code doc improvements
and small refactors I made while figuring out how subscriptions work.

- Adds doc comments, particularly clarifying protocol differences
- Use the `#[default]` attribute instead of manually implementing
  `Default`
- Prefix all message types unique to the legacy protocol with `Old`
- Use the same protocol order in all `match` statements
- Reduce nesting in test code and assert more error conditions

---

**Checklist**

- Tests added and passing[^4]
    - [ ] Unit tests
    - [ ] Integration tests

**Exceptions**

*Note any exceptions here*

**Notes**


apollo-router/src/services/execution/service.rs
```diff
@@ -70,26 +70,21 @@ pub(crate) struct ExecutionService {
...
 type CloseSignal = broadcast::Sender<()>;
-// Used to detect when the stream is dropped and then when the client closed the connection
-pub(crate) struct StreamWrapper(pub(crate) ReceiverStream<Response>, Option<CloseSignal>);
+/// Used to detect when the stream is dropped and then when the client closed the connection
+struct SubscriptionResponseStream {
+    stream: ReceiverStream<Response>,
+    /// When the response is dropped because of the client going away,
+    /// the broadcast handle is dropped. If all broadcast handles for
+    /// a deduplicated subscription go away, the associated subgraph
+    /// connection is closed.
```
goto-bus-stop: Ah this is actually wrong, this CloseSignal is not communicating over a shared broadcast channel, need to do this in a different place

apollo-router/tests/integration/subscriptions/ws_passthrough.rs
```diff
@@ -917,3 +937,159 @@ async fn test_subscription_ws_passthrough_dedup() -> Result<(), BoxError> {
...
+    // Create fixed payloads for consistent testing
+    let custom_payloads = vec![create_user_data_payload(1), create_user_data_payload(2)];
+    let interval_ms = 50;
+    let is_closed = Arc::new(AtomicBool::new(false));
+
+    // Start subscription server with fixed payloads, but do not terminate the connection
+    let (ws_addr, http_server) = start_subscription_server_with_payloads(
+        custom_payloads.clone(),
+        interval_ms,
+        false,
```
BrynCooke:
```suggestion
        true,
```
Otherwise the test will create more events than expected.

apollo-router/src/configuration/snapshots/apollo_router__configuration__tests__schema_generation.snap
```diff
@@ -11169,11 +11170,18 @@ expression: "&schema"
...
-      "type": "string"
+      "oneOf": [
+        {
+          "const": "graphql_ws",
+          "description": "The modern graphql-ws protocol. The subprotocol name is \"graphql-transport-ws\".\n\nSpec URL: https://github.com/enisdenjo/graphql-ws/blob/0c0eb499c3a0278c6d9cc799064f22c5d24d2f60/PROTOCOL.md",
+          "type": "string"
+        },
+        {
+          "const": "graphql_transport_ws",
+          "description": "The legacy subscriptions-transport-ws protocol. Confusingly, the subprotocol name is\n\"graphql-ws\".\n\nhttps://github.com/apollographql/subscriptions-transport-ws/blob/36f3f6f780acc1a458b768db13fd39c65e5e6518/PROTOCOL.md",
```
bnjjj, line 11181: Fucking nightmare...

apollo-router/src/protocols/websocket.rs
```diff
@@ -538,6 +554,8 @@ where
                         }
                         if let ServerMessage::Ping { .. } = server_message {
                             // Send pong asynchronously
+                            // XXX(@goto-bus-stop): We have to pull_flush() to ensure this thing
+                            // finishes, not sure if we're doing that right now?
```
bnjjj, line 561: No I don't think we need to flush because as described [here](https://docs.rs/futures/latest/futures/prelude/sink/trait.SinkExt.html#method.send) send also executes the flush

apollo-router/src/protocols/websocket.rs
```diff
@@ -627,6 +645,8 @@ where
     ) -> Poll<Result<(), Self::Error>> {
         let mut this = self.project();
         if !*this.completed {
+            // XXX(@goto-bus-stop): We have to pull_flush() to ensure this thing
+            // finishes, not sure if we're doing that right now?
```
bnjjj, line 652: Read https://docs.rs/futures/latest/futures/prelude/sink/trait.SinkExt.html#method.send it executes the flush in send

apollo-router/src/protocols/websocket.rs
```diff
@@ -643,8 +663,12 @@ where
         if let WebSocketProtocol::SubscriptionsTransportWs = this.protocol
             && !*this.terminated
         {
-            match Pin::new(&mut Pin::new(&mut this.stream).send(ClientMessage::ConnectionTerminate))
-                .poll(cx)
+            // XXX(@goto-bus-stop): We have to pull_flush() to ensure this thing
+            // finishes, not sure if we're doing that right now?
```
bnjjj, line 670: Same https://docs.rs/futures/latest/futures/prelude/sink/trait.SinkExt.html#method.send

apollo-router/src/protocols/websocket.rs
```diff
@@ -657,6 +681,8 @@ where
 
         if !*this.closed {
             // instead of just calling poll_close we also send a proper CloseWebsocket event to indicate it's a normal close, not an error
+            // XXX(@goto-bus-stop): We have to pull_flush() to ensure this thing
+            // finishes, not sure if we're doing that right now?
```
bnjjj, line 688: Same https://docs.rs/futures/latest/futures/prelude/sink/trait.SinkExt.html#method.send

apollo-router/src/protocols/websocket.rs
goto-bus-stop: It returns a _future_ that also executes the flush. But we only poll that future once, which isn't guaranteed (at least by docs) to drive the future to _completion_.

goto-bus-stop:
I think after this, the most important part that is still largely missing from tests is error handling, i.e. when the subgraph sends `error` type messages at various points. This should be another follow up.

apollo-router/src/services/subgraph_service.rs
```diff
@@ -635,7 +628,7 @@ async fn call_websocket(
         "http.url" = %uri,
         "net.transport" = "ip_tcp",
         "apollo.subgraph.name" = %service_name,
-        "graphql.operation.name" = %operation_name,
+        "graphql.operation.name" = body.operation_name.as_deref().unwrap_or(""),
```
BrynCooke, line 631: Nit: Suggest use Empty here: https://docs.rs/tracing/latest/tracing/field/struct.Empty.html otherwise users will see an empty attribute.

apollo-router/tests/integration/subscriptions/mod.rs
```diff
@@ -268,141 +279,387 @@ pub async fn verify_subscription_events(
...
+    State(config): State<SubscriptionServerConfig>,
     ws: WebSocketUpgrade,
     headers: axum::http::HeaderMap,
 ) -> Response {
     debug!("WebSocket upgrade requested");
     debug!("Headers: {:?}", headers);
-    ws.protocols(["graphql-ws"])
-        .on_upgrade(move |socket| handle_websocket(socket, config, is_closed))
+    // Speak both protocols
+    ws.protocols(["graphql-ws", "graphql-transport-ws"])
```
BrynCooke: Can we use the contants that have been defined elsewhere? `GRAPHQL_WS_SUBPROTOCOL`

apollo-router/src/protocols/websocket.rs
```diff
@@ -27,119 +29,127 @@ use crate::graphql;
 
 const CONNECTION_ACK_TIMEOUT: Duration = Duration::from_secs(5);
 
-#[derive(Debug, Clone, PartialEq, Eq, Hash, Deserialize, Serialize, JsonSchema, Copy)]
+/// The WebSocket subprotocol name for the modern graphql-ws protocol.
+/// See [`WebSocketProtocol::GraphqlWs`].
+const GRAPHQL_WS_SUBPROTOCOL: &str = "graphql-transport-ws";
+/// The WebSocket subprotocol name for the legacy subscriptions-transport-ws protocol.
+/// See [`WebSocketProtocol::SubscriptionsTransportWs`].
+const SUBSCRIPTIONS_TRANSPORT_WS_SUBPROTOCOL: &str = "graphql-ws";
```
BrynCooke, line 37: Nit: elsewhere we refer to these as modern and legacy. Maybe consider changing the names of the constants to this? OR change references from modern and legacy to match these constant names.

apollo-router/tests/integration/subscriptions/mod.rs
goto-bus-stop: I didn't use them here because this is an integration test and the existing constants are private to the crate. We could introduce constants inside this module additionally, there are two uses of each string I guess so it could make sense to do so

apollo-router/src/protocols/websocket.rs
goto-bus-stop: I guess I kinda changed my mind halfway while I was working on making things "more consistent" and ended up not making them consistent 😅 
I don't love the modern/legacy naming but the full names are very long, I'll have a think overnight what's better

bnjjj (approved):

PR #8059, opened 2025-08-06
Fix (and more thoroughly test) JSONSelection single-path parsing regression in connect/v0.3

benjamn:
Before we had the `...` spread syntax (merged recently in #7957 but not yet released), it was possible to spread multiple object properties into a larger object by using an anonymous path with a trailing `SubSelection` (formerly called a `PathWithSubSelection` in the grammar) as a `NamedSelection`, alongside other `NamedSelection` elements in a `SubSelection`, like so:
```graphql
id
created
model
# This line spreads role and content into the larger object beside
# id, created, and model.
choices->first.message { role content }
```
Now that we have the `...` syntax, this selection can (and probably should) be written
```graphql
id
created
model
... choices->first.message { role content }
```
For backwards compatibility, we need to preserve the old `PathWithSubSelection` syntax for now, though we probably want to encourage using an explicit `...`, because it makes the selection code much easier to read (IMO).

In addition to being easy to miss, the `PathWithSubSelection` syntax was involved in a regression that I introduced in #7957. Specifically, if the `PathWithSubSelection` was the only selection, as in
```graphql
choices->first.message { role content }
```
then it would be mistakenly interpreted as a spreadable `PathWithSubSelection`, rather than a single `PathSelection` (which was the interpretation in connect/v0.2). This may not seem like much of a difference, since "spreading" a single `PathWithSubSelection` basically means using its value as the whole output value, which is the same behavior as evaluating a single `PathSelection`.

However, when we're actually spreading a path's value, there are some additional restrictions, e.g. the path value must be an object or `null` (or `None` with an error), given we can't spread other values (including arrays). To preserve backwards compatibility (and fix the regression), then, we need to reinterpret a single `PathWithSubSelection` element as a `PathSelection`, so it is not required to evaluate to an object/`null` in order to be spread into a larger object, but can return whatever value it produces.

Where does this matter? Consider the following example:
```graphql
$.array { id name }
```
Before this PR, with `connect/v0.3` (unreleased), this looks like an anonymous path expression (due to the `$.`) with a trailing `SubSelection`, so it could potentially be parsed as a `PathWithSubSelection`. But assuming `$.array` evaluates to an array, "spreading" the result of this path means attempting to spread an array, which doesn't make sense in an object context. However, if we treat the selection as a single `PathSelection` that simply returns its value, then there's no problem returning an array as that value. That was the behavior in v0.2, so that's the behavior we need to restore.

In short, this PR restores the ability to use a single anonymous path with a trailing `SubSelection` as the only selection in a JSONSelection string, regardless of what value type it returns.

Note that explicitly using `...` will enforce the object/`null` requirement:
```graphql
# Fine because there's only one { role content } object to be spread:
... choices->first.message { role content }

# Runtime errors due to attempting to spread a non-object (array or other JSON type):
... choices.message { role content }
... $.array { id name }
... $("can't spread strings")->slice(0, 5)
... $([1, 2, 3])->first
```
However, any of these expressions should work as a single `PathSelection` without the `...`, as long as it's used by itself, with no other selections.

I would say this regression crept in because we were not actually running most of our JSONSelection tests with `ConnectSpec::V0_3`, since they were implicitly running with `ConnectSpec::latest()`, which is still `ConnectSpec::V0_2`. Thankfully, @andrewmcgivery noticed the problem, and sent me the reproduction that led to this PR. Without this fix, we would have seen many test failures upon attempting to update `ConnectSpec::latest()` to `ConnectSpec::V0_3`, which would have been surprising and bad. To ensure that doesn't happen, I've started converting more tests to use `#[rstest]` cases to run against both v0.2 and v0.3. These strengthened tests are all passing now, as expected.

### DIFF
--- a/apollo-federation/src/connectors/json_selection/apply_to.rs
+++ b/apollo-federation/src/connectors/json_selection/apply_to.rs
@@ -1094,8 +1094,10 @@ mod tests {
     use super::*;
     use crate::selection;
 
-    #[test]
-    fn test_apply_to_selection() {
+    #[rstest]
+    #[case::v0_2(ConnectSpec::V0_2)]
+    #[case::v0_3(ConnectSpec::V0_3)]
+    fn test_apply_to_selection(#[case] spec: ConnectSpec) {
         let data = json!({
             "hello": "world",
             "nested": {
@@ -1109,16 +1111,18 @@ mod tests {
             ],
         });
 
-        let check_ok = |selection: JSONSelection, expected_json: JSON| {
-            let (actual_json, errors) = selection.apply_to(&data);
+        #[track_caller]
+        fn check_ok(data: &JSON, selection: JSONSelection, expected_json: JSON) {
+            let (actual_json, errors) = selection.apply_to(data);
             assert_eq!(actual_json, Some(expected_json));
             assert_eq!(errors, vec![]);
-        };
+        }
 
-        check_ok(selection!("hello"), json!({"hello": "world"}));
+        check_ok(&data, selection!("hello", spec), json!({"hello": "world"}));
 
         check_ok(
-            selection!("nested"),
+            &data,
+            selection!("nested", spec),
             json!({
                 "nested": {
                     "hello": "world",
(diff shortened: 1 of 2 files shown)


andrewmcgivery (approved):
Approved pending CI!

PR #8067, opened 2025-08-08
Revert #7879 to fix a subscriptions regression in v2.5.0

goto-bus-stop:
This PR is _not_ intended to be merged, but just to cut a nightly release for select customers who require a quick fix.

## Problem

#7879 fixed a bug where deduplicated subscription connections to the subgraph were closed too early, leading to clients having a subscription open but not receiving any events.

Unfortunately, it introduced a regression the other way around: now when a client disconnects, the router doesn't close the subgraph connection, or at least, not cleanly. This can cause resource leakage in subgraphs.

## Workarounds

We're still working on a proper fix to the problem. In the mean time, there are some workarounds for affected users:

- Stay on 2.4.0. This means missing out on hot reload and deduplication fixes. You can disable subscription deduplication and hot reload. Both issues existed for a long time, so if you haven't experienced problems, you can also just wait it out.
- Update to the nightly released from this branch. This reverts a subscription deduplication fix, so the recommendation would be to disable subscription deduplication when using this branch.

Subscription deduplication is enabled by default. You can check if you're actually _using_ it heavily in production using the `subscriptions.deduplicated` attribute on the `apollo.router.operations.subscriptions` metric. If it's only `true` for a small or zero amount of subscriptions, you can safely disable subscription deduplication without increasing the load on subgraph services. If you _do_ have many deduplicated subscriptions, then consider if your subgraphs can handle increased load from disabling the feature. Otherwise you may have to wait it out until the next 2.5.x patch release.

### DIFF
--- a/apollo-router/src/services/subgraph_service.rs
+++ b/apollo-router/src/services/subgraph_service.rs
@@ -27,6 +27,7 @@ use opentelemetry::Key;
 use opentelemetry::KeyValue;
 use rustls::RootCertStore;
 use serde::Serialize;
+use tokio::select;
 use tokio::sync::oneshot;
 use tokio_tungstenite::connect_async;
 use tokio_tungstenite::connect_async_tls_with_config;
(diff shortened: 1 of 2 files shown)


abernix:
@goto-bus-stop should we close this now?

PR #8074, opened 2025-08-08
Benjamn/fix connect spec disregard in nontest code

benjamn:
This PR continues the process of auditing usages of `JSONSelection::parse` and (usually) discouraging its use in favor of `JSONSelection::parse_with_spec`, since `::parse(&str)` must guess the default `ConnectSpec` to use, whereas that information is specified explicitly with `::parse_with_spec(&str, ConnectSpec)`. Eventually, we may want to merge these methods together into one `::parse(&str, ConnectSpec)` method, but there are lot of remaining callers of `JSONSelection::parse` to update.

This PR uses the technique of temporarily configuring `JSONSelection::parse` as `#[cfg(test)]`, in order to find non-test code within the `apollo-federation` crate that was potentially ignoring the `ConnectSpec` version. Also I removed some helpers like the `impl TryFrom<Valid<FieldSet>> for JSONSelection`, since such patterns don't generalize to providing more information than the input value (e.g. the `ConnectSpec`). That said, the `TryFrom` implementation for `ErrorsArguments` was already using a tuple of inputs, so I decided to add `ConnectSpec` there as another tuple element.

You can already see from the snapshot updates that `connect/v0.n` versions chosen by the `@link` directive in various connectors test schemas are now respected more accurately (in many cases the schema actually does say `connect/v0.1`, so `spec: V0_1` in the snapshot is correct).


apollo-federation/src/connectors/runtime/responses.rs
```diff
@@ -348,16 +348,23 @@ impl MappedResponse {
...
                         ));
                     };
 
-                    let key_selection: Result<JSONSelection, _> = keys.try_into();
-                    let key_selection = key_selection
-                        .map_err(|e| HandleResponseError::MergeError(e.to_string()))?;
+                    let spec = selection.spec();
+                    let key_selection = JSONSelection::parse_with_spec(
+                        &keys.serialize().no_indent().to_string(),
+                        spec,
```
benjamn, line 365: Imagine how thrilled I was when I realized this `ResponseKey::BatchEntity` enum variant already had exactly the information I needed hidden behind that `..`: namely `selection: Arc<JSONSelection>`, which provides `JSONSelection::spec(&self) -> ConnectSpec`, returning whatever `ConnectSpec` was used to parse the original `JSONSelection`.

andrewmcgivery (approved):

PR #8081, opened 2025-08-11
Add graph artifact validation and aligned env variables

graytonio:
---
Adds a validation step for OCI references to only allow a digest referenced graph artifact to be configured. Will be updated when tag reloading is implemented.

https://apollographql.atlassian.net/browse/VUL-1360

https://apollographql.atlassian.net/browse/VUL-1350

**Checklist**

- Tests added and passing[^4]
    - [x] Unit tests
    - [x] Integration tests

**Exceptions**

*Note any exceptions here*

**Notes**

[ROUTER-1415]: https://apollographql.atlassian.net/browse/ROUTER-1415?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apollo-router/src/executable.rs
```diff
@@ -296,13 +295,27 @@ impl Opt {
...
+                    .clone()
+                    .ok_or(Self::err_require_opt("APOLLO_GRAPH_ARTIFACT_REFERENCE"))?,
+            )?,
         })
     }
 
+    fn validate_oci_reference(reference: &str) -> std::result::Result<String, anyhow::Error> {
+        // Currently only shas are allowed to be passed as graph artifact references
+        // TODO Update when tag reloading is implemented
+        let valid_regex = Regex::new(r"@sha256[0-9a-fA-F]{64}$").unwrap();
```
zw428: this should be `@sha256:` I think? missing a colon?

graytonio: Yes your right fixed now

apollo-router/src/executable.rs
```diff
@@ -227,7 +227,6 @@ pub struct Opt {
 
     /// An OCI reference to an image that contains the supergraph schema for the router.
     #[clap(long, env, action = ArgAction::Append)]
```
BrynCooke:
```suggestion
    #[clap(long, env = "APOLLO_GRAPH_ARTIFACT_REFERENCE", action = ArgAction::Append)]
```

BrynCooke: You need to also update clap.
Please add an integration test for this env variable. `IntegrationTest` supports env variables.

BrynCooke (changes requested):
Needs a test. Also technically this is a breaking change as this has been public for a long time. Are we sure no-one is using this?

apollo-router/src/executable.rs
goto-bus-stop: Oh, we actually don't have a way to specify env variables in `IntegrationTest`. I'm quite surprised by that...

To add support, I'd add an `env: HashMap<String, String>` argument to the integration test [builder](https://docs.rs/buildstructor/latest/buildstructor/) method here: https://github.com/apollographql/router/blob/a0acf2958a56bbd3d2a3dddb89acdcbad5b030aa/apollo-router/tests/common.rs#L530-L540

& then propagate them to the router subprocess in here
https://github.com/apollographql/router/blob/a0acf2958a56bbd3d2a3dddb89acdcbad5b030aa/apollo-router/tests/common.rs#L703

You can check usage sites of `IntegrationTest` for how tests are written with it. It might not be super obvious, @graytonio you can DM me on slack or put some time on my calendar if you want to pair a bit.

apollo-router/src/executable.rs
```diff
@@ -292,13 +291,27 @@ impl Opt {
...
+        // TODO Update when tag reloading is implemented
+        let valid_regex = Regex::new(r"@sha256:[0-9a-fA-F]{64}$").unwrap();
+
+        if valid_regex.is_match(reference) {
+            Ok(reference.to_string())
+        } else {
+            Err(anyhow!("Invalid graph artifact reference: {reference}"))
+        }
+    }
+
```
lrlna: this needs some unit tests as well

lrlna:
@BrynCooke the env var only got added on June 24 in #7474. this is the change to the docs, [for example](https://github.com/apollographql/router/pull/7474/files#diff-eeb39af37d7a07375d5289def3e061b1d8e62336242f94339dc9ff0f0daf79e1). It sounds like no one aside from us internally was able to actually use it, since it required a LaunchDarkly flag to be enabled and it was not enabled for anyone but our own deployment of the router. We probably shouldn't have added anything to the docs in #7474 if we knew it was an internal change, but it is what it is!

lrlna:
@graytonio give me a shout when you're ready for another review.

.github/CODEOWNERS
```diff
@@ -1,10 +1,9 @@
...
 /.changesets/ @apollographql/docs
 /apollo-federation/ @apollographql/fed-core @apollographql/rust-platform
-/apollo-federation/src/connectors @apollographql/graph-dev
-/apollo-router/ @apollographql/router-core
-/apollo-router/src/plugins/connectors @apollographql/graph-dev
-/apollo-router/src/plugins/fleet_detector.rs @apollographql/cloud-fleet
+/apollo-federation/src/sources/connect @apollographql/connectors
+/apollo-router/ @apollographql/router-core @apollographql/fed-core
 /apollo-router-benchmarks/ @apollographql/router-core @apollographql/fed-core
+/apollo-router-scaffold/ @apollographql/router-core @apollographql/fed-core
```
lrlna: why did this file change? The changes are not accurate based on actual teams' ownership.

apollo-router/src/registry/mod.rs
```diff
@@ -121,9 +122,34 @@ async fn pull_oci(
...
 /// Fetch an OCI bundle
 pub(crate) async fn fetch_oci(oci_config: OciConfig) -> Result<OciContent, OciError> {
+    tracing::debug!("OCI 1");
     let reference: Reference = oci_config.reference.as_str().parse()?;
+    tracing::debug!("OCI 2");
     let auth = build_auth(&reference, &oci_config.apollo_key);
-    pull_oci(&mut Client::default(), &auth, &reference).await
+    tracing::debug!("OCI 3");
+    let host = reference.registry();
+    tracing::debug!("OCI 4");
```
lrlna: From debugging POV, I am not sure what the numbering - OCI 1, OCI 2 - refers to. Can we make the debug messages a bit more clear?

apollo-router/tests/integration/fixtures/minimal-no-oci.router.yaml
lrlna, line 1: this file and `minimal-oci.router.yaml` are identical. do we need both?

apollo-router/tests/integration/oci.rs.backup
lrlna, line 1: was this file meant to be committed?

apollo-router/tests/integration/fixtures/minimal.router.yaml
lrlna, line 1: I think either this file name needs to be changed, or we should have a minimal, globally applicable router config. My suggestion would be to make a new directory for oci fixtures in `tests/integration/fixtures` so you could have an oci-specific minimal router config.

I also notice that this test is only being used in `.backup` file. Is that intentional?

apollo-router/tests/integration/oci.rs
```diff
@@ -0,0 +1,172 @@
...
+                String::from(
+                    "service:test-graph:abc123def456ghi789jkl012mno345pqr678stu901vwx234yz", // gitleaks:allow
+                )
+                .into(),
+            ),
+            (
+                String::from("APOLLO_GRAPH_ARTIFACT_REFERENCE"),
+                artifact_reference.into(),
+            ),
+        ]))
```
lrlna: the way we've done this before, is have the tests read `TEST_APOLLO_KEY` and `TEST_APOLLO_GRAPH_REF` from either local or CI, that way a key like this (for onlookers, this is a dummy key, and is gibberish) doesn't have to be hardcoded. [This is a good example](https://github.com/apollographql/router/blob/32eb88c87cf124caddc4fd1c70186a94810a67e0/apollo-router/tests/integration/traffic_shaping.rs#L183-L216). Any chance we can stick to the same pattern with this test?

apollo-router/tests/integration/oci.rs
```diff
@@ -0,0 +1,172 @@
...
+        )
+        .mount(&mock_server)
+        .await;
+
+    let artifact_reference = format!("{}/{}@{}", mock_server.address(), graph_id, manifest_digest);
+    (mock_server, artifact_reference)
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn test_router_boots_with_oci_config() -> Result<(), BoxError> {
```
lrlna, line 144: this is only the happy path test. do we expect this to fail in any way that would need an integration test?

apollo-router/src/executable.rs
```diff
@@ -292,13 +291,27 @@ impl Opt {
...
 
+    pub fn validate_oci_reference(reference: &str) -> std::result::Result<String, anyhow::Error> {
+        // Currently only shas are allowed to be passed as graph artifact references
+        // TODO Update when tag reloading is implemented
+        let valid_regex = Regex::new(r"@sha256:[0-9a-fA-F]{64}$").unwrap();
+        if valid_regex.is_match(reference) {
+            tracing::debug!("Validated OCI configuration");
+            Ok(reference.to_string())
+        } else {
+            Err(anyhow!("Invalid graph artifact reference: {reference}"))
```
lrlna: A nit on log and error messages: 

The standard in Rust is to not capitalise at the start of a message (logs, errors, asserts, panics etc), unless multiple sentences are absolutely a `must`. From [the docs](https://rustc-dev-guide.rust-lang.org/diagnostics.html#errors-and-lints):
>The text should be matter of fact and avoid capitalization and periods, unless multiple sentences are needed

So all of these `tracing::debug`, `Err(anyhow!)`, etc, should be like this:  

```suggestion
            tracing::debug!("validated OCI configuration");
            Ok(reference.to_string())
        } else {
            Err(anyhow!("invalid graph artifact reference: {reference}"))
```

apollo-router/src/executable.rs
```diff
@@ -857,4 +870,78 @@ mod tests {
...
+
+        for hash in valid_hashes {
+            let result = super::Opt::validate_oci_reference(hash);
+            assert!(result.is_ok(), "Hash '{}' should be valid", hash);
+            assert_eq!(result.unwrap(), hash);
+        }
+    }
+
+    #[test]
+    fn test_validate_oci_reference_invalid_cases() {
```
lrlna, line 893: thank you for this test! 🙏

.github/CODEOWNERS
sirdodger: That's a good question; these came as part of rebasing to the lastest main. It's probably merge damage. I'll revert it. I don't actually know the team names yet. :(

apollo-router/src/registry/mod.rs
sirdodger: I'll rip these out; they were for debugging only. I think they got reintroduced as part of the rebase.

apollo-router/tests/integration/oci.rs
sirdodger: Is the expectation that devs will always have TEST_APOLLO_KEY and TEST_APOLLO_GRAPH_REF defined locally? I didn't see any setup instructions on how to do that, and never realized that was an expectation.

This test doesn't explicitly depend on having a working uplink configuration, but it doesn't hurt either. By providing the fake (but properly formed) test key, it allows execution of the graph artifact launch scenario locally without contacting uplink. But on CI, both KEY and REF are set to real values, and so the test works there, also.

I guess I was leaning towards having the test execute properly locally in the absence of those env vars, but if the prevailing pattern is to silently fail, I can switch it over.

apollo-router/tests/integration/fixtures/minimal.router.yaml
sirdodger: I managed to create multiple files through several iterations through testing; I'll clean them up so there is only a single, properly named oci config.

apollo-router/tests/integration/oci.rs
sirdodger: I'll add a failure test for the common case just to catch regressions. Most of the failure cases just fail to launch with a different error message, though, so I don't see a lot of value in enumerating them all at the integration test level.

sirdodger:
> 5. And this is a meta question - do we want metrics on whether the schema got pulled from OCI or not? Any other metrics that would be useful?

@lrlna That is a great question. Just for my education, metrics in router are set up to deliver otel metrics to the customer, not to Apollo, correct? So introducing a metric at this level would help a customer who is running a large fleet of routers to know how many were configured to launch via graph artifact?

I think that would only be valuable if e.g., we tracked all the different ways you could launch a router, so that if they were migrating from local files to oci registry, they would have an easy way to watch the count of routers launched via file go down and the number launched by registry go up. Even that seems like a bit of an edge case metric to me. I'd expect most fleets to cut over from one configuration to another, instead of being a gradual rollout.

That metric is VERY interesting to Apollo though, so tracking on our side how often the artifact registry is used is critical. I don't think that is the router's problem, though.


Other metrics:

Latency in querying for a graph artifact? Tracking one part of the startup latency doesn't seem to add much value, when what I would care about would probably be total startup time. (And even then, I'd be more focused on runtime stats than startup stats.) 

Schema size? Doesn't seem useful in isolation to artifacts. Total schema size might be a fun router metric to track over time, and could maybe flag major unintended changes in the supergraph schema. We track that Apollo-side because of the correlation with performance of the build, but I am not sure what a customer might find there. I think if I were tracking graph size, I'd want per-subgraph sizes so I could see what services were generating the largest API exposure.


Nothing is jumping out at me as being high value, but I'd love other opinions. Maybe if there is a significant metric we want to add that includes graph artifacts we can add another PR for it.

apollo-router/tests/integration/oci.rs
sirdodger: I've gone ahead and added the short-circuit. I can still test locally without a valid key by only setting the TEST_APOLLO_KEY variable.

apollo-router/tests/integration/oci.rs
lrlna: Sounds good to me, thank you!

lrlna:
> metrics in router are set up to deliver otel metrics to the customer, not to Apollo, correct? So introducing a metric at this level would help a customer who is running a large fleet of routers to know how many were configured to launch via graph artifact?

All the `apollo.router.*` metrics can be sent to Apollo as well as the user. We just need to make sure the data team knows to ingest them. I agree with you that it's only valuable if we look at whole picture and the various ways schema can be provided to the router:

* schema is available at startup either via:
    * env var (the router won't be able to know if that env var did get populated via the oci registry though)
    * command-line argument
* schema is hot-reloaded either via:
    * oci registry
    * previous uplink registry

I am not sure though if that information is valuable to users. Like you said, a given fleet, or even just one router is likely to just change from one method to another, rather than doing a rollout.

We do have a couple of existing `uplink` metrics which will be good to mirror for the OCI registry as well:
* `apollo.router.uplink.fetch.duration.seconds` (histogram)
* `apollo.router.uplink.fetch.count.total` (counter)

apollo-router/src/registry/mod.rs
```diff
@@ -123,7 +124,28 @@ async fn pull_oci(
 pub(crate) async fn fetch_oci(oci_config: OciConfig) -> Result<OciContent, OciError> {
     let reference: Reference = oci_config.reference.as_str().parse()?;
     let auth = build_auth(&reference, &oci_config.apollo_key);
-    pull_oci(&mut Client::default(), &auth, &reference).await
+    let host = reference.registry();
+    let protocol = if host.starts_with("localhost") || host.starts_with("127.0.0.1") {
+        ClientProtocol::Http
+    } else {
```
lrlna: Should we use this protocol outside of tests? I take it this was here while we were testing locally, I am not sure if we should merge it as is.

apollo-router/tests/integration/oci.rs
lrlna: >Is the expectation that devs will always have TEST_APOLLO_KEY and TEST_APOLLO_GRAPH_REF defined locally? I didn't see any setup instructions on how to do that, and never realized that was an expectation.

Yea, we have those vars set. We should add that to our dev docs 🙈 

> but if the prevailing pattern is to silently fail, I can switch it over.

It silently "passes", which is I guess makes all of them Schrödinger's tests. I'll take an action item to at least emit a message to devs where we don't do anything without those vars set. But it'll be easier to fix this if all tests follow the same pattern.

>By providing the fake (but properly formed) test key, it allows execution of the graph artifact launch scenario locally without contacting uplink. But on CI, both KEY and REF are set to real values, and so the test works there, also.

Ah I see, there isn't a specific dependence of graphos being enabled then. That seems all right, then.

lrlna: ah, look at that, @aaronArinder is adding a note about these vars to the docs already! https://github.com/apollographql/router/pull/8198

apollo-router/src/registry/mod.rs
sirdodger: The CI tests require it as well. Because the registry ref can include hostname but not protocol, the choice was either to expose protocol as a top-level variable, or infer it from the URL. The former would be more explicit and flexible, but it also adds top-level configuration for minimal gain.

It seems that a casual user who is debugging/testing things out might also want the ability to fire up a local OCI registry and not have to bother with SSL certs.

Pros: Easy local testing. No confusing/largely unused top-level config.
Cons: Not particularly discoverable. Looks like weird special case. Unexpected behavior if someone creates a domain like localhost.foo.com, though maybe they deserve the hassle then.

sirdodger:
I looked at the changeset docs and it seemed like that was an independent process that gets its own PR?

lrlna:
> Open a pull request for it. Note that your PR title and body will be used to pre-populate the changeset.

@sirdodger, ah, I see how the wording on that is confusing. The `it` in `Open a pull request for it` refers to the change you're making, aka this existing PR. That's why the checklist [in the description of your PR](https://github.com/apollographql/router/pull/8081#issue-3311561163) includes:

> [ ] Changeset is included for user-facing changes

apollo-router/src/registry/mod.rs
```diff
@@ -95,29 +96,73 @@ async fn pull_oci(
...
+        }
+        Err(e) => {
+            u64_counter!(
+                "apollo.router.artifact.fetch.count.total",
+                "Total number of requests to Apollo Graph Artifact Registry",
+                1u64,
+                status = "failed",
+                reference = reference.whole(),
+                error=e.to_string()
+            );
```
BrynCooke: Remove the counter. Histograms already include a count.

apollo-router/src/registry/mod.rs
```diff
@@ -95,29 +96,73 @@ async fn pull_oci(
...
+
+    }.await {
+        Ok(content) => {
+            u64_counter!(
+                "apollo.router.artifact.fetch.count.total",
+                "Total number of requests to Apollo Graph Artifact Registry",
+                1u64,
+                status = "success",
+                reference = reference.whole()
+            );
```
BrynCooke: Remove the counter. Histograms already include a count.

apollo-router/src/registry/mod.rs
```diff
@@ -95,29 +96,73 @@ async fn pull_oci(
...
+            f64_histogram!(
+                "apollo.router.artifact.fetch.duration.seconds",
```
BrynCooke: Remove the `.seconds` suffix. All metrics are always in si units and otel semantic conventions don't use such suffixes.

apollo-router/src/registry/mod.rs
```diff
@@ -95,29 +96,73 @@ async fn pull_oci(
...
+            u64_counter!(
+                "apollo.router.artifact.fetch.count.total",
+                "Total number of requests to Apollo Graph Artifact Registry",
+                1u64,
+                status = "failed",
+                reference = reference.whole(),
+                error=e.to_string()
+            );
+            f64_histogram!(
+                "apollo.router.artifact.fetch.duration.seconds",
```
BrynCooke: Remove the `.seconds` suffix. All metrics are always in si units and otel semantic conventions don't use such suffixes.

sirdodger:
@BrynCooke As it turns out, the metrics don't work there because telemetry hasn't been set up yet. Thanks for the info, though; it'll come in handy next time.

docs/source/routing/observability/index.mdx
```diff
@@ -59,7 +59,7 @@ Not an enterprise customer? [Learn about GraphOS for Enterprise.](https://www.ap
 
 - Learn how to use [GraphOS Insights](/graphos/platform/insights/) to monitor and improve your graph's performance.
 
-- Learn how to [configure router telemetry](/graphos/routing/observability/telemetry)
+- Learn how to [configure router telemetry](/graphos/routing/observability/otel)
```
lrlna: This now points to `otel` docs for [GraphOS and Federation as a whole](https://www.apollographql.com/docs/graphos/routing/observability/otel), and doesn't really explain how to configure telemetry in the router, which is what the [original link was pointing to](https://www.apollographql.com/docs/graphos/routing/observability/telemetry). Was there a reason to change this to point to "OpenTelemetry in Apollo Federation"?

docs/source/routing/observability/telemetry/instrumentation/standard-instruments.mdx
```diff
@@ -144,6 +145,24 @@ The initial call to Uplink during router startup is not reflected in metrics.
...
+  - `status`: (`success`, `failure`)
+  - `reference`: The OCI reference used to fetch the schema
+  - `result`: The content object or error result
+
+<Note>
+
+Fetch duration includes both the time spent querying the OCI manifest and the time spent retrieving the schema blob.
+
+</Note>
+
```
lrlna: Since these metrics are no longer added, we should remove these docs.

```suggestion
```

.changesets/config_graph_artifacts.md
```diff
@@ -0,0 +1,7 @@
+### Add validation and metrics when fetching the schema from the Apollo Graph Artifact Registry ([PR #8081](https://github.com/apollographql/router/pull/8081))
+
+Rename the `GRAPH_ARTIFACT_REFERENCE` configuration environment variable to `APOLLO_GRAPH_ARTIFACT_REFERENCE` for consistency and added validation.
+
+Adds `apollo.router.artifact.fetch.count.total` and `apollo.router.artifact.fetch.duration.seconds` metrics.
```
lrlna: This is no longer accurate since we are not adding the metrics, right? And since Graph Artifacts have not been publicly available, this is going to be the last PR that makes them available to everyone else (in the router, at least)? Should this be a more descriptive changelog describing what they are and where users can learn more about them? If so, we might potentially need help from @davidstaheli here (if David is indeed the PM responsible for graph artifacts, and I don't have my wires crossed).

apollo-router/src/registry/mod.rs
lrlna: I needed some time to think about this. You're right - having a config option, at least for now, is too much. So having an inference like this makes a bit more sense. Could you add some comments to this code block explaining what and why?

The `.registry()` is meant to return the name of the registry. So we expect that in local testing that will only be `localhost` and `127.0.0.1`, to triple check?

docs/source/routing/observability/index.mdx
sirdodger: Oh that's strange; I updated it because I was getting 404s on the link, but I see the page and it is working now. I'll back out the change.

.changesets/config_graph_artifacts.md
sirdodger: @davidstaheli can comment, but I think there is some tension between wanting the full docs here and waiting for the public announcement of the feature.

apollo-router/src/registry/mod.rs
sirdodger: `.registry()` returns both host and optionally port. I can split and check only direct matches against the host portion to be a little cleaner.

We obviously can't infer all cases of how someone would want to test without SSL, but these are the most common.

.changesets/config_graph_artifacts.md
davidstaheli: Hi @lrlna and @sirdodger. I'd prefer we avoid mentioning Graph Artifacts before Summit. Can we remove these lines?

lrlna: @davidstaheli so you all make a separate PR afterwards, closer to summit, to add detailed docs on how to use artifacts with the router and what router release it's available in? Because a bunch of it is already in the docs, as this PR (correctly) makes changes to those docs with the new env var.

apollo-router/src/registry/mod.rs
sirdodger: Split out an explicit function for the inference and added a docstring and tests.

PR #8082, opened 2025-08-11
Connectors: Add `??` and `?!` operators

andrewmcgivery:
---

**Checklist**

- Tests added and passing[^4]
    - [x] Unit tests
    - [ ] Integration tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-federation/src/connectors/json_selection/README.md
+++ b/apollo-federation/src/connectors/json_selection/README.md
@@ -96,7 +96,7 @@ PathStep             ::= "." Key | "->" Identifier MethodArgs?
 Key                  ::= Identifier | LitString
 Identifier           ::= [a-zA-Z_] NO_SPACE [0-9a-zA-Z_]*
 MethodArgs           ::= "(" (LitExpr ("," LitExpr)* ","?)? ")"
-LitExpr              ::= LitPath | LitPrimitive | LitObject | LitArray | PathSelection
+LitExpr              ::= LitPath | LitPrimitive | LitObject | LitArray | PathSelection | LitExpr "??" LitExpr | LitExpr "?!" LitExpr
 LitPath              ::= (LitPrimitive | LitObject | LitArray) NonEmptyPathTail
 LitPrimitive         ::= LitString | LitNumber | "true" | "false" | "null"
 LitString            ::= "'" ("\\'" | [^'])* "'" | '"' ('\\"' | [^"])* '"'
@@ -896,6 +896,31 @@ powerful ways, e.g. `page: list->slice(0, $limit)`.
 Also, as a minor syntactic convenience, `LitObject` literals can have
 `Identifier` or `LitString` keys, whereas JSON objects can have only
 double-quoted string literal keys.
+
+#### Null-coalescing operators
+
+The `LitExpr` syntax also supports two null-coalescing operators: `??` and `?!`. These operators provide fallback values when expressions evaluate to `null` or are missing entirely (`None`):
+
+- `A ?? B` (null-coalescing): Returns `B` if `A` is `null` or `None`, otherwise returns `A`
+- `A ?! B` (none-coalescing): Returns `B` if `A` is `None`, otherwise returns `A` (preserving `null` values)
+
+These operators are left-associative and can be chained:
+
+```graphql
+# Basic usage
+fallback: $($.missingField ?? "default")
+preserveNull: $($.nullField ?! "default")  # keeps null as null
+
+# Chaining (left-to-right evaluation)
+multiLevel: $($.first ?? $.second ?? $.third ?? "final fallback")
+
+# Combined with other expressions  
+computed: $($.value ?? 0->add(10))
+```
+
+The key difference between the operators is how they handle `null` values:
+- `??` treats both `null` and `None` (missing) as "falsy" and uses the fallback
+- `?!` only treats `None` (missing) as "falsy", preserving explicit `null` values
 
 ### `LitPath ::= (LitPrimitive | LitObject | LitArray) PathStep+`
 
--- a/apollo-federation/src/connectors/json_selection/apply_to.rs
+++ b/apollo-federation/src/connectors/json_selection/apply_to.rs
@@ -894,6 +894,50 @@ impl ApplyToInternal for WithRange<LitExpr> {
                 .and_then_collecting_errors(|value| {
                     subpath.apply_to_path(value, vars, input_path, spec)
                 }),
+            LitExpr::NullCoalescing(left, right) => {
+                let (left_value, left_errors) = left.apply_to_path(data, vars, input_path, spec);
+
+                match left_value {
+                    // If left is null or None, evaluate right operand
+                    None | Some(JSON::Null) => {
+                        let (right_value, right_errors) =
+                            right.apply_to_path(data, vars, input_path, spec);
+                        if right_value.is_some() {
+                            // If right operand succeeds, suppress left operand errors
+                            (right_value, right_errors)
+                        } else {
+                            // If right operand also fails, include both errors
+                            let mut all_errors = left_errors;
+                            all_errors.extend(right_errors);
+                            (right_value, all_errors)
+                        }
+                    }
+                    // Otherwise, return left value
+                    Some(value) => (Some(value), left_errors),
+                }
+            }
+            LitExpr::NoneCoalescing(left, right) => {
+                let (left_value, left_errors) = left.apply_to_path(data, vars, input_path, spec);
+
+                match left_value {
+                    // If left is None (but not null), evaluate right operand
+                    None => {
+                        let (right_value, right_errors) =
+                            right.apply_to_path(data, vars, input_path, spec);
+                        if right_value.is_some() {
+                            // If right operand succeeds, suppress left operand errors
+                            (right_value, right_errors)
+                        } else {
+                            // If right operand also fails, include both errors
+                            let mut all_errors = left_errors;
+                            all_errors.extend(right_errors);
+                            (right_value, all_errors)
+                        }
+                    }
+                    // Otherwise, return left value (including null)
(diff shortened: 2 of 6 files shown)


apollo-federation/src/connectors/json_selection/README.md
```diff
@@ -96,7 +96,7 @@ PathStep             ::= "." Key | "->" Identifier MethodArgs?
 Key                  ::= Identifier | LitString
 Identifier           ::= [a-zA-Z_] NO_SPACE [0-9a-zA-Z_]*
 MethodArgs           ::= "(" (LitExpr ("," LitExpr)* ","?)? ")"
-LitExpr              ::= LitPath | LitPrimitive | LitObject | LitArray | PathSelection
+LitExpr              ::= LitPath | LitPrimitive | LitObject | LitArray | PathSelection | LitExpr "??" LitExpr | LitExpr "?!" LitExpr
```
benjamn: I think we should design the grammar to support more infix operators in the future, without adding more unnamed cases to `LitExpr`.

An alternative grammar I've been considering looks like this:
```diff
-LitExpr    ::= LitPath | LitPrimitive | LitObject | LitArray | PathSelection
-LitPath    ::= (LitPrimitive | LitObject | LitArray) PathStep+
+LitExpr    ::= LitPath | LitOpChain | LitPrimitive | LitObject | LitArray | PathSelection
+LitPath    ::= (LitPrimitive | LitObject | LitArray) NonEmptyPathTail
+LitOpChain ::= LitExpr LitOp LitExpr
+LitOp      ::= "??" | "?!" | "&&" | "||" | "==" | "!=" | "<" | "<=" | ">" | ">=" | "+" | "-" | "*" | "/" | "%"
```
As you can see, `??` and `?!` are just two examples of potentially many infix operators, so it would be convenient if the rest of the grammar didn't have to change as we add more operators.

apollo-federation/src/connectors/json_selection/README.md
```diff
@@ -893,10 +895,66 @@ The `LitExpr` mini-language diverges from JSON by allowing symbolic
...
+
+**Important**: Mixing operators like `??` and `?!` in the same expression chain is not allowed. Each operator chain must use only one type of operator. This restriction ensures predictable evaluation semantics and leaves room for future operator precedence rules.
+
+The key difference between the operators is how they handle `null` values:
+- `??` treats both `null` and `None` (missing) as "falsy" and uses the fallback
+- `?!` only treats `None` (missing) as "falsy", preserving explicit `null` values
+
+### `LitOpChain ::= LitExpr LitOp LitExpr`
+
+![LitOpChain](./grammar/LitOpChain.svg)
```
andrewmcgivery, line 945: These SVG's obviously don't exist... will need some help from Ben to craft these. :)

apollo-federation/src/connectors/json_selection/lit_expr.rs
```diff
@@ -102,6 +184,13 @@ impl LitExpr {
         }
     }
 
+    fn parse_binary_operator(input: Span) -> ParseResult<LitOp> {
+        alt((
+            map(ranged_span("??"), |_| LitOp::NullCoalescing),
+            map(ranged_span("?!"), |_| LitOp::NoneCoalescing),
```
benjamn:
```suggestion
    fn parse_binary_operator(input: Span) -> ParseResult<WithRange<LitOp>> {
        alt((
            map(ranged_span("??"), |qq| WithRange::new(LitOp::NullCoalescing, qq.range())),
            map(ranged_span("?!"), |qi| WithRange::new(LitOp::NoneCoalescing, qi.range())),
```
Unless you want to make `LitOp` carry its own range information somehow, the `WithRange<T>` wrapper is the utility we use around here to preserve source range information.

> [!WARNING]
> Code above hand-written in GitHub web UI, not compiled or tested, and may require other adjustments to work properly.

apollo-federation/src/connectors/json_selection/lit_expr.rs
```diff
@@ -54,13 +55,94 @@ pub(crate) enum LitExpr {
     // of the path, which is never PathList::Empty, because that would mean the
     // LitExpr could stand on its own, using one of the other variants.
     LitPath(WithRange<LitExpr>, WithRange<PathList>),
+
+    // Operator chains: A op B op C ... where all operators are the same type
+    // OpChain contains the operator type and a vector of operands
+    // For example: A ?? B ?? C becomes OpChain(NullCoalescing, [A, B, C])
+    OpChain(LitOp, Vec<WithRange<LitExpr>>),
```
benjamn:
```suggestion
    // Operator chains: A op B op C ... where all operators are the same type
    // OpChain contains the operator type and a vector of operands
    // For example: A ?? B ?? C becomes OpChain(NullCoalescing, [A, B, C])
    OpChain(WithRange<LitOp>, Vec<WithRange<LitExpr>>),
```
In order to store source range information for `LitOp` operators, I think you want another `WithRange<_>` wrapper at this level (the `LitExpr::OpChain` enum variant).

apollo-federation/src/connectors/json_selection/lit_expr.rs
benjamn: Alternatively, you might just want to implement
```rust
impl LitOp {
    fn parse(input: Span) -> ParseResult<WithRange<Self>> {
        // ... contents of LitExpr::parse_binary_operator here...
    }
}
```

apollo-federation/src/connectors/json_selection/README.md
benjamn: I'll push a commit with your SVGs today, but for anyone who wants to try their hand at railroad diagram generation, here are the tools I use:
* Source: https://github.com/GuntherRademacher/rr
* Hosted: https://www.bottlecaps.de/rr/ui

apollo-federation/src/connectors/json_selection/apply_to.rs
```diff
@@ -954,6 +1010,23 @@ impl ApplyToInternal for WithRange<LitExpr> {
                     literal.compute_output_shape(context, input_shape, dollar_shape.clone());
                 subpath.compute_output_shape(context, literal_shape, dollar_shape)
             }
+
+            LitExpr::OpChain(_, operands) => {
+                // For any coalescing operator chain, the result can be any of the operands
+                let shapes: Vec<Shape> = operands
```
benjamn: Let's go ahead and break this shape code into the different known `LitOp` cases, something like
```rust
LitExpr::OpChain(op, operands) => {
    match op {
        LitOp::NullCoalescing | LitOp::NoneCoalescing => {
            // Same shape code for both of these cases
        }
        // Keep the match exhaustive, with no _ => {} catch-all case
    }
```
Keeping the `match` exhaustive means we will be forced to update it when we add more operators, which is good.

apollo-federation/src/connectors/json_selection/lit_expr.rs
```diff
@@ -54,13 +55,94 @@ pub(crate) enum LitExpr {
...
+}
+
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub(crate) enum LitOp {
+    NullCoalescing, // ??
+    NoneCoalescing, // ?!
```
benjamn: Naming discussion/idea:
```suggestion
#[derive(Debug, PartialEq, Eq, Clone)]
pub(crate) enum LitOp {
    NullCoalescing, // ??
    NoneDefaulting, // ?!
```
As I understand it, the word "coalescing" in the [ECMAScript/JavaScript nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing) syntax refers to the way `null` and `undefined` get handled the same way (coalesced). That's what we're aiming for with `??`, so `NullCoalescing` works for that operator name. If we want to lean into the JS precedent even further, we might call `??` the `LitOp::NullishCoalescing` operator, where "nullish" values are `null` and `None`, but I don't think `NullCoalescing` is wrong or misleading.

However, with our `?!` operator, multiple input values are not being coalesced, but instead we are defaulting to another value when the first value is `None`/missing. I don't think `NoneCoalescing` quite fits for that, but I like `NoneDefaulting` (or possibly just `NoneDefault`). Thoughts?

apollo-federation/src/connectors/json_selection/README.md
```diff
@@ -893,10 +895,66 @@ The `LitExpr` mini-language diverges from JSON by allowing symbolic
...
+**Important restriction**: All operators in a chain must be the same type. Mixed operators like `A ?? B ?! C` will fail to parse as a single `LitOpChain`, with the parser stopping after the first operator change and leaving unparsed remainder.
+
+### `LitOp ::= "??" | "?!"`
+
+![LitOp](./grammar/LitOp.svg)
+
+The `LitOp` rule defines the binary operators currently supported in `LitOpChain` expressions:
+
+- `??` (null-coalescing): Returns the right operand if the left operand is `null` or `None` (missing)
+- `?!` (none-coalescing): Returns the right operand if the left operand is `None` (missing), but preserves `null` values
```
benjamn, line 969: Maybe call `?!` the `None-defaulting` operator (see https://github.com/apollographql/router/pull/8082#discussion_r2270551895 for discussion)?

apollo-federation/src/connectors/json_selection/lit_expr.rs
andrewmcgivery: After discussion on slack, we are settling on "NullishCoalescing" and "NoneCoalescing".

apollo-federation/src/connectors/json_selection/README.md
benjamn: Just pushed https://github.com/apollographql/router/pull/8082/commits/f0c564f0c717b804c882055a17bcc924b23aae4f to update the SVGs

benjamn (approved):
Approved pending CI passing!

PR #8083, opened 2025-08-12
(feat) Enhance chaos plugin

BrynCooke:
The experimental chaos plugin was not actually useful because the router detects that a schema reload event is the same as the last seen schema and skips reloading.

This PR fixes and enhances the chaos plugin so that we can actually use it for testing.

Changes:
- Split out chaos reload and sighup reload into separate units of functionality.
- Chaos plugin:
  - Fix ineffective schema reload.
  - Add force_schema_reload and force_config_reload with independent timers
  - Automatically capture and replay events with modifications to force reloads:
    - Schema events: inject timestamp comments into SDL
    - Config events: clone and re-emit configuration

As experimental chaos is not officially supported there is no changeset, documentation and the config is allowed to have breaking changes.

---

**Checklist**

- Tests added and passing[^4]
    - [x] Unit tests
    - [x] Integration tests

**Exceptions**

*Note any exceptions here*

**Notes**

### DIFF
--- a/apollo-router/src/axum_factory/tests.rs
+++ b/apollo-router/src/axum_factory/tests.rs
@@ -23,11 +23,14 @@ use http::header::CONTENT_TYPE;
 use http::header::{self};
 #[cfg(unix)]
 use http_body_util::BodyExt;
+#[cfg(unix)]
 use hyper::rt::ReadBufCursor;
+#[cfg(unix)]
 use hyper_util::rt::TokioIo;
 use mime::APPLICATION_JSON;
 use mockall::mock;
 use multimap::MultiMap;
+#[cfg(unix)]
 use pin_project_lite::pin_project;
 use reqwest::Client;
 use reqwest::Method;
@@ -44,8 +47,10 @@ use serde_json::json;
 use test_log::test;
 use tokio::io::AsyncRead;
 use tokio::io::AsyncReadExt;
+#[cfg(unix)]
 use tokio::io::AsyncWrite;
 use tokio::io::AsyncWriteExt;
+#[cfg(unix)]
 use tokio::io::ReadBuf;
 use tokio::sync::mpsc;
 use tokio_util::io::StreamReader;
--- a/apollo-router/src/router/event/reload.rs
+++ b/apollo-router/src/router/event/reload.rs
@@ -1,3 +1,4 @@
+#[cfg(unix)]
 use std::task::Poll;
 
 use futures::prelude::*;


apollo-router/src/router/mod.rs
```diff
@@ -231,29 +232,19 @@ fn generate_event_stream(
...
-            })
-            .boxed(),
+        configuration.into_stream(uplink_config).boxed(),
         shutdown_receiver
             .into_stream()
             .map(|_| Event::Shutdown)
             .boxed(),
     ])
+    .with_sighup_reload()
+    .with_chaos_reload()
```
BrynCooke, line 246: Look how neat this is now!

apollo-router/src/plugins/chaos/mod.rs
```diff
@@ -0,0 +1,139 @@
...
+/// configuration is cloned and re-emitted, which triggers the router's configuration change
+/// detection and reload logic.
+///
+/// ### Automatic Operation
+/// The chaos system requires no manual intervention once configured:
+/// - It automatically captures schema and configuration events as they flow through the system
+/// - Timer-based reloads only occur if there's a "last known" event to replay
+/// - Each chaos timer operates independently - you can enable schema reloading without
+///   configuration reloading, or vice versa
+/// - The system gracefully handles cases where no events are available to replay
```
goto-bus-stop: this mostly restates things that were already mentioned, I think it would be clearer if this list was deleted and the idea from the second point (it 'starts working' when config/schema become available) is merged up into the "How Chaos Reloading Works" paragraph.

apollo-router/src/plugins/chaos/mod.rs
```diff
@@ -0,0 +1,139 @@
...
+/// - Merges the upstream events with periodic chaos reload events
+pub(crate) trait ChaosEventStream: Stream<Item = Event> + Sized {
+    /// Add chaos reload functionality to an event stream.
+    ///
+    /// This method wraps the event stream to automatically capture schema and configuration
+    /// events for later replay during chaos testing. The reload source will emit modified
+    /// versions of these events at configured intervals to force hot reloads.
+    ///
+    /// The chaos reload timers are automatically configured when configuration events
+    /// flow through the stream - no manual setup is required.
```
goto-bus-stop: arguably calling `.with_chaos_reload`(`_state`) _is_ the manual setup required ;)

apollo-router/src/plugins/chaos/mod.rs
BrynCooke: [14356e1](https://github.com/apollographql/router/pull/8083/commits/14356e10e118272939e79cf5f78565df4f4d2925)

apollo-router/src/plugins/chaos/mod.rs
BrynCooke: [14356e1](https://github.com/apollographql/router/pull/8083/commits/14356e10e118272939e79cf5f78565df4f4d2925)

PR #8095, opened 2025-08-18
Caroline/response caching sets v2 no reverse index

bnjjj:
WIP


apollo-router/src/cache/redis.rs
```diff
@@ -723,6 +744,33 @@ impl RedisCacheStorage {
...
+            Box::pin(client.scan_cluster_buffered(pattern, None, None))
+        } else {
+            Box::pin(client.scan_buffered(pattern, None, None))
+        };
+
+        let mut keys = Vec::new();
+        while let Some(res) = stream.next().await {
+            if let Ok(key) = res {
+                keys.push(key)
+            }
```
bnjjj: Should we throw an error if there's an error ?

apollo-router/src/plugins/response_cache/plugin.rs
```diff
@@ -231,8 +183,8 @@ const fn default_lru_private_queries_size() -> NonZeroUsize {
 #[derive(Clone, Debug, JsonSchema, Deserialize, Serialize)]
 #[serde(rename_all = "snake_case", deny_unknown_fields, default)]
 pub(crate) struct Subgraph {
-    /// PostgreSQL configuration
-    pub(crate) postgres: Option<PostgresCacheConfig>,
+    /// Redis configuration
+    pub(crate) redis: Option<RedisCacheConfig>,
```
bnjjj, line 181: We have to double check there's a configuration for idle_timeout and acquire_timeout, I think have 2 distinct timeout config is really useful.

apollo-router/src/plugins/response_cache/plugin.rs
```diff
@@ -312,61 +264,61 @@ impl PluginPrivate for ResponseCache {
...
-                        return Err(e.into());
+                        return Err(e);
                     } else {
-                        let pg_cache_storage = Arc::new(OnceLock::new());
+                        let redis_cache_storage = Arc::new(OnceLock::new());
                         task_aborts.push(
-                            tokio::spawn(check_pg_connection(
-                                postgres_config,
-                                pg_cache_storage.clone(),
+                            tokio::spawn(check_redis_connection(
```
bnjjj, line 263: Is the behavior with Redis and fred the same than with pg and sqlx ? I mean if we already opened connections and we loose connection in the middle of the process, is it trying to automatically reconnect once it's available ?

apollo-router/src/plugins/response_cache/snapshots/apollo_router__plugins__response_cache__tests__private_and_public-3.snap.new
bnjjj, line 1: This one should be deleted ? Or accepted ?

apollo-router/src/plugins/response_cache/storage/mod.rs
```diff
@@ -0,0 +1,195 @@
...
+        let now = Instant::now();
+        let result = self._invalidate(invalidation_keys, subgraph_names).await;
+
+        let elapsed = now.elapsed().as_secs_f64();
+        f64_histogram_with_unit!(
+            "apollo.router.operations.response_cache.storage.invalidate",
+            "Time to get invalidate data in cache",
+            "s",
+            elapsed,
+            "kind" = "specific"
```
bnjjj, line 232: Maybe we should be more granular here

apollo-router/src/plugins/response_cache/storage/redis.rs
```diff
@@ -0,0 +1,427 @@
...
+}
+
+impl Storage {
+    pub(crate) async fn new(config: &Config) -> Result<Self, BoxError> {
+        // TODO: make the 'caller' parameter include the namespace? or subgraph name?
+        let storage = RedisCacheStorage::new(config.clone(), "response-cache").await?;
+
+        let mut long_config = config.clone();
+        long_config.timeout = Some(Duration::from_secs(10));
+        let long_storage = RedisCacheStorage::new(long_config, "response-cache-long").await?;
```
bnjjj: I'm not sure to understand this part

apollo-router/src/plugins/response_cache/storage/redis.rs
```diff
@@ -0,0 +1,427 @@
...
+
+    pub(crate) async fn perform_periodic_maintenance(&self) {
+        let storage = self.long_storage.clone();
+        let key = self.make_key("cache-tags");
+
+        // maintenance 1: take random members from cache-tags and use zremrangebyscore on them
+        tokio::spawn(async move {
+            let mut interval = tokio::time::interval(Duration::from_secs(1));
+            interval.set_missed_tick_behavior(MissedTickBehavior::Skip);
+            loop {
```
bnjjj: NIT: I think converting this interval to an IntervalStream is easier to read

apollo-router/src/plugins/response_cache/storage/redis.rs
```diff
@@ -0,0 +1,427 @@
...
+                    .client()
+                    .zremrangebyscore(&cache_tag_key, f64::NEG_INFINITY, cutoff as f64)
+                    .await
+                    .unwrap_or_else(|err| {
+                        // TODO error handling
+                        eprintln!("error while removing keys from cache-tag: {err:?}");
+                        0
+                    });
+
+                u64_counter!(
```
bnjjj: with_unit

bnjjj: Do we need a counter or a gauge ?

apollo-router/src/plugins/response_cache/storage/redis.rs
```diff
@@ -0,0 +1,427 @@
...
+                        // TODO error handling
+                        eprintln!("error while removing cache-tags: {err:?}");
+                        0
+                    });
+
+                u64_counter!(
+                    "apollo.router.operations.response_cache.storage.maintenance.removed_cache_tags",
+                    "Counter for removed items",
+                    removed_items
+                );
```
bnjjj: Same comment

apollo-router/src/plugins/response_cache/storage/redis.rs
```diff
@@ -0,0 +1,427 @@
...
+            let value = CacheValue {
+                data: document.data,
+                cache_control: document.cache_control,
+            };
+
+            // TODO: figure out if this is actually how we want to store the values
+            let _: () = pipeline
+                .set(
+                    document.cache_key,
+                    &serde_json::to_string(&value).unwrap(),
```
bnjjj, line 352: Handle error

apollo-router/src/plugins/response_cache/storage/redis.rs
bnjjj, line 1: I think we should build a lot of tests to test these methods. We didn't need it for pg because we don't have to test pg indexes and tables. But here we're doing it ourself so I think we should test it properly

apollo-router/src/plugins/response_cache/plugin.rs
carodewig: Yes, it'll automatically reconnect - right now the policy is hardcoded, but we could make it configurable if we wanted: https://github.com/apollographql/router/blob/af324c7ba2233087ffb90bd36d10e8f649156fcf/apollo-router/src/cache/redis.rs#L331

apollo-router/src/plugins/response_cache/plugin.rs
carodewig: Unfortunately fred doesn't really support that -  since it uses connection multiplexing, commands are just added to each client's internal pipeline and sent, rather than having a client which can be acquired :(
But I'll take a look to see how else we could improve that experience

apollo-router/src/plugins/response_cache/storage/redis.rs
carodewig: This was a test I made to see if splitting up the commands to different client pools would help with GET performance - there was one pool for the long commands (insert/invalidation/maintenance) and one pool for the gets. I don't think it really made a difference so will probably end up removing this

apollo-router/src/plugins/response_cache/storage/redis.rs
carodewig: 1d232568291f6bcb6cb04f506688266490092100

carodewig:
Working on creating several distinct PRs to make the review process easier.

PR #8101, opened 2025-08-19
fix(composition): fix check for root type extensions

dariuszkuc:
Fix logic to correctly check for root type extensions ([source](https://github.com/apollographql/federation/blob/main/internals-js/src/schemaUpgrader.ts#L308)). We were incorrectly checking for `object || !root_type` whereas we should be checking for `object && root_type`.


tninesling (approved):

duckki:
Wait a sec. The previous logic doesn't look wrong to me.

JS: `A && B && C`
Rust (previously):
```rust
  if !A || !B { return false; }
  C
```

Now, the updated code:
```rust
  if !A && B { return false; }
  C
```

dariuszkuc:
While I'm taking another look I created revert PR https://github.com/apollographql/router/pull/8107

dariuszkuc:
Indeed that check was fine. The issue is with incorrect response from `ty.has_non_extension_elements()`

PR #8109, opened 2025-08-19
fix(federation): removed `RebaseError::InterfaceObjectTypename` error

duckki:
### Summary

- The `rebase_on` method was prohibiting __typename field from being rebased on an interface object type by raising `InterfaceObjectTypename` error.
- But, it seemed unnecessarily prohibitive, rejecting plannable queries.
- Although it's odd to fetch `__typename` from interface object types, there are already several other cases where such `__typename` selections are added to interface object types.
- Such `__typename` selections are benign and we currently don't plan to remove all of those cases at this time.
- Thus, this PR removes the error variant and the check for the error altogether.
- Also, added a regression test that caused the error previously.

---

**Checklist**

- Tests added and passing[^4]
    - [x] Unit tests
    - [ ] Integration tests

[FED-678]: https://apollographql.atlassian.net/browse/FED-678?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changesets/fix_duckki_fed_678.md
+++ b/.changesets/fix_duckki_fed_678.md
@@ -1,0 +1,5 @@
+### (Federation) Removed `RebaseError::InterfaceObjectTypename` variant ([PR #8109](https://github.com/apollographql/router/pull/8109))
+
+Fixed an uncommon query planning error, "Cannot add selection of field `X` to selection set of parent type `Y` that is potentially an interface object type at runtime". Although fetching `__typename` selections from interface object types are unnecessary, it is difficult to avoid them in all cases and the effect of having those selections in query plans is benign. Thus, the error variant and the check for the error have been removed.
+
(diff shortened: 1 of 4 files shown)


sachindshinde (approved):
This is fine given our current limitations of interface object handling; for folks who may not know why we need to do this, see [this comment from a previous PR](https://github.com/apollographql/router/pull/6250#issuecomment-2471729100).

Note that this change cannot be mirrored to the JS codebase as-is, as the JS codebase still implements the fragment-reuse optimization which uses `errorIfCannotRebase: false`, and that codepath needs `__typename` to be "skipped" to prevent the issue described in [this federation repo PR](https://github.com/apollographql/federation/pull/2886). This fragment-reuse codepath was removed from the Rust codebase, so it's fine here.

